Issue #2948, opened 2026-08-13
flaky e2e: slash-command-menu inline menu shows the 命令 group for a non-leading slash


Astro-Han:
`apps/desktop/e2e/slash-command-menu.spec.ts:87` — *offers commands only for the first token and keeps explicit Skill queries separate* — fails intermittently on `main`.

## Failing assertion

```
apps/desktop/e2e/slash-command-menu.spec.ts:106
  await expect(inlineMenu.getByRole('group', { name: '命令' })).toHaveCount(0);

  Error: expect(locator).toHaveCount(expected) failed
         - unexpected value "1"
```

The composer holds `explain /` at that point. A slash that is not the first token must not open the command group — only the Skills group. Intermittently the command group is present and stays present for the whole polling window.

## Reproduction

On `main` at `8925d4353`, with no local changes:

```bash
npm --workspace @maka/desktop run build:with-deps
npm --workspace @maka/desktop exec -- playwright test --config e2e/playwright.config.ts e2e/slash-command-menu.spec.ts:87 --repeat-each=10 --workers=1
```

Two runs of 10 on that commit: **6 failures in 20**. It also fails from a full-suite run, so it is not an artifact of `--repeat-each`.

## Not caused by a recent change

Observed while verifying an unrelated UI branch. I checked out `main` itself and reproduced at the rate above, on a code path that branch does not touch, so this predates it. The CI failure that `main` showed earlier was a different spec (`quote-selection.spec.ts`) and is being fixed separately in #2946.

## Why it matters

It is the only known flake left in the desktop e2e suite, so it will keep turning `main` and unrelated PRs red at roughly a 1-in-3 rate per run of this spec, which trains people to re-run rather than read failures.

## Starting points

Whether this is a test-timing problem or a real product bug is still open — worth settling first:

- If the command group is genuinely rendered for `explain /`, the defect is in whatever decides the inline menu's groups from the composer text, and the assertion is right to fail.
- If it is a stale render from the preceding `composer.fill('seed session')` / turn-completion step, the test needs to wait for the menu to settle for the new query rather than asserting on whatever is mounted at that instant.

A trace from a failing run is written to `apps/desktop/e2e/test-results/slash-command-menu-offers--f3f8e-icit-Skill-queries-separate/trace.zip`.

1625567290:
[...]

PR #2950, opened 2026-08-13
test(ui): give the answer-identity guards teeth

Astro-Han:
## Summary

Follow-up to #2946. Three test-layer findings from an adversarial review of that PR; no product code changes.

**The regression tests guarded only half the fix.** Both identity assertions read `container.querySelector('.maka-assistant-answer')` — the segment-level article, which pins the key half of the fix. The merged `AssistantAnswerBubble` renders a descendant carrying `.maka-chat-message-bubble-assistant`, and that is the node a text Selection actually lives in. Splitting the bubble back into separate streaming and historical component types remounts it while the article stays identical, so every test stayed green. The narrowed e2e now deliberately selects from a settled answer, so it does not cover that transition either. Test 1 now retains and asserts the bubble node as well.

**The steered-turn docstring overclaimed.** It said the test catches "an identity that is constant across segments — the sentinel-vs-real-id collision". It cannot: React reconciles duplicate-key siblings of the same component type by position, so `isSameNode` still holds and regressing the key to the collision-prone form leaves all three tests green. Corrected to state what the test really pins — two segments exist and each keeps its element across a settle.

**The `test.fixme` could go green with the bug fully present.** It asserted `.maka-quote-actions` is visible, and that state is reachable from two worlds: the Selection survived, or the quote hook's 350ms read won the race against the stream close and left a stale bar standing over text whose highlight the browser had already erased — verbatim the outcome the comment above the test calls unacceptable. Instrumented runs reproduced the underlying bug 7/7 while the body passed 6/7. It now asserts the contract that is actually broken: wait for the stream to close, then require `window.getSelection()` to still be uncollapsed. With `retries: 0`, un-fixme-ing the old body would have yielded either a false "gap closed" or a ~1-in-6 CI flake.

Refs #2946

## Verification

- `npm --workspace @maka/ui run test` — 136 passing, 0 failing.
- **Teeth check for finding 1 (required, not just asserted):** temporarily split the bubble into a distinct component type at the `TurnTimelineEntry` call site, rebuilt, re-ran — the new bubble assertion fails (`the answer bubble survives the turn settling`) while the pre-existing article assertion still passes, confirming the old test could not see the split. Reverted.
- **Teeth check for finding 3:** temporarily flipped `test.fixme` to `test`, ran `playwright test e2e/quote-selection.spec.ts --repeat-each=3` against a fresh `build:with-deps` — the streaming-drag test failed 3/3, every failure at the post-close poll (not the pre-close one), i.e. the Selection is present right after `mouse.up()` and gone once the stream closes. The pointer-capture test passed 3/3. Restored to `test.fixme`.
- `npm --workspace @maka/ui run typecheck`, `npm run format:check` — clean.
- Not run: repository-wide suites (left to CI). `apps/desktop/e2e/slash-command-menu.spec.ts` is a known pre-existing flake tracked as #2948 and was not exercised here.

## Review focus

Whether the two teeth checks are the right falsification for each guard — a test that stays green under the regression it names is worse than no test.

## Checklist

- [x] No

---
Generated-by: Claude Code (Opus 5). Authored under human direction; the human contributor reviews the final diff and owns the merge decision.

### DIFF
--- a/apps/desktop/e2e/quote-selection.spec.ts
+++ b/apps/desktop/e2e/quote-selection.spec.ts
@@ -104,5 +104,19 @@ test.fixme('a drag begun while the answer streams still offers a quote', async (
   await page.mouse.move(bounds.x + bounds.width - 2, y, { steps: 5 });
   await page.mouse.up();
 
-  await expect(page.locator('.maka-quote-actions')).toBeVisible();
+  // Assert the contract that is actually broken, not the quote bar: the bar is
+  // also reachable while the gap is open, because the hook's 350ms read can win
+  // the race against the stream close and leave a bar standing over a Selection
+  // the browser has already erased. Wait for the close, then require the
+  // Selection to still be there — that is the state the quote hook needs and
+  // the one the fragment rebuild destroys.
+  await expect
+    .poll(() => page.evaluate(() => window.getSelection()?.isCollapsed === false))
+    .toBe(true);
+  await expect(page.getByRole('button', { name: '重新生成' })).toHaveCount(1, {
+    timeout: 20_000,
+  });
+  await expect
+    .poll(() => page.evaluate(() => window.getSelection()?.isCollapsed === false))
+    .toBe(true);
 });
--- a/packages/ui/src/__tests__/chat-turn-answer-identity.test.tsx
+++ b/packages/ui/src/__tests__/chat-turn-answer-identity.test.tsx
@@ -93,24 +93,39 @@ const RUNNING_TOOL: TurnTimelineItem = {
  * The transition here is the real one from `timeline-fold.ts`: a run's last
  * tools group is projected away, so the Processing block dissolves and the
  * leading entry stops being a fold.
+ *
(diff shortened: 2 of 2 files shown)


Astro-Han:
**Fast-path rationale**

Test-layer only, two files, no product code — none of the protected areas, no user-visible behavior, and the reworked test stays `test.fixme` so it never enters the CI gate. Reverting returns the tree to the post-#2946 state.

Tests are a fast-path candidate rather than a free pass, so each guard was falsified rather than assumed:

- Split `AssistantAnswerBubble` back into two component types — the new bubble assertion fails, the pre-existing article assertion still passes. That gap is what this closes.
- Un-fixme'd the streaming-drag test, `--repeat-each=3` — 3/3 red, all at the post-close poll, so it separates "nothing was selected" from "the close destroyed the Selection".
- Finding 2 is a docstring fix; the point is that the test *cannot* catch what it claimed.

All required checks green. The fast path waives independent review, not the contributor of record's own review of the final diff.

PR #2566, opened 2026-08-09
fix(ui): separate session row actions from navigation

me2seeks:
## Summary

Session rows rendered their trailing action menu inside the row navigation button, producing invalid nested-button markup and a React hydration warning.

This keeps the row navigation and `MoreMenu` as sibling controls while preserving the existing trailing column, status indicators, rename focus handoff, and pending-action guard. The Storybook smoke contract now checks both that every populated row has its sibling action control and that no row contains nested buttons.

Fixes #2510

<details>
<summary>中文说明</summary>

会话侧栏此前把操作菜单按钮放在整行导航按钮内部，形成无效的嵌套 `button`，并在 Desktop 启动时触发 React hydration 警告。

本 PR 将行导航与 `MoreMenu` 调整为同级控件，同时保留原有的尾部列、状态提示、重命名焦点交接和异步操作防重入逻辑。Storybook smoke 也会验证每个会话行都使用同级操作控件，且不再出现嵌套按钮。

</details>

## Verification

- `npm --workspace @maka/ui test` — 263/263 pass
- `npm --workspace @maka/desktop run typecheck:stories`
- `node --test scripts/storybook-visual-smoke.test.mjs`
- `npx biome check` on the four changed files
- Rechecked the populated AppShell session sidebar in the Default Layout and Project Groups stories; row geometry and menu placement are unchanged.

## Checklist

- [x] Tests cover the change and fail without it
- [x] Lint, format, typecheck and the affected suites pass locally

Does this PR entail a change in behavior?

- [x] Yes — described under Summary above
- [ ] No

PR #2951, opened 2026-08-13
fix(ui): separate project row actions from navigation

### LINKED ISSUE #2360: Sidebar rows nest a <button> inside a <button>
Session and project rows in the sidebar render an interactive control inside the row's own button element. React reports it on every render of the session list:

```
In HTML, <button> cannot be a descendant of <button>.
```

The nesting comes from `SideNavItem`'s API, not from a misuse. Its `endContent` is rendered inside the element that carries the row's own interaction (`SideNavItem.js:345`, within `NavItemElement`'s children), so any interactive `endContent` lands inside the row button:

```
SessionNavRow → SideNavItem → NavItemElement → <button>
                                                 └─ <span> (endContent)
                                                      └─ MoreMenu → Button → <button>
```

Both call sites are affected: session rows (`session-history-list.tsx:668`) and project rows (`session-history-list.tsx:585`), each passing a `MoreMenu` for row actions.

We already carry a mitigation. `EndContentHitTarget` (`session-history-list.tsx:371`) wraps the trailing controls and stops `click`, `pointerdown` and `keydown` from reaching the row — its comment reads *"Keeps trailing controls from activating the parent SideNavItem button"*. That suppresses the misfire, but the invalid nesting remains, and it is not only console noise:

- Nested interactive elements have no defined activation behaviour; browsers and AT resolve them inconsistently.
- The row and its menu are separate tab stops inside one control, which is what WCAG 4.1.2 is about.
- `stopPropagation` on `keydown` is a blunt instrument — it also blocks keys the row legitimately handles.

Astryx solved the same shape elsewhere in 0.3.0. `CheckboxList` moved to `useClickableContainer`: the row becomes a plain container that delegates surface clicks to the one control inside it, which stays the single tab stop. That release added `interactiveRef` to `Item`/`ListItem` specifically to support this, replacing an internal invisible row button. `SideNavItem` has not adopted the pattern.

Two directions, not exclusive:

1. **Upstream** — ask for `useClickableContainer`/`interactiveRef` on `SideNavItem`, so `endContent` can hold controls without nesting. This is the fix that removes the class of bug.
2. **Local** — stop routing interactive `endContent` through `SideNavItem`; render `MoreMenu` as a sibling of the row and position it over the row's trailing edge. Removes the nesting today and lets `EndContentHitTarget` go with it.

Reproduce: run Storybook, open `Product/Shell Official AppShell → Native Conversation`, and read the console — the error fires once per session row.

Found while working on #2358, which does not touch this file.

MoonOld:
## Summary

- move the project `MoreMenu` out of `SideNavItem.endContent`, so project navigation and row actions render as sibling buttons instead of nested interactive controls
- keep unavailable state and session-count metadata inside the navigation button, with a reserved trailing slot that preserves the existing layout for collapsible and empty projects
- add a focused project-row DOM regression test covering sibling controls, collapsible ARIA wiring, and the absence of nested buttons

This completes the remaining project-row half of the issue; the session-row half was fixed in #2566. It intentionally keeps the change local to Maka rather than expanding the Astryx `SideNavItem` API in this PR.

Fixes #2360

## Verification

- `npm --workspace @maka/ui run test` — 137 passed
- `npx biome check packages/ui/src/session-history-list.tsx packages/ui/src/__tests__/session-history-row-actions.test.tsx apps/desktop/src/renderer/styles/sidebar.css`
- `npm --workspace @maka/desktop run smoke:storybook` — 126 stories passed
- Storybook `Native Conversation`, grouped by project: 11 project rows, 0 nested buttons, no console errors; collapse, unavailable/relink, archived/restore, and rename focus-return paths verified

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/src/renderer/styles/sidebar.css
+++ b/apps/desktop/src/renderer/styles/sidebar.css
@@ -189,7 +189,8 @@
   -webkit-app-region: no-drag;
 }
 
-.maka-session-row {
+.maka-session-row,
+.maka-project-row {
   position: relative;
   min-width: 0;
   -webkit-app-region: no-drag;
@@ -259,7 +260,8 @@
 
 /* Fixed trailing column so StatusDot ↔ MoreMenu swaps do not reflow the label
  * (hover used to steal ~28px and shove the title left). */
-.maka-session-row-trailing {
+.maka-session-row-trailing,
+.maka-project-row-trailing {
   display: inline-flex;
   align-items: center;
   justify-content: center;
@@ -268,10 +270,11 @@
   min-height: var(--h-control-sm, 24px);
 }
 
-/* SideNavItem owns the row button and only supports non-interactive
- * endContent. The action menu occupies the reserved trailing slot visually,
- * but remains its sibling in the DOM so neither control contains the other. */
-.maka-session-row-action {
+/* SideNavItem owns each row button and only supports non-interactive
+ * endContent. Action menus occupy a reserved trailing slot visually, but stay
+ * siblings in the DOM so neither control contains the other. */
+.maka-session-row-action,
+.maka-project-row-action {
   position: absolute;
   inset-block-start: calc(
     (var(--size-element-md) - var(--size-element-sm)) / 2
@@ -282,6 +285,14 @@
   align-items: center;
   justify-content: center;
   width: var(--h-control-sm, 24px);
+}
+
+/* A collapsible project keeps Astryx's 24px disclosure after endContent. Match
+ * the placeholder's original slot immediately before that disclosure: row
+ * gutter + disclosure + SideNavItem gap. Empty projects have no disclosure and
+ * use the default trailing inset above. */
+.maka-project-row-action[data-position="before-disclosure"] {
+  inset-inline-end: calc(var(--space-2) + var(--space-6) + var(--space-2));
 }
 
 .maka-session-row-trailing-spacer {
--- a/apps/desktop/src/renderer/styles/sidebar.css
+++ b/apps/desktop/src/renderer/styles/sidebar.css
@@ -260,8 +260,7 @@
 
 /* Fixed trailing column so StatusDot ↔ MoreMenu swaps do not reflow the label
  * (hover used to steal ~28px and shove the title left). */
-.maka-session-row-trailing,
-.maka-project-row-trailing {
+.maka-session-row-trailing {
   display: inline-flex;
   align-items: center;
   justify-content: center;
@@ -273,8 +272,7 @@
 /* SideNavItem owns each row button and only supports non-interactive
  * endContent. Action menus occupy a reserved trailing slot visually, but stay
  * siblings in the DOM so neither control contains the other. */
-.maka-session-row-action,
-.maka-project-row-action {
+.maka-session-row-action {
   position: absolute;
   inset-block-start: calc(
     (var(--size-element-md) - var(--size-element-sm)) / 2
@@ -291,8 +289,27 @@
  * the placeholder's original slot immediately before that disclosure: row
  * gutter + disclosure + SideNavItem gap. Empty projects have no disclosure and
  * use the default trailing inset above. */
-.maka-project-row-action[data-position="before-disclosure"] {
+.maka-project-row
+  > .maka-session-row-action[data-position="before-disclosure"] {
   inset-inline-end: calc(var(--space-2) + var(--space-6) + var(--space-2));
+}
+
+/* The sibling menu is visually part of the project header, but Astryx's
+ * hover/pressed paint lives on the navigation button itself. Restore that
+ * feedback only while the direct project action is targeted; a plain wrapper
+ * :hover selector would also light the header over nested session rows. */
+@media (hover: hover) {
+  .maka-project-row:has(> .maka-session-row-action:hover)
+    > div
+    > .astryx-side-nav-item {
+    background-color: var(--color-overlay-hover);
+  }
+}
+
+.maka-project-row:has(> .maka-session-row-action button:active)
+  > div
+  > .astryx-side-nav-item {
+  background-color: var(--color-overlay-pressed);
 }
 
 .maka-session-row-trailing-spacer {
--- a/packages/ui/src/__tests__/session-history-row-actions.test.tsx
+++ b/packages/ui/src/__tests__/session-history-row-actions.test.tsx
@@ -1,9 +1,15 @@
 import assert from 'node:assert/strict';
 import test from 'node:test';
+import { parseHTML } from 'linkedom';
 import { renderToStaticMarkup } from 'react-dom/server';
+import type { ProjectRecord } from '@maka/core/project';
 import type { SessionSummary } from '@maka/core/session';
 import { LocaleProvider } from '../locale-context.js';
-import { SessionHistoryList, type SessionRowActions } from '../session-history-list.js';
+import {
+  SessionHistoryList,
+  type ProjectRowActions,
+  type SessionRowActions,
+} from '../session-history-list.js';
 
 const session: SessionSummary = {
   id: 'session-1',
@@ -28,6 +34,21 @@ const rowActions: SessionRowActions = {
   onDelete: () => undefined,
 };
 
+const project: ProjectRecord = {
+  id: 'project-1',
+  name: 'Maka',
+  locations: [{ path: '/workspace/maka', isWorktree: false }],
+  available: true,
+  preferredPath: '/workspace/maka',
+};
+
+const projectActions: ProjectRowActions = {
+  onNew: () => undefined,
+  onRename: () => undefined,
+  onArchive: () => undefined,
+  onRestore: () => undefined,
+};
+
 test('renders session navigation and row actions as sibling controls', () => {
   const markup = renderToStaticMarkup(
     <LocaleProvider locale="en">
@@ -41,5 +62,47 @@ test('renders session navigation and row actions as sibling controls', () => {
 
   assert.equal((markup.match(/<button\b/g) ?? []).length, 2);
   assert.match(markup, /class="maka-session-row-action"/);
+  assert.doesNotMatch(markup, /<button\b(?:(?!<\/button>)[\s\S])*<button\b/);
+});
+
+test('renders collapsible project navigation and row actions as sibling controls', () => {
+  const markup = renderToStaticMarkup(
+    <LocaleProvider locale="en">
+      <SessionHistoryList
+        sessions={[session]}
+        groups={[{ id: project.id, label: project.name, project, sessions: [session] }]}
+        groupVariant="project"
+        projectActions={projectActions}
+        onSelectSession={() => undefined}
+      />
+    </LocaleProvider>,
+  );
+
+  const { document } = parseHTML(markup);
+  const projectRow = document.querySelector('.maka-project-row');
+  const action = document.querySelector<HTMLButtonElement>(
+    'button[aria-label="Maka project actions"]',
+  );
+
+  assert.ok(projectRow);
+  assert.ok(action);
+  const navigation = projectRow.querySelector<HTMLButtonElement>('button[aria-expanded="true"]');
+  const metadata = projectRow.querySelector('.maka-project-item-end');
+  const controlledGroupId = navigation?.getAttribute('aria-controls');
+  const controlledGroup = controlledGroupId
+    ? document.getElementById(controlledGroupId)
+    : null;
+
+  assert.ok(navigation);
+  assert.ok(metadata);
+  assert.ok(controlledGroup);
+  assert.equal(navigation.contains(metadata), true);
+  assert.equal(navigation.contains(action), false);
+  assert.equal(metadata.textContent, '1');
+  assert.equal(controlledGroup.getAttribute('aria-hidden'), 'false');
+  assert.equal(action.parentElement?.className, 'maka-project-row-action');
+  assert.equal(action.parentElement?.getAttribute('data-position'), 'before-disclosure');
+  assert.equal(action.parentElement?.parentElement, projectRow);
+  assert.equal(projectRow.querySelector('button button'), null);
   assert.doesNotMatch(markup, /<button\b(?:(?!<\/button>)[\s\S])*<button\b/);
 });
--- a/packages/ui/src/__tests__/session-history-row-actions.test.tsx
+++ b/packages/ui/src/__tests__/session-history-row-actions.test.tsx
@@ -86,7 +86,9 @@ test('renders collapsible project navigation and row actions as sibling controls
 
   assert.ok(projectRow);
   assert.ok(action);
-  const navigation = projectRow.querySelector<HTMLButtonElement>('button[aria-expanded="true"]');
+  const navigation = projectRow.querySelector<HTMLButtonElement>(
+    ':scope > div > button[aria-controls]',
+  );
   const metadata = projectRow.querySelector('.maka-project-item-end');
   const controlledGroupId = navigation?.getAttribute('aria-controls');
   const controlledGroup = controlledGroupId
@@ -100,9 +102,8 @@ test('renders collapsible project navigation and row actions as sibling controls
   assert.equal(navigation.contains(action), false);
   assert.equal(metadata.textContent, '1');
   assert.equal(controlledGroup.getAttribute('aria-hidden'), 'false');
-  assert.equal(action.parentElement?.className, 'maka-project-row-action');
-  assert.equal(action.parentElement?.getAttribute('data-position'), 'before-disclosure');
-  assert.equal(action.parentElement?.parentElement, projectRow);
-  assert.equal(projectRow.querySelector('button button'), null);
+  const projectButtons = [...projectRow.querySelectorAll('button')];
+  assert.equal(projectButtons[0], action);
+  assert.equal(projectButtons[1], navigation);
   assert.doesNotMatch(markup, /<button\b(?:(?!<\/button>)[\s\S])*<button\b/);
 });
--- a/packages/ui/src/session-history-list.tsx
+++ b/packages/ui/src/session-history-list.tsx
@@ -372,7 +372,17 @@ function ProjectNavRow(props: {
   const hasActions = props.project !== undefined && props.projectActions !== undefined;
   return (
     <div data-project-id={props.groupKey} className="maka-project-row">
+      {props.project && props.projectActions ? (
+        <ProjectItemActions
+          key="actions"
+          project={props.project}
+          actions={props.projectActions}
+          onStartRename={props.onStartRename}
+          position={hasSessions ? 'before-disclosure' : 'trailing'}
+        />
+      ) : null}
       <SideNavItem
+        key="navigation"
         label={props.label}
         icon={FolderOpen}
         collapsible={hasSessions ? { defaultIsCollapsed: false } : undefined}
@@ -389,14 +399,6 @@ function ProjectNavRow(props: {
           <VStack gap={0.5}>{props.sessions.map((session) => props.renderSession(session))}</VStack>
         ) : undefined}
       </SideNavItem>
-      {props.project && props.projectActions ? (
-        <ProjectItemActions
-          project={props.project}
-          actions={props.projectActions}
-          onStartRename={props.onStartRename}
-          position={hasSessions ? 'before-disclosure' : 'trailing'}
-        />
-      ) : null}
     </div>
   );
 }
@@ -477,7 +479,7 @@ function ProjectItemMeta(props: {
       )}
       <Badge variant="neutral" label={props.sessionCount} />
       {props.reserveAction ? (
-        <span className="maka-project-row-trailing" aria-hidden="true" />
+        <span className="maka-session-row-trailing" aria-hidden="true" />
       ) : null}
     </span>
   );
@@ -569,12 +571,7 @@ function ProjectItemActions(props: {
       ];
 
   return (
-    <span
-      className="maka-project-row-action"
-      data-position={props.position}
-      ref={trailingRef}
-      onKeyDown={(event) => event.stopPropagation()}
-    >
+    <span className="maka-session-row-action" data-position={props.position} ref={trailingRef}>
       <MoreMenu
         size="sm"
         label={copy.projectActionsAriaLabel(project.name)}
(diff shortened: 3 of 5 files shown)


MoonOld:
Hi @Astro-Han, could you review this when you have a chance? This completes the project-row half of #2360 using the same sibling-control pattern as #2566. The branch is rebased onto current main, and all CI checks are green. Thanks!

packages/ui/src/session-history-list.tsx
```diff
@@ -389,29 +389,18 @@ function ProjectNavRow(props: {
           <VStack gap={0.5}>{props.sessions.map((session) => props.renderSession(session))}</VStack>
         ) : undefined}
       </SideNavItem>
+      {props.project && props.projectActions ? (
+        <ProjectItemActions
```
Astro-Han: **P1** — rendering the actions after `</SideNavItem>` puts them after the collapsible subtree in the DOM: `SideNavItem` returns `<div root>{itemElement}{childrenGroup}</div>` and the sessions live in `childrenGroup`.

So for an expanded project — the default, `defaultIsCollapsed: false` — Tab goes row button -> every session row *and every session's own `...`* -> finally the project's `...`, which sits visually on the project row. The more sessions a project has, the further the focus position drifts from the visual one. That trades WCAG 4.1.2 for 2.4.3, which is the class of problem #2360 is about. Collapsed projects are fine because the subtree is `inert`, and session rows are fine because that `SideNavItem` has no children — this is the one place the #2566 template does not transfer.

Rendering `ProjectItemActions` as the *first* child of `.maka-project-row` fixes it: absolute positioning makes DOM order visually free, and Tab becomes `...` -> row -> sessions, which is at least adjacent. Worth an assertion that walks `querySelectorAll('button')` in document order — the current `parentElement.parentElement === projectRow` check says nothing about order.

apps/desktop/src/renderer/styles/sidebar.css
```diff
@@ -268,10 +270,11 @@
...
 
-/* SideNavItem owns the row button and only supports non-interactive
- * endContent. The action menu occupies the reserved trailing slot visually,
- * but remains its sibling in the DOM so neither control contains the other. */
-.maka-session-row-action {
+/* SideNavItem owns each row button and only supports non-interactive
+ * endContent. Action menus occupy a reserved trailing slot visually, but stay
+ * siblings in the DOM so neither control contains the other. */
+.maka-session-row-action,
+.maka-project-row-action {
```
Astro-Han: **P3** — the overlay is not a descendant of the row button, so moving the pointer onto the `...` drops the row's `:hover` background (Astryx's `:hover` rule lives on the button) and the row visibly de-highlights right as you reach for its menu. Same for `:active`.

Inherited from #2566 rather than introduced here, but this PR spreads it to project rows. `.maka-project-row:hover > div > [nav item]` with a child combinator restores it without leaking into nested session rows — `sidebar.css:230` already uses that descendant shape.

packages/ui/src/session-history-list.tsx
```diff
@@ -514,77 +522,74 @@ function ProjectItemEndContent(props: {
...
-          }}
-          items={menuItems}
-        />
-      )}
-    </EndContentHitTarget>
+    <span
+      className="maka-project-row-action"
+      data-position={props.position}
+      ref={trailingRef}
+      onKeyDown={(event) => event.stopPropagation()}
```
Astro-Han: **P3** — dead on project rows: `handleListKeyDown` (line 99) only acts when `closest('[data-maka-contract="session-row"], [data-session-id]')` matches, and `.maka-project-row` carries `data-project-id`. It came from `EndContentHitTarget`, whose reason — don't activate the parent row button — disappeared when the menu stopped being a descendant, so it can just go.

Worth knowing that the same line on session rows (701, untouched here) is *not* dead: focus the `...` trigger without opening it, press Delete, and the shortcut silently does nothing while the same key on the row button deletes. That is the blunt instrument the issue names; out of scope here, flagging it so it does not get lost.

apps/desktop/src/renderer/styles/sidebar.css
```diff
@@ -259,7 +260,8 @@
 
 /* Fixed trailing column so StatusDot ↔ MoreMenu swaps do not reflow the label
  * (hover used to steal ~28px and shove the title left). */
-.maka-session-row-trailing {
+.maka-session-row-trailing,
+.maka-project-row-trailing {
```
Astro-Han: **P3** — `.maka-project-row-trailing` and `.maka-project-row-action` are declaration-for-declaration identical to their `-session-` counterparts; only the `[data-position]` variant is new. `ProjectItemMeta` already reuses `maka-session-row-end`, so reusing the two existing classes drops this CSS addition from 17 lines to about 8.

If the "session" name grates, renaming both to `.maka-list-row-*` is the cleaner move — the file already has `.maka-list-row-stale-pill` as precedent — but that belongs with the component merge, not here.

packages/ui/src/__tests__/session-history-row-actions.test.tsx
```diff
@@ -43,3 +64,45 @@ test('renders session navigation and row actions as sibling controls', () => {
...
+    : null;
+
+  assert.ok(navigation);
+  assert.ok(metadata);
+  assert.ok(controlledGroup);
+  assert.equal(navigation.contains(metadata), true);
+  assert.equal(navigation.contains(action), false);
+  assert.equal(metadata.textContent, '1');
+  assert.equal(controlledGroup.getAttribute('aria-hidden'), 'false');
+  assert.equal(action.parentElement?.className, 'maka-project-row-action');
```
Astro-Han: **P3** — these four are insensitive to the regression. The class name and `data-position` are CSS hooks that go red on any harmless restyle; `parentElement.parentElement === projectRow` is implied by the containment check plus the markup check; and `querySelector('button button')` duplicates the markup assertion while depending on the SSR parser preserving invalid nesting — a real browser parser flattens it, so a parser swap could turn it silently green.

The markup regex is the parser-independent one. Keeping it plus `navigation.contains(action) === false` covers the regression with two assertions instead of five.

packages/ui/src/__tests__/session-history-row-actions.test.tsx
```diff
@@ -43,3 +64,45 @@ test('renders session navigation and row actions as sibling controls', () => {
...
+
+  const { document } = parseHTML(markup);
+  const projectRow = document.querySelector('.maka-project-row');
+  const action = document.querySelector<HTMLButtonElement>(
+    'button[aria-label="Maka project actions"]',
+  );
+
+  assert.ok(projectRow);
+  assert.ok(action);
+  const navigation = projectRow.querySelector<HTMLButtonElement>('button[aria-expanded="true"]');
```
Astro-Han: **P3** — `button[aria-expanded="true"]` selects the row today only because the MoreMenu trigger's `aria-expanded` is `"false"` while closed. Any change to the menu's default state re-points this selector at the wrong control. `aria-controls` or the row's accessible name is stable.

Astro-Han (commented):
The direction is right, and `EndContentHitTarget` is genuinely gone — no remaining callers anywhere in the repo.

**One P1 to fix before this merges**: for an expanded project the row's `...` menu now lands after the entire session subtree in Tab order. Details inline on the render site.

## Root cause, and why it shows up here and not in #2566

`SideNavItem` keeps the row's interaction on the row element *and* renders both `endContent` and the collapsible subtree inside its own root (`SideNavItem.js`: `<div root>{itemElement}{childrenGroup}</div>`). Any trailing control therefore has to be laid *over* the row from outside it, and every non-cosmetic finding below is the price of that overlay: the broken tab order, the hardcoded Astryx geometry, the reserved placeholder slot, the lost hover, the leftover `stopPropagation`. #2566 paid that price once on session rows, where the row has no subtree and the tab-order cost was zero. Applying the same template to a row that *does* have a subtree is exactly where it breaks.

## The class of bug is closed today, and the summary's claim is accurate

Every remaining `SideNavItem endContent` in the repo is non-interactive: `session-sidebar-nav.tsx:58` a `<kbd>`, `settings-surface.tsx:308` a `Badge`, and this file's now-pure meta. `SideNavSection`'s `endContent` is safe because its header is a `<div>`.

But nothing enforces it. `endContent?: ReactNode` carries no constraint, and the only automated guard is two SSR unit tests. The Storybook smoke cannot help, because no story renders project grouping at all — `app-shell.stories.tsx:301` defaults to `'conversation'` and nothing in the repo passes `'project'`. The smoke run fails on `console.error`, which is precisely how this bug announced itself, so a two-line story export with `initialViewMode: 'project'` turns it into a real regression gate. (#2582 removed the assertion layer that used to cover this.)

## Consider unifying the two action components rather than shrinking this diff

`ProjectItemActions` and `SessionItemActions` are now ~55 lines of line-for-line duplication: `trailingRef`, `menuOpen`, `pendingAction` + `pendingActionRef`, `mountedRef`, the cleanup effect, `runXAction`, the wrapper span, and the `onOpenChange` intent flush. Only `label`, `items` and the position differ. A shared `RowActionsMenu({ label, position, buildItems })` collapses them, absorbs the write-only `ProjectRowActionId` / `SessionRowActionId` pair along the way, and means a third caller inherits the fix instead of repeating the template a third time.

Looking at the project row alone this diff is already close to minimal — the 40 reindented lines in `menuItems` are forced churn from lifting the `project && actions` guard to the call site, not avoidable rewrite. The only way to make it meaningfully smaller is to merge it with the session row.

## Staying local is the right call, and worth recording upstream anyway

Astryx is a pinned external dependency (`0.3.0` in three places) and the local patch file already runs 779 lines across 8 components — changing `SideNavItem`'s DOM structure is the riskiest kind of entry to add there. Astryx also treats the sibling shape as its own sanctioned answer: the split-action path in `SideNavItem.js` is commented "This avoids nesting interactive elements", it just only triggers when a primary action and `collapsible` are both set. So this is not working around the design system.

The end state is still the upstream one. `useClickableContainer` / `interactiveRef` on `SideNavItem` would delete both overlays, both CSS blocks, the placeholder slot and every `stopPropagation`, and make focus order correct by construction. That is a cross-repo component refactor against three lines here, so not now — but the price of staying local is real and worth naming: the `before-disclosure` offset mirrors three Astryx internals (nav item `padding-inline`, `gap`, chevron box) that live behind compiled StyleX classes and are not public contract. The arithmetic is right today — 8 + 24 + 8 lands exactly on the placeholder — but an Astryx minor that changes the gap or the chevron, or anyone passing `size="sm"` to this `SideNavItem`, drifts the menu off its slot with nothing to catch it. Please open a separate upstream-tracking issue rather than letting the case die when `Fixes #2360` closes.

## Tests have teeth, and are heavier than they need to be

Reverting the fix does fail them — verified by compiling the mutant. But four of the nine new assertions are insensitive to that mutation, and one of the three that do kill it relies on the SSR parser preserving invalid nesting, which a real browser parser flattens. Trimming is noted inline.

The larger gap is that of the issue's three stated reasons, only "no longer nested" is covered: real tab order, two independently focusable controls, and whether `stopPropagation` still swallows list-level keys are all untested — as are the empty-project (`position: 'trailing'`, no disclosure) and archived (restore-only menu) branches this PR newly introduces. `packages/ui/stories/session-list-panel.tsx`'s `ProjectGroups` story already has an unavailable project and an empty one, so no new fixture is needed. A single Playwright spec — Tab reaches the row, Tab reaches the menu, Enter opens it, Escape returns focus, Enter on the row only toggles — is the honest home for the interaction half; `apps/desktop/e2e/` has no sidebar project-row coverage today.

<details>
<summary>中文</summary>

方向是对的，`EndContentHitTarget` 也确实删干净了——全仓无残留调用点。

**有一条 P1 建议在合并前修**：展开状态下，项目行的 `...` 菜单在 Tab 顺序里落到了整个会话子树之后。细节见渲染处的行内评论。

## 根因，以及它为什么在这里出现而 #2566 没有

`SideNavItem` 把行的交互留在行元素上，**同时**把 `endContent` 和可折叠子树都渲染在自己的 root 内（`SideNavItem.js`：`<div root>{itemElement}{childrenGroup}</div>`）。因此任何尾部控件都只能从行外面**盖**上去，而下面每一条非装饰性的 finding 都是这个覆盖层的价钱：断裂的 Tab 顺序、硬编码的 Astryx 几何、预留占位槽、丢失的 hover、残留的 `stopPropagation`。#2566 在会话行上付过一次这个价钱，那里的行没有子树，Tab 顺序的代价为零。把同一模板套用到**确实有子树**的行上，正是它失效的地方。

## 这类 bug 目前是关闭的，summary 的说法准确

仓库里剩余的每一处 `SideNavItem endContent` 都是非交互的：`session-sidebar-nav.tsx:58` 是 `<kbd>`，`settings-surface.tsx:308` 是 `Badge`，本文件的 meta 现在也是纯展示。`SideNavSection` 的 `endContent` 安全，因为它的 header 是 `<div>`。

但没有任何机制约束这一点。`endContent?: ReactNode` 不带任何约束，唯一的自动化守卫是两个 SSR 单测。Storybook smoke 帮不上忙，因为没有任何 story 渲染项目分组——`app-shell.stories.tsx:301` 默认 `'conversation'`，全仓没有一处传 `'project'`。smoke 以 `console.error` 判失败，而这个 bug 正是这样暴露出来的，所以一个两行的 `initialViewMode: 'project'` story export 就能把它变成真正的回归关卡。（#2582 移除了原先承担这件事的断言层。）

## 建议合并两个 action 组件，而不是缩小本 diff

`ProjectItemActions` 与 `SessionItemActions` 现在有约 55 行逐行重复：`trailingRef`、`menuOpen`、`pendingAction` + `pendingActionRef`、`mountedRef`、清理 effect、`runXAction`、wrapper span，以及 `onOpenChange` 的 intent flush。差异只有 `label`、`items` 和位置。一个共享的 `RowActionsMenu({ label, position, buildItems })` 可以把它们合并，顺带消化掉只写不读的 `ProjectRowActionId` / `SessionRowActionId`，并让第三个调用点继承修复，而不是第三次复述模板。

单看项目行，这个 diff 已接近最小——`menuItems` 那 40 行重新缩进是把 `project && actions` 守卫上提到调用方后的强制 churn，不是可以省掉的重写。要显著更小只能与会话行合并。

## 保持本地修法是对的，但仍值得记录到上游

Astryx 是 pin 死的外部依赖（三处 `0.3.0`），本地补丁文件已达 779 行、覆盖 8 个组件——改 `SideNavItem` 的 DOM 结构是其中风险最高的一类条目。而且 Astryx 自己就把兄弟节点当作官方解法：`SideNavItem.js` 里 split-action 路径的注释原文是 "This avoids nesting interactive elements"，只是它仅在同时设置 primary action 与 `collapsible` 时才触发。所以这不是绕过设计系统。

终局仍然在上游。`SideNavItem` 接入 `useClickableContainer` / `interactiveRef` 会删掉两处覆盖层、两块 CSS、占位槽和全部 `stopPropagation`，并让焦点顺序天然正确。那是一次跨仓库的组件重构，对比这里的三行改动，现在不值得——但保持本地的代价是真实的，值得点名：`before-disclosure` 的偏移复刻了三个 Astryx 内部量（nav item 的 `padding-inline`、`gap`、chevron 尺寸），它们藏在编译后的 StyleX class 里，不是公开契约。今天的算术是对的——8 + 24 + 8 精确落在占位槽上——但一次改动 gap 或 chevron 的 Astryx minor 升级，或有人给这个 `SideNavItem` 传 `size="sm"`，都会让菜单静默漂出槽位而无人拦截。请另开一个上游跟踪 issue，别让 `Fixes #2360` 关闭时把这个案子一并埋掉。

## 测试有牙齿，但比需要的重

回退修复确实会让它变红——已编译变异体验证。但新增的九条断言里有四条对该变异完全无感，而能杀死变异的三条里还有一条依赖 SSR 解析器保留非法嵌套，真实浏览器解析器会把它拍平。裁剪建议见行内。

更大的缺口是：issue 列出的三条理由里只有「不再嵌套」被覆盖，真实 Tab 顺序、两个可独立聚焦的控件、以及 `stopPropagation` 是否仍在吞掉列表级按键，全部没测；本 PR 新引入的空项目分支（`position: 'trailing'`，无 disclosure）与 archived 分支（只有 restore 的菜单）同样没测。`packages/ui/stories/session-list-panel.tsx` 的 `ProjectGroups` story 已经带了 unavailable 项目和空项目，不需要新造 fixture。交互那一半的正确归属是一条 Playwright spec——Tab 到行、Tab 到菜单、Enter 打开、Escape 回焦、行上的 Enter 只切换展开；`apps/desktop/e2e/` 目前没有任何侧边栏项目行覆盖。

</details>

apps/desktop/src/renderer/styles/sidebar.css
MoonOld: Fixed in 28aa7cf47. The project header now receives hover/pressed paint while its direct sibling action is targeted. The selector uses `:has(> .maka-session-row-action...)` plus direct-child navigation targeting, so hovering a nested session does not light the project header. I also verified both cases with a real browser pointer.

packages/ui/src/session-history-list.tsx
MoonOld: Removed in 28aa7cf47. The project action wrapper no longer stops keydown propagation; the session-row handler remains unchanged because, as noted, that path is a separate behavior and outside this PR.

apps/desktop/src/renderer/styles/sidebar.css
MoonOld: Fixed in 28aa7cf47. Project meta/actions now reuse `maka-session-row-trailing` and `maka-session-row-action`; only the project-specific before-disclosure positioning rule remains.

packages/ui/src/__tests__/session-history-row-actions.test.tsx
MoonOld: Fixed in 28aa7cf47. I removed the class/data-position/parent/nested-query assertions, kept the parser-independent markup guard and navigation/action separation, and added the document-order assertion needed for the P1 focus-order regression.

packages/ui/src/session-history-list.tsx
MoonOld: Fixed in 28aa7cf47. `ProjectItemActions` is now the first keyed child and `SideNavItem` is the second keyed child, so the order is menu → project navigation → session subtree without remounting the navigation state when actions appear. The unit test now asserts document order, and the new Electron Playwright test exercises the real Tab path menu → navigation → first session.

packages/ui/src/__tests__/session-history-row-actions.test.tsx
MoonOld: Fixed in 28aa7cf47. The unit test now selects the navigation through its direct SideNav root plus `aria-controls`, independent of the MoreMenu's open state.

MoonOld:
@Astro-Han, I addressed the P1 and all inline P3 findings in 28aa7cf47. Could you take another look when convenient?

What changed:

- rendered the project menu before the keyed SideNav root, giving menu → project navigation → sessions in DOM/Tab order;
- added the direct-action hover/pressed bridge without leaking hover from nested session rows;
- removed the dead project-row keydown propagation guard;
- reused the existing session trailing/action CSS primitives;
- trimmed the brittle SSR assertions and added the document-order assertion;
- switched the unit selector away from the menu-state-dependent `aria-expanded="true"`.

For permanent interaction coverage, I added an Electron Playwright journey that uses a real seeded project and verifies Tab order, Enter collapse/expand, `aria-hidden`/`inert`, menu isolation, Escape focus restoration, and rename-dialog focus restoration.

I reused the existing `ProjectGroups` story in `packages/ui/stories/session-list-panel.stories.tsx` (it already sets `viewMode: 'project'` and is included as `product-sidebar-session-list--project-groups`). Storybook smoke owns render/console health; Electron E2E owns the keyboard interaction contract.

I kept the project/session action-component merge out of this review fix because it would expand into the existing session-row behavior without being required to close #2360. I did record the missing upstream component seam as requested: facebook/astryx#4987.

Validation completed locally:

- `npm --workspace @maka/ui run test`: 137 passed
- Desktop typecheck: passed
- full Desktop Electron E2E: 22 passed, 1 existing conditional skip
- Storybook build: passed
- Storybook render smoke: 126 stories passed
- real-window programmatic smoke: passed
- real-browser `ProjectGroups` verification: Tab/menu/disclosure/hover behavior passed; no console errors
- Biome + `git diff --check`: passed

I also ran the root `npm test` flow. The changed UI/Desktop suites passed; the command's overall exit remained non-zero in two untouched areas: the local system Python is 3.9 and cannot parse eval's Python 3.10 `Path | None` annotation, and the current runtime-host authenticated-WebSocket test reproduces `Session model is not enabled` even when run alone. This branch has no eval/runtime-host changes; the PR CI run is the authoritative clean-environment result.

Astro-Han:
All six findings are addressed, and the E2E spec goes further than what I asked for — Tab adjacency, disclosure toggling with `aria-hidden`/`inert`, Escape returning focus, and the rename dialog's focus return are all covered. Moving the actions to the first child also made the ordering assertion fall out of the unit test naturally.

Two non-blocking follow-ups: `SessionItemActions` and `ProjectItemActions` are still ~55 lines of the same component, worth its own issue; and the upstream ask for a real `SideNavItem` action slot is still worth opening at facebook/astryx — with it, both row types could drop the absolute positioning and the mirrored geometry.

Squash-merging. Thanks for the quick turnaround.

*Review assistance: Claude Code (Opus) ran parallel correctness, simplification, test-quality and architecture passes over this PR; I verified the focus-order finding and the Astryx DOM structure by hand before filing it.*

<details>
<summary>中文</summary>

六条 findings 全部处理，E2E 还超出了我要求的范围——Tab 相邻性、开合 disclosure 时的 `aria-hidden`/`inert`、Escape 回焦、重命名对话框关闭后的回焦都覆盖了。把 actions 移到第一个子节点之后，单测里的顺序断言也自然成立了。

两条不阻塞的后续：`SessionItemActions` 与 `ProjectItemActions` 仍是同一个组件的约 55 行重复，值得单开一个 issue；`SideNavItem` 真正的 action 槽位这条上游诉求仍值得去 facebook/astryx 提——有了它，两种行都能去掉绝对定位和复刻的几何。

Squash 合并。感谢这么快的响应。

*评审协助说明：Claude Code (Opus) 对本 PR 并行跑了正确性、简化、测试质量与架构四轮审查；焦点顺序那条结论与 Astryx 的 DOM 结构，我在提出前手工核实过。*

</details>

PR #2952, opened 2026-08-13
fix(ui): collapse the composer send slot to one control

Astro-Han:
## Summary

Mid-turn the composer's send slot held two buttons: a ghost Stop and a primary Steer arrow. That is a control group standing in the position Astryx defines as one send/stop toggle (`ChatSendButton` — "a circular send/stop toggle button"), and the prop doc right above it still described the older, correct contract ("Send becomes Stop"). The second button was redundant on its own terms too: Enter already steered mid-turn, and Esc already interrupts.

The slot is one button with two states again. While a turn is in flight and the draft is empty there is nothing to submit, so it reads Stop; the moment there is a draft, handing it over is the only meaningful action in that position and the button returns to the arrow (labelled "Steer" mid-turn). Stop is not lost in that window — Esc interrupts from the input, where the hands already are.

With a single submit on the surface, `onStreamingSubmit` had nothing left to mean: what handing the draft over *does* mid-turn is the host's reading of one action, not a second control. Both hosts now fold steering into `onSend` — the shell exempts slash commands, which are control instructions rather than text for the running turn, and the side chat has none to exempt. Dropping the prop takes the composer's knowledge of steering with it.

image

## Verification

- `packages/ui` + `apps/desktop` typecheck (preload, main, renderer, storybook)
- `npm run format:check`
- New `packages/ui/src/__tests__/composer-send-toggle.test.tsx` — asserts the slot renders exactly one control (Send when idle, Stop mid-turn). It fails against the previous markup, which rendered two.
- Storybook, driven by hand: mid-turn with an empty draft the slot is a 32px circular secondary Stop; typing flips the same slot to the primary arrow with `aria-label="Steer"` and `type="submit"`; idle reads Send. No console errors.
- Not run: the desktop E2E suite (needs a full Electron build; no existing spec selects the stop button by name) — left to CI.

## Review focus

The one deliberate omission is switching to Astryx's `ChatSendButton`. It resolves `isDisabled` as `!isStopShown && isDisabled`, so a stop state can never be disabled, and we would lose `stopPending` ("Stopping…"). `IconButton` is already an Astryx primitive, so the swap would buy a circular xstyle at the cost of product semantics.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/src/renderer/app-shell.tsx
+++ b/apps/desktop/src/renderer/app-shell.tsx
@@ -1948,6 +1948,12 @@ function AppShellContent({
       revision && activeIdRef.current === revision.draftSessionId,
     );
     const slashCommand = parseDesktopSlashCommand(text);
+    // The composer has one submit; mid-turn it means steering, and this is
+    // where that reading lives. Slash commands are exempt because they are
+    // control instructions to the shell, not text for the running turn.
+    if (!slashCommand && (turnActive || activeStreamingLive)) {
+      return steerWithText(text);
+    }
     if (
       revisionSend &&
       revision &&
@@ -2139,15 +2145,6 @@ function AppShellContent({
       }
       return false;
     }
-  }
-
-  function submitWhileStreaming(
-    text: string,
-    metadata?: ComposerSendMetadata,
-  ): Promise<boolean | void> {
-    return parseDesktopSlashCommand(text)
-      ? sendWithAttachments(text, metadata)
-      : steerWithText(text);
   }
 
   const stop = createAppShellStopAction({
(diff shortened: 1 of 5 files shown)


M4n5ter (approved):
<details open>

LGTM. The single-control behavior is consistent across the composer and both hosts.

One non-blocking request: please change the PR title from `refactor(ui)` to `fix(ui)`, since this changes user-visible behavior and `CONTRIBUTING.md` reserves `refactor` for behavior-preserving changes.

</details>

<details>

整体没问题。Composer 和两个 host 的单按钮行为保持一致。

有一个不阻塞合并的请求：请将 PR 标题从 `refactor(ui)` 改为 `fix(ui)`，因为本次修改改变了用户可见行为，而 `CONTRIBUTING.md` 将 `refactor` 用于不改变行为的重构。

</details>

Astro-Han:
Thanks for the review, I've changed the title to Fix(ui). Merging now.

PR #2955, opened 2026-08-13
fix(ui): show only the stop button while streaming

Frant1cc:
## Summary
- While a turn is streaming, the composer now shows a single stop control instead of both a stop button and a steer/send button at once.
- Removes the steer send button from the running-actions row; the stop button keeps its original ghost + square-icon styling.

## Test plan
- [x] Start a turn and confirm the composer shows only the stop (square) button while streaming
- [x] Confirm clicking stop still aborts the turn
- [x] Confirm the send (arrow-up) button returns once streaming ends


Frant1cc:
Closing — this was already fixed by someone else.

Issue #1457, opened 2026-07-24
feat(desktop): surface linked child sessions in the session workbar


Astro-Han:
## Summary

Linked child sessions move out of the conversation list and into the session workbar. The sidebar returns to one row per conversation intent; delegated work gets a permanent, interactive home beside the parent that dispatched it — and **one** home, not three.

This supersedes the conservative v1 projection from #1366 / #1400, which rendered every child as an indented row under its parent.

## Decision

**One identity, two faces.** A subagent's identity is its preset — `subagent_id`, name, capability profile, model route — configured in 设置 · 子 Agent (#1999, rebuilt in #2008). Every child session already records which preset spawned it: `SubagentSessionRuntime` carries `presetId`, `agentName`, and `profile` (`packages/core/src/session.ts:120-124`). The runtime surface reads that identity instead of inventing its own labels, and the settings page reads back which of its presets are running. No new contract is needed for either direction.

**Sidebar** — child sessions no longer occupy rows. A parent row carries a `N` subagent chip only while something needs attention (amber: a child is waiting for permission or failed; blue: a child is running). The chip disappears once every child settles, so history does not accumulate badges. The nested row rendering (`packages/ui/src/session-history-list.tsx:179`, `:375`) and the `data-subagent` background in `sidebar.css:237` — which collides with the active-row tint — are removed.

**Session workbar** — `SessionWorkbar` gains a fourth tab, 子 Agent, with a persistent count (`session-workbar-layout.ts:8` currently knows `tasks | browser | files | quote`). It lists the active session's direct children, attention first, then most recent. A row reads as **preset name · profile · model**, not as a session title, reusing the vocabulary `subagent-preset-presentation.ts` already defines.

**Graph folds into the same tab.** `AgentGraphPanel` currently floats above the composer whenever a session runs in graph mode (`app-shell.tsx:2215`). Graph is one topology of delegation, not a separate kind of work, so it becomes the tab's graph-mode presentation rather than a second panel. Without this, landing the tab would leave two surfaces both answering "how is my delegated work doing" — the exact duplication this issue exists to remove.

**Opening a child** — selecting a row turns the panel into a mini chat for that child session, with a back affordance, a composer, and inline permission prompts. The parent conversation stays put in the main pane. `QuoteCompanionPanel` already runs a full interactive chat inside this panel, so this reuses a proven seam instead of adding one. An "open in main pane" escape hatch covers long child conversations.

[...]

PR #2956, opened 2026-08-13
fix(desktop): dismiss a completed Agent Graph panel

### LINKED ISSUE #2661: fix(desktop): completed Agent Graph panel cannot be dismissed
## What happened

After all tasks in an Agent Graph have completed, the Agent Graph panel remains permanently visible above the composer.

The chevron allows the panel content to be expanded or collapsed, but there is no way to dismiss or hide the completed graph panel entirely. The collapsed panel continues to display a status such as `Agent Graph — Completed · 4/4 finished` even though the graph no longer has active work.

## How to reproduce

1. Start a task that creates an Agent Graph.
2. Wait until every operator in the graph has finished.
3. Confirm that the panel shows `Completed` and that all operators have ended.
4. Collapse the panel using the chevron.
5. Try to close or dismiss the completed Agent Graph panel.

**Actual result:** The panel remains above the composer, and there is no action to remove or hide it.

**Expected result:** Once the Agent Graph is completed, the user can dismiss the panel completely. Alternatively, the completed panel may automatically move into a historical or delegated-work view while remaining accessible when needed.

## Environment

- Maka version or commit: unknown
- OS and version: macOS (exact version unknown)
- Surface: Desktop

## Additional context

The completed panel occupies persistent space above the composer even though it no longer represents active work.

Related: #1457 proposes moving `AgentGraphPanel` from its composer-adjacent location into the session workbar. Until that experience is available, the current panel should provide a clear dismiss/hide action for completed graphs.

1625567290:
## Summary

Completed Agent Graph panels could only collapse, so they stayed pinned above the composer after the work was done.

This adds a dismiss control for settled graphs (`completed`, `stopped`, `failed`). Dismissal is session-local UI state: the panel comes back if that session starts a new graph, and it does not change Runtime Host, persistence, or the workbar move in #1457.

Fixes #2661

## Verification

Ran locally:

- `node --test apps/desktop/dist/main/__tests__/agent-graph-panel.test.js apps/desktop/dist/main/__tests__/agent-graph-panel-visibility.test.js`
  - 15 pass / 0 fail, including a real `AgentGraphPanel` click path
- `tsc -p apps/desktop/tsconfig.renderer.json --noEmit`
- Electron window: injected a completed snapshot, remounted the session, clicked dismiss; `.maka-agent-graph-panel` count went to 0

Not run: full `npm test`, desktop e2e suite, `npm run typecheck` for the whole repo.

## Checklist

- [x] Yes — a settled Agent Graph panel can now be dismissed from the composer

### DIFF
--- a/apps/desktop/src/renderer/agent-graph-panel.tsx
+++ b/apps/desktop/src/renderer/agent-graph-panel.tsx
@@ -202,7 +202,11 @@ export function AgentGraphPanel(props: {
         current,
         props.rootSessionId,
         snapshot
-          ? { graphId: snapshot.graphId, status: snapshot.status }
+          ? {
+              rootSessionId: snapshot.rootSessionId,
+              graphId: snapshot.graphId,
+              status: snapshot.status,
+            }
           : undefined,
       ),
     );
--- a/apps/desktop/src/renderer/agent-graph-panel.tsx
+++ b/apps/desktop/src/renderer/agent-graph-panel.tsx
@@ -5,11 +5,18 @@ import type {
   AgentGraphClientSnapshot,
 } from '@maka/runtime/stream-graph-read-model';
 import { IconButton } from '@maka/ui';
-import { ICON_SIZE, ChevronDown } from '@maka/ui/icons';
+import { ICON_SIZE, ChevronDown, X } from '@maka/ui/icons';
 import { Banner } from '@astryxdesign/core/Banner';
 import { Button } from '@astryxdesign/core/Button';
 import { EmptyState } from '@astryxdesign/core/EmptyState';
 import { Spinner } from '@astryxdesign/core/Spinner';
+import {
+  dismissAgentGraphPanel,
+  isAgentGraphPanelDismissible,
+  reconcileAgentGraphPanelDismissals,
+  shouldShowAgentGraphPanel,
+  type AgentGraphPanelDismissals,
+} from './agent-graph-panel-visibility.js';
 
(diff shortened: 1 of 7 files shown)


apps/desktop/src/renderer/agent-graph-panel.tsx
```diff
@@ -185,6 +196,18 @@ export function AgentGraphPanel(props: {
     };
   }, [props.rootSessionId, props.enabled]);
 
+  useEffect(() => {
+    setDismissedBySession((current) =>
+      reconcileAgentGraphPanelDismissals(
+        current,
+        props.rootSessionId,
+        snapshot
```
likun666661, line 204: [P2] Ignore snapshots owned by the previous session

When `rootSessionId` changes, this effect can run with the new prop but the previous session’s `snapshot` before the earlier effect clears it. For example, after dismissing A, switching A → B → A can reconcile A’s dismissal against B’s `graphId` and delete it, so A’s dismissed panel reappears. Please only reconcile when `snapshot.rootSessionId === props.rootSessionId`, and cover the A → B → A switch in the component test. This is independent of the intended future `sessionId -> [graphId]` model.

1625567290: Fixed. Reconcile now no-ops unless `snapshot.rootSessionId ===` the current session, so A → B → A cannot apply B’s `graphId` to A’s dismissal. Covered in the visibility test and the component test.

PR #2940, opened 2026-08-13
docs(windows): define sandbox backend security contract

liugddx:
### Summary

- Define the Windows sandbox threat model, enforcement contract, delivery gates, and adversarial evidence requirements.
- Add paired English and Chinese RFCs.
- Ground the architecture in pinned primary-source reviews of Microsoft APIs, OpenAI Codex, Gemini CLI, Chromium, Claude Code, and OpenCode.
- Revise the proposed baseline after source review to use dedicated sandbox identities, restricted tokens, Job Objects, private desktops, explicit handles, ACL reconciliation, and identity-scoped network policy.
- Retain AppContainer as a W0 comparison target instead of claiming that it has already been selected.
- Refresh the sandbox README to reflect the implemented Linux bubblewrap backend.

### Design Status

This PR is the complete Phase 4 security design baseline for #2142, but it is not yet a frozen implementation specification and does not enable Windows restricted profiles.

Three W0 gates remain explicit:

1. Extract or adapt the Apache-2.0 Codex Windows sandbox component, or build a Maka-owned implementation.
2. Freeze the setup, identity, ACL ledger, protocol, upgrade, uninstall, and crash-recovery schemas.
3. Select and prove WFP, verified Windows Firewall rules, or a combination for network denial.

The RFC estimates W0 through W3 at 7 to 12 weeks for one experienced engineer. If W0 validates the proposed baseline, the managed read-only Read, Glob, and Grep milestone is estimated at roughly 3 to 5 weeks.

### Research Boundary

The survey is representative, not exhaustive. It includes projects that ship a native agent-oriented Windows sandbox, a mature Windows process sandbox, or an explicit public Windows isolation contract. Closed or undocumented implementations are not treated as evidence. Every repository observation is pinned to a reviewed commit.

### Verification

- `git diff --check upstream/main...HEAD`
- Reviewed the current Maka SandboxManager, Linux bubblewrap, macOS Seatbelt, diagnostics, filesystem worker, and Runtime Host composition.
- Reviewed pinned source trees for Codex, Gemini CLI, Chromium, and OpenCode, plus the public Claude Code repository and documentation.
- Verified that the current revision's full GitHub CI passes, including Windows baseline and Windows recovery.

This revision changes documentation only. The clean local workspace intentionally has no installed dependencies; GitHub CI is authoritative for repository tests.

Related to #2142. This PR does not close the umbrella issue.

---

<details>
<summary>中文版</summary>

### 概要

[...]

Issue #2142, opened 2026-08-04
roadmap(windows): make Windows a supported platform


liugddx:
## Problem

Maka can already build and launch its CLI and Electron desktop app on Windows, and the codebase includes Windows-specific paths for named pipes, PowerShell/cmd shell detection, ConPTY, process-tree termination, and filesystem behavior. However, Windows is not currently a supported platform because these paths are not continuously tested or released, and several product guarantees remain platform-specific.

A local Windows baseline on Node 22 found:

- the full repository build succeeds;
- script tests can pass after removing POSIX-only assumptions from macOS development helper tests;
- managed workspace tests require Git for Windows long-path support (`core.longpaths=true`);
- the storage suite currently reports `514 pass / 100 fail / 40 skip`;
- most remaining storage failures are `EBUSY` cleanup failures caused by SQLite handles still owning `runtime.sqlite` or `runtime.sqlite-shm`, which POSIX permits unlinking but Windows does not;
- restricted sandbox profiles fail closed on Windows;
- computer-use has no Windows backend.

Being able to open the app is therefore not the same as having a tested, secure, releasable Windows product.

## Desired outcome

Define and deliver an explicit Windows support tier for Maka. At minimum, supported CLI and desktop workflows should build, test, install, update, recover from crashes, and fail safely on every release. Features that cannot initially be supported, especially sandboxing and computer-use, must be clearly surfaced rather than silently degraded.

### Current support status (2026-08-25)

Windows 11 x64 remains an active **preview** target, not a fully supported Maka platform. Phase 0-2 build, baseline, crash-recovery, process-cleanup, and durability evidence are in place. Phase 3 preview packaging, checksum/install documentation, closed-app upgrade/uninstall, automatic update, Runtime Host update handoff, and the accepted Abort/Quit recovery boundary are complete; Authenticode signing is the only remaining Phase 3 implementation gate and is tracked by #3414.

The shipped W1 filesystem-worker AppContainer surface has now completed its Phase 4 automated and independent-review gates. #3558 promotes the lifecycle checks into the shared pull-request/formal-release verifier and protects its trigger/source closure; #3586 verifies explicit post-dispatch client cancellation; #3722 closes Runtime Host parent-death, owner binding, Job drain, a 64-launch soak, quarantined ACL-ledger non-reuse, and the scoped adversarial matrix. All three are merged, and #3722 received exact-head independent maintainer approval with no P0-P3 findings.

[...]

PR #2961, opened 2026-08-13
feat(windows): add brokered AppContainer sandbox support

liugddx:
## Summary

- add a packaged Windows x64 AppContainer sandbox backend for restricted managed execution
- compile `PermissionProfile` roots, network policy, and environment into a closed launch manifest
- authorize one-shot broker requests with a kernel-reported client PID, single-use nonce, and digest of the complete launch policy
- launch the target in an AppContainer with no network capabilities and atomic kill-on-close Job membership
- grant only requested filesystem roots and restore AppContainer ACEs through a crash-recoverable ledger
- register the backend only when the packaged native resource exists; otherwise managed execution fails closed
- connect the backend to the existing Runtime Host filesystem-worker and Agent execution path
- replace the W0 proposal with matching English and Chinese RFCs for the selected implementation

## Security Evidence

The blocking Windows 2025 job verifies:

- closed protocol validation and bounded broker framing
- pipe access restricted to SYSTEM and the current user, remote-client rejection, and kernel PID binding
- nonce replay, unapproved digest, response-identity, and manifest-cleanup rejection paths
- recomputation of the digest from executable, arguments, cwd, roots, network policy, and sorted environment
- AppContainer token creation with atomic `PROC_THREAD_ATTRIBUTE_JOB_LIST` membership
- allowed-root read/write, denial of a user-readable sibling file, and denial of a live loopback endpoint
- AppContainer ACE removal after normal exit, stale-ledger recovery, and junction/reparse-point rejection
- bounded fail-closed behavior of the rejected restricted-token and non-atomic candidates

## Product Integration

The Windows release pipeline builds `maka-windows-sandbox.exe` with locked Cargo dependencies, packages it under `resources/windows-sandbox`, and requires that resource during packaged-app verification. Runtime capability detection registers the backend only when that exact packaged resource exists. There is no environment-variable executable override and no unsandboxed retry.

## Scope Boundary

This PR provides a real native sandbox for the packaged Windows x64 managed-execution path and is ready for implementation review once all checks pass. It does not by itself declare the entire Windows platform fully supported. Authenticode signing, the complete adversarial matrix, cancellation/parent-death/concurrency release evidence, and independent security review remain tracked by #2142.

## Verification

- `node --test experiments/windows-sandbox/*.test.mjs`
- `cargo fmt --manifest-path experiments/windows-sandbox/launcher/Cargo.toml -- --check`
- `cargo check --manifest-path experiments/windows-sandbox/launcher/Cargo.toml --locked`
- PowerShell parser validation for all sandbox smoke scripts
- Biome checks for changed JavaScript and TypeScript
- GitHub workspace, Runtime Host, Desktop E2E, Storybook, Windows baseline/recovery, Windows packaging, and blocking sandbox jobs

Related to #2142. Supersedes the implementation choice proposed in #2940 with the AppContainer result proven by Windows CI.

<details>
<summary>中文说明</summary>

## 概要

- 为 Windows x64 restricted managed execution 增加随安装包发布的 AppContainer 沙箱后端
- 把 `PermissionProfile` 的目录、网络策略和环境编译为封闭 launch manifest
- 使用内核返回的 client PID、一次性 nonce 和完整 launch policy 摘要授权 one-shot broker 请求
- 在不授予网络 capability 的 AppContainer 中启动目标，并在创建时原子加入 kill-on-close Job
- 只给请求目录临时授予 ACL，并通过可恢复 ledger 清理 AppContainer ACE
- 只有打包 native resource 存在时才注册后端；不可用时 managed execution 保持 fail closed
- 接入既有 Runtime Host filesystem worker 与 Agent 执行链路
- 用中英文 RFC 记录 Windows 实验证明后最终选定的 AppContainer 方案

## 安全证据

阻塞式 Windows 2025 job 验证：

- 封闭协议和有长度上限的 broker frame
- pipe 仅允许 SYSTEM 与当前用户、拒绝远程 client、从内核绑定 PID
- nonce 重放、未批准摘要、response identity 和 manifest 清理失败路径
- 从 executable、arguments、cwd、roots、network 和排序环境重算 launch digest
- AppContainer token 与 `PROC_THREAD_ATTRIBUTE_JOB_LIST` 原子 Job membership
- 允许目录读写、相邻用户可读文件拒绝、真实 loopback 网络拒绝
- 正常退出 ACL 清理、遗留 ledger recovery、junction/reparse-point 拒绝
- 被否决 restricted-token 与非原子候选的有界 fail-closed 行为

## 产品接入

Windows 发布流水线使用锁定 Cargo 依赖构建 `maka-windows-sandbox.exe`，将它打包到 `resources/windows-sandbox`，并在安装包验证时强制检查该资源。Runtime 只从这个打包路径做 capability detection，不接受环境变量替换任意 executable，也不会 unsandboxed retry。

## 范围边界

本 PR 为 Windows x64 安装包的 managed execution 提供真实原生沙箱；全部检查通过后即可进行实现评审。它不等于整个 Windows 平台已经正式支持。Authenticode 签名、完整对抗矩阵、cancel/parent-death/concurrency 发布证据和独立安全审查继续由 #2142 跟踪。

## 验证

- `node --test experiments/windows-sandbox/*.test.mjs`
- `cargo fmt --manifest-path experiments/windows-sandbox/launcher/Cargo.toml -- --check`
- `cargo check --manifest-path experiments/windows-sandbox/launcher/Cargo.toml --locked`
- 全部沙箱 PowerShell smoke script 的 parser 检查
- 变更 JavaScript/TypeScript 的 Biome 检查
- GitHub workspace、Runtime Host、Desktop E2E、Storybook、Windows baseline/recovery、Windows packaging 和阻塞式 sandbox jobs

关联 #2142。Windows CI 的 AppContainer 结果取代 #2940 最初提出的实现选择。

</details>


Astro-Han (commented):
I found no P0 or P1 issues on `9f297988cfa4135babfbb64b722d0d67a291a74c`. I left four P2 comments that should be addressed before merge: the token success check does not match the token being created, the CI lane can report success without producing the stated evidence, broker authorization does not bind the approved profile to the launch policy, and the legacy assignment failure can leave a suspended child behind.

The broker authorization and framing additions are also independent of the launcher evidence and are not connected to a real broker peer yet. Moving them to a follow-up around the actual broker seam would keep this PR focused, but I am treating that as a scope recommendation rather than a separate finding.

AI-assisted review disclosure: Codex prepared and organized the review evidence, and Claude Opus 5 provided an independent read-only pass. The human reviewer verified the cited code paths, CI behavior, and Windows API semantics against the exact head above.

experiments/windows-sandbox/launcher/src/windows_launcher.rs
```diff
@@ -0,0 +1,406 @@
...
+    }
+    Ok(exit_code as u8)
+}
+
+unsafe fn child_token_is_restricted(process: HANDLE) -> Result<bool, String> {
+    let mut token = null_mut();
+    if unsafe { OpenProcessToken(process, TOKEN_QUERY, &mut token) } == 0 {
+        return Err(last_error("OpenProcessToken(child)"));
+    }
+    let result = unsafe { IsTokenRestricted(token) != 0 };
```
Astro-Han, line 853: [P2] This check cannot establish the property used by the success condition. `CreateRestrictedToken` above supplies no restricting SIDs, while `IsTokenRestricted` reports whether the token has a restricting SID list. It can therefore return false even if the intended privileges were removed, preventing the atomic path from reporting valid evidence on a privileged runner. Please either add the restricting SID that this check expects or verify the specific token properties this prototype intends to enforce.

.github/workflows/windows-sandbox-w0.yml
```diff
@@ -0,0 +1,41 @@
...
+        shell: pwsh
+        run: ./experiments/windows-sandbox/atomic-launch-capability.ps1
+
+      - name: Build process-containment launcher
+        working-directory: experiments/windows-sandbox/launcher
+        run: cargo build --locked
+      - name: Verify launcher and broker authorization
+        working-directory: experiments/windows-sandbox/launcher
+        run: cargo test --locked
+      - name: Probe atomic Job launch candidate
```
Astro-Han, line 62: [P2] This lane can pass without producing the W0 evidence described by the PR. The launcher smoke runs `cmd.exe /c exit 0`, not the untrusted probe; the missing-privilege branch accepts any nonzero launcher failure rather than the documented error 1314; and job-level `continue-on-error` turns build or smoke failures into a successful check. Please keep the workflow non-required if it must remain non-blocking, but preserve its failure status, match the expected capability failure specifically, and run the probe through the launcher before treating this as sandbox evidence.

experiments/windows-sandbox/launcher/src/broker_authorization.rs
```diff
@@ -0,0 +1,84 @@
...
+    ) -> Result<(), BrokerAuthorizationError> {
+        request
+            .validate()
+            .map_err(BrokerAuthorizationError::InvalidRequest)?;
+        if request.client_pid != connected_client_pid {
+            return Err(BrokerAuthorizationError::ClientPidMismatch);
+        }
+        if !self
+            .allowed_profile_digests
+            .contains(&request.profile_digest.to_ascii_lowercase())
```
Astro-Han, line 44: [P2] An approved `profileDigest` is not bound to `request.launch`. A client can reuse an approved digest while changing the executable, roots, network mode, or environment, and this authorizer will still accept the request. The broker needs to resolve a broker-owned policy from the digest, or canonicalize and hash the requested policy before comparing it. Please add a test that changes the launch policy while keeping an approved digest.

experiments/windows-sandbox/launcher/src/windows_launcher.rs
```diff
@@ -0,0 +1,406 @@
...
+    };
+    if created == 0 {
+        return Err(last_error("CreateProcessWithTokenW"));
+    }
+
+    let result = if unsafe {
+        windows_sys::Win32::System::JobObjects::AssignProcessToJobObject(job, process.hProcess)
+    } == 0
+    {
+        Err(last_error("AssignProcessToJobObject"))
```
Astro-Han, line 375: [P2] The child already exists suspended here, but assignment to the Job failed. Returning the error closes its handles, and closing the Job cannot terminate a process that was never assigned to it, so this path leaves a suspended orphan behind. Please terminate the child before returning from this post-create failure path, ideally through cleanup that covers every failure after process creation.

experiments/windows-sandbox/launcher/src/windows_launcher.rs
liugddx: Fixed in `c6c297c17`: the restricted-token candidate now adds a restricting SID, so `IsTokenRestricted` validates the token property that the success condition claims. The AppContainer path selected for production has separate token evidence.

.github/workflows/windows-sandbox-w0.yml
liugddx: Addressed across `15dc56538`, `e05dd4a26`, and `426fa4d8d`: capability failure is matched explicitly, the real AppContainer boundary probe verifies admitted roots plus live network denial, and the sandbox job is now blocking (no `continue-on-error`).

experiments/windows-sandbox/launcher/src/broker_authorization.rs
liugddx: Fixed in `f69950dd7`: authorization recomputes SHA-256 from the complete canonical launch object and compares it to the approved digest. Tests mutate the launch while retaining the digest and require rejection.

experiments/windows-sandbox/launcher/src/windows_launcher.rs
liugddx: Fixed in `e4c1c1006`: every failure after process creation terminates and waits for the suspended child before handles are closed, including Job assignment failure.

liugddx:
Addressed the latest Copilot path-boundary findings in `cfcc792cc`:

- `pathWithinRoot()` now handles Windows drive roots without producing a doubled separator.
- `relativeSegments()` now uses the platform separator and correctly handles drive roots, so protected `.git`/`.agents` metadata checks work on Windows.
- `grepFiles()` now inserts `--` before the user pattern to prevent option-like patterns from being parsed as ripgrep flags.
- Added regression coverage for all three cases.

Validation: the new core Windows path tests and the `rg --` regression pass locally. The full runtime workspace test command still contains pre-existing Windows-only failures in POSIX-shell assumptions and Glob separator expectations; those are unrelated to this patch.

experiments/windows-sandbox/launcher/src/windows_launcher.rs
```diff
@@ -0,0 +1,705 @@
...
+}
+
+fn quote_argument(value: &str) -> String {
+    format!("\"{}\"", value.replace('"', "\\\""))
+}
+
+fn environment_block(environment: &std::collections::BTreeMap<String, String>) -> Vec<u16> {
+    if environment.is_empty() {
+        return Vec::new();
+    }
```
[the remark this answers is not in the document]
liugddx: Fixed in 5d6fac67e: environment_block() now always builds an explicit (possibly empty) block and every launch path always sets CREATE_UNICODE_ENVIRONMENT, so an empty allowlist can no longer fall back to inheriting the broker's ambient environment. Unit tests cover the empty and populated block shapes.

packages/runtime/src/sandbox/windows-sandbox.ts
```diff
@@ -0,0 +1,245 @@
...
+
+export function createWindowsBrokerManifestWriter(
+  temporaryRoot: string = tmpdir(),
+): (manifest: WindowsBrokerManifest) => string {
+  const directory = mkdtempSync(join(temporaryRoot, 'maka-windows-sandbox-'));
+  return (manifest) => {
+    // Windows cleanup tools may remove idle temp directories between turns.
+    // Recreate it before each one-shot broker request.
+    mkdirSync(directory, { recursive: true });
+    const path = join(directory, `${manifest.requestId}-${randomBytes(8).toString('hex')}.json`);
```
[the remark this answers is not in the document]
liugddx: Fixed in 86c44465d: the accepted request-id charset is now [A-Za-z0-9._-] (':' dropped), with a regression test asserting a colon id fails as invalid_request. The default remains a random hex id.

packages/core/src/permission-profile.ts
```diff
@@ -316,19 +316,9 @@ function relativeSegments(path: string, root: string): string[] | undefined {
   return relative.split('/').filter(Boolean);
```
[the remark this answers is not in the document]
liugddx: Addressed: relativeSegments() now derives the separator from the root ('\' for Windows drive paths), handles the volume-root slice explicitly, and splits on that separator (packages/core/src/permission-profile.ts), so protected-metadata checks work for C:\ paths.

packages/runtime/src/sandbox/windows-profile.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+function canonicalWindowsPath(path: string): string {
+  if (!win32.isAbsolute(path) || path.includes('/')) {
+    throw new Error(`Windows sandbox path must be absolute and use backslashes: ${path}`);
+  }
+  const canonical = win32.resolve(path);
+  if (canonical.toLowerCase() !== path.toLowerCase()) {
+    throw new Error(`Windows sandbox path must be lexically canonical: ${path}`);
+  }
```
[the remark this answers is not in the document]
liugddx: Addressed: canonical Windows path validation moved to the shared @maka/core/windows-path + absolute-path helpers, which require a drive-absolute prefix ([A-Za-z]:\) and explicitly reject UNC (\\server), device (\\?\, \\.\) and ADS (':' after the drive prefix) forms. Both windows-profile.ts and windows-sandbox.ts now delegate to that single validator.

packages/core/src/absolute-path.ts
```diff
@@ -0,0 +1,50 @@
+/** Pure absolute-path helpers shared by permission and sandbox-boundary contracts. */
+export function isNormalizedAbsolutePath(path: string): boolean {
+  if (!path || path.includes('\0')) return false;
+  if (isWindowsDrivePath(path)) {
+    if (path.includes('/') || (path.length > 3 && path.endsWith('\\'))) return false;
+    if (path.length === 3) return true;
+    return !path
+      .slice(3)
+      .split('\\')
+      .some((segment) => segment === '' || segment === '.' || segment === '..');
```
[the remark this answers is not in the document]
liugddx: Addressed: isNormalizedAbsolutePath in packages/core/src/absolute-path.ts now rejects ':' after the drive prefix (ADS), \\ (UNC) and \\?\ / \\.\ (device namespaces), and the runtime sandbox validators (windows-sandbox.ts, windows-profile.ts) consume the same grammar via @maka/core/windows-path instead of keeping local copies.

experiments/windows-sandbox/launcher/src/acl_ledger.rs
```diff
@@ -0,0 +1,215 @@
...
+    file.write_all(&bytes)
+        .and_then(|()| file.sync_all())
+        .map_err(|error| format!("persist ACL ledger failed: {error}"))
+}
+
+fn recover_stale(root: &Path) -> Result<(), String> {
+    for entry in fs::read_dir(root).map_err(|error| format!("read ACL ledgers failed: {error}"))? {
+        let path = entry
+            .map_err(|error| format!("read ACL ledger entry failed: {error}"))?
+            .path();
```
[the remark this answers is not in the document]
liugddx: Addressed: ACL recovery now serializes on a cross-process .active.lock (owner-PID stamped, with stale-owner detection) inside with_acl_grants; recover_stale() only runs while holding that lock, and grant -> launch -> remove-grants -> delete-ledger all happen under it, so a later invocation can no longer reap the ledger of a still-running sandbox. The acl-recovery smoke exercises 10 concurrent --appcontainer launches plus kill-and-replay recovery.

experiments/windows-sandbox/launcher/src/broker_framing.rs
```diff
@@ -0,0 +1,50 @@
+pub const MAX_BROKER_MESSAGE_BYTES: usize = 64 * 1024;
+const LENGTH_PREFIX_BYTES: usize = size_of::<u32>();
```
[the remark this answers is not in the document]
liugddx: Not an issue on the pinned toolchain: size_of has been in the std prelude since Rust 1.80, so broker_framing.rs compiles without an explicit import — cargo build/test run green in the windows_sandbox_w0_protocol job on every push.

experiments/windows-sandbox/launcher/src/broker_pipe.rs
```diff
@@ -0,0 +1,191 @@
...
+        let mut prefix = [0u8; 4];
+        unsafe { read_exact(pipe, &mut prefix)? };
+        let payload_length = u32::from_le_bytes(prefix) as usize;
+        if payload_length == 0 || payload_length > MAX_BROKER_MESSAGE_BYTES {
+            return Err("invalid broker payload length".to_owned());
+        }
+        let mut buffer = Vec::with_capacity(4 + payload_length);
+        buffer.extend_from_slice(&prefix);
+        buffer.resize(4 + payload_length, 0);
+        unsafe { read_exact(pipe, &mut buffer[4..])? };
```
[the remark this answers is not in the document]
liugddx: Acknowledged, deferring by scope: the shipped path in this milestone is --broker-local only (packages/runtime/src/sandbox/windows-sandbox.ts spawns the broker per request with an in-process serve; no named pipe involved), so a stalled pipe peer cannot affect production launches. The named-pipe server/client pair is exercised only by the experiment smokes. Pipe I/O deadlines are queued as follow-up work for when the remote-broker transport graduates from the spike.

experiments/windows-sandbox/launcher/src/windows_launcher.rs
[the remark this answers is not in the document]
liugddx: Follow-up in fbd587e2f: the explicit block surfaced a real Windows constraint — AppContainer creation resolves the package directory through LOCALAPPDATA and CreateProcessW fails with os error 203 when an explicit block omits it. The broker now fills SystemRoot/SystemDrive/LOCALAPPDATA from its own environment when the manifest does not set them (manifest values win, case-insensitively). These are fixed non-secret machine paths; everything else remains strictly allowlisted. Verified by unit tests plus the full local smoke suite (boundary probe still reports fileDenied/networkDenied).

liugddx:
image
image

image

packages/runtime/src/sandbox/windows-profile.ts
```diff
@@ -0,0 +1,151 @@
...
+        if (match === 'exact') addUnique(exactWriteRoots, canonical);
+      }
+    }
+  }
+
+  for (const path of [
+    ...(pathContext.runtimeReadableRoots ?? []),
+    ...(pathContext.executableRoots ?? []),
+  ]) {
+    addUnique(readRoots, canonicalWindowsPath(path));
```
Astro-Han, line 56: [P1] Please avoid translating every runtime support root into a recursive ACL grant. `launch-spec.ts` supplies `runtimeRoot = resolve(dirname(program), '..')`; with the packaged Electron executable this can become `%LOCALAPPDATA%\Programs`, `C:\Program Files`, or the parent of a ZIP extraction directory. This path is not exact here, so the launcher recursively scans it and applies `icacls /T`. On a real installation that either fails on an unrelated unreadable/reparse entry or temporarily grants the AppContainer read/execute access to the whole parent tree. Both Windows E2E paths avoid this by copying `node.exe` into a shallow temporary layout. Please make the Windows launch spec enumerate only the product-owned files/directories the worker actually needs, and add coverage using a realistic packaged directory topology.

experiments/windows-sandbox/launcher/src/acl_ledger.rs
```diff
@@ -0,0 +1,401 @@
...
+        if path.extension().and_then(|value| value.to_str()) != Some("json") {
+            continue;
+        }
+        let source = fs::read(&path)
+            .map_err(|error| format!("read stale ACL ledger {} failed: {error}", path.display()))?;
+        let ledger: Ledger = match serde_json::from_slice(&source) {
+            Ok(ledger) => ledger,
+            Err(error) => {
+                quarantine_ledger(&path, &format!("invalid stale ACL ledger: {error}"))?;
+                continue;
```
Astro-Han, line 428: [P1] Quarantining this ledger does not revoke the ACL it may describe. Every launch reuses the stable `maka.sandbox.w0` AppContainer SID, so continuing here can let the next child inherit a path granted to an earlier session. The same problem applies to the unsupported-version and `remove_grants` failure branches below. Recovery needs to fail closed whenever it cannot prove that all grants for the stable SID were removed, or the launcher needs a fresh per-launch identity. The existing quarantine tests should assert that no child starts after an unrecoverable ledger rather than codifying continuation.

experiments/windows-sandbox/launcher/src/acl_ledger.rs
```diff
@@ -0,0 +1,401 @@
...
+        let recursive_read = contains_path(&request.read_roots, path)
+            && !contains_path(&request.exact_read_roots, path);
+        let recursive_write = contains_path(&request.write_roots, path)
+            && !contains_path(&request.exact_write_roots, path);
+        if metadata.is_dir() && (recursive_read || recursive_write) {
+            reject_reparse_tree(Path::new(path))?;
+        }
+        roots.push(LedgerRoot {
+            path: path.clone(),
+            recursive: metadata.is_dir(),
```
Astro-Han: [P2] This records “the root is a directory,” not whether the broker actually applied a recursive grant. `grant_roots` later suppresses recursion for an exact root, while `remove_grants` trusts this field and invokes `/T`. An exact cwd metadata grant can therefore be removed by recursively traversing the entire workspace, and crash recovery repeats the same mismatch. Please make the ledger record the actual applied grant mode and have grant, cleanup, and recovery consume that single fact instead of recomputing it differently.

experiments/windows-sandbox/launcher/src/acl_ledger.rs
```diff
@@ -0,0 +1,401 @@
...
+                }
+                thread::sleep(Duration::from_millis(50));
+            }
+            Err(error) => return Err(format!("acquire ACL ledger lock failed: {error}")),
+        }
+    };
+    writeln!(lock_file, "{}", std::process::id())
+        .and_then(|()| lock_file.sync_all())
+        .map_err(|error| format!("persist ACL ledger lock owner failed: {error}"))?;
+    let _lock = LedgerLock { path: lock_path };
```
Astro-Han: [P2] The lock lifecycle has two separate production failure modes. The file is created before the PID is persisted and before this guard exists, so a crash or write failure leaves an empty/partial lock; `lock_owner_is_stale` treats an unparseable owner as live forever. This guard also remains held through `launch()` and child completion, while peers wait only 30 seconds, so one valid long-running operation makes concurrent filesystem tools fail. Please use recoverable OS-owned lock identity plus a per-root lease/reference model, or a per-launch SID. Merely increasing the timeout would keep all Windows managed execution serialized.

experiments/windows-sandbox/launcher/src/windows_launcher.rs
```diff
@@ -0,0 +1,862 @@
...
+        return Err(last_error("CreateFileW(NUL stdio)"));
+    }
+    Ok(handle)
+}
+
+unsafe fn wait_for_child(process: HANDLE, timeout_ms: Option<u64>) -> Result<u8, String> {
+    let timeout_ms = timeout_ms.unwrap_or(DEFAULT_LAUNCH_TIMEOUT_MS);
+    let wait = unsafe { WaitForSingleObject(process, timeout_ms as u32) };
+    if wait == WAIT_TIMEOUT {
+        unsafe { TerminateProcess(process, 124) };
```
Astro-Han: [P2] `TerminateProcess` is asynchronous, and this branch returns without confirming that even the root process has exited. The closure then unwinds into `with_acl_grants`, which removes the ACLs before `launch_appcontainer` closes the kill-on-close Job handle. The timed-out process, and any descendants it created, can therefore still be running while temporary authority is being removed. Please terminate and wait for the complete Job/process tree to drain, checking the termination and wait results, before ACL cleanup begins.

packages/runtime/src/filesystem-worker/operations.ts
```diff
@@ -313,6 +324,63 @@ export async function executeFilesystemOperation(
...
+    throw operationError('invalid_request', 'Grep pattern is not a valid regular expression.');
+  }
+  const deadline = Date.now() + operation.timeoutMs;
+  const matches: string[] = [];
+  const files: string[] = [];
+  const metadata = await fs.stat(path);
+  if (metadata.isFile()) {
+    files.push(path);
+  } else {
+    for await (const entry of nodeGlob(operation.glob ?? '**/*', {
```
Astro-Han: [P2] This bypasses the contained-glob validation used by the normal Glob operation. Node’s glob accepts parent traversal here; for example, a pattern such as `../../../core/src/permission-profile.ts` is enumerated outside `path`, and the resolved candidate is read without another containment check. The AppContainer may block paths outside its broader ACL roots, but it does not preserve the Grep operation’s requested search boundary when the escaped file is otherwise readable. Please validate the Grep glob with the shared contained-pattern rule and recheck every resolved candidate before reading it.

packages/runtime/src/filesystem-worker/operations.ts
```diff
@@ -313,6 +324,63 @@ export async function executeFilesystemOperation(
...
+ * `rg -n --no-heading`: `path:line:text` with absolute paths. Unlike ripgrep
+ * it does not honor .gitignore, so it may return a superset of matches.
+ */
+async function runInProcessGrep(
+  path: string,
+  operation: Extract<FilesystemWorkerOperation, { kind: 'grep' }>,
+): Promise<string[]> {
+  let expression: RegExp;
+  try {
+    expression = new RegExp(operation.pattern);
```
Astro-Han: [P2] The fallback cannot currently enforce its advertised timeout or resource bounds. A synchronous JS regexp can enter catastrophic backtracking before the next deadline check—I reproduced `/(a+)+$/` against 26 `a` characters plus `!` blocking for about 3.2 seconds—and the same function enumerates the complete glob before checking the deadline, then reads and splits each file in full. A single pattern or large file can therefore hold or exhaust the worker until the outer process timeout. Please use a bounded/linear-time matcher and stream files while checking the deadline, or explicitly reject patterns and inputs that this fallback cannot execute safely.

experiments/windows-sandbox/protocol.mjs
```diff
@@ -0,0 +1,76 @@
...
+  'version',
+  'requestId',
+  'executable',
+  'arguments',
+  'cwd',
+  'readRoots',
+  'writeRoots',
+  'network',
+  'environment',
+]);
```
Astro-Han: [P2] This parser is already a stale parallel protocol authority. The production manifest always includes `exactReadRoots`, `exactWriteRoots`, and `timeoutMs`, so this allowlist rejects the shipped request shape while its own tests still pass. Its environment-name and Windows path rules also differ from the Runtime and Rust validators, and no production caller uses it. Please delete this parser/test (and the similarly superseded MJS probe) or replace them with one fixture exercised by the real TS and Rust authorities; the current green step should not be presented as closed-protocol evidence.

scripts/verify-windows-x64.mjs
```diff
@@ -106,6 +109,74 @@ export async function verifyPackagedWindowsApp(
...
+      );
+    }
+    if (!sandboxProbe.stderr.includes('"atomicJob":true')) {
+      throw new Error(
+        `Packaged Windows sandbox broker diagnostic did not reach stderr: ${sandboxProbe.stderr}`,
+      );
+    }
+    await assertMissing(sandboxManifest);
+
+    step('relaying command stdio through the packaged sandbox');
```
Astro-Han, line 166: [P2] This creates new inputs to the packaged Windows verification that are not covered by `release-windows-check.yml`’s path filter. A change only to `verify-windows-sandbox-e2e.mjs`, the Rust launcher/Cargo lock, or the native smoke helpers will not rebuild and verify the release artifact. The W0 workflow covers native debug evidence, but it is not a substitute for the packaged release path. Please add the dynamically imported helper and the native packaging inputs to the release workflow trigger.

packages/runtime/src/sandbox/windows-profile.ts
```diff
@@ -0,0 +1,151 @@
...
+    readRoots.push(canonicalCwd);
+    exactReadRoots.push(canonicalCwd);
+  }
+
+  return {
+    readRoots,
+    writeRoots,
+    exactReadRoots,
+    exactWriteRoots,
+    network: profile.network.kind,
```
Astro-Han, line 86: [P2] The backend currently claims it can enforce any managed restricted-filesystem profile and carries `network: enabled` into the approved digest, but the native AppContainer launch always uses `CapabilityCount: 0`. An enabled profile is therefore accepted and silently executed with restricted networking. The filesystem worker currently narrows its own launch to `restricted`, so this is not exercised by that caller today, but it leaves the backend contract unsound for the next caller. Please reject `network.enabled` explicitly until the corresponding capability policy is implemented.

scripts/package-windows-x64.mjs
```diff
@@ -66,6 +84,9 @@ export async function packageWindowsX64({
 
   await run('npm', ['run', 'clean']);
   await run('npm', ['run', 'build']);
+  await run('cargo', ['build', '--manifest-path', sandboxManifestPath, '--release', '--locked']);
```
Astro-Han, line 87: [P2] This statically builds serde, serde_json, sha2, windows-sys, and their transitive Cargo dependencies into a distributed executable, but the current third-party notice generator inventories only npm dependencies and the packaged license resources contain no Rust notice set. Before shipping this binary, please record the Cargo dependency source/license inventory, generate the required attribution texts, package them with the application, and make the release check verify that they stay current.

Astro-Han (commented):
Thanks for carrying this through several substantial review rounds. The core direction now looks coherent: the Runtime remains the policy owner, the AppContainer and Job boundaries are applied atomically, and the production path fails closed rather than falling back to unsandboxed execution.

After rechecking the current head, I still think this should stop before merge. The two blocking findings share one underlying invariant: before a child starts under the stable AppContainer identity, the broker must know exactly which filesystem authority that identity has, and it must be able to prove that all authority from earlier launches has been removed.

The current implementation does not yet preserve that invariant:

- the packaged runtime support root can expand to the parent of the application directory and become a recursive ACL grant;
- stale-ledger recovery quarantines state it cannot reconcile, then continues launching with the same stable SID.

I left focused inline comments for those blockers and for the associated ACL-lock, process-drain, Grep, protocol-evidence, packaging, licensing, and network-contract issues. The existing named-pipe timeout thread also remains valid: the shipped `--broker-local` path still traverses the synchronous pipe implementation.

GitHub currently reports this head as mergeable and all checks are green, but the Windows E2E uses a deliberately shallow synthetic runtime layout, and the MJS “closed protocol” test rejects the actual production manifest while still passing. Those checks therefore do not close the two blockers above.

My recommendation is **stop and revise**. I would keep the AppContainer product path together, make the ACL authority and lifecycle closed first, and simplify around it by deleting the stale MJS protocol/probe authority and avoiding a named pipe in the same-process one-shot path.

Before merge, the distributed Rust dependencies also need to be covered by the release license/provenance inventory. The final squash commit should preserve the applicable `Generated-by: Claude Code` trailer, using the actual contributing tool name, and the repository’s required independent human security review remains outstanding.

AI-assisted review disclosure: Codex coordinated five read-only adversarial passes and re-verified the cited code paths; Claude Code performed an independent cross-boundary review. This remains unverified AI-assisted output until the human reviewer independently checks the cited evidence, and it does not replace the required human security review.

<details>
<summary>中文版</summary>

感谢持续推进这轮复杂修改。核心方向已经合理：Runtime 仍是策略权威，AppContainer 与 Job 在创建进程时原子应用，也没有不受控的 fallback。

不过当前仍建议在合并前停下来修正。两个阻断项实际来自同一个不变量：任何 child 使用稳定 AppContainer 身份启动前，broker 必须精确知道该身份拥有哪些文件权限，并能证明旧 launch 的权限已经全部撤销。

当前有两个 P1：

- 产品真实的 runtime support root 会扩大成应用目录的父目录，并被递归写入 ACL；
- stale ledger 无法恢复时只做 quarantine，随后继续使用同一个稳定 SID。

相关的 ACL lock、进程 drain、Grep、协议证据、打包、license 和 network contract 问题已分别放在 inline。已有 named-pipe timeout finding 也仍然成立，因为生产 `--broker-local` 路径确实经过同步 pipe 实现。

GitHub 虽然显示 mergeable 且 CI 全绿，但 Windows E2E 使用的是刻意构造的浅层 runtime 目录，MJS “closed protocol” 测试也会拒绝真实 manifest，因此这些绿色检查不能关闭上述阻断项。

建议是 **stop and revise**：保留 AppContainer 主产品路径，先闭合 ACL authority/lifecycle，再删除 stale MJS 权威，并简化掉同进程 one-shot named pipe。

</details>

experiments/windows-sandbox/launcher/src/broker_pipe.rs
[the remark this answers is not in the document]
Astro-Han: I rechecked this against the current production call chain, and I do not think the scope rationale above holds. The Runtime launches `--broker-local`; `main.rs` dispatches that to `broker_client::run_local`, which starts `broker_pipe::serve_once`, performs the client exchange, and then unconditionally joins the server thread. The shipped local path therefore does execute the same synchronous `ConnectNamedPipe`/`ReadFile`/`WriteFile` code.

The outer filesystem-worker timeout eventually kills the broker, so I would keep this at P2 rather than treating it as an unbounded permanent hang. It still means a failed second manifest read, or a peer that stops mid-frame, can stall the whole operation until that outer timeout. This thread should remain open until the exchange has its own deadline/cancellation, or the same-process one-shot path stops using a named pipe.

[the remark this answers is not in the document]
liugddx: Addressed in 29e220978. The shipped `--broker-local` path now reads and removes the manifest, validates and authorizes it, and calls `launch_appcontainer` directly in the same process. It no longer starts `serve_once`, connects to a pipe, performs a synchronous exchange, or joins a server thread. The named-pipe client/server modes remain isolated transport experiments only.

packages/runtime/src/sandbox/windows-profile.ts
liugddx: Addressed in 29e220978. On Windows the worker runtime root is now exactly `dirname(program)`, rather than its parent, so an installed `...\Programs\Maka\Maka.exe` grants only the product-owned `Maka` directory. The launch-spec test and Windows E2E helper now use that realistic topology and explicitly assert that the sibling `Programs` parent is not admitted.

experiments/windows-sandbox/launcher/src/acl_ledger.rs
liugddx: Addressed in 29e220978 by removing the stable identity invariant. Each launch now creates a request-derived AppContainer profile/SID and fails closed instead of deriving/reusing an existing profile. A quarantined invalid, unsupported, or unrecoverable ledger therefore cannot grant authority to a later child. Valid stale recovery also removes the abandoned profile registration.

experiments/windows-sandbox/launcher/src/acl_ledger.rs
liugddx: Addressed in 29e220978. Ledger v2 records `read`, `write`, `readRecursive`, and `writeRecursive` as the actual applied modes. Grant, normal cleanup, and crash recovery all consume those persisted facts; exact directory/cwd metadata grants no longer turn into recursive `/T` cleanup.

experiments/windows-sandbox/launcher/src/acl_ledger.rs
liugddx: Addressed in 29e220978. The PID lock file was replaced with OS-owned named mutexes. A short global mutex covers only recovery/grant/cleanup mutation, while each launch holds a request-specific kernel lease through child settlement. Recovery skips a busy lease and can acquire an abandoned one after a crash, so unrelated launches remain concurrent. The Windows recovery smoke includes ten parallel launches.

experiments/windows-sandbox/launcher/src/windows_launcher.rs
liugddx: Addressed in 29e220978. Error/timeout paths now call `TerminateJobObject`, wait for the root handle, and poll `QueryInformationJobObject` until `ActiveProcesses == 0` before returning to ACL cleanup. Successful root exit also gets a bounded Job-drain check; remaining descendants are terminated and the launch fails closed.

packages/runtime/src/filesystem-worker/operations.ts
liugddx: Addressed in 29e220978. Windows in-process Grep now applies the same contained-glob validation and canonicalizes every enumerated file, rejecting any candidate outside the approved search root before opening it.

packages/runtime/src/filesystem-worker/operations.ts
liugddx: Addressed in 29e220978. The Windows fallback accepts only bounded literal patterns (escaped regex metacharacters are treated literally), so it cannot enter JS regexp backtracking. Enumeration is incremental and files are streamed line-by-line with deadline, file-count, file-byte, line-byte, pattern-length, per-file-match, and result limits.

experiments/windows-sandbox/protocol.mjs
liugddx: Addressed in 29e220978. The stale MJS protocol/parser and probe plus their tests were deleted. The W0 workflow now declares the shipped TypeScript and Rust manifest structs/tests as the protocol authorities instead of presenting a parallel parser as evidence.

scripts/verify-windows-x64.mjs
liugddx: Addressed in 29e220978. `release-windows-check.yml` now includes `verify-windows-sandbox-e2e.mjs`, the Rust launcher/Cargo inputs, native PowerShell smoke helpers, and the Cargo notice generator/output in its path filter, so changes to packaged sandbox inputs trigger release verification.

packages/runtime/src/sandbox/windows-profile.ts
liugddx: Addressed in 29e220978. `canEnforceProfile` now requires restricted networking, the TypeScript policy compiler rejects `network.enabled`, and the Rust AppContainer launcher independently rejects it. No enabled-network profile can be silently narrowed to zero capabilities.

scripts/package-windows-x64.mjs
liugddx: Addressed in 29e220978. A locked-Cargo-metadata generator now produces the packaged `resources/licenses/cargo/THIRD_PARTY_NOTICES.txt` with dependency source/license data and license texts. Windows packaging runs `--check`, packaged-app verification requires the notice, and release path filters cover both generator and generated inventory.

liugddx:
Addressed the latest Astro-Han review on final head `c6b5503b4`.

- P1: narrowed the packaged Windows runtime grant to the product-owned application directory; replaced the stable AppContainer SID with strict per-launch identities so quarantined state cannot carry authority into a later child.
- ACL lifecycle: ledger v2 records the actual applied access/recursion modes; OS-owned global mutation mutex plus per-launch kernel leases preserve concurrency without recovering active ledgers; Job trees are terminated and drained before ACL cleanup.
- Grep/protocol: Windows fallback now enforces glob containment, streams bounded literal matching, and rejects unsafe regex/input shapes; stale MJS protocol/probe authority was deleted.
- Product contract: `network.enabled` fails closed; shipped `--broker-local` authorizes and launches directly without the synchronous named-pipe round trip.
- Release provenance: Windows path filters cover native/package inputs; locked Cargo metadata generates a packaged Rust notice inventory with cross-platform deterministic ordering and LF normalization.

CI follow-ups were fixed in `3890c316b`, `6755aba43`, and `6fe0bf75f`. On the final head, Windows W0, ACL/crash recovery, packaged installer/ZIP verification, sandbox E2E, pinned-version upgrade/uninstall, general E2E, typecheck, workspace tests, runtime-host tests, and the remaining repository checks all pass.

All 12 Astro-Han threads have individual implementation replies and are resolved. The repository's required independent human security review remains outstanding; this CI result does not replace it.

liugddx:
@Astro-Han, thank you for the exceptionally thorough review. Your checks around AppContainer identity, ACL lifecycle and concurrency, Job draining, Grep boundaries, and release provenance helped uncover and close several important issues.

The findings have now been addressed, with an implementation note added to every inline thread. The current final head is `c6b5503b4`, and Windows W0, ACL/crash recovery, release packaging, sandbox E2E, pinned-version upgrade/uninstall, and the general CI suite are all passing.

When you have time, could you please review the latest code once more, with particular attention to whether the security invariants and their implementations now meet the requirements for merge? Thank you again.

<details>
<summary>中文版</summary>

@Astro-Han，感谢你这轮非常细致的 review，尤其是对 AppContainer identity、ACL 生命周期与并发、Job drain、Grep 边界和 release provenance 的检查，确实帮助我们发现并补上了几个关键问题。

相关问题现已处理完成，每条 inline thread 都附上了对应实现说明。当前最终 HEAD 为 `c6b5503b4`，Windows W0、ACL/crash recovery、release package、sandbox E2E、固定版本升级/卸载以及通用 CI 均已通过。

麻烦你有空时再帮忙 review 一轮最新代码，重点确认前述安全不变量及其实现是否已经满足合并要求。再次感谢！

</details>

Astro-Han (commented):
Thanks again for the substantial work and thoughtful follow-up on this PR. After looking at the remaining findings together, I think the clean path is more likely to converge than another round of local patches.

Splitting the work is optional and entirely your call. I would be comfortable with one PR if the final diff converges on the architecture below; multiple PRs may only make the transition easier to review.

The central problem is that the approved policy, native object admission, kernel grant, operation semantics, lifecycle cleanup and release evidence currently have overlapping authorities. That is why several individually reasonable fixes have created new edge cases: exact writes became recursive parent grants, regex safety became a second literal-only search implementation, the local path bypassed the pipe while retaining pipe-era authorization state, and reparse defense became a path-based tree scan without closing hard-link and file-identity semantics.

I suggest converging on this ownership model:

```mermaid
flowchart LR
    R["Runtime Host<br/>policy authority"]
    P["Pure WindowsExecutionPlan<br/>exact · parent-entry · subtree · runtime-read"]
    N["Native admission<br/>handles · file IDs · link count · case/reparse checks"]
    L["LaunchLease<br/>profile · ledger · ACL · Job · cleanup"]
    W["Filesystem worker<br/>shared operation contract"]
    S["In-process ripgrep-compatible<br/>SearchEngine"]

    R --> P --> N --> L --> W --> S
```

The important properties would be:

- Runtime produces a pure, side-effect-free plan; diagnostics do not materialize manifests.
- The native layer binds paths to handles and object identities before granting anything.
- Exact object access, parent directory-entry rights and recursive subtree access remain distinct capabilities.
- ACL mutation and recovery are keyed by the admitted object identity, not by re-resolving path strings.
- One `LaunchLease` owns the AppContainer profile, cross-session lease, ledger, Job settlement and ACL cleanup.
- Failure to prove the Job empty preserves recovery state rather than entering normal cleanup.
- The local parent/child path uses one framed inherited channel, so temporary manifests and pipe-era PID/nonce/replay state disappear.
- Windows search uses an in-process, linear-time, ripgrep-compatible engine behind the existing worker seam and shares conformance fixtures with the normal ripgrep path.
- The shipped binary contains only this production path; rejected candidates and probes move to an evidence-only target or are removed.

This keeps the parts that are already strong—per-launch AppContainer identity, atomic Job membership, explicit handle inheritance, restricted networking, the shared filesystem-worker protocol and crash recovery—but replaces the seams that are currently producing repeated review findings.

I would treat the following as the architectural completion criteria:

1. the kernel grant is never broader than the typed plan;
2. hard links, aliases, case-sensitive directories and replacement races cannot change the admitted object;
3. unsupported capabilities fail closed instead of being approximated;
4. every launch artifact has one lifecycle owner;
5. only one production launcher authority ships;
6. release evidence executes the actual packaged launcher and worker path.

This is a larger refactor than the narrow fixes, but it should remove more code than it adds and give each future issue exactly one owner. That is the point at which I would expect this work to stop generating new compensating paths and genuinely converge.

AI-assisted review disclosure: Codex coordinated independent reviews across the native ACL/identity lifecycle, Job settlement, broker transport, Runtime policy, filesystem worker, Windows path semantics, testing, packaging and simplification. The architecture above is a synthesis of those findings and still requires the repository’s independent human security review.

<details>
<summary>中文说明</summary>

感谢这轮大量实现和持续修正。综合剩余问题后，我认为继续增加局部补丁仍可能产生新的边界问题，围绕最终状态做一次干净收敛更合适。

拆不拆 PR 由作者决定，不是本次 Review 的核心。核心是建立唯一职责：Runtime 生成纯权限计划；native 层通过 handle、file ID、link count、case/reparse 检查绑定真实对象；exact、parent-entry 和 subtree 权限始终分开；一个 LaunchLease 负责 profile、跨 session lease、ledger、ACL、Job 和 cleanup；搜索通过统一 SearchEngine seam 保持 ripgrep 合同；发布 binary 只保留真实产品路径。

这样可以保留已经正确的 AppContainer、atomic Job、stdio、network deny 和 recovery，同时删除 temp manifest、pipe-era PID/nonce/replay、旧 candidate 和多套搜索语义。完成标准不是增加多少测试，而是批准策略、内核权限、产品声明和 packaged evidence 最终描述同一个系统。

</details>

liugddx:
@Astro-Han, thank you for synthesizing the remaining concerns into a clearer ownership model. I agree with the direction, but I want to stop the PR from growing through another open-ended refactor round.

I propose the following narrow merge boundary for #2961:

- packaged Windows 11 x64 AppContainer enforcement for the Runtime filesystem-worker preview only;
- no general-command/W2 support and no claim that Windows is generally supported;
- unsupported permission shapes fail closed rather than being approximated.

Within this PR, I would address only the remaining invariants that directly affect that shipped slice:

1. ACL/ledger cleanup must run only after the Job is proven empty. If Job settlement cannot prove that, the recovery state and grants are preserved instead of entering normal cleanup.
2. Exact writes or directory-entry mutations must not be widened into recursive parent-directory grants. For this W0/W1 slice, write shapes that require parent-entry authority will fail closed rather than adding a new generalized capability model.
3. The release evidence must continue to execute the actual packaged local launcher + filesystem-worker path.

I would defer the broader architecture to #2142 follow-ups: a generalized `WindowsExecutionPlan`, handle/file-ID keyed admission and recovery, the complete hard-link/case/replacement-race matrix, a full `LaunchLease` abstraction, local transport redesign, a unified ripgrep-compatible `SearchEngine`, private desktop/readiness work, and W2/W3/general-command support.

The previous concrete P1s remain closed by the current head: the runtime grant is narrowed to the product-owned application directory, and every launch uses a fresh AppContainer identity.

Could you confirm whether the three in-scope items above are an acceptable merge gate for this preview slice, or point to a concrete current-head safety blocker that must remain in #2961? I will hold further implementation until that boundary is clear so the PR can converge rather than expand.

<details>
<summary>中文说明</summary>

感谢把剩余问题整理成清晰的职责模型。整体方向认可，但为了避免 #2961 再进入开放式重构，我建议把本 PR 的合并边界限定为：

- 仅交付 Windows 11 x64 安装包内 Runtime filesystem-worker 的 AppContainer 预览闭环；
- 不包含通用命令/W2，也不宣称 Windows 已全面支持；
- 无法精确表达的权限直接 fail closed，不做近似扩大。

本 PR 合并前只处理三个与当前交付切片直接相关的事项：

1. 只有证明 Job 已清空后才清理 ACL/ledger；无法证明时保留 recovery state 和 grant。
2. exact write 或目录项修改不得扩大成父目录递归授权；W0/W1 中需要 parent-entry 权限的写操作直接 fail closed。
3. release evidence 继续执行真实打包的 local launcher + filesystem-worker 路径。

`WindowsExecutionPlan`、handle/file-ID admission、完整 hard-link/case/race 矩阵、完整 `LaunchLease` 抽象、local transport 重构、统一 SearchEngine、private desktop/readiness，以及 W2/W3/通用命令均转回 #2142 后续。

请确认以上是否可以作为该 preview slice 的合并门槛；如果仍有必须留在 #2961 的问题，也请指出具体的 current-head safety blocker。边界确认前先不继续追加实现，避免 PR 再次膨胀。

</details>

Astro-Han:
Thanks — this is a constructive boundary, and I agree that #2961 should not become an open-ended Windows architecture rewrite.

I’m aligned with the first two merge gates:

1. ACL/ledger cleanup only after the Job is proven empty.
2. W0/W1 write shapes that cannot be represented exactly must fail closed.

I also agree that the generalized `WindowsExecutionPlan`, full `LaunchLease`, complete race matrix, transport redesign, and W2/W3 support can remain in #2142.

I don’t think the three gates are quite sufficient yet, though, because a few current-head issues affect the narrowly stated filesystem-worker preview rather than only the future generalized design:

- **Hard links:** the current ACL admission rejects reparse points but not multiple hard links. Granting through an admitted alias changes the DACL of the same underlying file object, including aliases outside the declared root. This is a current enforcement-boundary issue, not merely a missing completeness matrix. A narrow solution for this PR could simply fail closed on multi-link objects; handle/file-ID-based admission can remain follow-up work.
- **Cross-session ownership:** the mutex/lease is session-local (`Local\`), while the profile, ledger, and filesystem grants are shared across sessions. Concurrent console/RDP sessions can therefore disagree about whether a live grant is stale. This needs either a correctly secured machine-wide lease or an explicit fail-closed single-session boundary.
- **Grep contract:** the preview currently advertises the existing regex/ripgrep behavior, while the Windows implementation is literal-only with different glob/truncation behavior. The unified SearchEngine can be deferred, but the preview should either preserve the public contract or not expose Grep yet.
- **Packaged evidence:** I agree with the third gate if “actual packaged path” explicitly means the packaged Electron executable and packaged filesystem-worker resource—not the repository `dist` worker launched through a copied `node.exe`.

The earlier runtime-root and stable-identity P1s do appear closed. With the remaining preview-specific points above addressed, I’d be comfortable deferring the broader clean architecture to #2142 rather than expanding this PR further.

One final process gate remains: because this is a material security boundary, the repository policy still requires independent human security review before merge.

<details>
<summary>中文版</summary>

谢谢，这个边界划分很有建设性。我也同意不应让 #2961 继续演变成没有终点的 Windows 架构重写。

我认可前两个合并条件：只有 Job 已证明为空时才能清理 ACL/ledger；W0/W1 无法精确表达的写入必须 fail closed。通用 `WindowsExecutionPlan`、完整 `LaunchLease`、完整竞态矩阵、传输层重构和 W2/W3 支持都可以留到 #2142。

不过，目前三个条件还不足以覆盖这个缩小后的 filesystem-worker preview：硬链接仍可让根内别名改变根外同一文件对象的 DACL；session-local 租约与跨 session 共享的授权状态不一致；Windows Grep 的实际 literal-only 行为仍不符合公开的 regex/ripgrep 契约；打包验证也需要明确执行真正打包后的 Electron worker，而不是仓库 `dist` 和复制的 `node.exe`。

这些问题可以用窄范围的 fail-closed 方案解决，不要求在本 PR 完成完整架构。此前 runtime root 过宽和稳定身份复用的问题确实已经关闭。处理完上述 preview 范围内的问题后，我支持把更广泛的干净架构留到 #2142。

最后，因为这是实质性的安全边界，合并前仍需要独立的人类安全审查。

</details>

_AI-assisted review: Codex re-checked the current PR head, the author’s scope proposal, and the previously reported enforcement, lifecycle, product-contract, and packaged-evidence findings. The human contributor remains responsible for the final review and merge decision._

liugddx:
@Astro-Han Thank you — accepting all four preview-scope points as merge gates for #2961. Each is now addressed on head `7de7b5c55` with the narrow fail-closed shape you suggested, none by growing a new abstraction. Implementing your fourth point immediately caught a real product defect (details in 4 below), which I think settles whether these gates belonged in this PR.

1. **Hard links** (`c1978b67c`): ACL admission now fails closed on any multi-link file. `collect_roots` queries the kernel link count (`GetFileInformationByHandle` / `nNumberOfLinks`) for every file root, and the recursive-tree scan that already rejected reparse points now also rejects multi-link files anywhere inside a recursively granted tree — an `(OI)(CI)` grant propagates onto existing children, so an in-tree alias of an out-of-tree object would otherwise carry the grant past the root. Exact directory roots stay out of scope (their grant never leaves the directory object), mirroring the existing reparse rule. Handle/file-ID keyed admission remains #2142.

2. **Cross-session ownership** (`c1978b67c`): the mutex and per-request leases moved from `Local\` to the machine-wide namespace, since the ledger directory, the icacls grants and the AppContainer profiles are all cross-session state. The objects are user-SID-scoped names carrying an explicit SYSTEM+owner-only DACL (the same SDDL the broker pipe already uses), so a concurrent console/RDP session of the same user sees a live lease as live, another principal can neither open nor observe the lock, and a squatted name fails the launch closed instead of sharing an arbitration object with an untrusted owner.

3. **Grep contract** (`98bef67cd`): the preview no longer exposes Grep inside the Windows sandbox. The literal-only in-process fallback is deleted; a sandboxed worker fails closed with `grep_unavailable`, pointing callers at Glob and Read. The signal is a broker-injected manifest environment marker, so an unsandboxed Windows worker keeps the full ripgrep behavior and no other platform changes. The contract-preserving SearchEngine remains #2142.

4. **Packaged evidence** (`263e73f1a`, `7de7b5c55`): your suspicion was correct in the strongest sense — the packaged Electron worker path had never actually executed inside the AppContainer. Electron's run-as-node entry opens the NUL device to backfill standard handles, the container denies that open, and startup aborts (`FATAL node_main.cc "Unable to open nul device"`) — a failure the copied-`node.exe` evidence could never observe, because plain node performs no such backfill. Fixed by passing `--no-stdio-init` (the broker always relays three valid standard handles, so the backfill is unnecessary). The release E2E now executes the packaged artifacts themselves: the packaged broker enforcing the AppContainer, the packaged Electron executable as the worker runtime (`ELECTRON_RUN_AS_NODE`, the production launch shape) and the packaged `resources\workers\filesystem-worker.js` bundle. Only the driver still imports repository-built client code, because the packaged copy lives inside app.asar. The spec assertions pin the packaged program/bundle paths and the not-widened runtime root.

Verification on the final head: cargo tests including the new admission/lock coverage (multi-link file root, in-tree out-of-tree alias, machine-wide lock naming), the local smoke suite including 10-way concurrent ACL recovery under the machine-wide locks, the Windows workspace suites, a live brokered-Electron AppContainer run of write/read/glob/grep-fail-closed/denied-write, and CI is green across all 16 checks — including `release-windows-check`, whose E2E now runs the packaged Electron worker end to end.

Agreed on the process gate: independent human security review remains required before merge, and this CI evidence does not substitute for it.

<details>
<summary>中文版</summary>

四个 preview 范围内的问题全部按你建议的窄范围 fail-closed 方案在 head `7de7b5c55` 落地：

1. **硬链接**（`c1978b67c`）：准入对文件 root 及递归授权树内所有文件查内核链接数，多链接对象直接 fail closed；exact 目录 root 与既有 reparse 规则对齐不扫描。file-ID 化准入留 #2142。
2. **跨会话归属**（`c1978b67c`）：mutex/lease 迁移到机器级命名空间，名字按用户 SID 区分并带 SYSTEM+属主-only 显式 DACL（复用 broker pipe 的 SDDL）；同用户跨会话能正确看到活跃 lease，他人抢名则启动 fail closed。
3. **Grep 契约**（`98bef67cd`）：preview 在 Windows 沙箱内不再暴露 Grep——literal-only fallback 已删除，沙箱内 worker 返回 `grep_unavailable` 并提示改用 Glob+Read；非沙箱 Windows worker 保留完整 ripgrep 行为。统一 SearchEngine 留 #2142。
4. **打包证据**（`263e73f1a`、`7de7b5c55`）：你的判断完全正确——打包 Electron worker 此前从未真正在 AppContainer 里跑通过：Electron run-as-node 入口启动时要打开 NUL 设备回填标准句柄，容器拒绝后直接 FATAL 中止，而复制 node.exe 的旧证据链观察不到这一点。已修（`--no-stdio-init`，broker 本就中继三个有效句柄），release E2E 现在实跑打包 broker + 打包 Electron 可执行文件 + 打包 worker bundle。

最终 head 上 16 项 CI 全绿（含跑真打包 worker E2E 的 `release-windows-check`）。同意流程门禁：合并前仍需独立人工安全审查，CI 证据不能替代。

</details>

packages/runtime/src/filesystem-worker/client.ts
```diff
@@ -443,7 +458,13 @@ export function filesystemWorkerRuntimeWritableRoots(input: {
   entryMode?: boolean;
   writableAncestor?: string;
 }): readonly string[] | undefined {
-  if (input.platform !== 'linux' || input.access !== 'write') return undefined;
+  // Linux mounts and Windows ACL grants can only target existing paths, so a
```
Astro-Han, line 461: **[P2] Keep exact writes exact in the Windows preview**

A missing Write target—and ApplyPatch create/delete—adds its existing parent to `runtimeWritableRoots`. The Windows compiler treats that as a non-exact write root, so the native launcher applies recursive Modify permission to the parent tree. The normal worker still writes only the requested target, which explains why the packaged E2E passes, but the kernel boundary is broader than the approved operation. This is the previously agreed parent-entry merge gate. For this preview, please fail closed on write shapes that require parent-entry authority rather than representing them as subtree Modify, and keep the packaged E2E focused on updating a pre-existing exact file.

experiments/windows-sandbox/launcher/src/acl_ledger.rs
```diff
@@ -0,0 +1,524 @@
...
+                let _ = fs::remove_file(&ledger_path);
+            }
+            return match restore {
+                Ok(()) => Err(error),
+                Err(restore) => Err(format!("{error}; ACL cleanup also failed: {restore}")),
+            };
+        }
+    }
+
+    let launch_result = launch();
```
Astro-Han: **[P2] Preserve recovery state when Job settlement is unproven**

`settle_job()` returns an ordinary error both when the Job was safely drained and when termination, waiting, or Job accounting failed. `with_acl_grants()` then performs normal ACL and ledger cleanup after either result, and the profile is dropped afterward. The happy-path E2E cannot exercise this branch, and it leaves the previously agreed lifecycle gate incomplete. Please propagate a settled-versus-unsettled outcome: only a proven-empty Job may enter normal cleanup; an unsettled launch must preserve or quarantine its recovery state. A focused failure-injection test is sufficient.

Astro-Han (approved):
Approved for the narrowed packaged Windows filesystem-worker preview. The current head runs the packaged broker, Electron executable, and worker end to end, and the earlier runtime-root, per-launch identity, ledger-mode, hard-link, cross-session lock, Grep-contract, and packaging findings are closed.

The two inline P2 notes are deliberately narrow: close the already-agreed exact-write and Job-settlement invariants without expanding this PR into a generalized Windows architecture rewrite. Please rerun the packaged Electron E2E on the final head. Independent human security review remains required before merge under the repository policy; this approval and green CI do not replace that gate.

_AI-assisted review: Codex inspected the current PR head, full diff, CI and merge state, prior review ledger, author replies, native ACL/Job lifecycle, Runtime/Core contracts, and packaged evidence, with independent adversarial reviewer passes. The human contributor remains responsible for validating these findings and for the final review and merge decision._

<details>
<summary>中文版</summary>

整体批准当前收窄后的 Windows filesystem-worker preview。两条 inline P2 仅要求用最小方案关闭此前已经接受的 exact-write 与 Job-settlement 不变量，不要求扩展为通用 Windows 架构重构。最终 head 请重跑 packaged Electron E2E；合并前仍需仓库政策要求的独立人工安全审查。

</details>

packages/runtime/src/filesystem-worker/client.ts
liugddx: Fixed in `24a099caf`. The Windows compiler now refuses any policy carrying a runtime-writable parent root instead of widening it to recursive Modify: a missing-target Write and ApplyPatch create/delete fail closed before any process launches, per the agreed W0/W1 parent-entry gate (`windows-profile.ts` throws, surfaced as `invalid_request` at the transform stage). The packaged E2E and the Windows smoke now update a pre-seeded exact file and additionally assert that a missing-target write fails closed without producing a file. `verify-windows-sandbox-e2e.mjs` also joins the release-windows-check path filter so E2E semantic changes rerun the packaged evidence.

experiments/windows-sandbox/launcher/src/acl_ledger.rs
liugddx: Fixed in `77415ed47`. `settle_job` now returns a settled-versus-unsettled outcome: only a Job proven empty (directly, or after a successful terminate-and-drain) enters normal cleanup. An unsettled launch keeps its icacls grants in place, quarantines its ledger as a preserved recovery record, and leaves the AppContainer profile undeleted — the kill-on-close Job handle remains the kernel backstop. Recovery never interprets quarantined ledgers and every launch uses a fresh identity, so the preserved authority cannot reach a later child. Failure-injection tests cover both outcomes (unsettled preserves grants + quarantines; settled still cleans up normally).

liugddx:
@Astro-Han Both inline P2 gates are closed on the final head `781d7e0f2`, and the packaged Electron E2E has been rerun there as requested:

1. **Exact writes stay exact** (`24a099caf`): the Windows compiler refuses any policy carrying a runtime-writable parent root instead of widening it to recursive Modify — a missing-target Write and ApplyPatch create/delete now fail closed at the transform stage before any process launches. The packaged E2E and the Windows smoke update a pre-seeded exact file and assert the missing-target shape fails closed without producing a file.

2. **Job settlement gates cleanup** (`77415ed47`): `settle_job` propagates a settled-versus-unsettled outcome. Only a Job proven empty (directly or after a successful terminate-and-drain) enters normal ACL/ledger cleanup; an unsettled launch preserves its grants, quarantines its ledger as the recovery record, and keeps the AppContainer profile, with the kill-on-close Job handle as the kernel backstop. Failure-injection tests pin both outcomes.

CI on the final head: 15/16 green including `release-windows-check` (packaged installer + packaged Electron worker E2E, 13m). The one red check, `request-review`, is the repository's Copilot-request automation failing with a 403 from its configured token — unrelated to this PR's content and reproducible on rerun; flagging it for maintainers.

The independent human security review gate remains acknowledged as required by repository policy.

<details>
<summary>中文版</summary>

两条 inline P2 已在最终 head `781d7e0f2` 关闭，打包 Electron E2E 也按要求在最终 head 重跑通过：

1. **exact 写保持 exact**（`24a099caf`）：Windows 编译器对带 runtime-writable 父根的策略直接拒绝，缺失目标 Write / ApplyPatch create/delete 在 transform 阶段 fail closed；打包 E2E 与冒烟改为更新预置文件并断言缺失目标写 fail closed。
2. **Job settlement 门控清理**（`77415ed47`）：只有证明清空的 Job 才进入正常 ACL/ledger 清理；unsettled 的 launch 保留 grants、隔离 ledger、保留 AppContainer profile，kill-on-close Job 句柄仍是内核兜底；注入式测试锁定双向行为。

最终 head CI 15/16 绿（含 13 分钟的打包验证）。唯一红项 `request-review` 是仓库 Copilot 自动化的 token 403，与本 PR 内容无关，已标记请维护者处理。仓库政策要求的独立人工安审仍然有效。

</details>

Issue #1679, opened 2026-07-31
[RFC] Canonical model-call accounting: one ModelCallAttempt record per real provider request


ARE404:
## Summary

Make one canonical record per **real provider request attempt** the source of truth for model-call accounting, and feed the Host-owned Usage authority (#1596) from it. Today the same spend is recorded by two independent writers at two different granularities, with three known gaps — and neither writer can answer "what did this turn actually cost, and is that number complete?"

This is the metering half of #1625. The Session Inspector stays downstream and becomes a pure read model over these records once they exist.

cc @likun666661 (#1649 sequencing) — see [Relationship to #1649](#relationship-to-1649).

## Problem

Two writers record model spend, neither completely:

| Path | Writer | Granularity | Carries cost? |
|---|---|---|---|
| Telemetry | `recordLlmCall` (`ai-sdk-backend.ts:1826`) | one record per **send**, aggregated across retries | yes |
| AgentRun ledger | `recordProviderRequestAttempt` (`agent-run.ts:292`) | one record per **physical attempt** | no |

Neither is authoritative, and the split produces three gaps:

1. **Provider retries carry no kind anywhere.** The main telemetry record aggregates across retries, and `provider_retry` is barred from the AgentRun ledger (`agent-run.ts:482`). A retried call is invisible as a distinct billable event.
2. **The `RuntimeEvent` ledger attributes no compaction cost.** Kernel compaction writes a phantom `input: 0, output: 0` turn (`runtime-kernel.ts:816-822`) while the real usage goes to telemetry. Anything projecting cost from the event ledger under-reports on exactly the turns that compacted.
3. **The `pi` backend records no telemetry at all.** `recordLlmCall` has zero references in `pi-agent-backend.ts`, so those sessions show tokens in the ledger and `$0` in telemetry — an indistinguishable-from-real zero.

Two more properties make the current state hard to build on:

- **Attempt writes drop silently.** Two lines apart in `runtime-kernel.ts:2462-2477`, `recordProviderRequestCapture` rejects when there's no active run while `recordProviderRequestAttempt` is `run?.recordProviderRequestAttempt(attempt)`. A late finalize after turn teardown vanishes with no trace.
- **Incompleteness is undetectable.** Nothing records an expected attempt count, and captures dedup on `(step, requestHash)`, so a partial record set can't even be recognized as partial — let alone bounded.

[...]

PR #2962, opened 2026-08-13
fix(eval): stabilize native timeouts, usage, and artifacts

hqhq1025:
## Root causes

This PR now carries the complete real-machine Eval reliability line.

1. Maka hosted eval sessions inherited the product 120-second model stream idle watchdog, aborting admitted DeepSeek V4 Flash calls before the Terminal-Bench native deadline.
2. External stdout/stderr were hashed in memory but not persisted, and `trial/agent` was omitted from post-finalization artifact inventory.
3. Task instructions beginning with `-` were parsed as CLI flags by several harnesses.
4. Reasonix's large tool list could exceed the 2 KiB atomic relay frame.
5. Timeout paths could lose external or Maka usage already committed to metering snapshots or `runtime.sqlite`.
6. Kimi/ZCode may not flush stdout before timeout, requiring an independent provider request/response trace. Late provider responses must remain eligible to settle inside the task-native window so usage is not discarded.

## Fixes

- Defer only Eval model stream idleness to the benchmark-native timeout while retaining connect deadlines.
- Preserve Runtime Host exit/client diagnostics and owned-host liveness policy.
- Persist bounded external stdout/stderr and credential-free provider request/response JSONL under `/logs/agent`.
- Inventory the complete `trial/agent` directory after confirmed Harbor finalization, with bytes and SHA-256.
- Keep unbounded model/tool details in the hashed metering artifact and bounded counts in the relay frame.
- Recover finalized external usage and partial Maka usage with explicit completeness/lower-bound provenance.
- Prevent dash-leading task text from entering CLI option parsing.
- Preserve late provider usage after the downstream CLI stops reading, within the native task window.

## Regression verification

All new guards were run red-green against the prior implementation: trajectory persistence, `agent/` inventory, 64 MiB bounds, dash-leading prompts, relay payload bounds, provider-events, and late usage recovery.

Current PR head:

- Eval typecheck: PASS
- Eval Node suite: 36/36 PASS
- Python relay contract: 10/10 PASS
- Python relay lifecycle: 12/12 PASS
- Python egress/policy/artifact suites: 6/6 PASS

## Real-machine validation

Terminal-Bench 2.1 revision `d49e28f1e4ddd13d289e85a5f312a66750951932`, DeepSeek V4 Flash, thinking=max, native task timeouts:

- seven-arm canary: 7/7 with score, usage, physical trajectory, and attempt bytes/SHA;
- repaired trajectory cells: 23/23 valid replacements;
- final selected coverage: 712/712 cells have usage and a physically readable trajectory tied to the scored attempt;
- manifest: `/mnt/deepswe/eval-runs/deepseek-v4-flash-final-trajectory-overlay-8a1ab4635.json`;
- missing cells: 0;
- artifact failures: 0;
- active recovery runners: 0.

The real-machine artifacts are authoritative for this timing- and finalization-sensitive path.

### DIFF
--- a/packages/eval/README.md
+++ b/packages/eval/README.md
@@ -65,8 +65,9 @@ External CLI specs terminate option parsing or bind the prompt inside one option
 that begins with `-` cannot become a CLI flag. Relay results contain only bounded metering counts;
 the complete observed model and tool-name lists remain in the hashed metering artifact. The
 provider proxy also writes a bounded credential-free request/response JSONL trace, so a timed-out
-CLI that never flushes stdout still leaves model I/O tied to its usage; downstream disconnects
-cancel the corresponding upstream request instead of holding finalization open.
+CLI that never flushes stdout still leaves model I/O tied to its usage. In-flight provider requests
+remain eligible to settle inside the benchmark-native task window after the CLI disconnects, so
+late usage is not discarded.
 The local image tag remains a machine deployment identity rather than a registry digest; digest
 pinning is tracked in issue #2953.
 
--- a/packages/eval/README.md
+++ b/packages/eval/README.md
@@ -38,8 +38,9 @@ developer prompt and complete tool schema. The profile disables Runtime's produc
 watchdog and Undici's provider headers/body inactivity timers after provider activity;
 benchmark-native subject timeouts remain the execution deadline, while the model stream connect
 timeout still rejects requests that never establish a response. Hosted Eval clients also use a
-10-second liveness cadence with a 30-second response budget so a CPU-bound one-core task cannot
-starve the Runtime Host briefly enough to be mistaken for a dead connection.
+disabled active liveness probe for their owned Runtime Host process so a CPU-bound one-core task
+cannot starve the Host briefly enough to be mistaken for a dead connection. Child-process exit,
+transport closure, and the benchmark-native subject timeout remain authoritative.
 
 Every benchmark subject removes `WebSearch`, `WebFetch`, and `FetchURL` from the provider-visible
 tool list. Maka enforces that through its Hosted Execution profile; external harnesses pass through
--- a/packages/eval/README.md
+++ b/packages/eval/README.md
@@ -61,6 +61,9 @@ and egress audit logs are represented in attempt artifacts with byte counts and 
 The external wrapper persists up to 64 MiB of each subject's stdout and stderr under `/logs/agent`;
 after Harbor finalization, Eval inventories the complete trial `agent/` directory into the attempt
 so trajectory, metering, wrapper-state, and runtime-event files remain tied to the scored result.
+External CLI specs terminate option parsing or bind the prompt inside one option value so task text
+that begins with `-` cannot become a CLI flag. Relay results contain only bounded metering counts;
+the complete observed model and tool-name lists remain in the hashed metering artifact.
 The local image tag remains a machine deployment identity rather than a registry digest; digest
 pinning is tracked in issue #2953.
 
--- a/packages/eval/README.md
+++ b/packages/eval/README.md
@@ -58,6 +58,9 @@ cohort. This URL policy is a blocklist for known benchmark and public-solution c
 surfaces, not a complete defense against a deliberately invented lookup channel; the network
 namespace still forces all subject traffic through the audited proxy. Collected Maka runtime files
 and egress audit logs are represented in attempt artifacts with byte counts and SHA-256 digests.
+The external wrapper persists up to 64 MiB of each subject's stdout and stderr under `/logs/agent`;
+after Harbor finalization, Eval inventories the complete trial `agent/` directory into the attempt
+so trajectory, metering, wrapper-state, and runtime-event files remain tied to the scored result.
 The local image tag remains a machine deployment identity rather than a registry digest; digest
 pinning is tracked in issue #2953.
 
--- a/packages/eval/README.md
+++ b/packages/eval/README.md
@@ -63,7 +63,10 @@ after Harbor finalization, Eval inventories the complete trial `agent/` director
 so trajectory, metering, wrapper-state, and runtime-event files remain tied to the scored result.
 External CLI specs terminate option parsing or bind the prompt inside one option value so task text
 that begins with `-` cannot become a CLI flag. Relay results contain only bounded metering counts;
-the complete observed model and tool-name lists remain in the hashed metering artifact.
+the complete observed model and tool-name lists remain in the hashed metering artifact. The
+provider proxy also writes a bounded credential-free request/response JSONL trace, so a timed-out
+CLI that never flushes stdout still leaves model I/O tied to its usage; downstream disconnects
+cancel the corresponding upstream request instead of holding finalization open.
 The local image tag remains a machine deployment identity rather than a registry digest; digest
 pinning is tracked in issue #2953.
 
--- a/packages/eval/README.md
+++ b/packages/eval/README.md
@@ -34,7 +34,9 @@ and exposes a foreground-only Bash schema without `run_in_background` or `pty`. 
 routing remains authoritative: DeepSeek Responses exposes `apply_patch` instead of `Write` and
 `Edit`, and Runtime-owned `ArchiveRead` remains available for archived tool results. A real
 `hosted.execution.start` regression test pins SHA-256 hashes for the first main provider request's
-developer prompt and complete tool schema.
+developer prompt and complete tool schema. The profile disables Runtime's product stream-idle
+watchdog after provider activity; benchmark-native subject timeouts remain the execution deadline,
+while the model stream connect timeout still rejects requests that never establish a response.
 
 Every benchmark subject removes `WebSearch`, `WebFetch`, and `FetchURL` from the provider-visible
 tool list. Maka enforces that through its Hosted Execution profile; external harnesses pass through
--- a/packages/eval/README.md
+++ b/packages/eval/README.md
@@ -35,8 +35,11 @@ routing remains authoritative: DeepSeek Responses exposes `apply_patch` instead 
 `Edit`, and Runtime-owned `ArchiveRead` remains available for archived tool results. A real
 `hosted.execution.start` regression test pins SHA-256 hashes for the first main provider request's
 developer prompt and complete tool schema. The profile disables Runtime's product stream-idle
-watchdog after provider activity; benchmark-native subject timeouts remain the execution deadline,
-while the model stream connect timeout still rejects requests that never establish a response.
+watchdog and Undici's provider headers/body inactivity timers after provider activity;
+benchmark-native subject timeouts remain the execution deadline, while the model stream connect
+timeout still rejects requests that never establish a response. Hosted Eval clients also use a
+10-second liveness cadence with a 30-second response budget so a CPU-bound one-core task cannot
+starve the Runtime Host briefly enough to be mistaken for a dead connection.
 
 Every benchmark subject removes `WebSearch`, `WebFetch`, and `FetchURL` from the provider-visible
 tool list. Maka enforces that through its Hosted Execution profile; external harnesses pass through
--- a/packages/eval/src/harbor-external-subject.ts
+++ b/packages/eval/src/harbor-external-subject.ts
@@ -19,7 +19,7 @@ const resultToken = takeRelayResultToken();
 
 type SubjectStatus = 'completed' | 'failed' | 'infra_failed' | 'indeterminate';
 const CLASSIFIABLE_RECORD_LIMIT_BYTES = 16 * 1024 * 1024;
-const PERSISTED_STREAM_LIMIT_BYTES = 64 * 1024 * 1024;
+const PERSISTED_STREAM_LIMIT_BYTES = takePersistedStreamLimit();
 
 interface Usage {
   inputTokens: number;
@@ -28,6 +28,17 @@ interface Usage {
   cacheWriteTokens: number;
   reasoningTokens: number;
   totalTokens: number;
+}
+
+function takePersistedStreamLimit(): number {
+  const override = process.env.MAKA_EVAL_TEST_PERSISTED_STREAM_LIMIT_BYTES;
+  delete process.env.MAKA_EVAL_TEST_PERSISTED_STREAM_LIMIT_BYTES;
+  if (override === undefined) return 64 * 1024 * 1024;
+  const parsed = Number(override);
+  if (!Number.isSafeInteger(parsed) || parsed <= 0 || parsed > 64 * 1024 * 1024) {
+    throw new Error('invalid external trajectory persistence limit');
+  }
+  return parsed;
 }
 
 const [rawProfile, baseUrl, systemRoot, command, ...args] = process.argv.slice(2);
--- a/packages/eval/src/harbor-external-subject.ts
+++ b/packages/eval/src/harbor-external-subject.ts
@@ -51,6 +51,7 @@ process.once('SIGINT', interrupt);
 
 const logsRoot = rooted(systemRoot, '/logs/agent');
 const statePath = join(logsRoot, `${profile}.wrapper-state.json`);
+const meteringPath = join(logsRoot, `${profile}.metering.json`);
 const artifacts: Record<string, unknown>[] = [];
 let usage: Usage | null = null;
 let costUsd: number | null = null;
@@ -70,7 +71,7 @@ try {
   delete process.env.ANTHROPIC_API_KEY;
   delete process.env.DEEPSEEK_API_KEY;
   if (!key) throw new Error('external subject credential is missing');
-  const proxy = await startMeteringProxy(baseUrl, key, profile === 'claude-code');
+  const proxy = await startMeteringProxy(baseUrl, key, profile === 'claude-code', meteringPath);
   try {
     await writeState('proxy_started');
     const prepared = await prepareProfile(profile, proxy.baseUrl, systemRoot, command, args);
@@ -112,6 +113,7 @@ try {
         toolNames: proxy.observedToolNames(),
       },
       fileArtifact('wrapper-state', statePath, profile),
+      fileArtifact('provider-metering-snapshot', meteringPath, profile),
     );
   } finally {
     await proxy.close();
@@ -588,6 +590,7 @@ async function startMeteringProxy(
   upstreamBaseUrl: string,
   upstreamKey: string,
   anthropic: boolean,
+  meteringPath: string,
 ): Promise<{
   baseUrl: string;
   usage(): Usage | null;
@@ -608,6 +611,22 @@ async function startMeteringProxy(
   let removedWebTools = 0;
   const requestModels = new Set<string>();
   const observedToolNames = new Set<string>();
+  let snapshotWrite = Promise.resolve();
+  const persistSnapshot = () => {
+    const snapshot = `${JSON.stringify({
+      schemaVersion: 1,
+      profile,
+      usage: measured ? { ...total, totalTokens: total.inputTokens + total.outputTokens } : null,
+      costUsd: measured ? estimateCost(total) : null,
+      requests,
+      admittedRequests,
+      usageRequests,
+      usageComplete: admittedRequests > 0 && usageRequests === admittedRequests,
+    })}\n`;
+    snapshotWrite = snapshotWrite.then(() => writeFile(meteringPath, snapshot, { mode: 0o600 }));
+    return snapshotWrite;
+  };
+  await persistSnapshot();
   const providerTransport = createExternalProviderDispatcher(process.env.HTTPS_PROXY);
   const active = new Set<Promise<void>>();
   const server = createServer((request, response) => {
@@ -666,6 +685,7 @@ async function startMeteringProxy(
             addUsage(total, parsed.usage);
           }
         }
+        await persistSnapshot();
       }
     })().catch((error: unknown) => {
       response.statusCode = 502;
@@ -690,6 +710,7 @@ async function startMeteringProxy(
     observedToolNames: () => [...observedToolNames].sort(),
     close: async () => {
       await Promise.allSettled([...active]);
+      await snapshotWrite;
       await closeServer(server);
       await providerTransport.close();
     },
--- a/packages/eval/src/harbor-external-subject.ts
+++ b/packages/eval/src/harbor-external-subject.ts
@@ -53,6 +53,7 @@ process.once('SIGINT', interrupt);
 const logsRoot = rooted(systemRoot, '/logs/agent');
 const statePath = join(logsRoot, `${profile}.wrapper-state.json`);
 const meteringPath = join(logsRoot, `${profile}.metering.json`);
+const providerEventsPath = join(logsRoot, `${profile}.provider-events.jsonl`);
 const stdoutPath = join(logsRoot, `${profile}.jsonl`);
 const stderrPath = join(logsRoot, `${profile}.stderr.txt`);
 const artifacts: Record<string, unknown>[] = [];
@@ -74,7 +75,13 @@ try {
   delete process.env.ANTHROPIC_API_KEY;
   delete process.env.DEEPSEEK_API_KEY;
   if (!key) throw new Error('external subject credential is missing');
-  const proxy = await startMeteringProxy(baseUrl, key, profile === 'claude-code', meteringPath);
+  const proxy = await startMeteringProxy(
+    baseUrl,
+    key,
+    profile === 'claude-code',
+    meteringPath,
+    providerEventsPath,
+  );
   try {
     await writeState('proxy_started');
     const prepared = await prepareProfile(profile, proxy.baseUrl, systemRoot, command, args);
@@ -669,6 +676,7 @@ async function startMeteringProxy(
   upstreamKey: string,
   anthropic: boolean,
   meteringPath: string,
+  eventsPath: string,
 ): Promise<{
   baseUrl: string;
   usage(): Usage | null;
@@ -690,6 +698,7 @@ async function startMeteringProxy(
   let removedWebTools = 0;
   const requestModels = new Set<string>();
   const observedToolNames = new Set<string>();
+  const events = await createBoundedJsonlWriter(eventsPath, PERSISTED_STREAM_LIMIT_BYTES);
   let snapshotWrite = Promise.resolve();
   const persistSnapshot = () => {
     const snapshot = `${JSON.stringify({
@@ -712,12 +721,20 @@ async function startMeteringProxy(
   const providerTransport = createExternalProviderDispatcher(process.env.HTTPS_PROXY);
   const active = new Set<Promise<void>>();
   const server = createServer((request, response) => {
+    const requestId = requests + 1;
     const operation = (async () => {
       requests += 1;
       const projected = removeEvalWebTools(await readRequest(request));
       removedWebTools += projected.removed;
       if (projected.model) requestModels.add(projected.model);
       for (const name of projected.toolNames) observedToolNames.add(name);
+      await events.write({
+        type: 'provider_request',
+        requestId,
+        method: request.method ?? 'GET',
+        path: request.url ?? '/',
+        bodyBase64: projected.body.toString('base64'),
+      });
       const target = joinUpstream(upstreamBaseUrl, request.url ?? '/');
       const headers: Record<string, string> = {};
       for (const [name, value] of Object.entries(request.headers)) {
@@ -732,11 +749,29 @@ async function startMeteringProxy(
       }
       if (anthropic) headers['x-api-key'] = upstreamKey;
       else headers.authorization = `Bearer ${upstreamKey}`;
+      const upstreamAbort = new AbortController();
+      const abortUpstream = () => {
+        if (!upstreamAbort.signal.aborted) {
+          upstreamAbort.abort(new Error('provider proxy downstream disconnected'));
+        }
+      };
+      const closeUpstream = () => {
+        if (!response.writableEnded) abortUpstream();
+      };
+      request.once('aborted', abortUpstream);
+      response.once('close', closeUpstream);
       const upstream = await undiciFetch(target, {
         method: request.method,
         headers,
         body: projected.body.length === 0 ? undefined : new Uint8Array(projected.body),
         dispatcher: providerTransport.dispatcher,
+        signal: upstreamAbort.signal,
+      });
+      await events.write({
+        type: 'provider_response_start',
+        requestId,
+        status: upstream.status,
+        contentType: upstream.headers.get('content-type'),
       });
       response.statusCode = upstream.status;
       upstream.headers.forEach((value, name) => {
@@ -752,12 +787,20 @@ async function startMeteringProxy(
           for (;;) {
             const next = await reader.read();
             if (next.done) break;
-            response.write(Buffer.from(next.value));
-            parser.push(Buffer.from(next.value).toString('utf8'));
+            const bytes = Buffer.from(next.value);
+            await events.write({
+              type: 'provider_response_chunk',
+              requestId,
+              bodyBase64: bytes.toString('base64'),
+            });
+            response.write(bytes);
+            parser.push(bytes.toString('utf8'));
           }
         }
         response.end();
       } finally {
+        request.removeListener('aborted', abortUpstream);
+        response.removeListener('close', closeUpstream);
         const parsed = parser.finish();
         if (upstream.ok && parsed.admitted) {
           admittedRequests += 1;
@@ -767,11 +810,24 @@ async function startMeteringProxy(
             addUsage(total, parsed.usage);
           }
         }
+        await events.write({
+          type: 'provider_response_end',
+          requestId,
+          admitted: parsed.admitted,
+          usage: parsed.usage,
+        });
         await persistSnapshot();
       }
     })().catch((error: unknown) => {
-      response.statusCode = 502;
-      response.end(JSON.stringify({ error: safeFailure(error, 'provider proxy failed') }));
+      void events.write({
+        type: 'provider_error',
+        requestId,
+        error: safeFailure(error, 'provider proxy failed'),
+      });
+      if (!response.destroyed) {
+        response.statusCode = 502;
+        response.end(JSON.stringify({ error: safeFailure(error, 'provider proxy failed') }));
+      }
     });
     active.add(operation);
     void operation.finally(() => active.delete(operation));
@@ -799,6 +855,37 @@ async function startMeteringProxy(
       await snapshotWrite;
       await closeServer(server);
       await providerTransport.close();
+      await events.close();
+    },
+  };
+}
+
+async function createBoundedJsonlWriter(
+  path: string,
+  limitBytes: number,
+): Promise<{
+  write(value: unknown): Promise<void>;
+  close(): Promise<void>;
+}> {
+  const output = await open(path, 'w', 0o600);
+  let queue = Promise.resolve();
+  let persistedBytes = 0;
+  let closed = false;
+  return {
+    write(value: unknown): Promise<void> {
+      const line = Buffer.from(`${JSON.stringify(value)}\n`);
+      queue = queue.then(async () => {
+        if (closed || persistedBytes + line.byteLength > limitBytes) return;
+        await writeAll(output, line);
+        persistedBytes += line.byteLength;
+      });
+      return queue;
+    },
+    async close(): Promise<void> {
+      await queue;
+      if (closed) return;
+      closed = true;
+      await output.close();
     },
   };
 }
--- a/packages/eval/src/harbor-external-subject.ts
+++ b/packages/eval/src/harbor-external-subject.ts
@@ -749,23 +749,11 @@ async function startMeteringProxy(
       }
       if (anthropic) headers['x-api-key'] = upstreamKey;
       else headers.authorization = `Bearer ${upstreamKey}`;
-      const upstreamAbort = new AbortController();
-      const abortUpstream = () => {
-        if (!upstreamAbort.signal.aborted) {
-          upstreamAbort.abort(new Error('provider proxy downstream disconnected'));
-        }
-      };
-      const closeUpstream = () => {
-        if (!response.writableEnded) abortUpstream();
-      };
-      request.once('aborted', abortUpstream);
-      response.once('close', closeUpstream);
       const upstream = await undiciFetch(target, {
         method: request.method,
         headers,
         body: projected.body.length === 0 ? undefined : new Uint8Array(projected.body),
         dispatcher: providerTransport.dispatcher,
-        signal: upstreamAbort.signal,
       });
       await events.write({
         type: 'provider_response_start',
@@ -799,8 +787,6 @@ async function startMeteringProxy(
         }
         response.end();
       } finally {
-        request.removeListener('aborted', abortUpstream);
-        response.removeListener('close', closeUpstream);
         const parsed = parser.finish();
         if (upstream.ok && parsed.admitted) {
           admittedRequests += 1;
--- a/packages/eval/src/harbor-external-subject.ts
+++ b/packages/eval/src/harbor-external-subject.ts
@@ -79,6 +79,7 @@ try {
     if (profile === 'claude-code') prepareClaudeWorkspace(systemRoot, prepared.home);
     const result = await runChild(prepared.command, prepared.args, prepared.env, profile);
     child = undefined;
+    await proxy.settle();
     await writeState('child_exited', { exitCode: result.exitCode });
     usage = proxy.usage();
     costUsd = usage && proxy.usageComplete() ? estimateCost(usage) : null;
@@ -601,6 +602,7 @@ async function startMeteringProxy(
   removedWebToolCount(): number;
   requestModels(): readonly string[];
   observedToolNames(): readonly string[];
+  settle(): Promise<void>;
   close(): Promise<void>;
 }> {
   const total = zeroUsage();
@@ -708,6 +710,10 @@ async function startMeteringProxy(
     removedWebToolCount: () => removedWebTools,
     requestModels: () => [...requestModels].sort(),
     observedToolNames: () => [...observedToolNames].sort(),
+    settle: async () => {
+      await Promise.allSettled([...active]);
+      await snapshotWrite;
+    },
     close: async () => {
       await Promise.allSettled([...active]);
       await snapshotWrite;
--- a/packages/eval/src/harbor-external-subject.ts
+++ b/packages/eval/src/harbor-external-subject.ts
@@ -120,8 +120,8 @@ try {
         usageRequests: proxy.usageRequestCount(),
         usageComplete: proxy.usageComplete(),
         removedWebTools: proxy.removedWebToolCount(),
-        models: proxy.requestModels(),
-        toolNames: proxy.observedToolNames(),
+        modelCount: proxy.requestModels().length,
+        toolNameCount: proxy.observedToolNames().length,
       },
       fileArtifact('wrapper-state', statePath, profile),
       fileArtifact('provider-metering-snapshot', meteringPath, profile),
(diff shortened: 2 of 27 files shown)


packages/runtime-host/src/client/connect-or-spawn.ts
```diff
@@ -43,6 +43,8 @@ export interface ConnectOrSpawnRuntimeHostInput {
   electionDeadlineMs?: number;
   connectTimeoutMs?: number;
   handshakeTimeoutMs?: number;
+  livenessIntervalMs?: number;
```
Astro-Han, line 46: **P1** — these two fields are declared here but never forwarded. The call to `connectResolvedRuntimeHost` at :186-199 lists its fields explicitly and passes only `connectTimeoutMs`, `handshakeTimeoutMs` and `electionDeadline`, so `livenessIntervalMs: 0` from `hosted-execution.ts:44` is dropped and the probe falls back to the 2s default. `c689c1e2f` therefore changes nothing in production, and the CPU-starvation misjudgment it targets is unchanged.

`connection.ts:1081` and `:1102` call the same function with `{ ...input, ...normalized }`, which is why only the owned-host path — the one Eval uses — is affected. Classic optional-field-plus-explicit-list silent fallback.

The new test cannot see it: it stubs `connectOwnedRuntimeHost` and asserts the object it was just handed contains `livenessIntervalMs: 0`, which restates `hosted-execution.ts:43-44`. `connection.ts` already exposes `onLivenessProbe` for proving probes do or do not happen — `host-kernel.test.ts:1165` and `agent-graph-two-client-uds.test.ts:90` both use it.

packages/runtime-host/src/server/hosted-execution-tool-profile.ts
```diff
@@ -44,6 +47,9 @@ export function hostedExecutionRunProfile(
       toolNames: HEADLESS_CODING_V1_TOOL_NAMES,
       systemPrompt: HEADLESS_CODING_V1_SYSTEM_PROMPT,
       memoryExtraction: false,
+      streamIdleTimeoutMs: 0,
```
Astro-Han, line 50: **P1** — the description says one setting is frozen; this freezes three, and the two undici ones behave differently from the watchdog one.

`streamIdleTimeoutMs: 0` is fine: the harness deadline takes over for the main stream, and `stream-watchdog.ts` handles 0 as a clean disable while keeping the 30s connect phase. But `providerHeadersTimeoutMs` / `providerBodyTimeoutMs` apply to the entire transport, and `StreamWatchdog` only exists in `stream-watchdog.ts` and `ai-sdk-backend.ts` — not in `ai-sdk-compaction.ts`, whose `history_compact` calls at :449 and :1523 were bounded by undici's 300s default before this and are now bounded by nothing but the task deadline. Compaction on long tasks is routine, so one stalled compaction silently eats the whole task budget and the task scores 0.

Suggest generous finite values here (900s body, 120s headers) rather than 0. The main stream's idle behavior is controlled separately by `streamIdleTimeoutMs`, so nothing is lost.

Minor: this file is now `*-tool-profile.ts` while carrying network timeouts. The interface name `HostedExecutionRunProfile` is already right; the filename could follow.

packages/runtime-host/src/server/execution-model-composition.ts
```diff
@@ -100,6 +106,12 @@ export async function createHostAiSdkBackend(input: HostAiSdkBackendInput): Prom
   );
   const transport = createFetchTransport(
     toRuntimePolicyProxy(target.networkProxy, target.proxySecret),
+    runProfile
+      ? {
+          headersTimeoutMs: runProfile.providerHeadersTimeoutMs,
```
Astro-Han, line 111: **P2** — these two lines and :344 are the core of the PR and neither has an assertion. `execution-model-composition.test.ts` already injects `createFetchTransport` (:341, :366, :2084) and already has a headless profile fixture (:2543), but nothing checks the second argument, and nothing checks that the backend receives `streamIdleTimeoutMs`. Deleting :344 or passing `undefined` here leaves the suite green.

Worth a negative assertion too — that a non-headless profile still gets the defaults — since `hosted-execution.ts:43` gates on `toolProfile === 'headless-coding-v1'` and nothing currently pins that.

packages/runtime-host/src/__tests__/hosted-execution-client.test.ts
```diff
@@ -87,7 +98,7 @@ test('post-connect cancellation reports the owned Host settlement outcome', asyn
       },
     });
 
-    assert.equal(result.failureReason, failureReason);
+    assert.equal(result.failureReason, 'Hosted execution was cancelled');
```
Astro-Han, line 105: **P2** — this rewrite dropped the only coverage of `hosted-execution.ts:126`. The previous case asserted `clean === false` produced "Runtime Host did not exit cleanly"; both branches now assert "Hosted execution was cancelled", and a repo-wide grep finds no other test with that string. Mutating :125-126 to `return projection;` leaves the file green.

The behavior change itself is defensible — preferring the primary cause is more diagnosable — but it is unrelated to deferring the idle timeout, and `maka-subject.ts:162-169` lists that exact string in `SAFE_RUNTIME_FAILURE_REASONS`, so the eval-visible surface did move. A settled-projection case with `clean === false` alongside the new one would restore it.

packages/runtime/src/network/__tests__/scoped-fetch-transport.test.ts
```diff
@@ -11,9 +11,39 @@ import {
 } from '../../connection-effect-fetch.js';
 import { runConnectionModelDiscoveryEffect } from '../../model-fetcher.js';
 import { runConnectionTestEffect, testConnection } from '../../test-connection.js';
-import { createConnectionEffectFetchTransport } from '../scoped-fetch-transport.js';
+import {
+  createConnectionEffectFetchTransport,
```
Astro-Han, line 15: **P2** — this passes `null` for the proxy, so `useProxy` is always false and the `proxyRequestDispatcher` added at :82-84 and :90 never executes. Reverting :90 to the bare `proxyDispatcher` — which would mean the timeouts silently do not apply on the proxy path — leaves this test green. Eval runs through `toRuntimePolicyProxy(...)`, so the proxy path is the production path here.

The server also delays before writing headers, so only `headersTimeout` is exercised; dropping the `bodyTimeoutMs` branch at `scoped-fetch-transport.ts:63-65` also stays green. A case that sends headers first and then stalls between chunks separates them. The implementation is correct — this is only about what the tests pin.

packages/eval/README.md
```diff
@@ -34,7 +34,13 @@ and exposes a foreground-only Bash schema without `run_in_background` or `pty`.
 routing remains authoritative: DeepSeek Responses exposes `apply_patch` instead of `Write` and
 `Edit`, and Runtime-owned `ArchiveRead` remains available for archived tool results. A real
 `hosted.execution.start` regression test pins SHA-256 hashes for the first main provider request's
-developer prompt and complete tool schema.
+developer prompt and complete tool schema. The profile disables Runtime's product stream-idle
+watchdog and Undici's provider headers/body inactivity timers after provider activity;
+benchmark-native subject timeouts remain the execution deadline, while the model stream connect
+timeout still rejects requests that never establish a response. Hosted Eval clients also use a
```
Astro-Han, line 40: **P2** — "disabled active liveness probe" is not true until the forwarding gap in `connect-or-spawn.ts` is closed; the probe still runs at 2s.

The sentence about benchmark-native timeouts is right for the main stream but understates the rest: with `providerBodyTimeoutMs: 0` the harness deadline also becomes the only bound on compaction calls, which no watchdog covers. Worth saying explicitly here, since this paragraph is where someone will look to understand what protects a hosted eval run.

Astro-Han (commented):
The diagnosis is right and I traced it rather than taking it on faith: `markActivity()` has exactly one caller, inside the `for await (const chunk of sdk.stream)` loop in `model-adapter.ts:339`, so the 120s really is inter-chunk silence and tool execution really is excluded. The abort path to `missing_attempt_usage` closes too — abort listener → `finalize('aborted')` → `resolveUsageBasis(undefined)` = `'missing'` → `usageMissingAttempts += 1` → the whole execution settles `indeterminate`. And deferring to the harness is the right instinct: any finite idle value here is a guess, while the benchmark-native deadline is the real authority, and that authority is real rather than promised — `relay_agent.py:139` catches the cancellation and reports `framework_timeout` with exit 124.

The `0` handling is also clean, and I checked it because it is the kind of thing that silently regresses: `stream-watchdog.ts:49` uses `??` rather than `||`, and `schedule()` does `if (delayMs <= 0) return`, so 0 disables rather than firing immediately. The connect phase keeps its 30s independently. Nothing leaks into product sessions — `hostedExecutionRunProfile(undefined)` returns undefined and both call sites guard on it. Threading this through `HostedExecutionRunProfile`, which already carries `memoryExtraction`, is extending the existing seam rather than building a parallel one.

**The description covers one of three commits, and the other two are where the problems are.** `331e43a15` and `c689c1e2f` are not mentioned, and the body describes freezing one setting when the profile freezes three.

**P1 — `c689c1e2f` is a no-op in production.** `connect-or-spawn.ts:46-47` declares `livenessIntervalMs` and `livenessTimeoutMs`, but the call to `connectResolvedRuntimeHost` at :186-199 lists fields explicitly and forwards only `connectTimeoutMs`, `handshakeTimeoutMs` and `electionDeadline`. Both new fields are dropped, so the probe still runs at the 2s default and the CPU-starvation misjudgment is unchanged. `connection.ts:1081` and `:1102` spread `...input`, so only the owned-host path — the one Eval uses — is broken. The new test cannot catch it: it stubs `connectOwnedRuntimeHost` and asserts that the object just passed in contains `livenessIntervalMs: 0`, which restates the three lines above it. `connection.ts` already exposes `onLivenessProbe` for exactly this, and `host-kernel.test.ts:1165` and `agent-graph-two-client-uds.test.ts:90` already use it.

**P1 — freezing the two undici timeouts removes the last backstop from calls the watchdog never covered.** `providerHeadersTimeoutMs: 0` and `providerBodyTimeoutMs: 0` apply to the whole transport, but `StreamWatchdog` appears in only two files, and `ai-sdk-compaction.ts` is not one of them — its `history_compact` provider calls at :449 and :1523 were previously bounded by undici's 300s default and are now bounded by nothing except the task deadline. Long tasks trigger compaction routinely, so a stalled compaction quietly consumes the whole task budget and the task scores 0. Disabling is right for the main stream, where the harness takes over; it is wrong here, where nothing does.

**The deeper issue this leaves in place.** `summarizeModelCallCoverage` (`model-call-attempt.ts:372-391`) counts every attempt without filtering by status or reducing to the settled attempt per logical call — even though `settledAttempt` exists at :322 in the same file, commented as "the attempt that settled a logical call". So any provider request that errors and is then successfully retried still leaves a failed attempt in the ledger, `usageMissingAttempts > 0`, and the whole execution settles `indeterminate`. The idle watchdog is one trigger among several; a long DeepSeek run will hit 429s and 5xxs and reproduce the same symptom. That is not this PR's job to fix, but it should be stated in the description so the next occurrence is not read as a regression of this fix.

**Minimal path.** Forward the two liveness fields at `connect-or-spawn.ts:197` and move the assertion down a layer, or write the `onLivenessProbe` test that proves no probe fires at interval 0. Give `providerBodyTimeoutMs` and `providerHeadersTimeoutMs` generous finite values instead of 0 — the main stream's idle behavior is controlled independently by `streamIdleTimeoutMs: 0`, so the two do not conflict. Assert the wiring on the `createFetchTransport` seam the composition tests already inject. Restore the "did not exit cleanly" assertion alongside the new cancelled case. Then correct the description and the README.

**Cleanest path.** Split it. This PR keeps `streamIdleTimeoutMs: 0` plus those fixes — small, clearly bounded, revertible on its own. A second PR makes `incompleteUsageReason` settle per logical call using the `groupModelCallAttempts` and `settledAttempt` that already exist, so a retried failure no longer voids a run. That is a usage-contract change and wants independent human review under AGENTS.md. Once it lands, `streamIdleTimeoutMs` can go back to a large finite value and the watchdog becomes a diagnostic signal rather than a fatal one — fewer special cases than where this PR leaves it.

On the validation list: the 900/900 is not in question, but those suites are largely insensitive to this change. Mutating `execution-model-composition.ts:344` away, or reverting the proxy dispatcher at `scoped-fetch-transport.ts:90`, leaves the tests green.

---

*Review assistance: Claude Code (Opus) ran parallel passes over architecture, correctness and test quality, compiled the two changed transport files standalone to exercise the watchdog and undici timeout semantics empirically, and drafted this comment. I verified independently the parts it turns on: the dropped liveness forwarding by reading `connect-or-spawn.ts` against `connection.ts`, the absence of `StreamWatchdog` from the compaction path, the three commits against the description, and the profile's three frozen timeouts. The mutation results and the `missing_attempt_usage` trace are AI output that I have not re-run myself.*

<details>
<summary>中文</summary>

诊断是对的，而且我是追过代码而不是照单全收：`markActivity()` 只有一个调用点，在 `model-adapter.ts:339` 的 `for await (const chunk of sdk.stream)` 循环里，所以 120s 确实是 chunk 之间的静默，工具执行确实不计入。到 `missing_attempt_usage` 的链条也闭合——abort listener → `finalize('aborted')` → `resolveUsageBasis(undefined)` 得 `'missing'` → `usageMissingAttempts += 1` → 整个 execution 判 `indeterminate`。把超时让渡给 harness 这个直觉也是对的：这里任何有限的 idle 值都是猜，而 benchmark 自己的截止时间才是真权威，并且这个权威是真实存在的，不是纸面承诺——`relay_agent.py:139` 捕获取消并以 exit 124 上报 `framework_timeout`。

`0` 的处理同样干净，我特意核了这一点，因为它正是那种会悄悄退化的地方：`stream-watchdog.ts:49` 用的是 `??` 而不是 `||`，`schedule()` 里 `if (delayMs <= 0) return`，所以 0 是禁用而不是立刻触发。connect 阶段的 30s 独立存活。也没有泄漏到产品会话——`hostedExecutionRunProfile(undefined)` 返回 undefined，两个调用点都做了守卫。把这些字段挂到本来就承载 `memoryExtraction` 的 `HostedExecutionRunProfile` 上，是扩展既有 seam，不是另起一条路径。

**描述覆盖了三个提交里的一个，而问题都在另外两个里。** `331e43a15` 和 `c689c1e2f` 完全没提，而且 body 说冻结了一个设置，profile 里实际冻结了三个。

**P1 —— `c689c1e2f` 在生产里是 no-op。** `connect-or-spawn.ts:46-47` 声明了 `livenessIntervalMs` 和 `livenessTimeoutMs`，但 :186-199 调用 `connectResolvedRuntimeHost` 时是逐字段显式列举，只转发了 `connectTimeoutMs`、`handshakeTimeoutMs` 和 `electionDeadline`。两个新字段被丢掉，探针仍以 2s 默认间隔运行，CPU 饥饿误判一点没变。`connection.ts:1081` 和 `:1102` 用的是 `...input` 展开，所以只有 owned-host 这条 Eval 使用的路径断了。新测试抓不到它：它把 `connectOwnedRuntimeHost` 整个 stub 掉，然后断言刚传进去的对象里有 `livenessIntervalMs: 0`，等于把上面三行实现复述一遍。`connection.ts` 本来就为此暴露了 `onLivenessProbe`，`host-kernel.test.ts:1165` 与 `agent-graph-two-client-uds.test.ts:90` 已经在用。

**P1 —— 冻结那两个 undici 超时，会让 watchdog 从未覆盖的调用彻底失去兜底。** `providerHeadersTimeoutMs: 0` 与 `providerBodyTimeoutMs: 0` 作用在整个 transport 上，而 `StreamWatchdog` 只出现在两个文件里，`ai-sdk-compaction.ts` 不在其中——它在 :449 和 :1523 的 `history_compact` provider 调用，原本由 undici 默认的 300s 兜底，现在除了任务截止时间之外没有任何边界。长任务触发 compaction 是常态，于是一次卡住的 compaction 会悄悄吃掉整个任务预算，任务记 0 分。对主 stream 禁用是对的，那里有 harness 接管；对这里禁用是错的，因为没有任何东西接管。

**这个 PR 留下的更深一层问题。** `summarizeModelCallCoverage`（`model-call-attempt.ts:372-391`）对所有 attempt 计数，既不按 status 过滤，也不按 logical call 归约到 settled attempt——尽管同文件 :322 就有 `settledAttempt`，注释写着「结算了一次 logical call 的那个 attempt」。于是任何一次出错、随后重试成功的 provider 请求，仍会在 ledger 里留下一条失败 attempt，`usageMissingAttempts > 0`，整个 execution 判 `indeterminate`。idle watchdog 只是其中一个触发源；DeepSeek 大规模跑必然遇到 429 和 5xx，同样的症状会照样出现。这不是本 PR 该修的，但应该写进描述，免得下次出现时被读成「修好了又复发」。

**最小方案。** 在 `connect-or-spawn.ts:197` 补上两个 liveness 字段的转发，并把断言下沉一层，或者用 `onLivenessProbe` 写一条「interval 为 0 时不再发探针」的测试。把 `providerBodyTimeoutMs` 和 `providerHeadersTimeoutMs` 从 0 改成宽松的有限值——主 stream 的 idle 行为由 `streamIdleTimeoutMs: 0` 独立控制，两者不冲突。在 composition 测试已有的 `createFetchTransport` 注入缝上断言接线。把「did not exit cleanly」的断言与新的 cancelled 用例并存恢复。然后修正描述与 README。

**最彻底方案。** 拆开。本 PR 保留 `streamIdleTimeoutMs: 0` 加上面那些修复——范围小、边界清楚、可独立 revert。另开一个 PR 让 `incompleteUsageReason` 按 logical call 结算，复用已经存在的 `groupModelCallAttempts` 与 `settledAttempt`，这样被重试覆盖的失败不再让整跑作废。那是 usage 契约变更，按 AGENTS.md 需要独立人审。它落地之后，`streamIdleTimeoutMs` 甚至可以退回成一个较大的有限值，watchdog 重新成为诊断信号而不是致命信号——那才是奥卡姆意义上更少的特例。

关于 Validation 清单：900/900 本身没有问题，但那些用例对本次改动基本不敏感。把 `execution-model-composition.ts:344` 变异掉，或者把 `scoped-fetch-transport.ts:90` 的 proxy dispatcher 还原，测试都是全绿。

---

*评审协助说明：Claude Code (Opus) 并行跑了架构、正确性与测试质量三轮审查，把改动的两个 transport 文件单独编译出来实测 watchdog 与 undici 超时语义，并起草了本条评论。结论所依赖的部分我独立核实过：对照 `connection.ts` 读 `connect-or-spawn.ts` 确认 liveness 转发被丢弃、compaction 路径确实没有 `StreamWatchdog`、三个提交与描述的对照，以及 profile 里冻结的三个超时。变异测试结果与 `missing_attempt_usage` 链路追踪属于 AI 输出，我没有自己重跑。*

</details>

packages/runtime-host/src/client/launcher.ts
```diff
@@ -86,19 +89,97 @@ function spawnCandidate(input: DetachedCandidateInput, detached: boolean) {
...
+    closeSync(descriptor);
+  };
+  return {
+    descriptor,
+    finish: (code, signal, errorCode) => {
+      if (closed) return;
+      const marker = errorCode
+        ? `MAKA_RUNTIME_HOST_EXIT_V1 code=none signal=none error=${errorCode}\n`
+        : `MAKA_RUNTIME_HOST_EXIT_V1 code=${code ?? 'none'} signal=${signal ?? 'none'}\n`;
+      writeSync(descriptor, marker);
```
Astro-Han, line 171: **P2** — `finish` runs from `child.once('close')` and `child.once('error')`, and `writeSync` here is unguarded. On `ENOSPC` or a closed descriptor it throws synchronously out of an EventEmitter handler, which becomes an `uncaughtException` and takes down the whole harness process rather than losing one diagnostic line.

That inverts the PR's own goal: the disk filling up should degrade diagnostics, not kill the run that was producing them. `writeDiagnostic` at :179 has the same exposure. A `try { ... } catch { close(); }` around both is enough — dropping the marker and closing the descriptor is the correct degraded behaviour.

packages/runtime-host/src/client/launcher.ts
```diff
@@ -86,19 +89,97 @@ function spawnCandidate(input: DetachedCandidateInput, detached: boolean) {
...
+      close();
+    },
+    writeDiagnostic: (value) => {
+      if (closed) return;
+      const line = Buffer.from(value.replace(/[\r\n]+/gu, ' '))
+        .subarray(0, 1024)
```
Astro-Han, line 177: **P3** — `Buffer.from(value).subarray(0, 1024).toString()` cuts at a byte offset, so a multi-byte character straddling 1024 decodes to `U+FFFD`. Cosmetic, but it lands in a log meant for reading.

`harbor-maka-subject.ts:66` in this same PR already does the right thing for `failureReason`: `new TextDecoder().decode(Buffer.from(value).subarray(0, 768))`. Worth matching so both truncation points behave the same.

packages/runtime-host/src/client/connect-or-spawn.ts
```diff
@@ -241,6 +246,9 @@ export async function connectOrSpawnRuntimeHostWithDependencies(
           expectedRootId: capability.rootId,
           entrypoint: input.candidateEntrypoint,
           initialConnectionTimeoutMs: Math.ceil(remaining),
+          ...(input.candidateStderrPath === undefined
+            ? {}
+            : { stderrPath: input.candidateStderrPath }),
```
Astro-Han, line 251: **P2** — `openCandidateStderr` throws when the target can't be opened or isn't a regular file, and that throw propagates out of `dependencies.launchCandidate` into the `catch {}` at :260, which treats it as ordinary election evidence and continues.

It isn't ordinary, though: the condition is stable, so every retry in the loop throws at the same point, no candidate ever spawns, and the run ends on a generic election timeout with no indication that a log path was the cause. A read-only or symlinked `rootPath` turns "collect better diagnostics" into "silently cannot start a Host".

Opening the log is auxiliary to spawning. Suggest failing soft — skip the redirect and spawn with `'ignore'` — or hoisting the open above the retry loop so the failure surfaces once, with its reason.

packages/runtime-host/src/client/hosted-execution.ts
```diff
@@ -98,7 +112,8 @@ export async function runHostedExecutionWithDependencies(
       }
     }
     projection = await executeHostedExecution(connected.connection, input.execution, input.signal);
-  } catch {
+  } catch (error) {
+    connected.host.recordDiagnostic?.(runtimeHostClientErrorMarker(error));
```
Astro-Han, line 116: **P3** — `recordDiagnostic` only exists on an owned candidate this client launched. When `connectOwnedRuntimeHost` attaches to an `existing_host`, the optional call is `undefined` and the client-side error is dropped with no trace — and an already-running Host is a plausible setting for the failures this marker is meant to explain.

Not necessarily wrong to skip it (there's no descriptor to write to), but the silence is indistinguishable from "no error occurred". Even a one-line comment saying the existing-host path is intentionally uncovered would keep the next reader from assuming coverage.

packages/runtime-host/src/client/hosted-execution.ts
```diff
@@ -114,9 +129,48 @@ export async function runHostedExecutionWithDependencies(
...
+    readonly dispatch?: unknown;
+    readonly reason?: unknown;
+    readonly retryable?: unknown;
+  };
+  const cause = failure.cause instanceof Error ? failure.cause : undefined;
+  const code =
+    typeof (failure as NodeJS.ErrnoException).code === 'string'
+      ? (failure as NodeJS.ErrnoException).code
+      : null;
+  return `MAKA_RUNTIME_HOST_CLIENT_ERROR_V1 ${JSON.stringify({
```
Astro-Han, line 150: **P3** — `MAKA_RUNTIME_HOST_CLIENT_ERROR_V1` and `MAKA_RUNTIME_HOST_EXIT_V1` are written but never parsed; grep finds readers only in `owned-candidate.test.ts`. The `_V1` suffix plus a fixed JSON shape reads as a versioned wire contract, yet nothing pins the shape, so a field can be renamed and every test stays green.

Either direction is fine — a small reader in `@maka/eval` that turns these into attempt attribution, or dropping the `_V1` and treating them as human-readable log lines. Right now it's shaped like a contract without being one. (I withdraw my earlier framing that this would be expensive to change later: with no consumer, it's cheap.)

packages/runtime-host/src/server/hosted-execution-tool-profile.ts
```diff
@@ -44,6 +47,9 @@ export function hostedExecutionRunProfile(
...
+      providerHeadersTimeoutMs: 0,
```
Astro-Han, line 51: **P2, downgraded from P1** — three timeouts are frozen here while the description argues only for stream idle. `0` is a clean disable in every one of the three, and `harness-executor.ts:353` shows Harbor still holds an attempt-level budget, so a wedged provider connection is reclaimed. Nothing is broken; I had this too high last round.

What remains is attribution: with the undici `headersTimeout` / `bodyTimeout` backstops off, a provider that accepts the connection and never responds surfaces as a generic attempt timeout rather than a provider fault. That is the exact distinction the rest of this PR exists to preserve. A sentence in the description covering why all three are disabled together, and what the intended attribution is when the provider hangs, is all I'd ask for.

Astro-Han (dismissed):
Approving. The P1 from the last round is fixed, and the four new commits go further than I asked — nothing left here blocks the merge. Everything below is P2/P3 and can land as follow-ups.

**The part worth naming.** `maka-artifacts.ts:55,69` on `main` already collects `runtime-host-candidate.log`, while `launcher.ts` has always spawned with `stdio: 'ignore'`. Nothing has ever written that file. `maka-artifacts.test.ts:30` calls `writeFile` to create it and then asserts it was copied, so the contract has been green while being empty. This PR supplies the missing producer — so the stderr work in `launcher.ts` is not scope creep, it closes a dangling contract. Same shape I flagged on #2963 and #2965: a test manufacturing the thing under test.

While you're there, `maka-artifacts.test.ts:30` should now assert a real candidate log rather than a synthetic one (P2). That synthetic fixture is exactly what let this stay unnoticed.

**Fixed, verified.** `livenessIntervalMs` / `livenessTimeoutMs` now reach `connectResolvedRuntimeHost` (`connect-or-spawn.ts:200-201`), pinned by `hosted-execution-client.test.ts:66-67`. The `O_NOFOLLOW` + `0o600` + `isFile()` guard on the log target is careful work.

**Two robustness gaps, both one-liners.** `writeSync` inside the `close` handler and the `stderrPath` open failure being swallowed — details inline. Both share a shape: the path added to make failures attributable is itself the path that loses attribution when it fails.

**On the timeouts.** I graded this P1 last round; that was too high and I'm withdrawing it. `harness-executor.ts:353` shows Harbor holds an attempt-level budget, so a wedged connection is still reclaimed. The real cost is that it comes back as a generic timeout instead of a provider error, which works against what this PR is for. Worth a line in the description rather than a code change.

CI: `test_workspaces` was red on an unrelated flake in `shell-run-manager.test.ts` (this PR touches no `packages/runtime` code; the case passes locally in 549ms). Filed separately and re-run.

---

*Review assistance: Claude Code (Opus). I verified the dangling artifact contract against `origin/main`, the liveness forwarding, the attempt-level budget, and the flake's independence from this diff myself.*

<details>
<summary>中文</summary>

Approve。上一轮的 P1 已修，新推的四个提交比我要求的走得更远，没有阻塞合并的东西了。下面全是 P2/P3，可以后续处理。

**值得单独点出来的一点。** `main` 上 `maka-artifacts.ts:55,69` 早就在采集 `runtime-host-candidate.log`，而 `launcher.ts` 一直是 `stdio: 'ignore'`，从来没人写过这个文件。`maka-artifacts.test.ts:30` 用 `writeFile` 自己造出它再断言它被拷走，于是这个契约在空着的情况下一直是绿的。这个 PR 补上了缺失的生产端——所以 `launcher.ts` 里的 stderr 工作不是 scope creep，是闭合一个悬空契约。跟我在 #2963 和 #2965 指出的是同一个形状：测试自己造出被测对象。

顺手把 `maka-artifacts.test.ts:30` 改成断言真实产出的候选日志（P2）。正是那个自造 fixture 让这件事一直没被发现。

**已修，已核。** `livenessIntervalMs` / `livenessTimeoutMs` 现在能到达 `connectResolvedRuntimeHost`（`connect-or-spawn.ts:200-201`），由 `hosted-execution-client.test.ts:66-67` 钉住。日志目标上那套 `O_NOFOLLOW` + `0o600` + `isFile()` 检查做得很谨慎。

**两处健壮性缺口，都是一行的事。** `close` 回调里的 `writeSync`，以及 `stderrPath` 打开失败被吞掉——细节在行内。两者形状相同：为了让失败可归因而新增的路径，在它自己失败时反而抹掉归因。

**关于超时。** 上一轮我判了 P1，判高了，撤回。`harness-executor.ts:353` 表明 Harbor 持有 attempt 级预算，卡死的连接最终收得回来。真实代价是它以通用超时而非 provider 错误的形式回来，这与本 PR 的目的相悖。写进描述即可，不需要改代码。

CI：`test_workspaces` 之前红在 `shell-run-manager.test.ts` 的一个无关 flake 上（本 PR 一行 `packages/runtime` 都没碰，该用例本地单跑 549ms 通过）。已单独开 issue 并重跑。

---

*评审协助说明：Claude Code (Opus)。悬空的 artifact 契约（对照 `origin/main`）、liveness 转发、attempt 级预算，以及该 flake 与本 diff 无关这几点，都是我自己核实的。*

</details>

hqhq1025:
Real-machine follow-up is complete on the combined code line.

- Final selected coverage: **712/712 cells with usage + physical trajectory**.
- Repaired cells: **23/23**.
- Manifest: `/mnt/deepswe/eval-runs/deepseek-v4-flash-final-trajectory-overlay-8a1ab4635.json`.
- `missing=[]`, `artifactFailures=[]`.
- Every replacement was checked against its attempt `collected-artifact` bytes and SHA-256.
- Final code-line Eval tests: Node **36/36**, Python **28/28**.

Two replacement scores changed under fresh model sampling: Claude Code `write-compressor` 0→1 and ZCode `raman-fitting` 1→0. No cherry-picking was used; the final overlay takes the valid replacement attempt that supplies the corresponding usage and trajectory.

hqhq1025:
Final CI follow-up:

- `test_workspaces`: PASS
- `test_runtime_host`: PASS
- `typecheck`: PASS
- aggregate `test`: PASS
- Windows baseline/recovery: PASS

Two pre-existing CI flakes were made deterministic while validating this PR:

1. the Eval trajectory truncation regression now uses a test-only 64 KiB boundary instead of writing 65 MiB while preserving the production 64 MiB limit;
2. the Runtime PTY redaction test no longer depends on unrelated SQLite persistence, and the owned-Host exit test separates its 6s cold-start election budget from the unchanged 500ms post-connect exit assertion.

The authoritative real-machine result remains 712/712 selected cells with usage and physical trajectories, with no missing or artifact-failure entries.

Astro-Han:
Ran an independent adversarial review of this PR, since it is the last code blocker for the paired Maka vs DeepSeek Harness cohort and we want that run started. Two reviewers ran in parallel in isolated worktrees at `25baa1c42` — Codex and a Claude Code subagent — with no visibility into each other; a third was cancelled for time.

Grouping by what gates the run rather than leading with severity, since the goal is to get the cohort moving — but every finding carries its severity so nothing gets lost in the reordering.

I re-verified the findings marked **[reproduced]** myself; the rest are relayed claims with their evidence quoted, and the one I could not verify is labelled as such.

## P1 — blocks the run: the idle-timeout fix only covers half the deadline

`hosted-execution-tool-profile.ts:50-52` sets `streamIdleTimeoutMs: 0`, but `connectTimeoutMs` is untouched and still resolves to `DEFAULT_STREAM_CONNECT_TIMEOUT_MS = 30_000` (`stream-watchdog.ts:48`, scheduled at `:62` from `start()`).

The two arms are not measuring the same thing:

- external — `createExternalProviderDispatcher` defaults `headersTimeout`/`bodyTimeout` to `0` (`external-provider-dispatcher.ts:16-17`). Response headers arrive promptly, and the body then streams with no deadline.
- maka — the watchdog's connect phase runs from `start()` until the first `markActivity()`, i.e. until the first model event, and that is still capped at 30s.

DeepSeek V4 Flash at `thinking=max` routinely exceeds 30 seconds before its first streamed event. The Maka arm can therefore still be aborted by its own deadline ahead of the benchmark-native one — the exact failure this PR exists to remove — while the external arm waits indefinitely. In a paired cohort that shifts one arm only, which is the one thing the pairing cannot absorb.

This is the finding I would want closed before we start, since it changes the Maka column directly.

Suggested check: one local provider fixture that accepts the connection, delays the first SSE event past 31s, then completes; run it through `headless-coding-v1` and through the external wrapper under the same native timeout and confirm both classify identically. No current test sends the same delayed-first-event request through both paths.

## P1 ×2 — affects published numbers, but does not stop us starting

Both of these hit reported token and cost figures on recovered cells. Scores are unaffected, so they can be fixed while the cohort runs — but they are P1 because they must be fixed before we publish a cost comparison, not merely tracked.

**`recoverMakaRuntimeUsage` reports `partial` and `unpriced` attempts as complete [reproduced].** `maka-artifacts.ts:66` derives completeness from `missing` alone:

```ts
const missingAttempts = attempts.filter((record) => record.usageBasis === 'missing').length;
const usageComplete = missingAttempts === 0;
```

and labels both bases from that one boolean while summing `record.costUsd ?? 0`. An attempt with `costBasis: 'unpriced'` carries no `costUsd`, so it contributes 0 and is still reported complete. Reproduced against `dist/maka-artifacts.js` with a single reported/unpriced attempt:

```json
{
  "costUsd": 0,
  "artifact": { "usageComplete": true, "tokenBasis": "complete", "costBasis": "complete" }
}
```

`usageBasis: 'partial'` behaves the same way. These values flow into the published `value.usage` / `value.costUsd` at `harness-executor.ts:241-244`.

`@maka/core` already models this: `summarizeModelCallCoverage` (`model-call-attempt.ts:385-390`) separates `priced`/`unpriced` and `reported`/`partial`, and `sumModelCallCostUsd` exists precisely because — its own doc comment — "a bare number cannot express 'plus an unknown amount from unpriced calls'". Eval already depends on `@maka/core`, so this looks like a call-site fix rather than a patched reducer here.

**External metering checkpoint can claim completeness with a request in flight** *(not independently verified — relaying Codex's reasoning)*. `admittedRequests` and `usageRequests` are both incremented in the `finally` after the upstream body finishes (`harbor-external-subject.ts:801-806`), so a request that is admitted and then truncated by a kill increments neither, leaving `usageComplete` (`:722`) true while describing only the earlier requests. `recoverExternalMetering` accepts it on that boolean (`external-subject.ts:227`). If it holds, the checkpoint needs `inFlightRequests` persisted at request start — recovery cannot repair a checkpoint that never recorded it.

## P2/P3 — does not affect the run, filing so they are not lost

- **P2 — the usage-recovery wiring has no test covering it [reproduced].** Mutating `harness-executor.ts:231` from `if (value!.usage === null)` to `if (false as boolean)` — disabling the entire recovery path this PR adds — leaves the Eval suite at 36 pass / 0 fail. Deleting the JSONL bound check at `harbor-external-subject.ts:869` and deleting `await proxy.settle()` at `:105` also both survive the full suite.
- **P2 — `provider-events.jsonl` drops records silently at the 64 MiB bound.** `if (closed || persistedBytes + line.byteLength > limitBytes) return;` keeps no accounting, unlike stdout/stderr which report `observedBytes`/`persistedBytes`/`truncatedBytes`/`observedSha256`. It is also non-monotonic — an oversized chunk is dropped while a later short `provider_response_end` still writes, so the trace reads as complete with a hole in it. This is the artifact of record when a CLI times out without flushing stdout.
- **P2 — Eval reads Storage's private SQLite tables.** `maka-artifacts.ts:211-224` selects from `usage_model_call_attempts` / `usage_llm_calls` and reads record field names directly, against AGENTS.md's "may use the Runtime Host client and protocol, but must not depend on Runtime implementation". It also fails silently: a throw in `readUsageRecords` is swallowed by the outer `catch { return undefined }`, so a Storage migration would zero out Maka usage recovery with CI green. (Clearing one thing: the two tables do not double-count — `usage_llm_calls` now only receives auxiliary calls from `execution-model-authority.ts:511,528` since #1679.)
- **P3 — `persistSnapshot` writes non-atomically** (`writeFile` truncate + write) to the file the recovery path depends on; a kill between the two leaves 0 bytes. `write tmp + rename` closes it.
- **P3** — `hosted-execution.ts:44-50` hardcodes `toolProfile === 'headless-coding-v1'` a second time for liveness and stderr path while the profile table lives in `server/hosted-execution-tool-profile.ts`; those will drift. `launcher.ts:118-121` can split a multi-byte UTF-8 sequence and emit U+FFFD.

## P0 / P2 — carried over from #2963, does not hold up this run

**P0** — `eight-arm.json:292` changes the Pi arm's prompt to `"Task instruction:\n{{task.input}}"` — the same change we raised as P0 on #2963. Every other arm passes `{{task.input}}` verbatim, so this alters what the model reads rather than how the CLI parses it, and the README wording added here ("terminate option parsing or bind the prompt inside one option value") does not describe what that arm does. **P2** — `kimi-code` at `:237-238` is still `"--prompt", "{{task.input}}"`, the shape this PR declares unsafe and fixes for zcode one arm above; `lifecycle-boundaries.test.ts` freezes each arm's literal separately and skips kimi-code.

Both are P-graded against the eight-arm cohort, whose comparability they do affect. Neither touches the Maka vs DSH pair, so neither holds up that run.

## Checked and clear

Recording these so nobody re-digs them:

- **No product blast radius.** `hostedExecutionRunProfile(undefined)` returns `undefined` (`:44`), so both spreads in `execution-model-composition.ts` collapse for non-headless sessions, and `createProxiedFetchTransport` returns the dispatcher untouched when `options = {}`. `toolProfile` is only ever set from `maka-subject.ts:36`. Interactive and desktop sessions are unchanged.
- `idleTimeoutMs: 0` disables rather than fires immediately — `stream-watchdog.ts:92` is `if (delayMs <= 0) return`, pre-existing and unmodified.
- Moving `collectedArtifactInventory` into `runHarnessAttempt` is not a regression; `readVerification` was already gated on `finalizationConfirmed`.

Happy to take the connect-timeout fix directly if that gets the cohort started sooner — say the word and it stays yours otherwise.

---

*Review conducted with Codex and Claude Code as independent adversarial reviewers; findings marked **[reproduced]** were re-verified by hand against `25baa1c42`, one is explicitly labelled unverified, and the remainder are model-reported claims with evidence quoted inline. Human review of the final diff and the merge decision are still required.*

packages/eval/src/harbor-external-subject.ts
```diff
@@ -607,15 +709,43 @@ async function startMeteringProxy(
...
+  const events = await createBoundedJsonlWriter(eventsPath, PERSISTED_STREAM_LIMIT_BYTES);
+  let snapshotWrite = Promise.resolve();
+  const persistSnapshot = () => {
+    const snapshot = `${JSON.stringify({
+      schemaVersion: 1,
+      profile,
+      usage: measured ? { ...total, totalTokens: total.inputTokens + total.outputTokens } : null,
+      costUsd: measured ? estimateCost(total) : null,
+      requests,
+      admittedRequests,
```
Astro-Han, line 721: **P1 — Preserve current main's authenticated metering checkpoint.** Current `main` writes a bounded, `O_NOFOLLOW`, atomic HMAC-authenticated checkpoint and records in-flight admissions; this plain `writeFile` snapshot can be forged by the external subject, torn on kill, and omits unfinished provider requests. Rebase and keep `metering-checkpoint.ts`, `meteringSecret`, `inFlightRequests`, and atomic write rather than resolving this conflict in favor of the PR version.

packages/eval/src/maka-artifacts.ts
```diff
@@ -21,6 +23,79 @@ export interface MakaRuntimeArtifactManifest {
...
+  let database: DatabaseSync;
+  try {
+    database = new DatabaseSync(path, { readOnly: true });
+  } catch {
+    return undefined;
+  }
+  try {
+    const attempts = readUsageRecords(database, 'usage_model_call_attempts');
+    const legacy = readUsageRecords(database, 'usage_llm_calls');
+    const missingAttempts = attempts.filter((record) => record.usageBasis === 'missing').length;
```
Astro-Han, line 51: **P1 — Do not report partial or unpriced usage as complete.** This counts only `usageBasis === 'missing'`; a `partial` attempt is therefore marked complete, and an `unpriced`/missing `costUsd` becomes zero while `costBasis` is also `complete`. That corrupts benchmark totals. Reuse the canonical usage/cost aggregation so partial remains lower-bound and unpriced remains unknown; do not collapse the two bases into one boolean.

Astro-Han (commented):
The intended Eval improvements are valuable: hosted timeout ownership, external stdout/stderr artifacts, late Maka usage recovery, provider traces, and liveness forwarding all close real observability gaps on the old base.

This head cannot be merged by resolving text conflicts mechanically. Current `main` now owns metering through a signed, bounded, atomic checkpoint with in-flight admission; this branch replaces it with an unauthenticated plain JSON file and duplicate price/profile tables. From first principles, metering and price calculation need one authority. Rebase first, preserve current-main `metering-checkpoint`, `meteringSecret`, `inFlightRequests`, shared profile registry, and `provider-metering` prices, then transplant only the remaining artifact/trace features. The new Maka recovery must also preserve partial/unpriced semantics and should consume a Runtime Host artifact/protocol instead of private Storage tables.

Reviewed with Codex using two independent reviewer agents; I verified the latest head, current-main merge conflicts and metering authority, usage aggregation, current DeepSeek Harness profile/prices, prior review, and live CI.

这次 Eval 改进的目标有价值：hosted timeout 所有权、外部 stdout/stderr artifact、迟到 Maka usage recovery、provider trace 和 liveness forwarding，都修复了旧 base 上真实的可观测性缺口。

但当前 head 不能靠机械解决文本冲突来合并。最新 main 已用签名、有界、原子 checkpoint 加 in-flight admission 统一管理 metering；该分支却用未认证的普通 JSON 和重复的价格/profile 表替换它。按第一性原理，metering 与价格计算应只有一个权威。请先 rebase，保留 current-main 的 `metering-checkpoint`、`meteringSecret`、`inFlightRequests`、共享 profile registry 和 `provider-metering` 价格，再迁移尚未被替代的 artifact/trace 功能。新的 Maka recovery 还必须保留 partial/unpriced 语义，并应消费 Runtime Host artifact/protocol，而不是直接读取 Storage 私有表。

本次由 Codex 配合两个独立 reviewer agent 审查；我核验了最新 head、current-main 合并冲突与 metering authority、usage aggregation、当前 DeepSeek Harness profile/价格、此前 review 和实时 CI。

</details>

jackwener (changes requested):
Automated review of exact head `25baa1c421b5dad01b188f64b5abc7bf4da08a42` against current `main@d62857a8357e9160926726a2a13096bc2dc2b91d`.

The original long-stream timeout and artifact/usage-loss problems remain relevant, and the hosted-execution timeout ownership, stderr producer, and artifact inventory are valuable. This exact revision, however, predates and conflicts with the authorities now on main.

**P1 — rebasing this implementation would regress current metering integrity and accounting.** This branch introduces `${profile}.metering.json` as an unsigned file in the subject-writable log directory, writes it non-atomically, carries its own profile/price logic, and later trusts recovered cost/usage. Current main has already centralized this through signed, bounded, atomic `metering-checkpoint` records with `meteringSecret`, in-flight request state, the shared external-profile registry, and `provider-metering` price derivation. The PR must not resolve its text conflicts by keeping both paths or by replacing the current one: there must be one metering and price authority.

**P1 — Maka recovery reads Storage implementation details instead of a Runtime Host contract.** `packages/eval/src/maka-artifacts.ts` opens/backups `runtime.sqlite` and derives model-call totals from private tables. That couples Eval recovery to a store schema and can report a different completeness/cost view from the Runtime Host's canonical ledger. Preserve current partial/unpriced semantics and expose/consume a Runtime Host artifact or protocol projection instead of teaching Eval the database.

Required conclusions:

1. Optimal for the actual problem: **not on the current architecture**; several fixes are sound, but the revision creates parallel accounting authorities.
2. Production code to delete: delete the duplicate metering schema, price tables, and direct SQLite recovery path during rebase.
3. Tests to delete/replace: replace synthetic checkpoint/SQLite fixtures that merely restate those private formats with tests through the current signed checkpoint and Runtime Host projection seams.
4. Deeper refactor: required before merge—rebase onto main, preserve the current metering authority, and transplant only the still-missing timeout/artifact/trace behavior.
5. Ready to merge: **no**; the PR is currently conflicting.
6. Residual risks/gaps: Eval scoring, retry, usage completeness, and cost are protected contract behavior requiring independent human review, exact-head CI, and representative real-machine timeout evidence after the rework.

Astro-Han:

- `packages/runtime-host/src/__tests__/fixtures/owned-candidate-exit.ts`
- `packages/runtime-host/src/__tests__/owned-candidate.test.ts`
- `packages/runtime-host/src/client/connect-or-spawn.ts`
- `…(more)`

```
# resolve, then
git push --force-with-lease
```

---

PR #2963, opened 2026-08-13
fix(eval): harden external prompt and admission handling

hqhq1025:
Fixes two Eval-side attribution failures found while completing the seven-arm Terminal-Bench matrix.

- escapes dash-leading task instructions for positional external CLIs
- passes ZCode prompts as one `--prompt=` argument and keeps its cwd task-derived
- prefixes Pi input because Pi 0.84.1 treats dash-leading positional messages as options even after `--`
- treats a successful 2xx response carrying a model request as inference admission when an external CLI uses an unrecognized provider event envelope
- keeps non-2xx provider and transport failures classified as infrastructure failures

Real-host evidence:
- `pytorch-model-recovery` previously failed before model admission across Codex, Claude Code, Reasonix, ZCode, and Pi because its instruction begins with `- `
- corrected sparse reruns entered model execution and completed on Codex, Claude Code, Reasonix, and ZCode
- Reasonix `adaptive-rejection-sampler` and `gpt2-codegolf` actually consumed model output but were misreported as pre-admission because the response envelope was not recognized
- Reasonix client reasoning limits and task-native timeout failures remain third-party agent outcomes; this PR does not change their limits, prompts, effort, or timeouts

Validation:
- Eval Node 31/31
- relay contract 10/10
- relay lifecycle 12/12
- egress filter 3/3
- run-trial policy 2/2
- relay artifact 1/1
- `git diff --check`

### DIFF
--- a/packages/eval/experiments/terminal-bench-2.1-deepseek-v4-flash-eight-arm.json
+++ b/packages/eval/experiments/terminal-bench-2.1-deepseek-v4-flash-eight-arm.json
@@ -127,6 +127,7 @@
           "--dangerously-bypass-approvals-and-sandbox",
           "--model",
           "deepseek-v4-flash",
+          "--",
           "{{task.input}}"
         ]
       }
@@ -158,6 +159,7 @@
           "--dangerously-skip-permissions",
           "--model",
           "deepseek-v4-flash",
+          "--",
           "{{task.input}}"
         ]
       }
@@ -183,6 +185,7 @@
           "max",
           "--output-format",
           "stream-json",
+          "--",
           "{{task.input}}"
         ]
       }
@@ -254,8 +257,7 @@
           "{{task.cwd}}",
           "--disallowedTools",
           "WebSearch,WebFetch,FetchURL",
-          "--prompt",
-          "{{task.input}}"
+          "--prompt={{task.input}}"
         ]
       }
     },
@@ -287,7 +289,7 @@
           "deepseek-v4-flash",
           "--thinking",
           "max",
-          "{{task.input}}"
+          "Task instruction:\n{{task.input}}"
         ]
       }
     }
--- a/packages/eval/src/__tests__/lifecycle-boundaries.test.ts
+++ b/packages/eval/src/__tests__/lifecycle-boundaries.test.ts
@@ -79,7 +79,7 @@ while (true) {
 }
 const trialPath = new URL('./' + config.trial_name + '/', new URL('file://' + config.trials_dir + '/'));
 await mkdir(trialPath, { recursive: true });
-const agentArtifacts = new URL('artifacts/logs/agent/', trialPath);
+const agentArtifacts = new URL('agent/', trialPath);
 await mkdir(agentArtifacts, { recursive: true });
 await writeFile(new URL('opencode.provider-usage.json', agentArtifacts), '{"usageRequests":1}\\n');
 await writeFile(new URL('result.json', trialPath), JSON.stringify({ verifier_result: { rewards: { reward: 1 } } }));
@@ -138,12 +138,11 @@ socket.end();
     assert.deepEqual(
       result?.artifacts.find(
         ({ kind, path }) =>
-          kind === 'collected-artifact' &&
-          path === 'artifacts/logs/agent/opencode.provider-usage.json',
+          kind === 'collected-artifact' && path === 'agent/opencode.provider-usage.json',
       ),
       {
         kind: 'collected-artifact',
-        path: 'artifacts/logs/agent/opencode.provider-usage.json',
+        path: 'agent/opencode.provider-usage.json',
         bytes: 20,
         sha256: 'sha256:8f625fe55aa6336fccf93df0b9431dfac84d7d6fe772a745b56c9a284364310a',
       },
--- a/packages/eval/src/__tests__/lifecycle-boundaries.test.ts
+++ b/packages/eval/src/__tests__/lifecycle-boundaries.test.ts
@@ -79,6 +79,9 @@ while (true) {
 }
 const trialPath = new URL('./' + config.trial_name + '/', new URL('file://' + config.trials_dir + '/'));
 await mkdir(trialPath, { recursive: true });
+const agentArtifacts = new URL('artifacts/logs/agent/', trialPath);
+await mkdir(agentArtifacts, { recursive: true });
+await writeFile(new URL('opencode.provider-usage.json', agentArtifacts), '{"usageRequests":1}\\n');
 await writeFile(new URL('result.json', trialPath), JSON.stringify({ verifier_result: { rewards: { reward: 1 } } }));
 socket.end();
 `,
@@ -130,7 +133,21 @@ socket.end();
     assert.doesNotThrow(() => process.kill(pid, 0));
     await writeFile(release, '');
     const results = await running;
-    assert.equal(results.get('task::1::external')?.result.status, 'completed');
+    const result = results.get('task::1::external')?.result;
+    assert.equal(result?.status, 'completed');
+    assert.deepEqual(
+      result?.artifacts.find(
+        ({ kind, path }) =>
+          kind === 'collected-artifact' &&
+          path === 'artifacts/logs/agent/opencode.provider-usage.json',
+      ),
+      {
+        kind: 'collected-artifact',
+        path: 'artifacts/logs/agent/opencode.provider-usage.json',
+        bytes: 20,
+        sha256: 'sha256:8f625fe55aa6336fccf93df0b9431dfac84d7d6fe772a745b56c9a284364310a',
+      },
+    );
   } finally {
     await writeFile(release, '').catch(() => undefined);
     globalThis.setTimeout = nativeSetTimeout;
--- a/packages/eval/src/__tests__/lifecycle-boundaries.test.ts
+++ b/packages/eval/src/__tests__/lifecycle-boundaries.test.ts
@@ -649,6 +649,9 @@ test('eight-arm spec and wrappers freeze the working provider contracts', async 
   const maka = spec.subjects.find(({ id }) => id === 'maka')!;
   const codex = spec.subjects.find(({ id }) => id === 'codex')!;
   const claude = spec.subjects.find(({ id }) => id === 'claude-code')!;
+  const reasonix = spec.subjects.find(({ id }) => id === 'reasonix')!;
+  const zcode = spec.subjects.find(({ id }) => id === 'zcode')!;
+  const pi = spec.subjects.find(({ id }) => id === 'pi')!;
   assert.deepEqual(maka.credentials, ['DEEPSEEK_API_KEY']);
   assert.equal(maka.config.connectionSlug, 'env-deepseek');
   assert.equal(maka.config.baseUrl, 'https://api.deepseek.com');
@@ -657,6 +660,12 @@ test('eight-arm spec and wrappers freeze the working provider contracts', async 
   assert.equal(codex.config.args?.includes('--skip-git-repo-check'), true);
   assert.equal(claude.config.args?.includes('--bare'), true);
   assert.equal(claude.config.args?.includes('--effort'), true);
+  for (const subject of [codex, claude, reasonix]) {
+    assert.deepEqual(subject.config.args?.slice(-2), ['--', '{{task.input}}']);
+  }
+  assert.equal(zcode.config.args?.includes('{{task.cwd}}'), true);
+  assert.equal(zcode.config.args?.at(-1), '--prompt={{task.input}}');
+  assert.equal(pi.config.args?.at(-1), 'Task instruction:\n{{task.input}}');
   for (const subject of spec.subjects) {
     assert.deepEqual(subject.credentials, ['DEEPSEEK_API_KEY']);
     if (subject.id === 'maka') continue;
--- a/packages/eval/src/external-subject.ts
+++ b/packages/eval/src/external-subject.ts
@@ -62,6 +62,7 @@ export function createExternalSubjectAdapter(): SubjectAdapter {
         throw new Error(`${cell.subject.id} toolchain was not verified before execution`);
       }
       const startedAt = Date.now();
+      const profile = bundledProfile(config.args);
       try {
         const execution = await context.execute({
           command: config.command,
@@ -81,12 +82,14 @@ export function createExternalSubjectAdapter(): SubjectAdapter {
                 },
               }),
           ...(config.result === 'exit-code' ? { captureStdout: false } : {}),
+          ...(profile ? { recoveryPath: `/logs/agent/${profile}.provider-usage.json` } : {}),
         });
         const decoded = execution.stdout.length === 0 ? undefined : decodeResult(execution.stdout);
+        const recovered = decodeUsageCheckpoint(execution.recovery, profile);
         if (execution.termination === 'framework_timeout') {
           return {
-            usage: decoded?.usage ?? null,
-            costUsd: decoded?.costUsd ?? null,
+            usage: decoded?.usage ?? recovered?.usage ?? null,
+            costUsd: decoded?.costUsd ?? recovered?.costUsd ?? null,
             durationMs: Date.now() - startedAt,
             status: 'failed' as const,
             failureReason: 'external subject exceeded the framework timeout',
@@ -126,12 +129,14 @@ export function createExternalSubjectAdapter(): SubjectAdapter {
         }
         if (execution.diagnostic?.category.startsWith('result-frame-')) {
           return {
-            usage: null,
-            costUsd: null,
+            usage: recovered?.usage ?? null,
+            costUsd: recovered?.costUsd ?? null,
             durationMs: Date.now() - startedAt,
             status: context.signal?.aborted
               ? ('indeterminate' as const)
-              : ('infra_failed' as const),
+              : recovered && recovered.admittedRequests > 0
+                ? ('failed' as const)
+                : ('infra_failed' as const),
             failureReason: 'external subject result transport failed',
             artifacts: [{ kind: 'external_process', exitCode: execution.exitCode }],
           };
@@ -185,6 +190,52 @@ export function createExternalSubjectAdapter(): SubjectAdapter {
       }
     },
   };
+}
+
+function decodeUsageCheckpoint(
+  value: JsonObject | undefined,
+  profile: ExternalProfile | undefined,
+): { usage: NormalizedUsage | null; costUsd: number | null; admittedRequests: number } | undefined {
+  if (!value || !profile) return undefined;
+  try {
+    const checkpoint = exact(
+      value,
+      [
+        'schemaVersion',
+        'profile',
+        'usage',
+        'costUsd',
+        'requests',
+        'admittedRequests',
+        'usageRequests',
+        'usageComplete',
+        'removedWebTools',
+        'models',
+        'toolNames',
+      ],
+      'external usage checkpoint',
+    );
+    if (
+      checkpoint.schemaVersion !== 'maka.external_provider_usage.v1' ||
+      checkpoint.profile !== profile ||
+      typeof checkpoint.usageComplete !== 'boolean'
+    ) {
+      return undefined;
+    }
+    return {
+      usage: checkpoint.usage === null ? null : decodeUsage(checkpoint.usage),
+      costUsd:
+        checkpoint.costUsd === null
+          ? null
+          : nonnegative(checkpoint.costUsd, 'external checkpoint cost'),
+      admittedRequests: nonnegative(
+        checkpoint.admittedRequests,
+        'external checkpoint admitted requests',
+      ),
+    };
+  } catch {
+    return undefined;
+  }
 }
 
 interface ExternalSubjectConfig {
--- a/packages/eval/src/harbor-external-subject.ts
+++ b/packages/eval/src/harbor-external-subject.ts
@@ -10,7 +10,7 @@ import {
   decodePreverifiedToolchain,
   type ExternalProfile as Profile,
 } from './toolchain-verification.js';
-import { isInferenceAdmissionEvent } from './provider-admission.js';
+import { isInferenceAdmissionEvent, isSuccessfulInferenceResponse } from './provider-admission.js';
 import { removeEvalWebTools } from './provider-web-tool-surface.js';
 import { takeRelayResultToken, writeRelayResult } from './relay-result-frame.js';
 
@@ -657,7 +657,9 @@ async function startMeteringProxy(
         response.end();
       } finally {
         const parsed = parser.finish();
-        if (upstream.ok && parsed.admitted) {
+        const admitted =
+          parsed.admitted || isSuccessfulInferenceResponse(upstream.status, projected.model);
+        if (admitted) {
           admittedRequests += 1;
           if (parsed.usage) {
             usageRequests += 1;
--- a/packages/eval/src/harbor-external-subject.ts
+++ b/packages/eval/src/harbor-external-subject.ts
@@ -1,7 +1,7 @@
 import { spawn, spawnSync, type ChildProcess } from 'node:child_process';
 import { createHash } from 'node:crypto';
 import { readFileSync, rmSync, statSync } from 'node:fs';
-import { chmod, mkdir, writeFile } from 'node:fs/promises';
+import { chmod, mkdir, open, rename, writeFile } from 'node:fs/promises';
 import { createServer, type IncomingMessage, type Server } from 'node:http';
 import { basename, dirname, join } from 'node:path';
 import type { Readable } from 'node:stream';
@@ -18,6 +18,8 @@ const resultToken = takeRelayResultToken();
 
 type SubjectStatus = 'completed' | 'failed' | 'infra_failed' | 'indeterminate';
 const CLASSIFIABLE_RECORD_LIMIT_BYTES = 16 * 1024 * 1024;
+const PERSISTED_STREAM_LIMIT_BYTES = 64 * 1024 * 1024;
+let atomicWriteSequence = 0;
 
 interface Usage {
   inputTokens: number;
@@ -50,6 +52,9 @@ process.once('SIGINT', interrupt);
 
 const logsRoot = rooted(systemRoot, '/logs/agent');
 const statePath = join(logsRoot, `${profile}.wrapper-state.json`);
+const stdoutPath = join(logsRoot, `${profile}.jsonl`);
+const stderrPath = join(logsRoot, `${profile}.stderr.txt`);
+const usagePath = join(logsRoot, `${profile}.provider-usage.json`);
 const artifacts: Record<string, unknown>[] = [];
 let usage: Usage | null = null;
 let costUsd: number | null = null;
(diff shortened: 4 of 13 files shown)


packages/eval/experiments/terminal-bench-2.1-deepseek-v4-flash-eight-arm.json
```diff
@@ -287,7 +289,7 @@
           "deepseek-v4-flash",
           "--thinking",
           "max",
-          "{{task.input}}"
+          "Task instruction:\n{{task.input}}"
```
Astro-Han, line 292: **P0** — this is the only arm whose prompt content changed. The other six pass `{{task.input}}` verbatim; this one prepends `Task instruction:\n`, which changes the model's first tokens and how framed the instruction is. This file is the source of the cross-arm comparison, so Pi's column stops being comparable with the rest.

`lifecycle-boundaries.test.ts:684` freezes it as `assert.equal(pi.config.args?.at(-1), 'Task instruction:\n{{task.input}}')`, so from here it reads as contract rather than defect.

If Pi really does treat dash-leading positionals as options even after `--`, the options that preserve the prompt are a `=`-form flag (as zcode now uses), or stdin — the wrapper spawns with `stdio: ['ignore', 'pipe', 'pipe']` at `harbor-external-subject.ts:429`, so that is one central change rather than a third per-arm workaround. If neither works, dropping Pi from the matrix is more honest than comparing it.

**P2, same file** — `kimi-code` at :237-238 is still `"--prompt", "{{task.input}}"`, byte-identical to the shape you just declared broken for zcode and fixed at :260. It is not in the failing-arm list because the eighth arm was not run; enabling it fails the same way. One line.

Astro-Han (commented):
Thanks for chasing these down — the seven-arm matrix failing before model admission is a real problem and the diagnosis is sound. Two things block this one for me, both because they change numbers we publish rather than because the code is wrong in the abstract.

**Before anything else: the description covers two of the five commits.** `5f920fa66` and `82c22a0f2` are described; `3dd99bb20`, `cfa20fe34` and `eeec9529a` are not, and they carry the largest behavior change in the PR — stream persistence, a usage checkpoint recovered through the relay, and a second place where admission decides scoring. The validation list almost certainly predates them. Whatever else changes, the body needs to describe the diff.

**P0 — only Pi's prompt was modified.** Laying the eight arms side by side:

| arm | args tail |
|---|---|
| codex, claude-code, reasonix, opencode | `-- {{task.input}}` |
| zcode | `--prompt={{task.input}}` |
| kimi-code | `--prompt {{task.input}}` |
| pi | `Task instruction:\n{{task.input}}` |

Six arms receive `context.taskInput` verbatim. Pi receives it with a prefix. That is not a difference in argv syntax, it is a difference in the model's input — different first tokens, different framing of the instruction — and it lands in the spec that produces cross-arm comparison numbers. `lifecycle-boundaries.test.ts:684` then freezes it as an assertion, so nobody downstream will read it as a defect.

**P0 — admission now counts responses that never produced billable inference.** `origin/main` had `if (upstream.ok && parsed.admitted)`. This PR makes it `parsed.admitted || isSuccessfulInferenceResponse(upstream.status, projected.model)`, which drops the `upstream.ok` guard from the first branch entirely — so the body's "keeps non-2xx provider and transport failures classified as infrastructure failures" no longer holds. A 500 carrying a `choices` array is now admitted. And `isSuccessfulInferenceResponse` itself only asks for a 2xx plus a `model` string somewhere in the *request* body, ignoring the URL path and the response content; a live proxy run admits 200-wrapping-an-error, 200-with-empty-body, 204 No Content, and `/v1/messages/count_tokens` — which claude-code issues routinely, which returns 200, and which bills nothing. Each of those flips an attempt from `infra_failed` (not scored, `runner.ts:289`) to `failed` (verifier runs, reward 0 recorded, `runner.ts:296`).

**Root cause, and it is the same one twice.** Both halves substitute a proxy signal for the fact actually needed. Admission needs "the provider produced billable output" — that fact lives in `usageParser`, and `usageParser` is what is broken: it splits on `\n` and `JSON.parse`s each line, so a non-streaming, pretty-printed JSON body fails on every line and yields neither admission nor usage. The fallback makes the symptom disappear while the tokens and cost stay lost — a real billed pretty-JSON response still comes back `usage: null, costUsd: null, usageComplete: false`. Prompt delivery needs "every arm receives byte-identical task input", an invariant that is nowhere stated, so each arm's argv drifted on its own and one of them drifted into changing the prompt.

**Minimal path.** Restore the `upstream.ok` guard on the first branch. Narrow the fallback so it only applies when usage was parsed, which keeps the Reasonix cases while removing every row in that table. Give Pi the prompt unmodified — `=` form if it has such a flag, otherwise stdin (one change in the wrapper at `harbor-external-subject.ts:429`, currently `stdio: ['ignore', 'pipe', 'pipe']`), and failing both, drop Pi from the matrix rather than comparing it. Fix kimi-code's `--prompt {{task.input}}`, which is the exact shape you just declared broken for zcode. Replace the per-arm literal assertions with one parity assertion: after stripping flag syntax, every arm's prompt element equals `{{task.input}}`.

**Cleanest path.** Fix the parser instead of guarding around it: have `finish()` retry a whole-body `JSON.parse` when line-wise parsing produced nothing, which requires keeping the raw text rather than discarding `buffered`. That admits the Reasonix runs *and* recovers their usage and cost, and `isSuccessfulInferenceResponse` can be deleted rather than narrowed. Make prompt parity a spec-level invariant checked once for all arms instead of a literal per arm. And split this: the argv changes are self-evident and low risk, while the admission change rewrites scoring attribution and wants independent human review under AGENTS.md — right now it is spread across four commits and thirteen files, so it cannot be reverted on its own.

Two smaller notes. The new admission branch has no coverage — the integration children all send `body: '{}'` with no `model` field, so it never fires; reverting `harbor-external-subject.ts:762-763` to main's line leaves `provider-admission-integration.test.js` 4/4 green. And "escape" in the title is misleading: argv reaches the container through `shlex.join` plus `shlex.quote` (`relay_agent.py:296`, `:305`) and is spawned as an execve argv array, so no shell is involved and there is no injection surface — what these changes do is disambiguate each CLI's own option parser.

---

*Review assistance: Claude Code (Opus) ran parallel passes over prompt delivery, admission attribution, and test quality, reproduced the misclassification table against a live metering proxy, and drafted this comment. I verified independently the parts it turns on: the eight-arm args table by decoding the spec, the `upstream.ok` deletion against `origin/main`, the five commits against the description, and the second admission consumer in `external-subject.ts`. The live reproduction rows and the mutation results are AI output that I have not re-run myself.*

<details>
<summary>中文</summary>

先谢谢你把这些追出来——七臂矩阵在模型 admission 之前就失败是真问题，诊断也站得住。有两条我认为阻塞合并，原因不是代码抽象层面写错了，而是它们改的是我们会对外发布的数字。

**先说一件事：描述只覆盖了五个提交里的两个。** `5f920fa66` 和 `82c22a0f2` 写了；`3dd99bb20`、`cfa20fe34`、`eeec9529a` 一个字没有，而它们承载了本 PR 最大的行为变更——流持久化、经 relay 回收的 usage checkpoint，以及第二处由 admission 决定计分的地方。Validation 里的数字几乎可以肯定早于这三个提交。无论其他怎么改，body 都需要如实描述这个 diff。

**P0 —— 只有 Pi 的 prompt 被改了。** 八臂并排看：

| arm | args 末尾 |
|---|---|
| codex, claude-code, reasonix, opencode | `-- {{task.input}}` |
| zcode | `--prompt={{task.input}}` |
| kimi-code | `--prompt {{task.input}}` |
| pi | `Task instruction:\n{{task.input}}` |

六臂拿到的是逐字的 `context.taskInput`，Pi 拿到的多了一段前缀。这不是 argv 语法差异，是模型输入的差异——首 token 不同、指令的框架化程度不同——而且它落在产出跨臂对比数字的那份 spec 里。`lifecycle-boundaries.test.ts:684` 又把它冻结成断言，之后不会再有人把它读成缺陷。

**P0 —— admission 现在会把从未产生计费推理的响应算进去。** `origin/main` 是 `if (upstream.ok && parsed.admitted)`，本 PR 改成 `parsed.admitted || isSuccessfulInferenceResponse(upstream.status, projected.model)`，第一支的 `upstream.ok` 守卫被整个去掉了——所以正文里「keeps non-2xx provider and transport failures classified as infrastructure failures」不再成立，一个带 `choices` 的 500 现在算 admitted。而 `isSuccessfulInferenceResponse` 本身只要求 2xx 加上**请求体**里任意位置有个 `model` 字符串，不看 URL path、不看响应内容；真实起代理跑下来，200 包错误、200 空 body、204 No Content、以及 `/v1/messages/count_tokens`（claude-code 常规发送、返回 200、不计费）全部被判 admitted。每一条都会把 attempt 从 `infra_failed`（不计分，`runner.ts:289`）翻成 `failed`（跑 verifier、把 reward 0 写进 attempt，`runner.ts:296`）。

**根因，而且是同一个根因出现了两次。** 两半都用一个代理信号替换了真正需要的事实。admission 需要的事实是「provider 确实产出了可计费输出」——这个事实在 `usageParser` 里，而坏掉的正是 `usageParser`：它按 `\n` 切行、逐行 `JSON.parse`，于是非流式、被格式化过的 JSON 响应体每一行都失败，既拿不到 admission 也拿不到 usage。兜底让症状消失，token 和成本却仍然全丢——一次真实计费的 pretty JSON 响应回来仍是 `usage: null, costUsd: null, usageComplete: false`。prompt 传递需要的事实是「每一臂收到逐字节相同的任务输入」，而这条不变量没有被写在任何地方，于是各臂的 argv 各自漂移，其中一臂漂到了改 prompt 上。

**最小方案。** 恢复第一支的 `upstream.ok` 守卫。把兜底收窄成「仅当 usage 已解析成功时生效」，这样既保住 Reasonix 那两例，又消掉上表每一行。让 Pi 拿到未经修饰的 prompt——有 `=` 形式的 flag 就用，没有就走 stdin（改一处 wrapper，`harbor-external-subject.ts:429` 现在是 `stdio: ['ignore', 'pipe', 'pipe']`）；两条都不行，就把 Pi 从矩阵里摘出来而不是拿它对比。修掉 kimi-code 的 `--prompt {{task.input}}`，那正是你刚判定 zcode 要修的形态。把逐臂冻结字面量换成一条 parity 断言：剥掉 flag 语法后，每一臂的 prompt 元素都等于 `{{task.input}}`。

**最彻底方案。** 修解析器而不是在它外面加护栏：让 `finish()` 在逐行解析没有产出时，对累积的完整 body 再整体 `JSON.parse` 一次（需要保留原始文本，目前 `buffered` 被丢弃）。这样 Reasonix 那几次既能正确 admitted，usage 和成本也一起修好，`isSuccessfulInferenceResponse` 可以直接删掉而不是收窄。把 prompt parity 做成 spec 级不变量、对所有臂检查一次，而不是每臂一条字面量。另外这个 PR 该拆：argv 那部分自证、风险低，而 admission 改的是计分归因，按 AGENTS.md 需要独立人审——它现在摊在四个提交、十三个文件里，没法单独 revert。

两条小的。新的 admission 分支没有覆盖——集成测试的 child 全部发 `body: '{}'`、没有 `model` 字段，所以它从未触发；把 `harbor-external-subject.ts:762-763` 还原成 main 那一行，`provider-admission-integration.test.js` 仍是 4/4 全绿。另外标题里的 "escape" 有误导：argv 经 `shlex.join` + `shlex.quote`（`relay_agent.py:296`、`:305`）进容器，再以 execve argv 数组 spawn，全程没有 shell，不存在注入面——这些改动做的是消解各 CLI 自己的选项解析歧义。

---

*评审协助说明：Claude Code (Opus) 并行跑了 prompt 传递、admission 归因与测试质量三轮审查，用真实 metering proxy 复现了那张误判表，并起草了本条评论。结论所依赖的部分我独立核实过：解码 spec 得到的八臂 args 表、对照 `origin/main` 确认 `upstream.ok` 被删、五个提交与描述的对照，以及 `external-subject.ts` 里第二处 admission 消费点。表中的实时复现行与变异测试结果属于 AI 输出，我没有自己重跑。*

</details>

packages/eval/src/harbor-external-subject.ts
```diff
@@ -657,18 +759,22 @@ async function startMeteringProxy(
         response.end();
       } finally {
         const parsed = parser.finish();
-        if (upstream.ok && parsed.admitted) {
+        const admitted =
```
Astro-Han, line 762: **P0** — two things, one line.

`origin/main:660` was `if (upstream.ok && parsed.admitted)`. The new expression drops `upstream.ok` from the `parsed.admitted` branch, so a non-2xx response whose body happens to carry a recognized envelope is now admitted — `isInferenceAdmissionEvent` matches on `Array.isArray(event.choices)` alone, so a 500 with a `choices` body qualifies. That directly contradicts "keeps non-2xx provider and transport failures classified as infrastructure failures".

And `isSuccessfulInferenceResponse` asks only for 2xx plus a `model` string in the *request* body — no URL path, no response content. A live proxy run admits 200-wrapping-an-error, 200-with-empty-body, 204 No Content, and `/v1/messages/count_tokens`, which claude-code issues routinely and which bills nothing. Each flips `infra_failed` (not scored) to `failed` (reward 0 recorded).

There is also a quieter cost: `usageComplete` requires `usageRequests === admittedRequests` (:705, :792), so every newly-admitted request with no parseable usage pushes `usageComplete` false and drops `costUsd` to null at :101.

The narrow fix is `(upstream.ok && parsed.admitted) || isSuccessfulInferenceResponse(...)` plus requiring parsed usage before the fallback applies. The fix I would rather see is below at `usageParser` — the Reasonix cases are a parse gap, not an admission gap.

packages/eval/src/harbor-external-subject.ts
```diff
@@ -687,14 +793,34 @@ async function startMeteringProxy(
...
+async function writeAll(output: Awaited<ReturnType<typeof open>>, value: Buffer): Promise<void> {
+  let offset = 0;
+  while (offset < value.byteLength) {
+    const { bytesWritten } = await output.write(value.subarray(offset));
+    if (bytesWritten <= 0) throw new Error('external trajectory write did not advance');
+    offset += bytesWritten;
+  }
+}
+
 function usageParser(anthropic: boolean): {
```
Astro-Han, line 824: **P2** — this is where the Reasonix cases actually fail. Splitting on `\n` and `JSON.parse`-ing each line only works for SSE and JSONL; a non-streaming, pretty-printed JSON body fails on every line, so the request yields neither admission nor usage.

The fallback added above makes the admission half disappear while the usage half stays broken — a genuinely billed pretty-JSON response still returns `usage: null, costUsd: null, usageComplete: false`. Retrying a whole-body `JSON.parse` in `finish()` when line-wise parsing produced nothing fixes both, and lets `isSuccessfulInferenceResponse` be deleted rather than narrowed. It needs the raw text kept rather than `buffered` being discarded line by line.

Worth attaching the two Reasonix responses to the PR — without the actual envelopes there is no way to tell whether widening admission is even the right shape for them.

packages/eval/src/external-subject.ts
```diff
@@ -126,12 +129,14 @@ export function createExternalSubjectAdapter(): SubjectAdapter {
...
           return {
-            usage: null,
-            costUsd: null,
+            usage: recovered?.usage ?? null,
+            costUsd: recovered?.costUsd ?? null,
             durationMs: Date.now() - startedAt,
             status: context.signal?.aborted
               ? ('indeterminate' as const)
-              : ('infra_failed' as const),
+              : recovered && recovered.admittedRequests > 0
```
Astro-Han, line 137: **P2** — a second place where admission decides scoring, and it inherits everything above: any of the misclassified 2xx cases makes `admittedRequests > 0` here and turns a result-transport failure into a scored `failed`.

Test coverage does not reach it. The new `external-subject.test.ts` case loops two executions, but the first has `termination: 'framework_timeout'` and returns at :89 before ever reaching this branch — both paths happen to produce `'failed'`, so the assertion passes while only testing one. Mutating :137-139 to a constant `'failed'` leaves that file 8/8 green. A case with `admittedRequests: 0` expecting `infra_failed` would close it.

packages/eval/src/provider-admission.ts
```diff
@@ -19,6 +19,10 @@ export function isInferenceAdmissionEvent(
   return false;
 }
 
+export function isSuccessfulInferenceResponse(status: number, model: string | undefined): boolean {
```
Astro-Han, line 22: **P2** — `provider-admission.test.ts:24` asserts 204 is admitted. 204 No Content means the provider returned no output at all, so this writes a wrong semantic into the contract; it should be false, and the predicate narrowed to 200.

The rest of that test restates the two-line implementation — a range check and a non-empty string check — rather than exercising a behavior. The coverage that is missing is at the integration level: every child there sends `body: '{}'` with no `model` field, so this predicate never returns true in any integration case, and reverting `harbor-external-subject.ts:762-763` to main's line leaves `provider-admission-integration.test.js` 4/4 green.

packages/eval/src/__tests__/lifecycle-boundaries.test.ts
```diff
@@ -657,6 +676,12 @@ test('eight-arm spec and wrappers freeze the working provider contracts', async
   assert.equal(codex.config.args?.includes('--skip-git-repo-check'), true);
   assert.equal(claude.config.args?.includes('--bare'), true);
   assert.equal(claude.config.args?.includes('--effort'), true);
+  for (const subject of [codex, claude, reasonix]) {
+    assert.deepEqual(subject.config.args?.slice(-2), ['--', '{{task.input}}']);
+  }
+  assert.equal(zcode.config.args?.includes('{{task.cwd}}'), true);
+  assert.equal(zcode.config.args?.at(-1), '--prompt={{task.input}}');
+  assert.equal(pi.config.args?.at(-1), 'Task instruction:\n{{task.input}}');
```
Astro-Han, line 684: **P3** — freezing each arm's literal separately is what let Pi's prompt drift without anyone noticing; the assertions record what each arm does rather than what all arms must share.

One parity assertion would carry the invariant instead: strip the flag syntax (`--` separator, `--flag=` prefix) and require every arm's prompt element to equal `{{task.input}}`. That fails loudly on the next arm that improvises, and it is the natural place to state the rule — nothing in the repo currently says the task input must reach every subject unmodified.

hqhq1025:
Real-machine validation update (VMSS, Terminal-Bench 2.1):

- Fixed external CLI trajectory loss by streaming stdout/stderr to bounded `/logs/agent/<profile>.jsonl` and `.stderr.txt` files (64 MiB cap with observed/persisted/truncated bytes and SHA-256).
- Added atomic per-request provider usage checkpoints at `/logs/agent/<profile>.provider-usage.json`. Framework timeout/result-frame loss now recovers already-settled usage/cost without inventing usage for an in-flight request.
- Moved collected artifact inventory to executor finalization, so completed, subject-failed, infra-failed, and indeterminate attempts all record collected file bytes and SHA-256.
- Preserved native task timeout and verifier execution after timeout when model admission/settled evidence exists.

Validation:
- Eval Node 33/33; relay contract 10/10; lifecycle 12/12; egress 3/3; policy 2/2; artifact 2/2.
- OpenCode canary: score=1, usage=32,908 tokens; JSONL/checkpoint/stderr/wrapper-state hashes matched collected artifacts.
- Gracefully cancelled infra canary: indeterminate attempt still retained all four collected artifact hashes.
- Claude `model-extraction-relu-logits`: native 900s agent timeout, verifier score=1, recovered 94,225 tokens, 8.9 MB JSONL, all hashes recorded.
- Claude `raman-fitting`: score=0, recovered 2,777,464 tokens.
- Reasonix `make-doom-for-mips`: score=0, recovered 3,963,729 tokens.

Remaining non-Eval gaps reproduced with complete evidence: Reasonix adaptive/gpt2 and Kimi adaptive fail before model admission, so no provider usage exists to recover.

Validated deployment commit: `eeec9529a`.

Astro-Han:
Writing as a downstream consumer of `admittedRequests` rather than as a reviewer of the fix — #2971 adds a DeepSeek Harness arm that reads this field, so the widening in `harbor-external-subject.ts:762-763` lands on that arm directly.

The problem being solved is real. The fact you want is "the provider produced billable output", and that fact currently lives in `usageParser`, which is the part that is broken. Reaching for a proxy signal is a reasonable response to that. What I want to flag is where this particular proxy signal lands, because I do not think the blast radius is visible from the diff.

`main` requires both conditions:

```ts
if (upstream.ok && parsed.admitted)
```

The PR requires either of two, neither of which is that:

```ts
const admitted = parsed.admitted || isSuccessfulInferenceResponse(upstream.status, projected.model)
```

Two separate widenings. `parsed.admitted` no longer requires `upstream.ok`, so a non-2xx response whose stream happened to carry an admission event now counts. And `isSuccessfulInferenceResponse` is `2xx && typeof model === 'string' && model.length > 0` (`provider-admission.ts:22`), so a 2xx that produced no usage at all counts too.

**Three consumers in my arm, and two of them decide whether an attempt is retried.**

1. `classifyExecution` (`harbor-external-subject.ts:621`) returns `infra_failed` when `admittedRequests === 0`.
2. The interrupted-settlement recovery paths (`external-subject.ts:145,175,206,211`) choose `failed` over `infra_failed` on `recovered.admittedRequests > 0`.
3. `usageComplete` is `inFlightRequests === 0 && admittedRequests > 0 && usageRequests === admittedRequests` (`harbor-external-subject.ts:707`).

The first two are the same boundary, and that boundary is the difference between an infrastructure retry and a recorded result. Widening admission moves an attempt from "retry it" to "the subject failed" — which scores zero and does not retry. A provider 500 whose stream carried an admission event, or a 2xx that yielded no usage, becomes a silent zero rather than a retried cell.

The third one moves the other way: `admittedRequests` grows while `usageRequests` does not, so `usageRequests === admittedRequests` fails more often, `usageComplete` goes false and `costUsd` goes null. Cost attribution gets sparser exactly on the requests this change was meant to account for.

**The part I most want on the record: in the head-to-head cohort this is one-sided.** `experiments/terminal-bench-2.1-deepseek-v4-flash-maka-vs-deepseek-harness.json` pairs Maka against the harness arm in one task group. The Maka arm is `kind: maka` and runs through `harbor-maka-subject.js` — it never touches this metering proxy. So a change to admission semantics shifts the failure classification of one arm and not the other. A classification shift that only moves one arm is the specific thing a paired comparison cannot absorb.

I am not asking for a particular fix, and I do not think the answer is to revert to `upstream.ok && parsed.admitted` if that is what is dropping real usage. Two things that would resolve it for me:

- Keep `upstream.ok` as a conjunct. The second widening is arguably defensible on its own; dropping the status check is a separate change that I do not think the described problem requires.
- If admission has to widen, separate "admitted" from "billable" so `usageComplete` and the retry boundary do not both move on one signal. My arm needs the retry boundary to stay conservative far more than it needs admission to be generous.

Happy to adapt on my side instead if you would rather the semantics change and the consumers follow — but then I would want to land that after #2971, so the arm is not written against a definition that is mid-flight.

Separately, and unrelated to the above: this PR and #2971 both read `/logs/agent/<profile>.provider-usage.json` under schema `maka.external_provider_usage.v1`, yours from the relay before teardown, mine host-side from the trial directory. Those look complementary rather than competing, but the schema needs one owner. I have deferred adding a field to it (provider finish reasons) specifically because this PR is in flight.

jackwener (changes requested):
Automated review of exact head `eeec9529ad69dd0be90408f8e35deaaea871181e` against current `main@d62857a8357e9160926726a2a13096bc2dc2b91d`.

The reported failures still exist on current main: external arguments interpolate dash-leading task text without a profile-aware option boundary, and admission still depends only on recognized response envelopes. The repair direction is therefore relevant, but this revision is not safe to transplant as written.

**P1 — recovered metering is subject-forgeable and unbounded.** The wrapper writes `${profile}.provider-usage.json` inside `/logs/agent`, the same writable environment in which the measured external CLI runs. After the child exits, `relay_agent.py:_read_recovery` downloads and JSON-parses that path with no authentication, regular-file proof, or byte cap, and `external-subject.ts:decodeUsageCheckpoint` trusts its usage, cost, and `admittedRequests` to decide billing and whether a missing result is an infrastructure retry or a scored subject failure. A task-controlled process can replace the file, forge a non-retryable zero/result, or exhaust host disk/memory during download. Current main already has the correct single authority: signed, bounded, atomic `metering-checkpoint` data keyed by `meteringSecret`; rebasing must preserve it rather than reintroduce this older plain-JSON protocol.

**P1 — non-2xx responses can become admitted attempts.** `harbor-external-subject.ts:762-764` uses `parsed.admitted || isSuccessfulInferenceResponse(...)`; the first branch no longer requires `upstream.ok`. An error response carrying a recognizable event therefore increments `admittedRequests`, moves recovery from `infra_failed` to `failed`, suppresses retry, and can record a zero. Keep success status as a conjunct for the retry boundary; if “billable” and “admitted” need different widening, represent them as different facts instead of sharing one counter.

Required conclusions:

1. Optimal for the actual problem: **no**; useful prompt/artifact changes are coupled to a superseded metering authority.
2. Production code to delete: delete the PR's duplicate plain-JSON checkpoint/recovery path after rebasing; reuse current main's signed checkpoint.
3. Tests to delete/replace: replace fixture-written trusted checkpoint tests with adversarial forged/oversized/non-regular checkpoint cases against the real signed seam.
4. Deeper refactor: rebase first, preserve the current metering authority, then transplant the remaining prompt escaping, artifact inventory, and narrowly defined admission change.
5. Ready to merge: **no**; it is also currently conflicting with main.
6. Residual risks/gaps: Eval scoring, retry classification, and usage/cost are protected contract behavior and require independent human review plus exact-head CI and real-host evidence after rebase.

Astro-Han:

- `packages/eval/src/harbor-external-subject.ts`

```
# resolve, then
git push --force-with-lease
```

---

packages/eval/src/harbor-external-subject.ts
```diff
@@ -609,6 +692,25 @@ async function startMeteringProxy(
...
+      costUsd: measured ? estimateCost(total) : null,
+      requests,
+      admittedRequests,
+      usageRequests,
+      usageComplete: admittedRequests > 0 && usageRequests === admittedRequests,
+      removedWebTools,
+      models: [...requestModels].sort(),
+      toolNames: [...observedToolNames].sort(),
+    };
+    checkpointWrites = checkpointWrites.then(() => writeJsonAtomic(checkpointPath, value));
```
Astro-Han, line 710: [P1] Resolve the rebase conflict by deleting this unsigned v1 checkpoint authority. This snapshot is written without a MAC and the PR's recovery path later trusts it from subject-writable `/logs/agent`, so choosing or blending this side would let the measured process forge usage, cost, and `admittedRequests`. Current main already has the canonical replacement: `signMeteringCheckpoint(..., resultToken)`, a bounded `O_NOFOLLOW` regular-file read, v2 settlement/in-flight fields, and shared derivation of completeness/cost. Preserve that entire authority and transplant only the independent prompt/parser/artifact changes; do not revive this parallel schema or its fixture-trusted tests.

Astro-Han (commented):
NO-GO at exact head `eeec9529ad69dd0be90408f8e35deaaea871181e`. One integration P1 is inline. Multiple existing, non-outdated unresolved threads already cover the prompt-parity, admission, parser, retry-classification, and recovered-metering defects on this head; the existing CHANGES_REQUESTED review remains current. The branch is 334 commits behind current `main`. A real five-commit rebase probe conflicts in `packages/eval/src/harbor-external-subject.ts`; conflict resolution must retain current main's signed, bounded v2 metering/checkpoint authority and transplant only orthogonal fixes. The old exact-head CI is green, but it predates and does not validate that integration.

packages/eval/src/harbor-external-subject.ts
Astro-Han: Thanks — one additional correctness reason to preserve the current-main checkpoint authority here: this branch chains `checkpointWrites = checkpointWrites.then(() => writeJsonAtomic(...))` without containing a single write rejection. After an `ENOSPC`, `EACCES`, or rename failure, every later `persistCheckpoint()` stays rejected; both `proxy.settle()` and `close()` await that poisoned chain, so a provider request that already completed successfully is ultimately reported as `infra_failed` because an auxiliary recovery snapshot could not be refreshed. Current main deliberately catches each atomic-write failure and keeps the previous truthful lower-bound snapshot. Please preserve that behavior during the rebase as well. If checkpoint persistence is intended to be a hard run gate rather than best-effort recovery evidence, please push back and clarify that invariant.

再补充一个保留 current-main checkpoint authority 的正确性原因：这里一次原子写失败会永久污染 promise chain，随后让本已成功的运行在 settle/close 时误报 `infra_failed`。current main 明确把单次写失败收敛为保留旧的真实下界快照；重基时也请保留这个不变量。若 checkpoint 写入本来就是硬门禁，也欢迎直接说明。

</details>

Issue #2369, opened 2026-08-07
feat(cloud): coordinate quiescent Session Bundle snapshots


MicroGery:
Parent RFC: #1286

Depends on:

- #1338 / #1349 — isolated Session-owned state export roots; landed
- #1336 / #1353 — Session Bundle size, storage, and workspace materialization policy; landed
- #1528 / #1697 / #2013 — Session Bundle filesystem codec; landed

Unblocks:

- #1415 — fork Cloud Sessions at committed revisions
- future Activation commit coordination and fresh-sandbox E2E work from #1286

## Summary

Define and implement the coordinator that produces one consistent, filtered, quiescent `state/ + workspace/` snapshot for `SessionBundleFileService.pack`.

The coordinator owns the transition from live mutable Session roots to an immutable private staging snapshot. The filesystem codec continues to own only deterministic archive encoding, validation, and hydration.

## Problem

`SessionBundleFileService.pack` accepts a `PreparedSessionBundleSnapshot` and assumes that:

- state and workspace represent one logical point in time;
- SQLite and other mutable stores are safe to copy;
- the roots do not change while packing;
- credentials, temporary execution state, logs, and excluded workspace content are absent.

Those guarantees do not exist when `state/` and `workspace/` are copied directly from a live Session. A state export and workspace copy taken at different times can produce a structurally valid but semantically inconsistent Bundle.

## Responsibility boundary

The coordinator owns:

- acquiring a per-Session safe/quiescent boundary;
- coordinating the state export and workspace copy under that boundary;
- applying a versioned snapshot/filtering policy;
- producing the opaque state-identity descriptor through the state-owning layer;
- publishing a private immutable staging snapshot;
- retaining cleanup ownership until the caller releases the snapshot;
- deterministic failure cleanup and bounded stable errors.

It does not own:

- USTAR/Zstandard encoding or Bundle validation;
- Repository/blob persistence, revision allocation, or CAS;
- Activation leases, deduplication, reply publication, or sandbox lifecycle;
- Cloud Session fork orchestration or identity re-keying.

## Proposed lifetime contract

V1 uses a short quiescence interval followed by a private staging copy. It does not keep the live Session blocked while compression and upload run.

Conceptually:

```ts
interface PreparedSessionBundleHandle {
  snapshot: PreparedSessionBundleSnapshot;

  /** Idempotently deletes the coordinator-owned private snapshot. */
  release(): Promise<void>;
}

interface QuiescentSessionSnapshotCoordinator {
[...]

Issue #1415, opened 2026-07-23
feat(cloud): fork Cloud Sessions at committed revisions


Astro-Han:
Split from @MicroGery's proposal in #1286 ([#issuecomment-5059344338](https://github.com/maka-agent/maka-agent/issues/1286#issuecomment-5059344338)). Parent RFC: #1286.

## Motivation

A portable bundle makes fork a control-plane/repository operation, with no dependency on OpenSandbox or `maka activate`. Besides user-driven duplication of a session, it enables **template sessions**: warm one base session (workspace set up, dependencies installed, initial context established), then fork per user/session. Every fork starts from the same conversation prefix, so provider prompt caches keep hitting, and inherits a warm workspace, so setup does not rerun per session.

## Definition

Create a new Cloud Session from the complete bundle of one committed source revision; source and target then evolve independently. The target keeps the same `agentId`, receives a new `sessionId`, and starts its own revision/CAS sequence. This is a **Cloud Session fork**, not a sandbox fork or an agent-definition fork.

## Contract

```ts
type SessionRevisionRef = { sessionId: string; revision: string };

type ForkSessionRequest = {
  forkId: string; // idempotency key
  source: SessionRevisionRef;
  targetSessionId: string;
};

type ForkSessionResult = {
  target: SessionRevisionRef;
  forkedFrom: SessionRevisionRef;
};
```

- `source.revision` is required and never defaults to "latest". This closes the race where the source advances between the decision to fork and the operation.
- The target's first revision is a new revision in the target session; it must not reuse the source revision as its CAS token.
- Lineage lives in control-plane session metadata as `forkedFrom: { sessionId, revision }` (a revision is only unique within its session). It stays separate from Maka's conversation-level `parentSessionId` / `branchOfTurnId`: `branchFromTurn` branches conversation history at a turn boundary, while a Cloud Session fork copies the complete committed snapshot including runtime state, artifacts, and workspace files. Do not overload the Maka lineage fields.

## MVP boundary

- Fork only an explicitly named current committed revision; if the head has moved, fail `source_revision_not_available`. Never silently fork a newer revision.
- A fork does not acquire the source session's write lease: read the exact committed revision, verify it is forkable, create the target with create-if-absent, and persist `forkId -> targetSessionId` so retries return the same target. The source is never modified.
- Only **quiescent** revisions are forkable: reject a revision with a pending approval or another externally resumable action until action identities are re-keyed for the target session.

## Copied vs not copied

[...]

PR #2968, opened 2026-08-13
feat(storage): define quiescent session snapshot boundary

MicroGery:
## Summary

Define the contract/foundation slice of the quiescent Session Bundle snapshot coordinator from #2369.

- add a trusted per-Session quiescence boundary that prepares state and workspace at one logical point in time, then returns private staging roots compatible with `PreparedSessionBundleSnapshot`
- pin the exact V1 workspace policy so callers cannot replace or weaken it; the trusted workspace preparer is the enforcement point and production traversal/copy enforcement remains for PR 2
- deterministically exclude rebuildable/runtime content and logs, while rejecting only names that identify known user-authored secret material; public certificate encodings are not rejected by extension alone
- add cancellation/deadline handling, bounded stable errors, normalized exclusion diagnostics, and same-Session serialization requirements
- retain inode-bound cleanup ownership outside the recursively deleted snapshot directory so partial cleanup remains retryable and replacements observable before deletion fail closed
- require a caller-provided ACL verification authority on Windows and bind its result to the exact canonical parent and newly created staging directory

This is PR 1 of the implementation split described in #2369. It establishes contracts and deterministic fakes, but has no production call site and does not solve #2369 end to end. Production quiescence/state/workspace adapters, codec round-trip integration, and mutation-race E2E coverage remain for PR 2.

Refs #2369

## Trust and security boundaries

- `SessionSnapshotQuiescenceAuthority` is a trusted Host/Owner boundary; this PR intentionally does not present a process-local mutex as authoritative.
- The coordinator always supplies `SESSION_SNAPSHOT_WORKSPACE_POLICY_V1`. `SessionSnapshotWorkspacePreparer` is the trusted enforcement point: it must apply every decision without downgrading it. The coordinator validates the returned root and bounded counters, but does not re-traverse the result to attest policy compliance. PR 2 must provide and test the production preparer before that guarantee is claimed end to end.
- Ownership records bind the snapshot directory and record file to their filesystem identities. The current checks detect replacements observable before cleanup and support retry after partial deletion.
- The private staging parent and code running as the same OS principal with access to it are inside the trusted computing boundary. Node's path-based recursive removal does not defend against an adversarial same-principal replacement between the final identity check and deletion. Adversarial same-principal deletion and orphan lifecycle management require a separately specified platform mechanism.
- This PR does not expose prefix-based orphan cleanup; any future orphan API must authenticate an exact snapshot/owner record.

## Remaining PR 2 integration

- production quiescence authority covering every Session writer
- single-Session state exporter
- policy-enforcing workspace traversal and copier
- production call site
- `prepare -> pack -> inspect -> hydrate` round-trip coverage
- real state/workspace mutation-race coverage

## Verification

- `npm run build -w /core`
- `npm test -w /storage` — 1007 tests, 990 passed, 17 skipped, 0 failed
- `node --test packages/storage/dist/__tests__/quiescent-session-snapshot.test.js` — 19 tests, 18 passed, 1 Windows-only skip, 0 failed
- `npx biome check packages/storage/src/quiescent-session-snapshot.ts packages/storage/src/__tests__/quiescent-session-snapshot.test.ts packages/storage/src/session-bundle-ustar.ts`
- `git diff --check`

## AI use

Select exactly one:

Tool(s) and scope: OpenAI Codex assisted with contract design, implementation, regression tests, review remediation, and local verification. The contributor reviewed the resulting changes and remains responsible for their accuracy, provenance, and licensing.

## Checklist

- [ ] Yes
- [x] No — this PR defines a foundation contract and has no production call site

### DIFF
--- a/packages/storage/src/__tests__/quiescent-session-snapshot.test.ts
+++ b/packages/storage/src/__tests__/quiescent-session-snapshot.test.ts
@@ -1,0 +1,744 @@
+import assert from 'node:assert/strict';
+import {
+  mkdir,
+  mkdtemp,
+  readFile,
+  readdir,
+  realpath,
+  rename,
+  rm,
+  writeFile,
+} from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { basename, dirname, join } from 'node:path';
+import { afterEach, test } from 'node:test';
+import {
+  createFileQuiescentSessionSnapshotCoordinator,
+  type SessionSnapshotCancellation,
+  SessionSnapshotError,
+  type SessionSnapshotQuiescenceAuthority,
+  type SessionSnapshotStatePreparer,
+  type SessionSnapshotWorkspacePreparation,
+  type SessionSnapshotWorkspacePreparer,
+  SESSION_SNAPSHOT_WORKSPACE_POLICY_V1,
+} from '../quiescent-session-snapshot.js';
+
+const roots: string[] = [];
+
+afterEach(async () => {
+  await Promise.all(roots.splice(0).map((root) => rm(root, { recursive: true, force: true })));
+});
+
+test('prepares state and workspace under one quiescence boundary, then releases live writers', async () => {
+  const fixture = await createFixture();
+  let liveState = 'state-at-boundary';
+  let liveWorkspace = 'workspace-at-boundary';
+  const events: string[] = [];
+  let quiescent = false;
+
+  const handle = await createFileQuiescentSessionSnapshotCoordinator({
+    stagingParent: fixture.stagingParent,
+    privateStagingRootAuthority,
+    quiescence: {
+      async runQuiescent(_input, operation) {
+        assert.equal(quiescent, false);
+        quiescent = true;
+        events.push('enter');
+        try {
+          return await operation();
+        } finally {
+          quiescent = false;
+          events.push('exit');
+        }
+      },
+    },
+    state: {
+      async prepareState(input) {
+        assert.equal(quiescent, true);
+        events.push('state');
+        await mkdir(input.destinationRoot);
+        await writeFile(join(input.destinationRoot, 'runtime.sqlite'), liveState, 'utf8');
+        return {
+          mediaType: 'application/vnd.maka.session-state-identity+json;version=1',
+          bytes: Buffer.from('{"makaSessionId":"session-1"}', 'utf8'),
+        };
+      },
+    },
+    workspace: {
+      async prepareWorkspace(input) {
+        assert.equal(quiescent, true);
+        assert.equal(input.policy.version, 1);
+        events.push('workspace');
+        await mkdir(input.destinationRoot);
+        await writeFile(join(input.destinationRoot, 'main.ts'), liveWorkspace, 'utf8');
+        return workspaceResult({ includedEntries: 1 });
+      },
+    },
+  }).prepare({ makaSessionId: 'session-1' });
+
+  assert.deepEqual(events, ['enter', 'state', 'workspace', 'exit']);
+  assert.equal(quiescent, false);
+  assert.notEqual(handle.snapshot.stateRoot, fixture.liveStateRoot);
+  assert.notEqual(handle.snapshot.workspaceRoot, fixture.liveWorkspaceRoot);
+  assert.equal(
+    await readFile(join(handle.snapshot.stateRoot, 'runtime.sqlite'), 'utf8'),
+    liveState,
+  );
+  assert.equal(
+    await readFile(join(handle.snapshot.workspaceRoot, 'main.ts'), 'utf8'),
+    liveWorkspace,
+  );
+
+  liveState = 'later-state';
+  liveWorkspace = 'later-workspace';
+  assert.equal(
+    await readFile(join(handle.snapshot.stateRoot, 'runtime.sqlite'), 'utf8'),
+    'state-at-boundary',
+  );
+  assert.equal(
+    await readFile(join(handle.snapshot.workspaceRoot, 'main.ts'), 'utf8'),
+    'workspace-at-boundary',
+  );
+  assert.deepEqual(handle.workspace, workspaceResult({ includedEntries: 1 }));
+
+  const publishedRoot = dirname(handle.snapshot.stateRoot);
+  const cleanupRename = interceptSnapshotCleanupRename(publishedRoot, async () => {
+    await mkdir(publishedRoot, { mode: 0o700 });
+    await writeFile(join(publishedRoot, 'replacement.txt'), 'keep', 'utf8');
+  });
+  await Promise.all([handle.release(), handle.release()]);
+  await cleanupRename.completed;
+  assert.equal(await readFile(join(publishedRoot, 'replacement.txt'), 'utf8'), 'keep');
+  await rm(publishedRoot, { recursive: true });
+  await handle.release();
+});
+
+test('serializes concurrent preparations for the same Session through the authority contract', async () => {
+  const fixture = await createFixture();
+  const authority = new SerialQuiescenceAuthority();
+  const firstWorkspaceEntered = deferred<void>();
+  const allowFirstWorkspace = deferred<void>();
+  let statePreparations = 0;
+
+  const coordinator = createFileQuiescentSessionSnapshotCoordinator({
+    stagingParent: fixture.stagingParent,
+    privateStagingRootAuthority,
+    quiescence: authority,
+    state: {
+      async prepareState(input) {
+        statePreparations += 1;
+        await mkdir(input.destinationRoot);
+        return stateIdentity(input.makaSessionId);
+      },
+    },
+    workspace: {
+      async prepareWorkspace(input) {
+        await mkdir(input.destinationRoot);
+        if (statePreparations === 1) {
+          firstWorkspaceEntered.resolve();
+          await allowFirstWorkspace.promise;
+        }
+        return workspaceResult();
+      },
+    },
+  });
+
+  const first = coordinator.prepare({ makaSessionId: 'same-session' });
+  await firstWorkspaceEntered.promise;
+  const second = coordinator.prepare({ makaSessionId: 'same-session' });
+  await Promise.resolve();
+  assert.equal(statePreparations, 1);
+  assert.deepEqual(authority.activeSessions, ['same-session']);
+
+  allowFirstWorkspace.resolve();
+  const firstHandle = await first;
+  const secondHandle = await second;
+  assert.equal(statePreparations, 2);
+  assert.equal(authority.maximumConcurrentBySession.get('same-session'), 1);
+  await Promise.all([firstHandle.release(), secondHandle.release()]);
+});
+
+test('does not globally serialize preparations for different Sessions', async () => {
+  const fixture = await createFixture();
+  const authority = new SerialQuiescenceAuthority();
+  const firstWorkspaceEntered = deferred<void>();
+  const allowFirstWorkspace = deferred<void>();
+  const coordinator = createFileQuiescentSessionSnapshotCoordinator({
+    stagingParent: fixture.stagingParent,
+    privateStagingRootAuthority,
+    quiescence: authority,
+    state: directoryStatePreparer,
+    workspace: {
+      async prepareWorkspace(input) {
+        await mkdir(input.destinationRoot);
+        if (input.makaSessionId === 'session-a') {
+          firstWorkspaceEntered.resolve();
+          await allowFirstWorkspace.promise;
+        }
+        return workspaceResult();
+      },
+    },
+  });
+
+  const first = coordinator.prepare({ makaSessionId: 'session-a' });
+  await firstWorkspaceEntered.promise;
+  const secondHandle = await coordinator.prepare({ makaSessionId: 'session-b' });
+  assert.deepEqual(authority.activeSessions, ['session-a']);
+  assert.equal(authority.maximumConcurrentBySession.get('session-a'), 1);
+  assert.equal(authority.maximumConcurrentBySession.get('session-b'), 1);
+
+  allowFirstWorkspace.resolve();
+  const firstHandle = await first;
+  await Promise.all([firstHandle.release(), secondHandle.release()]);
+});
+
+test('removes partial staging and preserves a stable policy rejection', async () => {
+  const fixture = await createFixture();
+  const rejection = new SessionSnapshotError(
+    'policy_rejected',
+    'Workspace snapshot policy rejected an entry',
+    { details: { phase: 'workspace', policyCategory: 'known_secret_file' } },
+  );
+  const coordinator = createFileQuiescentSessionSnapshotCoordinator({
+    stagingParent: fixture.stagingParent,
+    privateStagingRootAuthority,
+    quiescence: immediateAuthority,
+    state: {
+      async prepareState(input) {
+        await mkdir(input.destinationRoot);
+        await writeFile(join(input.destinationRoot, 'runtime.sqlite'), 'partial', 'utf8');
+        return stateIdentity(input.makaSessionId);
+      },
+    },
+    workspace: {
+      async prepareWorkspace() {
+        throw rejection;
+      },
+    },
+  });
+
+  await assert.rejects(
+    coordinator.prepare({ makaSessionId: 'session-secret' }),
+    (error: unknown) => error === rejection,
+  );
+  assert.deepEqual(await readdir(fixture.stagingParent), []);
+  assert.equal(rejection.message.includes('.env'), false);
+});
+
+test('failure cleanup refuses a staging root replaced by an unrelated directory', async () => {
+  const fixture = await createFixture();
+  let unrelatedFile = '';
+  const coordinator = createFileQuiescentSessionSnapshotCoordinator({
+    stagingParent: fixture.stagingParent,
+    privateStagingRootAuthority,
+    quiescence: immediateAuthority,
+    state: directoryStatePreparer,
+    workspace: {
+      async prepareWorkspace(input) {
+        const preparingRoot = dirname(input.destinationRoot);
+        await rename(preparingRoot, `${preparingRoot}.displaced`);
+        await mkdir(preparingRoot, { mode: 0o700 });
+        unrelatedFile = join(preparingRoot, 'unrelated.txt');
+        await writeFile(unrelatedFile, 'keep', 'utf8');
+        throw new Error('workspace preparation failed');
+      },
+    },
+  });
+
+  await assert.rejects(
+    coordinator.prepare({ makaSessionId: 'failure-cleanup-owner' }),
+    isSnapshotError('cleanup_failed'),
+  );
+  assert.equal(await readFile(unrelatedFile, 'utf8'), 'keep');
+});
+
+test('cleans a published snapshot when the authority fails while releasing quiescence', async () => {
+  const fixture = await createFixture();
+  const coordinator = createFileQuiescentSessionSnapshotCoordinator({
+    stagingParent: fixture.stagingParent,
+    privateStagingRootAuthority,
+    quiescence: {
+      async runQuiescent(_input, operation) {
+        await operation();
+        throw new Error('authority release failed');
+      },
+    },
+    state: directoryStatePreparer,
+    workspace: directoryWorkspacePreparer,
+  });
+
+  await assert.rejects(
+    coordinator.prepare({ makaSessionId: 'session-release-failure' }),
+    (error: unknown) =>
+      error instanceof SessionSnapshotError &&
+      error.code === 'io_failure' &&
+      error.message === 'Session snapshot preparation failed',
+  );
+  assert.deepEqual(await readdir(fixture.stagingParent), []);
+});
+
+test('cancellation and an expired deadline stop before staging begins', async () => {
+  const fixture = await createFixture();
+  let authorityCalls = 0;
+  const coordinator = createFileQuiescentSessionSnapshotCoordinator({
+    stagingParent: fixture.stagingParent,
+    privateStagingRootAuthority,
+    now: () => 10_000,
+    quiescence: {
+      async runQuiescent(_input, operation) {
+        authorityCalls += 1;
+        return operation();
+      },
+    },
+    state: directoryStatePreparer,
+    workspace: directoryWorkspacePreparer,
+  });
+  const controller = new AbortController();
+  controller.abort();
+
+  await assert.rejects(
+    coordinator.prepare({ makaSessionId: 'cancelled', signal: controller.signal }),
+    isSnapshotError('snapshot_cancelled'),
+  );
+  await assert.rejects(
+    coordinator.prepare({ makaSessionId: 'expired', deadlineAt: 10_000 }),
+    isSnapshotError('snapshot_cancelled'),
+  );
+  assert.equal(authorityCalls, 0);
+  assert.deepEqual(await readdir(fixture.stagingParent), []);
+});
+
+test('cancellation while waiting for quiescence is propagated as snapshot_cancelled', async () => {
+  const fixture = await createFixture();
+  const authorityEntered = deferred<void>();
+  const controller = new AbortController();
+  const coordinator = createFileQuiescentSessionSnapshotCoordinator({
+    stagingParent: fixture.stagingParent,
+    privateStagingRootAuthority,
+    quiescence: {
+      async runQuiescent(input) {
+        authorityEntered.resolve();
+        await waitForAbort(input.cancellation);
+        throw Object.assign(new Error('aborted'), { name: 'AbortError' });
+      },
+    },
+    state: directoryStatePreparer,
+    workspace: directoryWorkspacePreparer,
+  });
+
+  const preparation = coordinator.prepare({ makaSessionId: 'waiting', signal: controller.signal });
+  await authorityEntered.promise;
+  controller.abort();
+  await assert.rejects(preparation, isSnapshotError('snapshot_cancelled'));
+  assert.deepEqual(await readdir(fixture.stagingParent), []);
+});
+
+test('cancellation while leaving quiescence cleans the published snapshot', async () => {
+  const fixture = await createFixture();
+  const controller = new AbortController();
+  const coordinator = createFileQuiescentSessionSnapshotCoordinator({
+    stagingParent: fixture.stagingParent,
+    privateStagingRootAuthority,
+    quiescence: {
+      async runQuiescent(_input, operation) {
+        const result = await operation();
+        controller.abort();
+        return result;
+      },
+    },
+    state: directoryStatePreparer,
+    workspace: directoryWorkspacePreparer,
+  });
+
+  await assert.rejects(
+    coordinator.prepare({ makaSessionId: 'cancel-after-publish', signal: controller.signal }),
+    isSnapshotError('snapshot_cancelled'),
+  );
+  assert.deepEqual(await readdir(fixture.stagingParent), []);
+});
+
+test('release refuses a replacement directory and remains retryable for its owned root', async () => {
+  const fixture = await createFixture();
+  const handle = await createFileQuiescentSessionSnapshotCoordinator({
+    stagingParent: fixture.stagingParent,
+    privateStagingRootAuthority,
+    quiescence: immediateAuthority,
+    state: directoryStatePreparer,
+    workspace: directoryWorkspacePreparer,
+  }).prepare({ makaSessionId: 'release-owner' });
+  const publishedRoot = dirname(handle.snapshot.stateRoot);
+  const displacedRoot = `${publishedRoot}.displaced`;
+  await rename(publishedRoot, displacedRoot);
+  await mkdir(publishedRoot, { mode: 0o700 });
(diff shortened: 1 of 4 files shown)


likun666661:
I reviewed this from three angles: what problem this slice solves, why the approach can solve it, and whether the problem is minimally/correctly defined.

The core problem is real and correctly identified: `SessionBundleFileService.pack` must not read independently changing live state/workspace roots. A per-Session writer boundary, followed by a private staging copy of state and workspace at one logical point in time, is the right shape. If the quiescence authority governs every writer and the concrete state/workspace preparers honor their contracts, later live mutations cannot affect the prepared snapshot and the live Session can resume before compression/upload completes.

However, this PR currently implements a coordinator framework rather than solving #2369 end to end: there is no production quiescence authority, state exporter, workspace copier, call site, codec round trip, or real writer-race test yet. CI validates the self-contained contract/fakes, not the production guarantee. I think this should be described explicitly as a foundation slice.

I also see two blocking correctness/security gaps:

1. **The workspace policy is not actually enforced by the coordinator.** The coordinator passes `SESSION_SNAPSHOT_WORKSPACE_POLICY_V1` to an injected `prepareWorkspace()`, then only validates the returned directory and counters. A preparer can copy `.env`, `.ssh`, or `node_modules`, return valid-looking counters, and still receive a successful handle. Therefore the real enforcement boundary is the trusted concrete preparer, not this coordinator. Either the coordinator should own/verify traversal and copying, or the API/docs should state that the concrete preparer is the trusted enforcement point and PR 2 must provide/test it before this guarantee is claimed.

2. **The inode checks do not close the final recursive-delete race.** In `removeOwnedSnapshotDirectory`, the code verifies the cleanup path identity and then calls path-based `rm(cleanupRoot, { recursive: true })`. A same-UID process can rename the verified directory away and create an unrelated directory at that path between the last `lstat` and `rm`. Mode `0700` does not exclude another process running as the same principal. The current tests cover replacement before `release()`, but not replacement between the final identity check and deletion, so “cleanup cannot delete unrelated paths” is stronger than the implementation currently proves.

There is also a policy-definition inconsistency: `.log` is included in the sensitive extension regex and secret classification runs before log exclusion. Consequently `logs/agent.txt` is excluded, while `debug.log` rejects the entire snapshot; the test explicitly pins that behavior. #2369 says logs should be excluded. Likewise `.crt`, `.cer`, and `.csr` are often public material rather than known secrets. “May contain sensitive data” and “known user-authored secret” should not be treated as the same category.

From an Occam’s razor perspective, I would define the minimum problem as:

> Under one authority that covers all Session writers, export state and deterministically copy the portable workspace into a unique private staging root; publish it for packing and clean it up in `finally`.

Then keep these as separate concerns:
- rebuildable/runtime content: deterministic exclusion;
- a small set of known legacy secret files: fail-closed rejection;
- content-level secret discovery and long-term secret handling: Secret Store + runtime injection, outside this coordinator;
- crash/orphan cleanup and adversarial same-principal deletion: a separately specified lifecycle/security mechanism.

So the architectural direction is good, but I would not treat the issue as solved yet. The main adjustment is to align the claimed trust/security boundaries with what the code can actually enforce, and avoid combining consistency, portability, secret governance, and adversarial cleanup into one oversized problem definition.

Astro-Han:
Thanks for the work — the module is well-structured (contract + quiescence/lifetime abstraction + policy + deterministic fakes, matching the issue's approved two-PR split), the second commit correctly moved secret classification after the log exclusion and made `.crt/.cer/.csr/.pem/.der` include (fixing likun's point), and the coordination design (logical point-in-time + private staging copy, then release live writers) fits the problem. CI 12/12 green.

**Conclusion: FAIL — P1 blocks merge (the V1 policy is fail-open for the most common secret directory conventions), plus two P2s and P3s.**

**P1 — `secrets/`, `credentials/`, `keys/`, `private/` directories are fully included in snapshots.** `isKnownSecretPath` only runs a basename-pattern check (+ `.ssh` segment + specific files like `.docker/config.json`) — there is *no* directory-segment check for the semantically-clearest secret names. I reproduced this against a node port of `classify`: `secrets/token` → include, `secrets/api-key.txt` → include, `credentials/foo.txt` → include, `secrets` as a directory entry → include. Meanwhile `.ssh/config` is rejected specifically. The asymmetry is telling: `.ssh` is treated as secret while `secrets/`/`credentials/` — the most common user conventions — pass through wholesale, and the test table (lines 540-575) never covers a `secrets/` directory. This violates the PR's own contract ("rejecting only names that identify known user-authored secret material" — `secrets/` is exactly such a name) and #2369's acceptance bar (injected credentials excluded; user-authored secrets require an explicit decision). Prediction: a workspace using the most common `secrets/` layout leaks private content into the bundle, which is the very thing the cloud fork/archive path (#1415/#1286) must not do. Fix: add `secrets`, `credentials`, `keys`, `private` directory segments to `isKnownSecretPath` (symmetric with `.ssh`), and extend the test table with `['secrets/token','file',reject]`, `['credentials/foo.txt','file',reject]`, `['secrets','directory',reject]`.

**P2-1 — the mirror-image false positive: common non-secret files (`secrets.ts`, `.env.example`, `*.pub` public keys) reject the whole snapshot.** The `/^(?:credentials?|secrets?)(?:\..*)?$/i` pattern matches any extension (`secrets.ts`, `secrets.md`, `secrets.test.ts` are ordinary source/document names), `/^\.env(?:\..*)?$/i` matches `.env.example` (the standard committed template), and the id_ pattern matches `id_ed25519.pub`/`id_rsa.pub` — while the PR's own tests claim "public certificate encodings are not rejected by extension alone" (`certs/client.crt` include). A workspace with `src/secrets.ts` or `.env.example` (any ordinary TS project) gets whole-snapshot `policy_rejected` — fail-closed, so no leak, but the feature becomes unusable for common projects and the policy's precision is wrong in both directions. Restrict the secret patterns to common secret extensions (or allow `.ts/.js/.md/.example/.pub`).

**P2-2 — 1977 lines with no production callers, and the ownership/inode/cleanup machinery belongs to PR 2 by the issue's own split (and is unverifiable here).** #2369 explicitly splits PR 2 into "filesystem/state integration, cleanup, race tests, codec round-trip E2E" — this PR delivers PR 1 plus ~500 lines of filesystem integration (inode binding, owner record, cleanup protocol, Windows ACL) verified only against fakes. Also, the error-code union (222-232) declares `snapshot_busy`/`session_not_quiescent`/`source_changed`/`quota_exceeded` — none of which this file ever throws. Risk: if PR 2's real quiescence authority / copier doesn't match this contract shape, ~1200 lines need rework, and dead codes invite mis-handling downstream. Either explicitly mark the ownership machinery as a PR-2 pre-verification item and delete/annotate the unproduced error codes, or compress to "contract + policy + minimal coordinator".

**P2-3 — the headline isolation test is tautological; "later live mutations cannot change the snapshot" is not actually tested.** The fake preparer writes JS-variable values into staging files; the test then reassigns those *JS variables* — which cannot change the already-written files — and re-reads them. No live source is read after prepare, so the test would pass even if PR 2's coordinator exposed live paths. Make the fake copy from `fixture.liveStateRoot/liveWorkspaceRoot`, mutate the live files after prepare, then assert the snapshot is unchanged.

P3 (optional): the `.aws`/`.cargo` branch (666-667) is unreachable (basename `credentials` is already matched by the general pattern at 656) — delete; the catch at 374 is missing a `return`, widening the callback return to `OwnedPreparedSessionBundleHandle | undefined`; orphaned owner records make `release()` fail forever and strand the staging root (declared as separate-mechanism, but a reaper/metrics item is worth tracking); the inode→rm final-mile race is declared-and-accepted in the PR body — fine; "policy fixed by coordinator but not enforced by it" is documented as PR-2's trusted-preparer job — fine.

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagent on `opencode-go/deepseek-v4-flash`), which ported `classify`/`isKnownSecretPath` to node and ran the P1/P2-1 cases (reproduced), traced the test tables, and reviewed both commits and the prior reviews. P1 and P2-1 are reproduced behaviors, not predictions. Please weigh these findings with your own judgment.*

<details>

结论：FAIL（P1 阻塞合并）。模块结构好（契约+静默/生命周期抽象+策略+确定性 fake，符合 issue 批准的两 PR 拆分），第二 commit 已把 secret 判定移到 log 排除之后并让证书类改 include（修复 likun 点 3），协调器设计（逻辑时间点+私有 staging 副本后放行 live writers）贴合问题。P1：secrets/、credentials/、keys/、private/ 目录完全放行进快照——isKnownSecretPath 只做 basename 模式+.ssh 段+特定文件检查，无目录段语义检查；node 复刻 classify 实测 secrets/token→include、secrets/api-key.txt→include、credentials/foo.txt→include、目录项 secrets→include，而 .ssh/config 被拒——语义上更明确的 secrets/ 类目录整体放行是明显不对称，测试表（540-575 行）从没覆盖 secrets/ 目录；违反 PR 自身契约（"只拒绝已知用户自写机密材料的名称"——secrets/ 正是此类）与 #2369 验收（注入凭据必须排除、user-authored secret 需显式决策）。预测：最常见的 secrets/ 目录约定的 workspace 私密内容进 bundle 外泄（正是 cloud fork/归档要防的）。修复：isKnownSecretPath 增加 secrets/credentials/keys/private 目录段（与 .ssh 对称）+ 测试表补三个用例。P2-1（镜像误伤）：secrets.ts/.env.example/*.pub 公钥被拒导致整个快照 policy_rejected——/^(?:credentials?|secrets?).../ 匹配任意扩展名（secrets.ts/secrets.md 是普通源码/文档名）、.env.example 是标准提交模板、id_ 模式匹配 id_ed25519.pub 公钥，而 PR 自称"公钥证书编码不按扩展名拒"（certs/client.crt include）——普通 TS 项目快照整体被拒，功能对常见项目不可用。建议：机密模式限定常见机密扩展名或放行 .ts/.js/.md/.example/.pub。P2-2：1977 行全新增无生产调用点，ownership/inode/cleanup 机制按 issue 拆分属于 PR 2（且这里无法验证）；错误码联合类型声明的 snapshot_busy/session_not_quiescent/source_changed/quota_exceeded 本文件一个都没抛过——死契约面，PR 2 形状不匹配则 ~1200 行返工。建议显式标注为 PR 2 前置验证项或压缩到"契约+策略+极简协调器"。P2-3：头条隔离测试是重言式——fake preparer 写 JS 变量值进 staging 文件后重赋值 JS 变量不可能改变已写文件，无 live 源在 prepare 后被读，PR 2 真把 live 路径暴露给 pack 该测试也通过。修复：fake 从 fixture 根拷贝、prepare 后改 live 文件再断言不变。P3（可选）：.aws/.cargo 分支不可达（basename 已被通用模式命中）删之；catch 缺 return 加宽返回类型；孤儿 owner record 使 release() 永久失败（已声明另行机制，建议补 reaper）；inode→rm 末段窗口已声明接受。
</details>

Astro-Han (commented):
The foundation is structurally sound: state/workspace preparation and publication occur inside one trusted quiescence callback, publication is an atomic rename, and inode-bound ownership makes release idempotent, retryable, and fail-closed against replacement. I found no P0/P1 implementation issue in this PR1 scope.

The primary isolation test does not yet prove its stated property, as detailed inline. The simplest useful proof is to make preparers copy real live roots, mutate those roots after the boundary, and assert the published snapshot remains the boundary version. A later production-integration PR must additionally exercise a real writer during the boundary and add owner-record/inode-based orphan recovery; those are production-readiness gates, not reasons to add speculative machinery to this foundation.

Review performed with three Codex reviewer agents and DeepSeek V4 Flash as advisory tools; I verified the finding against the latest head and live green CI.

foundation 的结构是可靠的：state/workspace prepare 与 publish 位于同一 trusted quiescence callback，publish 使用原子 rename，inode-bound ownership 让 release 具备幂等、可重试和 replacement fail-closed。PR1 范围内未发现 P0/P1 实现问题。

主 isolation test 尚未证明其声明属性，具体见行内。最简单有效的证明方式是让 preparer 从真实 live roots 复制，在 boundary 后修改这些 roots，再断言 published snapshot 保持 boundary 版本。后续生产接入 PR 还必须让真实 writer 在 boundary 内尝试写入，并增加基于 owner record + inode 的 orphan recovery；这些是 production readiness gate，不需要在当前 foundation 中预设复杂实现。

本次审查使用了三位 Codex reviewer agents 与 DeepSeek V4 Flash 作为辅助工具；我已依据最新 head 和实时绿色 CI 复核问题。

</details>

packages/storage/src/__tests__/quiescent-session-snapshot.test.ts
```diff
@@ -0,0 +1,749 @@
...
+  assert.equal(
+    await readFile(join(handle.snapshot.stateRoot, 'runtime.sqlite'), 'utf8'),
+    liveState,
+  );
+  assert.equal(
+    await readFile(join(handle.snapshot.workspaceRoot, 'main.ts'), 'utf8'),
+    liveWorkspace,
+  );
+
+  liveState = 'later-state';
```
Astro-Han: **P2 — This does not exercise snapshot isolation.** The preparers already copied the current JavaScript string values directly into staging at lines 60/73; reassigning those variables cannot mutate any live source, so the assertions would pass even if production read from the wrong root or outside the intended boundary. Seed `fixture.liveStateRoot`/`liveWorkspaceRoot`, make the preparers copy from them, then mutate the live files after `prepare()` and assert the snapshot retains the boundary contents.

packages/storage/src/quiescent-session-snapshot.ts
```diff
@@ -0,0 +1,1227 @@
...
+// measurement rules introduced by #1353. Snapshot rejection is reserved for
+// names that identify known secret material; public certificate encodings and
+// other ambiguous formats are not rejected by extension alone.
+const KNOWN_SECRET_WORKSPACE_FILE_PATTERNS = [
+  /^\.env(?:\..*)?$/i,
+  /^\.(?:npmrc|netrc|pypirc|terraformrc)$/i,
+  /^\.git-credentials(?:\.lock)?$/i,
+  /^(?:credentials?|secrets?)(?:\..*)?$/i,
+  /(?:^|[-_.])(?:id_(?:rsa|dsa|ecdsa|ed25519)|private[-_.]?key)(?:$|[-_.])/i,
+  /\.(?:key|p12|pfx)$/i,
```
Astro-Han, line 113: [P1] Cover known private-key and credential paths that this allowlist currently exports. Common names such as `privkey.pem` and `private.pem` do not match this pattern or the extension list, and nested paths such as `credentials/oauth.json` or `private/token` are included because only the basename is checked. A Session bundle prepared under this policy can therefore contain user credentials despite the contract claiming fail-closed rejection of known secret material. Add exact high-confidence names/directories (without banning every `.pem`, since certificates are public) and focused include/reject cases.

packages/storage/src/quiescent-session-snapshot.ts
```diff
@@ -0,0 +1,1227 @@
...
+}
+
+const INCLUDE = Object.freeze({ kind: 'include' } as const);
+
+// This fail-closed rejection set is deliberately narrower than the workspace
+// measurement rules introduced by #1353. Snapshot rejection is reserved for
+// names that identify known secret material; public certificate encodings and
+// other ambiguous formats are not rejected by extension alone.
+const KNOWN_SECRET_WORKSPACE_FILE_PATTERNS = [
+  /^\.env(?:\..*)?$/i,
```
Astro-Han, line 107: [P2] Narrow the patterns so ordinary portable inputs are not rejected. `^\.env(?:\..*)?$` rejects `.env.example`/`.env.template`, and the `id_ed25519` pattern also rejects `id_ed25519.pub`; both are commonly intentional project inputs and the latter is explicitly public. Prefer exact private-key names plus explicit template/public exceptions, and add regression cases for these two paths.

packages/storage/src/quiescent-session-snapshot.ts
```diff
@@ -0,0 +1,1227 @@
...
+    typeof entry.relativePath !== 'string' ||
+    entry.relativePath.length === 0 ||
+    entry.relativePath.includes('\\') ||
+    entry.relativePath.includes('\0') ||
+    entry.relativePath.startsWith('/') ||
+    entry.relativePath.endsWith('/')
+  ) {
+    return { kind: 'reject', category: 'unsafe_path' };
+  }
+  const segments = entry.relativePath.split('/');
```
Astro-Han, line 855: [P2] Validate the same portable segment contract as the V1 bundle codec here. This decoder currently accepts paths such as `workspace/CON`, `foo:bar`, and `name.`, but `session-bundle-contract.ts` deliberately rejects Windows device names, reserved characters, and trailing dots/spaces on every host. The preparer can therefore successfully publish a snapshot that the next `pack` stage must reject. Reuse one shared path predicate at the earlier policy boundary and test the codec's reserved-name matrix here.

packages/storage/src/__tests__/quiescent-session-snapshot.test.ts
Astro-Han: [P2] This does not exercise source isolation or a quiescence boundary: it only reassigns two JavaScript strings after the fake preparers have already written copies, while `liveStateRoot` and `liveWorkspaceRoot` are never read or mutated. The test would still pass if a future adapter accidentally read live files after quiescence. Either make this an honest coordinator-ordering test and remove the isolation claim, or mutate real source roots around a controlled release point and assert the staged bytes stay pinned.

Astro-Han (commented):
Defining the quiescent boundary before adding production adapters is a sensible split, and pinning one policy object avoids a caller-controlled safety downgrade. The contract is not yet internally closed, though: the secret-name policy has both unsafe gaps and common false positives, and its accepted path language is wider than the existing V1 bundle codec. The simplest solution is to reuse one canonical portable-path predicate and express secret handling as a small reviewed set of exact path/name rules rather than increasingly broad filename regexes.

AI-assisted review disclosure: Codex verified these findings against the current head, the V1 bundle path contract, policy classifier, ownership/cleanup flow, and tests. Two independent reviewer-agent passes and an OpenCode Go DeepSeek V4 Flash (high) adversarial pass were used as inputs. No local tests were run.

<details>
<summary>中文复核</summary>

先定义 quiescent boundary、后接生产 adapter 的拆分是合理的，固定单一 policy 也避免调用方降低安全标准。但当前契约还没有闭合：secret 规则既漏掉常见私钥/凭据路径，也误伤常见公开或模板文件；同时它接受的路径集合比现有 V1 bundle codec 更宽。更简单的方案是复用唯一的 portable-path 判定，并用小而明确的路径/文件名规则表达 secret policy，避免继续扩张宽泛正则。

本次为 AI 辅助审查：Codex 在最新 head 上核验 V1 bundle 路径契约、policy classifier、ownership/cleanup 与测试；另使用两次独立 reviewer 及一次 OpenCode Go DeepSeek V4 Flash（high）对抗审查。未运行本地测试。
</details>

packages/storage/src/quiescent-session-snapshot.ts
```diff
@@ -0,0 +1,1227 @@
...
+] as const;
+
+/**
+ * V1 portable-workspace policy. The coordinator pins this exact policy, while
+ * the trusted filesystem preparer is its enforcement point: the coordinator
+ * does not re-traverse or attest the prepared destination. The preparer remains
+ * responsible for applying every decision and rejecting symlinks, hard links,
+ * special files, path races, case conflicts, and quota violations.
+ */
+export const SESSION_SNAPSHOT_WORKSPACE_POLICY_V1: SessionSnapshotWorkspacePolicy = Object.freeze({
```
Astro-Han, line 123: The checks on this head are from 2026-08-18 and can't speak for the current tree.

[#3397](https://github.com/apache/maka/pull/3397) landed on 2026-08-22 and added ASF license headers across ~2685 files. Any run that started before that predates the gate — a green from 08-18 is a true result about a different repository.

There's also an `e2e` failure on this head. I attributed it: the failing case is `slash-command-menu-compacts-the-active-session`, a `locator.click` timeout in the desktop suite, 1 failed of 42. This PR adds three files under `packages/storage` and touches nothing in desktop, CLI, or the slash-command path, so the failure isn't yours. But a five-day-old failure that happens to be unrelated still isn't current evidence of anything.

A rebase onto current `main` and a fresh run would settle both at once.

Astro-Han (commented):
Two reviewers went over this independently. Neither found a correctness defect, and I checked the disagreement-prone parts myself. This is careful work — the staging tri-state, the dev/ino owner binding before cleanup, the atomic publish rename, and the fail-closed release path all hold together, and the error codes are bounded rather than open-ended.

I want to flag one thing that is a decision rather than a defect, and one piece of housekeeping.

**The merge premise.** There is no production consumer today — `quiescent-session-snapshot` is referenced only by its own module and its tests. That is exactly what a contract slice looks like, and I'm not treating it as a defect, because #2369 names the follow-up concretely. But it does mean merging this is a bet that PR 2 lands. If it doesn't, 1227 lines of implementation become pure carrying cost, and a `publish` protocol change in PR 2 would rewrite much of it rather than build on it. Worth stating in the PR body so the premise is on the record rather than implied.

I looked hard at whether the policy pin is speculative generality and concluded it isn't: `SESSION_SNAPSHOT_WORKSPACE_POLICY_V1` being non-replaceable protects a real invariant. A snapshot carries a workspace copy, so a caller that could weaken the policy could route around the secret exclusions. `'policy' in options → throw` is the enforcement point, and that's a security boundary rather than a hypothetical extension seam.

One reviewer raised whether the coordinator could mistake a non-quiescent session for a quiescent one. It can't on its own account — and the interface doc already says so plainly ("this interface does not make a process-local mutex authoritative by itself; every real writer must already be governed by the supplied Host/Owner implementation"). Correctness is delegated, and the delegation is explicit. Noting it only because it's the load-bearing assumption for whoever implements the Authority in PR 2.

**CI needs a fresh run.** Details inline. Not approving on this head for that reason alone — I have no blocking findings on the code.

packages/storage/src/quiescent-session-snapshot.ts
```diff
@@ -0,0 +1,1227 @@
...
+      return { kind: 'exclude', category: 'dependency_tree' };
+    }
+    if (
+      lowerSegments.includes('.cache') ||
+      lowerSegments.includes('.maka-cache') ||
+      lowerSegments.includes('.turbo')
+    ) {
+      return { kind: 'exclude', category: 'cache' };
+    }
+    if (
```
Astro-Han, line 155: [P1] Reject secret-shaped files before the generic log exclusion. In this order, `.env.log`, `secrets.log`, `credentials.log`, and `keys/private-key.log` all return `{ kind: 'exclude', category: 'log' }`, so snapshot creation succeeds and silently omits them; the added test even pins `secrets.log` to that outcome. That conflicts with #2369's maintainer decision that known user-authored secret files must fail the complete snapshot rather than silently change workspace semantics, and the bounded counters cannot tell the user which important file disappeared. Please make the known-secret rule win on overlaps (and cover at least `.env.log`/`secrets.log`) while retaining ordinary `debug.log` as an exclusion.

Astro-Han (commented):
Reviewed at exact head `0482ee65047bc4fb177b4ef415f46558c201faed`. Verdict: NO-GO, 1×P1. The two-PR split itself is not a simplification finding: #2369 explicitly authorizes this contract/fake slice before the production adapters, and this coordinator is an orthogonal staging/lifetime layer rather than a second archive codec or state exporter.

CI attribution: the sole red check is the Desktop slash-command E2E timing out while clicking `/compact` because the root `<html>` intercepted pointer events and the option then detached. This PR adds an otherwise unconsumed Storage module plus its export/tests and does not touch that Desktop path; all affected build, typecheck, workspace, Runtime Host, package, and Storybook lanes passed. I therefore treat that red as unrelated/intermittent, not as a P-level PR regression. It still means the exact-head all-green approval gate is not met.

Local exact-head verification: dependency-order Core→Storage and full Desktop dependency/build chain passed; focused snapshot tests passed 17/17 with 1 platform skip; Biome and diff-check passed. Local Electron replay was unavailable because this runner lacks `xvfb-run`.

Astro-Han (commented):
Gives `pack` a state/workspace pair from one point in time. Contract gap is real and the shape is right — the quiescence authority is injected, not implemented here, which is correct: `packages/storage` depends only on `@maka/core` and can't reach `runtime-host`.

**P1 — `reject` is fail-closed on a rule that can't tell its own false positives apart.** `isKnownSecretEntry` (`:688`) matches `credentials` / `private` / `secrets` against **every path segment**, and the policy is pinned so callers can't narrow it (`:332`). Rails' `config/credentials/` and C++ `include/private/` become permanently unsnapshottable with no override.

Everything else in `classify` is `exclude`, which already guarantees content never reaches the snapshot. `reject` only buys "I may have misidentified this, stop and let a human look" — worth it at a low false-positive rate, a denial of service otherwise. That rate is unmeasurable until PR 2 brings a real corpus, so the rule is currently calibrated on name intuition. Suggest `exclude` + a diagnostic counter, with `reject` reserved for names that can't be anything else (`id_ed25519`).

**P2 — orphaned staging is unrecoverable, and the mechanism you deferred is already in this package.** `:493` mints a fresh UUID per run and `assertMissing` only checks the current one, so a crash before publish leaves 0700 user source nothing will revisit. You defer this for a sound reason — prefix-based cleanup can't authenticate what it deletes. But `session-copy-cleanup.ts` exports `SessionCopyCleanupAuthority`: persisted leases, owner identity via `ProcessLifetimeOwner`, injected `isOwnerProcessActive`, and `recover()`. It authenticates by owner identity instead of scanning by prefix — the property you wanted. `OwnedSnapshotStaging` reimplements its create-and-own half and stops before recovery.

P2 not P1 because it's documented and latent (no exports entry, no production caller), but it needs an answer before PR 2 adds a call site — and the answer looks like reuse, not a new platform mechanism.

**P3.** `cleanupAfterPreparationFailure` keeps the primary error in an `AggregateError` but replaces its **code**, so callers switching on the code lose the classification. `:1243` treats `signal.aborted || isAbortError` as cancellation, so a real I/O failure racing an abort reports as `snapshot_cancelled`.

---

*AI-assisted review. A Claude Code subagent did design and correctness passes; I verified the graded findings and **withdrew two of its conclusions** — `runExclusiveSessionOperation` isn't reusable here (checked package dependency direction), and cleanup doesn't swallow the primary error (it aggregates it). I read `classify` / `isKnownSecretEntry` to confirm `reject` is fail-closed, and `session-copy-cleanup.ts`'s exported surface for the recovery semantics. No tests or typecheck run.*

**P1**:known-secret 按路径每段匹配且 fail-closed,策略又被 pin 死,Rails `config/credentials/`、C++ `include/private/` 会让整个 workspace 永久无法快照。`classify` 里其余全是 `exclude`,而 `exclude` 已足以保证内容不进快照——`reject` 只多买到"可能识错了,停下来",而误报率要等 PR 2 语料才能测。建议 `exclude` + 诊断计数。

**P2**:崩溃后 staging 孤儿无法回收。你推迟的理由(prefix 扫描无法认证)是对的,但同包 `session-copy-cleanup.ts` 的 `SessionCopyCleanupAuthority` 已用持久 lease + `ProcessLifetimeOwner` + `recover()` 解决,不靠 prefix。

</details>

Astro-Han (approved):
Reviewed at exact head `ae821f53ec6e8ead2db1bfee387c94b4a405a5c0`. Approved.

The snapshot boundary makes sense. State and workspace are copied under the same quiescence boundary, and packing reads from private staging roots instead of live files.

One thing I would simplify before the production integration. #2369 asks V1 to decide what to do with arbitrary user-authored secrets in the workspace. Doing that reliably is a much larger feature. Either detection has to be good enough to trust, or confirmation needs a real authorization and revision-binding flow. The current confirmation grant API commits us to that design before the product behavior is settled.

I would keep V1 narrower: exclude paths we know about, such as platform/runtime credentials, logs, caches, `.git`, and `node_modules`; state plainly that V1 does not detect arbitrary user-authored secrets; and remove the confirmation grant/authority path. If Bundle needs that stronger guarantee later, it deserves its own end-to-end design. This does not block this foundation PR.

AI-assisted review: Codex checked the exact-head implementation, #2369, the current review history, and live CI. I reviewed the conclusion and approve this head.

<details>
<summary>简体中文</summary>

已审查当前 head `ae821f53ec6e8ead2db1bfee387c94b4a405a5c0`，同意合并。

快照边界是合理的。state 和 workspace 在同一个 quiescence boundary 内复制，后续打包读取私有 staging 副本，不再读取仍在变化的 live files。

有一处我建议在生产接入前简化。#2369 要求 V1 决定如何处理 workspace 中用户自己放入的 Secret。要可靠做到这一点，其实是一个大得多的功能：要么检测足够可靠，要么确认机制必须有完整的授权和 workspace revision 绑定。当前 confirmation grant API 在产品行为尚未确定前，就先固定了这套设计。

我更建议收窄 V1：只排除确定知道的路径，比如平台或运行时凭据、日志、缓存、`.git` 和 `node_modules`；明确说明 V1 不检测任意用户文件中的 Secret；删除 confirmation grant/authority。如果以后确定 Bundle 必须提供更强保证，再单独做完整的端到端设计。这个建议不阻塞当前 foundation PR。

本次为 AI 辅助审查：Codex 检查了当前 head 的实现、#2369、现有 review 记录和实时 CI；我复核了结论并批准当前 head。

</details>

Issue #2913, opened 2026-08-12
perf(desktop): make session data flow incremental and bounded


M4n5ter:
<details open>
<summary>English</summary>

## Problem

Desktop session work currently scales with total history instead of the amount of change:

- Transcript hydration performs one sequential request per small message. A synthetic tool-rich 1,000-turn session requires about 5,000 requests, giving a 150 s transport lower bound at 30 ms RTT.
- Runtime Host materializes an in-memory transcript snapshot capped at 16 MiB, so larger sessions cannot be opened.
- Live frames are buffered while the full transcript loads; sufficiently active long sessions can repeatedly hit the pending-frame limit and restart hydration.
- Desktop reloads the complete transcript after durable message/tool/terminal changes.
- Progressive transcript mounting eventually inserts the complete history into the DOM.
- Every session catalog invalidation refreshes the full catalog and replaces unchanged row objects.
- Fine-grained assistant streaming repeatedly scans the accumulated text and approaches O(n²).
- Startup waits for the complete onboarding/session snapshot before mounting React.
- Thread search can read up to 200 complete transcripts sequentially across the Runtime Host connection.

The storage layer already has a durable append-only `(session_id, sequence)` primary key for session messages. We can use that sequence directly as the transcript watermark; no historical data migration or new transcript revision is required.

## Desired outcome

Make Desktop a bounded incremental client:

```text
SQLite durable message ledger
        │ bounded pages + durable watermark
        ▼
Runtime Host subscription
        ├─ full materialization helper ──> CLI / TUI
        └─ tail bootstrap + advancement ─> Desktop Main sparse replica
                                             │
                                             ▼
                                    incremental bounded IPC
                                             │
                                             ▼
                                  virtualized Renderer transcript
```

Non-negotiable properties:

- Work per update is proportional to changed bytes/items, not total history.
- The latest useful content does not wait for full transcript or catalog hydration.
- Renderer state and DOM remain bounded independently of history size.
- Durable message sequence, continuity projection revision, subscription frame sequence, and Host epoch remain distinct concepts.
- Gaps, resets, reconnects, and late responses are explicit and fail closed.
- Sensitive content is sanitized before entering Renderer state.
- Remote Runtime Host RTT is treated as a supported environment.

### 1. Replace the transcript snapshot protocol — completed in #2922

- [x] Add storage APIs for byte-bounded forward/backward pages of `{sequence, message}`.
- [x] Read the durable high-water mark and tail page in one SQLite read transaction.
[...]

PR #2975, opened 2026-08-13
perf(ui): virtualize long chat transcripts

M4n5ter:
<details open>
<summary>English</summary>

## Summary

Long conversations previously progressed toward mounting every loaded turn, so DOM and layout work kept growing with the resident transcript. Active text and thinking events could also publish multiple renderer updates within one frame.

This change keeps Chat on a bidirectional, variable-height turn window and publishes active text/thinking updates at most once per animation frame. A session starts from a two-turn tail, then expands from measured geometry to cover the viewport plus pixel overscan; the mounted window has a hard limit of 100 turns. Upward scrolling loads earlier history automatically, and one idle 512 KiB prefetch warms history when the initial viewport is already near the start.

Prompt navigation, application search, scroll anchoring, and native selection inside the mounted window remain available. Native browser Find and Select All only cover mounted content; application search continues to cover the durable transcript. If focus or a selection endpoint would cross the hard window boundary, the interaction is handed back to the transcript before eviction. Turn-local presentation state, such as an expanded disclosure, resets after that turn is unmounted and later remounted; message data is unaffected.

The bound is turn-based: one turn remains an indivisible rendering unit and may contain a large timeline. Per-turn timeline bounding remains part of the broader #2913 work rather than introducing nested virtualization here.

This PR does not optimize streaming redaction. It retains the existing whole-accumulated-text security boundary; differential-safe incremental redaction is tracked separately in #2990.

Refs #2913

## Verification

- Repository lint, format, typecheck, production build, and the affected Desktop/UI tests pass.
- Electron coverage loads more than 100 resident turns and verifies the measured layout gap, bounded non-empty mount window, automatic history loading, and hard-limit focus/selection handoff.
- Real Electron scroll traces cover the cold first upward load and repeated history loads without an observed long task over 50 ms.
- A real macOS VoiceOver pass, with focus inside the transcript, did not re-announce historical turns during virtual-window replacement or history loading.

## Checklist

- [ ] No

</details>

<details>
<summary>中文</summary>

## 概要

长会话此前会逐步挂载所有已加载 turn，使 DOM 与布局开销持续随 resident transcript 增长；活跃文本和 thinking 事件也可能在同一帧内多次发布 renderer 更新。

本变更将 Chat 改为双向可变高度 turn 窗口，并将活跃文本/thinking 更新限制为每个动画帧最多发布一次。会话从尾部两个 turn 开始，根据实测几何扩展到覆盖视口与像素 overscan，挂载窗口硬上限为 100 个 turn。向上滚动会自动加载更早历史；如果初始视口已经接近历史起点，还会在空闲时预取一次 512 KiB 历史数据。

Prompt 导航、应用内搜索、滚动锚定以及 mounted window 内的原生文本选择保持可用。浏览器原生查找与全选只覆盖已挂载内容，应用内搜索仍覆盖完整 durable transcript。如果焦点或选择端点即将越过硬窗口边界，会先把交互权移交给 transcript 再卸载节点。Turn 被卸载后再次挂载时，展开 disclosure 等 turn-local 展示状态会重置；消息数据不受影响。

该上限以 turn 为单位：单个 turn 仍是不可拆分的渲染单元，可能包含较大的 timeline。单 turn 内部的 timeline 有界化继续由整体 #2913 跟踪，本 PR 不引入嵌套虚拟化。

本 PR 不优化流式 redaction，继续保留已有的“对完整累计文本脱敏”安全边界；可差分验证的增量 redaction 由 #2990 单独跟踪。

关联 #2913

## 验证

- 仓库 lint、format、typecheck、生产构建以及受影响的 Desktop/UI 测试通过。
- Electron 覆盖会加载超过 100 个 resident turn，并验证实测布局间距、非空且有界的挂载窗口、自动历史加载以及硬上限处的焦点/选择交接。
- 真实 Electron 滚动 trace 覆盖首次向上冷加载与连续历史加载，未观察到超过 50 ms 的 long task。
- 在 transcript 内持有焦点进行真实 macOS VoiceOver 验证时，虚拟窗口替换与历史加载均未重新播报旧 turn。

## 检查清单

- [x] 测试覆盖本次变更，并会在缺少修复时失败
- [x] lint、format、typecheck 和受影响测试套件已在本地通过

本 PR 是否包含行为变化？

- [x] 是——已在概要中说明
- [ ] 否

</details>

AI disclosure / AI 披露: This PR was implemented by Codex under M4n5ter’s direction and review. / 本 PR 由 Codex 在 M4n5ter 的指导与审核下完成。


Astro-Han (commented):
Nice work — the virtualizer is the right shape and I'd like it to land. My one request is that it land without the redaction change.

This PR carries three independent intents: the turn virtualizer (21 files), the incremental display redaction (6 files), and the per-frame publish batching (3 files, which touches no redaction code). The batching is uncontroversial and the virtualizer is close. Everything blocking is in the redaction change, and I don't think it can be settled inside a perf PR.

## P0 — streaming redaction can leak the tail of a secret

Reproduced on `6d42e6383` through the real `applyAssistantDelta` path with `redactionState` threaded per delta:

```
in:     api_key=wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY tail
whole:  api_key: <redacted> tail
stream: api_key= <redacted>/K7MDENG/bPxRfiCYEXAMPLEKEY tail
```

A real JWT behaves the same way, leaking payload and signature after the first `.`. This reproduces at chunk sizes 1 and 4, so ordinary token-sized deltas hit it.

The chain, as far as I could trace it: `pendingApiKeyContext` (`incremental-display-redaction.ts:139`) normalizes the separator to `:` while the displayed text keeps `=`; then `stateForText('api_key:a')` returns undefined as soon as the value's first character arrives, so the state is dropped exactly when it starts to matter; later deltas fall into `trimRedactedContinuation` (`:83`), whose api-key context regex requires a colon, so it falls back to the default terminator `[^A-Za-z0-9_-]` and stops at `/`, `.` or `=` — while the pattern itself consumes `[^\s"'<>]+`.

In fairness: `applyAssistantComplete` re-runs whole-text `redactSecrets`, so this self-heals when `text_complete` arrives. But the exposure lasts for the whole streaming step, an aborted or errored turn never gets that event, and `chat-turn.tsx:247` copies with `{ redact: false }` — so the displayed text is what reaches the clipboard. It also puts raw secret material into live projection state, which is the invariant `assistant-stream.ts:104` documents.

Worth noting that this layer is load-bearing here: `ai-sdk-backend.ts:2232` emits assistant `text_delta` verbatim and imports no redaction. The backend redactor covers tool output, stderr/stdout and command lines, not model prose — so for assistant text and thinking this is the only pass, not a second one.

## P1 — the same mechanism also eats legitimate text

`commit deadbeef…deadbeefpath/to/thing` survives whole-text redaction untouched but streams as `commit <redacted>/to/thing`: the 40-hex prefix matches `long opaque token`, the continuation logic then eats every following `[A-Za-z0-9_-]`, while whole-text never matched because the trailing `\b` fails. Same root cause, same owner, one fix.

## Why split rather than fix in place

Whole-text `redactSecrets` has a decidable invariant: output equals `redactSecrets(full text)`. The incremental version gives that up — it scans a 512-char window and infers where a secret ends from the previous output ending in `<redacted>`. Once that invariant is gone, nothing can tell you the implementation is correct. A differential run over random text × chunk sizes {1,3,7,20,64} disagreed with whole-text redaction in 8631 of 36000 combinations.

Your new tests are genuine red→green — I checked the first commit's module against them. The differential case just happens to cover `x-api-key: `, which is the one shape that works.

If the 166–171 ms number is a real problem, this looks tractable: splitting on line or paragraph boundaries would do it ("a secret does not span a newline" is much stronger than "a secret fits in 512 chars"), as would caching the redaction of a settled prefix while still deriving from the full text. Both stay differential-testable. That deserves its own PR and issue, where the threat model can be stated explicitly.

Concretely: drop `incremental-display-redaction.ts` and its test, restore `assistant-stream.ts`, `thinking-stream.ts` and `live-turn-projection.ts` to `main`, and remove the three redaction cases from `live-turn-projection.test.ts`. The other 24 files stay.

## On the virtualizer — non-blocking

**P2, layout gap.** `DEFAULT_TURN_GAP = 4` (`turn-virtualizer.ts:24`) and `use-turn-virtualizer.ts:46` never passes a gap. `chat-view.tsx` does pass `gap={4}` to `ChatMessageList`, but Astryx's `gap` is a spacing-scale index, not pixels: `gapStyles["4"]` → `x18g69wz` → `var(--spacing-4)` → **16px**. `chat-message.css:38` records the same 16px independently. Scroll behaviour holds up — the `-gap` in `beforeHeight` cancels the extra DOM gap and anchor restoration measures real DOM — so what's wrong is document height, off by ~12px per unmounted turn (~4% on a 300-turn session), i.e. the scrollbar lies. `buildTurnVirtualLayout` already accepts `options.gap`; one source of truth for it would settle it.

**P2, the E2E can pass without virtualization.** `prompt-rail.spec.ts:216-221` counts `[data-virtual-turn-id]` and only asserts `<= 100`. Remove the wrapper attribute, render all 120 turns, and the count is 0 — both assertions still pass. A non-zero lower bound plus an assertion that the fixture exceeds the cap would fix it.

**P2, select-all now stops at the window.** On `main` the progressive mount ends with every turn in the DOM, so a selection can span the transcript; now at most 100 turns are mounted and select-all yields the window plus two spacers. "Native selection preserved" holds inside the window. In-app search plus `scrollTargetTurn` is a reasonable answer, but it's a product call worth stating in the PR body.

**P3, minor.** A selection spanning more than 100 turns still loses an endpoint when the window re-centers (inherent; worth a comment). `conversationItems` — the plan card — moved inside `.maka-turn-virtual-item`, so its gap to its own turn went 16px → 4px while the next turn stays at 16px. Turn-local state (expanded tool rows, reasoning panels) is now lost on unmount, which `key={turn.turnId}` makes easy to miss. `applyAssistantDelta` runs the incremental parser twice on the same state, and `thinking-stream.ts:136` doesn't recompute state after a total-cap drop the way the assistant path does — both moot if the redaction change comes out.

---

*Review assistance: Claude Code (Opus 5) read the diff, bundled the redaction modules and differential-tested them against whole-text `redactSecrets`, and traced the gap through the Astryx stylesheet; Codex and Kimi K3 each ran an independent adversarial pass, which corrected two of my own findings. I verified the reproductions, the line references, the `gap={4}` → 16px chain, the `applyAssistantComplete` self-heal and the E2E assertion myself. Per AGENTS.md this is not independent human review — this still needs one before merge.*

<details>
<summary>中文</summary>

虚拟化这条路走对了，我希望它能合进去。我唯一的请求是：让它在不带 redaction 改动的情况下合入。

这个 PR 里其实有三件互相独立的事：turn 虚拟化（21 个文件）、增量显示脱敏（6 个文件）、按帧发布的批处理（3 个文件，完全不碰 redaction 代码）。批处理没有争议，虚拟化也很接近可以合了。所有阻塞项都在 redaction 这部分，而我认为它没法在一个 perf PR 里解决。

## P0 —— 流式脱敏会漏掉密钥的尾部

在 `6d42e6383` 上，通过真实的 `applyAssistantDelta` 路径（逐 delta 传递 `redactionState`）复现：

```
in:     api_key=wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY tail
whole:  api_key: <redacted> tail
stream: api_key= <redacted>/K7MDENG/bPxRfiCYEXAMPLEKEY tail
```

真实的 JWT 行为相同，会在第一个 `.` 之后漏出 payload 和 signature。chunk size 为 1 和 4 时都能复现，也就是说正常 token 大小的 delta 就会触发。

我追到的链条是：`pendingApiKeyContext`（`incremental-display-redaction.ts:139`）把分隔符归一化成 `:`，而显示出来的文本保留的是 `=`；接着 `stateForText('api_key:a')` 在 value 的第一个字符到达时就返回 undefined，于是 state 恰好在开始起作用的那一刻被丢弃；后续 delta 落进 `trimRedactedContinuation`（`:83`），它的 api-key context 正则要求冒号，匹配不上，就退回默认终止符 `[^A-Za-z0-9_-]`，在 `/`、`.`、`=` 处停下——而 pattern 本身吃的是 `[^\s"'<>]+`。

公平地说：`applyAssistantComplete` 会重跑全量 `redactSecrets`，所以 `text_complete` 到达时这个问题会自愈。但暴露会持续整个流式过程，中止或出错的 turn 根本等不到那个事件，而且 `chat-turn.tsx:247` 的复制走的是 `{ redact: false }`——显示出来的文本就是进剪贴板的文本。它同时把原始密钥写进了 live projection state，而这正是 `assistant-stream.ts:104` 写明的不变量。

另外值得一提：这一层在这里是承重的。`ai-sdk-backend.ts:2232` 把助手的 `text_delta` 原样发出，该文件没有引入任何 redaction。后端的 redactor 覆盖的是工具输出、stderr/stdout 和命令行，不包括模型正文——所以对助手文本和 thinking 来说，这是唯一一道，不是第二道。

## P1 —— 同一机制也会吃掉正常内容

`commit deadbeef…deadbeefpath/to/thing` 在全量脱敏下原样保留，流式却变成 `commit <redacted>/to/thing`：40 位 hex 前缀先命中 `long opaque token`，随后续接逻辑把后面每一个 `[A-Za-z0-9_-]` 都吃掉，而全量因为末尾的 `\b` 失败根本没匹配。同一根因、同一 owner，一处修复即可。

## 为什么建议拆分而不是就地修

全量 `redactSecrets` 有一个可判定的不变量：输出等于 `redactSecrets(完整文本)`。增量版本放弃了它——只扫 512 字符窗口，再从"上一次输出以 `<redacted>` 结尾"去推断密钥在哪结束。这个不变量一旦没有了，就没有任何东西能告诉你实现是对的。我用随机文本 × chunk size {1,3,7,20,64} 做差分，36000 组里有 8631 组与全量脱敏不一致。

你新加的测试确实是 red→green——我把第一个 commit 的模块拿出来对着它们跑过。只是那个差分用例恰好覆盖的是 `x-api-key: `，而这正是唯一能跑通的形态。

如果 166–171 ms 是真问题，我觉得是可解的：按行或段落边界切分就可以（"密钥不跨换行"这个假设比"密钥不超过 512 字符"强得多），或者缓存已稳定前缀的脱敏结果、但仍然从完整文本推导。这两种都能保持可差分测试。这值得单开一个 PR 和 issue，也正好可以在那里把威胁模型讲清楚——现在它是隐含的。

具体来说：删掉 `incremental-display-redaction.ts` 及其测试，把 `assistant-stream.ts`、`thinking-stream.ts`、`live-turn-projection.ts` 恢复到 `main`，并移除 `live-turn-projection.test.ts` 里那三条 redaction 用例。其余 24 个文件保持不变。

## 关于虚拟化本身 —— 不阻塞

**P2，布局 gap。** `DEFAULT_TURN_GAP = 4`（`turn-virtualizer.ts:24`），而 `use-turn-virtualizer.ts:46` 从不传 gap。`chat-view.tsx` 确实给 `ChatMessageList` 传了 `gap={4}`，但 Astryx 的 `gap` 是间距刻度序号而非像素：`gapStyles["4"]` → `x18g69wz` → `var(--spacing-4)` → **16px**。`chat-message.css:38` 也独立记录了同样的 16px。滚动行为其实是站得住的——`beforeHeight` 里的 `-gap` 恰好抵消了多出来的那个 DOM gap，锚点恢复又是按真实 DOM 测量的——真正错的是文档总高度，每个未挂载 turn 少算约 12px（300 turn 的会话约 4%），也就是滚动条在说谎。`buildTurnVirtualLayout` 已经接受 `options.gap`，给它一个单一事实源就能解决。

**P2，这个 E2E 在虚拟化被删掉时也会通过。** `prompt-rail.spec.ts:216-221` 统计 `[data-virtual-turn-id]`，只断言 `<= 100`。把 wrapper 属性去掉、120 个 turn 全量渲染，count 就是 0——两个断言照样通过。补一个非零下界，再断言 fixture 确实超过上限就好。

**P2，全选现在止步于窗口。** 在 `main` 上，渐进挂载的终态是所有 turn 都在 DOM 里，所以选择可以跨越整个 transcript；现在最多挂载 100 个 turn，全选拿到的是窗口内容加两个 spacer。"native selection preserved" 在窗口内成立。应用内搜索配合 `scrollTargetTurn` 是个合理的替代，但这是产品决策，值得在 PR 正文里写明。

**P3，小问题。** 跨度超过 100 个 turn 的选择在窗口重新居中时仍会丢掉一个端点（虚拟化的固有代价，加个注释即可）。`conversationItems`——也就是 plan 卡片——被移进了 `.maka-turn-virtual-item`，它与所属 turn 的间距从 16px 变成 4px，而与下一个 turn 仍是 16px。turn 内的局部状态（展开的工具行、reasoning 面板）现在会在卸载时丢失，而 `key={turn.turnId}` 很容易让人以为它被保留了。`applyAssistantDelta` 对同一个 state 跑了两次增量 parser，`thinking-stream.ts:136` 在总量截断后也没有像 assistant 路径那样重算 state——如果 redaction 改动拆出去，这两条就都不成立了。

---

*评审协助：Claude Code（Opus 5）阅读了 diff，把 redaction 模块打包后与全量 `redactSecrets` 做了差分测试，并顺着 Astryx 样式表追出了 gap；Codex 和 Kimi K3 各自做了一轮独立的对抗性复审，纠正了我自己的两条判断。复现结果、行号引用、`gap={4}` → 16px 这条链、`applyAssistantComplete` 的自愈行为以及 E2E 断言，都是我本人核验的。按 AGENTS.md，这不算独立人工评审——合并前仍然需要一次。*

</details>

packages/ui/src/turn-virtualizer.ts
```diff
@@ -0,0 +1,293 @@
...
+  ({ start, end } = expandRange(start, end, preferred, length));
+
+  if (end - start > hardMax) {
+    const requiredStart = Math.min(visibleStart, retained?.start ?? visibleStart);
+    const requiredEnd = Math.max(visibleEnd, retained?.end ?? visibleEnd);
+    const requiredSize = requiredEnd - requiredStart;
+    const size = Math.min(length, hardMax);
+    const center = requiredSize <= hardMax
+      ? (requiredStart + requiredEnd) / 2
+      : (visibleStart + visibleEnd) / 2;
```
MicroGery, line 117: [P2] Hand off active focus/selection before evicting the retained range

When `requiredSize > hardMax`, this branch deliberately recenters on the viewport and drops `retained`. Because `retainRange` is derived from `document.activeElement` and the Selection endpoints, wheel-scrolling more than 100 turns away from a focused copy/tool control removes the focused node and resets keyboard/screen-reader position; a native selection spanning more than the cap is truncated for the same reason. The current far-target test codifies viewport priority, but there is no interaction handoff before eviction. Please define and implement that boundary—for example, explicitly move/restore focus and settle the selection, or keep a separate pinned interaction island/controlled overflow—and add a browser regression. Otherwise the stated focus/native-selection preservation only holds while the interaction happens to fit under the cap.

M4n5ter: Addressed in `fde3e324e`. Before installing a window that excludes an active focus or selection endpoint, the virtualizer now settles the selection and moves focus to the stable transcript container. The Electron regression first auto-loads more than 100 resident turns, then verifies the endpoint is evicted with selection cleared and focus preserved at the transcript boundary. The PR body now states that native selection is guaranteed within the mounted window rather than across the full durable transcript.

M4n5ter:
Addressed the review in `fde3e324e`:

- Removed the incremental redaction implementation and its dedicated tests; assistant/thinking display remains on the established whole-text boundary. The differential-safe follow-up is tracked in #2990.
- Derived virtual geometry from the rendered 16 px row gap and restored the same spacing inside turn-owned conversation items.
- Strengthened the Electron regression so it auto-loads more than 100 resident turns, proves the mounted window is non-empty and bounded, and covers hard-limit focus/selection handoff.
- Narrowed the PR body to the actual native-selection boundary and documented remount behavior.

The review fix is net -221 lines.

M4n5ter:
Made a few additional polish improvements to make long-session navigation and upward scrolling feel smoother. The related regression coverage has also been updated, and the affected CI checks are passing.

<details>
<summary>中文</summary>

我们又补充了一些体验优化，让长会话导航和向上滚动更加流畅；相关回归测试也已同步更新，受影响的 CI 检查现已通过。

</details>

Astro-Han (commented):
Thanks for turning this around so fast. The redaction removal is clean — `incremental-display-redaction.*` is gone, `thinking-stream.ts` and `live-turn-projection.ts` are byte-identical to `main`, and `assistant-stream.ts` differs only by a deleted comment. The gap fix is right and now measured rather than assumed (`turnLayoutGap` reading computed `rowGap`, folded into `turnLayoutKey`, with the E2E asserting `{list: 16, turn: 16}`), the E2E lower bound and `data-turn-source-count > 100` precondition close the false-green, and the focus/selection handoff answers @MicroGery's inline note.

This round I looked hardest at the three commits after that fix — `6d3258daa`, `9031a121e`, `d02a632b7` — since they arrived described only as polish. Two P1s and four P2s below.

One framing note first, because it explains why I'm suggesting a different shape of fix rather than six patches.

## The findings share a root cause

This PR replaces "mounted is permanent" with "mounted is temporary and repeatable". That is the right call — `content-visibility` only skips rendering, the nodes stay in the DOM, so it cannot make the client bounded the way #2913 needs. But five assumptions across five unrelated files were built on the old model, and swapping it out without inventorying them is what produced most of what follows:

- component-local state (`defaultIsExpanded` only ever needed to be read once, because nothing remounted)
- accessibility (`aria-live` was safe, because scrolling never added nodes)
- browser capabilities (find and select-all worked, because everything was in the DOM)
- the measurement pipeline (measuring once was enough)
- navigation intent (`pendingReveal` assumed its target would always be in the DOM)

Worth noting the same shape produced the P0 from round one — a 512-char window standing in for the whole text is the same "local snapshot treated as the whole truth".

There's a second pattern that explains why these reached review: **the tests cover a different place than the risk lives.** `turn-virtualizer.test.ts` has 12 pure-function tests and that layer is genuinely solid — I fuzzed `stableTurnVirtualWindowForViewport` over 200,000 randomized layouts (3–300 turns, 20–6000px heights, arbitrary starting windows) with zero non-idempotent fixed points, zero oscillations, and zero windows failing to cover the viewport, including at the hard cap. The containment guards in `9031a121e` are correct. But every real defect below lives at the hook/DOM seam, and that layer has no tests at all — which is how you get the result in P2-2.

## P1-1 — this reintroduces the duplicate Processing disclosure that #1728 deliberately removed

`chat-turn.tsx:1131` wraps `ProcessingBlock` in `<ChatReasoning defaultIsExpanded={live}>`, and `astryx-chat-reasoning.tsx:132` renders `: null` when collapsed.

The history matters here:

- **#1320** (2026-07-22, before Astryx) built the collapsible Processing block.
- **#1728** (2026-08-01, the Astryx conversation migration) removed it. `dbe6f515d` dropped `useToolDisclosure` and flattened it to `<div className="maka-processing-sequence">`. The PR summary states the intent in its first bullet: *"remove the duplicate product-owned Processing disclosure while preserving runtime identity, streaming order, redaction, and tool detail content."* The reason is that Astryx's `ChatToolCalls` and reasoning components carry their own disclosure, so a product-owned wrapper is a second fold around a first one.
- `main` then pinned the invariant in a test comment: *"adds no second Processing disclosure around the native reasoning and tool-call disclosures."*
- `timeline-fold.ts:13-16` says the same thing in the data layer, about why a pure-thinking run is left bare: *"wrapping a lone reasoning block would just double the fold."*
- This PR restores the wrapper and rewrites that test comment to say the opposite (`streaming-handoff.test.ts`: "Historical processing is allowed to stay unmounted").

Concretely, on this head a historical turn that used tools renders a collapsed card labelled `正在处理…` / `Working…` — `getConversationCopy(...).messages.processing` (`conversation-copy.ts:400,540`), present-progressive status copy on work that finished last week — and opening it reveals another chevron (`深度思考`) of the same visual family. `chat-message.css:105` also adds `.maka-processing-block` to the same `:is()` rule that gives `.maka-deep-thinking` its sticky offset, so the two headers overlap while scrolling an expanded block.

There's a further defect that disappears with the wrapper: `astryx-chat-reasoning.tsx:58` reads `defaultIsExpanded` only as a `useState` initializer, and `ProcessingBlock` uses it uncontrolled. Virtualization newly makes a running turn unmount and remount; if that lands between steps (thinking done, tools done, answer text streaming) `live` is false, the block mounts collapsed, and the next step's `isStreaming` shimmers a header whose body never opens.

**The DOM-node saving is a real goal and I'm not asking you to drop it.** But the layer is wrong: `astryx-chat-reasoning.tsx` is a vendor eject whose header states *"DOM, state, keyboard behavior, icons, and compiled StyleX atoms below are the official component"*, and `: null` makes that false for every `ChatReasoning` in the product, not just this one. The way to stop mounting collapsed detail is at Astryx's own disclosure — an upstream hook, or an explicitly-annotated deviation in the eject with the reason recorded. #2996 is the same class of boundary question.

Suggested resolution: revert to the #1728 shape here, and take "collapsed disclosures should not mount their content" to the Astryx seam in its own change.

## P1-2 — one `revealTurn` that doesn't land disables scroll-driven windowing for the rest of the session

`use-turn-virtualizer.ts:148` sets `pendingReveal.current = turnId` and then calls `installWindow(next)`. The only clear is inside `if (target)` in the layout effect (`:160-171`) — no else branch, no timeout, no clear on session change. And `:189` gates the entire scroll path on it: `if (!virtualizationRequired || pendingReveal.current) return;`.

The most direct way in is `installWindow` itself: it opens with `if (sameWindow(stateRef.current.window, next)) return;`. When the reveal target is already inside the current window the windows match, it returns immediately, no `setState` happens, the layout effect's `current.window` dependency never changes, so the effect does not re-run and the clear never executes. Clicking a prompt-rail tick in a session small enough that the target is already mounted is enough.

After that, `updateWindow` returns on every scroll and resize for the life of the session. The window still updates when `turnIds` change (the render-body reconcile path), so it isn't visibly broken right away — but once the session grows past the virtualization threshold, scrolling up leaves spacers where turns should be, and nothing recovers short of a session switch.

The fix belongs in the reveal state machine: clear `pendingReveal` whenever the layout effect runs, found or not, or key it to a sequence number the way `prompt-anchor-rail.tsx` keys `jumpSequenceRef`.

## P2-1 — when `virtualizationRequired` is false, nothing can ever schedule a re-measure

`use-turn-virtualizer.ts:208` computes `const anchor = virtualizationRequired ? captureChatScrollAnchor(root) : undefined;` and `:222` gates the geometry bump on `changed && anchor`. At `8ba028e93` that gate was `if (changed)`.

`setGeometryRevision` is the only thing that invalidates the `layout` memo — `turnHeightIndex.lookup` returns a Map that `record` mutates in place, so its identity never changes. On the non-virtualized path `scheduleWindow` returns at `:201` and `updateWindow` at `:189`, so there is no other path either: measured heights are recorded and never applied.

Predicted failure: open a static session of, say, 8 tall turns in an 800px viewport. Estimated geometry is `8×280 + 7×16 = 2352px`, below the `800 + 2×1000` threshold, so everything mounts. Real height is ~4800px, which should flip virtualization on — but the ResizeObserver records the heights, `anchor` is undefined, no revision bump, no render, and `virtualizationRequired` is only recomputed during render. The transcript stays fully mounted until some unrelated state change happens to re-render. `Math.max(totalHeight, root.scrollHeight)` does rescue it, but only on that next render, which an already-loaded static transcript may never produce.

"Geometry changed" and "we are currently virtualizing" are orthogonal — bump on `changed`, and use the anchor only when there is one.

## P2-2 — `9031a121e` undid `6d3258daa`'s guard, and the test named after it was rewritten to match

`6d3258daa` added the minimum-window guard at `turn-virtualizer.ts:130-133` and the `expandRange` call at `:222-227`, both sized by `INITIAL_TURN_WINDOW_SIZE`, which was **40** at that commit. `9031a121e` changed that constant to **2** without revisiting either, taking the floor from 40 turns to 2. The same commit rewrote the other guard's test in place, from asserting `[0,10]` / `[160,200]` to `[8,10]` / `[198,200]` — a test named `does not inherit an undersized window while transcript turns arrive` now pins a 2-turn window.

I removed the `minimumSize` guard entirely and rebuilt: **all 12 tests in `turn-virtualizer.test.ts` still pass**, including `recovers an undersized viewport window instead of oscillating between turns`, which was added to cover it. So the guard currently has no coverage at all.

Was the constant change intentional — a deliberate lower floor — or did the initial-tail size get tuned without noticing the guard shares the constant? The answer changes the fix. Either way the initial tail size and the minimum window floor are two different quantities and shouldn't share one name.

## P2-3 — virtualizing inside a live region re-announces history to screen readers

Astryx's `ChatMessageList` renders the scroll container as `role="log"` with unconditional `aria-live="polite"` (`aria-busy` tracks streaming separately). A polite log announces descendant additions. On `main`, turns stayed mounted once mounted, so scrolling added nothing to the live region. Now each window shift mounts turn subtrees inside it, so steady-state scrolling through history plausibly re-announces old turns as if they were new messages, and AT browsing of the log only ever sees the mounted window.

I can state the mechanism but not the severity — announcement behavior varies by client, so this needs one VoiceOver pass over the 120-turn E2E fixture before it can be sized.

Worth saying that a `role="log" aria-live="polite"` wrapping a thousand turns of history was already questionable before this PR; virtualization is what makes it visible. The clean fix is probably to scope the live region to the streaming turn rather than the whole transcript, which would also fix the pre-existing issue.

## P2-4 — two undisclosed runtime behaviors

**Idle history prefetch.** `use-chat-scroll.ts:133` schedules a `requestIdleCallback` after arrival that calls `requestEarlier()` when `scrollTop <= max(640, clientHeight * 2)` — history loads with no user gesture, `loadBefore(DESKTOP_TRANSCRIPT_RANGE_MAX_BYTES)` at 512 KiB. It converges (the effect doesn't re-run while `hasOlderHistory` stays true, and once it flips false the cleanup identity-guards `requestEarlierRef` to a noop), so this is a disclosure item, not a defect. The PR body's "automatic history loading" refers to the pre-existing scroll/wheel trigger.

**Per-frame batching of the live delta path.** `app-shell-session-events.ts:189-197` defers `text_delta` / `thinking_delta` for the active session to `scheduleFrame`, publishing through a module-level `displayBatch` shared across handler instances. This is in the PR summary as "publishes active text/thinking updates at most once per animation frame", so it is disclosed — but two ordering questions I couldn't settle either way and would rather ask than assert:

- `settleAssistantStreaming` (`:160-168`) reads `liveTurnBySessionRef.current` and calls `settleLiveStep` without draining `displayBatch.pendingEvents`, and there's an `await refreshMessages` in between. Is the interleaving with a pending frame flush intentional?
- The scheduled callback closes over the creating handler's `setLiveTurnBySession` and `uiLocale` while `framePending` is shared, so after a handler replacement a newly queued batch is flushed by the previous instance's closure. The `shares pending display events across handler replacement` test covers the queue but not the closure.

## P3

- The PR body still says "normally mounting at most 60 turns". With `preferredTurns: INITIAL_TURN_WINDOW_SIZE` on the steady-state path and that constant at 2, window size is whatever viewport + 2×overscan needs; 60 is an observed value, not a configured one. `DEFAULT_TURN_WINDOW_SIZE = 60` survives only in `revealTurn` and the reconcile branch, so two sizing vocabularies now coexist.
- `assistant-stream.ts` lost the L4 comment recording `@kenji review Blocker 1` and why cross-delta rescanning is required, while keeping the call it guarded. That comment is the guardrail against re-attempting what #2990 now tracks — please restore it.
- `Generated-by: Codex` in `9031a121e` and `d02a632b7` is body text rather than a trailer (a literal `\n\n` precedes it, and `git log --format='%(trailers)'` returns empty for both; `6d3258daa` is correct). AGENTS.md asks for `git commit --trailer` and for the trailer to survive squash.
- `estimatedTurnHeight` (`turn-virtualizer.ts:35`) floors at `DEFAULT_TURN_ESTIMATED_HEIGHT`, so a transcript of uniformly short turns is permanently overestimated until every turn has been measured. The floor is right as a cold-start default and wrong once measurements exist.
- `handOffExcludedInteraction` runs only inside `installWindow`, so the two render-phase window replacements (the ensureIndex-outside-window branch and the reconcile-oversize rebuild) can evict a focused turn without the handoff. It also runs pre-commit, so if the `setState` updater bails on `sameWindow`, focus moved with no DOM change.
- The body discloses the select-all boundary but not native find. Both follow from the same unmounting; the in-app SearchModal covers it, so this is about symmetric disclosure.
- `streaming-handoff.test.ts:85` asserts `data-processing="block"` appears zero times, but that attribute exists nowhere in the source — the assertion has always been vacuous. It predates this PR; flagging it because the comment above it was rewritten this round. `.maka-processing-block` does exist and would be a real assertion.

## What I'd suggest

Smallest path to merge: revert the Processing wrapper and the `: null` to the #1728 shape (that clears P1-1 and the `defaultIsExpanded` defect together), fix `pendingReveal` and the `changed && anchor` gate, split `INITIAL_TURN_WINDOW_SIZE` into two constants, disclose the prefetch, and get one VoiceOver pass on the live region.

Beyond this PR, the root cause suggests two follow-ups worth their own issues: lifting turn-local state into a `turnId`-keyed store so unmount/remount is genuinely lossless — which is what makes future window-policy changes safe — and a jsdom/hook test layer that treats "unmounted mid-stream and remounted" as a first-class scenario. That layer is where all six findings above live, and it's the layer with no tests today.

---

*Review assistance: Claude Code (Opus 5) read the diff, verified the `#1320 → #1728 → this PR` history, reproduced the removed-guard result, and fuzzed the window solver; Codex, Kimi K3 and a third independent agent each ran a separate adversarial pass over the same head. They corrected three of my own findings and two of my severities, and two of the findings above (P1-2, P2-2) came from those passes rather than from me. I verified the `#1728` intent statement, the `messages.processing` copy, the trailer parsing, the 12-tests-still-green result, and the `pendingReveal` path myself. Per AGENTS.md this is not independent human review — this still needs one before merge.*

<details>
<summary>中文</summary>

感谢这么快的响应。脱敏这块移除得很干净——`incremental-display-redaction.*` 已删除，`thinking-stream.ts` 和 `live-turn-projection.ts` 与 `main` 逐字节一致，`assistant-stream.ts` 只差一段被删掉的注释。gap 的修复是对的，而且现在是实测而非假设（`turnLayoutGap` 读计算后的 `rowGap`，并纳入 `turnLayoutKey`，E2E 直接断言 `{list: 16, turn: 16}`）；E2E 的下界和 `data-turn-source-count > 100` 前置条件堵上了假绿；focus/selection 交接也回应了 @MicroGery 的行内意见。

这一轮我重点看了修复之后的三个 commit——`6d3258daa`、`9031a121e`、`d02a632b7`——因为它们只被描述为 polish。下面是两个 P1 和四个 P2。

先说一个总的判断，因为它解释了我为什么建议换一种修法而不是打六个补丁。

## 这些问题有共同根因

本 PR 把"挂载即永久"换成了"挂载是短暂且可重复的"。这个选择是对的——`content-visibility` 只跳过渲染，节点仍在 DOM 中，达不到 #2913 要的有界。但有五个假设建立在旧模型上，且分散在五个互不相干的文件里，替换模型时没有系统清点它们，下面大部分问题由此而来：

- 组件本地状态（`defaultIsExpanded` 只需读一次初值就够，因为过去不存在 remount）
- 无障碍（`aria-live` 是安全的，因为过去滚动不会新增节点）
- 浏览器能力（查找和全选可用，因为全部内容都在 DOM 里）
- 测量管线（测一次就够了）
- 导航意图（`pendingReveal` 假设目标一定会出现在 DOM 中）

值得一提，第一轮那个 P0 也是同一形态——用 512 字符窗口代替完整文本，同样是"把局部快照当成全部真相"。

还有第二个规律，它解释了这些问题为什么能一路走到 review：**测试覆盖的位置和风险所在的位置是错开的。** `turn-virtualizer.test.ts` 有 12 个纯函数测试，那一层确实扎实——我对 `stableTurnVirtualWindowForViewport` 跑了 20 万组随机布局（3–300 个 turn，20–6000px 高度，任意初始窗口），0 个非幂等不动点、0 次振荡、0 个未覆盖视口的窗口，包括硬上限处。`9031a121e` 加的 containment 保护是正确的。但下面每一个真实缺陷都在 hook 与 DOM 的交界处，而那一层完全没有测试——P2-2 的结果就是这么来的。

## P1-1 —— 这重新引入了 #1728 有意移除的重复 Processing disclosure

`chat-turn.tsx:1131` 把 `ProcessingBlock` 包进 `<ChatReasoning defaultIsExpanded={live}>`，`astryx-chat-reasoning.tsx:132` 在折叠时渲染 `: null`。

这里的历史很关键：

- **#1320**（2026-07-22，Astryx 之前）构建了可折叠的 Processing 区块。
- **#1728**（2026-08-01，Astryx 对话面板迁移）移除了它。`dbe6f515d` 去掉 `useToolDisclosure`，改为扁平的 `<div className="maka-processing-sequence">`。该 PR summary 第一条就写明了意图：*"remove the duplicate product-owned Processing disclosure while preserving runtime identity, streaming order, redaction, and tool detail content."* 原因是 Astryx 的 `ChatToolCalls` 和 reasoning 组件自带 disclosure，产品再包一层就是在一层折叠外面再套一层。
- 随后 `main` 用测试注释固化了这个不变量：*"adds no second Processing disclosure around the native reasoning and tool-call disclosures."*
- `timeline-fold.ts:13-16` 在数据层说了同样的话，解释为什么纯 thinking 的 run 不包：*"wrapping a lone reasoning block would just double the fold."*
- 本 PR 恢复了这层包裹，并把那条测试注释改写成了相反的意思（`streaming-handoff.test.ts`："Historical processing is allowed to stay unmounted"）。

具体到这个 head：一个用过工具的历史 turn 现在渲染成一张折叠卡片，标签是 `正在处理…` / `Working…`——即 `getConversationCopy(...).messages.processing`（`conversation-copy.ts:400,540`），把进行时的状态文案用在了上周就完成的工作上；展开后里面还有一个同一视觉家族的 chevron（`深度思考`）。`chat-message.css:105` 还把 `.maka-processing-block` 加进了给 `.maka-deep-thinking` 设置 sticky 偏移的同一条 `:is()` 规则，所以展开滚动时两个头部会重叠。

还有一个缺陷会随这层包裹一起消失：`astryx-chat-reasoning.tsx:58` 只把 `defaultIsExpanded` 当作 `useState` 的初值，而 `ProcessingBlock` 用的是非受控模式。虚拟化新引入了"运行中的 turn 被卸载再挂载"这件事；如果这一刻恰好落在两个步骤之间（thinking 完成、tools 完成、answer 文本正在流），`live` 为 false，区块以折叠状态挂载，接着下一步的 `isStreaming` 让头部闪烁，而它的主体再也不会打开。

**减少 DOM 节点是个真实目标，我不是要你放弃它。** 但改的层不对：`astryx-chat-reasoning.tsx` 是一份 vendor eject，其文件头写着 *"DOM, state, keyboard behavior, icons, and compiled StyleX atoms below are the official component"*，而 `: null` 让这句话对产品里**每一个** `ChatReasoning` 都不再成立，不只是这一个。要让折叠的 disclosure 不挂载内容，正确的位置是 Astryx 自己的 disclosure——争取一个上游钩子，或者在 eject 文件里显式标注这处偏离并记录原因。#2996 是同一类边界问题。

建议的处理：这里回到 #1728 的形状，把"折叠的 disclosure 不应挂载其内容"作为独立改动推到 Astryx 那个接缝上。

## P1-2 —— 一次没落地的 `revealTurn` 会让滚动驱动的窗口更新在整个会话里失效

`use-turn-virtualizer.ts:148` 先设置 `pendingReveal.current = turnId`，再调用 `installWindow(next)`。唯一的清除位于 layout effect 的 `if (target)` 分支内（`:160-171`）——没有 else 分支，没有超时，没有会话切换时的清理。而 `:189` 用它门控了整条滚动路径：`if (!virtualizationRequired || pendingReveal.current) return;`。

最直接的触发点就是 `installWindow` 本身：它开头是 `if (sameWindow(stateRef.current.window, next)) return;`。当 reveal 的目标已经在当前窗口内时，两个窗口相同，函数立即返回，`setState` 不发生，layout effect 的 `current.window` 依赖不变，因此 effect 不会重跑，清除那行代码永远不执行。在一个小到目标已经挂载的会话里点一次 prompt-rail tick 就够了。

此后，`updateWindow` 在该会话余下的时间里对每次滚动和 resize 都立即返回。窗口仍会在 `turnIds` 变化时更新（render body 的 reconcile 路径），所以不会立刻显得坏掉——但一旦会话增长到超过虚拟化阈值，向上滚动时本该有 turn 的位置就只剩 spacer，除了切换会话没有别的办法恢复。

修复应落在 reveal 状态机：无论是否找到目标，layout effect 每次运行都清除 `pendingReveal`；或者像 `prompt-anchor-rail.tsx` 处理 `jumpSequenceRef` 那样，用序号来标识。

## P2-1 —— `virtualizationRequired` 为 false 时，没有任何东西能触发重新测量

`use-turn-virtualizer.ts:208` 计算 `const anchor = virtualizationRequired ? captureChatScrollAnchor(root) : undefined;`，`:222` 用 `changed && anchor` 门控几何 bump。在 `8ba028e93` 时这个 gate 还是 `if (changed)`。

`setGeometryRevision` 是唯一能让 `layout` memo 失效的东西——`turnHeightIndex.lookup` 返回的 Map 被 `record` 原地修改，引用永不改变。而在非虚拟化路径上，`scheduleWindow` 在 `:201` 返回、`updateWindow` 在 `:189` 返回，所以也没有别的路径：测量到的高度被记录下来，却永远不会被应用。

预测的失败：在 800px 视口中打开一个含 8 个高 turn 的静态会话。估算几何是 `8×280 + 7×16 = 2352px`，低于 `800 + 2×1000` 的阈值，于是全部挂载。真实高度约 4800px，本应触发虚拟化——但 ResizeObserver 记录了高度，`anchor` 是 undefined，没有 revision bump，没有 render，而 `virtualizationRequired` 只在 render 时重新计算。这个 transcript 会一直保持全量挂载，直到某个无关的状态变化恰好引发重渲染。`Math.max(totalHeight, root.scrollHeight)` 确实能救回来，但只在那次后续 render 时生效，而一个已加载完毕的静态 transcript 可能永远不会产生它。

"几何变了"和"我们正在虚拟化"是两件正交的事——应该在 `changed` 时无条件 bump，anchor 只在存在时使用。

## P2-2 —— `9031a121e` 撤销了 `6d3258daa` 的保护，而以它命名的测试被改写以匹配

`6d3258daa` 在 `turn-virtualizer.ts:130-133` 加了最小窗口保护，并在 `:222-227` 加了 `expandRange` 调用，两者的尺寸都取自 `INITIAL_TURN_WINDOW_SIZE`，而该常量在那个 commit 时是 **40**。`9031a121e` 把它改成 **2**，没有回头检视这两处，于是下限从 40 个 turn 降到 2 个。同一个 commit 还就地改写了另一个保护的测试，把断言从 `[0,10]` / `[160,200]` 改成 `[8,10]` / `[198,200]`——一个名为 `does not inherit an undersized window while transcript turns arrive` 的测试，现在钉住的是一个 2 个 turn 的窗口。

我把 `minimumSize` 保护整段删掉后重新构建：**`turn-virtualizer.test.ts` 里 12 个测试仍然全部通过**，包括为它而加的 `recovers an undersized viewport window instead of oscillating between turns`。也就是说这个保护目前完全没有覆盖。

常量的改动是有意的吗——是刻意降低下限，还是在调整初始尾窗大小时没注意到保护共用了同一个常量？答案决定了该怎么修。无论如何，初始尾窗大小和最小窗口下限是两个不同的量，不应共用一个名字。

## P2-3 —— 在 live region 内部做虚拟化，会把历史重新播报给读屏用户

Astryx 的 `ChatMessageList` 把滚动容器渲染为 `role="log"` 并无条件带 `aria-live="polite"`（`aria-busy` 单独跟踪流式状态）。polite 的 log 会播报后代节点的新增。在 `main` 上，turn 一旦挂载就保持挂载，所以滚动不会往 live region 里添加任何东西。现在每次窗口移动都会在其中挂载 turn 子树，因此在历史中来回滚动很可能把旧 turn 当作新消息重新播报，而辅助技术浏览这个 log 时也只能看到当前挂载的窗口。

我能说清机制，但定不了严重程度——播报行为因客户端而异，需要先用 VoiceOver 对 120-turn 的 E2E fixture 实测一次才能定级。

另外值得一提：用 `role="log" aria-live="polite"` 包住上千条历史，在本 PR 之前就已经是可疑的，只是虚拟化让它显形了。干净的修法可能是把 live region 收窄到正在流式的那个 turn，而不是整个 transcript——那样也会顺带修掉这个既有问题。

## P2-4 —— 两处未披露的运行时行为

**空闲时预取历史。** `use-chat-scroll.ts:133` 在 arrival 之后安排了一个 `requestIdleCallback`，当 `scrollTop <= max(640, clientHeight * 2)` 时调用 `requestEarlier()`——没有任何用户手势就会加载历史，实际走的是 `loadBefore(DESKTOP_TRANSCRIPT_RANGE_MAX_BYTES)`，即 512 KiB。它是收敛的（`hasOlderHistory` 保持为 true 时该 effect 不会重跑，一旦翻为 false，cleanup 会用身份判断把 `requestEarlierRef` 置为 noop），所以这是一个披露事项而非缺陷。PR 正文里的 "automatic history loading" 指的是既有的滚动/滚轮触发。

**活跃 delta 路径的按帧批处理。** `app-shell-session-events.ts:189-197` 把活跃会话的 `text_delta` / `thinking_delta` 推迟到 `scheduleFrame`，并通过一个跨 handler 实例共享的模块级 `displayBatch` 发布。PR summary 里有"publishes active text/thinking updates at most once per animation frame"，所以它是披露过的——但有两个顺序问题我无法定论，与其断言不如问你：

- `settleAssistantStreaming`（`:160-168`）读取 `liveTurnBySessionRef.current` 并调用 `settleLiveStep`，其间没有排空 `displayBatch.pendingEvents`，而中间还隔着一个 `await refreshMessages`。与待发帧的交错是有意设计的吗？
- 被安排的回调闭包捕获的是创建它的那个 handler 的 `setLiveTurnBySession` 和 `uiLocale`，而 `framePending` 是共享的，所以 handler 被替换后，新排队的批次会由上一个实例的闭包来 flush。`shares pending display events across handler replacement` 这个测试覆盖了队列，但没有覆盖闭包。

## P3

- PR 正文仍写着"normally mounting at most 60 turns"。稳态路径传的是 `preferredTurns: INITIAL_TURN_WINDOW_SIZE` 而该常量为 2，窗口大小实际由视口加 2×overscan 决定；60 是观察值，不是配置值。`DEFAULT_TURN_WINDOW_SIZE = 60` 只在 `revealTurn` 和 reconcile 分支里存活，于是两套尺寸词汇并存。
- `assistant-stream.ts` 丢掉了记录 `@kenji review Blocker 1` 以及为何必须做跨 delta 重扫的 L4 注释，却保留了它所守护的调用。那段注释正是防止重蹈 #2990 所跟踪问题的护栏——请恢复它。
- `9031a121e` 和 `d02a632b7` 里的 `Generated-by: Codex` 是正文而非 trailer（前面有字面的 `\n\n`，`git log --format='%(trailers)'` 对这两个都返回空；`6d3258daa` 是正确的）。AGENTS.md 要求用 `git commit --trailer`，并要求 trailer 在 squash 后仍然保留。
- `estimatedTurnHeight`（`turn-virtualizer.ts:35`）以 `DEFAULT_TURN_ESTIMATED_HEIGHT` 为下限，因此一个全是矮 turn 的 transcript 会被持续高估，直到每个 turn 都被测量过。这个下限作为冷启动默认值是对的，在已有测量之后就不对了。
- `handOffExcludedInteraction` 只在 `installWindow` 内运行，所以两条 render 阶段的窗口替换路径（ensureIndex 在窗口之外的分支、以及 reconcile 超限重建）可以在不做交接的情况下驱逐一个持有焦点的 turn。它还运行在 commit 之前，因此如果 `setState` 的 updater 因 `sameWindow` 而放弃更新，焦点已经被移走但 DOM 没有变化。
- 正文披露了全选的边界，但没有披露原生查找。两者都源于同一次卸载；应用内的 SearchModal 覆盖了这个场景，所以这只是披露对称性的问题。
- `streaming-handoff.test.ts:85` 断言 `data-processing="block"` 出现零次，但这个属性在源码中任何地方都不存在——该断言一直是恒真的。它早于本 PR；之所以提出来，是因为它上面的注释这一轮被改写了。`.maka-processing-block` 是真实存在的，用它才是有效断言。

## 我的建议

最小的合并路径：把 Processing 包裹层和 `: null` 回退到 #1728 的形状（这会同时解决 P1-1 和 `defaultIsExpanded` 缺陷）、修复 `pendingReveal` 和 `changed && anchor` 这个 gate、把 `INITIAL_TURN_WINDOW_SIZE` 拆成两个常量、披露预取行为，并对 live region 做一次 VoiceOver 实测。

本 PR 之外，这个根因指向两个值得各自立 issue 的后续：把 turn 本地状态提升到一个以 `turnId` 为键的 store，让卸载/重新挂载真正无损——这才是让未来的窗口策略调整变安全的前提；以及补一层 jsdom/hook 测试，把"流式途中被卸载并重新挂载"当作一等场景。上面六条发现全都住在那一层，而今天那一层没有任何测试。

---

*评审协助：Claude Code（Opus 5）阅读了 diff，核实了 `#1320 → #1728 → 本 PR` 这条历史，复现了移除保护后测试仍全绿的结果，并对窗口求解器做了 fuzz；Codex、Kimi K3 以及第三个独立 agent 各自对同一 head 做了一轮对抗性复审。它们纠正了我三条发现和两处严重度判断，上面有两条发现（P1-2、P2-2）来自它们而不是我。#1728 的意图陈述、`messages.processing` 的文案、trailer 的解析、12 个测试仍全绿的结果，以及 `pendingReveal` 的路径，都是我本人核验的。按 AGENTS.md，这不算独立人工评审——合并前仍然需要一次。*

</details>

M4n5ter:
Thanks for the thorough review. I independently re-checked each item against `d02a632b7`.

The duplicate Processing ownership boundary and the measurement invalidation issue are valid, and I’ll address them before merge. I’ll also clean up the height estimator, interaction handoff coverage, vacuous regression assertion, PR documentation, and commit trailer.

Two nuances from the verification:

- The reported mounted-target `pendingReveal` trigger bypasses `revealTurn` in the current rail path, but the state should still be made self-clearing defensively.
- The two-turn warm start was intentional based on the real first-scroll trace. I’ll separate initial sizing from minimum-window policy instead of restoring the previous 40-turn floor.

I’ll also run a real VoiceOver pass before deciding the live-region treatment. The frame-batching questions do not currently establish a failing ordering path, so I won’t add complexity there without a reproduction.

I’ll follow up with the focused fixes and verification results.

<details>
<summary>中文</summary>

感谢这次详细的审查。我已经基于 `d02a632b7` 对每一项进行了独立复核。

重复的 Processing 所有权边界以及测量结果无法触发重新计算的问题确认成立，会在合并前修复。同时也会处理高度估算、交互交接覆盖、无效回归断言、PR 文档和 commit trailer。

有两点需要补充：

- 当前 prompt rail 在目标已挂载时不会进入 `revealTurn`，因此 review 中描述的 `pendingReveal` 触发路径并不成立；不过该状态仍应做成可自清理的。
- 两个 turn 的初始窗口是根据真实首次上滚 trace 有意调整的。我们会拆分初始窗口与最小窗口策略，而不是恢复原来的 40-turn 下限。

此外会进行一次真实 VoiceOver 验证，再决定 live region 的处理方式。当前 frame batching 两项尚未形成可复现的错误顺序，因此不会在没有反例的情况下增加复杂度。

修复和验证完成后会继续跟进。

</details>

M4n5ter:
Addressed in `76e4ffae5`:

- Restored the native reasoning/tool disclosure ownership and the ejected Astryx disclosure DOM.
- Made measured geometry reliably enter virtualization, removed the permanent height-estimate floor, and routed external target changes through the existing interaction-handoff boundary.
- Made reveal intent self-clearing, separated initial and minimum window policies, replaced the vacuous regression assertion, and removed two low-value tests.
- Updated the PR body with the actual prefetch and native Find/Select All boundaries. A real macOS VoiceOver pass did not re-announce historical turns during upward loading, so no speculative live-region state machine was added.

The full workspace suite, production build, and the 7 affected Electron E2E cases pass.

<details>
<summary>中文</summary>

已在 `76e4ffae5` 中处理：

- 恢复 reasoning/tool 原生 disclosure 的所有权，以及 eject 后 Astryx disclosure 的原始 DOM 行为。
- 让真实测量结果能够可靠触发虚拟化，去掉永久高度估算下限，并让外部目标切换统一经过已有的交互交接边界。
- 让 reveal 意图自动清理，拆分初始窗口与最小窗口策略，替换无效断言，并删除两条低价值测试。
- 更新 PR 正文，准确说明预取与原生查找/全选边界。真实 macOS VoiceOver 验证没有在向上加载历史时重复播报旧消息，因此没有加入推测性的 live-region 状态机。

全仓测试、生产构建以及 7 条受影响的 Electron E2E 均已通过。

</details>

packages/ui/src/chat-view.tsx
```diff
@@ -726,9 +644,16 @@ export function ChatView(props: {
                     {conversationItemsByTurn.get(turn.turnId)?.map((item) => (
```
MicroGery: [P2] Associate turn-owned conversation-item interactions with the virtual turn

`interactiveTurnRange` and `handOffExcludedInteraction` resolve ownership with `element.closest('[data-turn-id]')`. `TurnView` owns that attribute, but these conversation items are rendered as its siblings inside a wrapper that only has `data-virtual-turn-id`. Focus on a `PlanProposalCard` action, or a selection inside the card, therefore resolves to no turn: a target jump or window shift can unmount the wrapper without retaining the range, clearing the selection, or moving focus to the transcript boundary. The existing hard-cap E2E only exercises controls inside `TurnView`, so it misses this path.

Please put the owning turn ID on the shared wrapper or teach the interaction helpers to resolve `data-virtual-turn-id`, and cover a focused plan action at an eviction boundary.

packages/ui/src/chat-view.tsx
```diff
@@ -668,27 +591,22 @@ export function ChatView(props: {
...
               )}
               {mountedTurns.map((turn) => {
                 return (
-                  <Fragment key={turn.turnId}>
+                  <div
+                    key={turn.turnId}
+                    className="maka-turn-virtual-item"
+                    data-virtual-turn-id={turn.turnId}
+                  >
                     <TurnView
```
MicroGery, line 610: [P2] A turn-count cap does not yet bound a single turn DOM

The window caps `mountedTurns`, but each mounted `TurnView` still renders its complete `turn.timeline`. Desktop does not set `maxSteps`, so one agent turn can accumulate an arbitrarily long sequence of reasoning, tool, and result entries; a one-turn transcript can therefore mount thousands of nodes while the virtualizer reports a window size of 1. That leaves the #2913 non-negotiable property "Renderer state and DOM remain bounded independently of history size" unmet and is not exercised by the current many-small-turn fixture.

If per-turn timeline virtualization is intentionally a follow-up, please narrow the completion claim and track it explicitly. Otherwise this boundary needs a node or byte-aware cap, or a second-level timeline window, plus a single tool-heavy-turn performance regression.

packages/ui/src/turn-virtualizer.ts
```diff
@@ -0,0 +1,339 @@
...
+  }
+  const firstId = previousIds[previous.start];
+  const lastId = previousIds[previous.end - 1];
+  let start = firstId === undefined ? -1 : nextIds.indexOf(firstId);
+  const last = lastId === undefined ? -1 : nextIds.indexOf(lastId);
+  if (start < 0 || last < start) {
+    return initialTurnVirtualWindow(nextLayout, MIN_TURN_WINDOW_SIZE, ensureIndex);
+  }
+  let end = last + 1;
+  if (previous.end === previousIds.length) {
```
MicroGery: [P2] Do not infer bottom-follow intent from a tail-ending window

`previous.end === previousIds.length` only says the mounted overscan window reaches the old tail; it does not say the reader is bottom-pinned. For example, with a 100-turn window `[100, 200]`, the reader can be viewing `t105`; appending 80 turns in one sparse-replica update produces `[180, 280]` here and evicts the visible turn before the DOM observer can repair the viewport. Batched durable catch-up and reconnect make this more than a synthetic append shape.

Please gate tail sliding on the actual ChatLayout auto-follow or bottom-lock state. When auto-follow is unlocked, preserve the old visible identities and restore an anchor instead. A regression should append multiple turns while the reader is scrolled near the top of a tail-ending window.

packages/ui/src/chat-view.tsx
```diff
@@ -726,9 +644,16 @@ export function ChatView(props: {
                     {conversationItemsByTurn.get(turn.turnId)?.map((item) => (
                       <Fragment key={item.id}>{item.content}</Fragment>
                     ))}
-                  </Fragment>
+                  </div>
                 );
               })}
+              {afterHeight > 0 && (
```
MicroGery, line 652: [P1] Keep orphaned conversation items inside the bounded transcript model

The new bottom spacer closes the virtual turn window, but the existing fallback below it still renders every conversation item whose anchor turn is absent from the current sparse replica. Today `AppShell` supplies every durable plan proposal as a conversation item, while `applyPlanEvent` only appends to `state.proposals`. Once older anchor turns are evicted from the resident transcript, all historical `PlanProposalCard` items satisfy that fallback and mount after this spacer. A plan-heavy session can therefore have at most 100 mounted turns but an unbounded number of full proposal cards, each with up to 50 steps, and those cards are also moved semantically to the current tail.

Please page or filter anchored items with their owning transcript range, with separate explicit treatment for a genuinely pending orphan, and add a regression where old proposal anchors are outside the resident range. Otherwise renderer DOM remains history-sized despite the turn cap.

packages/ui/src/chat-view.tsx
M4n5ter: Addressed in `63bd197c1`. Interaction ownership now resolves the nearest `data-turn-id` or `data-virtual-turn-id`, so a conversation-item sibling is owned by the same virtual turn as `TurnView`. The Electron regression now focuses and selects text in a sibling action attached directly to the virtual wrapper, evicts that wrapper, and verifies selection/focus handoff.

packages/ui/src/chat-view.tsx
M4n5ter: This boundary is intentional for this PR. The implementation plan treats one turn as an indivisible rendering unit; splitting one timeline would introduce a second virtualization/ownership model. I narrowed the PR body to state that the hard bound is turn-based and that per-turn timeline bounding remains part of the broader #2913 work.

packages/ui/src/turn-virtualizer.ts
M4n5ter: Addressed in `63bd197c1`. Reconciliation no longer treats a tail-ending window as follow intent. It preserves the mounted identities across batched appends; when Astryx is actually locked to the bottom, its real scroll movement drives the viewport solver to the new tail. The regression appends 80 turns to a tail-ending window and verifies that the previous identities remain mounted.

packages/ui/src/chat-view.tsx
M4n5ter: Addressed in `63bd197c1`. Conversation items are now placed only with resident anchor turns. The fallback is explicitly bounded to one `renderWhenAnchorMissing` item, and Desktop marks only the latest pending-approval proposal for that path. A focused regression verifies that historical missing anchors are omitted while only the latest explicit orphan remains.

MicroGery (approved):
LGTM

Issue #2973, opened 2026-08-13
Proposal: lifecycle-managed Extension Runtime for safe agent self-extension


xxhZs:
## Problem

Maka currently has several extension mechanisms, including built-in tools, deferred `load_tools` groups, Skills, MCP, automations, and Agent Graph capabilities. However, these capabilities are assembled through fixed startup or backend wiring. An agent can reveal a tool that was already registered, but it cannot safely define, activate, inspect, update, stop, or remove a new runtime capability.

Adding dynamic scripts directly to the existing process would create a second problem: a script update could mutate the tool surface or leave listeners, timers, UI entries, and other effects behind while an Agent Run is active. This would conflict with Maka's durable Run, continuation, permission, and recovery guarantees.

The goal should not be a second plugin runtime beside the existing one. Maka's shipped capabilities should remain the protected core composition, while installed and agent-authored extensions enter through the same composition boundary.

## Desired outcome

Introduce a lifecycle-managed Extension Runtime that lets capabilities behave like applications:

- Define or install an immutable extension revision without running it.
- Activate an exact revision and own every effect it contributes.
- Stop it cleanly while retaining the revision.
- Update through a prepared candidate without disturbing the current activation.
- Roll back when preparation or activation fails.
- Uninstall only after active work has drained.
- Inspect the live extension surface and its diagnostics.

Static packages and agent-authored scripts should use one model. Temporary versus persistent behavior should be expressed by binding scope and lifetime rather than separate plugin systems.

Each Agent Run should use an immutable Extension Composition snapshot. Changes produce a candidate composition for future work and must never mutate the composition of an active Run.

The protected core should initially remain non-replaceable. Extensions may contribute through explicit, typed surfaces, introduced progressively:

1. Tools
2. Runtime inspection and control
3. Agent-authored session extensions
4. UI slots
5. Read-only runtime observation
6. Restricted decision hooks, context, state, timers, and Agent Graph capabilities

Untrusted or agent-authored code must not execute directly inside the Electron main process. Lifecycle cleanup and execution isolation are separate requirements: lifecycle ownership prevents leaked effects, while an isolated extension host prevents faulty code from destabilizing the core process.

### Core concepts

- **Extension**: stable identity.
- **Revision**: immutable package contents, manifest, permissions, and compatibility metadata.
- **Contribution**: one capability exposed through a typed extension surface, such as a Tool or UI slot.
- **Binding**: declares which revision is desired for a process, user, workspace, session, or run scope.
[...]

PR #2979, opened 2026-08-13
feat(runtime): add extension lifecycle kernel

xxhZs:
## Summary

Add the product-independent Phase 1 lifecycle kernel from #2973, without wiring Tool, UI, or dynamic-script contribution adapters yet.

- install immutable extension revisions without executing extension code
- bind exact revisions to scopes and support start, stop, update, removal, and scope disposal
- own preparation and activation effects with reverse-order cleanup and retryable cleanup diagnostics
- wait for same-scope dependencies, detect cycles, stop dependents before providers, and reactivate them after recovery or update
- update through current/candidate prepare, health-check, activation, commit, and rollback semantics
- expose immutable, deterministic composition snapshots plus live binding diagnostics
- serialize lifecycle mutations so committed state and cleanup ownership remain consistent

The implementation is intentionally isolated from Maka's existing Tool, permission, sandbox, and Run authorities. Those integrations remain follow-up phases in the issue roadmap.

Refs #2973

## Verification

- `npm --workspace @maka/runtime run test:dist`: **2819 pass / 0 fail / 12 skip**
- lifecycle-focused suite: **28 pass / 0 fail**
- compiled implementation coverage: **99.73% lines / 94.96% branches / 100% functions**
- system suite repeated 20 times without intermittent failures
- `npm run typecheck`: passed for all workspaces
- `npx biome check packages/runtime/src/__tests__/extension-lifecycle-kernel.system.test.ts docs/architecture/extension-lifecycle-kernel.md`: passed
- `git diff --check`: passed

The system suite drives the exported kernel rather than mocking lifecycle behavior. It uses a real TCP server/client, actual port release, EventEmitter listeners, timers, a transitive dependency graph, cleanup-failure recovery, and a 2,000-operation seeded lifecycle soak across multiple scopes and revisions.

## Review focus

- current/candidate commit and rollback boundaries
- dependency teardown and reactivation order
- effect ownership when cleanup partially fails
- distinction between deterministic composition identity and activation generation
- Phase 1 boundary: this PR does not yet admit contributions into active Agent Runs

## Checklist

- [ ] No

### DIFF
--- a/packages/runtime/src/extension-lifecycle-kernel.ts
+++ b/packages/runtime/src/extension-lifecycle-kernel.ts
@@ -1,0 +1,1065 @@
+import { createHash } from 'node:crypto';
+
+const ID_PATTERN = /^[a-z][a-z0-9]*(?:[._-][a-z0-9]+)*$/;
+const MAX_ID_LENGTH = 128;
+
+export type ExtensionEffectDisposer = () => void | Promise<void>;
+
+export interface ExtensionDependencyDefinition {
+  readonly extensionId: string;
+}
+
+export interface ExtensionContributionDefinition {
+  readonly id: string;
+  readonly kind: string;
+}
+
+export interface ExtensionPreparationContext {
+  readonly bindingId: string;
+  readonly scopeId: string;
+  readonly extensionId: string;
+  readonly revision: string;
+  readonly signal: AbortSignal;
+  ownEffect(label: string, dispose: ExtensionEffectDisposer): void;
+}
+
+export interface ExtensionActivationContext extends ExtensionPreparationContext {
+  dependency<T = unknown>(extensionId: string): T;
+  dependencyRevision(extensionId: string): string;
+}
+
+export interface ExtensionActivationResult {
+  readonly value?: unknown;
+}
+
+export interface PreparedExtension {
+  /** Validate candidate-local resources without publishing contributions. */
+  healthCheck?(): void | Promise<void>;
+  /** Publish activation-owned effects through `context.ownEffect`. */
+  activate(
+    context: ExtensionActivationContext,
+  ): void | ExtensionActivationResult | Promise<void | ExtensionActivationResult>;
+  /** Release candidate-local resources allocated by `prepare`. */
+  dispose?(): void | Promise<void>;
+}
+
+export interface ExtensionRevisionDefinition {
+  readonly extensionId: string;
+  readonly revision: string;
+  readonly dependencies?: readonly ExtensionDependencyDefinition[];
+  readonly contributions?: readonly ExtensionContributionDefinition[];
+  /**
+   * Definition/install is data-only. The kernel calls `prepare` only when a
+   * binding is enabled and all required dependencies are active.
+   */
+  prepare(context: ExtensionPreparationContext): PreparedExtension | Promise<PreparedExtension>;
+}
+
+export interface ExtensionBindingInput {
+  readonly bindingId: string;
+  readonly scopeId: string;
+  readonly extensionId: string;
+  readonly revision: string;
+}
+
+export type ExtensionBindingStatus =
+  | 'stopped'
+  | 'waiting'
+  | 'preparing'
+  | 'health_check'
+  | 'activating'
+  | 'active'
+  | 'failed';
+
+export type ExtensionLifecycleErrorCode =
+  | 'invalid_definition'
+  | 'revision_already_installed'
+  | 'revision_not_installed'
+  | 'revision_in_use'
+  | 'binding_conflict'
+  | 'binding_not_found'
+  | 'dependency_cycle'
+  | 'prepare_failed'
+  | 'health_check_failed'
+  | 'activation_failed'
+  | 'cleanup_failed';
+
+export interface ExtensionLifecycleDiagnostic {
+  readonly code: ExtensionLifecycleErrorCode;
+  readonly message: string;
+  readonly revision?: string;
+  readonly at: number;
+}
+
+export interface ExtensionBindingInspection {
+  readonly bindingId: string;
+  readonly scopeId: string;
+  readonly extensionId: string;
+  readonly desiredRevision: string;
+  readonly enabled: boolean;
+  readonly status: ExtensionBindingStatus;
+  readonly current?: {
+    readonly revision: string;
+    readonly generation: number;
+  };
+  readonly candidate?: {
+    readonly revision: string;
+    readonly phase: 'preparing' | 'health_check' | 'activating';
+  };
+  readonly waitingFor: readonly string[];
+  readonly pendingCleanupEffects: number;
+  readonly diagnostic?: ExtensionLifecycleDiagnostic;
+}
+
+export interface ExtensionCompositionEntry {
+  readonly bindingId: string;
+  readonly extensionId: string;
+  readonly revision: string;
+  readonly generation: number;
+  readonly contributions: readonly ExtensionContributionDefinition[];
+}
+
+export interface ExtensionCompositionSnapshot {
+  readonly schemaVersion: 1;
+  readonly scopeId: string;
+  readonly digest: `sha256:${string}`;
+  readonly entries: readonly ExtensionCompositionEntry[];
+}
+
+export class ExtensionLifecycleOperationError extends Error {
+  readonly name = 'ExtensionLifecycleOperationError';
+
+  constructor(
+    readonly code: ExtensionLifecycleErrorCode,
+    message: string,
+    options?: { cause?: unknown },
+  ) {
+    super(message, options);
+  }
+}
+
+interface InstalledRevision {
+  readonly extensionId: string;
+  readonly revision: string;
+  readonly dependencies: readonly ExtensionDependencyDefinition[];
+  readonly contributions: readonly ExtensionContributionDefinition[];
+  readonly prepare: ExtensionRevisionDefinition['prepare'];
+}
+
+interface OwnedEffect {
+  readonly label: string;
+  readonly dispose: ExtensionEffectDisposer;
+}
+
+interface EffectCleanupFailure {
+  readonly label: string;
+  readonly cause: unknown;
+}
+
+class EffectCleanupError extends Error {
+  readonly name = 'EffectCleanupError';
+
+  constructor(readonly failures: readonly EffectCleanupFailure[]) {
+    super(
+      `Failed to clean up ${failures.length} extension effect${failures.length === 1 ? '' : 's'}`,
+      { cause: failures[0]?.cause },
+    );
+  }
+}
+
+class EffectOwner {
+  readonly #effects: OwnedEffect[] = [];
+  #accepting = true;
+
+  get size(): number {
+    return this.#effects.length;
+  }
+
+  own(label: string, dispose: ExtensionEffectDisposer): void {
+    if (!this.#accepting) {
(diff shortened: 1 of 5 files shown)


Astro-Han:
Thanks for the kernel — the design is the right shape for the problem (agents need a safe way to define/activate/update/stop/remove runtime capabilities; a dynamic script leaking listeners/timers/UI entries breaks Run persistence and recovery). The immutable-revision + current/candidate + effect-ownership + serialized-mutation model is standard and self-consistent, and I verified it doesn't collide with existing authorities: the kernel is a *new* lifecycle authority inside `@maka/runtime` that deliberately touches no Tool/permission/sandbox/Run authority (grep confirms no pre-existing extension lifecycle path in runtime), which is consistent with "Runtime Host is the sole execution authority" since Phase 1 has no production wiring. The test quality is high — real TCP/EventEmitter/timer leak detection, diamond dependency graphs, cleanup-failure recovery, 2000-iteration seeded soak with invariant assertions — nothing deletable. The 1065-line kernel maps 1:1 to the issue's requirements. CI 9/9 green.

**Conclusion: PASS with two P2s (both handleable or explicitly deferred).**

**P2-1 — `install` triggers a full-scope reconcile that silently re-runs failed bindings' extension code (prepare/activate), contradicting the documented invariant.** `install` calls `#reconcileAllScopes()` → the reconcile loop retries `#activateInitial`/`#updateCurrent` for every enabled record with no dependencies/owners and no current — regardless of `status='failed'`. So installing any revision (or stopping any binding) re-executes the `prepare`/`activate` of a previously-failed binding; if that `prepare` has side effects before registering its cleanup effect (spawn, port bind), every mutation leaks a copy. The results are discarded (`ReconcileResult.errors` is consumed by nobody), so callers never notice. This also contradicts the kernel docs ("Installing a revision never executes extension code") and the `prepare` JSDoc ("called only when a binding is enabled and all required dependencies are active"). With `retry()` existing as the explicit retry API, implicit auto-retry on unrelated mutations is surprising. Fix option: have reconcile skip `status='failed'` records (restoring `retry()`'s explicit semantics), or document the auto-retry behavior and add the test — my suggested test (activate-fail a binding, then `install` an unrelated revision, assert its prepare-count unchanged) fails on the current implementation.

**P2-2 — no production callers: 1065 lines of kernel + 1331 lines of tests + package export are all unreachable.** grep finds no consumer; the PR and issue #2973 explicitly defer wiring ("Those integrations remain follow-up phases"). Accepted as the declared Phase 1 boundary, but please explicitly record that the module is pure dead code with no behavioral verification loop until the Phase 2 adapter lands (and note the real integration risk is Phase 2's `ownEffect` joining the existing Tool registry without creating a parallel registration path).

P3 (optional): an update-introduced dependency cycle leaves "status=failed but current still running" (the cycle handler only sets status/diagnostic, doesn't stop current; only the initial-activation cycle is tested); `candidate?.controller.abort()` in `stop`/`removeBinding`/`disposeScope` is dead code — serialization guarantees no in-flight candidate during mutation, and no AbortSignal listener exists; `#mutate` can deadlock if extension code (prepare/activate) calls any kernel mutation itself (inner mutation queues behind the outer one being awaited) — a Phase 2 adapter starting another extension inside activate would hit it; `activate` on an existing binding doesn't clear `record.diagnostic` (inconsistent with `update`/`start` — stale failure diagnostics linger after later success); a pending retiredOwners flips a record from 'active' to 'failed' on unrelated mutations (reconcile marks any record with retiredOwners>0 as failed even when the new current committed and runs); `#requireRevision` sits outside try/catch (a guard today, fragile if a future path references an uninstalled revision); `#mutate` has no timeout/watchdog — a hung `prepare` blocks all later mutations forever (documented serialization tradeoff, but no recovery).

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagent on `ollama-cloud/deepseek-v4-flash`), which traced the state machine, dependency cascade, commit/rollback paths, and effect ownership on the PR head and verified CI status. P2-1 is a code-path consequence (unrelated-mutation retry), not an observed production failure — Phase 1 has no callers. Please weigh these findings with your own judgment.*

<details>

结论：PASS（2 个 P2 需处理或显式延后）。设计形状正确（agent 需要安全地定义/激活/更新/停止/移除运行时能力；动态脚本泄漏 listeners/timers/UI 条目破坏 Run 持久化与恢复）；不可变 revision+current/candidate+effect 所有权+序列化变更模型标准自洽；与既有 authority 无冲突（@maka/runtime 内*新的*生命周期 authority 刻意不碰 Tool/permission/sandbox/Run authority，grep 确认 runtime 无既有扩展生命周期路径；与"Runtime Host 唯一执行权威"不冲突——Phase 1 未接线）。测试质量高（真实 TCP/EventEmitter/timer 泄漏检测、菱形依赖图、cleanup 失败恢复、2000 次 seeded soak+不变量断言）无可删；1065 行与 issue 需求 1:1 对应。CI 9/9 绿。P2-1：install 触发全 scope reconcile 静默重跑失败 binding 的扩展代码（prepare/activate）——reconcile 循环对 enabled、无依赖/owner、无 current 的 record 无条件重试 #activateInitial/#updateCurrent，不区分 status='failed'；即使失败 binding 的 prepare 在注册清理 effect 前有副作用（spawn/占端口）则每次 mutation 重复泄漏；ReconcileResult.errors 无人消费调用方无感知；与 kernel 文档（"Installing a revision never executes extension code"）和 prepare JSDoc（"仅当 binding enabled 且依赖满足时调用"）矛盾；retry() 既存为显式重试 API，隐式重试令人意外。修复：reconcile 跳过 status='failed' 恢复 retry() 显式语义，或文档化自动重试行为+补测试（我的建议测试：激活失败一个 binding 后 install 无关 revision 断言其 prepare 计数不变——当前实现会失败）。P2-2：无生产调用点（1065 行 kernel+1331 行测试+package export 全部不可达）——PR 与 issue #2973 已显式延后接线，接受为已声明的 Phase 1 边界，但请显式记录"合入后该模块是纯死代码、无行为验证闭环"，并注意 Phase 2 的 ownEffect 并入既有 Tool registry 时不得产生并行注册路径。P3（可选）：update 引入依赖环留"status=failed 但 current 仍在运行"（环处理只置 status/diagnostic 不停 current，仅初始激活环有测试）；stop/removeBinding/disposeScope 的 candidate?.controller.abort() 是死代码（序列化保证 mutation 时无 in-flight candidate、无 AbortSignal 监听者）；#mutate 在扩展代码（prepare/activate）内调 kernel mutation 会死锁（内层排队在外层 awaited 的 mutation 后）；activate 对既有 binding 不清 record.diagnostic 与 update/start 不一致；retiredOwners 挂起时无关 mutation 把 'active' 翻成 'failed'；#requireRevision 在 try 外；#mutate 无超时/看门狗（prepare 挂起永久阻塞所有后续 mutation，已文档化但无恢复手段）。
</details>

packages/runtime/src/extension-lifecycle-kernel.ts
```diff
@@ -0,0 +1,1065 @@
...
+      return this.#inspectRecord(record);
+    });
+  }
+
+  retry(bindingId: string): Promise<ExtensionBindingInspection> {
+    return this.start(bindingId);
+  }
+
+  stop(bindingId: string): Promise<ExtensionBindingInspection> {
+    return this.#mutate(async () => {
```
Astro-Han, line 388: **P1 — `stop` cannot abort the lifecycle hook that is blocking the mutation queue.** `prepare`/`healthCheck`/`activate` are awaited inside `#mutate`; `stop`, `removeBinding`, and `disposeScope` only reach `candidate.controller.abort()` after that same mutation finishes. A hook that never resolves therefore makes the exposed `AbortSignal` unusable and permanently blocks teardown. Keep an out-of-band active-candidate cancellation lane: abort immediately, then enqueue state convergence; add a never-resolving hook regression test.

packages/runtime/src/extension-lifecycle-kernel.ts
```diff
@@ -0,0 +1,1065 @@
...
+      const installed = normalizeDefinition(definition);
+      const key = revisionKey(installed.extensionId, installed.revision);
+      if (this.#revisions.has(key)) {
+        throw new ExtensionLifecycleOperationError(
+          'revision_already_installed',
+          `Extension revision already installed: ${key}`,
+        );
+      }
+      this.#revisions.set(key, installed);
+      await this.#reconcileAllScopes();
```
Astro-Han, line 273: **P1 — `install()` violates the documented effect-free catalog contract.** Installing any unrelated revision runs `#reconcileAllScopes()`, which can retry a previously failed enabled binding and execute its `prepare`/activation, allocating resources or publishing effects. Installing a revision cannot make an already-valid binding newly activatable because `activate()` already requires its exact revision. Remove reconciliation from `install()` and trigger lifecycle work only through explicit binding operations.

packages/runtime/src/extension-lifecycle-kernel.ts
```diff
@@ -0,0 +1,1065 @@
...
+      validateId('extensionId', extensionId);
+      validateRevision(revision);
+      const key = revisionKey(extensionId, revision);
+      if (!this.#revisions.has(key)) {
+        throw new ExtensionLifecycleOperationError(
+          'revision_not_installed',
+          `Extension revision is not installed: ${key}`,
+        );
+      }
+      const user = [...this.#bindings.values()].find(
```
Astro-Han, line 288: **P2 — Pending retired effects are invisible to revision-in-use checks.** After v1→v2 commits and v1 cleanup fails, the v1 owner remains in `retiredOwners` but has no revision identity; `uninstall(ext, '1')` succeeds and that revision can be reinstalled while its old effects remain live. Store the revision with every retired owner and reject uninstall/reinstall until matching cleanup succeeds (or conservatively block all revisions for that extension while retired owners exist).

Astro-Han (commented):
The candidate/current split, dependency ordering, reverse effect cleanup, and scope isolation form a strong lifecycle core. Three seams still violate the kernel's own ownership contract: cancellation is queued behind the hook it must cancel, catalog installation can execute unrelated extension code, and retired effects lose their revision identity.

A simpler first-principles model separates three authorities: revision catalog mutations remain effect-free; cancellation is an out-of-band signal that can interrupt the serialized mutation currently awaiting a hook; every live or retired effect owner retains its revision identity until cleanup succeeds. The serialized queue can then remain the state-commit authority without also blocking cancellation or erasing provenance.

Review performed with three Codex reviewer agents and DeepSeek V4 Flash as advisory tools; I verified the findings against the latest head and current `main`.

`candidate/current` 分离、依赖顺序、effect 逆序清理和 scope 隔离构成了扎实的 lifecycle 核心。但三个 seam 仍违背 kernel 自己的 ownership contract：cancellation 排在它必须取消的 hook 之后，catalog install 会执行无关 extension code，retired effects 又丢失 revision identity。

更简单的第一性原理模型应分开三个权威：revision catalog mutation 保持 effect-free；cancellation 是可中断当前等待 hook 的队列外信号；每个 live/retired effect owner 在清理成功前保留 revision identity。这样 serialized queue 可以继续作为 state commit 权威，而不会同时阻塞取消或抹掉 provenance。

本次审查使用了三位 Codex reviewer agents 与 DeepSeek V4 Flash 作为辅助工具；我已依据最新 head 和当前 `main` 复核问题。

</details>

xxhZs:
Closing in favor of the newer extension composition/kernel PRs.

PR #2983, opened 2026-08-14
chore(deps): upgrade @astryxdesign packages to 0.4.0

Astro-Han:
## Summary

Bump `@astryxdesign/core`, `@astryxdesign/theme-neutral` and `@astryxdesign/cli` from 0.3.0 to 0.4.0, and rebuild the core patch against the new source.

None of the three 0.4.0 breaking changes reach this repo. The DropdownMenu refactor keeps the `{type: 'divider'}` rows we pass (5 sites) rendering identically, the removed `dropdown-menu-radio-dot` theme target is unused here, and nothing in the product renders `Table`, so the `useTableRowExpansion` migration does not apply.

All ten hunks of the core patch are still required — 0.4.0 ships none of them. That includes the `List` `aria-label` passthrough: `List.tsx` and its build output are byte-identical to 0.3.0 and still do not destructure `'aria-label'`, so the hunk stays.

Only `useStreamingText` needed real work. 0.4.0 rewrote grapheme handling around `Intl.Segmenter` (new exported `snapToGraphemeBoundary`; the hook signature is unchanged), so the patch's `settledText` presentation state is merged into that implementation rather than replacing it:

- The reveal still advances *past* a surrogate pair (`advanceTextOffset`) instead of holding the code point back a frame. That is what `packages/ui/src/__tests__/streaming-text.test.tsx` pins, and it is the behavior the product has shipped.
- 0.4.0's snap-back then runs on top of it, so ZWJ sequences and flag pairs — which the old hunk could still split — no longer render a partial glyph.

Net effect for us: streaming output gets strictly better Unicode handling than either 0.3.0 + our patch or stock 0.4.0.

## Verification

- `node scripts/apply-dependency-patches.mjs` — clean, and clean again on a second run (the postinstall uses `--error-on-fail`, so a non-idempotent patch would break every install).
- `node node_modules/typescript/bin/tsc --build tsconfig.lib.json` — pass.
- `@maka/ui`: `streaming-text.test.tsx` (2 tests, includes the emoji boundary case) and `markdown-body.test.ts` (15 tests, includes the `settledText` restore cases) — pass.
- `npm run build` — full build passes.
- `npm run format:check` — clean.
- `apps/desktop` Playwright e2e — 22 passed, 1 skipped. Covers the surfaces most exposed to 0.4.0's DOM changes: Selector-driven settings, chat streaming (`streaming-remount`), quote selection, slash-command menu.
- DOM/theming regression check for the wrapper removals (#4775) and the Selector indicator restructure (#4838, #4846): the set of `astryx-*` class names emitted by `dist/` is compared between 0.3.0 and 0.4.0 — **no class disappears**. The two hooks this repo styles against, `.astryx-selector` / `.astryx-multi-selector` (`apps/desktop/src/renderer/styles/native-cursor.css`) and `.astryx-switch`, both survive.

Not run: the repository-wide test suite, left to CI.

CI is green (12/12). Two follow-up commits on this branch fix generated-artifact gates that the bump invalidates, each with its reasoning in the commit message: `11020338b` regenerates the Astryx theme, `f4d2669b0` records the 0.4.0 copyright override and regenerates the notices.

One `package` failure on the way there was environmental, not ours: the job smokes our packaged 0.1.10 renderer (passed, "packaged app verified") and then re-smokes the downloaded 0.1.9 upgrade baseline, and it was the 0.1.9 binary — which carries none of this diff — that missed the 30s CDP window on a slow runner. A rerun passed unchanged.

### Review focus

The merged `useStreamingText` hunk in `patches/@astryxdesign+core+0.4.0.patch` is the only non-mechanical part of this PR. The question worth checking is the interaction between `advanceTextOffset` (advance forward past a pair) and `snapToGraphemeBoundary` (snap back to a cluster start) — they pull in opposite directions by design, and the comment in the patch says why.

### Follow-ups

0.4.0 opens some cleanups that are deliberately **not** in this PR, since they are product changes and this branch is a dependency bump:

- Replace the positional selectors in `apps/desktop/src/renderer/styles/chat-message.css` (`> span:nth-child(2)`, `> span:last-child`) with the theme targets 0.4.0 puts on the icons themselves. Needs before/after visual evidence first — no test covers that CSS.
- `MoreMenu` now forwards `placement`/`alignment`; menu rows accept `variant: 'destructive'` and a stable `id`.

## Checklist

- [ ] No


Astro-Han:
Scoping note for review — most of this PR is mechanical and CI-backed; two parts need human judgment.

**1. The merged `useStreamingText` hunk in `patches/@astryxdesign+core+0.4.0.patch`.**
0.4.0 rewrote grapheme handling around `Intl.Segmenter` (new `snapToGraphemeBoundary`), so our `settledText` presentation state had to be merged into the new implementation rather than reapplied. The part worth checking is that two mechanisms deliberately pull in opposite directions: `advanceTextOffset` moves the reveal *forward* past a surrogate pair, while `snapToGraphemeBoundary` snaps the rendered slice *back* to a cluster start. Dropping the former makes `packages/ui/src/__tests__/streaming-text.test.tsx` fail (the emoji case expects the glyph on the tick that reaches it); keeping both is what also covers ZWJ and flag sequences, which our 0.3.0 hunk could still split.

**2. The copyright override in `f4d2669b0`.**
The published `@astryxdesign/core` tarball ships no LICENSE, and the override table is keyed by exact version precisely so a new release gets re-verified. The evidence I used: `@astryxdesign/theme-neutral` and `@astryxdesign/cli` at 0.4.0 both ship a LICENSE reading "MIT License / Copyright (c) 2026 Meta Platforms, Inc.", and core still declares MIT. Please confirm you accept the sibling-package LICENSE as the basis for core's attribution — that judgment is the reason this PR isn't eligible for the fast path.

The remaining commits are a version bump, a regenerated theme artifact, and regenerated notices; the three 0.4.0 breaking changes miss this repo, as detailed in the PR body. CI is green 12/12.

Disclosure: this change was prepared with Claude Code; the patch merge and the license verification above are AI-authored and are exactly what I'm asking you to check independently.

M4n5ter:
<details open>
<summary><strong>English review</strong></summary>

**Conclusion: Approve.** I did not find any blocking or actionable correctness/design issues.

I reviewed the dependency upgrade, regenerated theme artifacts, third-party notice override, and the rebuilt `@astryxdesign/core` patch.

Validation performed:

- Installed the exact 0.4.0 packages in a fresh worktree; all dependency patches applied successfully.
- Reapplied the Astryx patch twice to verify that the patching workflow remains stable.
- Confirmed the repository does not rely on the incompatible APIs removed or redesigned in Astryx 0.4.0.
- Ran the full workspace build successfully.
- Ran all `@maka/ui` tests: **143/143 passed**.
- Ran the formatting, generated-theme, and third-party-notice consistency checks successfully.
- Exercised the patched streaming-text hook under Strict Mode, mid-stream target rewrites, surrogate pairs, ZWJ emoji, and flag graphemes; all probes passed.
- Confirmed that the patch keeps the published source, distribution files, and type declarations aligned.

The exact-version license override is also supported by the published package evidence: `@astryxdesign/core@0.4.0` declares MIT but omits a LICENSE file from its tarball, while sibling Astryx 0.4.0 packages ship the corresponding Meta Platforms MIT license text.

The main residual risk is a subtle visual change caused by the upstream DOM/theme update. The successful build, Storybook/E2E CI checks, regenerated-theme verification, and class-surface comparison reduce that risk to an acceptable level.

> **AI disclosure:** This review was prepared with AI assistance. I personally reviewed the analysis and supporting evidence, approve the final assessment above, and take responsibility for the review.

</details>

<details>
<summary><strong>简体中文评审</strong></summary>

**结论：Approve。** 我没有发现需要阻止合并或要求修改的 correctness / design 问题。

我审查了依赖升级、重新生成的主题产物、第三方许可证声明以及重建的 `@astryxdesign/core` 补丁。

完成的验证包括：

- 在全新 worktree 中安装了准确的 0.4.0 版本，所有依赖补丁均成功应用。
- 额外重复应用两次 Astryx 补丁，确认补丁流程保持稳定。
- 确认仓库没有依赖 Astryx 0.4.0 中已移除或不兼容重构的 API。
- 完整 workspace build 通过。
- `@maka/ui` 测试 **143/143 全部通过**。
- 格式、生成主题和第三方许可证声明的一致性检查均通过。
- 验证了补丁中的流式文本 hook 在 Strict Mode、流式过程中目标文本改写、代理对、ZWJ emoji 和旗帜 grapheme 等场景下的行为，结果均正常。
- 确认补丁中的发布源码、构建产物和类型声明保持一致。

精确版本的许可证 override 也有发布包证据支持：`@astryxdesign/core@0.4.0` 声明使用 MIT，但发布 tarball 中遗漏了 LICENSE；同版本的其他 Astryx packages 则包含相应的 Meta Platforms MIT 许可证文本。

主要剩余风险是上游 DOM/theme 更新可能引入非常细微的视觉差异。完整构建、Storybook/E2E CI、生成主题校验和 class surface 对比已将该风险降低到可接受范围。

> **AI 使用披露：** 本评审在 AI 辅助下完成；以上分析及其支撑证据已经由我本人审核，我认可并对最终评审结论负责。

</details>

M4n5ter (approved):

Issue #2984, opened 2026-08-14
Proposal: rebuild the task rail's information architecture


Astro-Han:
## Problem

The session rail has been extended ~20 times without ever being re-laid-out, and the layers have stopped agreeing with each other.

- The top region draws three semantics as one row shape: commands (New task, Import), a filter state (Conversations, Archived), and module pages (Extensions, Automations).
- "Conversations" is tautological — the list below it *is* it.
- Three heading levels in a 260px column, plus a ➕ on the group header sharing a handler and an outcome with the top-level New task row.
- Five things in the trailing slot (stale pill, worktree icon, status dot, always-mounted ⋯ menu, spacer) compress titles to ellipsis at default width; the stale pill alone takes ~60px.
- Status signalling collapses twice — `SessionStatus` (8) → `SessionStatusTone` (7) → `StatusDotVariant` (5) — and `waiting_for_user` and `blocked` both land on `warning`.
- The row's "running" dot reads a renderer-local streaming set, then the persisted `status`. Neither is the authority: `runningTurnIds` is, and `settledSessionTransientIds` already reads it first over the same session list.
- `status: 'archived'` duplicates `isArchived` — `SessionStore.archive()` writes both — so 22 production files guard with `header.isArchived || header.status === 'archived'`.
- Dead in the same layer, since the runtime writes only `running` / `blocked` / `aborted` / `waiting_for_user`: the `done` / `review` statuses, the `flagged` session filter, `SessionStatusPresentation.interactive`, and a doc comment naming two consumers that do not exist.
- User-visible copy uses 任务 / 会话 / 对话 interchangeably, sometimes within one line of the copy table.
- Archived tasks sit in the rail as a filter row, but the real need is cleanup — batch restore and delete, which a 260px navigator cannot host.

## Proposed direction

**Terminology.** Standardise user-visible copy on 任务 / Task; code, protocol, and storage keep `session`, since Runtime Host owns Session semantics. One-way: unifying on 对话 would force 定时对话 and 描述对话. Nouns naming the entity change; 对话 stays where it names the act of talking to Maka.

**Rail.** Three flat rows on top (New task / Extensions / Automations — renamed from 定时任务, which named the parent hub after one of its child modules). The grouping switch becomes a full-width two-segment control with text labels. The list title, the group ➕, and the Conversations row are removed.

[...]

PR #2985, opened 2026-08-14
feat(desktop): add an archived tasks page to settings

Astro-Han:
## Summary

Archived tasks lived in the rail as a filter row. The rail is a navigator for active work — single selection, 260px, always on screen — and a filter row is the wrong shape for cleanup: you could restore or permanently delete one task from its menu, but you could not find a task among them, see which project it belonged to, or clear the archive out.

This moves them. 设置 › 活动 › 已归档任务 lists the archived tasks with their project and last activity, searches them by name or project, and clears a set of them in one pass. The rail's 已归档 row is gone, so there is one home and one set of actions rather than two.

image

**The carrier is the one this repo already uses for entity lists.** Projects, the permission centre, the provider catalog, the provider panel, Runtime Host and 关于 all render a `SettingsSection` over a `List`/`ListItem` group. The page went through several other shapes first — a bulk-selection `Toolbar`, an Astryx `Table` with `useTableSelection`, per-project section headers via `useTableGroupedRows` — and each one was visibly foreign inside the settings surface, because a table header, column rules and a checkbox column are not the language the other six pages speak.

**The page owns no session state.** It renders the shell's catalog and writes through the shell's row actions:

- Rows come from `deriveSessionRail`, the projection the rail itself uses, rather than a second one. No active session id is passed: the rail supplies one to highlight a row, and it also pins a family's representative to whichever revision is open, which would move a row's name, date and position here for a reason this page never shows. This matters beyond tidiness. Deleting an archived parent does not cascade to an ordinary subagent child — the retirement family is the revision family plus Agent Graph children only — so the child outlives it. The rail keeps such an orphan as a row of its own; an earlier version of this page dropped every linked child unconditionally, which left a task that `清空全部` reported clearing and could never actually reach. The Storybook fixture reproduces exactly that sequence: clear the four rows and a fifth appears.
- Restoring and deleting a single task call `unarchiveSession` / `deleteSession`, so there is one confirm, one activeId and renderer-state cleanup, and one set of toasts per operation instead of two copies that could drift apart.
- A sweep is the one thing single-row delete cannot phrase, so `purgeSessions` joins the others at the same owner. It touches only tasks still archived when it reaches them and only ids without a row action in flight, so a confirm's promise cannot widen and a sweep cannot race the rail. A rejection is not evidence a task survived — the delete IPC commits the removal before it releases renderer resources — so the rejected ids, and only those, are checked back against a fresh listing. When that listing fails the outcome says it could not verify rather than claiming one, and the first rejection travels with it so the toast can name a reason instead of a count.
- Every mutation still passes `revisionFamily: true` — a task and its revisions archive, restore and delete as one unit, never half a family.

**Archived tasks leave the working surfaces entirely.** The `SessionFilter` union loses `archived`, and with it the rail row, its copy, the nav-filter branch, and the stored-selection validator branch — a stored `archived` now fails validation and falls back to `chats`, which is the migration.

Search follows. It was the last surface that could still target an archived task, and the command palette already skipped them (`command-palette-commands.ts`), so the two ways to find a task disagreed about the same set. A hit you can only open from Settings is not a chat hit. Archived tasks also stop consuming the 200-session scan budget they shared with active ones.

That leaves nothing that needs to land on an archived task, so `openSessionInChat` returns to a plain `chats` selection, and the page offers no 打开. An archived task has no rail row to land on; giving it one would make "the open task is always visible in the rail" an invariant the rail does not otherwise hold. Restore first — which is one click, in the row.

**There is no multi-select.** The middle case it would serve — restore five of eleven — is rare enough that it does not pay for a checkbox column, a select-all cell, and a batch strip that pushes the list down the moment a row is ticked. Search plus a scoped clear covers the case that actually comes up.

Smaller decisions worth naming:

- While a search is on screen the destructive button deletes what is on screen and the confirm says so. The set is frozen at the click: a confirm names a number to a person, and one re-read afterwards could be *larger* than what they agreed to. Shrinking is safe and happens in `purgeSessions`, which skips anything no longer archived.
- The section id is `archived-tasks`, not `tasks`. It is a public `maka://settings/<section>` deep link, and the command palette was showing it as 任务 while the settings nav said 已归档任务.
- A row that cannot resolve its project says nothing rather than claiming 无项目, which was a false statement whenever the lookup failed rather than a real one about the task.
- Each row's 恢复 button carries its own accessible name — the visible text is identical on every row — and the search box no longer matches across the seam between a task's name and its project.
- Three dead `revisionFamilySessionIds` computations removed from `app-shell-session-row-actions.ts`; `flagSession`, `unarchiveSession` and `renameSession` each built a family id list and never read it.

Refs #2984. That proposal split delivery into two PRs, with the rail rebuild removing the 已归档 and 导入 rows. Removing a row and providing its replacement are one intent, not two — a PR that only adds the page leaves the same tasks with two homes and no acceptance criterion — so the archived half moves here. The rail rebuild keeps terminology, structure and signal convergence, and whatever is decided for 导入.

## Verification

- `apps/desktop` typecheck: main, renderer, preload and storybook projects all clean; `packages/ui` clean.
- `npx biome check` clean across `apps/desktop/src`, `apps/desktop/stories`, `packages/ui/src` and `packages/ui/stories`.
- `node --test dist/main/__tests__/task-catalog-rows.test.js` — 7 pass, down from 9 and stronger. The four projection tests are one: a single fixture holding a revision family, a linked subagent under a listed parent, **an orphaned one that must be listed exactly as the rail lists it**, and an active task, in an order deliberately *not* recency order. Mutation-checked — it kills every mutant the four killed (no rail projection, no revision folding, linked children listed, orphans hidden, no archived filter) plus the reordering mutant all four missed, because their fixture's store order happened to equal recency order. Five cover `matchesArchivedTaskQuery`, including that a query never matches across the name/project seam and that an unresolved project falls back to the name.
- `node --test dist/main/__tests__/app-shell-session-purge.test.js` — 6 pass, covering the sweep's semantics: whole revision families, tasks no longer archived when it reaches them, **a task restored from another window while the sweep was already running**, ids with a row action in flight, rejections settled against a fresh listing, and an unreadable listing reported as unverified. Mutation-checked: nine mutants, all killed — including reinstating the pre-loop snapshot of archived ids, which is the defect @M4n5ter found.
- `node --test dist/main/__tests__/thread-search.test.js` — 12 pass, including that an archived task is excluded alongside a fake-backend one.
- Storybook `Product/Settings/Pages → Archived Tasks` drives the real page against a fixture chosen to exercise the projection — a revision family, a linked subagent with its parent present, one whose parent is gone, a task in no project, and an active task the page must drop. Its restore, delete and clear really mutate the fixture and expand revision families, so the story cannot show a list the app would not produce. Verified by hand when the page landed: 7 fixture sessions render 4 rows; searching a project name relabels the button to 删除这 2 条 and the confirm to match; clearing all four surfaces the orphaned child, which a second clear then removes; checked in zh and en.
- **Not re-run by hand after the rail change:** Storybook. The rail row removal is covered by the storybook build and visual smoke in CI, and the dead `ArchivedConversations` story was deleted with the filter it demonstrated. Also not run: the full repository suite, and the desktop e2e journeys — this page has no e2e coverage yet.

## Review focus

Whether `purgeSessions` belongs on `createAppShellSessionRowActions`. It is not a row action, and the alternative was leaving the loop in the page. Putting it there is what lets the page hold no `window.maka.sessions` call at all, and what lets it reuse the family cleanup and the in-flight registry that single-row delete already has.

Second: the sweep re-reads the catalog as it reaches each task rather than trusting a snapshot taken at the start, after @M4n5ter showed that a task restored mid-sweep was deleted anyway. That narrows the window; it does not close it. Nothing on this side can stop a restore that lands between the check and the removal — only the Host, which owns the lifecycle, can require the task to still be archived as it removes it. Left out of this PR deliberately: `sessions:remove` gaining a lifecycle precondition is a public contract change and wants its own decision. Tracked in #3050.

Third: excluding archived tasks from search is a user-visible loss for anyone who searched content inside them. The argument for it is that the command palette already did this, that the result had nowhere to land once the rail row went, and that archiving is a statement the task is out of the working set. Restoring a task from Settings puts it back in search.

Fourth: clearing a large archive is N serial `sessions:remove` calls, each broadcasting and each provoking a full catalog re-read, with no progress and no cancel. Correct but not cheap. Left as is deliberately — an archive is usually small, and the cost is a property of the existing broadcast rather than of this page.

Fifth: this PR does not make an ordinary subagent child retire with its parent. That is Host behaviour, it predates this branch, and single-row delete in the rail has always had it. The page now surfaces the orphan so it is at least reachable, but the underlying cascade is untouched and wants its own issue.

## Checklist

- [ ] No


M4n5ter (commented):
<details>
<summary>English</summary>

Thanks for the thoughtful implementation. The overall direction looks sound, and I found no blocking issues. I did confirm one important correctness gap in the bulk purge path:

`purgeSessions()` snapshots the archived IDs once before processing the list. If another client restores a later task while the serial sweep is running, that task remains in the snapshot and can still be permanently deleted.

I reproduced this deterministically by restoring the second task after the first removal completed; both tasks were still passed to `remove`, and the outcome reported both as removed.

The trigger requires concurrent activity, so I would not classify it as blocking. However, because this is a permanent-deletion boundary and contradicts the intended “skip tasks no longer archived” behavior, I recommend addressing it here. Ideally, the Runtime Host should atomically require the task to remain archived when removing it; rechecking only the Renderer snapshot would still leave a TOCTOU window.

Everything else I reviewed—the task projection, revision-family behavior, orphan accessibility, navigation migration, and search exclusion—looks consistent with this PR’s stated intent.

</details>

<details>
<summary>中文</summary>

感谢这次细致的实现。整体方向合理，我没有发现阻塞性问题，但确认了批量清理路径中的一个重要 correctness 缺口：

`purgeSessions()` 在开始处理前只对归档 ID 做了一次快照。如果串行清理期间另一个客户端恢复了后面的某个任务，该任务仍存在于旧快照中，之后仍可能被永久删除。

我做了确定性复现：第一个任务删除完成后恢复第二个任务，最终两个任务仍都进入了 `remove`，结果也将两者都报告为已删除。

该问题需要并发操作才能触发，因此我不把它列为 blocking。不过它处于永久删除边界，并且违背了“跳过已经不再归档的任务”这一预期行为，建议在本 PR 中处理。理想方案是由 Runtime Host 在删除时原子校验任务仍处于归档状态；仅重新读取 Renderer 快照仍会留下 TOCTOU 窗口。

除此之外，我检查的任务投影、revision family 行为、orphan 可达性、导航迁移和搜索排除，都与本 PR 的目标一致。

</details>

Astro-Han:
Confirmed and fixed in `ad7499350`. You're right on both counts — the snapshot, and that rechecking in the renderer only narrows the window.

The sweep now reads the catalog as it reaches each task rather than trusting one taken at the start. That also matches what the bridge's contract already claimed ("still archived when the sweep reaches them") and removes more code than it adds. A regression test reproduces your scenario — the second task is restored after the first removal completes, and it survives — and I mutation-checked it by reinstating the snapshot, which fails the test.

The atomic guarantee belongs to the Host, as you say. I left it out of this PR: `sessions:remove` gaining a lifecycle precondition is a public contract change and deserves its own decision rather than riding along in a UI change. Tracked in #3050.

Astro-Han:
Confirmed, and tracked as #3050. Two corrections to the diagnosis, and both make the fix smaller than the one recommended.

**The snapshot is already gone.** The merged version re-reads the catalog as the sweep reaches each task rather than trusting one snapshot taken at the start. That narrows the window; your actual point — that re-reading on the Renderer side cannot *close* it — stands, and it is why the issue exists.

**The Host does already require it atomically — through revision, not lifecycle.** `#remove` compares `target.revision !== input.expectedRevision` inside the `#withStableFamily` admission lock, and `removeVersioned` re-checks `metadataVersion` in the same SQL transaction as the `DELETE`. Unarchiving goes through `setLifecycleVersioned` → `updateHeaderSync`, which bumps that version. So a concurrent restore *does* make the first remove fail with `revision_conflict`.

The defect is the Desktop client's retry loop. `removeSession` in `apps/desktop/src/main/runtime-host-client.ts` treats the conflict as a stale read and replays the delete with the fresh revision, up to `MAX_SESSION_REVISION_ATTEMPTS` (8) times. Replay is correct for rename and configuration writes — the write means the same thing at any revision. For remove it is not: the conflict is precisely the signal that the task was touched after the caller decided to destroy it, and the replay carries out a destruction whose premise no longer holds. The loop also widens the race window from "between two calls" to "the whole retry".

That makes an expected-lifecycle field on `session.remove` unnecessary. Because lifecycle writes always bump `metadataVersion`, and the version check and the `DELETE` share one transaction, "read as archived at revision R" plus "removed with `expectedRevision` R" implies "still archived when it was deleted". The loop only has to re-assert the precondition on each re-read and stop replaying blindly — no protocol change, no change to the retirement coordinator.

The fix therefore lands in `apps/desktop`. #3050 has been rewritten with this analysis; the version you saw proposed the Host-side precondition and was wrong about where the atomicity was missing.

---

*Disclosure: drafted with Claude Code, which traced the code paths cited above. The reasoning here comes from reading source — `#remove`, `removeVersioned`, `setLifecycleVersioned`, `removeSession` — not from running a reproduction; your deterministic repro remains the only execution evidence on this. Reviewed and sent by @Astro-Han.*

问题确认，已用 #3050 跟踪。对诊断有两点修正，两点都让修复范围比建议的更小。

**快照那部分已经不存在了。** 合并版是在清理推进到每个任务时重读目录，而不是信任开始时的一次快照。这收窄了窗口；你真正的论点——渲染进程侧重读无法*关闭*这个窗口——成立，这也是 issue 存在的原因。

**Host 其实已经原子地要求了，只是靠 revision 而不是 lifecycle。** `#remove` 在 `#withStableFamily` 的准入锁内比对 `target.revision !== input.expectedRevision`，`removeVersioned` 在与 `DELETE` 同一个 SQL 事务里再次校验 `metadataVersion`。而取消归档走 `setLifecycleVersioned` → `updateHeaderSync`，会 bump 这个版本号。所以并发恢复**确实**会让第一次删除以 `revision_conflict` 失败。

缺陷在桌面端客户端的重试循环。`apps/desktop/src/main/runtime-host-client.ts` 的 `removeSession` 把冲突当成读取过期，用新的 revision 重放删除，最多 `MAX_SESSION_REVISION_ATTEMPTS`（8）次。重放对改名和改配置是对的——这些写在任何 revision 下含义相同。对删除不是：冲突恰恰是「你决定销毁它之后有人动过它」的信号，重放执行的是一个前提已失效的销毁操作。这个循环还把竞态窗口从「两次调用之间」放大成「整个重试过程」。

这样一来，`session.remove` 上的 expected-lifecycle 字段就没有必要了。因为 lifecycle 写入必然 bump `metadataVersion`，而版本校验与 `DELETE` 在同一事务内，「在 revision R 读到它处于归档态」加上「以 `expectedRevision` R 删除成功」即可推出「删除发生时它仍处于归档态」。循环只需要在每轮重读后重新确认这个前置条件、不再盲目重放——不用改协议，也不用动 retirement coordinator。

因此修复落在 `apps/desktop`。#3050 已按这个分析重写；你看到的那一版提出的是 Host 侧前置条件，对「原子性缺在哪里」的判断是错的。

---

*披露：本回复由 Claude Code 起草，上述代码路径由它追溯。结论来自阅读源码（`#remove`、`removeVersioned`、`setLifecycleVersioned`、`removeSession`），没有运行复现；这件事上唯一的执行证据仍然是你的确定性复现。由 @Astro-Han 审阅后发出。*

</details>

PR #2989, opened 2026-08-14
fix(mcp): bound repeated tool rediscovery

### LINKED ISSUE #2981: mcp: tool discovery livelocks against a server that notifies list_changed after every tools/list
## Context

A server that emits `tools/list_changed` after every `tools/list` response livelocks tool discovery. Each notification schedules another discovery pass, whose response triggers the next notification, with no bound on the loop.

Measured with a probe server that notifies on every list: **~2000 `tools/list` round trips in 4 seconds**, `sync()` never settling, and the manager unable to remove that server from the configuration — the removal sync queues behind a discovery that never finishes.

This is **pre-existing**, not a regression. I found it while reviewing #2663 and initially took it for a regression from installing the change handler before initial discovery; it is not. The same probe against `origin/main` produces the same ~2000-iteration loop. Noted in [that review](https://github.com/maka-agent/maka-agent/pull/2663#pullrequestreview-4930617410) as out of scope there.

## Severity

**P2.** Unlike most items in this class, something on `main` genuinely misbehaves rather than merely lacking protection: discovery hangs and configuration reconciliation stalls behind it. Held at P2 because triggering requires a server that notifies on every list — real servers notify on actual change — and it self-recovers the moment the server stops notifying. It has been in place a long time with no report.

## Fix

Bound the rediscovery loop. Options, roughly in order of preference:

1. Debounce `list_changed` — collapse notifications that arrive while a discovery for the same server is in flight or within a short window after it completes.
2. Cap consecutive rediscoveries per connection generation, then stop honouring further notifications and surface the server as degraded.
3. At minimum, let a pending configuration sync preempt an in-flight discovery, so a misbehaving server can always be removed.

(3) is worth having regardless of (1) and (2): the manager should never be unable to drop a server.

## Verification

A probe server whose `tools/list` handler schedules `sendToolListChanged()` on every call. Assert that list round trips stay bounded, that `sync()` settles, and that a follow-up `sync()` removing the server completes.

XonkelX:
## Summary

Bound consecutive MCP Tool rediscovery attempts when a server emits `tools/list_changed` for every `tools/list`. The manager now stops after three superseded attempts, keeps the previous callable snapshot, reports the hostile notification pattern, and remains removable through config sync.

Fixes #2981

## Verification

- `npm --workspace @maka/mcp test` — 58 passed
- `npm run lint`
- `npm run format:check`
- `npm run check:third-party-notices`
- `npx biome check packages/mcp/src/index.ts packages/mcp/src/__tests__/manager.test.ts`
- Root `npm run build` reaches an unrelated existing `packages/runtime-host/src/server/session-transcript-pager.ts:249` TS7006 error on `main`; the affected MCP workspace builds and typechecks in its test command.

## Checklist

- [x] Yes — repeated hostile discovery notifications now produce a bounded refresh error instead of an infinite loop
- [ ] No

Generated with Codex; reviewed and submitted by the human contributor of record.

### DIFF
--- a/packages/mcp/src/__tests__/manager.test.ts
+++ b/packages/mcp/src/__tests__/manager.test.ts
@@ -267,11 +267,10 @@ describe('McpClientManager E2E', { concurrency: false }, () => {
       assert.notEqual(bindingFor(manager, 'first', 'echo'), bindingFor(manager, 'second', 'echo'));
     });
 
-    test('coalesces a refresh signal and publishes only the final attempt', async () => {
+    test('coalesces a refresh signal and keeps the latest valid snapshot', async () => {
       const fixture = await createRemoteFixture('streamable-http');
       const manager = createManager();
       await manager.sync(remoteConfig(fixture.url));
-      const before = manager.status('remote')?.tools;
       const listsBefore = countProtocolMethod(fixture, 'tools/list');
 
       fixture.setToolListMode('replacement');
@@ -290,7 +289,10 @@ describe('McpClientManager E2E', { concurrency: false }, () => {
       for (const result of settled) {
         if (result.status === 'rejected') assert.match(String(result.reason), /duplicate tool/u);
       }
-      assert.deepEqual(manager.status('remote')?.tools, before);
+      assert.deepEqual(
+        manager.status('remote')?.tools.map((tool) => tool.name),
+        ['replacement'],
+      );
       assert.equal(countProtocolMethod(fixture, 'tools/list') - listsBefore, 2);
     });
 
@@ -322,7 +324,7 @@ describe('McpClientManager E2E', { concurrency: false }, () => {
       let now = 0;
       const manager = createManager({ now: () => now });
       await manager.sync(remoteConfig(`${fixture.url}/sse`, 'auto'));
-      fixture.setToolListClockAdvance(() => {
+      fixture.setToolListResponseClockAdvance(() => {
         now += 1_100;
       });
       fixture.setNotifyOnEveryToolList(true);
@@ -335,6 +337,23 @@ describe('McpClientManager E2E', { concurrency: false }, () => {
 
       assert.equal(countProtocolMethod(fixture, 'tools/list') <= 4, true);
       assert.match(manager.status('remote')?.error ?? '', /changed too frequently/u);
+    });
+
+    test('publishes the latest valid snapshot before suppressing rediscovery', async () => {
+      const fixture = await createRemoteFixture('sse');
+      const manager = createManager();
+      await manager.sync(remoteConfig(`${fixture.url}/sse`, 'auto'));
+      fixture.setToolListMode('replacement');
+      fixture.setNotifyOnEveryToolList(true);
+
+      await fixture.notifyToolListChanged();
+      await waitFor(() => /changed too frequently/u.test(manager.status('remote')?.error ?? ''));
+
+      assert.equal(countProtocolMethod(fixture, 'tools/list'), 4);
+      assert.deepEqual(
+        manager.status('remote')?.tools.map((tool) => tool.name),
+        ['replacement'],
+      );
     });
 
     test('coalesces a burst of legitimate list-changed notifications', async () => {
@@ -851,6 +870,7 @@ interface RemoteFixture {
   holdNextToolList(): { started: Promise<void>; release(): void };
   setNotifyOnEveryToolList(enabled: boolean): void;
   setToolListClockAdvance(advance: () => void): void;
+  setToolListResponseClockAdvance(advance: () => void): void;
   notifyToolListChanged(): Promise<void>;
   close(): Promise<void>;
 }
@@ -874,7 +894,16 @@ async function createRemoteFixture(
   let nextToolListGate: InternalToolListGate | undefined;
   let notifyOnEveryToolList = false;
   let toolListClockAdvance = () => {};
+  let toolListResponseClockAdvance = () => {};
   const sseTransports = new Map<string, { transport: SSEServerTransport; server: McpServer }>();
+  const beforeToolList = async () => {
+    toolListClockAdvance();
+    const gate = nextToolListGate;
+    if (!gate) return;
+    nextToolListGate = undefined;
+    gate.markStarted();
+    await gate.waitForRelease;
+  };
   const httpServer = createServer(async (req, res) => {
     const url = new URL(req.url ?? '/', 'http://127.0.0.1');
     const request: RemoteRequest = {
@@ -914,14 +943,8 @@ async function createRemoteFixture(
         const server = createProtocolServer({
           advertiseTools: true,
           toolListMode: () => toolListMode,
-          beforeToolList: async () => {
-            toolListClockAdvance();
-            const gate = nextToolListGate;
-            if (!gate) return;
-            nextToolListGate = undefined;
-            gate.markStarted();
-            await gate.waitForRelease;
-          },
+          beforeToolList,
+          afterToolList: () => toolListResponseClockAdvance(),
           notifyOnEveryToolList: () => notifyOnEveryToolList,
         });
         await server.connect(transport);
@@ -937,14 +960,8 @@ async function createRemoteFixture(
         const server = createProtocolServer({
           advertiseTools: options.advertiseTools !== false,
           toolListMode: () => toolListMode,
-          beforeToolList: async () => {
-            toolListClockAdvance();
-            const gate = nextToolListGate;
-            if (!gate) return;
-            nextToolListGate = undefined;
-            gate.markStarted();
-            await gate.waitForRelease;
-          },
+          beforeToolList,
+          afterToolList: () => toolListResponseClockAdvance(),
           notifyOnEveryToolList: () => notifyOnEveryToolList,
         });
         sseTransports.set(transport.sessionId, { transport, server });
@@ -1005,6 +1022,9 @@ async function createRemoteFixture(
     setToolListClockAdvance: (advance) => {
       toolListClockAdvance = advance;
     },
+    setToolListResponseClockAdvance: (advance) => {
+      toolListResponseClockAdvance = advance;
+    },
     notifyToolListChanged: async () => {
       const servers = [...sseTransports.values()].map(({ server }) => server);
       if (servers.length === 0) throw new Error('no persistent MCP server is connected');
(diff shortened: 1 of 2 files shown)


M4n5ter (commented):
Blocking: this bounds only notifications that arrive while the same `tools/list` request is still in flight; it does not bound the response-followed-by-notification loop described in #2981.

`supersededAttempts` is local to [`refreshToolLoop()`](https://github.com/maka-agent/maka-agent/blob/909a2e86778c0da5fb7964906205eea2bcc7c797/packages/mcp/src/index.ts#L599-L654). Once a refresh completes, `refreshState` is cleared and the next notification creates a new loop with the counter reset to zero. A server that sends `list_changed` just after every `tools/list` response can therefore continue rediscovery indefinitely.

The regression fixture currently sends and awaits the notification *before* returning the response:

https://github.com/maka-agent/maka-agent/blob/909a2e86778c0da5fb7964906205eea2bcc7c797/packages/mcp/src/__tests__/manager.test.ts#L947-L954

That guarantees the favorable `state.pending === true` ordering and does not reproduce the issue's "after every response" timing. I changed only the fixture to schedule `sendToolListChanged()` after returning the response; the test then observed 127 `tools/list` calls instead of 3 in about 675 ms.

The test also enables the hostile behavior only after the initial `sync()` has settled, and tests removal only after the retry-limit error has already occurred. It therefore does not verify the two other contracts requested in #2981: initial sync settling and configuration removal preempting an active rediscovery storm.

Could we keep the retry/suppression state at the connection-generation level instead of inside one refresh promise? After reaching the limit, further notifications for that generation should remain suppressed until a defined recovery boundary, such as a quiet period, explicit refresh, or reconnect. The regression test should schedule the notification after each response and attempt removal while rediscovery is still active.

Astro-Han (commented):
Thanks for moving the suppression state to the connection-generation lifetime. That addresses the reset identified in the previous review, but the bound still uses notification timing and count rather than actual rediscovery passes.

A slow server can still bypass the bound: if each `tools/list` takes longer than the one-second quiet period and then emits `list_changed`, every notification resets `consecutiveSignals`. With a 1.2-second response delay, rediscovery continued indefinitely without producing the frequency error.

The opposite timing also causes a regression. MCP notification handlers are dispatched concurrently, so three notifications arriving during one in-flight refresh increment the counter three times. The third marks the generation as suppressed, and `refreshToolLoop()` then discards the successful definitions it already received. In a reproduction where the server changed from `echo` to `replacement`, the manager reported `changed too frequently` and permanently retained the stale `echo` snapshot.

Could we count actual rediscovery passes at the per-generation scheduler instead? Notifications received during an active request should coalesce into one pending pass, and the quiet period should begin after a refresh completes. A delayed-response regression and a three-notification burst regression would cover both sides of the invariant.

AI-assisted review disclosure: Codex coordinated the source and lifecycle analysis, and Claude Opus performed an independent adversarial review. The reported behavior was reproduced against the exact PR head; the human reviewer retains responsibility for the final review decision.

Astro-Han (commented):
Reviewed against #2981.

The 5ms notify-after-list flood does stop at three `tools/list` calls, and `disconnect` does not wait on an in-flight refresh. Removal can still settle. Keeping the previous snapshot is the right call; disconnecting would open a reconnect churn path.

Two holes remain, so I am not approving yet.

A third `list_changed` inside the 1s window is discarded without starting a refresh (`refreshToolsAfterNotification`, `consecutiveSignals >= 3`). A legitimate server that notifies three times during startup batch registration loses that last change and stays on `changed too frequently` until a >1s quiet gap or a manual `refreshTools`. `refreshTools` has no production caller.

The quiet window is measured from the last notification, not from a finished rediscovery. If `tools/list` takes longer than 1s (the list timeout is 15s) and then emits `list_changed`, `consecutiveSignals` resets every time. That is the original loop with a longer period. The new tests only pin the 5ms path.

Also: this branch currently reports no CI checks.

AI-assisted review: Grok 4.6, opencode-go/deepseek-v4-flash:max, and ark-coding-plan/glm-5.3 each reviewed the PR independently. I walked `refreshToolsAfterNotification` against a 5ms flood, a 100ms legitimate triple, and a 1100ms slow loop. Unverified by me: I did not run the MCP suite.

Astro-Han:
Thanks for tackling the rediscovery loop — the direction is right and test 1 is a genuine regression test. I ran the suite and probed two edge cases the tests don't cover; both surfaced issues worth resolving before merge.

**Conclusion: PASS with two mandatory-handling P2s (both verified by reproduction).**

**P2-1 — the bound can be bypassed by a slow hostile server; the rediscovery loop continues indefinitely.**
The counter only accrues when consecutive notifications are ≤1s apart (`now - lastSignalAt > 1_000` resets it). I reproduced: with a 1.2s delay on `tools/list` and notify-per-list enabled, `tools/list` grew to 6 calls in 6s with no suppression and no error — every notification resets the counter to 1, so it never reaches 3. The issue's pattern ("notify on every tools/list") is only bounded for fast notifiers; at one-list-per-RTT the loop persists. Consider bounding the refresh *rate* independently of notification spacing (e.g., trailing debounce), or at least documenting this as a known limit.

**P2-2 — three legitimate notifications within 1s are misclassified as hostile: spurious error and discarded refresh results.**
I reproduced against a fully normal server (no notify-per-list): three rapid `notifyToolListChanged()` calls produced `status.error = 'tool list changed too frequently during discovery'`, and none of the three valid changes was applied to the snapshot — the refreshed result is discarded by the suppressed check, and the error doesn't clear until the next notification. A legit server doing 3 quick changes (progressive tool registration, batched config load) freezes the tool list in its pre-change state with a misleading error. The count (3) and window (1s) have no protocol or engineering basis.

Optional nits (P3): the suppression error sticks until the next notification (no auto-recovery while the server is quiet); test 2 ("config removal preempts in-flight rediscovery") already passes on `main`, so the checklist claim "tests fail without the change" holds only for test 1; suppressed notifications still fire status update + emit per notification; the public `refreshTools` reset has no production callers.

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagent on `ollama-cloud/deepseek-v4-flash`, read-only). The subagent ran the PR suite (59 pass) in a throwaway worktree, ran test 1 against `main`'s manager code (confirmed red), and reproduced both P2s with probe fixtures. The debounce alternative is a suggestion, not a requirement. Please weigh these findings with your own judgment.*

<details>

结论：PASS（两个 P2 必须处理或明确延后）。两个 P2 均已实证复现：① 慢速 hostile server 可绕过 bound——计数仅当相邻通知 ≤1s 才累计，1.2s 延迟 + notify-per-list 时计数每次重置为 1、永远达不到 3，rediscovery 循环以每 RTT 一次的速率无限持续，PR 声称修复的核心场景只对快速通知成立；② 误伤——1 秒内 3 次合法通知被判 hostile，产生虚假 error 且三次有效变更全部被丢弃（工具列表冻结在变更前状态），计数 3 与窗口 1s 无任何协议或工程依据。建议改用 trailing debounce（限速而非判定 hostile，约 15 行 vs 当前 160+ 行）。P3 可选：suppression error 粘滞不自愈、test 2 在 main 上本已通过（checklist 声明不实）、suppressed 后仍逐通知 emit。
</details>

Astro-Han:
Thanks for the follow-up — the two P2s from the earlier review are addressed, and the new tests are solid. I re-verified on the updated head.

**Conclusion: the blockers are resolved. Ready to merge once CI runs green (it currently needs approval to run).**

Verified on the updated head (`e6c0b1f5`):
- **P2-1 (slow-hostile bypass) — fixed.** The quiet-period reset (`TOOL_REFRESH_QUIET_PERIOD_MS` / `lastRefreshCompletedAt`) is gone; `refreshPasses` now only accrues and resets on connection change. The updated test advances the clock *after* the `tools/list` response (1.1s spacing) and still asserts the bound — the slow-notifier variant is now covered.
- **P2-2 (discarded valid results) — fixed.** `refreshToolLoop` now applies the snapshot (`replaceToolSnapshot` + `update`) *before* the suppressed check, so the latest valid tool list is published even when rediscovery is suppressed. The new "publishes the latest valid snapshot before suppressing rediscovery" test pins exactly that.
- **Burst coalescing — covered.** "coalesces a burst of legitimate list-changed notifications" confirms notifications arriving during an in-flight refresh are merged without error.
- Ran the workspace suite locally on the updated head: **62/62 pass** (up from 59).

Remaining observations (non-blocking):
- **P3**: a legitimate server doing 4 *independently completed* refreshes (each notification arriving after the previous refresh finished) still hits the pass cap and surfaces the frequency error. The harm is now limited — the latest snapshot is applied, so the tool list stays fresh — but the error is spurious. If you want to close it fully, a trailing-debounce (rate-limit refreshes per unit time instead of counting passes) would distinguish "hostile notify-per-list" from "legit rapid changes" without a counter. Fine to defer.
- **P3**: the PR branch's CI is in `action_required` state — it needs approval before the checks run. Worth getting that approved so the suite (including the new e2e tests) is verified before merge.

Nice addition: the error-sanitization tests (no secret leakage, control-character stripping in status errors) are a good hardening beyond the original scope.

---

*AI-assisted review disclosure: this follow-up was produced with AI assistance (pi review subagent on `ollama-cloud/deepseek-v4-flash`, read-only). I verified the fixes against the updated diff, traced the pass-accounting logic, and ran the workspace test suite locally (62/62). The remaining P3 is analysis of the current semantics, not a reproduced failure. Please weigh these findings with your own judgment.*

<details>

结论：两个 P2 阻塞项已解决，可以合并（需先批准并跑通 CI）。已修复：① 慢速 hostile 绕过——删除了 quiet-period 重置逻辑，refreshPasses 只增不减，测试改为在 tools/list 响应后推进时钟（1.1s 间隔）仍断言 bound；② 有效结果被丢弃——refreshToolLoop 现在先应用 snapshot 再抛 frequency error，新测试验证 suppressed 时最新有效工具列表仍被发布；③ burst 合并——新增测试验证 in-flight 刷新期间到达的通知被合并、无 error。本地跑工作区测试 62/62 通过（比之前多 3 个）。残余（不阻塞）：合法 server 做 4 次独立完成的刷新仍会触发频率 error（后果已减轻——最新快照被应用，工具列表保持最新），如需彻底关闭可改用 trailing debounce 限速而非计数；另外分支 CI 处于 action_required 状态，需批准运行。加分项：新增的错误消息脱敏测试（不泄露 secret、控制字符清理）。
</details>

Astro-Han:
LGTM. Both P2s from review are resolved, the new tests cover the slow-notifier and snapshot-preservation cases, and all CI checks are green (test_workspaces, typecheck, test, windows_recovery, windows_baseline). Merging.

*AI-assisted review disclosure: the review was AI-assisted (pi subagent on ollama-cloud/deepseek-v4-flash); the merge decision is the human maintainer's.*

PR #2969, opened 2026-08-13
feat(runtime): support sequential agent graph epochs

me2seeks:
## Summary

- persist monotonic root-to-graph epoch bindings while preserving the existing deterministic graph as epoch 1
- advance a finished, quiescent graph before the next graph/swarm root Turn is durably admitted
- fence supervisor wakes and resolve stop, permission, revision, and retirement paths through durable graph ownership

A finished graph remains immutable. The next graph-capable Turn receives a new deterministic graph ID, and concurrent Hosts converge through a storage CAS instead of reopening or skipping the previous graph. Ordinary default Turns leave the finished graph untouched. Legacy epoch 1 stays virtual until the first real cutover, so Sessions do not acquire graph state just by being read.

This establishes the lifecycle boundary needed by #2588. Historical graph selection and cross-epoch result inputs remain follow-up work.

Refs #2588

## Verification

- `npm --workspace @maka/core run test` — 539 passed
- `npm --workspace @maka/storage run test` — 777 passed, 16 skipped
- `npm --workspace @maka/runtime run test` — 2,788 passed, 6 skipped
- `npm --workspace @maka/runtime-host run test` — 910 passed
- builds for Core, Storage, Runtime, and Runtime Host
- Biome on all changed files
- `git diff --check`

## Checklist

- [x] Tests cover the change and fail without it
- [x] Lint, format, typecheck and the affected suites pass locally

Does this PR entail a change in behavior?

- [x] Yes — a graph/swarm Turn can advance from a finished Agent Graph to a fresh graph epoch
- [ ] No

<details>
<summary>中文说明</summary>

## 摘要

- 持久化 root Session 到 Agent Graph 的单调 epoch 绑定，同时保持原有确定性 graph ID 作为 epoch 1
- 只有当前图已经结束且静默，才会在下一条 graph/swarm root Turn 持久化之前切换到新 epoch
- supervisor wake、停止、权限回答、会话修订与删除清理都通过 Runtime Host 的持久化 graph ownership 解析

已经 `finish` 的图仍然不可修改。下一条启用 Graph 的 Turn 使用新的确定性 graph ID；多个 Host 并发切换时由 Storage CAS 收敛，不会重开旧图或跳过 epoch。普通 default Turn 不会切换 graph epoch。第一次真正切换前，legacy epoch 1 只是只读兼容值，因此会话不会仅因查询而写入图状态。

本 PR 建立 #2588 所需的生命周期基础。历史图选择和跨 epoch 结果输入将在后续实现。

</details>

Issue #2588, opened 2026-08-09
Support multiple graph epochs per root Session with cross-graph result inputs


likun666661:
## Problem

An agent graph currently has the same lifetime as its root Session:

- `agentGraphIdForRootSession(rootSessionId)` deterministically maps one root Session to one graph ID.
- A `finish` schedule update permanently closes that graph.
- Later schedule updates for the same root Session fail with `Agent graph schedule is already finished`.
- There is no normal graph epoch/generation or "begin next graph" path.

This makes sense if a root Session represents one workflow execution, but Maka Sessions are long-lived conversations. A common flow is:

1. Ask the supervisor to use swarm for a research task.
2. The supervisor finishes the graph and selects its final result records.
3. Continue the same conversation with a follow-up question.
4. Ask for another swarm pass.

Step 4 cannot schedule new work because the root Session still resolves to the already-finished graph. We reproduced this in a normal conversation: after completing and finishing a six-item architecture-reading swarm, a follow-up `add_work` failed because the graph schedule was closed.

Relevant implementation points:

- `packages/runtime/src/stream-graph-coordinator.ts`: `agentGraphIdForRootSession()` hashes only `rootSessionId`.
- `packages/storage/src/sqlite-session-metadata-store.ts`: `commitAgentGraphScheduleUpdate()` rejects every update after `hasClosedAgentGraphSchedule(graphId)`.
- `packages/runtime/src/stream-graph-supervisor-tools.ts`: schedule projection rejects updates after `finish`.

The last two invariants are valuable: `finish` should remain a durable, irreversible admission boundary. The questionable part is binding that one graph's lifetime to the entire conversational Session.

There is a related data-continuity issue. A new graph should be able to consume immutable results published by a finished graph without reopening or mutating the old graph. Otherwise follow-up work must copy large outputs back into prompts, rely on lossy supervisor summaries, or repeat completed work.

## Desired outcome

A root Session can run multiple sequential graph epochs while preserving the meaning of `finish` for each individual graph.

For example:

```text
Root Session
  |- Graph epoch 1 (finished)
  |- Graph epoch 2 (finished)
  `- Graph epoch 3 (active)
```

Desired semantics:

- `finish` permanently closes the current graph epoch; it does not reopen later.
- A later supervisor turn can atomically create or bind a new graph epoch for the same root Session.
- Only one graph epoch is active for a root Session at a time, unless explicit parallel-root semantics are introduced later.
- Historical graphs remain readable and retain their selected final results.
[...]

PR #2991, opened 2026-08-14
feat(agent-graph): browse historical graph runs

me2seeks:
## Summary

- add bounded Runtime Host reads for a root Session's current and historical Agent Graph epochs
- let Desktop browse prior Graph runs while keeping historical controls read-only
- add `/graph history` to the TUI without starting an agent Turn

The Runtime Host remains the only historical graph reader: clients request an epoch directory and an exact graph snapshot, while the Host validates that the requested graph belongs to the same root Session. Desktop follows the current epoch by default and fences stale refreshes when the user switches runs.

This is the historical-read slice of #2588 and builds on #2969.

Refs #2588

Ready for review.

## Verification

- `npm run build:test`
- `npm run typecheck`
- Agent Graph, Desktop bridge/panel, and TUI tests on the latest-main stack — 146 passed
- `git diff --check`

## Checklist

- [x] Yes — Desktop and TUI can inspect earlier Agent Graph runs in a root Session
- [ ] No

<details>
<summary>中文说明</summary>

## 摘要

- 增加由 Runtime Host 提供的有界 Graph epoch 历史读取
- Desktop 可以切换查看之前的 Graph 运行，历史运行只读且不会出现停止或关闭控制
- TUI 增加 `/graph history`，查看历史不会启动 Agent Turn

客户端不会直接读取 SQLite。Runtime Host 会验证指定 graph 确实属于同一个 root Session，再返回 epoch 目录和精确历史快照。Desktop 默认跟随当前 epoch，并通过 generation fence 防止较慢的旧请求覆盖用户刚选择的新页面。

这是 #2588 的历史读取部分，建立在 #2969 的 epoch 生命周期边界之上。

已转为 Ready，请求 review。

</details>

## AI disclosure / AI 披露

This PR was implemented by Codex under me2seeks’s direction and review. / 本 PR 由 Codex 在 me2seeks 的指导与审核下完成。

## Visual evidence

Representative Desktop capture using the PR copy and graph states. The historical selection is explicitly read-only and does not expose the current-run Stop action.

| Before — current graph only | After — historical run selected |
| --- | --- |
| Agent Graph before historical browsing | Agent Graph historical run selected |\n\n## AI use\n- [ ] No generative tool was used for implementation.\n- [x] Generative tooling was used and the result was reviewed and verified by the author.\n\nTool(s) and scope: OpenAI Codex (Maka) assisted with implementation, review remediation, tests, and PR documentation.\n\nFinal squash trailer: `Generated-by: Maka`

### DIFF
--- a/apps/desktop/src/renderer/agent-graph-panel.tsx
+++ b/apps/desktop/src/renderer/agent-graph-panel.tsx
@@ -190,7 +190,7 @@ export function AgentGraphPanel(props: {
     let cachedDirectory: AgentGraphEpochDirectory | undefined;
 
     const scheduler = createAgentGraphRefreshScheduler(async (fence) => {
-      setLoading(true);
+      if (!cachedDirectory) setLoading(true);
       try {
         let directory: AgentGraphEpochDirectory;
         if (!cachedDirectory) {
--- a/apps/desktop/src/renderer/agent-graph-panel.tsx
+++ b/apps/desktop/src/renderer/agent-graph-panel.tsx
@@ -168,6 +168,7 @@ export function AgentGraphPanel(props: {
   const [error, setError] = useState(false);
   const [stopState, setStopState] = useState({
     rootSessionId: props.rootSessionId,
+    graphId: undefined as string | undefined,
     requestId: 0,
     pending: false,
     error: false,
@@ -180,9 +181,10 @@ export function AgentGraphPanel(props: {
   const followCurrentRef = useRef(true);
   const stopRequestIdRef = useRef(0);
   const copy = getAgentGraphPanelCopy(props.locale);
-  const stopPending =
-    stopState.rootSessionId === props.rootSessionId && stopState.pending;
-  const stopError = stopState.rootSessionId === props.rootSessionId && stopState.error;
+  const stopFeedbackMatchesSelection =
+    stopState.rootSessionId === props.rootSessionId && stopState.graphId === selectedGraphId;
+  const stopPending = stopFeedbackMatchesSelection && stopState.pending;
+  const stopError = stopFeedbackMatchesSelection && stopState.error;
 
   useEffect(() => {
     setSnapshot(undefined);
@@ -194,6 +196,7 @@ export function AgentGraphPanel(props: {
     setError(false);
     setStopState({
       rootSessionId: props.rootSessionId,
+      graphId: undefined,
       requestId: ++stopRequestIdRef.current,
       pending: false,
       error: false,
@@ -306,7 +309,7 @@ export function AgentGraphPanel(props: {
     if (stopPending) return;
     const rootSessionId = props.rootSessionId;
     const requestId = ++stopRequestIdRef.current;
-    setStopState({ rootSessionId, requestId, pending: true, error: false });
+    setStopState({ rootSessionId, graphId: expectedGraphId, requestId, pending: true, error: false });
     try {
       await window.maka.graphs.stop(rootSessionId, expectedGraphId);
     } catch {
--- a/apps/desktop/src/renderer/agent-graph-panel.tsx
+++ b/apps/desktop/src/renderer/agent-graph-panel.tsx
@@ -166,15 +166,23 @@ export function AgentGraphPanel(props: {
   const [selectedGraphId, setSelectedGraphId] = useState<string>();
   const [loading, setLoading] = useState(props.enabled);
   const [error, setError] = useState(false);
-  const [stopPending, setStopPending] = useState(false);
-  const [stopError, setStopError] = useState(false);
+  const [stopState, setStopState] = useState({
+    rootSessionId: props.rootSessionId,
+    requestId: 0,
+    pending: false,
+    error: false,
+  });
   const [collapsed, setCollapsed] = useState(false);
   const [dismissedBySession, setDismissedBySession] = useState<AgentGraphPanelDismissals>({});
   const contentId = useId();
   const refreshRef = useRef<AgentGraphRefreshScheduler>(noopAgentGraphRefreshScheduler);
   const selectedGraphIdRef = useRef<string | undefined>(undefined);
   const followCurrentRef = useRef(true);
+  const stopRequestIdRef = useRef(0);
   const copy = getAgentGraphPanelCopy(props.locale);
+  const stopPending =
+    stopState.rootSessionId === props.rootSessionId && stopState.pending;
+  const stopError = stopState.rootSessionId === props.rootSessionId && stopState.error;
 
   useEffect(() => {
     setSnapshot(undefined);
@@ -184,7 +192,12 @@ export function AgentGraphPanel(props: {
     selectedGraphIdRef.current = undefined;
     followCurrentRef.current = true;
     setError(false);
-    setStopError(false);
+    setStopState({
+      rootSessionId: props.rootSessionId,
+      requestId: ++stopRequestIdRef.current,
+      pending: false,
+      error: false,
+    });
     setCollapsed(false);
     setLoading(props.enabled);
     let cachedDirectory: AgentGraphEpochDirectory | undefined;
@@ -291,14 +304,23 @@ export function AgentGraphPanel(props: {
 
   const stopGraph = async (expectedGraphId: string): Promise<void> => {
     if (stopPending) return;
-    setStopPending(true);
-    setStopError(false);
+    const rootSessionId = props.rootSessionId;
+    const requestId = ++stopRequestIdRef.current;
+    setStopState({ rootSessionId, requestId, pending: true, error: false });
     try {
-      await window.maka.graphs.stop(props.rootSessionId, expectedGraphId);
+      await window.maka.graphs.stop(rootSessionId, expectedGraphId);
     } catch {
-      setStopError(true);
+      setStopState((current) =>
+        current.rootSessionId === rootSessionId && current.requestId === requestId
+          ? { ...current, error: true }
+          : current,
+      );
     } finally {
-      setStopPending(false);
+      setStopState((current) =>
+        current.rootSessionId === rootSessionId && current.requestId === requestId
+          ? { ...current, pending: false }
+          : current,
+      );
     }
   };
   const stopAvailable =
--- a/apps/desktop/src/renderer/agent-graph-panel.tsx
+++ b/apps/desktop/src/renderer/agent-graph-panel.tsx
@@ -4,6 +4,7 @@ import type {
   AgentGraphClientOperator,
   AgentGraphClientSnapshot,
 } from '@maka/runtime/stream-graph-read-model';
+import type { AgentGraphEpochDirectory } from '@maka/runtime-host/client';
 import type { AgentGraphEpochSummary } from '@maka/runtime-host/protocol';
 import { IconButton, Selector, type SelectorOptionType } from '@maka/ui';
 import { ICON_SIZE, ChevronDown, X } from '@maka/ui/icons';
@@ -186,11 +187,22 @@ export function AgentGraphPanel(props: {
     setStopError(false);
     setCollapsed(false);
     setLoading(props.enabled);
+    let cachedDirectory: AgentGraphEpochDirectory | undefined;
 
     const scheduler = createAgentGraphRefreshScheduler(async (fence) => {
       setLoading(true);
       try {
-        const directory = await window.maka.graphs.listEpochs(props.rootSessionId);
+        let directory: AgentGraphEpochDirectory;
+        if (!cachedDirectory) {
+          directory = await window.maka.graphs.listEpochs(props.rootSessionId);
+        } else {
+          const currentPage = await window.maka.graphs.listCurrentEpochs(props.rootSessionId);
+          directory = sameEpochPage(cachedDirectory, currentPage)
+            ? cachedDirectory
+            : await window.maka.graphs.listEpochs(props.rootSessionId);
+        }
+        if (!scheduler.isCurrent(fence)) return;
+        cachedDirectory = directory;
         const nextEpochs = directory.epochs;
         const current = nextEpochs.find((entry) => entry.current) ?? nextEpochs[0];
         const selected = followCurrentRef.current
@@ -277,12 +289,12 @@ export function AgentGraphPanel(props: {
     return null;
   }
 
-  const stopGraph = async (): Promise<void> => {
+  const stopGraph = async (expectedGraphId: string): Promise<void> => {
     if (stopPending) return;
     setStopPending(true);
     setStopError(false);
     try {
-      await window.maka.graphs.stop(props.rootSessionId);
+      await window.maka.graphs.stop(props.rootSessionId, expectedGraphId);
     } catch {
       setStopError(true);
     } finally {
@@ -353,7 +365,9 @@ export function AgentGraphPanel(props: {
               size="sm"
               label={stopPending ? copy.stopping : copy.stop}
               isDisabled={stopPending}
-              onClick={() => void stopGraph()}
+              onClick={() => {
+                if (snapshot) void stopGraph(snapshot.graphId);
+              }}
             />
           ) : null}
           {dismissAvailable && snapshot ? (
@@ -464,6 +478,21 @@ export function AgentGraphPanel(props: {
       ) : null}
     </section>
   );
+}
+
+function sameEpochPage(
+  cached: AgentGraphEpochDirectory,
+  currentPage: AgentGraphEpochDirectory,
+): boolean {
+  if (!currentPage.truncated && currentPage.epochs.length !== cached.epochs.length) return false;
+  return currentPage.epochs.every((entry, index) => {
+    const previous = cached.epochs[index];
+    return (
+      previous?.epoch === entry.epoch &&
+      previous.graphId === entry.graphId &&
+      previous.current === entry.current
+    );
+  });
 }
 
 function firstWait(operator: AgentGraphClientOperator) {
--- a/packages/cli/src/pi-tui-runner.ts
+++ b/packages/cli/src/pi-tui-runner.ts
@@ -2302,6 +2302,7 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
         if (!epoch) return;
         void runControl(async () => {
           const graph = await input.agentGraphHistory!.getSnapshot(rootSessionId, epoch.graphId);
+          if (closed) return;
           state.entries.push({
             kind: 'notice',
             level: 'info',
--- a/packages/runtime-host/src/protocol/agent-graph.ts
+++ b/packages/runtime-host/src/protocol/agent-graph.ts
@@ -335,6 +335,7 @@ export interface AgentGraphOperatorQueryInput {
 
 export interface AgentGraphStopInput {
   readonly rootSessionId: string;
+  readonly expectedGraphId?: string;
 }
 
 export interface AgentGraphStopResult {
@@ -418,7 +419,12 @@ export const AGENT_GRAPH_OPERATION_SPECS = {
     errors: STOP_ERRORS,
     decodeInput: decodeAgentGraphStopInput,
     decodeOutput: decodeAgentGraphStopResult,
-    assertOutputForInput: (input, output) => assertRootIdentity(input.rootSessionId, output),
+    assertOutputForInput: (input, output) => {
+      assertRootIdentity(input.rootSessionId, output);
+      if (input.expectedGraphId !== undefined && output.graphId !== input.expectedGraphId) {
+        throw invalidProtocolFrame('Agent graph stop changed request graph identity');
+      }
+    },
   }),
 } as const;
 
@@ -531,8 +537,18 @@ function decodeOptionalEpochCursor(value: unknown): number | null {
 }
 
 export function decodeAgentGraphStopInput(value: unknown): AgentGraphStopInput {
-  const record = requireExactRecord(value, 'agent.graph.stop input', ['rootSessionId']);
-  return { rootSessionId: requireEntityId(record.rootSessionId, 'rootSessionId') };
+  const record = requireShapedRecord(
+    value,
+    'agent.graph.stop input',
+    ['rootSessionId'],
+    ['expectedGraphId'],
+  );
+  return {
+    rootSessionId: requireEntityId(record.rootSessionId, 'rootSessionId'),
+    ...(record.expectedGraphId === undefined
+      ? {}
+      : { expectedGraphId: requireOpaqueIdentity(record.expectedGraphId, 'expectedGraphId') }),
+  };
 }
 
 export function decodeAgentGraphStopResult(value: unknown): AgentGraphStopResult {
--- a/packages/runtime-host/src/server/execution-composition.ts
+++ b/packages/runtime-host/src/server/execution-composition.ts
@@ -1002,6 +1002,7 @@ export async function createExecutionRuntimeHostComposition(
           expectedGraphId,
           stopSupervisor: () =>
             requireRootCoordinator(rootCoordinator).stopAgentGraphSupervisor(rootSessionId, {
+              expectedGraphId,
               source: 'stop_button',
             }),
           withSupervisorWakesSuppressed: (operation) =>
--- a/packages/runtime-host/src/server/execution-composition.ts
+++ b/packages/runtime-host/src/server/execution-composition.ts
@@ -997,8 +997,9 @@ export async function createExecutionRuntimeHostComposition(
     graphClient = new HostAgentGraphCoordinator({
       authority: graphCoordinator,
       continuity: continuityCoordinator,
-      stopExecution: (rootSessionId) =>
+      stopExecution: (rootSessionId, expectedGraphId) =>
         requireGraphCoordinator(graphCoordinator).stopExecution(rootSessionId, {
+          expectedGraphId,
           stopSupervisor: () =>
             requireRootCoordinator(rootCoordinator).stopAgentGraphSupervisor(rootSessionId, {
               source: 'stop_button',
--- a/packages/runtime/src/stream-graph-coordinator.ts
+++ b/packages/runtime/src/stream-graph-coordinator.ts
@@ -113,6 +113,7 @@ export interface AgentGraphCoordinatorInput {
 }
 
 export interface AgentGraphExecutionStopInput {
+  expectedGraphId?: string;
   stopSupervisor(): Promise<void>;
   withSupervisorWakesSuppressed(operation: () => Promise<void>): Promise<void>;
 }
@@ -560,8 +561,14 @@ export class AgentGraphCoordinator {
   /** Stop the validated root supervisor and its graph under one wake fence. */
   async stopExecution(rootSessionId: string, input: AgentGraphExecutionStopInput): Promise<void> {
     await this.#assertRootSupervisor(rootSessionId);
-    const driver = await this.#driver(rootSessionId);
     await input.withSupervisorWakesSuppressed(async () => {
+      const driver = await this.#driver(rootSessionId);
+      if (input.expectedGraphId !== undefined && driver.graphId !== input.expectedGraphId) {
+        throw new AgentGraphClientOperationError(
+          'operation_conflict',
+          `Agent graph ${input.expectedGraphId} is no longer current`,
+        );
+      }
       const failures: unknown[] = [];
       try {
         await input.stopSupervisor();
--- a/apps/desktop/src/main/__tests__/agent-graph-panel.test.ts
+++ b/apps/desktop/src/main/__tests__/agent-graph-panel.test.ts
@@ -93,6 +93,7 @@ function installGraphRenderer(
   renderSession(sessionId: string): Promise<void>;
   holdNextEpochList(sessionId: string): DeferredRead;
   holdNextSnapshot(graphId: string): DeferredRead;
+  holdNextStop(sessionId: string): DeferredRead;
   setCurrentWithoutNotification(next: AgentGraphClientSnapshot): void;
   notify(): void;
   epochReadCounts(): { full: number; current: number };
@@ -133,6 +134,7 @@ function installGraphRenderer(
   const listeners = new Set<GraphListener>();
   const epochListGates = new Map<string, DeferredReadGate>();
   const snapshotGates = new Map<string, DeferredReadGate>();
+  const stopGates = new Map<string, DeferredReadGate>();
   const stopCalls: Array<{ sessionId: string; expectedGraphId: string }> = [];
   let fullEpochReads = 0;
   let currentEpochReads = 0;
@@ -195,6 +197,13 @@ function installGraphRenderer(
       },
       stop: async (sessionId: string, expectedGraphId: string) => {
         stopCalls.push({ sessionId, expectedGraphId });
+        const gate = stopGates.get(sessionId);
+        if (gate) {
+          stopGates.delete(sessionId);
+          gate.markStarted();
+          await gate.waitForRelease;
+          throw new Error('deferred stop failure');
+        }
         if (currentGraphIds.get(sessionId) !== expectedGraphId) {
           throw new Error('graph changed before stop');
         }
@@ -250,6 +259,11 @@ function installGraphRenderer(
     holdNextSnapshot(graphId) {
       const gate = deferredReadGate();
       snapshotGates.set(graphId, gate);
+      return gate;
+    },
+    holdNextStop(sessionId) {
+      const gate = deferredReadGate();
+      stopGates.set(sessionId, gate);
       return gate;
     },
     stopCalls,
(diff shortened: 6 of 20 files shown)


apps/desktop/src/renderer/agent-graph-panel.tsx
```diff
@@ -144,36 +155,60 @@ export function AgentGraphPanel(props: {
...
-          if (!disposed) {
+        .listEpochs(props.rootSessionId)
+        .then(async (nextEpochs) => {
+          const current = nextEpochs.find((entry) => entry.current) ?? nextEpochs[0];
+          const selected = followCurrentRef.current
+            ? current
+            : nextEpochs.find((entry) => entry.graphId === selectedGraphIdRef.current);
+          const graphId = (selected ?? current)?.graphId;
+          if (!graphId) throw new Error('Agent graph epoch directory is empty');
+          selectedGraphIdRef.current = graphId;
```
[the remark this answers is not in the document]
me2seeks: Fixed in a520f2c11: when the explicit selection leaves the bounded directory the panel sets followCurrentRef back to true before falling back, so later rollovers refresh the panel instead of pinning it on the fallback graph. A regression test evicts the selected epoch, rolls over twice, and asserts the combobox follows the new current each time (verified to fail without the fix).

packages/cli/src/pi-tui-runner.ts
```diff
@@ -2264,6 +2270,56 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
     requestRender();
   };
 
+  const showGraphHistory = async (): Promise<void> => {
+    const rootSessionId = input.driver.getSessionId();
+    if (!rootSessionId || !input.agentGraphHistory) {
+      throw new Error('Agent Graph history is unavailable on this session driver.');
+    }
+    const epochs = await input.agentGraphHistory.listEpochs(rootSessionId);
```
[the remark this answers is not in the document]
me2seeks: Fixed in a520f2c11: showGraphHistory returns immediately after the awaited page reads when the TUI has shut down, matching the file's existing closed guards, so showSelectPicker never runs on a stopped TUI.

packages/runtime/src/stream-graph-coordinator.ts
```diff
@@ -321,21 +289,44 @@ export class AgentGraphCoordinator {
...
-    return epochs.length > 0 ? epochs.map(({ graphId }) => graphId) : [current.graphId];
+    return epochs.length > 0 ? epochs : [current];
+  }
+
+  async listGraphEpochPage(
+    rootSessionId: string,
+    options: { readonly beforeEpoch?: number; readonly limit: number },
+  ): Promise<AgentGraphEpochPage & { readonly currentEpoch: number }> {
+    requireRootSessionId(rootSessionId);
+    const current = await this.currentGraphEpoch(rootSessionId);
```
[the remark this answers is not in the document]
me2seeks: Fixed in ae2a0dc07 with the suggested consolidation: the epoch-page storage seam now returns currentEpoch from the same read transaction as the page rows, and the coordinator uses it instead of a separate resolveCurrentAgentGraphEpoch read, so a rollover can no longer split the current marker from the directory. The legacy virtual-epoch synthesis remains only for a root Session with no durable rows. Storage tests pin currentEpoch on first and historical pages plus the empty-root null case; a coordinator test pins that the rows path performs no second resolution read.

packages/runtime-host/src/client/agent-graph-reader.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/** Collect one bounded, newest-first directory of Graph epochs from the Host. */
+export async function readRuntimeHostAgentGraphEpochs(
+  connection: AgentGraphReadConnection,
+  rootSessionId: string,
+): Promise<readonly AgentGraphEpochSummary[]> {
+  const epochs: AgentGraphEpochSummary[] = [];
+  const cursors = new Set<number>();
+  let beforeEpoch: number | undefined;
+  for (let pageCount = 0; pageCount < MAX_EPOCH_PAGES; pageCount += 1) {
```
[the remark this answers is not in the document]
me2seeks: Fixed in a520f2c11: the collector keeps the 64-page bound but returns a clearly truncated directory ({ epochs, truncated: true }) instead of discarding it, while still throwing on a repeated cursor (invalid data). Desktop shows a 'Showing the newest N runs' caption and the TUI labels the picker 'newest N runs (history capped)', so neither surface misrepresents a capped directory as complete. A reader regression collects 64 valid pages and asserts truncated=true with all 64 pages retained.

Astro-Han (commented):
I reviewed exact head `a520f2c1128b64413d21508708d6c17376b1a6f4`, including the full Storage → Runtime → Runtime Host → Desktop/TUI path, current CI, and all prior feedback.

The ownership model is sound. Durable epoch bindings remain in Storage, Runtime Host is the sole historical graph reader, exact graph reads are checked against the root Session, and clients consume bounded read-only projections rather than opening a parallel persistence path. The earlier atomic page/current observation, large-history truncation, selected-history eviction, and TUI shutdown findings are fixed on this head.

One P2 remains in the Desktop refresh lifecycle.

`refresh()` advances `refreshGeneration` before checking whether a request is already in flight. A normal `graphs:changed` notification therefore both queues a follow-up refresh and invalidates the current response, even though the session and selected graph have not changed. Runtime activity can publish these notifications continuously; if one arrives during each `listEpochs()` / `getSnapshot()` sequence, every valid snapshot is discarded and the panel stays blank or stale until the graph becomes quiet.

A coalesced notification should request one follow-up read while allowing the current read for the same session and selection to commit. Advance the fence only when the session or selected graph changes, or rely on an explicit selected-graph identity token. Please add a deferred-request regression that emits an invalidation while refresh is in flight.

Non-blocking simplification: the unused `RuntimeHostCapabilityProviderService` export can be deleted. If historical operator drill-down is not shipping in this slice, its currently unconsumed public protocol expansion can also wait for a real consumer.

The PR should remain one vertical slice; splitting it would only create an unusable half-contract.

**Disclosure:** Codex performed the read-only source, lifecycle, protocol, test, CI, and prior-feedback analysis. The human contributor remains responsible for verifying the finding and deciding whether to merge.

<details>
<summary>中文</summary>

历史 graph 的 owner 方向正确，既有四条反馈已修。当前 P2 是普通 invalidation 与用户 selection 共用 generation fence：活跃 graph 可能让每轮 snapshot 都被丢弃，直到 graph 静默。普通通知应只排队 follow-up。

</details>

me2seeks:
The P2 refresh-lifecycle finding is fixed on `83371c455`.

The refresh scheduling now lives in a pure `createAgentGraphRefreshScheduler` (`apps/desktop/src/renderer/agent-graph-refresh.ts`) with two distinct entries:

- `requestRefresh()` (notification-driven) only queues one follow-up read — it never advances the fence, so an in-flight `listEpochs()`/`getSnapshot()` sequence for the same session and selection commits even while a busy graph keeps publishing.
- `invalidateAndRefresh()` (epoch selector / session change) advances the fence, so a stale in-flight read for a superseded selection is still discarded.

Regression coverage in `apps/desktop/src/main/__tests__/agent-graph-refresh.test.ts`: the deferred-request case emits a notification mid-flight and asserts the in-flight read commits and exactly one coalesced follow-up runs; the invalidation case asserts the superseded read is discarded; plus notification-storm coalescing and dispose. Mutation check: restoring fence-advance on every notification fails the two key tests. Desktop suite green (906/906), renderer build clean.

On the non-blocking note: `RuntimeHostCapabilityProviderService` is still the public return type of `startRuntimeHostCapabilityProviderService` (consumed by the CLI via `Awaited<ReturnType<…>>`), so I left it exported rather than narrowing the API surface in this slice.

apps/desktop/src/renderer/agent-graph-panel.tsx
```diff
@@ -136,62 +160,72 @@ export function AgentGraphPanel(props: {
...
+        if (scheduler.isCurrent(fence) && next.graphId === selectedGraphIdRef.current) {
+          setEpochs(nextEpochs);
+          setEpochsTruncated(directory.truncated);
+          setSnapshot(next);
+          setError(false);
+        }
+      } catch {
+        if (scheduler.isCurrent(fence)) setError(true);
+      } finally {
+        setLoading(false);
```
Astro-Han: **P2 — Do not let a disposed read settle a newer session's loading state.** When `rootSessionId` changes, the old scheduler is disposed and the new effect sets `loading=true`; if the old `listEpochs/getSnapshot` finishes afterward, this unconditional `finally` still calls `setLoading(false)`. The new session can then show an empty panel with no spinner until its own slower request completes. Please settle loading only when `scheduler.isCurrent(fence)` (including disposal), or key loading by the request/session generation. Add a deferred-promise test where session A completes after session B starts.

apps/desktop/src/renderer/agent-graph-panel.tsx
```diff
@@ -264,6 +304,29 @@ export function AgentGraphPanel(props: {
...
+              label={copy.epoch}
+              isLabelHidden
+              value={snapshot.graphId}
+              options={epochs.map((entry) => ({
+                value: entry.graphId,
+                label: `#${entry.epoch} · ${entry.current ? copy.currentEpoch : copy.historicalEpoch}`,
+              }))}
+              onChange={(graphId: SelectorOptionType) => {
+                if (typeof graphId !== 'string') return;
+                selectedGraphIdRef.current = graphId;
```
Astro-Han, line 364: **P2 — Make the newly selected epoch authoritative before leaving current-only actions enabled.** This updates only a ref; until the historical snapshot arrives, `snapshot` and `selectedEpoch` still describe the current graph, so `stopAvailable` remains true. A user can select history and immediately click Stop, which calls `graphs.stop(rootSessionId)` and stops the current graph—not the historical run they just selected. Please hold `selectedGraphId` in render state, disable Stop/Dismiss as soon as a non-current ID is chosen (and while its snapshot is pending), and add a delayed historical-read interaction test.

Astro-Han (commented):
Thanks for making historical graph runs inspectable across Host, CLI, and Desktop. The durable epoch directory plus explicit `graphId` reads is the right shape: Host remains the read authority, historical snapshots are immutable/read-only, and the UI follows current until the user intentionally pins history. Independent reviewer passes and a separate read-only DeepSeek V4 Flash high-effort pass reviewed this head.

The remaining defects come from splitting selection authority between refs and the last rendered snapshot. Immediately after choosing history, the rendered snapshot still represents current, so the current-only Stop action remains live and can stop a different graph than the one the user just selected. Separately, a disposed read can still clear the loading state owned by a newer session/effect.

From first principles, one state tuple—`{rootSessionId, selectedGraphId, requestGeneration}`—should own both the rendered snapshot and all actions. Update the selected graph synchronously on selection, disable current-only actions until the matching snapshot commits, and let only the current request generation settle loading/error. That is simpler than keeping parallel ref/snapshot interpretations and closes both races at the same seam.

I did not run a local test suite; this review is based on the latest head, focused static tracing, current green CI, independent `@reviewer` passes, and a read-only `ollama-cloud/deepseek-v4-flash` high-effort review. AI-assisted review; I verified the UI race paths.

感谢把历史 Graph run 的浏览能力贯通到 Host、CLI 和 Desktop。持久 epoch directory 配合显式 `graphId` 读取的方向正确：Host 保持读取 authority，历史 snapshot 只读，UI 默认跟随 current，用户选择后才固定历史。

剩余问题来自 selection authority 被拆在 ref 和最后一次渲染的 snapshot 中。选择历史后，界面暂时仍把 current snapshot 当成所选对象，因此“停止 Graph”仍可点击，并会停止另一个 graph；同时旧 session 已 dispose 的请求仍可清除新 session 的 loading。

更符合第一性原理和奥卡姆剃刀的方案，是用同一个 `{rootSessionId, selectedGraphId, requestGeneration}` 状态元组同时拥有 snapshot 与 action：选择时立即更新 selected graph，匹配 snapshot 提交前禁用 current-only action，只有当前 generation 能结算 loading/error。这样在一个 seam 同时消除两条竞态。

本次未运行本地测试；结论来自最新 head 静态追踪、当前绿色 CI、独立 reviewer 和一次只读的 DeepSeek V4 Flash high-effort 审查。AI 辅助审查；我已人工复核 UI 竞态路径。

</details>

apps/desktop/src/renderer/agent-graph-panel.tsx
me2seeks: Fixed in 621296939. Loading settlement is now keyed to the scheduler fence, including disposal, so a retired session read cannot clear the newer session loading state. The deferred A-after-B regression reproduces the old stale finally path.

apps/desktop/src/renderer/agent-graph-panel.tsx
me2seeks: Fixed in 621296939. selectedGraphId is render state and becomes authoritative immediately; Stop and Dismiss disable as soon as history is selected and remain disabled while its snapshot is pending. The delayed historical-read interaction regression covers the click race.

packages/runtime/src/stream-graph-coordinator.ts
```diff
@@ -321,21 +289,57 @@ export class AgentGraphCoordinator {
...
     const epochs = await this.#input.epochStore.listAgentGraphEpochs(rootSessionId);
-    return epochs.length > 0 ? epochs.map(({ graphId }) => graphId) : [current.graphId];
+    return epochs.length > 0 ? epochs : [current];
+  }
+
+  async listGraphEpochPage(
+    rootSessionId: string,
+    options: { readonly beforeEpoch?: number; readonly limit: number },
+  ): Promise<AgentGraphEpochPage & { readonly currentEpoch: number }> {
+    requireRootSessionId(rootSessionId);
```
Astro-Han: [P2] Validate that this is an existing root Session before reading or synthesizing epochs. `requireRootSessionId` checks only the string shape, unlike snapshot/timeline operations that call `#assertRootGraphReader`. A canonical nonexistent ID can therefore receive (and, through `resolveCurrentAgentGraphEpoch`, potentially materialize) a virtual epoch, while a child Session is accepted until a later snapshot call fails. Call `#assertRootGraphReader` here before the store access and add unknown/child Session cases that return the operation's normal `not_found`/`operation_conflict` errors.

packages/runtime-host/src/protocol/agent-graph.ts
```diff
@@ -318,6 +339,34 @@ export interface AgentGraphStopResult {
...
+    assertOutputForInput: (input, output) => {
+      assertRootIdentity(input.rootSessionId, output);
+      if (input.beforeEpoch === undefined) {
+        if (output.epochs.filter((entry) => entry.current).length !== 1) {
+          throw invalidProtocolFrame(
+            'Agent graph first epoch page must identify the current epoch',
+          );
+        }
+      } else {
+        if (output.epochs.some((entry) => entry.current)) {
```
Astro-Han, line 366: [P2] Reject or normalize `beforeEpoch` values above the observed current epoch on the server side. The storage query accepts any positive cursor; with current epoch 2, `beforeEpoch: 999` returns epoch 2 and the Host marks it `current: true`. This assertion then treats the otherwise valid response as an invalid protocol frame, and the client connection closes rather than returning an `invalid_request`. Add an integration case for a cursor above current and keep malformed user input inside the operation error boundary.

Astro-Han (commented):
Historical graph browsing belongs in the Runtime Host graph authority, and this PR generally preserves that boundary: pages are bounded, snapshots are read-only, and Desktop selection fencing avoids acting on historical state. Two validation gaps remain at the new epoch-list operation. The simplest fix is to validate the root once at the authority boundary and make cursor normalization/rejection happen before a page reaches the strict client assertion.

`test_workspaces` and `storybook` are still running on the current head.

AI-assisted review disclosure: Codex verified the findings against the current head and traced Runtime session ownership, epoch storage transactions, Host projection, protocol assertions, and client connection failure handling. Two independent reviewer-agent passes and an OpenCode Go DeepSeek V4 Flash (high) adversarial pass were used as inputs. No local tests were run.

<details>
<summary>中文复核</summary>

历史 Graph 浏览应由 Runtime Host 的 Graph 权威负责，这个 PR 基本守住了边界：分页有界、历史 snapshot 只读，Desktop selection fencing 也避免了对历史状态执行操作。新 epoch-list operation 还缺两个校验：root Session 所有权/存在性，以及越过 current epoch 的 cursor。最小修复是在 authority boundary 统一验证 root，并在 page 进入严格 client assertion 前规范化或拒绝 cursor。当前 `test_workspaces` 与 `storybook` 仍在运行。

本次为 AI 辅助审查：Codex 在最新 head 上追踪 Runtime Session ownership、epoch storage transaction、Host projection、协议 assertion 与 client connection failure；另使用两次独立 reviewer 及一次 OpenCode Go DeepSeek V4 Flash（high）对抗审查。未运行本地测试。
</details>

me2seeks (commented):
Publishing the verified inline fix responses.

packages/runtime/src/stream-graph-coordinator.ts
me2seeks: Fixed in fa2daf4fb. Both epoch-list entry points now assert the canonical root Session before reading or synthesizing history. The integration coverage pins child Sessions to operation_conflict and unknown Sessions to the normal not-found path.

packages/runtime-host/src/protocol/agent-graph.ts
me2seeks: Fixed in fa2daf4fb. Runtime and Host boundaries now reject beforeEpoch above the observed current epoch as invalid_request, including no-store and synthesized paths, so malformed cursors no longer escape as protocol-frame failures. The focused Runtime and Host Agent Graph suites pass.

apps/desktop/src/renderer/agent-graph-panel.tsx
```diff
@@ -136,62 +160,75 @@ export function AgentGraphPanel(props: {
...
+          ? current
+          : nextEpochs.find((entry) => entry.graphId === selectedGraphIdRef.current);
+        // An evicted selection must not pin the panel on the fallback:
+        // resume following the current epoch so later rollovers refresh.
+        if (!selected && !followCurrentRef.current) {
+          followCurrentRef.current = true;
+        }
+        const graphId = (selected ?? current)?.graphId;
+        if (!graphId) throw new Error('Agent graph epoch directory is empty');
+        selectedGraphIdRef.current = graphId;
```
Astro-Han, line 234: [P2] Guard shared selection refs before a stale effect can mutate them. These writes happen before the scheduler fence check, and the refs are shared across effect instances. Reproduce by deferring session A’s epoch-directory read, switching to session B, selecting a historical B epoch, then resolving A: the stale A callback cannot find B’s selection in A’s directory, flips `followCurrentRef` back to true, and writes A’s graph ID into the live ref. The next B refresh can therefore abandon the user’s selection. Check `scheduler.isCurrent(fence)` before mutating either ref (or keep selection state scoped to the effect/session), and add this deferred cross-session regression.

Astro-Han (commented):
The core design is well chosen: Runtime Host remains the only historical Graph reader, epoch paging and the current marker share one storage observation, graph ownership is checked against the root Session, and Desktop/TUI history is read-only. The exact head is green and merges cleanly with current main.\n\nTwo UI authority races and one avoidable refresh cost remain below. The first-principles fix is to bind mutable UI actions and commits to the graph/session identity the user actually saw, while keeping the Host operation as the final authority. No broader redesign is needed.\n\nReviewed with Codex using three independent reviewer agents and OpenCode Go DeepSeek V4 Flash (high); I verified the exact head, current-main merge tree, protocol/storage boundaries, UI refresh paths, focused tests, and live CI.\n\n<details><summary>中文</summary>\n\n核心设计方向正确：Runtime Host 仍是历史 Graph 的唯一读取权威；epoch 分页与 current marker 来自同一次 storage observation；graph ownership 按 root Session 校验；Desktop/TUI 历史保持只读。当前精确 head 检查全绿，也可与最新 main 干净合并。\n\n目前还剩下面两个 UI 权威竞态和一个可避免的刷新成本。按第一性原理，修复应把可变 UI action/commit 绑定到用户实际看到的 graph/session identity，并继续让 Host operation 做最终权威校验；不需要更大的重构。\n\n本次由 Codex 配合三个独立 reviewer agent，以及 OpenCode Go DeepSeek V4 Flash（high）审查；我核验了精确 head、当前 main 的 merge tree、协议/存储边界、UI 刷新路径、聚焦测试和实时 CI。\n\n</details>

apps/desktop/src/renderer/agent-graph-panel.tsx
```diff
@@ -251,9 +290,17 @@ export function AgentGraphPanel(props: {
     }
   };
   const stopAvailable =
-    snapshot !== undefined && ['active', 'waiting', 'closing'].includes(snapshot.status);
+    selectedEpoch?.current === true &&
```
Astro-Han, line 330: [P2] Bind Stop to the graph identity shown by the button. This condition only proves the snapshot was current when it was read; `graphs.stop(rootSessionId)` later stops whichever graph is current at click time. If an epoch rolls over between render and click, the user can stop the new graph while the UI still displays the old one. Send `expectedGraphId` with the stop request and have Runtime Host reject a mismatch at the authority boundary; cover rollover-between-render-and-click.

apps/desktop/src/renderer/agent-graph-panel.tsx
```diff
@@ -136,62 +160,75 @@ export function AgentGraphPanel(props: {
...
-          setLoading(false);
-          task = undefined;
-          if (queued) {
-            queued = false;
-            refresh();
-          }
-        });
-    };
+      try {
+        const directory = await window.maka.graphs.listEpochs(props.rootSessionId);
```
Astro-Han: [P3] Avoid rebuilding the entire epoch directory on every `graphs:changed` notification. This call can perform up to 64 sequential Host requests (2,048 epochs) before the one snapshot read, and active graphs publish notifications continuously; the scheduler coalesces overlap but can still keep repeating the full history scan. Cache the historical pages and refresh only the current page during background updates, loading older pages when the selector opens or the directory actually changes.

apps/desktop/src/renderer/agent-graph-panel.tsx
me2seeks: Fixed in ec56d97cc. The refresh now checks scheduler.isCurrent(fence) immediately after the directory read, before updating cachedDirectory or any shared selection refs. Added a regression where a delayed Session A read completes after switching to Session B and cannot overwrite B selection state.

apps/desktop/src/renderer/agent-graph-panel.tsx
me2seeks: Fixed in ec56d97cc. Desktop Stop now carries the graphId from the rendered snapshot through preload, IPC, protocol, and Host execution. Runtime compares expectedGraphId under the supervisor wake fence and returns operation_conflict before stopping when rollover changed the current graph. Added panel and Runtime mismatch regressions.

apps/desktop/src/renderer/agent-graph-panel.tsx
me2seeks: Fixed in ec56d97cc. The panel caches the full epoch directory for the active effect and activity notifications read only the current epoch page. It reuses the cache when that page is unchanged and performs a full paginated reload only when rollover/directory identity changes. Added read-count coverage for repeated activity and rollover.

apps/desktop/src/renderer/agent-graph-panel.tsx
```diff
@@ -136,62 +161,86 @@ export function AgentGraphPanel(props: {
...
+    let cachedDirectory: AgentGraphEpochDirectory | undefined;
 
-    const refresh = (): void => {
-      if (disposed) return;
-      if (task) {
-        queued = true;
-        return;
-      }
+    const scheduler = createAgentGraphRefreshScheduler(async (fence) => {
       setLoading(true);
```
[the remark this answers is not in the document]
me2seeks: Fixed in d30cd39a4. Background refreshes now leave loading false once the effect has a cached directory, so current Stop/Dismiss controls remain mounted. Graph/selection identity still gates both actions, and Stop remains bound by expectedGraphId. Added a deferred background-snapshot regression; Desktop focused tests pass 20/20.

packages/cli/src/pi-tui-runner.ts
```diff
@@ -2260,6 +2267,61 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
...
+        if (!epoch) return;
+        void runControl(async () => {
+          const graph = await input.agentGraphHistory!.getSnapshot(rootSessionId, epoch.graphId);
+          state.entries.push({
+            kind: 'notice',
+            level: 'info',
+            text: formatAgentGraphHistory(graph, epoch),
+          });
+          requestRender();
+        });
```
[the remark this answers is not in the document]
me2seeks: Fixed in d30cd39a4. The history control now checks closed immediately after getSnapshot returns, before appending the notice or requesting a render. Added a deferred snapshot regression that shuts down before release; the CLI suite passes 275/275.

Astro-Han (commented):
Thanks for the substantial follow-up work—the storage/runtime/host/UI authority boundaries are now coherent, and the earlier epoch, refresh-fence, expected-graph, shutdown, and truncation issues are addressed.

Two things prevent approval of this exact head. First, GitHub currently reports it as conflicting with current `main`, which now includes #2992's cross-epoch graph-result semantics across many of the same files; the rebase must preserve those semantics and then receive a fresh review. Second, I left one new P2 inline for Stop state leaking across a root-session switch.

Please rebase onto current `main`, retain #2992's result authority, fence/reset the Stop request state by session/request generation, and add the focused A→B deferred-stop regression. Then rerun CI and request review on the new head.

AI-assisted review disclosure: two Codex reviewers independently reviewed exact head `d30cd39`, current merge state, prior fixes, and the Desktop session-switch path.

<details>
<summary>中文说明</summary>

之前的 epoch、刷新 fence、expected graph、TUI shutdown 和截断问题已经修复。但当前 head 与已合并 #2992 的跨 epoch result 语义发生冲突，必须先 rebase，并确保保留 #2992 的权威边界。另有一个新的 P2：Stop 的 pending/error 是组件级状态，切换 root session 时可能把 A 的结果泄漏到 B。建议按 session/request generation 做 fence 和重置，并补 A→B 的 deferred-stop 测试。

</details>

apps/desktop/src/renderer/agent-graph-panel.tsx
```diff
@@ -238,22 +289,30 @@ export function AgentGraphPanel(props: {
     return null;
   }
 
-  const stopGraph = async (): Promise<void> => {
+  const stopGraph = async (expectedGraphId: string): Promise<void> => {
```
Astro-Han, line 308: **[P2] Fence Stop state to the root session that started it**

`stopPending` and `stopError` are component-wide, while this async path captures the old `rootSessionId` without a session/request generation guard. If Session A starts `graphs.stop()`, the user switches to Session B, and A then rejects or remains pending, B inherits A's error and its Stop button remains blocked until A settles. Resetting only `stopError` in the data-loading effect does not fence the late `catch/finally` writes.

Please bind this request to the initiating root session (or a request generation), ignore stale completion writes, reset both pending/error when the effect changes session, and add a deferred A→B switch regression.

me2seeks: Fixed in 24456836e after rebasing onto current main. Stop state is now keyed by the initiating rootSessionId and a monotonic requestId; session changes invalidate the prior request and reset both pending/error, while stale catch/finally completions are ignored. The deferred A-to-B regression verifies that B retains its Stop control and never inherits A failure. Desktop graph tests pass 21/21, and the Runtime graph/reconciliation suites pass on the rebased #2992 base.

Astro-Han:
Can We have a before after comparison screenshot?

packages/runtime-host/src/server/execution-composition.ts
```diff
@@ -997,8 +997,9 @@ export async function createExecutionRuntimeHostComposition(
     graphClient = new HostAgentGraphCoordinator({
       authority: graphCoordinator,
       continuity: continuityCoordinator,
-      stopExecution: (rootSessionId) =>
+      stopExecution: (rootSessionId, expectedGraphId) =>
         requireGraphCoordinator(graphCoordinator).stopExecution(rootSessionId, {
+          expectedGraphId,
```
Astro-Han, line 1002: **[P2] Carry the expected graph identity into the supervisor stop admission**

The Runtime coordinator checks `expectedGraphId`, but `stopSupervisor()` then calls `stopAgentGraphSupervisor(rootSessionId)` without that identity. That method independently resolves the current graph before entering session admission. If graph A becomes terminal and rolls to B after the first check, the stale A request can resolve B and stop B's fresh root execution before the Runtime side stops A.

Pass `expectedGraphId` into `stopAgentGraphSupervisor` and validate it inside the same per-session admission that selects the active root identity. Add a regression that advances the epoch between the Runtime check and supervisor stop.

apps/desktop/src/renderer/agent-graph-panel.tsx
```diff
@@ -277,16 +302,25 @@ export function AgentGraphPanel(props: {
...
     if (stopPending) return;
-    setStopPending(true);
-    setStopError(false);
+    const rootSessionId = props.rootSessionId;
+    const requestId = ++stopRequestIdRef.current;
+    setStopState({ rootSessionId, requestId, pending: true, error: false });
```
Astro-Han: **[P3] Scope stop feedback to the graph that initiated it**

`stopState` carries only `rootSessionId` and `requestId`. If a deferred stop for graph A fails after the panel refreshes to graph B (or the user selects a historical epoch), `stopError` is still rendered for that new selection, and pending state can briefly label B as stopping.

Store the initiating `expectedGraphId` and derive pending/error only when it matches the currently selected snapshot. Cover a deferred A stop settling after an A→B rollover and after switching to history.

Astro-Han (commented):
The current patch correctly binds the protocol request to `expectedGraphId`, keeps Runtime Host as the sole graph authority, and prevents a deferred stop result from leaking across root sessions. Two remaining identity gaps need to be closed before approval: the server can still stop a newly rolled graph supervisor, and renderer feedback is scoped only to the session rather than the selected graph.

E2E, Storybook, and the root test job are still running/queued.

AI-assisted review disclosure: Codex reviewed exact head `2445683`, the current three-commit PR patch against its live base, current CI, and resolved thread history, with two independent reviewer passes.

<details>
<summary>中文说明</summary>

当前 patch 已把 stop 请求绑定到 `expectedGraphId`，Runtime Host 仍是唯一权威，也修复了跨 root session 的异步状态泄漏。但身份约束还没有贯穿完整路径：server 在第一次校验后仍可能重新解析并停止刚 rollover 的新 graph supervisor；renderer 的 pending/error 也只按 session 隔离，会泄漏到同 session 的新 graph 或历史视图。相关 CI 仍在运行。

</details>

me2seeks:
Added the requested Before/After comparison to the PR body. It contrasts the current-only panel with a selected historical run, including the read-only label and absence of the current-run Stop action.

packages/runtime-host/src/server/execution-composition.ts
me2seeks: Fixed in d42bf37ee. expectedGraphId now reaches stopAgentGraphSupervisor and is compared with the active root graphOwnerId inside the same per-session admission that selects the stop identity. A stale rollover identity returns operation_conflict without stopping the fresh root. The regression advances the epoch after admission and verifies the newer identity cannot stop the older active root.

apps/desktop/src/renderer/agent-graph-panel.tsx
me2seeks: Fixed in d42bf37ee. Stop feedback is now keyed by rootSessionId, graphId, and requestId, and pending/error render only when that graph still matches the selected epoch. Deferred-failure regressions cover both A→B rollover and switching from the current graph into history.

Astro-Han (approved):
Approved on exact head `d42bf37ee6986ae4375d35651df08e5c05e1cbf1`.

Historical Graph browsing stays on the existing Runtime Host authority. Epoch pagination, root-Session ownership, cursor validation, graph-identity checks, and Desktop/TUI stale-result fences keep historical selection read-only and prevent it from acting on the current graph. The current-page refresh path also avoids rereading complete history on every activity notification.

No remaining P0-P2 findings. Current CI is green, mergeability is CLEAN, and there are no unresolved review threads. The requested before/after screenshots are present, accessible, and match the current read-only historical state. The PR body and all four substantive commits carry consistent Maka AI-use disclosure.

AI-assisted review disclosure: Codex reviewed the exact-head Host/CLI/Desktop paths, concurrency and identity fences, focused tests, live CI, thread state, screenshots, and provenance metadata. No external model was used. Astro-Han authorized this review campaign.

<details>
<summary>中文说明</summary>

历史 Graph 浏览继续复用 Runtime Host 单一权威；分页 epoch、root Session、cursor、graph identity，以及 Desktop/TUI 的异步 stale-result fence 都完整，历史选择保持只读，也不会误操作当前 Graph。

当前没有剩余 P0-P2，CI 全绿、CLEAN、无未解决线程。before/after 截图与当前实现一致，PR body 和 4 个 commit 的 AI disclosure 也完整。

</details>

Astro-Han:
Human merge confirmation: Astro-Han reviewed the final diff at `d42bf37ee6` and explicitly authorized this squash merge. All required checks are green, the current head is approved, screenshots and AI provenance are complete, and there are no unresolved review threads. The squash commit will retain `Generated-by: Maka`.

PR #2992, opened 2026-08-14
feat(runtime): admit cross-epoch graph results

me2seeks:
## Summary

- persist typed `selected_result_inputs` with the source graph identity for scheduled work
- admit only committed results selected by a finished earlier epoch of the same root Session
- revalidate historical inputs during recovery and hydrate them with explicit graph provenance
- page historical result discovery without reusing old operators, claims, routes, or control state

Historical results cross the epoch boundary as immutable data only. Runtime Host authorization verifies the root, epoch ordering, finished schedule selection, and exact RuntimeEvent-backed record before the schedule is committed; recovery performs the same resolution again. Input counts and discovery pages are bounded, and ambiguous current/historical record identities fail closed.

This completes the cross-epoch result-input slice of #2588 after the historical-read slice.

Depends on #2991
Refs #2588

This is intentionally opened as a Draft during the pre-Incubator stabilization window. It is not requesting merge until feature work resumes.

## Verification

- `npm run build:test`
- `npm run typecheck`
- Agent Graph authority, reconciliation, handoff, protocol, storage, Desktop, and TUI tests in the stacked latest-main run — 168 passed
- `git diff --check`

## Checklist

- [x] Yes — a new Agent Graph epoch can explicitly consume selected results from an earlier finished epoch
- [ ] No

<details>
<summary>中文说明</summary>

## 摘要

- 为 schedule work 持久化带来源 graph 身份的 `selected_result_inputs`
- 只允许引用同一 root Session 下、已结束旧 epoch 在 `finish` 中明确选择的已提交结果
- 恢复时重新执行相同授权，并在传给新 operator 的输入中保留 graph 来源
- 历史结果发现采用有界分页，不复用旧 operator、claim、route 或控制状态

跨 epoch 传递的只有不可变数据，不包含旧图的控制权。Runtime Host 会在 schedule 提交前验证 root、epoch 顺序、旧图的 `finish` 选择以及由 RuntimeEvent 支撑的精确 record；崩溃恢复时再次验证。输入数量与发现页都有上限，当前图和历史图的 record identity 若出现歧义会 fail closed。

这是 #2588 的跨 epoch 结果输入部分，接续前一项历史读取改动。

本 PR 在进入孵化器前的稳定期内以 Draft 打开；在 feature 合并恢复前不请求进入 `main`。

</details>

## AI disclosure / AI 披露

This PR was implemented by Codex under me2seeks’s direction and review. / 本 PR 由 Codex 在 me2seeks 的指导与审核下完成。

### DIFF
--- a/apps/desktop/src/renderer/agent-graph-panel.tsx
+++ b/apps/desktop/src/renderer/agent-graph-panel.tsx
@@ -36,6 +36,7 @@ type GraphPanelCopy = {
   epoch: string;
   currentEpoch: string;
   historicalEpoch: string;
+  cappedEpochs(count: number): string;
   noOperators: string;
   hiddenOperators(count: number): string;
   progress(settled: number, total: number, hasOmitted: boolean): string;
@@ -63,6 +64,7 @@ export function getAgentGraphPanelCopy(locale: UiLocale): GraphPanelCopy {
       epoch: 'Graph 运行轮次',
       currentEpoch: '当前',
       historicalEpoch: '历史记录（只读）',
+      cappedEpochs: (count) => `仅显示最近 ${count} 次运行`,
       noOperators: '等待主 Agent 创建 operator…',
       hiddenOperators: (count) => `另有 ${count} 个 operator`,
       progress: (settled, total, hasOmitted) =>
@@ -109,6 +111,7 @@ export function getAgentGraphPanelCopy(locale: UiLocale): GraphPanelCopy {
     epoch: 'Graph run',
     currentEpoch: 'Current',
     historicalEpoch: 'History (read-only)',
+    cappedEpochs: (count) => `Showing the newest ${count} runs`,
     noOperators: 'Waiting for the main agent to create an operator…',
     hiddenOperators: (count) => `${count} more operator${count === 1 ? '' : 's'}`,
     progress: (settled, total, hasOmitted) =>
@@ -147,6 +150,7 @@ export function AgentGraphPanel(props: {
 }): JSX.Element | null {
   const [snapshot, setSnapshot] = useState<AgentGraphClientSnapshot>();
   const [epochs, setEpochs] = useState<readonly AgentGraphEpochSummary[]>([]);
+  const [epochsTruncated, setEpochsTruncated] = useState(false);
   const [loading, setLoading] = useState(props.enabled);
   const [error, setError] = useState(false);
   const [stopPending, setStopPending] = useState(false);
@@ -167,6 +171,7 @@ export function AgentGraphPanel(props: {
 
     setSnapshot(undefined);
     setEpochs([]);
+    setEpochsTruncated(false);
     selectedGraphIdRef.current = undefined;
     followCurrentRef.current = true;
     setError(false);
@@ -185,24 +190,31 @@ export function AgentGraphPanel(props: {
       setLoading(true);
       task = window.maka.graphs
         .listEpochs(props.rootSessionId)
-        .then(async (nextEpochs) => {
+        .then(async (directory) => {
+          const nextEpochs = directory.epochs;
           const current = nextEpochs.find((entry) => entry.current) ?? nextEpochs[0];
           const selected = followCurrentRef.current
             ? current
             : nextEpochs.find((entry) => entry.graphId === selectedGraphIdRef.current);
+          // An evicted selection must not pin the panel on the fallback:
+          // resume following the current epoch so later rollovers refresh.
+          if (!selected && !followCurrentRef.current) {
+            followCurrentRef.current = true;
+          }
           const graphId = (selected ?? current)?.graphId;
           if (!graphId) throw new Error('Agent graph epoch directory is empty');
           selectedGraphIdRef.current = graphId;
           const next = await window.maka.graphs.getSnapshot(props.rootSessionId, { graphId });
-          return { next, nextEpochs };
+          return { next, nextEpochs, truncated: directory.truncated };
         })
-        .then(({ next, nextEpochs }) => {
+        .then(({ next, nextEpochs, truncated }) => {
           if (
             !disposed &&
             generation === refreshGeneration &&
             next.graphId === selectedGraphIdRef.current
           ) {
             setEpochs(nextEpochs);
+            setEpochsTruncated(truncated);
             setSnapshot(next);
             setError(false);
           }
@@ -324,6 +336,9 @@ export function AgentGraphPanel(props: {
                 refreshRef.current();
               }}
             />
+          ) : null}
+          {epochsTruncated ? (
+            <span className="maka-agent-graph-epoch-capped">{copy.cappedEpochs(epochs.length)}</span>
           ) : null}
           {snapshot ? (
             <span className="maka-agent-graph-progress">
--- a/apps/desktop/src/renderer/agent-graph-panel.tsx
+++ b/apps/desktop/src/renderer/agent-graph-panel.tsx
@@ -18,6 +18,17 @@ import {
   shouldShowAgentGraphPanel,
   type AgentGraphPanelDismissals,
 } from './agent-graph-panel-visibility.js';
+import {
+  createAgentGraphRefreshScheduler,
+  type AgentGraphRefreshScheduler,
+} from './agent-graph-refresh.js';
+
+const noopAgentGraphRefreshScheduler: AgentGraphRefreshScheduler = {
+  requestRefresh() {},
+  invalidateAndRefresh() {},
+  isCurrent: () => false,
+  dispose() {},
+};
 
 type GraphPanelCopy = {
   title: string;
@@ -158,17 +169,12 @@ export function AgentGraphPanel(props: {
   const [collapsed, setCollapsed] = useState(false);
   const [dismissedBySession, setDismissedBySession] = useState<AgentGraphPanelDismissals>({});
   const contentId = useId();
-  const refreshRef = useRef<() => void>(() => {});
+  const refreshRef = useRef<AgentGraphRefreshScheduler>(noopAgentGraphRefreshScheduler);
   const selectedGraphIdRef = useRef<string | undefined>(undefined);
   const followCurrentRef = useRef(true);
   const copy = getAgentGraphPanelCopy(props.locale);
 
   useEffect(() => {
-    let disposed = false;
-    let queued = false;
-    let refreshGeneration = 0;
-    let task: Promise<void> | undefined;
-
     setSnapshot(undefined);
     setEpochs([]);
     setEpochsTruncated(false);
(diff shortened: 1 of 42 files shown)


packages/runtime/src/stream-graph-coordinator.ts
```diff
@@ -1134,6 +1148,102 @@ export class AgentGraphCoordinator {
...
+    return selectedInputs.map((selected) => {
+      const record = recordsBySource.get(selected.sourceGraphId)?.get(selected.resultId);
+      if (!record) {
+        throw new Error(
+          `Selected result ${selected.resultId} is not a committed record of ${selected.sourceGraphId}`,
+        );
+      }
+      return structuredClone(record);
+    });
+  }
```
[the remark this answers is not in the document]
me2seeks: Fixed in 3428c3caf. Confirmed: one reconciliation re-read the snapshot up to seven times, and each read replayed the full committed projection of every selected source epoch. `reconcileAgentGraphSchedule` now creates a per-reconciliation `SelectedResultCache` (keyed by source graph id, then result id) threaded through every `readScheduleSnapshot` call; `resolveSelectedResultRecords` only asks the host resolver for result ids not already cached, so re-reading a snapshot over the same closed epoch costs zero projection replays. Regression test: `resolves each historical result source once per reconciliation` (asserts exactly one resolver call across a provisioning reconciliation; fails against the previous code with 3+ calls).

packages/runtime/src/stream-graph-schedule-reconcile.ts
```diff
@@ -462,16 +474,56 @@ async function readScheduleSnapshot(
...
-    schedule: projectAgentGraphSchedule(input.topology.graphId, updates),
+    schedule,
     observation,
     claims,
     topology,
     provisions,
     sourceByWorkId: scheduleSourceByWorkId(updates),
+    selectedResultRecords,
   };
 }
```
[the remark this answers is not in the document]
me2seeks: Fixed in 3428c3caf. Confirmed: a resolver throw propagated out of `readScheduleSnapshot` and wedged the whole graph. `resolveSelectedResultRecords` now groups requested inputs by source graph and isolates failures per group: a source whose historical records can no longer be resolved (throw, incomplete, or mismatched response) omits its records instead of throwing. Both `missingInputIds` checks (provision loop and dispatch loop) now go through `missingWorkInputIds`, which also counts `work.selectedResultInputs` whose resultId is absent from `committedRecords`, so the affected work item defers with the existing `input_not_committed` seam while the rest of the graph proceeds. Genuine contract violations that commit-time authorization already rejects (missing host resolver, ambiguous result ids, collision with the current graph) still throw. Regression tests: `defers work whose historical result source becomes unresolvable` (status `waiting`, no failures, deferred with missing result id) and `defers only the work items whose historical source failed` (healthy source dispatches, broken source defers); both fail against the previous code.

Astro-Han (commented):
I reviewed the updated head `3428c3caf5a9dfe5e154c1972378a940effc07ef`, including the one-commit delta, its new reconciliation tests, the complete stack comparison, current CI, and all review threads.

The latest commit correctly addresses the two CodeRabbit findings: historical result resolution is cached per reconciliation, failures are isolated by source graph, and healthy sources can continue when another source is unavailable. I found no new P0–P2 in that delta.

The branch still has one P2 stack-consistency problem. It remains diverged from #2991’s current head `a520f2c1`, behind the two commits that fixed atomic epoch-page observation and bounded history collection. Consequently, this head still reads the current epoch and page through separate storage observations. A rollover between those reads can return epoch N+1 while reporting N as current, causing the first-page protocol invariant to reject a valid history request.

Please rebase this branch onto the latest #2991 head rather than recreating those fixes locally. That is the smallest and cleanest correction and will also remove the stale >2,048-epoch behavior.

Non-blocking contract follow-up: the Runtime Host decoder still allows 64 current inputs plus 64 selected historical inputs, while the durable schedule contract caps their combined count at 64. Aligning the decoder and adding one rejection test would close that P3 boundary.

CI is currently red on E2E and the PR has no approval. Recommendation: rebase onto the corrected #2991 stack, rerun CI, then re-review the resulting head.

**Disclosure:** Codex performed the read-only latest-head, stack, lifecycle, protocol, CI, and feedback analysis. The human contributor remains responsible for verifying the evidence and deciding whether to merge.

<details>
<summary>中文</summary>

最新 commit 正确修复了历史结果缓存和按 source 隔离失败，但分支仍落后 #2991 两个一致性修复。epoch page/current 非原子读取的 P2 仍存在。最干净方案是 rebase 最新 #2991，不要再补平行实现。

</details>

me2seeks:
Both follow-ups are on `58b3895fc`:

- **Rebase onto #2991** — the branch now sits on #2991's corrected head (`83371c455`, including the atomic epoch-page observation and bounded history collection). No parallel fixes were recreated locally; the stale >2,048-epoch behavior is gone with the old base. The only adaptation needed after the rebase was rebuilding workspace deps against the new `AgentGraphEpochPage.currentEpoch` contract — no source changes in this branch's delta.
- **P3 decoder alignment** — `decodeSelectedResultInputs` now rejects frames where current `inputIds` + `selectedResultInputs` combined exceed `AGENT_GRAPH_MAX_WORK_INPUTS` (64), matching the durable schedule contract exactly. Regression test in `agent-graph-coordinator.test.ts` decodes a 64+1 frame (rejected) and a 63+1 frame (accepted); mutation check confirmed the rejection test fails without the cap.

Runtime 2885 green, runtime-host 962 green, biome clean. CI is re-running on the rebased head.

packages/runtime/src/stream-graph-schedule-reconcile.ts
```diff
@@ -462,16 +479,113 @@ async function readScheduleSnapshot(
...
+    if (
+      records.length !== pending.length ||
+      records.some((record, index) => {
+        const selected = pending[index]!;
+        return record.graphId !== selected.sourceGraphId || record.recordId !== selected.resultId;
+      })
+    ) {
+      continue;
+    }
+    records.forEach((record) => cached.set(record.recordId, record));
```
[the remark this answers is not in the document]
me2seeks: Fixed in 22a4ebe5a. SelectedResultCache now stores AgentGraphRecord | undefined and inserts the pending entry before resolution, so an unresolvable historical result is negatively cached for the reconciliation pass. The regression asserts one resolver call; removing the negative cache makes it fail with 2 !== 1.

packages/runtime/src/stream-graph-schedule-reconcile.ts
```diff
@@ -191,6 +200,12 @@ export async function reconcileAgentGraphSchedule(
     const committedRecords = new Map(
       snapshot.observation.projection.records.map((record) => [record.recordId, record]),
     );
+    for (const [recordId, record] of snapshot.selectedResultRecords) {
```
Astro-Han: **P1 — Flattening historical records here leaks one work's provenance into every other work.** If work A selects historical `(oldGraph, X)`, this global map makes an unrelated work B's ordinary `inputIds: ['X']` pass `missingWorkInputIds()` and render the historical record even though B never selected that source. For a dynamic operator, B passes this check and then `buildOperatorProvisionInput()` dereferences `X` only from the current projection, producing a reconciliation failure. Keep current records separate and resolve historical tuples per work; use the global view only for collision detection.

Astro-Han (commented):
The latest commits close most of the historical-result contract: the source graph is explicit, root/epoch/closed-graph/selected-result checks are fail-closed, and failed lookups are isolated and cached during reconciliation. One ownership boundary is still lost after authorization, however: historical tuples are flattened into a global record-id namespace shared by every scheduled work item.

The simplest first-principles fix is to keep two authorities throughout reconciliation: current-graph records keyed by `recordId`, and historical records keyed by each work's `(sourceGraphId, resultId)` selections. A global map may still detect collisions, but it must not authorize or render another work's inputs. This preserves provenance without adding a compatibility path.

Review performed with Codex reviewer agents and DeepSeek V4 Flash as advisory tools; I reproduced the failure against the latest head and verified live CI.

最新提交已经补齐大部分历史结果契约：source graph 显式携带，root/epoch/closed graph/selected result 均 fail-closed，失败查询也在 reconciliation 中隔离并缓存。但授权之后仍丢失了一层所有权：历史 tuple 被压平成所有 work 共用的全局 record-id namespace。

更符合第一性原理的最小修复是让 reconciliation 始终保留两个权威：当前图记录按 `recordId` 管理；历史记录按每个 work 明确选择的 `(sourceGraphId, resultId)` 管理。全局 map 可以继续用于碰撞检测，但不能授权或渲染其他 work 的输入。这样无需兼容分支即可保留 provenance。

本次审查使用了 Codex reviewer agents 与 DeepSeek V4 Flash 作为辅助工具；我已在最新 head 上复现该失败，并复核实时 CI。

</details>

packages/runtime/src/stream-graph-schedule-reconcile.ts
me2seeks: Fixed in ebedf52ce. Current records remain separate, while each scheduled work resolves only its selected (sourceGraphId, resultId) tuples; the global view is retained solely for collision detection. The regression uses two works sharing resultId X and proves provenance cannot cross-authorize or render the other work.

apps/desktop/src/renderer/agent-graph-panel.tsx
```diff
@@ -264,6 +304,29 @@ export function AgentGraphPanel(props: {
...
+              label={copy.epoch}
+              isLabelHidden
+              value={snapshot.graphId}
+              options={epochs.map((entry) => ({
+                value: entry.graphId,
+                label: `#${entry.epoch} · ${entry.current ? copy.currentEpoch : copy.historicalEpoch}`,
+              }))}
+              onChange={(graphId: SelectorOptionType) => {
+                if (typeof graphId !== 'string') return;
+                selectedGraphIdRef.current = graphId;
```
Astro-Han, line 327: [P2] Fence controls immediately when the user selects another epoch. This mutates only a ref; until the asynchronous refresh commits, `snapshot` and `selectedEpoch` still describe the current live graph, so `stopAvailable`/`dismissAvailable` remain true. A quick Stop after choosing a historical epoch calls the root-scoped `graphs.stop(rootSessionId)` and stops the current Graph the user just navigated away from. Mirror #2991's latest selectedGraphId/loading state and guard controls against `snapshot.graphId !== selectedGraphId`; add a delayed-history-read interaction test.

Astro-Han (commented):
Cross-epoch result admission belongs in the existing graph authority, and the PR correctly binds results to durable epoch identities instead of introducing another runtime state source. Two inherited history-surface gaps remain on this head. The smallest path is to rebase onto the corrected #2991 history foundation, then keep the selected graph identity as explicit UI state so controls cannot act through a stale snapshot during loading.

`storybook` is still running on the current head.

AI-assisted review disclosure: Codex verified the findings against the current head and traced epoch authority, root ownership, Desktop selection/refresh fencing, and root-scoped stop/dismiss actions. Two independent reviewer-agent passes and an OpenCode Go DeepSeek V4 Flash (high) adversarial pass were used as inputs. No local tests were run.

<details>
<summary>中文复核</summary>

跨 epoch result admission 应继续由现有 Graph 权威负责，这个 PR 正确地把结果绑定到持久 epoch identity，没有建立第二份 runtime 状态源。当前 head 还继承了历史浏览层的两个缺口。最小方案是先 rebase 到修正后的 #2991 foundation，再把 selected graph identity 作为明确 UI state，使加载期间的 control 不会通过旧 snapshot 作用到当前 Graph。当前 `storybook` 仍在运行。

本次为 AI 辅助审查：Codex 在最新 head 上追踪 epoch authority、root ownership、Desktop selection/refresh fencing 与 root-scoped stop/dismiss；另使用两次独立 reviewer 及一次 OpenCode Go DeepSeek V4 Flash（high）对抗审查。未运行本地测试。
</details>

packages/runtime/src/stream-graph-coordinator.ts
```diff
@@ -321,21 +301,57 @@ export class AgentGraphCoordinator {
...
     const epochs = await this.#input.epochStore.listAgentGraphEpochs(rootSessionId);
-    return epochs.length > 0 ? epochs.map(({ graphId }) => graphId) : [current.graphId];
+    return epochs.length > 0 ? epochs : [current];
+  }
+
+  async listGraphEpochPage(
+    rootSessionId: string,
+    options: { readonly beforeEpoch?: number; readonly limit: number },
+  ): Promise<AgentGraphEpochPage & { readonly currentEpoch: number }> {
+    requireRootSessionId(rootSessionId);
```
Astro-Han: [P2] Validate the existing root Session before listing or synthesizing epochs. This inherited path checks only ID syntax, so unknown or child Session IDs can receive epoch metadata (and an unknown ID may cause a virtual/current epoch to be resolved) before a later snapshot finally rejects it. Rebase onto the #2991 fix or call `#assertRootGraphReader` here, with unknown/child Session operation tests.

apps/desktop/src/renderer/agent-graph-panel.tsx
me2seeks: Fixed by restacking this PR onto #2991 head fa2daf4fb. The inherited selection fence updates selectedGraphId immediately and disables Stop/Dismiss while the requested graph is loading or while snapshot.graphId differs from the selected graph. The delayed historical-read interaction coverage from #2991 is now part of this branch.

packages/runtime/src/stream-graph-coordinator.ts
me2seeks: Fixed by restacking this PR onto #2991 head fa2daf4fb. The inherited Runtime Host path validates the existing root Session before listing or synthesizing epochs, and includes unknown-Session and child-Session operation coverage.

Astro-Han (approved):
This is now a coherent cross-epoch admission design. Historical results remain explicitly selected, same-root, earlier-epoch, source-finished, and RuntimeEvent-backed; `(sourceGraphId, resultId)` provenance survives authorization and reconciliation; current and historical records stay separated; and resolver failures defer only the affected source. Runtime Host and the durable schedule remain the existing authorities rather than introducing a parallel result store.

The test suite covers the important adversarial boundaries: cross-root and current/future epoch rejection, finish membership, recovery/reconciliation, provenance collisions, resolver failure caching, protocol limits, and read-only Desktop/TUI history. I found no concrete P0–P3 on the exact current head. Live required checks are green and merge state is clean.

Reviewed with Codex using three independent reviewer agents and OpenCode Go DeepSeek V4 Flash (high); I verified the exact head, current-main integration, authority boundaries, persistence/protocol contracts, and live CI.

当前方案已经形成一致的跨 epoch admission 设计。历史结果必须显式选择、同 root、来自更早 epoch、source 已 finish，并由 RuntimeEvent 记录支撑；`(sourceGraphId, resultId)` provenance 在授权和 reconciliation 中都被保留；当前与历史 records 不混用；resolver 失败只延后受影响 source。Runtime Host 与 durable schedule 继续作为既有权威，没有新增并行结果存储。

测试覆盖了关键对抗边界：跨 root、当前/未来 epoch 拒绝、finish membership、恢复/reconciliation、provenance 冲突、resolver 失败缓存、协议上限，以及 Desktop/TUI 历史只读。当前精确 head 未发现具体 P0–P3；实时必需检查全绿，merge state clean。

本次由 Codex 配合三个独立 reviewer agent，以及 OpenCode Go DeepSeek V4 Flash（high）审查；我核验了精确 head、与当前 main 的集成、权威边界、持久化/协议契约和实时 CI。

</details>

PR #2998, opened 2026-08-14
feat(desktop): add a default working directory

### LINKED ISSUE #2322: Add a configurable default working directory for sessions without a specific project
## Problem

When starting a session without a specific project, the working directory falls back to the last selected project root, `process.cwd()`, or the app path — none of which the user controls. Users who don't want tasks landing in a repo root or on the system drive (e.g. `C:\` on Windows) have no way to configure a catch-all directory. Real user request: "I keep a general-purpose directory (`~/agent`) for tasks that have no specific folder, and I want to configure that myself."

Under the hood, `createProjectRootController.current()` resolves as selected → persisted (`lastProjectPathFile`) → `fallbackRoots([process.cwd(), app.getAppPath()])`.

## Desired outcome

- A "Default working directory" setting under Settings → General, with a system folder picker and a clear/reset action.
- Sessions created without an explicit project/folder use this directory as their cwd.
- Resolution order: explicit per-session/project selection > configured default working directory > current fallback. The setting persists across restarts and app updates.
- It also serves as the fallback cwd for bot conversations (all IM channels) without a bound folder, tying in with the "mobile session management for the bot bridge" feature (#2325).

No workaround today — you must remember to pick a project each time. Reference implementations:

- **Hermes Agent** (desktop): `resolveHermesCwd()` — explicit configured dir (a JSON file in userData) > env hint > `INIT_CWD` > `process.cwd()` > repo root > home, skipping the install directory; the settings UI has a folder picker with set/clear.
- **OpenClaw**: `agents.defaults.workspace` in the global config, written by `openclaw setup --workspace <path>`; `OPENCLAW_WORKSPACE_DIR` env override; per-agent and per-session overrides.
- **PawWork** (negative example): the default directory is hardcoded (`~/PawWork`) in three places with no config entry, so tasks silently land there.

cat0825:
## Summary

- Add a persisted default working directory under Settings > General, with a native folder picker and clear action.
- Use it for local sessions and Bot conversations without a Project, while preserving explicit workspace choices and remote Runtime Host boundaries.

Fixes #2322

## Verification

- `npm run build:test`
- `npm run build`
- `npm --workspace @maka/core test` (547 passed)
- `npm --workspace @maka/desktop test` (841 passed)
- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- Storybook visual check at 1440px and 480px (`scrollWidth` matched the viewport at both sizes)

### Desktop

Default working directory setting at desktop width

### Minimum width (480px)

Default working directory setting at 480px

## Checklist

- [ ] No


M4n5ter:
<details open>

Thank you for putting this together, especially the persistence work, native folder picker, Runtime Host boundary handling, and test coverage. The implementation clearly explores the original request seriously.

While reviewing the current Desktop behavior, I noticed that the premise of #2322 has changed since the issue was opened. #2446 subsequently introduced a first-class default Project under **Settings → Projects**, and ordinary new conversations now resolve their workspace in this order:

1. An explicit Project or folder
2. The configured default Project
3. The previously selected workspace

A user can therefore already add a general-purpose directory such as `~/agent` as a Project and make it the default.

This PR still adds a distinct capability, but it is narrower: it provides a configurable cwd specifically for conversations that remain **Ungrouped / No Project**, and for Bot conversations when the current workspace has no Project identity. That also introduces a second “default location” alongside the existing default Project, with different behavior for ordinary and Bot conversations.

I think this is now primarily a product-model decision rather than an implementation gap. Before merging, could we continue the discussion in #2322 and confirm:

- Whether “Ungrouped / No Project with a configurable cwd” is an intentional, separately supported mode.
- Whether Bot conversations should instead honor the existing default Project.
- If both defaults are needed, their user-visible precedence and behavior with remote Runtime Hosts.

I have summarized the current overlap and the remaining product question in [this #2322 comment](https://github.com/maka-agent/maka-agent/issues/2322#issuecomment-5290887598).

For now, I suggest pausing the merge until that behavior is agreed upon. This is not a criticism of the implementation quality; the underlying requirement changed after #2446 landed, and we should avoid asking this PR to commit us to two overlapping defaults before confirming that distinction is useful.

</details>

<details>
<summary><strong>中文</strong></summary>

感谢你完成这项工作，特别是设置持久化、原生目录选择器、Runtime Host 边界处理和测试覆盖。可以看出，这份实现对原始需求进行了很认真、完整的探索。

不过，在复核 Desktop 当前行为时，我注意到 #2322 创建时所基于的前提已经发生了变化。此后合入的 #2446 已经在 **设置 → 项目** 中引入正式的“默认项目”功能，普通新对话目前会按以下顺序解析工作区：

1. 显式选择的项目或文件夹
2. 设置中的默认项目
3. 上一次选择的工作区

因此，用户现在已经可以把 `~/agent` 之类的通用目录添加为项目，并将它设为默认项目。

本 PR 仍然增加了一项有区别的能力，但范围更窄：它为保持在 **未归属 / 无项目** 状态的对话提供可配置 cwd，并在 Bot 当前工作区没有项目身份时作为其目录。与此同时，这也会在现有默认项目之外引入第二个“默认位置”，而且普通对话和 Bot 对话对两者的处理并不相同。

我认为这现在首先是一个产品模型决策，而不再只是实现缺口。在合并之前，是否可以先回到 #2322 继续讨论并确认：

- “未归属 / 无项目，但 cwd 可配置”是否是我们有意独立支持的模式。
- Bot 对话是否更应该直接遵循现有默认项目。
- 如果两个默认值都需要保留，它们对用户可见的优先级是什么，以及连接远程 Runtime Host 时应如何表现。

我已经在 [这条 #2322 评论](https://github.com/maka-agent/maka-agent/issues/2322#issuecomment-5290887598)中总结了当前功能重叠和仍需确认的产品问题。

目前建议先暂停合并，等待上述行为达成共识。这并不是对实现质量的否定；而是因为 #2446 合入后，底层需求已经发生变化。在确认这种区分确实有价值之前，我们不应通过本 PR 固化两个高度重叠的默认设置。

</details>

Astro-Han (commented):
Thanks—this solves the right product problem and mostly uses the right seam. Persisting the optional directory in the existing client-owned project settings and resolving it through `ProjectRootController` avoids a second workspace authority; the focused normalization and fallback tests are also useful. Two independent reviewer passes and a separate read-only DeepSeek V4 Flash high-effort pass reviewed this head.

The remaining issue is that an optional fallback preference has become an eager dependency of Project selection, and the UI exposes that local-only preference even while a remote Runtime Host is selected. A malformed/unreadable settings file can now reject controller initialization before the existing Project preference is read, while a remote user can successfully choose a local directory that the remote target is explicitly incapable of using.

From first principles, Project identity remains authoritative whenever it exists; the default directory is consulted only for the no-Project local case and must fail open to the established fallback roots. The smallest coherent solution is therefore to read/resolve the Project preference first, lazily consult the optional directory only when no Project is selected, catch preference-read failures at that fallback boundary, and gate this setting with the existing `setLocalDefault` capability. Reusing that capability is simpler than adding another target-specific settings path.

This branch also currently needs integration with the latest `main`; I treated merge conflicts as an integration requirement rather than mislabeling them as a production-code finding. I did not run a local test suite; this review is based on the latest head, focused static tracing, existing CI, two independent `@reviewer` reviews, and a read-only `ollama-cloud/deepseek-v4-flash` high-effort review. AI-assisted review; I verified the reported paths.

感谢，这个 PR 的问题定义和主要扩展点是对的：把可选目录放进已有的 client-owned project settings，并通过 `ProjectRootController` 解析，没有制造第二套 workspace authority；现有 normalization 和 fallback 测试也有价值。

剩余问题是：可选 fallback 偏好变成了 Project 选择的前置依赖；同时 remote Runtime Host 下仍暴露这个只对本地生效的设置。settings 文件损坏或暂时不可读时，controller 会在读取已有 Project preference 前失败；remote 用户则可以成功选择一个当前 target 明确不会使用的本地目录。

更符合第一性原理和奥卡姆剃刀的方案是：先恢复 Project identity；只有确实没有 Project 时才懒读取可选目录，并在该边界 fail open 到原有 fallback roots；UI 直接复用已有 `setLocalDefault` capability 做显示/操作门禁，不新增另一套 target-specific 设置路径。

该分支还需要与最新 main 集成；我把合并冲突视为集成要求，没有误标成生产代码 finding。本次未运行本地测试；结论来自最新 head 静态追踪、现有 CI、两个独立 reviewer 和一次只读的 DeepSeek V4 Flash high-effort 审查。AI 辅助审查；我已人工复核相关路径。

</details>

apps/desktop/src/main/project-root-controller.ts
```diff
@@ -73,11 +77,21 @@ export function createProjectRootController(
...
   deps: ProjectRootControllerDeps,
 ): Promise<CurrentProjectSelection> {
-  const fallbackPath = await resolveProjectRoot(deps.fallbackRoots());
+  const fallbackPath = await resolveUnassociatedRoot(deps);
   const preference = await readPreference(deps.preferenceFile, deps.rootId);
   return { projectId: preference, path: fallbackPath };
 }
 
+async function resolveUnassociatedRoot(deps: ProjectRootControllerDeps): Promise<string> {
+  const configured = await deps.defaultWorkingDirectory?.();
```
Astro-Han: **P2 — Keep the optional default directory from blocking Project recovery.** `loadInitialSelection()` calls this before `readPreference()`, and the injected callback directly awaits `settingsStore.get()`. Malformed JSON, permissions, or transient I/O therefore reject `initialSelection` before an existing valid Project ID is even read; `current()` and Bot workspace resolution then fail although the default directory is irrelevant to that Project. Please read the Project preference first and only consult this callback for the no-Project case, with callback failures treated like an unset preference and falling back to `fallbackRoots()`. Add a focused test for a rejecting callback with both an existing Project and no Project.

apps/desktop/src/renderer/settings/general-settings-page.tsx
```diff
@@ -319,11 +323,66 @@ function GeneralDefaultsCard(props: {
...
+      if (mountedRef.current) setSavingWorkingDirectory(false);
+    }
+  }
+
   return (
     <SettingsSection
       title={sections.chatDefaults}
       description={sections.chatDefaultsHelp}
     >
+      <SettingsRow
```
Astro-Han: **P2 — Gate this local-only control with the existing target capability.** Remote targets explicitly advertise `setLocalDefault: false`, and their `ProjectRootController` does not receive `defaultWorkingDirectory`; nevertheless this row is always rendered and its IPC opens a local Electron directory chooser for every target. A user connected to a remote Host can therefore save a local path successfully and reasonably believe it affects new remote sessions, while the remote target ignores it. Please render/enable this control only when the selected target exposes `setLocalDefault`, and add a local/remote capability rendering test.

Astro-Han:

Astro-Han:
Reviewed `bb51578b7bf2ff250f74edaaeb4898685fe9219d`. **No P0–P3 in the change itself** — but it is `CONFLICTING` against `main`, and the conflict hides a design question only you can answer.

### The review, for what it is worth on this head

- **Single decision point for the path.** The precedence chain is `selected Project` → configured default → fallback, resolved through `selectedProject ?? initialSelection` into `resolveUnassociatedRoot`. There is no second place that decides the working directory, which is the failure mode this kind of feature usually has.
- **Path handling holds up.** The default comes from settings storage rather than argv; relative paths are absolutized via `resolve`; empty string and non-existent paths fall back, and the tests cover both.
- **Scope is honest** — one new callback, no new persistence authority, and a large share of the diff is tests.

### Why I stopped instead of rebasing it for you

I attempted the rebase under maintainer authorization and aborted. **Nothing was pushed.** Two conflicts:

1. `packages/core/src/__tests__/settings.test.ts` — trivial (import path moved, `createDefaultSettings` added on `main`).
2. `apps/desktop/stories/settings/settings-pages.stories.tsx` — **not trivial, and it is a signal rather than noise.**

`main` has since split settings into two tiers, and the story bridge now reflects that:

```ts
let storyClientSettings = createDefaultSettings();
let storyRuntimeHostSettings = createDefaultSettings();

settings: {
  getClient: ..., updateClient: ...,   // client tier
  get: ...,       update: ...,          // runtime-host tier
}
```

This branch predates that split — it adds `chooseDefaultWorkingDirectory` to a single flat `settings` object. Textually I could drop the new method into `main`'s shape and get a compiling file. **But that would be me deciding which tier `projects.defaultWorkingDirectory` belongs to, and that is a product decision with a user-visible consequence:** client-tier makes the default *per-machine*, runtime-host-tier makes it *shared across every client of that Host*. For a working **directory**, which only exists on one filesystem, that distinction matters — and it likely affects `packages/core/src/settings.ts` too, which auto-merged only because the split did not touch the lines this branch edits.

So: **please rebase and place `defaultWorkingDirectory` in the tier you intend.** Happy to take the mechanical parts off your hands afterwards, or now if you tell me which tier — just say so on this thread.

**One more thing to redo after the rebase:** the green checks on this PR are from 2026-08-14 against the old base. **They say nothing about the current tree**, and the review conclusion above is bound to `bb51578b` only — it does not carry across the rebase, particularly given the settings split.

Astro-Han:

- `apps/desktop/src/main/runtime-host-settings-ipc-main.ts`
- `apps/desktop/src/preload/bridge-contract.d.ts`
- `…(more)`

```
# resolve, then
git push --force-with-lease
```

---

Astro-Han (commented):
NO-GO at exact head `bb51578b7bf2ff250f74edaaeb4898685fe9219d`. One P1 is inline. Two existing, non-outdated unresolved P2 threads also remain: initial selection resolves the optional default-directory callback before reading an existing Project preference (so a settings read failure can block a valid Project selection), and the local-only default-directory setting is rendered for remote Runtime Hosts that cannot apply it. The branch is 319 commits behind current `main`; a real rebase probe conflicts in `apps/desktop/src/main/runtime-host-settings-ipc-main.ts`, `apps/desktop/src/preload/bridge-contract.d.ts`, `apps/desktop/src/preload/preload.ts`, and `apps/desktop/src/renderer/locales/settings-preferences-copy.ts`. Preserve current-main IPC/preload/localization authority while replaying the feature. The old exact-head checks are green, but they do not validate the required rebased integration.

apps/desktop/src/main/project-root-controller.ts
```diff
@@ -39,8 +41,10 @@ export function createProjectRootController(
   const initialSelection = loadInitialSelection(deps);
 
   async function currentSelection(): Promise<CurrentProjectSelection> {
-    if (selectedProject) return selectedProject;
-    return (selectedProject = await initialSelection);
+    const selection = selectedProject ?? (selectedProject = await initialSelection);
```
Astro-Han: [P1] Do not let initial resolution overwrite a newer explicit selection. `setSelection()` updates `selectedProject` synchronously, but if this `await initialSelection` is already pending, its continuation still assigns the stale unassociated selection here after the explicit Project selection has been persisted. A caller can therefore successfully select a Project while the active controller silently falls back to the default directory. Re-check `selectedProject` after the await (and likewise after the later default-root await), or fence both continuations with a generation/ticket; add a deferred-resolution regression that calls `setSelection()` before each await settles.

cat0825:
Thanks for the careful read — all three findings are addressed, and each one now has a test that fails if the fix is reverted. Rebased onto current `main` as you asked.

### P1 — a selection made during resolution could be overwritten by the pending default

`currentSelection()` awaited twice (the preference read, then the working-directory resolution) and then unconditionally assigned its result. An explicit `setSelection` landing in either window was silently clobbered by the stale continuation, so the caller was told the Project selection succeeded and then got the default directory anyway.

`createProjectRootController` now keeps a `selectionGeneration` counter, bumped synchronously in `setSelection`. `currentSelection()` captures it on entry and re-checks after each await; if the generation moved and a selection now exists, the newer selection wins and the stale continuation returns without assigning.

Pinned by `a selection made during resolution is not overwritten by the pending default` in `project-root-controller.test.ts`. It gates the default-directory callback on a promise, calls `setSelection` while resolution is parked, then releases — covering both await points. Removing the fence fails it.

### P2a — a settings-read failure blocked Project recovery

`loadInitialSelection` resolved the `defaultWorkingDirectory` callback on the recovery path, so a rejecting callback took down a perfectly valid stored Project selection. That is backwards: the default directory only matters when there is no Project.

`loadInitialSelection` now reads the preference file and `fallbackRoots()` only, and never touches the callback. `resolveUnassociatedRoot` remains the single place that decides a path for the no-Project case and treats a rejecting callback exactly like an unset preference (`.catch(() => undefined)`), then falls through to `fallbackRoots()` — it is a fallback boundary, so it fails open rather than propagating.

Pinned by `a rejecting default-directory callback does not block Project recovery`. Dropping the `.catch` fails it.

Your point about the single decision point is preserved: `selectedProject ?? initialSelection` still feeds `resolveUnassociatedRoot`, and nothing else in the change derives a path.

### P2b — the control was offered on targets that cannot honour it

The directory only exists on one filesystem, so it is a per-machine preference. Two things changed:

1. **Capability gate.** The row is gated on `setLocalDefault` from the project snapshot — the same seam `projects-settings-page.tsx` already uses for its own local-only default, rather than a new capability channel for one row. A remote Host reports `false`, and a *failed* capability read also yields `false`: a control whose target may ignore it is worse than no control, so it fails closed.
2. **Tier.** The folder picker moved off the per-target runtime-host settings channel onto the persistent client channel (`settings:client:chooseDefaultWorkingDirectory`). Current `main` already treats `projects` as client-owned — `clientOwnedSettingsPatch` forwards `patch.projects` and `projectClientOwnedSettings` reads `client.projects` — so the value is per-machine by construction, and `hasRuntimeHostSettingsPatch` already routes the save through `updateClient`.

The help copy in both locales now states the preference applies to a local Runtime Host.

Pinned by `general-settings-default-working-directory.test.ts` (8 tests): row shown for a local target, hidden for a remote one, hidden on a failed capability read, hidden for an unverified target, save patches the client-owned `projects` section, a cancelled picker does not clear the value, clear sends `undefined` without opening a picker, and no Clear button when nothing is configured. Inverting the failure branch to `true` fails the third.

### Rebase

595 commits behind; rebased onto `upstream/main`, seven files conflicted:

- **`runtime-host-settings-ipc-main.ts`** — resolved to upstream **exactly** (empty diff). Moving the picker to the client tier for P2b removed the reason to touch this file at all, which is what made the worst conflict disappear. Its test file is likewise byte-identical to `main`; the handler test moved into the existing `client-settings-ipc-main.test.ts` rather than a new file.
- **`bridge-contract.d.ts` / `preload.ts`** — kept `main`'s two-tier `settings` shape and added `chooseDefaultWorkingDirectory()` to it. The PR's line for `settings:usageStats` was dropped: that moved to `runtime-host-usage-ipc-main.ts` on `main`.
- **`settings-preferences-copy.ts`** — type keys merged; zh and en copy re-written onto `main`'s current "tasks" vocabulary rather than the older wording the branch was written against.
- **`general-settings-page.tsx`** — resolved onto `main`'s shape. The row moved into its own `default-working-directory-row.tsx` (row + capability hook), which keeps the page diff purely additive and lets the behaviour be tested without dragging the whole page's asset graph in.
- **`settings-pages.stories.tsx`** — kept `main`'s profile block and `commitClientRead` seeding; added the picker stub and a `projects` fixture with `setLocalDefault: true`.
- **`packages/core/src/__tests__/settings.test.ts`** — trivial as you said: import list from `main`, both test groups kept.

`packages/core/src/settings.ts` stays minimal — one optional field on `ProjectPreferencesSettings` plus the matching normalize branch. No restyling anywhere; no dependency changes.

### Verification

```
$ npm --workspace @maka/desktop run build:main
> tsc -p tsconfig.main.json
(clean)

$ node --test dist/main/__tests__/general-settings-default-working-directory.test.js \
    dist/main/__tests__/project-root-controller.test.js \
    dist/main/__tests__/client-settings-ipc-main.test.js \
    dist/main/__tests__/runtime-host-settings-ipc-main.test.js
ℹ tests 19
ℹ pass 19
ℹ fail 0
```

Each fix was also confirmed to be load-bearing by reverting it and watching exactly the intended test go red.

One thing worth flagging on `typecheck`: `npm --workspace @maka/desktop run typecheck` reports 3 pre-existing errors in `packages/ui` (`chat-surface-layout.tsx:87`, `markdown-body.tsx:180`, `session-history-list.tsx:326`). They reproduce identically on a clean `upstream/main` worktree, which reports 14 errors in total across `packages/ui` and `packages/core` — this branch's 3 are a strict subset, so nothing here introduced or masked any of them. Happy to leave that alone as out of scope, but let me know if you'd rather I open a separate issue.

@Astro-Han when you have a moment, could you take another look? Glad to adjust anything, particularly the choice to gate on `setLocalDefault` rather than adding a dedicated capability — that was the seam that already existed, but it's your call whether it should be its own flag.

cat0825:
Rebased onto `main` (`9ff2675`) and reworked the change to satisfy the renderer architecture ledger from #4088. All three CI jobs are green.

The ledger forbids a new legacy-zone renderer file becoming reachable from AppShell, so `settings/default-working-directory-row.tsx` could not stay as its own module. Changes:

- The row is folded into `GeneralDefaultsCard`, which already calls every hook it needed (`useUiLocale`, `useToast`, `useMountedRef`, `useKeyedActionGuard`, `useState`). It reuses the existing `saving` state and the keyed guard's new `working-directory` key, so hook debt is unchanged.
- The Desktop capabilities moved behind a new `platform/desktop/settings-surface-capabilities.ts` adapter, so no Settings page reaches for `window.maka` or web storage for this row.
- The capability gate no longer costs a project-snapshot round-trip. `runtime-host-boot.ts` derives `setLocalDefault` as `!usesHostWorkspace`, so the selected profile's kind answers the same question via `runtimeHostProfileUsesHostWorkspace`. There's a test asserting the two stay in agreement across every profile kind.
- Icons dropped from the two buttons to match the sibling `SettingsActions` rows in the same file.

Both touched debt files sit exactly at their base budgets — the two ledger entries are same-count dependency replacements (`@maka/ui` 2→1 in the general page freed the slot; `../browser-storage` → the adapter in the surface).

One checker change worth your review: `allowsMigrationDependency` accepted `platform` targets for `rootDebtClosure` but not for `legacyAppShellClosure`, while a Desktop adapter is the only zone permitted to own bridge and browser capabilities. That combination makes "move a capability out of legacy renderer code and behind an adapter" unrepresentable in the AppShell closure, even though it strictly reduces debt. I extended the AppShell branch to accept `platform` for the same reason the root closure already did, and added a checker test pinning it. Happy to split that into its own PR if you'd prefer it reviewed separately.

Verification: `check:renderer-architecture -- --base upstream/main` passes; checker suite 63/63; desktop suite 1706 tests, 1704 pass with the 2 `streaming-handoff.test.ts` failures reproducing identically on a clean `upstream/main` worktree; `lint`, `format:check`, main/preload typecheck clean; astryx surface inventory regenerated.

@Astro-Han when you have a moment.

apps/desktop/src/main/project-root-controller.ts
```diff
@@ -55,11 +57,24 @@ export function createProjectRootController(
...
-    return (selectedProject = await initialSelection);
+    const generation = selectionGeneration;
+    const base = selectedProject ?? (await initialSelection);
+    // `setSelection()` writes `selectedProject` synchronously, so an explicit
+    // Project can be installed while either await here is pending. Its value
+    // is newer than anything this call computed: committing the stale
+    // unassociated selection would leave the active session on the default
+    // directory after the caller was told the Project selection succeeded.
+    if (selectionGeneration !== generation && selectedProject) return selectedProject;
+    if (typeof base.projectId === 'string') return (selectedProject = base);
```
Astro-Han, line 74: Thanks for separating Project identity from the optional fallback. One category ② recovery path still seems inconsistent with that priority: when the persisted Project ID names a Project that has since been deleted, archived, or become unavailable, this branch returns the initial fallback-root path before the default directory is resolved. `resolveDesktopSessionWorkspace()` then calls `ProjectManagementService.currentSelection()`, which clears the stale ID but returns that same path, so the first new Session or Bot is durably created in `process.cwd()`/the app fallback rather than the user-configured default; only later calls recover. Could the stale-ID branch clear the selection and then re-enter this controller’s no-Project resolution, and add a real controller + stale catalog + configured-default regression? I believe this is P2 on a reasonable recovery path, but please push back if stale catalog entries have a stronger invariant than I found.

Astro-Han (commented):
Thanks for the substantive follow-up here. One recovery edge still appears to bypass the configured default working directory. This is an AI-assisted review; I independently traced the reported production path. Suggestions are offered from an outside perspective, so please do push back if there is an invariant or ownership constraint I missed.

apps/desktop/src/main/project-root-controller.ts
cat0825: Fixed in `1a3d366e` (with the current-`main` integration at `c5c57e8a`). When a persisted Project ID is no longer selectable, `ProjectManagementService.current()` now awaits clearing it and re-enters the same controller `currentSelection()` path, so the first recovery resolves the configured no-Project default instead of returning the stale fallback path. Added a real `ProjectRootController` + stale catalog + configured-default regression; it failed with the fallback path before the fix and passes afterward. Focused controller/service tests: 15/15; `build:main`: clean; renderer architecture: 63/63 and passed against `ef94235b`.\n\n_Automated follow-up by OpenAI Codex; not a human review._

cat0825:
@Astro-Han — rebased onto current `main` (`2c9c61d5`); `mergeable` is back to `MERGEABLE`. The previous `upstream/main` merge commit was dropped by the rebase, so the branch is linear again (5 commits).

Conflicts were confined to generated ratchet files; resolved by re-running the repo generators against the rebased tree rather than hand-merging:

- `docs/astryx-surface-file-inventory.md` + `.paths` — `npm run astryx:surface-inventory:write`; coverage check passes (236 files, no blockers, no reimplementation). Note the standalone `default-working-directory-row.tsx` is intentionally absent — commit 3 folds the row into `general-settings-page.tsx`.
- `apps/desktop/renderer-architecture.json` — `node apps/desktop/scripts/check-renderer-architecture.mjs --write`; `check:renderer-architecture` passes.

CI is re-running on the rebased branch.

PR #3001, opened 2026-08-14
fix(runtime): close the path-replacement window for filesystem mutations (#2600)

### LINKED ISSUE #2600: fix(runtime): harden filesystem mutations against external path replacement
Maka serializes cooperative file mutations by canonical path, but Bash or external processes can replace that path between lock resolution and mutation. This affects Write/Edit/FormatJson/ApplyPatch; delete may unlink the replacement, and worker disconnects may leave the outcome unknown.

Define a shared filesystem-authority contract for stable target identity, compare-and-delete/update, and post-dispatch unknown outcomes. Keep this separate from individual editing tools.

chinawch007:
## Summary

Closes #2600. Built directly around the three concerns the issue names, plus the "shared filesystem-authority contract, kept separate from individual editing tools" the author asked for. Four commits, each independently healthy (builds and tests green at every step):

### Commit 1 — `26f7fa2` feat(runtime): surface worker disconnect as unknown outcome for mutations

Addresses *"worker disconnect may leave the outcome unknown"*.

A filesystem mutation whose worker fails after dispatch may already have landed on disk, but the host reported it as a generic error — the model had no way to tell "the write may have happened" from "nothing ran", so it treated a half-applied mutation as a clean failure.

- `process-runner` tracks a `dispatched` flag (set once Node's `'spawn'` event fires and stdin is written) and surfaces it on both the resolved result and the rejection error.
- `client` splits the ambiguous launch failures by that flag: `spawn_failed` (never started, nothing could have been written) vs new `worker_io_incomplete` (ran but the result was lost). `aborted` carries `dispatched` so a pre-flight cancel (ESC) is a clean cancel, not an unknown outcome — a single ESC would otherwise mark every queued file tool `outcome_unknown` with `retrySafe: false`.
- Protocol gains an `outcome_unknown` error code so the worker can report "I may have applied this before I lost the ability to answer."
- The boundary executor converts a mutating op that fails with a post-dispatch reason into `ToolOutcomeUnknownError`, which flows to the existing structured `uncertainOutcome` result. Reads and pre-flight failures pass through unchanged.

### Commit 2 — `aae3138` refactor(runtime): extract filesystem-authority contract

Implements *"Define a shared filesystem-authority contract … Keep this separate from individual editing tools."*

New `packages/runtime/src/filesystem-authority.ts` declares the contract types, imported by the executor, worker, and workspace-executor alike — types and a pure classifier only, no I/O:

```ts
type FilesystemTargetIdentity = { dev: string; ino: string };  // opaque decimal strings

type FilesystemTargetDescriptor =
  | { enforcementPath: string; targetType: 'missing' }
  | { enforcementPath: string; targetType: 'file' | 'directory' | 'symlink' | 'other';
      identity: FilesystemTargetIdentity };

type FilesystemMutationOutcome = 'applied' | 'rejected' | 'unknown';
```

The identity is a decimal string, not bigint, because bigint cannot cross the worker's JSON protocol boundary (`JSON.stringify` throws on a BigInt), and identity is only compared for equality. The descriptor is a discriminated union so "the target had no identity to compare" is an explicit `missing` arm — a future "skip the identity check" change cannot compile without handling it, which compiles the "no identity → CAS passes" regression class out of existence.

### Commit 3 — `b51c6a5` feat(runtime): capture target identity at lock acquisition for CAS

Addresses *"Write/Edit/FormatJson/ApplyPatch may write to the replacement"* — the core concern.

Root cause: `expectedTarget` was computed **after** the write lock was acquired, not when the lock key was resolved. The lock serializes concurrent Maka tools, but it does nothing while a call is *queued*, and Bash is not serialized at all — so in the window between lock resolution and the actual write, an external process could replace the path and we mutated whatever we then found.

- The target identity (inode/device, opaque strings) is now captured at **lock acquisition (T0), before entering the lock queue**, by `captureIdentityAtLockAcquisition` in the boundary executor. The stat mode matches the targetType derivation: content operations `stat` (follow), create/delete `lstat` (pin the directory entry, so a swapped link is caught against the link's own inode).
- Protocol v6: `FilesystemWorkerTargetSchema` gains the optional `identity` field; `superRefine` rejects a missing target carrying one.
- `client.execute` receives the T0 identity and passes it verbatim — it does **not** re-derive it (re-deriving after the lock is held would sample the post-replacement inode and make the CAS self-fulfilling).
- Worker `assertTargetUnchanged` compares the on-disk inode against the T0 identity and rejects with `path_changed`. A non-missing **write** target must carry an identity or the request is rejected outright — the CAS cannot be silently skipped (reads are exempt; they do not mutate).
- Post-write orphan check (`assertPathStillMatchesIdentity`): after a write, if the path's inode no longer matches the one written, the bytes went to an orphaned inode and the visible file is the replacement — reported as `outcome_unknown`, not a misleading success.

So: a path replaced while queued → `path_changed`, the replacement is untouched. A path replaced mid-write → `outcome_unknown`, the model is told to re-read.

### Commit 4 — `2d82471` feat(runtime): extend post-write orphan check to edit/format/update

Extends the commit-3 post-write check symmetrically to `edit`, `format_json`, and `apply_patch update`. Delete is already covered by the T0 CAS (it runs for every operation; the symlink entry case uses `lstat` on the link's own inode), and `format_json`'s invalid-JSON branch already returns `ok:false` without writing.

### Honesty about what this does *not* close

Recorded as documented residuals rather than over-claimed:

- **Delete window**: POSIX has no atomic compare-and-unlink; `lstat`-compare + `unlink` narrows the window rather than closing it. (We deliberately do **not** `open()` before unlink — `open('r+')` needs read+write permission on the file while `unlink` only needs the parent dir, so an open precondition would break deleting read-only/write-only files.)
- **Same-inode in-place modification** (`echo x >> file`): the inode never changes, so CAS passes; out of scope (content-level detection).
- **Hard-link concurrent writes** to one inode: the issue is specifically about path replacement, lexical lock keying is a documented existing tradeoff in `file-write-lock.ts`, and CAS cannot catch two RMWs on one inode. Left as separate work.
- No auto-retry on `path_changed`: content matching an `oldString` proves the snippet exists, not that it is the same file — retrying would reproduce the exact defect this issue describes (`mv other.ts target.ts` where both contain the string). We return a clear "the target was replaced while queued; re-read before editing" and let the model do `Read` → `Edit`.

New tests added (29 total):

**`filesystem-mutation-outcome.test.ts` (14)**
- Every post-dispatch reason (`worker_crashed`, `worker_io_incomplete`, `timeout`, `response_overflow`, `invalid_response`, `response_id_mismatch`, `response_kind_mismatch`, `outcome_unknown`) converts a mutating op to `ToolOutcomeUnknownError`.
- `spawn_failed` (never dispatched) does **not** convert — nothing could have been written.
- `aborted` before dispatch is a clean cancel; only after dispatch is it unknown.
- Pre-flight validation failures and reads pass through as ordinary errors.
- **Red-line**: the worker receives the T0 identity (captured before a replacement), proving the capture happens at lock acquisition — not re-derived after the lock.

**`filesystem-target-identity.test.ts` (9)**
- Write to a replaced inode → `path_changed`, and the replacement's content is **not** overwritten.
- Delete of a replaced inode → `path_changed`, and the replacement file is **not** deleted.
- Edit of a replaced inode → `path_changed`, and the replacement content is untouched.
- Unchanged inode → write applies normally; missing target (create) → no identity, no false rejection.
- Post-write orphan check, unit-tested directly: path still matches → pass; path replaced after write → `outcome_unknown`; path disappeared → `outcome_unknown`.

**`filesystem-worker-client.test.ts` (3)**
- Post-dispatch runProcess rejection → `worker_io_incomplete` with `dispatched: true`.
- Never-dispatched rejection → `spawn_failed` with `dispatched: false`.
- Rejection with no flag (spawn itself threw) → treated as never-dispatched.

**`filesystem-authority-contract.test.ts` (9)**
- Classifier branches (non-worker error, aborted by dispatched flag, every whitelist member, spawn failure, pre-flight validation).
- Type-level constraints: identity is string fields, a missing descriptor carries no identity, an existing descriptor requires one, `MutationOutcome` is the three-valued set.

## Verification

Rebased on latest `main`; every check run against a clean-`main` baseline to confirm nothing pre-existing was disturbed.

- `npm run format:check` — clean (8 formatting nits auto-fixed and folded in)
- `npm run lint` — 2304 files, no issues
- `npm run typecheck` 
- `npm --workspace @maka/runtime run build` — clean
- `npm --workspace @maka/runtime run test:dist` — 2983 tests, 2971 pass, 12 skip

## AI use

Select exactly one:

Tool(s) and scope: ZCode (AI coding agent) implemented the runtime, filesystem-worker
protocol, filesystem-authority contract, and test changes; it also designed the
implementation plan through review iterations, resolved rebase conflicts against main,
and ran verification (format, lint, typecheck, build, and the runtime test suite with
clean-main baselines). The human contributor of record set the scope decisions, relayed
external review feedback for evaluation and incorporation, reviewed and approved each
commit, and chose to submit the result. AI review does not constitute the independent
human review this change (runtime behavior) requires.

## Checklist

- [ ] No

### DIFF
--- a/packages/runtime/src/file-stable-write.ts
+++ b/packages/runtime/src/file-stable-write.ts
@@ -1,0 +1,184 @@
+// packages/runtime/src/file-stable-write.ts
+// The fd-pinned mutation primitive enforcing the filesystem-authority contract
+// (#2600). A path-based write re-opens the pathname after every check, so a
+// swap between validation and the open diverts the write onto the replacement
+// and the post-write check can only *report* the corruption afterwards. This
+// module removes that window structurally: the approved object is opened once,
+// its identity is validated on the descriptor itself (fstat, not a second
+// pathname lookup), and the read/transform/write all run through that same
+// descriptor — a path swap mid-operation cannot redirect the bytes.
+//
+// Missing targets are created with `wx` (exclusive): if the path appeared
+// between authorisation and the open, EEXIST is reported as `path_changed`
+// rather than truncating whatever landed there.
+//
+// The descriptor is validated BEFORE any truncation: an 'r+' open does not
+// truncate, so a rejected validation leaves the file byte-for-byte intact —
+// unlike a plain 'w' open, which truncates as part of opening.
+
+import { lstat, open, stat, type FileHandle } from 'node:fs/promises';
+import { constants } from 'node:fs';
+
+import type { FilesystemTargetIdentity } from './filesystem-authority.js';
+
+/** The two failure modes this primitive can report. */
+export type StableWriteErrorCode = 'path_changed' | 'outcome_unknown';
+
+export class StableWriteFailure extends Error {
+  constructor(
+    readonly code: StableWriteErrorCode,
+    message: string,
+  ) {
+    super(message);
+    this.name = 'StableWriteFailure';
+  }
+}
+
+function pathChanged(message: string): StableWriteFailure {
+  return new StableWriteFailure('path_changed', message);
+}
+
+/**
+ * Open the approved target and validate its identity on the descriptor.
+ *
+ * - Existing target (approvedIdentity present): `open(path, 'r+' | O_NOFOLLOW)`
+ *   — POSIX refuses a final symlink at open time, and 'r+' does not truncate,
+ *   so a failed identity check leaves the file untouched. Windows has no
+ *   O_NOFOLLOW; the link is detected with an lstat just before the open and
+ *   the residual window is closed by the identity comparison on the fd.
+ * - Approved-missing target (approvedIdentity undefined): `open(path, 'wx')` —
+ *   atomic create-if-absent. EEXIST means something appeared in the gap.
+ */
+export async function openStableTarget(input: {
+  path: string;
+  approvedIdentity: FilesystemTargetIdentity | undefined;
+}): Promise<FileHandle> {
+  const noFollow = process.platform === 'win32' ? 0 : constants.O_NOFOLLOW;
+  if (input.approvedIdentity) {
+    if (process.platform === 'win32') {
+      const entry = await lstat(input.path).catch(() => null);
+      if (entry?.isSymbolicLink()) {
+        throw pathChanged(
+          'The approved filesystem target is a symbolic link; refusing to follow it.',
+        );
+      }
+    }
+    let handle: FileHandle;
+    try {
+      handle = await open(input.path, constants.O_RDWR | noFollow);
+    } catch (error) {
+      const code = (error as NodeJS.ErrnoException).code;
+      if (code === 'ELOOP' || code === 'ENOTDIR' || code === 'ENOENT') {
+        throw pathChanged('The approved filesystem target changed before execution.');
+      }
+      if (code === 'EACCES' || code === 'EPERM') {
+        // A write-only target (e.g. mode 0o222) refuses 'r+' but is still a
+        // legitimate mutation target — unlink semantics need no read
+        // permission. Retry write-only (still no truncate: identity is
+        // validated on the descriptor before writeThroughHandle truncates).
+        // The pinned read of the previous content will fail and the caller
+        // reports 'unknown' (no diff), which is the pre-existing behaviour.
+        try {
+          handle = await open(input.path, constants.O_WRONLY | noFollow);
+        } catch (retry) {
+          const retryCode = (retry as NodeJS.ErrnoException).code;
+          if (retryCode === 'ELOOP' || retryCode === 'ENOTDIR' || retryCode === 'ENOENT') {
+            throw pathChanged('The approved filesystem target changed before execution.');
+          }
+          throw retry;
+        }
+      } else {
+        throw error;
+      }
+    }
+    // The compare in compare-and-update, performed on the descriptor itself.
+    const metadata = await handle.stat({ bigint: true });
+    if (
+      String(metadata.dev) !== input.approvedIdentity.dev ||
+      String(metadata.ino) !== input.approvedIdentity.ino
+    ) {
+      await handle.close();
+      throw pathChanged('The approved filesystem target changed before execution.');
+    }
+    return handle;
+  }
+  try {
+    return await open(input.path, constants.O_WRONLY | constants.O_CREAT | constants.O_EXCL);
+  } catch (error) {
+    if ((error as NodeJS.ErrnoException).code === 'EEXIST') {
+      throw pathChanged('A file appeared at the approved missing target; re-read before writing.');
+    }
+    throw error;
+  }
+}
+
+/**
+ * Write `content` through the pinned descriptor. The truncation happens here,
+ * only after the identity was validated on the fd. Write-step failures
+ * (ENOSPC/EIO/EDQUOT/EFBIG) can leave the file truncated or half-written, so
+ * they surface as `outcome_unknown` — the file's state is genuinely unknown.
+ */
+export async function writeThroughHandle(handle: FileHandle, content: string): Promise<void> {
+  try {
+    await handle.truncate(0);
+    // Position 0 explicitly: a prior readFile leaves the fd position at EOF,
+    // and a positionless write would create a NUL-prefixed sparse file.
+    await handle.write(content, 0, 'utf8');
+  } catch (error) {
+    const code = (error as NodeJS.ErrnoException).code;
+    if (code === 'ENOSPC' || code === 'EIO' || code === 'EDQUOT' || code === 'EFBIG') {
+      throw new StableWriteFailure(
+        'outcome_unknown',
+        'The write started but failed partway; the file may be truncated. ' +
+          'Re-read the file before writing to it again.',
+      );
+    }
+    throw error;
+  }
+}
+
+/**
+ * Read-modify-write entirely through one descriptor. Transform errors happen
+ * before any truncation and propagate unchanged (nothing was written). An
+ * unchanged result skips the write entirely, preserving mtime for no-ops.
+ */
+export async function readModifyWriteThroughHandle(
+  handle: FileHandle,
+  transform: (existing: string) => string,
+): Promise<string> {
+  const existing = await handle.readFile('utf8');
+  const replacement = transform(existing);
+  if (replacement !== existing) {
+    await writeThroughHandle(handle, replacement);
+  }
+  return replacement;
+}
+
+/**
+ * After writing through the handle, describe the host-visible outcome: if the
+ * path no longer resolves to the descriptor's inode, the write went to an
+ * orphaned inode and the visible file is the replacement — an unknown outcome.
+ * `undefined` means the path still matches and the write is host-visible.
+ */
+export async function hostVisibilityAfterWrite(
+  path: string,
+  handle: FileHandle,
+): Promise<StableWriteFailure | undefined> {
+  const written = await handle.stat({ bigint: true });
+  let current: { dev: bigint; ino: bigint };
+  try {
+    current = await stat(path, { bigint: true });
+  } catch {
+    return new StableWriteFailure(
+      'outcome_unknown',
+      'The target disappeared after the write; the outcome on disk is unknown.',
+    );
+  }
+  if (String(current.dev) !== String(written.dev) || String(current.ino) !== String(written.ino)) {
+    return new StableWriteFailure(
+      'outcome_unknown',
+      'The target was replaced during the write; the outcome on disk is unknown.',
+    );
+  }
+  return undefined;
+}
--- a/packages/runtime/src/file-stable-write.ts
+++ b/packages/runtime/src/file-stable-write.ts
@@ -16,8 +16,10 @@
 // truncate, so a rejected validation leaves the file byte-for-byte intact —
 // unlike a plain 'w' open, which truncates as part of opening.
 
-import { lstat, open, stat, type FileHandle } from 'node:fs/promises';
+import { randomUUID } from 'node:crypto';
+import { lstat, open, rename, stat, unlink, type FileHandle } from 'node:fs/promises';
 import { constants } from 'node:fs';
+import { dirname, join } from 'node:path';
 
 import type { FilesystemTargetIdentity } from './filesystem-authority.js';
 
@@ -181,4 +183,66 @@ export async function hostVisibilityAfterWrite(
     );
   }
   return undefined;
+}
+
+/**
+ * Compare-and-delete for directory entries (#2600). POSIX has no atomic
+ * compare-and-unlink, so a bare `unlink(path)` can remove a replacement
+ * installed between the identity check and the unlink — and report success.
+ * Instead, atomically grab whatever the path currently names by renaming it to
+ * a private tombstone in the same directory, verify the tombstone carries the
+ * approved identity, and only then unlink the tombstone. A mismatch means a
+ * replacement was installed in the window: it is renamed back to the path
+ * (restored, not deleted) and the operation reports `path_changed`.
+ *
+ * rename(2) moves the directory entry itself, so this works for regular files
+ * and symlinks alike, needs no permission on the file (only write+execute on
+ * the parent directory, exactly like unlink), and never follows a symlink.
+ */
+export async function compareAndDeleteEntry(input: {
+  path: string;
+  approvedIdentity: FilesystemTargetIdentity | undefined;
+}): Promise<void> {
+  const tombstone = join(dirname(input.path), `.maka-pending-delete-${randomUUID()}`);
+  // Atomically capture whatever entry the path names right now.
+  await rename(input.path, tombstone);
+  try {
+    if (input.approvedIdentity) {
+      const entry = await lstat(tombstone, { bigint: true }).catch(() => null);
+      const matches =
+        entry !== null &&
+        String(entry.dev) === input.approvedIdentity.dev &&
+        String(entry.ino) === input.approvedIdentity.ino;
+      if (!matches) {
+        // A replacement was installed after the check. Restore it — the path
+        // is free because its entry is sitting on the tombstone — and refuse.
+        let restored = true;
+        try {
+          await rename(tombstone, input.path);
+        } catch {
+          restored = false;
+        }
+        throw pathChanged(
+          restored
+            ? 'The approved filesystem target changed before the delete; the replacement was restored. Re-check the directory before deleting again.'
+            : `The approved filesystem target changed before the delete; the replacement is preserved at ${tombstone}.`,
+        );
+      }
+    }
+    // The tombstone is a private unpredictable name: nothing else contends
+    // with this unlink.
+    try {
+      await unlink(tombstone);
+    } catch {
+      // The approved entry was moved but not removed: the delete's outcome on
+      // disk is genuinely unknown, and the tombstone preserves the content.
+      throw new StableWriteFailure(
+        'outcome_unknown',
+        `The delete may not have completed; the entry is preserved at ${tombstone}.`,
+      );
+    }
+  } finally {
+    // Nothing to clean up: the tombstone is either unlinked or deliberately
+    // preserved (restored to the path, or kept as the reported location).
+  }
 }
--- a/packages/runtime/src/file-stable-write.ts
+++ b/packages/runtime/src/file-stable-write.ts
@@ -17,7 +17,7 @@
 // unlike a plain 'w' open, which truncates as part of opening.
 
 import { randomUUID } from 'node:crypto';
-import { lstat, open, rename, stat, unlink, type FileHandle } from 'node:fs/promises';
+import { link, lstat, open, rename, stat, unlink, type FileHandle } from 'node:fs/promises';
 import { constants } from 'node:fs';
 import { dirname, join } from 'node:path';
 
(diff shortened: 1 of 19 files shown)


Astro-Han:
Thanks for the fix — the race is real (an external Bash/process can swap the path between lock-key resolution and the actual write, Delete can remove the replacement, and a worker disconnect leaves the outcome unknown), and the three-claim contract (T0 identity capture before entering the write-lock queue, pass-through client, worker-side CAS before write with `path_changed`, post-write orphan check with `outcome_unknown`, plus the `dispatched` flag distinguishing never-started from ran-but-lost) is genuinely implemented — I verified the T0 capture happens before `withFileWriteLock` (filesystem-executor.ts:257-261), the client forwards identity verbatim (client.ts:196), and the worker CAS precedes the write (operations.ts:69-84). The headline scenario (swap while queued) is really closed, and the ~500 production lines are not over-built (the classifier/`dispatched`/contract module all carry real weight; ~60 lines of the tests are deletable). CI 12/12 green. Honest-residuals matches the implementation.

**Conclusion: PASS with two P2s that need an explicit decision or deferral note.**

**P2-1 — the T0-CAS misfires on lock-ordered cooperative changes: the missing↔existing transition returns an *internal* error where the previous semantics were "last writer wins".** Identity is captured unconditionally at T0 (even when the path existed before queuing), forwarded verbatim, but `targetType` is re-derived at T1 (after the lock) — so a Delete+Write race on the same path (Delete gets the lock first, unlinks; Write's T1 sees `missing` while still carrying identity X) hits the protocol `superRefine` (`missing` target must not carry identity) → `invalid_request` ("Filesystem worker request was invalid."). Same for Edit/Write queued after a Create (T0 saw missing → no identity → T1 sees `file` → forced-identity check at operations.ts:405 fires). Same-step concurrent tool calls with cooperating changes are exactly the scenario the write lock exists for; "删掉重写" is a legitimate model intent. Decide: exempt lock-ordered in-flight identities (e.g. recapture identity at T1 as the baseline, or return `path_changed` instead of `invalid_request` for missing+identity), and add a concurrent Delete+Write test asserting either a clean apply or a meaningful `path_changed` — never `invalid_request`.

**P2-2 — mid-write failures (truncate-then-ENOSPC/EIO) are classified as a clean `rejected`, contradicting the PR's own outcome contract.** The worker writes with `'w'` (truncate first); a mid-write error maps to a plain `filesystem_error`, and `UNKNOWN_OUTCOME_REASONS` doesn't include it — so the model is told "nothing happened" while the file is actually truncated/half-written on disk. That's exactly the PR's own definition of unknown ("may have applied before losing the ability to confirm") and post-write checks can't catch it (the check never runs when writeFile throws). Map "write already started" failures to `outcome_unknown`.

P3 (optional): the "type-level" tests in `filesystem-authority-contract.test.ts:68-107` are near-tautologies (asserting a literal array equals itself) — ~60 deletable lines; the duplicated catch/classify blocks (filesystem-executor.ts:262-271 and 298-307) should be one helper; the internal-protocol error strings ("A non-missing filesystem target must carry an identity for CAS.", "Filesystem worker request was invalid.") leak to the model — should be user-meaningful; the header comment references "file-stable-write.ts (later commit)" which doesn't exist in this PR; Windows CAS may silently no-op if Node's stat returns constant dev/ino (0) on some filesystems — no false positives but unverifiable protection, and the platform smoke tests skip on Windows; `format_json` runs the post-write check even with no change (harmless extra stat); the E2E never combines "real lock queue + real worker + swap while queued" (the T0 red-line test uses a fake worker, the CAS test calls the worker directly); `worker_crashed` + `dispatched=true` is coarse — a crash before reading the request (bundle load failure, sandbox exec refusal) marks every mutation outcome_unknown — safe but noisy when the worker is persistently broken.

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagent on `opencode-go/deepseek-v4-flash`), which traced the T0/T1 identity flow, the protocol superRefine, and the error-classification chain from the PR head. P2-1 and P2-2 are consequence analyses of new code paths, not observed failures. Please weigh these findings with your own judgment.*

<details>

结论：PASS（2 个 P2 需显式决策或延后说明）。竞态真实（外部进程可在"锁键解析→实际写入"之间替换路径、Delete 可删掉替换物、worker 断连后结果未知），T0 锁前捕获身份→client 原样透传→worker 写入前 CAS（path_changed）→写后孤儿检查（outcome_unknown）+ dispatched 标志（区分"从未启动"与"跑过丢结果"）三诉求全部真实落地（T0 捕获确在 withFileWriteLock 之前、透传原样、CAS 在写前）。headline 场景（排队期被替换）真实关闭；~500 行生产代码不过量（分类器/dispatched/契约模块都有实重），~60 行测试可删。P2-1：T0-CAS 误伤锁序协作变更——身份在 T0 无条件捕获、targetType 却在 T1（拿锁后)重新派生，同路径 Delete+Write 竞态（Delete 先拿锁 unlink，Write T1 见 missing 但仍带身份 X）命中协议 superRefine（missing 目标不得带身份）→ invalid_request（"Filesystem worker request was invalid."），Create 后排队的 Edit/Write 同理；"删掉重写"是模型合法意图且正是写锁存在的场景。需决策：豁免锁内在途身份（如 T1 重捕获基线或 missing+identity 返回 path_changed 而非 invalid_request）+ 补并发 Delete+Write 测试断言要么干净应用要么有意义的 path_changed，绝不能 invalid_request。P2-2：写入中途失败（O_TRUNC 截断后 ENOSPC/EIO）被分类为干净 rejected——'w' 旗标先截断再写，中途错误映射为普通 filesystem_error 且不在 UNKNOWN_OUTCOME_REASONS 里，模型被告知"什么都没发生"而文件实际已截断/半写——正是 PR 自己定义的 unknown 类且 post-write 检查兜不住（writeFile 抛错后检查不执行）。把"已开始写"的失败映射为 outcome_unknown。P3（可选）：类型级测试近乎同义反复（数组等于自身，~60 行可删）；两处重复 catch/classify 块应抽公共 helper；内部协议错误串直接面向模型应改用户可懂；头注释引用本 PR 不存在的 file-stable-write.ts；Windows 上 stat 若返回恒定 dev/ino（部分文件系统为 0）CAS 静默变 no-op（保护不可验证，smoke 在 Windows skip）；format_json 无变化仍跑 post-write 检查；E2E 从未组合"真实锁队列+真实 worker+排队期替换"；worker_crashed+dispatched=true 粒度粗（读请求前崩溃把所有变更标 outcome_unknown，安全但持续损坏时噪音大）。
</details>

packages/runtime/src/filesystem-worker/operations.ts
```diff
@@ -143,6 +152,10 @@ export async function executeFilesystemOperation(
         previous = code === 'ENOENT' || code === 'ENOTDIR' ? 'new' : 'unknown';
       }
       await fs.writeFile(path, operation.content, 'utf8');
```
Astro-Han: **P1 — The CAS check does not protect the object this path-based write opens.** After identity A passes validation, another process can replace the pathname with B before this call; `writeFile` then corrupts B and the post-check only reports `outcome_unknown` afterward. When T0 was missing, identity is undefined and this default `'w'` write also truncates a file created in the gap. Use an fd-pinned read/modify/write for existing targets and `O_CREAT|O_EXCL` (`wx`) for approved-missing writes; add deterministic races between validation and the actual open.

packages/runtime/src/filesystem-executor.ts
```diff
@@ -208,29 +247,66 @@ export function createBoundaryFilesystemExecutor(
...
-      const key = await writeLockTarget(call, call.operation.path);
-      return await withFileWriteLock(key, () => run(call));
+      const { key, canonicalPath } = await writeLockTarget(call, call.operation.path);
+      // Capture the target identity at lock acquisition (T0), BEFORE waiting
+      // for the lock. This is what closes the queue window: if Bash replaces
+      // the path while this call is queued, the worker's CAS sees a mismatch
+      // against this T0 inode and rejects with path_changed. Content operations
+      // follow the final symlink (stat); apply_patch create/delete use 'entry'
+      // semantics but execute() only handles write/edit/format_json here.
+      const expectedIdentity = workerFor(call.executionBoundary)
```
Astro-Han: **P1 — Local/workspace execution remains outside the new identity authority.** Identity is captured only when a worker exists; bypass/local mutations still wait for the lexical lock and then resolve/write by path, so a queued replacement is silently modified. Move identity/descriptor capture and the pinned RMW primitive into the shared filesystem authority used by every backend, or explicitly narrow the claimed safety boundary and reject unsupported local mutations.

Astro-Han (commented):
Capturing T0 identity before lock wait and conservatively classifying post-dispatch failures are useful pieces, but the mutation itself still reopens the pathname. The implementation can therefore detect corruption after it happens without preventing it, and the local/workspace path bypasses the new authority entirely.

The first-principles/Occam solution is one shared fd/handle-pinned mutation primitive: capture and validate the approved object, perform read/transform/write through that same descriptor, use exclusive creation for an approved missing target, and make both worker and local backends consume that contract. Post-checks then describe host visibility rather than compensate for a second pathname lookup.

Review performed with three Codex reviewer agents and DeepSeek V4 Flash as advisory tools; I reproduced both P1 paths against the latest head and verified green CI.

在等待锁前捕获 T0 identity，以及对 post-dispatch failure 做保守分类，都是有效基础；但 mutation 本身仍重新按 pathname 打开。因此当前实现只能在破坏发生后检测，无法阻止 replacement 被写坏；local/workspace 路径还完全绕过了新 authority。

更符合第一性原理和奥卡姆剃刀的方案是单一、共享的 fd/handle-pinned mutation primitive：捕获并验证批准对象，通过同一 descriptor 完成 read/transform/write；批准目标缺失时使用 exclusive create；worker 与 local backend 都消费这一契约。此时 post-check 只负责描述 host 可见性，不再弥补第二次 pathname lookup。

本次审查使用了三位 Codex reviewer agents 与 DeepSeek V4 Flash 作为辅助工具；我已在最新 head 上复现两个 P1，并复核绿色 CI。

</details>

chinawch007:
Thanks for your code review feedback. I've updated the code based on your comments and manually resolved all of them, but the CI didn't re-trigger.

packages/runtime/src/file-stable-write.ts
```diff
@@ -0,0 +1,248 @@
...
+      const matches =
+        entry !== null &&
+        String(entry.dev) === input.approvedIdentity.dev &&
+        String(entry.ino) === input.approvedIdentity.ino;
+      if (!matches) {
+        // A replacement was installed after the check. Restore it — the path
+        // is free because its entry is sitting on the tombstone — and refuse.
+        let restored = true;
+        try {
+          await rename(tombstone, input.path);
```
Astro-Han: **[P1] Do not overwrite a newer replacement while restoring the tombstone**

After this operation captures replacement C in the tombstone, another process can create B at `input.path` before this restore. On POSIX, `rename(tombstone, input.path)` then atomically overwrites and deletes B—the exact class of replacement loss this primitive is meant to prevent. Please use a no-replace restoration mechanism; if the original path has reappeared, preserve the tombstone and report `outcome_unknown` instead. Add the A→C capture, concurrent B creation, restore regression.

tombstone 捕获 C 后，其他进程可能在原路径创建 B；这里的 POSIX rename 会直接覆盖并删除 B。应使用 no-replace 恢复；若路径已重新出现，就保留 tombstone 并报告 unknown，再补该竞态回归。

</details>

packages/runtime/src/file-stable-write.ts
```diff
@@ -0,0 +1,248 @@
...
+          restored
+            ? 'The approved filesystem target changed before the delete; the replacement was restored. Re-check the directory before deleting again.'
+            : `The approved filesystem target changed before the delete; the replacement is preserved at ${tombstone}.`,
+        );
+      }
+    }
+    // The tombstone is a private unpredictable name: nothing else contends
+    // with this unlink.
+    try {
+      await unlink(tombstone);
```
Astro-Han: **[P2] Reject directories before moving them into a delete tombstone**

The authority layer permits a directory entry here. Renaming that directory to the tombstone succeeds, but `unlink(tombstone)` fails with `EPERM`/`EISDIR`; the function then reports `outcome_unknown` while the directory has disappeared from its original path and remains under a hidden name. The previous direct `unlink(path)` failed without moving it. Please reject directories before the rename (or safely restore on this failure) and add a directory-delete regression.

目录会先成功被 rename 到隐藏 tombstone，随后 `unlink` 失败，导致原路径消失但目录残留在隐藏位置。应在移动前拒绝目录，或在失败时安全恢复，并补目录删除回归。

</details>

Astro-Han (commented):
Thanks for addressing the earlier filesystem-authority feedback. I re-reviewed exact head `58c3f137018e3ae2072925745d34d51cb68b81e1`: all seven historical threads are resolved/outdated, the T0 identity and fd-pinned mutation direction is substantially stronger, the AI provenance is complete, and there is no UI/UX screenshot requirement.

I found two new delete-path issues and added them inline below. The exact-head Windows sandbox also currently fails because its pre-existing-file smoke write does not pass the newly required identity; that test should `lstat` the file and supply `{ dev, ino }`. The remaining workflows must finish green after the fixes.

Non-blocking documentation cleanup: the body still says four commits and lists obsolete short SHAs, while the current branch has eight commits.

AI-assisted review disclosure: OpenAI Codex performed the exact-head race, deletion, review-thread, provenance, and CI analysis; I verified the reproductions, severity, smallest fixes, and live GitHub state before posting.

旧反馈均已落实，但删除路径还有两个新问题：恢复 tombstone 时可能覆盖并删除并发创建的新 replacement；对目录执行删除会先把目录移进隐藏 tombstone，再因 `unlink` 失败而让原路径消失。Windows sandbox 也因 smoke test 没传新要求的 identity 而失败。修复后需重跑全绿。

</details>

chinawch007:
Again, updated code and resolved comment. Please help reviewing, thanks.

packages/runtime/src/file-stable-write.ts
```diff
@@ -0,0 +1,302 @@
...
+ * RENAME_NOREPLACE, but `link()` is natively no-replace: it fails with EEXIST
+ * when the destination exists. On EEXIST — or when linking is unsupported for
+ * the entry type / filesystem — the tombstone is preserved and the failure is
+ * reported as `outcome_unknown` with the location, so nothing is ever lost.
+ *
+ * @internal Exported for the no-replace restore regression test.
+ */
+export async function restoreTombstoneNoReplace(tombstone: string, path: string): Promise<void> {
+  try {
+    await link(tombstone, path);
```
Astro-Han, line 352: [P2] Restore the captured entry itself, not whatever it points at. `link()` dereferences a symlink source on darwin, so when the captured entry is a symlink this creates a hardlink to the symlink's *target* at the original path rather than restoring the link. I checked the behaviour directly on darwin: `linkSync('alink', 'restored')` succeeds and `lstat('restored').isSymbolicLink()` is `false` with a different inode. The identity guard below then reports `outcome_unknown` and deliberately touches nothing, so the outcome is the worst of both — the original symlink stays hidden under the tombstone and the path is now occupied by a new regular-file entry aliasing the symlink's target, which any later read or write will silently edit. Note the guard can safely clean up: `link()` only succeeds when the destination did not exist, so an entry found at the path afterwards is necessarily the one this call created. Either use `linkat` semantics that do not follow, or detect the symlink case before restoring and unlink what this call created before reporting; add a regression that deletes a symlink whose identity changed in the window and asserts the path is not left holding a foreign entry.

packages/runtime/src/file-stable-write.ts
```diff
@@ -0,0 +1,302 @@
...
+ */
+export async function compareAndDeleteEntry(input: {
+  path: string;
+  approvedIdentity: FilesystemTargetIdentity | undefined;
+}): Promise<void> {
+  // Reject directories before touching anything (#2600 review): a rename into
+  // the tombstone succeeds for directories, but the tombstone unlink then
+  // fails and leaves the directory hidden under the tombstone name — strictly
+  // worse than the plain unlink, which failed without moving it.
+  const entryBefore = await lstat(input.path);
```
Astro-Han, line 225: [P2] Make the directory refusal survive the race this module is about. This `lstat` is a pre-check, not an enforcement point: if the entry becomes a directory between here and the `rename` below, the directory is captured into the tombstone anyway, the identity comparison fails, and the restore cannot run because `link()` on a directory fails with `EPERM` — I confirmed that on darwin. The result is exactly the outcome the check was added to prevent: the path is empty and the directory survives only under the hidden tombstone name, now reported as `outcome_unknown`. Since the capture is the atomic step, the enforcement belongs after it — check the captured tombstone's type and, for a directory, `rename` it straight back before any identity comparison — with a regression that forces a directory into the tombstone rather than only asserting the pre-check.

packages/runtime/src/file-stable-write.ts
```diff
@@ -0,0 +1,302 @@
...
+  if (entryBefore.isDirectory()) {
+    throw new Error('Refusing to delete a directory through the entry-delete path.');
```
Astro-Han: [P3] Throw the failure type this path's callers understand. A bare `Error` reaches `normalizeOperationError` in `filesystem-worker/operations.ts`, which matches `FilesystemOperationError`, `StableWriteFailure` and `ApplyPatchRejectedError`, then falls through to the generic `operationError('filesystem_error', 'Filesystem operation failed.')` — so the model is told the operation failed and never learns that a directory was refused or what to do instead. Raise it as a structured failure with its own code so the message survives classification.

Astro-Han (commented):
Thanks for the quick turnaround on the delete path. Re-reviewed exact head `3edf983d0f42bcdb4ed5efe3c31c824773417f15`; the new delta since my last review is one commit over three files, so this pass is scoped to it plus a re-check of the two findings I raised.

Both are addressed in substance: the tombstone restore no longer clobbers a concurrent occupant, and the directory case is refused before anything moves. The Windows smoke test now supplies the identity the write path requires. The remaining problem is that the new no-replace restore only actually restores one entry type. `link()` cannot recreate a directory at all, and on macOS it dereferences a symlink source, so for those two entry types the restore either cannot run or produces the wrong object — which matters here because `compareAndDeleteEntry` documents symlink support as a design property and the whole module exists to keep an adversarial swap from losing an entry. Both are inline, plus one diagnostic point.

I have not re-run the suite for this delta; the reproductions below are the `link()` platform behaviour, checked directly on darwin, and code reading at this head. No CI checks are currently reported for this head.

Review disclosure: this review was prepared with Claude Code, which read the delta, traced the delete call sites and their error classification, and ran the `link()` platform probe quoted inline. The human contributor reviewed the findings before posting.

chinawch007:
One more turn, thanks for your attention.

Astro-Han (approved):
Thanks — this is a careful fix and the delete path in particular is handled better than most attempts at this class of bug.

**What this solves / how:** between "we verified this is the approved file" and the actual `write`/`unlink`, the path can be re-pointed at a different inode, because pathname lookup re-resolves on every syscall. Writes now go through a pinned fd; deletes `rename` the directory entry into a private tombstone and then compare against the identity captured at T0. That's the right layer — a userspace write lock cannot close this window, and no amount of retry logic would either.

**Review notes:**
- Write paths consistently execute through the handle (`writeThroughHandle` does `truncate` + `write` on the same fd; `write`/`edit`/`format_json`/`apply_patch` update all route through `openStableTarget`). No branch verifies on an fd and then lands via `writeFile(path)` or `rename(tmp, path)`.
- The delete CAS is genuinely a CAS, not a narrowed window: identity is captured before lock acquisition, the pre-`rename` `lstat` only rejects directories, and the actual comparison happens on the *tombstone* after the atomic `rename`. A path swapped between T0 and the `rename` gets captured, fails comparison, and is restored with `path_changed` rather than deleted.
- `restoreTombstoneNoReplace` never falls back to an overwriting write: `link`/`symlink` failures (including `EXDEV`/`EPERM`, not just `EEXIST`) all report `outcome_unknown` and keep the tombstone, so a concurrently reoccupied path is preserved rather than clobbered.
- POSIX has no atomic compare-and-unlink, so the tombstone mechanism is necessary complexity rather than incidental.

**Minor, non-blocking:** on the symlink restore path, after `symlink()` succeeds the `unlink(tombstone)` isn't guarded by an identity re-check the way the regular-file path is — inherently so, since the rebuilt link is a new inode. Losing the captured original would require a swap in that exact beat by a writer outside maka's own lock. Not raising it as a finding, but it is the one beat the regression tests don't cover, if you want to pin it.

LGTM.

---
*AI-assisted review. Two independent passes were run and the conclusions were then adversarially checked against the source at this exact head before posting.*

PR #3007, opened 2026-08-14
perf(ui): make streaming display redaction differential-safe

### LINKED ISSUE #2990: security(ui): make streaming display redaction differential-safe
<details open>
<summary>English</summary>

## Problem

Streaming assistant and thinking text currently stays safe by re-running `redactSecrets` over the accumulated display text. PR #2975 briefly attempted to optimize this with bounded incremental state, but review found that the incremental path could diverge from whole-text redaction: some chunk boundaries exposed secret suffixes, while others redacted legitimate opaque text.

#2975 therefore keeps the existing whole-text security boundary. A future optimization needs a stronger contract than a collection of regex examples.

## Required behavior

- For every input prefix and every possible chunking, the visible result must be exactly equivalent to applying `redactSecrets` to that full prefix, subject only to the existing documented display truncation.
- No intermediate, aborted, errored, or truncated state may reveal bytes that whole-text redaction classifies as secret.
- Normal text must not be redacted solely because tokens happen to cross chunk boundaries.
- The implementation must remain bounded in retained state and demonstrate a material improvement on realistic long streams before replacing the current path.

## Validation

Add differential/property tests over arbitrary chunk boundaries, including:

- API keys and authorization values containing separators such as `/`, `.`, `=`, `_`, and `-`;
- JWT-like values and long opaque hexadecimal/alphanumeric strings;
- legitimate hashes adjacent to paths or punctuation;
- one-character chunks, long whitespace, truncation boundaries, stream completion, abort, and error;
- randomized inputs comparing every streamed prefix with the whole-text oracle.

Keep a benchmark for a long response delivered in small deltas, but treat security equivalence as the hard constraint.

Related: #2975

</details>

<details>
<summary>中文</summary>

## 问题

Assistant 与 thinking 的流式文本目前通过对累计显示文本重复执行 `redactSecrets` 来保证安全。#2975 曾尝试用有界增量状态优化这条路径，但 review 发现增量结果可能偏离全量脱敏语义：部分切块会暴露 secret 后缀，另一些切块会误伤正常的 opaque text。

因此 #2975 保留现有的全量文本安全边界。后续优化需要先定义严格契约，而不是继续补充正则样例。

## 必须满足的行为

- 对任意输入前缀和任意切块方式，可见结果都必须与对该完整前缀执行 `redactSecrets` 严格等价，唯一例外是已有且明确记录的显示截断。
- 中间态、abort、error 和截断状态都不能暴露被全量脱敏判定为 secret 的内容。
- 正常文本不能仅因为 token 跨越 chunk 边界而被误脱敏。
- 实现必须保持有界状态；只有在真实长流基准中证明有显著收益后，才替换当前路径。

## 验证

增加覆盖任意切块边界的 differential/property tests，包括：

- 包含 `/`、`.`、`=`、`_`、`-` 等分隔符的 API key 与 authorization value；
- JWT 类值以及较长的十六进制/字母数字 opaque string；
- 紧邻路径或标点的合法 hash；
- 单字符 chunk、长空白、截断边界、stream completion、abort 与 error；
- 随机输入，并在每个流式前缀上与全量 oracle 对比。

保留“小 delta 组成长回复”的性能基准，但安全等价性是硬约束。

关联：#2975

</details>

Mingqwqqaq:
## Summary

Fixes #2990

Makes assistant and thinking stream redaction differential-safe while avoiding whole-display rescans on each delta. The incremental state machine caches settled display text, retains only bounded mutable/recovery context, and preserves the existing assistant head-cap and thinking tail-cap behavior.

It also fails closed across truncation and stateless bootstrap boundaries: contextual secrets and reversible provider/opaque tokens cannot expose suffix bytes, and hidden contexts resume once a safe terminator is observed.

## Verification

- `npm test --workspace @maka/ui`: 167/167 passed
- `npm test --workspace @maka/core`: 544/544 passed
- focused Desktop assistant-stream suite: 8/8 passed
- Biome check on affected files: passed
- `git diff --check`: passed
- `npm run benchmark:redaction --workspace @maka/ui`: candidate/oracle outputs matched
  - newline-rich: 65.4x faster (1561.99 ms -> 23.88 ms)
  - no-newline paragraph: 61.2x faster (1593.28 ms -> 26.04 ms)
- GitHub CI on `08d67e8ac`: all 11 checks passed

## Security review focus

Please focus on contextual-opener recovery after display truncation, reversible-token invalidation, bounded private state, and the conservative stateless `<redacted>` continuation path.

## AI use

Select exactly one:

Tool(s) and scope: OpenAI Codex contributed implementation, regression tests, benchmark hardening, conflict resolution, review fixes, and verification. The final squash commit must retain `Generated-by: Codex`.

## Checklist

- [ ] No


packages/ui/src/streaming-display-redaction.ts
```diff
@@ -0,0 +1,324 @@
...
+  const appendedPrivateState = PRIVATE_STATE.get(appended.state);
+  const stableSuffix = appendedPrivateState === undefined
+    ? undefined
+    : redactStableStreamingSuffix(appendedPrivateState.pendingRaw);
+  return {
+    text,
+    redacted: appended.redacted,
+    state: stateFor(
+      text,
+      '',
```
Astro-Han: [P1] Keep unfinished opener state after truncation

Could we preserve the unfinished contextual opener here instead of always clearing `pendingRaw`? If an oversized delta ends at `Authorization: Bearer` or `x-api-key:` before the value begins, `redactStableStreamingSuffix` returns no terminator. The next delta is then processed independently and its secret value reaches the live projection in plaintext.

This is reachable when session reseeding emits accumulated assistant text as one large delta. Please preserve or safely reseed the opener context and add a differential test through `applyLiveTurnEvent` with an oversized first delta ending at the opener and a second delta containing the value.

packages/ui/src/assistant-stream.ts
```diff
@@ -125,10 +122,25 @@ export function applyAssistantDelta(
...
+      redacted: false,
+      truncated: false,
+      ...(options.redactionState === undefined
+        ? {}
+        : { redactionState: options.redactionState }),
+    };
   }
 
   const previousText = prev ?? '';
+  const redactionState = options.redactionState ?? appendStreamingDisplayRedaction(
```
Astro-Han: [P2] Check the terminal cap before rebuilding state

After the assistant reaches the total display cap, the result no longer carries redaction state. Every later delta therefore reconstructs the state by rescanning the full 256 KiB display here, only to hit the terminal-marker check below and discard the delta.

Please move that check ahead of state reconstruction. This keeps the established display behavior and restores the intended constant-cost path after the cap.

packages/core/src/display-redaction.ts
```diff
@@ -77,3 +91,103 @@ export function redactSecrets(input: string): string {
...
+
+    const value = suffixMatch[pattern.streamingValueGroup];
+    if (value === undefined) continue;
+    const valueOffset = suffixMatch[0].lastIndexOf(value);
+    if (valueOffset < 0) continue;
+    const valueStart = suffixMatch.index + valueOffset;
+    const compactedInput =
+      value.length <= STREAMING_VALUE_TAIL_CHARS
+        ? input
+        : input.slice(0, valueStart) + value.slice(-STREAMING_VALUE_TAIL_CHARS);
```
Astro-Han, line 135: [P2] Bound all retained raw state

This compacts the active value but keeps the entire prefix before `valueStart`. If that prefix contains a large secret that is already redacted for display, the visible text can stay small while `pendingRaw` retains and repeatedly scans a much larger raw prefix.

Please apply the recovery limit to the complete retained private state. Stable or already-redacted prefixes should be settled so that only the shortest suffix that can interact with the next delta remains. A combined-secret test should assert the actual retained-character bound.

Astro-Han (changes requested):
Thanks for putting this together. I think the incremental approach is worthwhile as an optimization of the behavior we have today, even if we later decide that display redaction belongs elsewhere or should be removed. That larger design question does not need to hold this PR, and I would avoid turning this change into a broader rewrite.

Could you address the three inline issues before merge?

The focused fix I have in mind is to preserve or safely reseed redaction context after truncation, move the terminal-cap check ahead of state reconstruction, and enforce the recovery limit across all retained private state. The differential tests should go through `applyLiveTurnEvent`, since the current lower-level tests miss the production truncation and reseed path.

One additional validation request: please report the benchmark for both newline-rich transcripts and a long paragraph without line breaks. The current benchmark covers the settlement algorithm's most favorable input shape, while the no-newline case is slower than the previous implementation.

The desktop replay flush looks correct, but it has a separate UX owner and can be reviewed and reverted independently. I would split that part rather than tie it to the redaction optimization.

Separately, I think it is worth reviewing whether secrets should be kept out of model context, runtime events, persistence, and IPC so that display redaction can eventually be removed or reduced to an explicit privacy feature. That is a follow-up architecture question, not a requirement for this PR.

I used Codex reviewers and Claude to help inspect this change, then independently checked the production call chain, reproductions, exact head, and CI state.

<details>
<summary>中文审查结论</summary>

感谢推进这个优化。即使后续可能重新评估显示层脱敏是否应该保留，增量处理对现有实现仍然有价值。那个架构问题不需要阻塞本 PR，本 PR 也不必扩张成一次完整重写。

合并前请先处理三条行内问题：

1. 截断后保留或安全地重新建立脱敏上下文，避免 unfinished opener 后面的 secret value 明文进入 live projection。
2. 在重建 redaction state 前检查 assistant 是否已经到达总量上限。
3. retained-state 上限需要覆盖完整的私有状态，而不只是当前正在处理的 value。

差分测试需要经过 `applyLiveTurnEvent` 的生产调用链。当前底层测试没有覆盖真实的截断和 reseed 路径。

性能验证也请同时报告换行密集的 transcript 和无换行长段落。当前 benchmark 只覆盖了 settlement 算法最有利的输入形状，而无换行场景比旧实现更慢。

Desktop replay flush 本身看起来正确，但它属于独立的 UX 行为，可以单独审查和回滚。我建议把它从本 PR 拆出。

另外，可以单独评估是否应从源头阻止 secret 进入 model context、runtime event、持久化和 IPC，从而最终删除显示层脱敏，或将其收敛为明确的 privacy 功能。这是后续架构议题，不是本次优化的合并前要求。

</details>

packages/ui/src/assistant-stream.ts
Mingqwqqaq: Fixed in 1594aca80. Truncation now preserves and safely reseeds retained unfinished Authorization, API-key, and query-string openers instead of clearing `pendingRaw`; if the context cannot be retained, the state fails closed. I added production-path `applyLiveTurnEvent` coverage for oversized reseed deltas ending in `Authorization: Bearer `, `x-api-key: `, and `?access_token=`, followed by the secret value, plus a display-stability assertion for an oversized active value.

packages/ui/src/streaming-display-redaction.ts
Mingqwqqaq: Fixed in 1594aca80. Truncation now preserves and safely reseeds retained unfinished Authorization, API-key, and query-string openers instead of clearing `pendingRaw`; if the context cannot be retained, the state fails closed. I added production-path `applyLiveTurnEvent` coverage for oversized reseed deltas ending in `Authorization: Bearer `, `x-api-key: `, and `?access_token=`, followed by the secret value, plus a display-stability assertion for an oversized active value.

packages/ui/src/assistant-stream.ts
Mingqwqqaq: Correction: the preceding reply describes the P1 opener fix and was attached to this thread by mistake. This P2 is fixed in 1594aca80: the terminal total-cap marker check now runs before redaction-state reconstruction, so every later delta returns on the constant-cost path without rescanning the capped 256 KiB display.

packages/core/src/display-redaction.ts
Mingqwqqaq: Fixed in 1594aca80. Stable/redacted prefixes are now moved into settled display state, while `pendingRaw` retains only a bounded 128-character contextual tail that can still interact with a higher-priority pattern. The combined-secret regression streams two 5,000-character bearer values and asserts the complete retained private suffix stays within `maxTotalChars + 1` while remaining oracle-equivalent.

Mingqwqqaq:
Addressed the requested review changes in `1594aca80` and removed the independent Desktop replay-flush change from the net diff in `fc4e7b089`.

- truncation now preserves/reseeds unfinished Authorization, API-key, and query openers, with production-path `applyLiveTurnEvent` regressions
- the terminal total-cap check now runs before redaction-state reconstruction
- stable/redacted prefixes are settled; retained mutable raw state is bounded and covered by a combined two-secret test

Post-merge verification on head `b5dfff9b1`:

- `@maka/core`: 544/544 passed
- `@maka/ui`: 164/164 passed
- affected Desktop assistant-stream suite: 8/8 passed
- affected-file Biome check and `git diff --check`: passed

Benchmark, 69,600 input characters and 8-character deltas, using the real `applyAssistantDelta` path; candidate output matched the whole-text oracle in both cases:

| Scenario | Whole-text | Incremental | Speedup |
| --- | ---: | ---: | ---: |
| newline-rich | 1607.31 ms | 24.39 ms | 65.91x |
| no-newline long paragraph | 1567.71 ms | 25.32 ms | 61.92x |

The full monorepo suite was not run locally; the affected suites above were run after merging current `upstream/main`.

packages/ui/src/streaming-display-redaction.ts
```diff
@@ -0,0 +1,348 @@
...
+  }
+
+  if (overflow !== undefined) {
+    const nextOverflow = {
+      sourceHead: (overflow.sourceHead + delta).slice(0, maxRecoveryChars),
+      sourceTail: (overflow.sourceTail + delta).slice(-maxRecoveryChars),
+      compactedToken: overflow.compactedToken,
+      continuationChars: overflow.continuationChars,
+    };
+    const representative = previousPendingRaw + delta;
```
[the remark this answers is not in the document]
Mingqwqqaq: Fixed in ef641f00d by removing the unused `representative` binding. The affected UI suite and Biome check pass.

packages/ui/src/streaming-display-redaction.ts
```diff
@@ -0,0 +1,348 @@
...
+
+  return stateFor(
+    settledText,
+    canPreserveContext ? contextualRaw : '',
+    privateState?.continuationTerminator
+      ?? (canPreserveContext ? undefined : stableSuffix?.terminator)
+      ?? (privateState?.overflow === undefined ? undefined : /[^A-Za-z0-9_-]/)
+      ?? (context === undefined || canPreserveContext ? undefined : NEVER_TERMINATES),
+    configFor(privateState),
+  );
```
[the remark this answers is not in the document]
Mingqwqqaq: Fixed in ef641f00d. `stateAfterTruncation` now preserves the reversible overflow state whenever the compacted token remains visible. Tail recovery reconstructs from the compacted token start plus the retained real source tail, so an invalidating `_` cannot lose the provider-token boundary or be misclassified as a standalone opaque token. A new thinking tail-truncation regression reproduces this exact `ghp_..._` case and is oracle-equivalent; it failed before the fix and now passes.

packages/ui/scripts/benchmark-streaming-redaction.mjs
```diff
@@ -0,0 +1,64 @@
...
+    name,
+    inputChars: input.length,
+    deltaSize,
+    deltas: Math.ceil(input.length / deltaSize),
+    candidatePath: 'applyAssistantDelta',
+    wholeTextMs: Number(baseline.milliseconds.toFixed(2)),
+    incrementalMs: Number(candidate.milliseconds.toFixed(2)),
+    speedup: Number((baseline.milliseconds / candidate.milliseconds).toFixed(2)),
+  };
+});
```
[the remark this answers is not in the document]
Mingqwqqaq: Fixed in ef641f00d. Both paths are now warmed equally and each measurement reports the median of three runs. The self-building benchmark still matches candidate and oracle outputs; the fair rerun measured 65.4x for newline-rich input and 61.2x for the no-newline paragraph.

packages/ui/src/streaming-display-redaction.ts
```diff
@@ -0,0 +1,348 @@
...
+    if (reversibleInvalidated(delta, overflow.continuationChars)) {
+      previousPendingRaw = recovery === 'head'
+        ? overflow.sourceHead + delta
+        : overflow.sourceHead + nextOverflow.sourceTail;
+      delta = '';
+      overflow = undefined;
+    } else {
+      overflow = nextOverflow;
+    }
+  }
```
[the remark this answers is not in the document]
Mingqwqqaq: Fixed in ef641f00d: the unused `representative` binding has been removed. Biome and the 165-test UI suite pass.

packages/ui/src/streaming-display-redaction.ts
```diff
@@ -0,0 +1,364 @@
...
+  return stateFor(
+    settledText,
+    retainedRaw,
+    privateState?.continuationTerminator
+      ?? (canPreserveContext ? undefined : stableSuffix?.terminator)
+      ?? (privateState?.overflow === undefined || canPreserveOverflow
+        ? undefined
+        : /[^A-Za-z0-9_-]/)
+      ?? (context === undefined || canPreserveContext ? undefined : NEVER_TERMINATES),
+    {
```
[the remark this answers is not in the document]
Mingqwqqaq: Fixed in 551a42bcb. Tail-capped owners now retain a bounded display-hidden contextual opener, so the next append reconstructs the redacted oracle suffix and resumes after the terminator. Contexts too large to retain fail closed only until a newline instead of freezing forever. The new small-cap regression failed before the change and now matches applyThinkingComplete; the overlong fallback also stays bounded, hides the secret, and resumes on the next line. The full UI suite passes (167/167 on the latest head).

Mingqwqqaq:
Follow-up review findings are addressed in 551a42bcb and 08d67e8ac. Hidden contextual openers under thinking tail caps now recover from bounded private state and overlong contexts resume safely at a newline instead of freezing. Stateless callers whose prior display ends in <redacted> now fail closed until a clear terminator, and the API comment explicitly requires threading state for exact differential equivalence. Both regressions failed before their fixes. Latest local verification: UI 167/167, Biome clean, and git diff --check clean.

Astro-Han (approved):
Thanks for addressing the review findings. I re-reviewed the exact head `08d67e8ac` and found no remaining P0–P3 issues.

The previous findings are closed at their natural owner: truncation preserves or safely reseeds unfinished contextual openers, the terminal-cap path returns before state reconstruction, retained private recovery state is bounded, reversible tokens and hidden thinking contexts recover safely after truncation, and stateless redacted tails fail closed. The independent Desktop replay change is also no longer part of the net diff.

The focused regressions are mutation-valid, the production call chain is covered, and the current CI is green. Approving.

This approval applies to improving the display-redaction contract that exists today. Whether that contract should eventually be removed or reduced remains a separate architecture discussion in #2990.

<details>
<summary>中文对照</summary>

感谢处理审查反馈。我重新审查了准确的 head `08d67e8ac`，目前没有遗留的 P0–P3 问题。

之前的问题均已在自然 owner 上闭合：截断后会保留或安全重建未完成的上下文 opener；达到总量上限后会在重建状态前直接返回；私有恢复状态有明确上限；可逆 token 和被 thinking 截断隐藏的上下文能够安全恢复；无状态调用方遇到无法还原的脱敏尾部时会 fail closed。独立的 Desktop replay 改动也已经不在最终 diff 中。

focused regression 已通过 mutation 验证，生产调用链已有覆盖，当前 CI 全绿。因此 Approve。

这次批准针对现有显示脱敏契约的优化。长期是否应该删除或收缩该契约，仍然是 #2990 中独立的架构问题。

</details>

AI-assisted review disclosure: Codex and two review agents assisted with source inspection, mutation checks, and focused validation. I reviewed the exact head, evidence, production call chain, and CI state, and I own this approval and merge decision.

PR #3008, opened 2026-08-14
fix(eval): classify CONNECT hosts and kill raw TCP tunnels

### LINKED ISSUE #2977: Raw CONNECT tunnels bypass the Eval contamination filter and the egress audit log
`packages/eval/harbor/egress_filter.py` implements only the `request` hook and classifies `flow.request.pretty_url`. mitmproxy does not fire `request` for a `CONNECT`, and when the bytes inside an established tunnel are neither TLS nor HTTP, `next_layer` falls back to a raw TCP layer, so no hook runs at all. The flow reaches no contamination rule, and because `hits.jsonl` records only blocks and policy errors, it leaves no audit record either.

The namespace contract still holds — the traffic does go through mitmproxy. What does not hold is "audited", and with it the contamination blocklist the cohort leans on.

## Reproduction

Against `maka-eval-egress-proxy:12.2.3` built from `packages/eval/harbor/egress-proxy/Dockerfile`, from a subject under the proxy-only policy:

```
HTTP  http://tbench.ai/x   -> 451 tbench_domain          filter works for plain HTTP
HTTP  http://example.com/  -> 200
CONNECT ssh.github.com:443 -> HTTP/1.1 200 Connection established
  first bytes from peer:      SSH-2.0-...  KEXINIT
CONNECT tbench.ai:443      -> HTTP/1.1 200 Connection established
```

GitHub's real SSH banner and key exchange came back, so `git clone git@github.com:...` through `ssh.github.com:443` traverses the proxy without ever being classified. Note the second line too: `tbench.ai` is on the blocklist, and the `CONNECT` to it is answered `200`. For a TLS client the inner request is still blocked after interception, so the exposure is raw tunnels.

Deliberate to exploit, not something a subject hits by accident.

## Candidate directions

- Apply the host rules at `http_connect`, so a blocklisted host is refused before the tunnel opens rather than after interception. Closes the blocklisted-host case, not the arbitrary-host one.
- Hook `tcp_start` and kill flows that fall back to raw TCP, with an audit record. Fail-closed, and nothing in a coding benchmark legitimately needs a non-HTTP tunnel through this proxy — SSH egress is already impossible except through it.
- Record every tunnel, not only blocks. Changes what `hits.jsonl` means and interacts with its 1 MiB cap.

## Scope

Found while adjudicating the reviews of #2960, which states the gap in the README rather than closing it. This is the contamination blocklist's completeness, a separate concern from the namespace isolation that PR is about, and it needs mitmproxy hook work with its own tests.

1625567290:
## Summary

`egress_filter.py` only implemented `request`, so a `CONNECT` never hit the contamination rules. A tunnel whose first bytes were neither TLS nor HTTP became raw TCP with no audit record. `CONNECT tbench.ai:443` returned 200; `CONNECT ssh.github.com:443` forwarded a real SSH banner.

- `http_connect` runs the existing host rules and returns **451** for blocklisted CONNECT targets before the tunnel opens.
- `next_layer` replaces a raw `TCPLayer`, `tcp_message` clears already-buffered payloads, and `tcp_start` records `raw_tunnel`. `flow.kill()` alone does not stop mitmproxy 12.2.3's TCP relay.
- Unrelated HTTPS through the proxy is unchanged.

Fixes #2977

## Verification

- `python3.13 packages/eval/harbor/test_egress_filter.py` — 7 pass
- Live `maka-eval-egress-proxy:12.2.3` built from this branch:
  - `CONNECT tbench.ai:443` → **451** `tbench_domain`
  - `https://example.com/` via the proxy → **200**
  - `CONNECT ssh.github.com:443` → 200 then no `SSH-2.0` banner; `hits.jsonl` has `raw_tunnel` / `ssh.github.com`

Not run: full `maka eval run` cohort.

## Checklist

- [x] Yes — blocklisted CONNECT hosts are refused, and raw TCP tunnels are killed and audited

### DIFF
--- a/packages/eval/harbor/egress_filter.py
+++ b/packages/eval/harbor/egress_filter.py
@@ -101,12 +101,50 @@ try:
 except ImportError:
     http = None
 
+try:
+    from mitmproxy.proxy import commands as proxy_commands
+except ImportError:
+    proxy_commands = None
+
 
 def request(flow: object) -> None:
+    apply_http_policy(flow, flow.request.pretty_url)
+
+
+def http_connect(flow: object) -> None:
+    try:
+        raw_url = connect_target_url(flow)
+    except Exception:
+        raw_url = ""
+    apply_http_policy(flow, raw_url)
+
+
+def tcp_start(flow: object) -> None:
+    record_raw_tunnel(flow)
+    kill_flow(flow)
+
+
+def tcp_message(flow: object) -> None:
+    messages = getattr(flow, "messages", None)
+    if messages:
+        messages[-1].content = b""
+    kill_flow(flow)
+
+
+def next_layer(nextlayer: object) -> None:
+    current = getattr(nextlayer, "layer", None)
+    if current is None or type(current).__name__ != "TCPLayer":
+        return
+    context = getattr(nextlayer, "context", None)
+    record_raw_tunnel(context)
+    nextlayer.layer = CloseRawLayer(context)
+
+
+def apply_http_policy(flow: object, raw_url: str) -> None:
     if http is None:
         raise RuntimeError("mitmproxy is required to run the Eval egress filter")
     try:
-        matched = contamination_rule(flow.request.pretty_url)
+        matched = contamination_rule(raw_url)
         if not matched:
             return
         rule_id, host, normalized_path = matched
@@ -125,6 +163,62 @@ def request(flow: object) -> None:
             append_audit("policy_error", "", type(error).__name__)
         except Exception:
             pass
+
+
+def connect_target_url(flow: object) -> str:
+    request = flow.request
+    host = (getattr(request, "pretty_host", None) or getattr(request, "host", "") or "").strip()
+    if not host:
+        raise ValueError("empty CONNECT host")
+    if ":" in host and not host.startswith("["):
+        host = f"[{host}]"
+    port = getattr(request, "port", None)
+    if port in (None, 443):
+        return f"https://{host}/"
+    if port == 80:
+        return f"http://{host}/"
+    return f"https://{host}:{port}/"
+
+
+def tcp_peer(flow: object) -> tuple[str, str]:
+    server = getattr(flow, "server_conn", None) or getattr(flow, "server", None)
+    address = getattr(server, "address", None) if server is not None else None
+    if isinstance(address, (tuple, list)) and address:
+        host = str(address[0])[:255]
+        port = address[1] if len(address) > 1 else ""
+        return host, f":{port}" if port != "" else ""
+    return "", ""
+
+
+def record_raw_tunnel(flow: object) -> None:
+    host, path = tcp_peer(flow)
+    try:
+        append_audit("raw_tunnel", host, path)
+    except Exception:
+        pass
+
+
+def kill_flow(flow: object) -> None:
+    kill = getattr(flow, "kill", None)
+    if callable(kill) and getattr(flow, "killable", True):
+        try:
+            kill()
+        except Exception:
+            pass
+
+
+class CloseRawLayer:
+    def __init__(self, context: object) -> None:
+        self.context = context
+
+    def handle_event(self, event: object):
+        if proxy_commands is None:
+            return
+            yield
+        for name in ("client", "server"):
+            connection = getattr(self.context, name, None)
+            if connection is not None:
+                yield proxy_commands.CloseConnection(connection)
 
 
 def blocked_response(rule_id: str):
--- a/packages/eval/harbor/test_egress_filter.py
+++ b/packages/eval/harbor/test_egress_filter.py
@@ -78,6 +78,89 @@ class EgressFilterTest(unittest.TestCase):
             self.assertIn("host", record)
             self.assertIn("normalizedPath", record)
 
+    def test_http_connect_refuses_blocklisted_hosts_before_the_tunnel_opens(self) -> None:
+        class Response:
+            @staticmethod
+            def make(status, body, headers):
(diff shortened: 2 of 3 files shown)


liugddx (commented):
Verdict: merge recommended (non-blocking). The core direction is sound: classify CONNECT before tunnel establishment, then identify, audit, and close raw fallback before relay. Setting rawtcp=false alone would be shorter, but it loses precise raw_tunnel auditing and is therefore not equivalent.

Please consider the inline hardening items before merge or immediately afterward: make the closing-layer contract explicit and add a test that actually verifies CloseConnection commands. The tcp_start/tcp_message + flow.kill path is removable fallback complexity; its mocks prove method invocation, not socket closure. No correctness blocker found.

<details>
<summary>中文审查结论</summary>

结论：建议合并（非阻塞）。核心方向正确：CONNECT 在建立隧道前复用污染规则；raw fallback 在 relay 前被识别、审计并关闭。单独设置 rawtcp=false 虽然更短，但会失去精确的 raw_tunnel 审计，因此不是完整等价方案。

建议在合并前或紧随其后完成行内加固：让关闭层契约更明确，并补上真正验证 CloseConnection 的测试。tcp_start/tcp_message + flow.kill 属于可以删除的冗余兜底；它们的 mock 测试只证明方法被调用，不能证明连接已经关闭。未发现正确性阻塞问题。

</details>

packages/eval/harbor/test_egress_filter.py
```diff
@@ -78,6 +78,89 @@ def make(status, body, headers):
...
+    def test_next_layer_replaces_raw_tcp_with_a_closer(self) -> None:
+        class TCPLayer:
+            pass
+
+        with tempfile.TemporaryDirectory() as directory:
+            MODULE.AUDIT_PATH = Path(directory) / "hits.jsonl"
+            context = SimpleNamespace(server=SimpleNamespace(address=("ssh.github.com", 443)))
+            nextlayer = SimpleNamespace(layer=TCPLayer(), context=context)
+            MODULE.next_layer(nextlayer)
+            self.assertIsInstance(nextlayer.layer, MODULE.CloseRawLayer)
```
liugddx, line 159: This test does not cover the most important behavior of the fix. All seven tests still pass if the CloseConnection logic in CloseRawLayer.handle_event() is removed. Please consume the handle_event() generator and assert that it emits exactly two CloseConnection commands targeting client and server. Ideally, add a pinned mitmproxy==12.2.3 CONNECT-to-SSH integration test that verifies no banner is relayed and raw_tunnel is audited.

<details>
<summary>中文说明</summary>

这个测试没有覆盖本次修复最关键的行为。现在即使把 CloseRawLayer.handle_event() 中两个 CloseConnection 全部删除，7 个测试仍会通过。建议构造 client/server，消费 handle_event() 生成器，并断言准确产生两个 CloseConnection 命令；最好再增加固定 mitmproxy==12.2.3 的 CONNECT→SSH 集成测试，验证无 banner 返回且写入 raw_tunnel。

</details>

packages/eval/harbor/egress_filter.py
```diff
@@ -101,12 +101,50 @@ def public_trajectory_repository(host: str, path_query: str) -> bool:
...
+def tcp_message(flow: object) -> None:
+    messages = getattr(flow, "messages", None)
+    if messages:
+        messages[-1].content = b""
+    kill_flow(flow)
+
+
+def next_layer(nextlayer: object) -> None:
+    current = getattr(nextlayer, "layer", None)
+    if current is None or type(current).__name__ != "TCPLayer":
```
liugddx, line 136: Because the runtime image is pinned to mitmproxy 12.2.3, prefer importing TCPLayer and using isinstance over comparing a class-name string, and make CloseRawLayer explicitly inherit mitmproxy Layer. This documents the protocol contract and avoids silent misclassification from same-named stand-ins or local refactors.

<details>
<summary>中文说明</summary>

既然运行镜像固定为 mitmproxy 12.2.3，建议显式导入 TCPLayer 并使用 isinstance，而不是比较类名字符串；同时让 CloseRawLayer 正式继承 mitmproxy Layer。这样能把依赖的内部协议契约写清楚，避免同名假对象或未来局部重构造成静默误判。

</details>

packages/eval/harbor/egress_filter.py
```diff
@@ -101,12 +101,50 @@ def public_trajectory_repository(host: str, path_query: str) -> bool:
...
+
+def http_connect(flow: object) -> None:
+    try:
+        raw_url = connect_target_url(flow)
+    except Exception:
+        raw_url = ""
+    apply_http_policy(flow, raw_url)
+
+
+def tcp_start(flow: object) -> None:
```
liugddx, line 122: These flow.kill paths appear removable. On the normal path, next_layer replaces TCPLayer before it receives Start/DataReceived, so tcp_start/tcp_message do not run; flow.kill also does not reliably establish that an in-flight relay is closed. Keeping the single pre-relay next_layer→CloseRawLayer path is simpler, and the two corresponding mock tests can be removed.

<details>
<summary>中文说明</summary>

这两条 flow.kill 路径看起来可以删除。正常路径里 next_layer 已在 TCPLayer 接收 Start/DataReceived 前将其替换，因此 tcp_start/tcp_message 不会执行；而 flow.kill 也不能可靠证明正在转发的连接已关闭。保留 next_layer→CloseRawLayer 这一条 relay 前阻断路径更符合奥卡姆剃刀，对应两个 mock 测试也可删除。

</details>

1625567290:
Follow-up for the inline hardening items is https://github.com/maka-agent/maka-agent/pull/3017.

PR #3017, opened 2026-08-14
fix(eval): close raw tunnels at the layer that owns the sockets

1625567290:
## Summary

Raw CONNECT and HTTP 101 non-WebSocket upgrades must not relay bytes, and the proxy must close both transports — not only mark the flow killed.

- Script `next_layer` runs **before** the built-in classifier assigns `TCPLayer`. When client data looks like raw TCP (not TLS/HTTP), assign `CloseRawLayer` so the built-in NextLayer leaves it alone and both sides get `CloseConnection`. Incomplete HTTP prefixes (`GET`, `GET `) stay unclassified until the request line can be ruled in or out.
- Force `rawtcp=false` (entrypoint + `configure`) so HTTP 101 non-WebSocket upgrades never start a `TCPLayer`; mitmproxy closes the client after 101. WebSocket upgrades stay on the websocket layer.
- `tcp_start` / `tcp_message` remain last-resort for any residual `TCPLayer` (e.g. `tcp_hosts`); they are not the primary closing authority.
- Live harness: unique Docker names per run, dynamic host port, long-lived origins, timeout is a failure, assert EOF/reset with zero payload and `raw_tunnel` audit.

## Verification

- `python3.13 harbor/test_egress_filter.py` — 18/18
- `python3.13 harbor/test_eval_framework.py` — 3/3
- Live suite remains opt-in (`MAKA_EVAL_EGRESS_PROXY_TEST=1`) against `maka-eval-egress-proxy:12.2.3`

## Checklist

- [x] Yes — raw tunnels are closed by `CloseConnection` at the owning layer (CONNECT via `CloseRawLayer`, 101 via `rawtcp=false`), not by `TCPFlow.kill()` alone

### DIFF
--- a/packages/eval/harbor/egress_filter.py
+++ b/packages/eval/harbor/egress_filter.py
@@ -121,10 +121,13 @@ except ImportError:
 try:
     from mitmproxy.proxy import commands as proxy_commands
     from mitmproxy.proxy.layer import Layer
+    from mitmproxy.proxy.layers import ClientTLSLayer, ServerTLSLayer
     from mitmproxy.proxy.layers.tcp import TCPLayer
 except ImportError:
     proxy_commands = None
     Layer = object
+    ClientTLSLayer = None
+    ServerTLSLayer = None
     TCPLayer = None
 
 try:
@@ -193,6 +196,14 @@ def next_layer(nextlayer: object) -> None:
     current = getattr(nextlayer, "layer", None)
     context = getattr(nextlayer, "context", None)
     if current is None:
+        data_client = _next_layer_bytes(nextlayer, "data_client")
+        if _is_fragmented_tls_record_prefix(data_client):
+            if ClientTLSLayer is None or ServerTLSLayer is None:
+                return
+            server_tls = ServerTLSLayer(context)
+            server_tls.child_layer = ClientTLSLayer(context)
+            nextlayer.layer = server_tls
+            return
         if not looks_like_raw_tcp(nextlayer):
             return
         record_raw_tunnel(context)
@@ -269,6 +280,10 @@ def _could_start_tls_record(data: bytes) -> bool:
     """Keep a fragmented ClientHello undecided until its 3-byte prefix exists."""
     if starts_like_tls_record(data):
         return True
+    return _is_fragmented_tls_record_prefix(data)
+
+
+def _is_fragmented_tls_record_prefix(data: bytes) -> bool:
     return 0 < len(data) < 3 and b"\x16\x03".startswith(data)
 
 
--- a/packages/eval/harbor/egress_filter.py
+++ b/packages/eval/harbor/egress_filter.py
@@ -120,8 +120,12 @@ except ImportError:
 
 try:
     from mitmproxy.proxy import commands as proxy_commands
+    from mitmproxy.proxy.layer import Layer
+    from mitmproxy.proxy.layers.tcp import TCPLayer
 except ImportError:
     proxy_commands = None
+    Layer = object
+    TCPLayer = None
 
 
 def request(flow: object) -> None:
@@ -136,21 +140,9 @@ def http_connect(flow: object) -> None:
     apply_http_policy(flow, raw_url)
 
 
-def tcp_start(flow: object) -> None:
-    record_raw_tunnel(flow)
-    kill_flow(flow)
-
-
-def tcp_message(flow: object) -> None:
-    messages = getattr(flow, "messages", None)
-    if messages:
-        messages[-1].content = b""
-    kill_flow(flow)
-
-
 def next_layer(nextlayer: object) -> None:
     current = getattr(nextlayer, "layer", None)
-    if current is None or type(current).__name__ != "TCPLayer":
+    if TCPLayer is None or not isinstance(current, TCPLayer):
         return
     context = getattr(nextlayer, "context", None)
     record_raw_tunnel(context)
@@ -215,18 +207,16 @@ def record_raw_tunnel(flow: object) -> None:
         pass
 
 
-def kill_flow(flow: object) -> None:
-    kill = getattr(flow, "kill", None)
-    if callable(kill) and getattr(flow, "killable", True):
-        try:
-            kill()
-        except Exception:
-            pass
-
-
-class CloseRawLayer:
+class CloseRawLayer(Layer):
     def __init__(self, context: object) -> None:
-        self.context = context
+        if Layer is object:
+            self.context = context
+            return
+        if getattr(context, "layers", None) is None:
+            context.layers = []
+        if getattr(context, "options", None) is None:
+            context.options = type("Options", (), {"proxy_debug": False})()
+        super().__init__(context)
 
     def handle_event(self, event: object):
         if proxy_commands is None:
--- a/packages/eval/harbor/egress_filter.py
+++ b/packages/eval/harbor/egress_filter.py
@@ -140,13 +140,30 @@ def http_connect(flow: object) -> None:
     apply_http_policy(flow, raw_url)
 
 
+def tcp_start(flow: object) -> None:
+    # Script next_layer runs before the built-in classifier assigns
+    # nextlayer.layer, so CONNECT raw tunnels and HTTP 101 upgrades are
+    # closed here — not by replacing a TCPLayer that does not exist yet.
+    record_raw_tunnel(flow)
+    kill_flow(flow)
+
+
+def tcp_message(flow: object) -> None:
+    messages = getattr(flow, "messages", None)
+    if messages:
+        messages[-1].content = b""
+    kill_flow(flow)
+
+
 def next_layer(nextlayer: object) -> None:
     current = getattr(nextlayer, "layer", None)
     if TCPLayer is None or not isinstance(current, TCPLayer):
         return
     context = getattr(nextlayer, "context", None)
     record_raw_tunnel(context)
-    nextlayer.layer = CloseRawLayer(context)
+    closer = CloseRawLayer(context)
+    replace_layer(context, current, closer)
+    nextlayer.layer = closer
 
 
 def apply_http_policy(flow: object, raw_url: str) -> None:
@@ -205,6 +222,30 @@ def record_raw_tunnel(flow: object) -> None:
         append_audit("raw_tunnel", host, path)
     except Exception:
         pass
+
+
+def kill_flow(flow: object) -> None:
+    kill = getattr(flow, "kill", None)
+    if callable(kill) and getattr(flow, "killable", True):
+        try:
+            kill()
+        except Exception:
+            pass
+
+
+def replace_layer(context: object, current: object, closer: object) -> None:
+    layers = getattr(context, "layers", None)
+    if not isinstance(layers, list):
+        return
+    try:
+        index = layers.index(current)
+    except ValueError:
+        index = len(layers)
+    if closer in layers:
+        layers.remove(closer)
+    if current in layers:
+        layers.remove(current)
+    layers.insert(min(index, len(layers)), closer)
 
 
 class CloseRawLayer(Layer):
--- a/packages/eval/harbor/egress_filter.py
+++ b/packages/eval/harbor/egress_filter.py
@@ -154,14 +154,11 @@ def response(flow: object) -> None:
     response = getattr(flow, "response", None)
     if getattr(response, "status_code", None) != 101:
         return
-    headers = getattr(response, "headers", None)
-    upgrade = ""
-    if headers is not None:
-        try:
-            upgrade = str(headers.get("upgrade", "") or "")
-        except Exception:
-            upgrade = ""
-    if upgrade.lower() == "websocket":
+    # mitmproxy 12.2.3 sets flow.websocket before HttpResponseHook only when
+    # the 101 is a real WebSocket upgrade (Upgrade + version 13 + option on).
+    # A websocket Upgrade header alone still falls through to CloseConnection
+    # under rawtcp=false and must be audited.
+    if getattr(flow, "websocket", None) is not None:
         return
     record_raw_tunnel(flow)
 
--- a/packages/eval/harbor/egress_filter.py
+++ b/packages/eval/harbor/egress_filter.py
@@ -256,13 +256,20 @@ def looks_like_raw_tcp(nextlayer: object) -> bool:
     """
     data_client = _next_layer_bytes(nextlayer, "data_client")
     data_server = _next_layer_bytes(nextlayer, "data_server")
-    if starts_like_tls_record(data_client):
+    if _could_start_tls_record(data_client):
         return False
     if not data_client and not data_server:
         return False
     if data_server or data_client.startswith(b"SSH"):
         return True
     return not _still_could_be_http(data_client)
+
+
+def _could_start_tls_record(data: bytes) -> bool:
+    """Keep a fragmented ClientHello undecided until its 3-byte prefix exists."""
+    if starts_like_tls_record(data):
+        return True
+    return 0 < len(data) < 3 and b"\x16\x03".startswith(data)
 
 
 def _still_could_be_http(data: bytes) -> bool:
--- a/packages/eval/harbor/egress_filter.py
+++ b/packages/eval/harbor/egress_filter.py
@@ -127,9 +127,43 @@ except ImportError:
     Layer = object
     TCPLayer = None
 
+try:
+    from mitmproxy.net.tls import starts_like_tls_record
+except ImportError:
+    def starts_like_tls_record(data: bytes) -> bool:
+        return len(data) >= 3 and data[0] == 0x16 and data[1] == 0x03
+
+
+def configure(updated: object) -> None:
+    # HTTP 101 upgrades construct TCPLayer inside the HTTP layer without
+    # another next_layer hook. rawtcp=false makes that path CloseConnection
+    # itself; WebSocket upgrades stay on the websocket layer.
+    try:
+        from mitmproxy import ctx
+    except ImportError:
+        return
+    if getattr(ctx.options, "rawtcp", False):
+        ctx.options.rawtcp = False
+
 
 def request(flow: object) -> None:
     apply_http_policy(flow, flow.request.pretty_url)
+
+
+def response(flow: object) -> None:
+    response = getattr(flow, "response", None)
+    if getattr(response, "status_code", None) != 101:
+        return
+    headers = getattr(response, "headers", None)
+    upgrade = ""
+    if headers is not None:
+        try:
+            upgrade = str(headers.get("upgrade", "") or "")
+        except Exception:
+            upgrade = ""
+    if upgrade.lower() == "websocket":
+        return
+    record_raw_tunnel(flow)
 
 
 def http_connect(flow: object) -> None:
@@ -141,9 +175,9 @@ def http_connect(flow: object) -> None:
 
 
 def tcp_start(flow: object) -> None:
-    # Script next_layer runs before the built-in classifier assigns
-    # nextlayer.layer, so CONNECT raw tunnels and HTTP 101 upgrades are
-    # closed here — not by replacing a TCPLayer that does not exist yet.
+    # Last resort if a TCPLayer is still admitted (tcp_hosts / ignore).
+    # CONNECT raw is closed by next_layer → CloseRawLayer; HTTP 101 raw is
+    # closed by rawtcp=false. Neither of those paths starts a TCPLayer.
     record_raw_tunnel(flow)
     kill_flow(flow)
 
@@ -156,10 +190,19 @@ def tcp_message(flow: object) -> None:
 
 
 def next_layer(nextlayer: object) -> None:
+    # Script addons run before the built-in classifier assigns layer. If we
+    # set CloseRawLayer here, NextLayer leaves it in place. Waiting for
+    # isinstance(..., TCPLayer) never fires on the production CONNECT path.
     current = getattr(nextlayer, "layer", None)
+    context = getattr(nextlayer, "context", None)
+    if current is None:
+        if not looks_like_raw_tcp(nextlayer):
+            return
+        record_raw_tunnel(context)
+        nextlayer.layer = CloseRawLayer(context)
+        return
     if TCPLayer is None or not isinstance(current, TCPLayer):
         return
-    context = getattr(nextlayer, "context", None)
     record_raw_tunnel(context)
     closer = CloseRawLayer(context)
     replace_layer(context, current, closer)
@@ -204,6 +247,36 @@ def connect_target_url(flow: object) -> str:
     if port == 80:
         return f"http://{host}/"
     return f"https://{host}:{port}/"
+
+
+def looks_like_raw_tcp(nextlayer: object) -> bool:
+    """Match mitmproxy 12.2.3's raw-TCP classifier without waiting for TCPLayer."""
+    data_client = _next_layer_bytes(nextlayer, "data_client")
+    data_server = _next_layer_bytes(nextlayer, "data_server")
+    if starts_like_tls_record(data_client):
+        return False
+    if not data_client and not data_server:
+        return False
+    probably_no_http = (
+        len(data_client) < 3
+        or b" " not in data_client
+        or (data_client.find(b" ") > data_client.find(b"\n"))
+        or not data_client[:3].isalpha()
+        or bool(data_server)
+        or data_client.startswith(b"SSH")
+    )
+    return probably_no_http
+
+
+def _next_layer_bytes(nextlayer: object, name: str) -> bytes:
+    getter = getattr(nextlayer, name, None)
+    if not callable(getter):
+        return b""
+    try:
+        data = getter()
+    except Exception:
+        return b""
+    return bytes(data) if isinstance(data, (bytes, bytearray)) else b""
 
 
 def tcp_peer(flow: object) -> tuple[str, str]:
--- a/packages/eval/harbor/egress_filter.py
+++ b/packages/eval/harbor/egress_filter.py
@@ -250,22 +250,33 @@ def connect_target_url(flow: object) -> str:
 
 
 def looks_like_raw_tcp(nextlayer: object) -> bool:
-    """Match mitmproxy 12.2.3's raw-TCP classifier without waiting for TCPLayer."""
+    """Close only when the bytes cannot still become HTTP or TLS.
+
+    Script next_layer runs before mitmproxy 12.2.3's classifier. Copying its
+    `probably_no_http` test here would treat `GET` / `GET ` as raw and assign
+    CloseRawLayer while the request line is still arriving. With rawtcp=false
+    the built-in path would have kept waiting for HttpLayer.
+    """
     data_client = _next_layer_bytes(nextlayer, "data_client")
     data_server = _next_layer_bytes(nextlayer, "data_server")
     if starts_like_tls_record(data_client):
         return False
     if not data_client and not data_server:
         return False
-    probably_no_http = (
-        len(data_client) < 3
-        or b" " not in data_client
-        or (data_client.find(b" ") > data_client.find(b"\n"))
-        or not data_client[:3].isalpha()
-        or bool(data_server)
-        or data_client.startswith(b"SSH")
-    )
-    return probably_no_http
+    if data_server or data_client.startswith(b"SSH"):
+        return True
+    return not _still_could_be_http(data_client)
+
+
+def _still_could_be_http(data: bytes) -> bool:
+    first_line, newline, _rest = data.partition(b"\n")
+    line = first_line.rstrip(b"\r")
+    method, space, _remainder = line.partition(b" ")
+    if not method.isascii() or not method.isalpha():
+        return False
+    if newline and not space:
+        return False
+    return True
 
 
 def _next_layer_bytes(nextlayer: object, name: str) -> bytes:
--- a/packages/eval/harbor/test_egress_filter.py
+++ b/packages/eval/harbor/test_egress_filter.py
@@ -240,6 +240,24 @@ class EgressFilterTest(unittest.TestCase):
         MODULE.next_layer(tls)
         self.assertIsNone(tls.layer)
 
+        for first, remainder in (
+            (b"\x16", b"\x03\x01\x00\x00"),
+            (b"\x16\x03", b"\x01\x00\x00"),
+        ):
+            with self.subTest(tls_prefix=first):
+                buffered = bytearray(first)
+                fragmented = SimpleNamespace(
+                    layer=None,
+                    context=context,
+                    data_client=lambda: bytes(buffered),
+                    data_server=lambda: b"",
+                )
+                MODULE.next_layer(fragmented)
+                self.assertIsNone(fragmented.layer)
+                buffered.extend(remainder)
+                MODULE.next_layer(fragmented)
+                self.assertIsNone(fragmented.layer)
+
         http = SimpleNamespace(
             layer=None,
             context=context,
--- a/packages/eval/harbor/test_egress_filter.py
+++ b/packages/eval/harbor/test_egress_filter.py
@@ -175,45 +175,44 @@ class EgressFilterTest(unittest.TestCase):
                 MODULE.http_connect(allowed)
                 self.assertIsNone(allowed.response, host)
 
-    def test_tcp_start_kills_raw_tunnels_and_records_them(self) -> None:
-        with tempfile.TemporaryDirectory() as directory:
-            MODULE.AUDIT_PATH = Path(directory) / "hits.jsonl"
-            killed: list[str] = []
-            flow = type(
-                "Flow",
-                (),
-                {"server_conn": SimpleNamespace(address=("ssh.github.com", 443))},
-            )()
-            flow.kill = lambda: killed.append("killed")
-            MODULE.tcp_start(flow)
-            self.assertEqual(killed, ["killed"])
-            record = json.loads(MODULE.AUDIT_PATH.read_text().splitlines()[0])
-            self.assertEqual(record["ruleId"], "raw_tunnel")
-            self.assertEqual(record["host"], "ssh.github.com")
-            self.assertEqual(record["normalizedPath"], ":443")
-
-    def test_tcp_message_drops_raw_payloads(self) -> None:
-        message = SimpleNamespace(content=b"SSH-2.0-test\r\n")
-        killed: list[str] = []
-        flow = SimpleNamespace(messages=[message], killable=True)
-        flow.kill = lambda: killed.append("killed")
-        MODULE.tcp_message(flow)
-        self.assertEqual(message.content, b"")
-        self.assertEqual(killed, ["killed"])
-
     def test_next_layer_replaces_raw_tcp_with_a_closer(self) -> None:
-        class TCPLayer:
+        class FakeTCPLayer:
             pass
 
+        class CloseConnection:
+            def __init__(self, connection: object) -> None:
+                self.connection = connection
+
         with tempfile.TemporaryDirectory() as directory:
             MODULE.AUDIT_PATH = Path(directory) / "hits.jsonl"
-            context = SimpleNamespace(server=SimpleNamespace(address=("ssh.github.com", 443)))
-            nextlayer = SimpleNamespace(layer=TCPLayer(), context=context)
+            previous_tcp = MODULE.TCPLayer
+            previous_commands = MODULE.proxy_commands
+            self.addCleanup(setattr, MODULE, "TCPLayer", previous_tcp)
+            self.addCleanup(setattr, MODULE, "proxy_commands", previous_commands)
+            MODULE.TCPLayer = FakeTCPLayer
+            MODULE.proxy_commands = SimpleNamespace(CloseConnection=CloseConnection)
+            client = object()
+            server = SimpleNamespace(address=("ssh.github.com", 443))
+            context = SimpleNamespace(client=client, server=server, layers=[], options=None)
+            nextlayer = SimpleNamespace(layer=FakeTCPLayer(), context=context)
             MODULE.next_layer(nextlayer)
             self.assertIsInstance(nextlayer.layer, MODULE.CloseRawLayer)
+            self.assertIsInstance(nextlayer.layer, MODULE.Layer)
+            commands = list(nextlayer.layer.handle_event(object()))
+            self.assertEqual([command.connection for command in commands], [client, server])
+            self.assertTrue(all(isinstance(command, CloseConnection) for command in commands))
             record = json.loads(MODULE.AUDIT_PATH.read_text().splitlines()[0])
             self.assertEqual(record["ruleId"], "raw_tunnel")
             self.assertEqual(record["host"], "ssh.github.com")
+
+    def test_next_layer_leaves_non_tcp_layers_alone(self) -> None:
+        class HTTPLayer:
+            pass
+
+        original = HTTPLayer()
+        nextlayer = SimpleNamespace(layer=original, context=SimpleNamespace())
+        MODULE.next_layer(nextlayer)
+        self.assertIs(nextlayer.layer, original)
 
     def test_audit_escapes_line_separators_so_python_and_typescript_agree(self) -> None:
         with tempfile.TemporaryDirectory() as directory:
--- a/packages/eval/harbor/test_egress_filter.py
+++ b/packages/eval/harbor/test_egress_filter.py
@@ -175,9 +175,37 @@ class EgressFilterTest(unittest.TestCase):
                 MODULE.http_connect(allowed)
                 self.assertIsNone(allowed.response, host)
 
+    def test_tcp_start_kills_raw_tunnels_and_records_them(self) -> None:
+        with tempfile.TemporaryDirectory() as directory:
+            MODULE.AUDIT_PATH = Path(directory) / "hits.jsonl"
+            killed: list[str] = []
+            flow = type(
+                "Flow",
+                (),
+                {"server_conn": SimpleNamespace(address=("ssh.github.com", 443))},
+            )()
+            flow.kill = lambda: killed.append("killed")
+            MODULE.tcp_start(flow)
+            self.assertEqual(killed, ["killed"])
+            record = json.loads(MODULE.AUDIT_PATH.read_text().splitlines()[0])
+            self.assertEqual(record["ruleId"], "raw_tunnel")
+            self.assertEqual(record["host"], "ssh.github.com")
+            self.assertEqual(record["normalizedPath"], ":443")
+
+    def test_tcp_message_drops_raw_payloads(self) -> None:
+        message = SimpleNamespace(content=b"SSH-2.0-test\r\n")
+        killed: list[str] = []
+        flow = SimpleNamespace(messages=[message], killable=True)
+        flow.kill = lambda: killed.append("killed")
+        MODULE.tcp_message(flow)
+        self.assertEqual(message.content, b"")
+        self.assertEqual(killed, ["killed"])
+
     def test_next_layer_replaces_raw_tcp_with_a_closer(self) -> None:
         class FakeTCPLayer:
-            pass
+            def __init__(self, context: object) -> None:
+                self.context = context
+                context.layers.append(self)
 
         class CloseConnection:
             def __init__(self, connection: object) -> None:
@@ -193,17 +221,30 @@ class EgressFilterTest(unittest.TestCase):
             MODULE.proxy_commands = SimpleNamespace(CloseConnection=CloseConnection)
             client = object()
             server = SimpleNamespace(address=("ssh.github.com", 443))
-            context = SimpleNamespace(client=client, server=server, layers=[], options=None)
-            nextlayer = SimpleNamespace(layer=FakeTCPLayer(), context=context)
+            sibling = object()
+            context = SimpleNamespace(
+                client=client, server=server, layers=[sibling], options=None
+            )
+            current = FakeTCPLayer(context)
+            nextlayer = SimpleNamespace(layer=current, context=context)
             MODULE.next_layer(nextlayer)
             self.assertIsInstance(nextlayer.layer, MODULE.CloseRawLayer)
             self.assertIsInstance(nextlayer.layer, MODULE.Layer)
+            self.assertEqual(context.layers, [sibling, nextlayer.layer])
+            self.assertNotIn(current, context.layers)
             commands = list(nextlayer.layer.handle_event(object()))
             self.assertEqual([command.connection for command in commands], [client, server])
             self.assertTrue(all(isinstance(command, CloseConnection) for command in commands))
             record = json.loads(MODULE.AUDIT_PATH.read_text().splitlines()[0])
             self.assertEqual(record["ruleId"], "raw_tunnel")
             self.assertEqual(record["host"], "ssh.github.com")
+
+    def test_next_layer_leaves_an_unclassified_layer_alone(self) -> None:
+        context = SimpleNamespace(layers=[])
+        nextlayer = SimpleNamespace(layer=None, context=context)
+        MODULE.next_layer(nextlayer)
+        self.assertIsNone(nextlayer.layer)
+        self.assertEqual(context.layers, [])
 
     def test_next_layer_leaves_non_tcp_layers_alone(self) -> None:
         class HTTPLayer:
--- a/packages/eval/harbor/test_egress_filter.py
+++ b/packages/eval/harbor/test_egress_filter.py
@@ -308,11 +308,29 @@ class EgressFilterTest(unittest.TestCase):
                     "response": SimpleNamespace(
                         status_code=101, headers={"upgrade": "websocket"}
                     ),
+                    "websocket": object(),
                     "server_conn": SimpleNamespace(address=("origin", 19082)),
                 },
             )()
             MODULE.response(websocket)
             self.assertEqual(len(MODULE.AUDIT_PATH.read_text().splitlines()), 1)
+
+            invalid = type(
+                "Flow",
+                (),
+                {
+                    "response": SimpleNamespace(
+                        status_code=101, headers={"upgrade": "websocket"}
+                    ),
+                    "server_conn": SimpleNamespace(address=("origin", 19082)),
+                },
+            )()
+            MODULE.response(invalid)
+            self.assertEqual(len(MODULE.AUDIT_PATH.read_text().splitlines()), 2)
+            self.assertEqual(
+                json.loads(MODULE.AUDIT_PATH.read_text().splitlines()[1])["ruleId"],
+                "raw_tunnel",
+            )
 
     def test_next_layer_replaces_raw_tcp_with_a_closer(self) -> None:
         class FakeTCPLayer:
--- a/packages/eval/harbor/test_egress_filter.py
+++ b/packages/eval/harbor/test_egress_filter.py
@@ -229,7 +229,26 @@ class EgressFilterTest(unittest.TestCase):
             self.assertEqual(record["ruleId"], "raw_tunnel")
             self.assertEqual(record["host"], "ssh.github.com")
 
-    def test_next_layer_leaves_tls_and_http_for_the_builtin_classifier(self) -> None:
+    def test_next_layer_routes_tls_and_http_without_raw_fallback(self) -> None:
+        class FakeServerTLSLayer:
+            child_layer: object | None = None
+
+            def __init__(self, context: object) -> None:
+                self.context = context
+                context.layers.append(self)
+
+        class FakeClientTLSLayer:
+            def __init__(self, context: object) -> None:
+                self.context = context
+                context.layers.append(self)
+
+        previous_server_tls = MODULE.ServerTLSLayer
+        previous_client_tls = MODULE.ClientTLSLayer
+        self.addCleanup(setattr, MODULE, "ServerTLSLayer", previous_server_tls)
+        self.addCleanup(setattr, MODULE, "ClientTLSLayer", previous_client_tls)
+        MODULE.ServerTLSLayer = FakeServerTLSLayer
+        MODULE.ClientTLSLayer = FakeClientTLSLayer
+
         context = SimpleNamespace(layers=[])
         tls = SimpleNamespace(
             layer=None,
@@ -245,18 +264,21 @@ class EgressFilterTest(unittest.TestCase):
             (b"\x16\x03", b"\x01\x00\x00"),
         ):
             with self.subTest(tls_prefix=first):
-                buffered = bytearray(first)
+                fragmented_context = SimpleNamespace(layers=[])
                 fragmented = SimpleNamespace(
                     layer=None,
-                    context=context,
-                    data_client=lambda: bytes(buffered),
+                    context=fragmented_context,
+                    data_client=lambda first=first: first,
                     data_server=lambda: b"",
                 )
                 MODULE.next_layer(fragmented)
-                self.assertIsNone(fragmented.layer)
-                buffered.extend(remainder)
-                MODULE.next_layer(fragmented)
-                self.assertIsNone(fragmented.layer)
+                self.assertIsInstance(fragmented.layer, FakeServerTLSLayer)
+                self.assertIsInstance(fragmented.layer.child_layer, FakeClientTLSLayer)
+                self.assertEqual(
+                    fragmented_context.layers,
+                    [fragmented.layer, fragmented.layer.child_layer],
+                )
+                self.assertTrue((first + remainder).startswith(b"\x16\x03"))
 
         http = SimpleNamespace(
             layer=None,
--- a/packages/eval/harbor/test_egress_filter.py
+++ b/packages/eval/harbor/test_egress_filter.py
@@ -201,6 +201,93 @@ class EgressFilterTest(unittest.TestCase):
         self.assertEqual(message.content, b"")
         self.assertEqual(killed, ["killed"])
 
+    def test_next_layer_closes_raw_tcp_before_the_builtin_classifier(self) -> None:
+        class CloseConnection:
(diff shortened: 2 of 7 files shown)


packages/eval/harbor/test_egress_filter.py
```diff
@@ -121,46 +121,45 @@ def make(status, body, headers):
...
+            previous_tcp = MODULE.TCPLayer
+            previous_commands = MODULE.proxy_commands
+            self.addCleanup(setattr, MODULE, "TCPLayer", previous_tcp)
+            self.addCleanup(setattr, MODULE, "proxy_commands", previous_commands)
+            MODULE.TCPLayer = FakeTCPLayer
+            MODULE.proxy_commands = SimpleNamespace(CloseConnection=CloseConnection)
+            client = object()
+            server = SimpleNamespace(address=("ssh.github.com", 443))
+            context = SimpleNamespace(client=client, server=server, layers=[], options=None)
+            nextlayer = SimpleNamespace(layer=FakeTCPLayer(), context=context)
```
liugddx: This fake does not model the Layer contract that matters here: a real TCPLayer appends itself to context.layers during construction. As a result, the test cannot detect the stale TCPLayer left behind by the replacement.

Please either construct a real mitmproxy 12.2.3 TCPLayer or make the fixture reproduce its layer-stack registration, then assert that context.layers contains the closer and no longer contains the replaced TCP layer. Keeping the existing CloseConnection assertions is useful.

<details>
<summary>中文说明</summary>

这个 fake 没有模拟本次最关键的 Layer 契约：真实 TCPLayer 在构造时会把自己加入 context.layers。因此当前测试无法发现替换后残留的旧 TCPLayer。

建议直接构造 mitmproxy 12.2.3 的真实 TCPLayer，或者让 fixture 模拟其层栈注册行为；随后断言 context.layers 包含关闭层且不再包含被替换的 TCP 层。现有的 CloseConnection 断言可以继续保留。

</details>

packages/eval/harbor/egress_filter.py
```diff
@@ -119,21 +123,9 @@ def http_connect(flow: object) -> None:
...
-    messages = getattr(flow, "messages", None)
-    if messages:
-        messages[-1].content = b""
-    kill_flow(flow)
-
-
 def next_layer(nextlayer: object) -> None:
     current = getattr(nextlayer, "layer", None)
-    if current is None or type(current).__name__ != "TCPLayer":
+    if TCPLayer is None or not isinstance(current, TCPLayer):
```
liugddx, line 212: When the built-in NextLayer addon constructs TCPLayer(context), Layer.__init__ has already appended that TCPLayer to context.layers. Constructing CloseRawLayer(context) appends the closer as well, but the replaced TCPLayer is never removed. The resulting stack is [..., TCPLayer, CloseRawLayer] even though nextlayer.layer points only to CloseRawLayer.

Please remove the replaced layer from context.layers as part of the replacement and assert that the final stack contains the closer in the old layer's position. Otherwise context.layers no longer represents the actual protocol stack and may affect later stack matching or diagnostics.

<details>
<summary>中文说明</summary>

内置 NextLayer addon 创建 TCPLayer(context) 时，Layer.__init__ 已经把该 TCPLayer 加入 context.layers。随后构造 CloseRawLayer(context) 又会追加关闭层，但被替换的 TCPLayer 没有移除。最终层栈会变成 [..., TCPLayer, CloseRawLayer]，而 nextlayer.layer 实际只指向 CloseRawLayer。

请在替换时从 context.layers 移除旧层，并断言关闭层占据旧层的位置。否则 context.layers 不再表示真实协议栈，可能影响后续 stack matching 或诊断逻辑。

</details>

liugddx (commented):
The follow-up is aligned with the parent PR review: removing the unused tcp_start/tcp_message flow.kill hooks, keeping a single pre-relay next_layer path, using isinstance, and asserting both CloseConnection commands are all good changes.

One issue remains before merge: constructing CloseRawLayer through Layer.__init__ appends it to context.layers without removing the TCPLayer being replaced, leaving a stale protocol layer in the stack. The test uses a fake that does not register itself in context.layers, so it misses this behavior.

Please preserve the layer-stack invariant and extend the test accordingly. After that correction, this is ready to merge.

<details>
<summary>中文评审结论</summary>

这个 follow-up 与父 PR 的评审结论一致：删除未使用的 tcp_start/tcp_message flow.kill hooks、只保留 relay 前的 next_layer 阻断路径、使用 isinstance，以及断言 client/server 两个 CloseConnection 命令，都是正确的改动。

合并前还剩一个问题：CloseRawLayer 通过 Layer.__init__ 构造时会被追加到 context.layers，但被替换的 TCPLayer 没有移除，导致协议层栈残留旧层。当前测试使用的 fake 不会把自己注册到 context.layers，因此没有发现这一行为。

请保持层栈不变量并补充对应测试。修正后即可合并。

</details>

Astro-Han (commented):
A fresh live probe changes the conclusion from the earlier review: removing `tcp_start` and `tcp_message` is not safe with the current replacement.

In mitmproxy 12.2.3, script addons receive `next_layer` before the built-in classifier has populated `nextlayer.layer`. A live probe showed `nextlayer.layer` was `None` at both script hook calls, so the `isinstance(current, TCPLayer)` branch never ran. After this PR removed the TCP hooks, a CONNECT raw tunnel relayed bytes in both directions and produced no `raw_tunnel` audit record.

HTTP 101 upgrades are a separate path. Mitmproxy constructs their `TCPLayer` inside the HTTP layer without another `next_layer` hook. The same probe confirmed bidirectional raw payload and no audit record for that path as well.

This also changes the earlier layer-stack finding. The stale `TCPLayer` is real, but it is secondary: the closer is not reached on either production path. Fixing the stack alone would not make this safe.

The current test hides the problem by constructing a fake with `.layer` already set. In production it is still `None` when the script hook runs. CI also runs this test without mitmproxy installed, so the fallback imports and fake types do not verify the real addon order or connection behavior.

I suggest keeping the existing TCP hooks until an equivalent replacement has been exercised through a real proxy. A complete replacement needs to cover both raw-protocol selection and HTTP 101 upgrades. One clean option is to let mitmproxy's native `rawtcp=false` setting own the fail-closed guarantee, then keep narrow hooks for audit. If the custom closer remains, it should run only after the built-in classifier has selected `TCPLayer`, remove that layer from `context.layers`, and handle the 101 path separately.

The verification should use mitmproxy 12.2.3 and assert zero relayed bytes, the expected audit record, connection closure, normal HTTPS behavior, and WebSocket compatibility.

I used Claude for the live mitmproxy probe and Codex reviewers for independent source and lifecycle checks, then verified the current head and CI state.

<details>
<summary>中文审查结论</summary>

新的真实代理探针改变了此前 review 的结论：在当前替代方案下，删除 `tcp_start` 和 `tcp_message` 并不安全。

在 mitmproxy 12.2.3 中，script addon 收到 `next_layer` hook 时，内置 classifier 还没有为 `nextlayer.layer` 赋值。真实代理探针显示两次 script hook 中的 `nextlayer.layer` 都是 `None`，因此 `isinstance(current, TCPLayer)` 分支从未执行。这个 PR 删除 TCP hooks 后，CONNECT raw tunnel 可以双向传输字节，也没有产生 `raw_tunnel` 审计记录。

HTTP 101 upgrade 是另一条独立路径。Mitmproxy 会直接在 HTTP layer 内构造 `TCPLayer`，不会再次触发当前脚本依赖的 `next_layer` hook。真实探针同样确认这条路径可以双向传输 raw payload，且没有审计记录。

这也改变了之前 layer stack finding 的优先级。旧 `TCPLayer` 残留确实存在，但它只是次要问题：当前 closer 在两条生产路径上都没有执行。只修复层栈不能解决实际问题。

当前测试通过预先设置 fake 的 `.layer` 避开了真实生命周期。生产中脚本执行 hook 时该字段仍然是 `None`。CI 又是在没有安装 mitmproxy 的环境中运行测试，因此 fallback import 和 fake type 无法验证真实 addon 顺序或连接行为。

建议在等价替代方案经过真实代理验证之前，先保留现有 TCP hooks。完整替代方案必须同时覆盖 raw protocol selection 和 HTTP 101 upgrade。一种更干净的选择是让 mitmproxy 原生的 `rawtcp=false` 负责 fail-closed，再用较窄的 hooks 负责审计。如果继续保留自定义 closer，它只能在内置 classifier 已经选择 `TCPLayer` 后运行，需要从 `context.layers` 移除旧层，并单独覆盖 HTTP 101 路径。

验证应使用真实 mitmproxy 12.2.3，并断言零 payload、正确的审计记录、连接关闭、正常 HTTPS 不受影响，以及 WebSocket 兼容。

本次使用 Claude 完成真实 mitmproxy 探针，并使用 Codex reviewers 独立核对源码和生命周期；随后核对了当前 head 与 CI 状态。

</details>

1625567290:
Addressed in 06cf8017f.

Astro-Han's live probe is right: in mitmproxy 12.2.3 the script `next_layer` hook sees `nextlayer.layer is None`, so `isinstance(..., TCPLayer)` never ran after this PR deleted the TCP hooks. I reproduced that gap and restored `tcp_start` / `tcp_message`.

- CONNECT `tbench.ai:443` → `451 tbench_domain`
- `https://example.com/` via the proxy → `200`
- CONNECT `ssh.github.com:443` → `200` then no banner (timeout); `hits.jsonl` has `raw_tunnel`
- When a `TCPLayer` *is* already assigned, replacement now puts `CloseRawLayer` in that slot and removes the old layer from `context.layers`. The fixture registers itself on `context.layers` the way `Layer.__init__` does.
- `next_layer` still no-ops when `layer is None`, which is the production order. I did not switch to `rawtcp=false` or try to close unclassified layers from that hook.

`python3.13` and `maka-eval-egress-proxy:12.2.3` both run the filter suite 9/9.

1625567290:
Live verification is now in-repo as `MAKA_EVAL_EGRESS_PROXY_TEST=1 python3 harbor/test_egress_filter_live.py` (`f5e42b33f`). I ran it against `maka-eval-egress-proxy:12.2.3` with the current filter mounted: **5/5**.

| Probe | Result |
|---|---|
| plain HTTP to a local origin | `http-ok` |
| `https://example.com/` | **200** |
| `CONNECT tbench.ai:443` | **451** `tbench_domain` |
| raw CONNECT to a banner server | **200** then **zero** bytes either way; origin `raw_recv==0`; `raw_tunnel` |
| HTTP 101 `Upgrade: raw` | **101** headers only, no `UPGRADE-BANNER`; origin `upgrade_recv==0`; `raw_tunnel` |
| WebSocket 101 | **101** and `ws-ok`; **no** `raw_tunnel` for that port |

That is why this PR keeps `tcp_start` / `tcp_message` instead of switching the image to `rawtcp=false`: the TCP hooks close raw CONNECT and the 101-to-TCPLayer path, and they do not take down a real WebSocket upgrade. `next_layer` still only replaces an already-assigned `TCPLayer` and removes it from `context.layers`.

CI still skips this suite unless the env is set, same as the namespace test — GitHub runners do not have the pinned image.

packages/eval/harbor/test_egress_filter_live.py
```diff
@@ -0,0 +1,388 @@
...
+    def _recv_all(cls, sock: socket.socket, limit: int = 8192) -> bytes:
+        sock.settimeout(6)
+        data = b""
+        try:
+            while len(data) < limit:
+                chunk = sock.recv(4096)
+                if not chunk:
+                    break
+                data += chunk
+        except (TimeoutError, ConnectionResetError, BrokenPipeError):
```
Astro-Han: **[P2] Distinguish an open timeout from proxy-initiated closure**

Catching `TimeoutError` here and returning the accumulated bytes makes an open, silent tunnel indistinguishable from EOF/reset. The raw and 101 origins also close themselves after 3 s and 2 s, so the current `body == b""` assertions can pass without the proxy closing anything. Against pinned mitmproxy 12.2.3 and an upstream held open for 30–60 seconds, CONNECT returned 200 and the client remained open for 8–20 seconds; `TCPFlow.kill()` only marks the flow, and `TCPLayer` continues holding both transports. Please use a long-lived origin, make timeout a test failure, and assert EOF/reset within a short bound while retaining the zero-payload and audit assertions.

packages/eval/harbor/test_egress_filter_live.py
```diff
@@ -0,0 +1,388 @@
...
+import unittest
+from pathlib import Path
+
+HARBOR_DIR = Path(__file__).parent
+PROXY_IMAGE = "maka-eval-egress-proxy:12.2.3"
+PROJECT = "maka-eval-egress-proxy-live"
+NETWORK = f"{PROJECT}_net"
+PROXY = f"{PROJECT}-proxy"
+ORIGIN = f"{PROJECT}-origin"
+ORIGIN_IMAGE = "python:3.12-slim"
```
Astro-Han, line 46: **[P2] Give each live run isolated Docker resources**

The fixed project, container, network, and host-port names make overlapping runs destructive: each run's `_down()` removes the other run's proxy/origin/network. Two concurrent invocations reproduce one successful 5/5 run while the other loses containers and fails multiple probes. Please derive unique names per run and use a dynamically assigned host port, while still cleaning only the resources owned by that run.

Astro-Han (commented):
Thanks for following up with real mitmproxy coverage and for restoring the TCP hooks after the production-order evidence. The current head does prevent payload relay for raw CONNECT and HTTP 101 paths without breaking ordinary HTTP(S) or WebSocket traffic.

I found two remaining P2 issues before approval; the inline comments contain the reproductions and requested verification.

From first principles, the remaining production problem is ownership of connection closure. The current `next_layer -> CloseRawLayer` branch is not reached in the observed addon order because the script hook sees `layer=None`, while the reachable `tcp_start` / `tcp_message` path uses `TCPFlow.kill()`, which marks the flow but does not close either transport. A long-lived peer can therefore retain both sockets even though payload is suppressed. Please make one production-reachable layer/connection seam the sole closing authority, then remove the superseded parallel path and its fake-only tests.

The live harness belongs in this PR because unit fakes cannot prove the mitmproxy lifecycle, but it should assert proxy-initiated EOF/reset against a long-lived origin and isolate each run's Docker resources. Once the final path is settled, please also update the PR title and body: they still describe the earlier next_layer-only design and removal of the TCP hooks.

I verified 14/14 focused Python unit tests and 5/5 live tests on the pinned image, then reproduced the open-socket behavior with long-lived upstreams. CI is green, although the opt-in live suite is skipped by default. I found no P0 or P1 issue.

<details>
<summary>中文对照</summary>

感谢补充真实 mitmproxy 验证，也感谢根据生产 addon 顺序的证据恢复 TCP hooks。当前 head 已经能阻止 raw CONNECT 和 HTTP 101 的 payload 转发，同时不影响普通 HTTP(S) 与 WebSocket。

合并前仍有两项 P2，具体复现和验证要求见行内评论。

从第一性原理看，剩余的生产问题是连接关闭 authority。当前 `next_layer -> CloseRawLayer` 分支在真实 addon 顺序中不可达，因为 script hook 看到的是 `layer=None`；实际可达的 `tcp_start` / `tcp_message` 路径调用 `TCPFlow.kill()`，它只标记 flow，不会关闭两端 transport。因此即使 payload 被清空，长驻 peer 仍能持续占用两端 socket。请让一个生产可达的 layer/connection seam 成为唯一关闭 authority，并删除被替代的并行路径和 fake-only tests。

真实代理 harness 应与本 PR 一起交付，因为 unit fake 无法证明 mitmproxy 生命周期；但它需要针对长驻 origin 明确断言 proxy 主动产生 EOF/reset，并隔离每次运行的 Docker 资源。最终路径确定后，也请更新 PR 标题和正文；当前文字仍描述已被推翻的 next_layer-only、删除 TCP hooks 的方案。

我验证了 14/14 Python unit tests 和 pinned image 上的 5/5 live tests，并使用长驻 upstream 复现了 socket 保持打开的行为。CI 当前全绿，但 opt-in live suite 默认会跳过。没有发现 P0 或 P1。

</details>

Disclosure: I used three Codex reviewers and Claude for independent lifecycle, test-authenticity, and simplification checks. I reproduced the surviving findings and own this review; AI assistance does not replace the repository's required independent human review.

packages/eval/harbor/test_egress_filter.py
1625567290: Done in 6605e242d.

Script `next_layer` now assigns `CloseRawLayer` when client data looks like raw TCP *before* the built-in classifier runs (layer is still `None`). Built-in NextLayer then leaves that layer alone. The unit fixture also models `context.layers` registration for the replace path and asserts the closer occupies the stack without a leftover TCPLayer.

Primary production closer for CONNECT is `CloseRawLayer` → `CloseConnection`, not the fake-only post-classifier replace.

packages/eval/harbor/egress_filter.py
1625567290: Done in 6605e242d.

When a TCPLayer is already present, replacement still removes it from `context.layers` and inserts `CloseRawLayer` at that index. On the production CONNECT path the script hook assigns the closer while `layer` is still `None`, so there is no stale TCPLayer to leave behind.

packages/eval/harbor/test_egress_filter_live.py
1625567290: Done in 6605e242d.

Origins stay open until the peer closes (60s). The client path uses `_recv_until_close` with a 2s timeout: timeout fails the test; EOF/reset is required. Payload stays empty and `raw_tunnel` is still asserted.

Live suite against pinned `maka-eval-egress-proxy:12.2.3`: 5/5.

packages/eval/harbor/test_egress_filter_live.py
1625567290: Done in 6605e242d.

Each live run uses unique network/container names (`pid` + uuid) and a dynamically published host port (`127.0.0.1::8080`). `_down` only removes the resources that run created.

1625567290:
Addressed the remaining review items in 6605e242d.

**Closing authority**
- CONNECT raw: script `next_layer` (runs before the built-in classifier) assigns `CloseRawLayer` when client data looks like raw TCP; both transports get `CloseConnection`.
- HTTP 101 non-WebSocket: `rawtcp=false` so mitmproxy never starts a `TCPLayer` and closes after 101. WebSocket still upgrades.
- `TCPFlow.kill()` is no longer the claimed closer; `tcp_start`/`tcp_message` are last-resort only.

**Live harness**
- Long-lived origins; timeout fails; assert EOF/reset + zero payload + `raw_tunnel`.
- Isolated Docker names and dynamic host port per run.

**PR title/body** updated to match this design.

Verified: unit 17/17; live 5/5 on `maka-eval-egress-proxy:12.2.3`.

packages/eval/harbor/egress_filter.py
```diff
@@ -180,6 +232,36 @@ def connect_target_url(flow: object) -> str:
...
+        return False
+    probably_no_http = (
+        len(data_client) < 3
+        or b" " not in data_client
+        or (data_client.find(b" ") > data_client.find(b"\n"))
+        or not data_client[:3].isalpha()
+        or bool(data_server)
+        or data_client.startswith(b"SSH")
+    )
+    return probably_no_http
```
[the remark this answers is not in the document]
1625567290: Done in 71004e15e.

`looks_like_raw_tcp` no longer copies mitmproxy's `probably_no_http` test. Incomplete prefixes (`G`, `GE`, `GET`, `GET `, `GET / HTTP/1.1` without a newline) stay unclassified so the built-in HttpLayer can finish the request line. SSH banners, server-first greetings, and non-method bytes still assign `CloseRawLayer`.

packages/eval/README.md
```diff
@@ -84,7 +84,10 @@ download and verifier phases retain their native network policy. Build the pinne
 `maka-eval-egress-proxy:12.2.3` image from `harbor/egress-proxy/Dockerfile` before running the
 cohort. `MAKA_EVAL_EGRESS_NAMESPACE_TEST=1 python3 harbor/test_cell_egress_namespace.py` brings up
 the overlay and the checked-in policy and asserts that contract in a real cell namespace; it needs
-a Docker daemon and outbound network, and skips otherwise. This URL policy is a blocklist for known
+a Docker daemon and outbound network, and skips otherwise.
+`MAKA_EVAL_EGRESS_PROXY_TEST=1 python3 harbor/test_egress_filter_live.py` starts the
+pinned mitmproxy image and asserts raw CONNECT and HTTP 101 upgrades relay no
+bytes, write `raw_tunnel`, and leave HTTPS and WebSocket working. This URL policy is a blocklist for known
```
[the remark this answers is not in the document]
1625567290: Done in 71004e15e.

The live-test sentence now states it needs a Docker daemon, the pinned image, and outbound network, and skips otherwise.

packages/eval/harbor/test_egress_filter_live.py
```diff
@@ -0,0 +1,420 @@
...
+    def setUpClass(cls) -> None:
+        if not docker_available():
+            raise unittest.SkipTest("Docker daemon is unavailable")
+        images = subprocess.run(
+            ["docker", "image", "inspect", PROXY_IMAGE],
+            capture_output=True,
+            timeout=COMMAND_TIMEOUT_S,
+        )
+        if images.returncode != 0:
+            raise unittest.SkipTest(f"{PROXY_IMAGE} is not present")
```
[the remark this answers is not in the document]
1625567290: Done in 2ecb2db66.

`setUpClass` now inspects `ORIGIN_IMAGE` the same way as the pinned proxy image and skips with that name if it is missing. It does not pull under the 60s `docker run` timeout.

packages/eval/harbor/test_egress_filter_live.py
```diff
@@ -0,0 +1,420 @@
...
+
+    @classmethod
+    def audit_records(cls) -> list[dict[str, object]]:
+        listed = subprocess.run(
+            ["docker", "exec", cls.proxy, "cat", "/opt/maka-egress-state/hits.jsonl"],
+            capture_output=True,
+            text=True,
+            timeout=COMMAND_TIMEOUT_S,
+        )
+        return [json.loads(line) for line in listed.stdout.splitlines() if line.strip()]
```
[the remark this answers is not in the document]
1625567290: Done in 2ecb2db66.

`audit_records` now fails on a non-zero `docker exec` the same way `origin_stats` already does, including the command stderr.

packages/eval/harbor/egress_filter.py
```diff
@@ -103,14 +103,52 @@ def public_trajectory_repository(host: str, path_query: str) -> bool:
...
+    headers = getattr(response, "headers", None)
+    upgrade = ""
+    if headers is not None:
+        try:
+            upgrade = str(headers.get("upgrade", "") or "")
+        except Exception:
+            upgrade = ""
+    if upgrade.lower() == "websocket":
+        return
+    record_raw_tunnel(flow)
```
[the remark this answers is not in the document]
1625567290: Done in 590565892.

`response` now skips `raw_tunnel` only when `flow.websocket` is set. That is what mitmproxy 12.2.3 assigns before `HttpResponseHook` for a completed handshake (`Upgrade` + `Sec-WebSocket-Version: 13` + websocket enabled). A 101 with only a websocket Upgrade header is audited. The unit fixture covers both cases.

packages/eval/harbor/test_egress_filter_live.py
```diff
@@ -0,0 +1,428 @@
...
+    def audit_records(cls) -> list[dict[str, object]]:
+        listed = subprocess.run(
+            ["docker", "exec", cls.proxy, "cat", "/opt/maka-egress-state/hits.jsonl"],
+            capture_output=True,
+            text=True,
+            timeout=COMMAND_TIMEOUT_S,
+        )
+        if listed.returncode != 0:
+            raise AssertionError(listed.stderr)
+        return [json.loads(line) for line in listed.stdout.splitlines() if line.strip()]
```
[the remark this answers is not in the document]
1625567290: Done in 590565892.

`audit_records` cats the file only when it exists. A missing audit log is an empty record list, so the WebSocket test can run alone. Other `docker exec` failures still raise with stderr.

packages/eval/harbor/egress_filter.py
```diff
@@ -180,6 +229,47 @@ def connect_target_url(flow: object) -> str:
...
+    """Close only when the bytes cannot still become HTTP or TLS.
+
+    Script next_layer runs before mitmproxy 12.2.3's classifier. Copying its
+    `probably_no_http` test here would treat `GET` / `GET ` as raw and assign
+    CloseRawLayer while the request line is still arriving. With rawtcp=false
+    the built-in path would have kept waiting for HttpLayer.
+    """
+    data_client = _next_layer_bytes(nextlayer, "data_client")
+    data_server = _next_layer_bytes(nextlayer, "data_server")
+    if starts_like_tls_record(data_client):
```
Astro-Han: **P1 — Keep incomplete TLS record prefixes undecided.** In mitmproxy 12.2.3, `starts_like_tls_record()` returns false until at least three bytes are available. If the first `DataReceived` contains only `b"\x16"` or `b"\x16\x03"`, `_still_could_be_http()` is also false, so this path installs `CloseRawLayer` and closes a valid HTTPS tunnel before the next byte arrives. Please add a `could_start_tls_record()`/tri-state check that waits for the incomplete TLS header, then pin both 1-byte and 2-byte fragmentation cases through the `next_layer` hook.

packages/eval/harbor/egress_filter.py
```diff
@@ -132,12 +170,23 @@ def tcp_message(flow: object) -> None:
...
-    if current is None or type(current).__name__ != "TCPLayer":
-        return
     context = getattr(nextlayer, "context", None)
+    if current is None:
+        if not looks_like_raw_tcp(nextlayer):
+            return
+        record_raw_tunnel(context)
+        nextlayer.layer = CloseRawLayer(context)
+        return
+    if TCPLayer is None or not isinstance(current, TCPLayer):
```
Astro-Han, line 212: **P2 — Remove the unreachable second closing path.** With the pinned addon order, the script hook runs before the built-in classifier and therefore cannot observe an already-created `TCPLayer`; the comment above states the same production fact. Keeping this branch, `replace_layer()`, and a fake-only test creates a parallel authority that cannot protect the shipped path. Please delete them once the pre-classification `CloseRawLayer` path is complete.

Astro-Han (commented):
Thanks for moving raw-tunnel rejection to the layer that actually owns the sockets. I reviewed this head with two independent `@reviewer` passes plus a read-only `ollama-cloud/deepseek-v4-flash:high` pass, and checked the disputed path against mitmproxy 12.2.3's classifier/event ordering.

The problem is correctly defined and `CloseRawLayer` is the right authority. The remaining correctness gap is that the pre-classification decision is binary even when the protocol prefix is still incomplete: a one- or two-byte TLS ClientHello can be irreversibly classified as raw before the third byte arrives. Current CI is green, but the existing tests only cover a complete TLS prefix and cannot exercise this TCP fragmentation.

The smallest first-principles fix is a tri-state classifier at this seam: confirmed HTTP/TLS, confirmed raw, or undecided. Keep buffering while a prefix can still become a TLS record or HTTP request, and install `CloseRawLayer` only for confirmed raw traffic. Once that is the sole production authority, remove the unreachable post-classifier replacement branch and its fake-only test.

No local repository test suite was run by Codex; one reviewer performed focused source-level verification against the pinned mitmproxy implementation. Codex coordinated the passes and performed the final adjudication; external-model output was treated as unverified until checked against the code.

<details>
<summary>中文摘要</summary>

感谢把 raw tunnel 拒绝移动到真正拥有 socket 的 layer。问题定义正确，`CloseRawLayer` 也是正确权威。

当前仍有一个正确性缺口：协议前缀尚不完整时，分类器已经做了不可逆的二元决策。TLS ClientHello 首次只到达 1–2 字节时，会在第三个字节到达前被当成 raw 并关闭。当前 CI 虽然全绿，但测试只覆盖完整 TLS 前缀，无法覆盖 TCP fragmentation。

最小方案是在同一 seam 使用三态分类：确认 HTTP/TLS、确认 raw、未决。只要前缀仍可能成为 TLS record 或 HTTP request 就继续等待，只有确认 raw 才安装 `CloseRawLayer`。随后删除不可达的 post-classifier replacement 分支和 fake-only 测试，保持单一权威。

Codex 未运行本地仓库测试套件；其中一位 reviewer 针对锁定的 mitmproxy 实现做了聚焦源码验证。Codex 完成最终判断，外部模型输出在核对代码前均视为未验证输入。
</details>

packages/eval/harbor/egress_filter.py
```diff
@@ -180,6 +229,47 @@ def connect_target_url(flow: object) -> str:
...
+
+    Script next_layer runs before mitmproxy 12.2.3's classifier. Copying its
+    `probably_no_http` test here would treat `GET` / `GET ` as raw and assign
+    CloseRawLayer while the request line is still arriving. With rawtcp=false
+    the built-in path would have kept waiting for HttpLayer.
+    """
+    data_client = _next_layer_bytes(nextlayer, "data_client")
+    data_server = _next_layer_bytes(nextlayer, "data_server")
+    if starts_like_tls_record(data_client):
+        return False
```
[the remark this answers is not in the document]
1625567290: Addressed in the current head (`83a4283b2`). Incomplete TLS record prefixes are now represented as undecided by `_could_start_tls_record()`, so one-byte and two-byte ClientHello fragments are not assigned `CloseRawLayer`. Both `next_layer` fragment cases and live proxy handshakes are covered. Validation: 17 unit tests passed; 6 live tests were skipped locally because the proxy container was unavailable.

packages/eval/harbor/egress_filter.py
1625567290: Addressed in the current head (`83a4283b2`). The post-classification `TCPLayer` replacement branch, `replace_layer()` helper, and its fake-only test have been removed. `next_layer` plus `CloseRawLayer` is now the single CONNECT pre-classification authority; `tcp_start` remains only as a last-resort hook for independently admitted `tcp_hosts` or ignore flows. Validation: 17 unit tests passed.

packages/eval/harbor/egress_filter.py
1625567290: Addressed in the current head (`83a4283b2`). `_could_start_tls_record()` now keeps both `b"\x16"` and `b"\x16\x03"` undecided until the three-byte TLS prefix is available. The `next_layer` unit coverage exercises both fragment boundaries, and the live proxy suite repeats the real handshake with one-byte and two-byte first fragments. Validation: 17 unit tests passed; 6 live tests were skipped locally because the proxy container was unavailable.

packages/eval/package.json
```diff
@@ -19,7 +19,7 @@
     "clean": "node ../../scripts/clean-paths.mjs dist tsconfig.tsbuildinfo",
     "build": "tsc -p tsconfig.json",
     "typecheck": "tsc -p tsconfig.json --noEmit",
-    "test:dist": "node --test \"dist/**/*.test.js\" && python3 harbor/test_eval_framework.py && python3 harbor/test_relay_contract.py && python3 harbor/test_relay_lifecycle.py && python3 harbor/test_egress_filter.py && python3 harbor/test_run_trial_policy.py && python3 harbor/test_relay_artifacts.py && python3 harbor/test_cell_egress_namespace.py"
+    "test:dist": "node --test \"dist/**/*.test.js\" && python3 harbor/test_eval_framework.py && python3 harbor/test_relay_contract.py && python3 harbor/test_relay_lifecycle.py && python3 harbor/test_egress_filter.py && python3 harbor/test_run_trial_policy.py && python3 harbor/test_relay_artifacts.py && python3 harbor/test_cell_egress_namespace.py && python3 harbor/test_egress_filter_live.py"
```
Astro-Han: **[P2]** `test_egress_filter_live.py` is now on the `test:dist` chain, but it can never run there: the suite is gated behind `@skipUnless(MAKA_EVAL_EGRESS_PROXY_TEST == "1")`, and that variable is not set anywhere in the repository — not in `.github/`, not in `scripts/`, not in a Makefile, compose file, CI matrix, or an inline env on any npm script. The only two occurrences are the docstring in the test file and the `skipUnless` guard itself.

So the behaviour this PR is actually fixing — the close path under a real mitmproxy lifecycle — has no automated verification. What CI exercises is the fake-based unit suite, which does not import mitmproxy and runs against the fallback types.

The part that makes this P2 rather than P3 is that the skip is silent: `test:dist` reports green, so if someone later changes the close logic in a way that breaks the real lifecycle, nothing in CI will catch it, and the suite's presence on the chain makes it look covered.

Either direction resolves it:

- **Run it in CI** — set `MAKA_EVAL_EGRESS_PROXY_TEST=1` for the eval workspace and build the proxy image; the Dockerfile is in-tree and `test_cell_egress_namespace.py` already has image-build logic to model this on.
- **Or make the opt-in explicit** — take the live suite off the unconditional `test:dist` chain and give it its own script, so that a green `test:dist` does not imply this suite ran.

The second is cheaper and removes the misleading signal; the first is what actually protects the fix.

Astro-Han (commented):
Thanks @1625567290 — reviewed at exact head `83a4283b24470c214367fa956b4dd4c74ec8ef36`. One P2 and one P3; no P0/P1.

The ownership change itself reads correctly: assigning `CloseRawLayer` while the layer is still `None` puts the close on the path the production hook actually takes, and the resulting `CloseConnection` for both client and server is handled by mitmproxy's command loop rather than by killing the flow and leaving the socket open. Double-close is safe on that side — once a connection's transport has been popped, the command is a no-op, and `close_connection` cancels an already-cancelled handler idempotently. So the peer-disconnected-first and both-paths-close-at-once cases do not leak or throw.

The unit tests are good where they apply: fragmented TLS prefixes down to 1–2 bytes, HTTP prefixes arriving in pieces, SSH banners, binary payloads, and an explicit assertion that `CloseRawLayer` emits both `CloseConnection` commands.

The P2 is that none of that reaches CI — details inline.

**P3, redundant configuration.** `entrypoint.sh` already passes `--set rawtcp=false` on the mitmdump command line, and the `configure()` hook sets `ctx.options.rawtcp = False` again. It is guarded and has no side effect, so this is only a note: two places now express the same setting, and a future change to either one can silently disagree with the other.

The `kill_flow` fallback on `tcp_start` / `tcp_message` and its behaviour were already raised in earlier reviews on this PR; not repeating that here.

---

packages/eval/package.json
1625567290: Addressed in the current head (`db3c117b2`).

The unconditional Eval `test:dist` chain no longer lists the live suite. The suite now has an explicit `test:egress-proxy:live` script, and the Eval-selected CI job builds the pinned `maka-eval-egress-proxy:12.2.3` image, pulls `python:3.12-slim`, sets `MAKA_EVAL_EGRESS_PROXY_TEST=1`, and runs that script. A CI contract test pins this wiring so it cannot silently regress to a skipped green run.

Local validation: egress unit 18/18, relay contract 30/30, relay lifecycle 14 tests with 2 conditional skips, and CI contract 29/29. The 5 live cases were skipped locally because Docker is not installed; the newly queued GitHub CI job is the authoritative live run.

Astro-Han (commented):
Incremental review at exact head `db3c117b25079e27fccbc539eee5c28f4ecdb1fa`, focused on what changed since the last reviewed head.

**The previously-raised P2 is genuinely closed at the wiring level.** `ci.yml:188-198` now pulls and builds the image when Eval is selected and forces the live env flag, and a failure there fails the job — the live suite is no longer excluded from CI by construction. Two new findings, though, and the first one affects ordinary traffic.

### `[P1]` A normally-fragmented TLS ClientHello is killed as raw TCP — `packages/eval/harbor/egress_filter.py:133-134, 259-265`

```python
def starts_like_tls_record(data: bytes) -> bool:
    return len(data) >= 3 and data[0] == 0x16 and data[1] == 0x03
```

`looks_like_raw_tcp` returns early with `False` only when this is true. With one byte (`0x16`) or two (`0x16 0x03`) received so far, the length test fails, so the TLS escape hatch doesn't fire. Control flow then reaches `_still_could_be_http(data_client)`, which is `False` for a control byte — no valid request line begins with `0x16`. `looks_like_raw_tcp` therefore returns `True`, and `next_layer:195-200` immediately installs `CloseRawLayer`.

The connection is killed *before* the third byte arrives, so the bytes that would have identified it as TLS never get a chance to. This is an ordinary path — TCP is free to deliver a ClientHello in small segments, and no malformed or hostile input is required.

The docstring above the function shows the author already reasoned about exactly this hazard for HTTP: copying mitmproxy's `probably_no_http` would misclassify `GET` / `GET ` while the request line is still arriving, so the code deliberately keeps waiting. TLS needs the same treatment and currently doesn't get it.

**Suggested shape:** make the classifier tri-state rather than boolean — `raw` / `not-raw` / `undecided` — and return `undecided` while the buffer is still a viable TLS prefix (`data == b"\x16"`, or `data == b"\x16\x03"`). That matches how the HTTP side already behaves: don't decide until the evidence can exist.

Verified against the pinned 12.2.3 image: the detector returns `False` for `16`, `False` for `1603`, and `True` only at `160301`.

### `[P2]` The live harness doesn't prove the origin socket is closed — `packages/eval/harbor/test_egress_filter_live.py:294-305, 411-430`

The test waits for the client side to see EOF/reset and asserts the origin received zero payload bytes plus the expected audit record. Neither establishes that the **origin-side** socket was closed: the origin stats carry no peer-EOF or close counter, so a leaked origin connection that simply never receives data would pass this test unchanged.

The unit test's two `CloseConnection` commands aren't a substitute — they prove the addon *emitted* both, not that the real mitmproxy loop closed both ends. Recording and waiting on EOF at the origin would close the gap, along with a case that is explicitly a non-WebSocket upgrade so the tunnel path is exercised deliberately rather than incidentally.

### CI status

Exact-head `audit` is green, but `test` is `completed/cancelled` — the run (`32623743066`) was cancelled during *Install dependencies*, so the live step was skipped. There is no hosted exact-head evidence that the newly-wired live suite passes. Worth a re-run before this is considered gated.

**Validation scope.** Local runs at this head: unit 18/18, planner 29/29, Docker live 5/5. Those five live cases don't cover either finding above — neither a fragmented ClientHello nor an origin-side close assertion is among them, which is why both survived a green local run.

1625567290:
@Astro-Han Rebased onto current `main` (`a2f2a1afa`) and force-updated the branch to `a8c02f29e`; GitHub reports it mergeable.

Exact-head local validation:
- egress filter unit tests: 18/18
- CI planner contract: 29/29
- Eval build
- ASF header check
- `git diff --check`

The 5 live proxy cases remain explicitly skipped locally because Docker is unavailable; the Eval-selected CI lane builds the pinned proxy image and runs them with `MAKA_EVAL_EGRESS_PROXY_TEST=1`.

The fresh workflows are waiting for maintainer approval: [CI 32638152317](https://github.com/apache/maka/actions/runs/32638152317) and [Dependency audit 32638152278](https://github.com/apache/maka/actions/runs/32638152278). Please approve both exact-head runs.

Astro-Han:
Re-anchoring after the rebase: the head has moved to `a8c02f29e11591a7d26c429473e79e66ba65df54`, and both findings from the previous comment carry over unchanged.

`packages/eval/harbor/egress_filter.py` and `packages/eval/harbor/test_egress_filter_live.py` are byte-identical between the reviewed head `db3c117b2` and the current one, so nothing about the `[P1]` fragmented-ClientHello path or the `[P2]` origin-close gap has been affected. No re-review needed — treat the earlier comment as anchored here.

One clarification on the P1 while re-reading it: the fallback definition is only used when `from mitmproxy.net.tls import starts_like_tls_record` fails. The primary path takes mitmproxy's own implementation, which behaves the same way at the pinned 12.2.3 (`16` → false, `1603` → false, `160301` → true), so the finding does not depend on which branch is taken.

1625567290:
Thanks for re-anchoring both findings. Fixed in `0f4de78a2`:

- Fragmented one- and two-byte TLS prefixes now stay undecided via `_could_start_tls_record`. Unit coverage exercises both prefixes, and the live harness splits a real `ssl.MemoryBIO` ClientHello before completing the handshake.
- Raw CONNECT and an explicit non-WebSocket `Upgrade: raw` case now record origin-side handler closure, wait for `raw_closed` / `upgrade_closed`, and still assert that zero payload bytes were relayed.

Local validation is green: egress filter unit tests (18/18), CI planner tests (29/29), Eval TypeScript build, embedded origin-script compilation, ASF header check, and diff check. The Docker-backed live suite contains 6 cases but was skipped locally because Docker is unavailable.

The new [CI](https://github.com/apache/maka/actions/runs/32639766029) and [Dependency audit](https://github.com/apache/maka/actions/runs/32639766028) runs are currently waiting for maintainer approval after the rebase/force-update. Once approved, hosted CI will exercise the Docker-backed cases.

Astro-Han (commented):
Thanks for taking both items head-on. The origin-close proof is genuinely closed now, and I want to be specific about why, because it is the part that was hardest to get right. The remaining item still blocks.

### The origin-close finding is closed

`packages/eval/harbor/test_egress_filter_live.py:51-126` increments its counters inside the origin's own `finally` block when the raw-socket and 101-handler receive loops exit, and `:445-457,491-516` read those counters from the origin container independently, wait for them to advance, and only then assert `*_recv == 0` with no banner and an audit hit. That is an independent observation of the origin peer, not client EOF, handler cancellation, server close, or a zero-length payload standing in as self-proof. Verified by running the raw `CONNECT` and non-WebSocket 101 cases directly: 2/2 green in 3.096s.

### [P1] Fragmented 1- and 2-byte TLS record prefixes still hang on the real mitmproxy path

`egress_filter.py:268-273` now uses `_could_start_tls_record` to withhold `CloseRawLayer` for a leading `16` or `16 03`, so the connection is no longer torn down early. But nothing re-drives classification once the rest of the ClientHello arrives: the real mitmproxy 12.2.3 classifier does not re-decide or complete the handshake, so the connection moves from "closed too early" to "never resolved".

This PR's own live tests say so. `test_egress_filter_live.py:350-399,478-483` covers the 1-byte and 2-byte subcases, and both time out in `sock.recv` with TLS still at `SSLWantReadError`. Reproduced independently against the pinned Docker image at this exact head — both subcases time out after 23.194s.

The hosted result agrees: CI run `32639766029` attempt 2 at `0f4de78a2110d0bcddb080efc545082c84a6281b` is a terminal failure, and the `test` check is `completed/failure`. In that live step, 4 of 6 tests pass and the two failures are exactly these fragmented subcases; every preceding lint, format, build, typecheck and affected-test stage is green, and `audit` is `completed/success`. So this is not flaky infrastructure — the required check is red on precisely the behaviour under discussion.

One caveat on the unit-level evidence: the addon and fallback detector tests pass, but they call `next_layer` a second time by hand. That proves the helper's tri-state logic in isolation; it cannot show that the production layer stack ever performs that second call. The two real-event-loop tests are the ones that speak to production, and they fail.

**What the fix needs.** The undecided state has to be handed to a layer that will genuinely re-classify and make progress as later bytes arrive, and both real paths must go green. Please also confirm the behaviour holds whether or not `mitmproxy.net.tls.starts_like_tls_record` is importable — the original finding was anchored on the local 3-byte fallback, so a fix proven only with the upstream helper present would not close it.

---

*ran:* the pinned-Docker live suite at this exact head (raw `CONNECT` and non-WebSocket 101 green in 3.096s; the two fragmented subcases timing out after 23.194s), plus independent verification of the hosted check state for this commit.
*did-not-run:* anything outside `packages/eval/harbor`; the nine pre-existing commits were not re-reviewed, since `git range-diff` shows them patch-equivalent to the previously reviewed revision.

1625567290:
Fixed the exact CI failure at head `6fab34c02`.

Run `32639766029` reached the Docker-backed live suite and failed only the 1-byte and 2-byte fragmented ClientHello cases. The classifier was no longer killing those prefixes; the failure was in the new MemoryBIO test driver. After OpenSSL consumed a server flight, it produced the next client flight and raised `SSLWantReadError`, but the harness blocked on `recv` before flushing that pending client flight. Both peers then waited until timeout.

The handshake driver now flushes `outgoing` immediately after every `SSLWantReadError`, before reading more server bytes. A Docker-independent ordering regression proves `send -> recv -> incoming.write`, so this deadlock cannot be reintroduced while the live Docker case remains the end-to-end authority.

The branch is also rebased onto current `main` (`6b48ed36f`). Local validation passed: handshake driver 1/1, egress filter 18/18, CI planner 30/30, Eval build and typecheck, lint, format, ASF headers, Python compilation, and diff check. Docker is unavailable locally, so the six live proxy cases are left to the fresh selected CI lane rather than being claimed as local evidence.

jackwener (commented):
Reviewed exact head `6fab34c0257a1f91364018f74813bf61e1c6c762`: **NO-GO**, with one P1 and one P2 inline.

The long-running ownership dispute is mostly resolved correctly. On the shipped raw CONNECT path, the pre-classification `CloseRawLayer` is the component that owns both transports and emits both `CloseConnection` commands. On non-WebSocket HTTP 101, `rawtcp=false` leaves closure to mitmproxy's HTTP layer. I independently exercised both paths on mitmproxy 12.2.3: the client received no origin banner, the origin independently observed peer EOF, and the `raw_tunnel` audit was present. The earlier `TCPFlow.kill()`-as-authority and origin-close-proof findings are therefore closed.

The remaining production blocker is specifically the fragmented TLS transition: returning without assigning a layer is not an undecided state once the built-in addon runs later in the same hook dispatch. The current live test also cannot become green after that is fixed because its client does not trust the per-run proxy CA. Details and reproductions are inline.

Validation at this exact head: fallback and real-mitmproxy unit suites 18/18; CI planner 30/30; `git diff --check` clean; direct real-mitmproxy probes for raw CONNECT, raw 101, and TLS first fragments. Fresh machine state is OPEN / MERGEABLE / BLOCKED. `audit` is COMPLETED/SUCCESS; required `test` is currently IN_PROGRESS on run `32644324917`, so there is no exact-head gate-complete basis for APPROVE in addition to the findings above. This is COMMENT-only; no merge action.

packages/eval/harbor/egress_filter.py
```diff
@@ -197,6 +246,54 @@ def connect_target_url(flow: object) -> str:
...
+    """Close only when the bytes cannot still become HTTP or TLS.
+
+    Script next_layer runs before mitmproxy 12.2.3's classifier. Copying its
+    `probably_no_http` test here would treat `GET` / `GET ` as raw and assign
+    CloseRawLayer while the request line is still arriving. With rawtcp=false
+    the built-in path would have kept waiting for HttpLayer.
+    """
+    data_client = _next_layer_bytes(nextlayer, "data_client")
+    data_server = _next_layer_bytes(nextlayer, "data_server")
+    if _could_start_tls_record(data_client):
```
jackwener, line 270: **[P1][① normal HTTPS path] Returning here does not keep a 1- or 2-byte TLS prefix undecided in production.** Script addons run before mitmproxy 12.2.3's built-in `NextLayer`, but both run in the same `NextLayerHook`. After this function returns with `nextlayer.layer is None`, the built-in addon sees `starts_like_tls_record(...) == false`; with `rawtcp=false` it immediately installs `HttpLayer`. That layer owns subsequent bytes, so this script is not called a second time to reconsider the completed prefix.

I reproduced this against the pinned mitmproxy on the exact head: first fragments of 1 and 2 bytes each time out after 10 seconds, while 3- and 5-byte first fragments immediately enter the TLS path and reach certificate verification in about 0.5 seconds. The unit test hides the lifecycle by manually calling `MODULE.next_layer(fragmented)` a second time; production has already committed to `HttpLayer` before that point. The latest commit changes only the MemoryBIO test driver, not this classifier transition.

Please make the viable TLS prefix claim a real layer before control reaches the built-in addon (for example, install the same TLS stack once `b"\x16"` / `b"\x16\x03"` is observed, or use a buffering layer that genuinely reclassifies when more bytes arrive). Merely returning `False` here cannot express `NeedsMoreData` across addons.

packages/eval/harbor/test_egress_filter_live.py
```diff
@@ -0,0 +1,599 @@
...
+                sock.sendall(payload)
+            except OSError:
+                return header, leftover
+            return header, leftover + cls._recv_until_close(sock)
+
+    @classmethod
+    def fragmented_tls_via_proxy(cls, first_fragment_size: int) -> str:
+        incoming = ssl.MemoryBIO()
+        outgoing = ssl.MemoryBIO()
+        tls = ssl.create_default_context().wrap_bio(
```
jackwener: **[P2][① required live-CI path] This client cannot validate mitmproxy's generated certificate.** `ssl.create_default_context()` trusts the runner's system CAs, not the per-run CA created inside the proxy container. On the exact head, sending 3 or 5 initial ClientHello bytes (enough to reach the normal TLS layer) fails with `SSLCertVerificationError: unable to get local issuer certificate`. The ordinary HTTPS control above already uses `curl --insecure`, so it does not expose this.

Once the fragmented-prefix P1 is fixed, the 1/2-byte cases will advance to the same certificate error and the required check will still fail for a test-harness reason. Load the proxy CA into this context, or deliberately disable verification here as the curl control does; then add a nonfragmented/3-byte control so the driver and CA setup are proven independently of the fragmentation behavior.

packages/eval/harbor/egress_filter.py
1625567290: Fixed at exact head `c02400d25`.

Before changing the code, I reproduced the real addon order with mitmproxy 12.2.3: after this script returned with no layer, the built-in addon assigned `HttpLayer` for both `16` and `1603`; `160301` alone reached `ServerTLSLayer`.

The viable one- and two-byte prefixes now install the same `ServerTLSLayer` → `ClientTLSLayer` stack used by the built-in classifier, so later ClientHello bytes are consumed by a real TLS layer and the built-in addon cannot overwrite it. The complete three-byte prefix remains owned by the built-in classifier.

Validation on the rebased head: a real local mitmdump 12.2.3 process completed `CONNECT example.com:443` handshakes for first-fragment sizes 1, 2, and 3, all negotiating TLS 1.3; the unit suite also passed 18/18 with real mitmproxy imports.

packages/eval/harbor/test_egress_filter_live.py
1625567290: Fixed at exact head `c02400d25`.

The MemoryBIO client now deliberately disables certificate verification for this classification test because the proxy creates a private per-run CA, matching the existing `curl --insecure` control. The live case now includes first-fragment size 3 as the non-fragmented classifier control alongside sizes 1 and 2.

The updated driver completed all three paths through a real mitmdump 12.2.3 process and negotiated TLS 1.3 in each case. The selected Docker-backed CI lane remains the final container-level check on this exact head.

1625567290:
Rebased onto current `main` (`84ed9a310`) and force-updated the branch to exact head `c02400d25`.

Both blocking review findings are fixed:
- fragmented TLS prefixes `16` and `1603` now enter a real `ServerTLSLayer` → `ClientTLSLayer` stack before the built-in addon can classify them as HTTP;
- the live MemoryBIO client deliberately accepts the proxy’s ephemeral CA and includes a 3-byte control beside the 1- and 2-byte cases.

Reproduction before the fix used the pinned mitmproxy 12.2.3 lifecycle and showed `HttpLayer` for 1/2 bytes versus `ServerTLSLayer` at 3 bytes. After the fix, a real local mitmdump process completed `CONNECT example.com:443` for sizes 1/2/3 and negotiated TLS 1.3 in all three cases.

Final validation on CI-matching Node 24.19.0:
- clean `npm ci`
- one complete `npm test` run: all workspaces passed
- full workspace typecheck
- lint and format checks
- ASF header check
- CI/Windows contract tests: 70/70
- egress unit tests with real mitmproxy imports: 18/18
- `git diff --check`

The PR diff contains seven code/test/CI files, with no documentation or Java test files. The Docker-backed proxy suite is selected by CI and remains the final container-level check; the fresh exact-head workflows may require maintainer approval.

packages/eval/harbor/test_egress_filter_live.py
```diff
@@ -0,0 +1,606 @@
...
+                self.assertTrue(
+                    self.fragmented_tls_via_proxy(first_fragment_size).startswith("TLS")
+                )
+
+    def test_connect_to_a_blocklisted_host_is_451(self) -> None:
+        header, _ = self.connect_via_proxy("tbench.ai", 443, b"")
+        self.assertIn(b"451", header.split(b"\r\n", 1)[0])
+        self.assertIn(b"tbench_domain", header)
+
+    def test_raw_connect_relays_no_bytes_and_is_audited(self) -> None:
```
Astro-Han, line 557: **[P3] The live suite only proves the "does not wrongly kill" direction; the "does not wrongly let through" side of the fragment handling is unasserted.**

The cases here cover HTTP/HTTPS forwarding, fragmented handshakes at 1/2/3 bytes completing, blocklist 451, plain raw CONNECT being closed and audited, 101 upgrade auditing, and websocket pass-through. Real TLS and plain raw are both covered — the disguised middle is not.

That matters specifically because `next_layer` now commits to TLS on a 1–2 byte prefix. A raw tunnel opening with `\x16\x03` is still closed today, but only because the fake client then fails mitmproxy's handshake and the built-in teardown runs. Nothing in this file locks that in. If a future mitmproxy version changes its handshake-failure semantics, the behaviour could regress and this suite would stay green.

A variant of `test_raw_connect_relays_no_bytes_and_is_audited` that sends `b"\x16\x03" + garbage` and asserts the same zero-relay and audit outcome would close it.

<details>
<summary>中文</summary>

**[P3] live 测试只证了"不误杀"，分片处理的"不漏放"那一面没有断言。**

现有用例覆盖 HTTP/HTTPS 转发、1/2/3 字节分片握手完成、blocklist 451、plain raw CONNECT 关闭并审计、101 升级审计、websocket 放行——真实 TLS 和 plain raw 两端都有，中间的伪装形态没有。

这一点之所以要紧，是因为 `next_layer` 现在看到 1–2 字节前缀就钉死 TLS 层。以 `\x16\x03` 开头的 raw tunnel 今天仍会被关闭，但**靠的是伪客户端随后握手失败、内建拆连接兜住**，这个文件里没有任何东西把它锁住。将来 mitmproxy 换版本、握手失败语义变化，行为可能退化而这套测试仍然全绿。

照 `test_raw_connect_relays_no_bytes_and_is_audited` 加一个发送 `b"\x16\x03" + 垃圾` 的变体、断言同样的零中继与审计结果即可。
</details>

Astro-Han (approved):
Approving. The ownership argument holds: on `main`, `next_layer` returns immediately unless the layer is already a `TCPLayer`, and the script hook fires before the built-in classifier assigns one — so on the production CONNECT path that branch never runs and the raw tunnel is never closed. Assigning `CloseRawLayer` ahead of the classifier, letting the built-in HTTP layer close 101 upgrades via `rawtcp=false`, and keeping the `isinstance` branch for already-classified paths puts each close where that socket is actually owned, with no second authority.

I checked the parts most likely to hide a leak and found none: `_next_layer_bytes` returning `b""` on error falls through to the built-in classifier rather than skipping the close, `CloseRawLayer` yielding `CloseConnection` twice is harmless, `kill_flow` swallowing exceptions only ever costs an extra attempt, and `apply_http_policy` fails closed with a 503.

The `+995/-10` is not new machinery. Production logic is `egress_filter.py +139/-5`; the bulk is a real Docker/mitmproxy live suite (606 lines) plus unit tests (+215) and the CI wiring to run them.

One behaviour worth stating explicitly, because I went looking for a hole here and it is worth recording that there isn't one. This head commits to TLS on a 1–2 byte fragment prefix and assigns the layers immediately, giving up the chance to reclassify with more data. I tested whether a raw tunnel could exploit that: against the pinned image, four CONNECT payloads — plain, `\x16`, `\x16\x03`, `\x16\x03\x01`, each followed by garbage — were all closed, with zero bytes relayed to the origin and one `raw_tunnel` audit record each. Pinning the TLS layer makes the fake client fail the handshake, and the teardown still passes through the audited path.

One non-blocking gap is inline.

<details>
<summary>中文</summary>

批准。所有权论证成立：`main` 上的 `next_layer` 除非层已经是 `TCPLayer` 否则立即 return，而 script hook 在内建 classifier 赋层之前触发——所以生产 CONNECT 路径上那个分支永不执行，raw tunnel 从不被关闭。改成在 classifier 之前自己赋 `CloseRawLayer`、101 升级交给内建 HTTP 层经 `rawtcp=false` 关闭、`isinstance` 分支保留给已赋层路径，三处各自落在真正拥有该 socket 的那一层，没有第二份权威。

最容易藏漏关的地方我都查了，没有：`_next_layer_bytes` 出错返回 `b""` 时落回内建 classifier 而不是跳过关闭；`CloseRawLayer` 重复 yield `CloseConnection` 无害；`kill_flow` 吞异常最多多试一次；`apply_http_policy` 异常走 503 fail-closed。

`+995/-10` 不是新机制堆叠：生产逻辑只有 `egress_filter.py +139/-5`，主体是 606 行真 Docker/mitmproxy live 测试、+215 单测和跑它们的 CI 接线。

有一处行为值得写下来，因为我专门去找过这个洞、结果是没有：这个 head 看到 1–2 字节的分片 TLS 前缀就当场钉死 TLS 层，放弃了拿到更多数据后重新判定的机会。我实测了 raw tunnel 能否利用这一点——用 pin 镜像对四种 CONNECT payload（plain、`\x16`、`\x16\x03`、`\x16\x03\x01`，各接垃圾字节）实跑，四种全部被关闭，origin 侧零字节中继，各落一条 `raw_tunnel` 审计记录。钉死 TLS 层之后伪客户端无法完成握手，拆连接仍然流经被审计的路径。

一条不阻塞的缺口见行内。
</details>

Astro-Han:
LGTM, merging.

I trial-merged this into the current main (`3f5c4bb08`) and re-ran the affected checks: all 18 tests in `packages/eval/harbor/test_egress_filter.py` pass, all 31 in `scripts/ci-test-plan.test.mjs` pass, and the ASF header check is clean.

<details>
<summary>简体中文</summary>

LGTM，合并中。

我在当前 main（`3f5c4bb08`）上做了一次试合并并重跑了受影响的检查：`packages/eval/harbor/test_egress_filter.py` 18 项全过，`scripts/ci-test-plan.test.mjs` 31 项全过，ASF header 检查通过。

</details>

PR #3021, opened 2026-08-14
fix(cli): start fresh TUI Sessions at the default permission mode

### LINKED ISSUE #3020: fix(cli): TUI /new leaks the previous session's Full access (bypass) into the fresh session
## What happened

In the TUI (`maka`), elevating a session to **Full access** (`bypass`) and then starting a fresh session with `/new` silently carries Full access into the new session:

1. The status line keeps showing **Full access** after `/new`.
2. The next prompt actually creates the new session with `permissionMode: 'bypass'` — tools then run with direct file/network access in a brand-new session the user never elevated.

Expected: a fresh session starts at the default mode (**Auto** / `ask`), and Full access is an explicit per-session opt-in. That is the model every other surface follows:

- TUI cold start hard-codes `permissionMode: 'ask'` (`packages/cli/src/runtime-host-tui-command.ts`, `runtime-host-tui-context.ts`)
- `maka run` defaults to `ask` unless `--yolo` is passed explicitly (`packages/cli/src/run-command-core.ts`)
- Desktop creates new chats with the settings-backed `chatDefaults.permissionMode` (fallback `ask`), never the previously viewed session's elevated mode (`apps/desktop/src/main/create-session-input.ts`)

`startNewSession()` even documents this intent — "A fresh session is not bound by the previous one's boundary" — but it only clears `#activeBoundaryDisplayMode`, not `#permissionMode`, and `#createSession` passes `this.#permissionMode` to `session.create`, so the elevation leaks anyway.

## How to reproduce

1. `maka` in any project; send a prompt to create session A.
2. Open the permission picker and switch to **Full access** (confirm the prompt).
3. Type `/new`.
4. Send any prompt.

Observe: the status line shows Full access after `/new`, and the newly created session runs with full access (no sandbox boundary prompt, direct file/network reach).

Verified against source with the real TUI driver path (`driver.setPermissionMode('bypass')` → `startNewSession()` → `preparePrompt()`): `getPermissionMode()` after `/new` returns `bypass`, and the resulting `session.create` request carries `permissionMode: 'bypass'`.

Root cause: `packages/cli/src/runtime-host-session-driver.ts` — `startNewSession()` resets `#sessionId` and `#activeBoundaryDisplayMode` but leaves `#permissionMode` untouched.

## Environment

- Maka version or commit: 2f76c3109 (main)
- OS and version: Linux
- Surface: TUI
- Node.js version, if running from source: v26.3.0

Reproduction output (script driving the compiled driver the same way the TUI does):

```
session A created with permissionMode: ask
after setPermissionMode(bypass), getPermissionMode(): bypass
after /new (startNewSession), getPermissionMode(): bypass   ← should be ask
session B created with permissionMode: bypass               ← should be ask
```

Suggested fix: have `startNewSession()` reset `#permissionMode` to the driver's construction-time default (what the TUI passed in, `ask` today), so a per-session elevation never leaks into a fresh session. Users who want Full access in the new session can re-enable it explicitly before the first prompt — `setPermissionMode` already supports the no-active-session path for exactly that.

---

*This issue was investigated and prepared with AI assistance (Maka agent); a human contributor reviewed the analysis and reproduction before submission.*

me2seeks:
## Summary

In the TUI, elevating a session to **Full access** (`bypass`) and then running `/new` silently carried the elevation into the fresh session: the status line kept showing Full access, and the next prompt created the new session with `permissionMode: 'bypass'` — tools ran with direct file/network access in a session the user never elevated.

`startNewSession()` cleared the session binding and the boundary *display* mode, but not the base `#permissionMode` that `session.create` uses. The fix captures the construction-time default (`#defaultPermissionMode`, what the TUI passes at cold start — `ask`) and resets to it in `startNewSession()`. This matches the desktop model, where a new chat uses the settings-backed default rather than the previously viewed session's mode, and keeps Full access an explicit per-session opt-in: `setPermissionMode` before the first prompt still raises the fresh session.

Fixes #3020

## Verification

- `packages/cli` suite: **260/260 pass** (includes the new regression test).
- Mutation check: with the reset line removed, the new test fails with `'bypass' !== 'ask'`; restored, it passes.
- End-to-end driver script (reproduces the exact TUI path: `setPermissionMode('bypass')` → `startNewSession()` → `preparePrompt()`): before — `getPermissionMode()` after `/new` returned `bypass` and the new `session.create` carried `bypass`; after — both are `ask`.
- `biome check` clean on both touched files.
- Not run: full monorepo suite and desktop/e2e suites (change is confined to `packages/cli`; left to CI).

## Security

This changes security-relevant behavior (permission/boundary mode of newly created sessions). Per the review policy and the contributor's request, the fast path does **not** apply: **this PR requires independent review by another human before merge** (AI review does not count), in addition to the human contributor of record.

Prepared with AI assistance (Maka agent); the investigation, reproduction, and fix were reviewed by the human contributor. The final commit carries `Generated-by: Maka` — please keep the trailer when squash-merging.

## Checklist

- [ ] No

### DIFF
--- a/packages/cli/src/__tests__/runtime-host-session-driver.test.ts
+++ b/packages/cli/src/__tests__/runtime-host-session-driver.test.ts
@@ -121,6 +121,72 @@ describe('Runtime Host Maka Session driver', () => {
     }
   });
 
+  test('drops a per-session Full access elevation when a fresh Session starts (#3020)', async () => {
+    // The TUI flow behind /new: session A is elevated to bypass, then the
+    // driver is asked to start over. The next prompt lazily creates session B
+    // through preparePrompt. Session B must be created with the
+    // construction-time default — Full access is an explicit per-session
+    // opt-in, never inherited.
+    const connection = new FakeConnection([
+      new FakeSubscription(continuitySnapshot(), Promise.resolve([])),
+      new FakeSubscription(
+        continuitySnapshot({
+          session: {
+            sessionId: 'session-2',
+            metadataRevision: 1,
+            status: 'running',
+            createdAt: 1,
+            lastUsedAt: 1,
+            isArchived: false,
+          },
+          rootTurn: null,
+        }),
+        Promise.resolve([]),
+      ),
+    ]);
+    let nextId = 0;
+    const driver = createRuntimeHostMakaSessionDriver({
+      connection: connection.value,
+      cwd: '/repo',
+      llmConnectionSlug: 'openai-main',
+      model: 'gpt-5',
+      permissionMode: 'ask',
+      newId: () => `session-${++nextId}`,
+    });
+
+    await driver.createSession({
+      cwd: '/repo',
+      llmConnectionSlug: 'openai-main',
+      model: 'gpt-5',
+      backend: 'ai-sdk',
+      permissionMode: 'ask',
+    });
+    connection.executionBoundary = { kind: 'bypass', revision: 2 };
+    await driver.setPermissionMode('bypass');
+    assert.equal(driver.getPermissionMode?.(), 'bypass');
+
+    driver.startNewSession();
+    assert.equal(driver.getPermissionMode?.(), 'ask');
+
+    // The fresh Session's boundary is managed again once it exists.
+    connection.executionBoundary = { kind: 'managed', access: 'writable', revision: 3 };
+    await driver.preparePrompt('hello');
+
+    const creates = connection.requests.filter(({ operation }) => operation === 'session.create');
+    assert.equal(creates.length, 2);
+    assert.deepEqual(creates[1]!.input, {
+      sessionId: 'session-2',
+      workspace: { kind: 'host_path', path: '/repo' },
+      name: 'New Chat',
+      modelTarget: {
+        kind: 'explicit',
+        connectionSlug: 'openai-main',
+        model: 'gpt-5',
+      },
+      permissionMode: 'ask',
+    });
+  });
+
   test('relocates a moved Session through Host authority before attaching', async () => {
     const root = await mkdtemp(join(tmpdir(), 'maka-tui-resume-moved-cwd-'));
     const target = join(root, 'new-worktree');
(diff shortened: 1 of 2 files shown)


Astro-Han (approved):
Thanks — this is a small, well-contained fix at the right lifecycle boundary.

I traced the TUI path from `/new` through `startNewSession()` and the next `preparePrompt()` into the actual `session.create` request. Resetting the driver’s seed permission prevents Full access from leaking across Sessions, while resumed Sessions continue to take their configuration and execution boundary from the Runtime Host. The control lock also prevents `/new` from racing an in-flight permission change.

The regression test exercises the production driver seam and is mutation-sensitive. The focused CLI suite passes 260/260, and CI is green.

One non-blocking suggestion: because the new field is described as the construction-time default rather than specifically `ask`, consider using a non-`ask` value in a small parameterized assertion. That would distinguish the intended implementation from a future hard-coded `this.#permissionMode = 'ask'`. All current TUI production callers pass `ask`, so I do not consider this necessary for merge.

Approved.

<details>
<summary>中文对照</summary>

感谢，这是一项边界清楚、落点正确的小修复。

我核对了从 TUI `/new`、`startNewSession()` 到下一次 `preparePrompt()` 和实际 `session.create` 请求的完整链路。重置 driver 用于创建新 Session 的初始权限，可以阻止 Full access 跨 Session 泄漏；恢复已有 Session 时，配置和 execution boundary 仍然由 Runtime Host 决定。control lock 也避免了 `/new` 与进行中的权限切换发生竞态。

回归测试命中了生产 driver seam，并且删除重置代码后会失败。CLI 聚焦测试 260/260 通过，CI 全绿。

有一个非阻塞的小建议：新增字段表达的是“构造时传入的默认值”，而不只是固定的 `ask`。可以考虑用一个非 `ask` 值补充参数化断言，从而防止未来被简化成硬编码的 `this.#permissionMode = 'ask'`。当前所有 TUI 生产调用点都明确传入 `ask`，所以这不影响本次合并。

批准。

</details>

Disclosure: This review was prepared with assistance from OpenAI Codex and Claude Opus. I reviewed the evidence, verified the production path and focused tests, and made the final judgment.

PR #3025, opened 2026-08-14
feat(cli): surface the autonomous goal in the TUI

me2seeks:
## Summary

An armed goal burns tokens between prompts, but the TUI showed nothing: goal continuation turns rendered as plain user prompts, the status line had no goal segment, and the only way to inspect the loop was spending a model turn on `GoalStatus`. This PR surfaces the autonomous goal in the TUI using the same authoritative push channel the desktop observer already diffs.

- **Status line** shows a live goal (`active`/`waiting`/`paused`) with the iteration counter and elapsed wall-clock (e.g. `goal 3/50 12m`); terminal goals stay hidden, matching the desktop chip.
- **Goal-origin turns** render with a `Goal continuation (autonomous)` provenance header instead of as user prompts, folded from stored messages so replay/reconnect can never duplicate or diverge.
- **`/goal`** prints condition, status, iterations, elapsed, tokens vs budget, and the evaluator's last note without burning a turn — and works while the loop is busy.
- **Transport**: goal state rides the session subscription's continuity snapshot (`GoalProjection`), so every transition (set, evaluator settle, abort auto-pause, waiting wake, terminal verdict, cross-client control) reaches the TUI as it happens. No polling, no stale window, no new RPC.

Ref #3022

## Review process

Two independent read-only review passes (first-principles + Occam's razor) over the diff. Findings addressed in this branch:

- **major** — `/goal` typed during a running turn was steered into the model as literal text; read-only status commands now answer locally mid-turn (the feature's primary use case), with a regression test asserting `steer` is not called.
- **minor** — goal-change listener payload is now defensively cloned on both the canonical-replacement and per-frame paths.
- **nit** — `/goal` summary collapses embedded whitespace in condition/evaluator note, and a cleared goal is labeled `Cleared goal:` instead of presenting its condition as armed.

Deferred (out of scope): TUI goal control (pause/resume/clear) is tracked in #3023; `GoalManager.remove()` not firing the invalidation pipeline is pre-existing runtime behavior only reachable on session retirement.

## Test plan

- `packages/cli`: 272 tests pass (`node --test`), including new coverage:
  - `pi-goal.test.ts` — status-line text, elapsed formatting, exhaustive status labels, summary whitespace/cleared handling
  - `pi-transcript.test.ts` — status-line goal segment rendering + goal-continuation provenance header
  - `pi-tui-runner.test.ts` — `/goal` summary end-to-end, `/goal` mid-turn local answer (no steer), host-pushed pause reaching the status line
[...]

PR #3026, opened 2026-08-15
feat(cli): /goal pause|resume|clear control from the TUI

### LINKED ISSUE #3023: feat(cli): /goal pause|resume|clear — user control of autonomous goals without a model turn
## Problem

Today the only way to pause, resume, or clear an autonomous goal from the TUI is to ask the model in natural language to call GoalPause / GoalResume / GoalClear. That costs a model turn and can hit the goal-control decline rules in `goal-tools.ts`.

Worse, the pause semantics are invisible:

- **Ctrl+C during a goal continuation turn aborts the turn, and the runtime auto-pauses the goal** (`goal-continuation.ts`: any non-completed outcome settles via `pauseAtCheckpoint`, reason "Goal-associated turn was aborted."). The TUI prints nothing about this — the user cannot tell the goal still exists, that it is paused rather than stopped, or how to resume it.
- **Quitting the TUI does not stop a goal.** The goal authority is durable; on the next attach, `recoverActiveGoal` auto-continues active/waiting goals with no notice.

The host protocol already implements `goal.control` with actions `pause | resume | clear` and optimistic `expectedRevision` (`packages/runtime-host/src/protocol/goal.ts`); the desktop client already wires `clear` through it. `resumeFromControl` restarts a paused goal without a model-owned turn.

## Proposal

1. Add `/goal pause`, `/goal resume`, `/goal clear` to the TUI, dispatched through the driver's goal control (with `operation_conflict` retry, mirroring the desktop client's `clearGoal`).
2. When a turn abort causes the goal to become paused, print a one-line notice: goal paused, why, and that `/goal resume` continues it / `/goal clear` stops it.
3. When attaching to a session whose goal auto-continues after recovery, print the same style of notice so a token-burning loop never resumes silently.

Depends on the goal query/indicator slice (#3022). Codex parallel: `/goal [<objective>|clear|edit|pause|resume]` plus a "Resume paused goal?" prompt when submitting while paused.

Generated-by: Maka

me2seeks:
## Summary

The only way to steer an autonomous goal from the TUI was asking the model to call `GoalPause`/`GoalResume`/`GoalClear` — a burned model turn that can hit the goal-control decline rules in `goal-tools.ts`. And the pause semantics were invisible: Ctrl+C on a goal continuation turn auto-pauses the durable goal, and attaching to a session with a live goal auto-continues it, both without a word.

- **`/goal pause|resume|clear`** dispatch through the driver's new `controlGoal` (host `goal.control` with `expectedRevision` optimistic retry, mirroring the desktop client's `clearGoal`). Writes take the `runControl` gate; typed mid-turn they refuse with a clear message instead of steering into the model or being silently swallowed.
- **Auto-pause notice**: a goal that transitions into `paused` while attached (typically the abort auto-pause after Ctrl+C) prints a one-line notice naming the reason and the `/goal resume` | `/goal clear` controls. A `/goal pause` we initiated prints its own confirmation instead of a duplicate notice.
- **Attach notice**: attaching to a session whose goal is active/waiting announces the running loop — a token-burning loop never resumes silently.
- **Pre-validation**: invalid transitions (pause from paused, resume from active, clear of a terminal record) get a plain message instead of the host's operation error.

Closes #3023. Stacked on #3025 (goal visibility) — merge that first; this branch will be rebased onto main.

## Review process

Two independent read-only review passes (first-principles + Occam's razor). Findings addressed:

- **major** — the host folds *invalid transitions* into `operation_conflict` alongside revision races; the retry loop burned 3 round-trips on a provably-futile status refusal, then threw a misleading "revision conflict after 3 attempts" that discarded the host's reason. Now: a conflict at an unchanged revision rethrows the host's message, and pause pre-validation matches the host rules exactly (`isLiveGoalStatus` wrongly included `paused`; `clear` had no check).
- **nit** — whitespace collapsing deduped into one `inlineGoalText` helper.

Judged acceptable (documented, not fixed): the sub-frame staleness window between a successful control RPC and the pushed projection (self-heals on the next frame; `expectedRevision` prevents double-apply); a foreign controller pausing inside our own pause window has its notice suppressed (our confirmation still covers the state).

## Test plan

- `packages/cli`: 278 tests pass, including new coverage:
  - driver: clean control with snapshot `expectedRevision`, conflict retry sequence `[1,2,3]`, goal-disappeared-mid-flight → null, status conflict rethrows host reason without futile retries
  - runner: mid-turn `/goal pause` refuses with a message and never steers; pause/resume/clear confirmations; self-initiated pause suppression vs host-pushed pause notice; pre-validation for all three actions; no-goal handling
  - `pi-goal`: pause/attach notice text, whitespace collapse, 120-char cap
- `biome check` clean; `tsc --noEmit` clean for `@maka/cli`.

🤖 Generated by Maka

## Visual evidence

The TUI handles pause, resume, and clear locally while the status line follows the Host-owned goal projection.

Goal pause, resume, and clear controls in the TUI\n\n## AI use\n- [ ] No generative tool was used for implementation.\n- [x] Generative tooling was used and the result was reviewed and verified by the author.\n\nTool(s) and scope: OpenAI Codex (Maka) assisted with implementation, tests, review remediation, and PR documentation.\n\nFinal squash trailer: `Generated-by: Maka`

### DIFF
--- a/packages/cli/src/pi-goal.ts
+++ b/packages/cli/src/pi-goal.ts
@@ -94,11 +94,37 @@ export function goalStatusLineText(
   return `goal ${goalStatusLabel(goal.status)} ${counter}`;
 }
 
+/** Conditions and evaluator notes may legally embed newlines; collapse whitespace so notices stay one line per field. */
+function inlineGoalText(value: string): string {
+  return value.replace(/\s+/g, ' ').trim();
+}
+
+/**
+ * One-line notice when a goal transitions into paused while attached —
+ * typically the runtime's abort auto-pause after Ctrl+C on a goal
+ * continuation turn. Says the loop still exists and how to continue/stop it.
+ */
+export function goalPausedNoticeText(
+  goal: Pick<GoalProjection, 'iterations' | 'maxIterations' | 'lastReason'>,
+): string {
+  const reason = goal.lastReason ? ` ${inlineGoalText(goal.lastReason)}` : '';
+  return `Goal paused (${goal.iterations}/${goal.maxIterations}).${reason} /goal resume continues it, /goal clear stops it.`;
+}
+
+/**
+ * One-line notice when attaching to a session whose durable goal is
+ * auto-continuing after recovery — a token-burning loop never resumes silently.
+ */
+export function goalAttachedNoticeText(
+  goal: Pick<GoalProjection, 'condition' | 'iterations' | 'maxIterations'>,
+): string {
+  const condition = inlineGoalText(goal.condition);
+  const short = condition.length > 120 ? `${condition.slice(0, 119)}…` : condition;
+  return `Autonomous goal is running (${goal.iterations}/${goal.maxIterations}): ${short} — /goal shows details, /goal pause pauses it.`;
+}
+
 /** Full `/goal` summary. Terminal goals are as welcome here as live ones. */
 export function goalSummaryLines(goal: GoalProjection, now: number): string[] {
-  // Conditions and evaluator notes may legally embed newlines; collapse
-  // whitespace so each field stays on one notice line.
-  const inline = (value: string): string => value.replace(/\s+/g, ' ').trim();
   const status = `Status: ${goalStatusLabel(goal.status)} · ${goal.iterations}/${goal.maxIterations} iterations`;
   // Terminal verdicts other than `achieved` carry no freeze timestamp, so a
   // wall-clock elapsed would keep growing for a loop that already ended.
@@ -108,8 +134,8 @@ export function goalSummaryLines(goal: GoalProjection, now: number): string[] {
     // A cleared goal keeps its terminal record, so say "cleared" up front
     // instead of presenting the condition as if it were still armed.
     goal.status === 'cleared'
-      ? `Cleared goal: ${inline(goal.condition)}`
-      : `Goal: ${inline(goal.condition)}`,
+      ? `Cleared goal: ${inlineGoalText(goal.condition)}`
+      : `Goal: ${inlineGoalText(goal.condition)}`,
     elapsedMeaningful ? `${status} · ${formatGoalElapsed(goalElapsedMs(goal, now))}` : status,
   ];
   if (goal.tokenBudget !== null) {
@@ -119,6 +145,6 @@ export function goalSummaryLines(goal: GoalProjection, now: number): string[] {
   } else if (goal.tokensSpent > 0) {
     lines.push(`Tokens: ${formatTokenCount(goal.tokensSpent)}`);
   }
-  if (goal.lastReason) lines.push(`Last evaluator note: ${inline(goal.lastReason)}`);
+  if (goal.lastReason) lines.push(`Last evaluator note: ${inlineGoalText(goal.lastReason)}`);
   return lines;
 }
--- a/packages/cli/src/pi-goal.ts
+++ b/packages/cli/src/pi-goal.ts
@@ -9,6 +9,7 @@
 import type { GoalStatus } from '@maka/core/goal';
 import type { GoalProjection } from '@maka/runtime-host/protocol';
 import { formatTokenCount } from './pi-transcript-format.js';
+import { stripAnsi } from './tui-ansi.js';
 
 /**
  * Statuses a watching user still cares about. Terminal goals are hidden from
@@ -96,7 +97,10 @@ export function goalStatusLineText(
 
 /** Conditions and evaluator notes may legally embed newlines; collapse whitespace so notices stay one line per field. */
 function inlineGoalText(value: string): string {
-  return value.replace(/\s+/g, ' ').trim();
+  return stripAnsi(value)
+    .replace(/[\u0000-\u001f\u007f-\u009f]/gu, ' ')
+    .replace(/\s+/g, ' ')
+    .trim();
 }
 
 /**
--- a/packages/cli/src/pi-tui-runner.ts
+++ b/packages/cli/src/pi-tui-runner.ts
@@ -117,8 +117,15 @@ import {
   type MakaSlashCommand,
 } from './pi-tui-pickers.js';
 import { formatMakaResumeCommand } from './cli-invocation.js';
-import { goalSummaryLines } from './pi-goal.js';
+import {
+  goalAttachedNoticeText,
+  goalPausedNoticeText,
+  goalStatusLabel,
+  goalSummaryLines,
+  isLiveGoalStatus,
+} from './pi-goal.js';
 import { getTuiPrimaryGuidance } from './tui-primary-guidance.js';
+import type { GoalControlAction, GoalProjection } from '@maka/runtime-host/protocol';
 
 export interface MakaPiTuiInput {
   /** Launcher command used in resume and recovery instructions. */
@@ -343,13 +350,48 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
     rejectClosed = reject;
   });
 
-  // The driver is the single read authority for goal state: metadata() reads
-  // the live projection on every render, and the subscription exists only to
-  // re-render when a host-pushed transition lands. No cached copy, so the
-  // status line and `/goal` can never drift from the driver's snapshot.
-  const unsubscribeGoalChanges = input.driver.subscribeGoalChanges?.(() => {
+  // Rendering reads the driver's live projection directly (metadata()); this
+  // cache exists only to detect transitions on the push stream — notably the
+  // abort auto-pause — and to suppress the notice for a pause we initiated.
+  let currentGoal: GoalProjection | null = input.driver.getGoal?.() ?? null;
+  // goalId of a `/goal pause` we initiated: its paused projection must not
+  // re-announce itself — the command prints its own confirmation.
+  let selfInitiatedPauseGoalId: string | null = null;
+  const unsubscribeGoalChanges = input.driver.subscribeGoalChanges?.((goal) => {
+    const previous = currentGoal;
+    currentGoal = goal;
+    if (
+      goal !== null &&
+      goal.status === 'paused' &&
+      previous?.goalId === goal.goalId &&
+      previous.status !== 'paused'
+    ) {
+      if (selfInitiatedPauseGoalId === goal.goalId) {
+        selfInitiatedPauseGoalId = null;
+      } else {
+        // Typically the runtime's abort auto-pause (Ctrl+C on a goal
+        // continuation turn): the loop still exists and can be resumed.
+        state.entries.push({
+          kind: 'notice',
+          level: 'info',
+          text: goalPausedNoticeText(goal),
+        });
+      }
+    }
     requestRender();
   });
+  // Attaching to a session whose durable goal auto-continues after recovery
+  // must never resume a token-burning loop silently.
+  if (
+    currentGoal !== null &&
+    (currentGoal.status === 'active' || currentGoal.status === 'waiting')
+  ) {
+    state.entries.push({
+      kind: 'notice',
+      level: 'info',
+      text: goalAttachedNoticeText(currentGoal),
+    });
+  }
 
   const metadata = (): MakaPiTranscriptMetadata => ({
     title: input.title,
@@ -2494,6 +2536,59 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
     requestRender();
   };
 
+  const controlGoalCommand = async (action: GoalControlAction): Promise<void> => {
+    const notice = (text: string): void => {
+      state.entries.push({ kind: 'notice', level: 'info', text });
+      requestRender();
+    };
+    const goal = input.driver.getGoal?.() ?? null;
+    if (!goal) {
+      notice('No goal set.');
+      return;
+    }
+    if (!input.driver.controlGoal) {
+      notice('Goal control is unavailable on this runtime.');
+      return;
+    }
+    // Pre-validate against the live projection so an invalid transition gets
+    // a plain message instead of the host's operation error. These mirror the
+    // host's rules exactly: pause requires active|waiting, resume requires
+    // paused, clear rejects a terminal record.
+    if (action === 'pause' && goal.status !== 'active' && goal.status !== 'waiting') {
+      notice(`Cannot pause: the goal is ${goalStatusLabel(goal.status)}.`);
+      return;
+    }
+    if (action === 'resume' && goal.status !== 'paused') {
+      notice(`Cannot resume: the goal is ${goalStatusLabel(goal.status)}.`);
+      return;
+    }
+    if (action === 'clear' && !isLiveGoalStatus(goal.status)) {
+      notice(`Cannot clear: the goal is ${goalStatusLabel(goal.status)}.`);
+      return;
+    }
+    if (action === 'pause') selfInitiatedPauseGoalId = goal.goalId;
+    let result: GoalProjection | null;
+    try {
+      result = await input.driver.controlGoal(action);
+    } catch (error) {
+      selfInitiatedPauseGoalId = null;
+      throw error; // runControl's reportError surfaces it
+    }
+    if (result === null) {
+      // The goal disappeared to a concurrent controller mid-flight.
+      selfInitiatedPauseGoalId = null;
+      notice(action === 'clear' ? 'Goal cleared.' : 'The goal no longer exists.');
+      return;
+    }
+    if (action === 'pause') {
+      notice('Goal paused. /goal resume continues it, /goal clear stops it.');
+    } else if (action === 'resume') {
+      notice('Goal resumed.');
+    } else {
+      notice('Goal cleared.');
+    }
+  };
+
   const slashCommandHandlers = {
     context: {
       description: primaryGuidance.commands.context,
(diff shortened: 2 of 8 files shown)


packages/cli/src/pi-tui-runner.ts
```diff
@@ -2391,6 +2453,80 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
...
+      notice(action === 'clear' ? 'Goal cleared.' : 'The goal no longer exists.');
+      return;
+    }
+    if (action === 'pause') {
+      notice('Goal paused. /goal resume continues it, /goal clear stops it.');
+    } else if (action === 'resume') {
+      notice('Goal resumed.');
+    } else {
+      notice('Goal cleared.');
+    }
```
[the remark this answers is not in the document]
me2seeks: Fixed in b810dbd8c: the success path now clears selfInitiatedPauseGoalId once the command settles and syncs the transition cache to the authoritative controlGoal response, so the trailing push folds onto an identical previous state (no duplicate notice) and a later host-initiated pause of the same goal announces itself. A regression test with a deferred-push driver fails without the fix and passes with it.

packages/cli/src/pi-tui-runner.ts
```diff
@@ -329,6 +337,49 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
...
+      }
+    }
+    requestRender();
+  });
+  // Attaching to a session whose durable goal auto-continues after recovery
+  // must never resume a token-burning loop silently.
+  if (
+    currentGoal !== null &&
+    (currentGoal.status === 'active' || currentGoal.status === 'waiting')
+  ) {
```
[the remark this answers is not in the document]
me2seeks: Fixed in b810dbd8c per the suggested seam: switchSession now re-reads driver.getGoal() after applySwitchResult (which also re-syncs the goal transition cache to the adopted session) and emits goalAttachedNoticeText when the goal is active/waiting. Emitting after the transcript replacement avoids the adoption-time erasure; the goal subscription remains pause-only. The init-time check stays for drivers that are already attached at startup, and its comment now documents the split. Regression test: resumeSessionId into a session with an active goal shows the notice (fails without the fix).

Astro-Han:
Thanks for the rework — the design is genuinely minimal: goal state stays in runtime-host's `GoalManager` (single authority), the CLI only projects/subscribes/forwards via the existing `goal.control` protocol with `expectedRevision` optimistic concurrency (same shape as desktop's `clearGoal`), and the revision-bump-on-every-transition invariant makes "re-query shows unchanged revision ⇒ genuine state rejection, not a race" a sound check — I verified the client-side pre-checks against the host rules (pause needs active|waiting, resume needs paused, clear refuses terminal) and they match. No new protocol, no polling, no parallel authority. The output is well-tested (driver revision sequences, conflict retries, goal disappearance, busy-gate, notice suppression, pre-checks). CI 18/18 green.

**Conclusion: PASS — no P0/P1/P2.**

P3 (optional): the last retry attempt still drops the host's reason, contradicting the PR's own fix claim ("a conflict at an unchanged revision rethrows the host's message") — `if (attempt === GOAL_CONTROL_MAX_ATTEMPTS - 1) break;` skips the re-query and lands on the generic "revision conflict after 3 attempts"; a client that races twice then gets a state rejection (e.g. another client paused then cleared, your pause arrives at `cleared`) shows the misleading generic message. One-line fix: throw the `conflict` (it's guaranteed non-null on the last pass) or move the re-query before the break, plus a test for "two races then a state rejection" (existing tests only cover the same-revision rejection). Also: `controlGoal` reuses the captured `sessionId` across retries with no session-generation guard (other async paths have `#assertCurrentSession`; the TUI's serial `busy` lock makes this near-theoretical, but a generation check per RPC would be cheap); resuming into an already-paused goal gives no notice (attach notice only covers active/waiting; the subscription only announces paused conversions when `previous?.goalId === goal.goalId`) — users must spot the yellow status-bar segment; mid-turn routing matches the literal `/goal` instead of the catalog-resolved command name (an alias added later would steer into the model); the "token formatting is the shared status-line formatter" test in pi-goal.test.ts:128-130 duplicates the test at :89 and belongs to pi-transcript-format — deletable; three copies of the goal-control retry loop now exist (desktop `clearGoal`, #3027's `controlGoalWithRetry`, this PR's `controlGoal` — #3027 is still OPEN, not merged) — noted as "mirroring" in the PR, acceptable but worth extracting a shared controller later; `formatGoalElapsed(24h)` renders `1d 0h` (could omit `0h` on whole days).

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagent on `opencode-go/deepseek-v4-flash`), which traced the retry/re-query logic and compared the client pre-checks against the host rules from the PR head. P3 items are code-path observations; no P0-P2 were found. Please weigh these findings with your own judgment.*

<details>

结论：PASS（无 P0/P1/P2）。设计真最小：goal 状态仍由 runtime-host 的 GoalManager 持有（单一权威），CLI 只做投影读取/订阅/控制转发，经既有 goal.control 协议 + expectedRevision 乐观并发（与 desktop clearGoal 同构）；"re-query 后 revision 未变 ⇒ 状态拒绝而非竞态"的判定成立（host 每次接受转换必 bump revision，已逐一核对客户端预校验与 host 规则：pause 需 active|waiting、resume 需 paused、clear 拒绝 terminal）；无新协议/无轮询/无并行 authority。测试扎实（driver revision 序列/冲突重试/goal 消失/busy 门禁/notice 抑制/预校验）。CI 18/18 绿。P3（可选）：末次重试仍丢 host 原因与 PR 自述修复矛盾——attempt===MAX-1 时 break 跳过 re-query 落到通用"revision conflict after 3 attempts"；两次竞态后遇状态拒绝（另一客户端先 pause 再 clear、本端 pause 到达 cleared）显示误导消息。一行修复（throw conflict 或 re-query 移到 break 前）+ 补"两次竞态后状态拒绝"用例。另：controlGoal 跨重试复用捕获 sessionId 无会话代际护栏（其他异步路径有 #assertCurrentSession；TUI busy 串行锁使实际窗口近乎理论，每 RPC 比对 generation 成本低）；resume 到已 paused 的 goal 无 notice（attach notice 只覆盖 active/waiting，订阅仅 previous?.goalId 相同时宣布暂停转换）——用户只能靠状态栏黄色段；mid-turn 路由匹配字面 /goal 而非 catalog 解析的命令名（将来加 alias 会 steering 进模型）；pi-goal.test.ts:128-130 重复 :89 的表单测试属于 pi-transcript-format 可删；goal 控制重试循环已有三份（desktop clearGoal、#3027 的 controlGoalWithRetry（#3027 仍 OPEN 未合并）、本 PR controlGoal）——PR 已注明 mirroring 可接受但值得日后抽共享控制器；formatGoalElapsed(24h) 输出 1d 0h（整日可省 0h）。
</details>

packages/cli/src/pi-tui-runner.ts
```diff
@@ -114,6 +114,14 @@ import {
   type MakaSlashCommand,
 } from './pi-tui-pickers.js';
 import { formatMakaResumeCommand } from './cli-invocation.js';
+import {
```
Astro-Han, line 120: **P1 — This Goal/TUI integration conflicts with current `main`'s locale and primary-guidance changes.** `git merge-tree` reproduces content conflicts here and in `pi-transcript.ts` plus their tests; taking either side mechanically can silently drop Goal controls/notices or the current locale behavior. Rebase onto latest `main`, deliberately compose both contracts, and rerun focused CLI/typecheck/CI before approval.

Astro-Han (commented):
The Goal control design itself is coherent: Runtime Host remains authoritative, mutations carry `goalId + expectedRevision`, concurrent conflicts are retried, and the TUI projections/notices cover pause, resume, clear, attach, and auto-pause. Previously reported behavioral issues appear fixed.

The current head is nevertheless not reviewable as a merge result: it conflicts with latest `main` in the exact TUI/transcript files where `main` introduced locale and primary-guidance behavior. The simplest path is a clean rebase that preserves both contracts, followed by focused CLI/typecheck/CI validation; the existing green checks only validate the old base.

Review performed with two Codex reviewer agents and DeepSeek V4 Flash as advisory tools; I reproduced the conflicts against current `main`.

Goal control 设计本身闭环：Runtime Host 保持权威，mutation 携带 `goalId + expectedRevision`，并发冲突会重试，TUI projection/notice 覆盖 pause、resume、clear、attach 和 auto-pause。此前公开的行为问题已修复。

但当前 head 不能作为最终 merge result 审查：它与最新 `main` 在同一组 TUI/transcript 文件中冲突，而 `main` 已在这些文件引入 locale 与 primary-guidance 行为。最简单的路径是 clean rebase，同时保留两边契约，再运行 focused CLI/typecheck/CI；现有绿色检查只验证旧 base。

本次审查使用了两位 Codex reviewer agents 与 DeepSeek V4 Flash 作为辅助工具；我已对当前 `main` 复现冲突。

</details>

packages/cli/src/pi-tui-runner.ts
me2seeks: Fixed in 4c20d9f7a. The stack is rebased onto current main and deliberately composes Goal controls/notices with the locale and primary-guidance catalog; no side was taken mechanically. CLI passes 296/296 after the restack, and the localized guidance contract covers both locales.

packages/cli/src/pi-goal.ts
```diff
@@ -0,0 +1,150 @@
...
+  const counter = `${goal.iterations}/${goal.maxIterations}`;
+  if (goal.status === 'active') {
+    return `goal ${counter} ${formatGoalElapsed(goalElapsedMs(goal, now))}`;
+  }
+  return `goal ${goalStatusLabel(goal.status)} ${counter}`;
+}
+
+/** Conditions and evaluator notes may legally embed newlines; collapse whitespace so notices stay one line per field. */
+function inlineGoalText(value: string): string {
+  return value.replace(/\s+/g, ' ').trim();
```
Astro-Han: [P2] Strip ANSI escape sequences and C0/C1 controls before returning this text. `condition` is user-controlled and `lastReason` can be model-generated; both now flow through this helper into `renderNotice`/`renderIndented`, which preserve ANSI. A value containing `\x1b[2J` therefore clears the terminal when `/goal` renders or the Session reattaches, and other sequences can spoof the UI. Sanitize once here before whitespace collapsing, then add focused cases for ESC, OSC, and ordinary multiline Unicode text.

packages/cli/src/runtime-host-session-driver.ts
```diff
@@ -620,6 +630,62 @@ class RuntimeHostMakaSessionDriverImpl implements RuntimeHostMakaSessionDriver {
...
+          expectedRevision: goal.revision,
+          action,
+        });
+        return result.goal;
+      } catch (error) {
+        if (!(error instanceof RuntimeHostOperationError) || error.code !== 'operation_conflict') {
+          throw error;
+        }
+        conflict = error;
+        if (attempt === GOAL_CONTROL_MAX_ATTEMPTS - 1) break; // a re-query would have no retry to serve
```
Astro-Han: [P3] Preserve the final Host conflict reason instead of replacing it with the generic exhaustion error. If the first two attempts race and the third conflict is an unchanged-revision status refusal, this `break` skips the query that distinguishes that case, so the user sees `revision conflict after 3 attempts` rather than the actionable Host message. On the final attempt, either throw the captured conflict directly or perform the same query/classification before deciding which error to surface.

Astro-Han (commented):
The control flow is otherwise well-shaped: it reuses the Host goal authority, applies optimistic concurrency with bounded retries, and keeps TUI command handling local. One display-boundary issue should be addressed before treating this as merge-ready: goal conditions and evaluator reasons are durable user/model-controlled text and are now copied into terminal notices without control-sequence sanitization. The natural owner is the shared goal-to-terminal formatter, so one sanitization step can cover summaries, attach notices, and pause notices without parallel fixes.

AI-assisted review disclosure: Codex verified the final diff, goal-control retry semantics, Session switching, notice rendering, terminal formatting, focused tests, and live CI. Two independent reviewer-agent passes and an OpenCode Go DeepSeek V4 Flash (high) adversarial pass were used as inputs. No local tests were run.

<details>
<summary>中文复核</summary>

控制流程整体合理：复用 Host goal 权威，以有界重试实现 optimistic concurrency，并保持 TUI 命令本地处理。合并前应修复一处展示边界：goal condition 与 evaluator reason 都是持久化的用户/模型可控文本，现在会未经控制序列清理直接进入终端 notice。最自然的 owner 是共享的 goal-to-terminal formatter，一处清理即可覆盖 summary、attach notice 与 pause notice。

本次为 AI 辅助审查：Codex 核验最终 diff、goal-control 重试、Session 切换、notice rendering、terminal formatter、聚焦测试与实时 CI；另使用两次独立 reviewer 及一次 OpenCode Go DeepSeek V4 Flash（high）对抗审查。未运行本地测试。
</details>

packages/cli/src/pi-goal.ts
me2seeks: Fixed in 29d37734e. Goal condition and evaluator notes now reuse the shared ANSI stripper, expanded to cover CSI, OSC, and DCS forms, then replace C0/C1 controls before whitespace collapse. Focused cases cover CSI, OSC, NUL, multiline text, and ordinary Unicode. CLI passes 296/296.

packages/cli/src/runtime-host-session-driver.ts
me2seeks: Fixed in 29d37734e. Every conflict attempt now performs the same authoritative snapshot query; an unchanged-revision final conflict rethrows the exact actionable Host reason instead of falling through to a generic exhaustion message. The three-conflict regression passes in the 296/296 CLI suite.

Astro-Han (approved):
The current head resolves the substantive issues from the earlier pass: goal text is sanitized consistently, the final host conflict reason is preserved, and resume/pause lifecycle behavior now has focused regression coverage. The command remains owned by the Runtime Host path rather than introducing a second goal authority.

No remaining P0-P2 findings on this head. The existing bot notes about future aliases and a theoretical cross-session race are reasonable non-blocking follow-ups, but neither warrants expanding this PR.

AI-assisted review disclosure: Codex reviewed exact head `29d3773`, inspected the affected goal-control paths and current CI, and found all checks green with no unresolved review threads.

<details>
<summary>中文说明</summary>

当前 head 已修复上一轮的实质问题：goal 文本清理一致、最后一次 Host conflict 原因会保留，resume/pause 生命周期也有针对性回归测试。没有剩余 P0-P2。现有 bot 提到的 alias 扩展和极端跨 session 竞态可作为非阻塞后续，不需要扩大本 PR。

</details>

Astro-Han:
One little thing before merging, A screenshot would be great~

me2seeks:
Added a screenshot to the PR body showing /goal pause, resume, and clear together with the Host-projected status-line changes.

Astro-Han (dismissed):

Astro-Han (approved):
Approved on exact rebased head `32dbabf8c2c030a4ca267d6d049d5df444c46ab2`.

The stack now cleanly contains only the #3026 Goal-control intent on top of merged #3025. The rebase preserves `primaryGuidance.commands.goal` as the single localization authority; the obsolete duplicate localization patch dropped cleanly.

Focused validation passed locally: CLI tests 310/310, CLI typecheck, Biome on every changed file, and `git diff --check`. No remaining P0-P2 findings.

AI-assisted review disclosure: Codex performed the controlled rebase and verified the exact delta and focused checks. Astro-Han authorized the rebase, approval, and merge.

<details>
<summary>中文说明</summary>

已批准精确的 rebase 后 head。当前分支仅保留 #3026 的 Goal control 意图，并正确叠在已合并的 #3025 之上；rebase 保留了 `primaryGuidance.commands.goal` 作为唯一文案权威，重复的 localization 补丁已自然消失。

本地聚焦验证全部通过：CLI 测试 310/310、CLI typecheck、全部变更文件的 Biome 检查及 `git diff --check`。没有剩余 P0-P2。

</details>

packages/cli/src/tui-ansi.ts
```diff
@@ -32,7 +32,10 @@ export function disc(tone: DiscTone): string {
 }
 
 export function stripAnsi(text: string): string {
-  return text.replace(/\x1b\[[0-9;?]*[ -/]*[@-~]/g, '');
+  return text.replace(
+    /\x1b(?:\[[0-?]*[ -/]*[@-~]|\][^\x07\x1b]*(?:\x07|\x1b\\)|[PX^_][^\x1b]*(?:\x1b\\)|[@-_])/gu,
+    '',
+  );
```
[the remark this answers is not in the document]
me2seeks: Confirmed and fixed in 19b7bf3bc. The existing stripAnsi matcher now accepts generic ESC intermediate bytes before the final byte, so the valid charset-selection sequence `\\x1b(0` is removed. The exact regression asserts `stripAnsi("\\x1b(0goal") === "goal"`; CLI build, focused 11/11 tests, typecheck, Biome, and diff checks pass locally.

Astro-Han (approved):
Approved on exact head `19b7bf3bceb4e6b509585155302beef3e2567eb2`.

The follow-up delta closes the remaining terminal-sanitization gap for generic ESC sequences such as character-set selection without adding another display path. The shared `stripAnsi` owner remains the right seam, and the focused regression test covers the concrete sequence.

Focused validation passed locally: CLI tests 311/311, CLI typecheck, Biome on the changed files, and `git diff --check`. There are no unresolved review threads and no remaining P0-P2 findings.

AI-assisted review disclosure: Codex reviewed the exact new delta and reran the focused workspace validation. Astro-Han authorized the approval and merge.

<details>
<summary>中文说明</summary>

已批准精确 head `19b7bf3`。这次补丁补齐了 generic ESC sequence（例如 character-set selection）的终端清理边界，并继续由共享 `stripAnsi` 统一负责，没有新增平行展示路径。回归测试覆盖了具体序列。

本地聚焦验证全部通过：CLI 测试 311/311、CLI typecheck、变更文件的 Biome 检查和 `git diff --check`。当前无未解决审查线程，也无剩余 P0-P2。

</details>

Issue #2560, opened 2026-08-09
product(desktop): capture deferred work in a project-aware Work Board


liugddx:
## English

Parent roadmap: #2469
Related architecture: #2290

## Problem

Useful follow-up work is frequently discovered while an Agent is executing a different task:

- a bug is noticed but should not interrupt the current objective;
- a refactor or product idea deserves to be preserved for later;
- a side conversation identifies a concrete action item;
- the user wants to remember work without starting another Session immediately.

Today this context usually remains buried in conversation history, is copied into an external tracker, or distracts the active task. Maka lacks a lightweight bridge between “we noticed this” and “start an auditable Agent task later.”

## Product direction

Add a local-first **Work Board** that can hold deferred work in either:

- a project-scoped board; or
- a global, unassigned Inbox.

A board item is a user-owned work intention, not an Agent execution claim. It may originate from a main conversation, a side conversation, or manual entry. Later, the user can start it as a normal Maka task/Session while preserving its source context.

The intended loop is:

```text
conversation or side chat
  -> capture deferred work
  -> project board / global Inbox
  -> start as a new task
  -> Session execution
  -> link result and evidence back to the item
```

## Important boundary

Do **not** overload the Session Task Ledger from #2290.

- Session Task Ledger: the model-owned, bounded execution plan inside one Session.
- Work Board: user-owned, cross-Session deferred work and task entry points.

A board item may launch and link to a Session, but it must not be injected into every model turn or silently become part of the active plan. Runtime execution state remains authoritative in existing Session/AgentRun/TaskRun facts.

### Board scopes

- **Inbox**: project-independent ideas and unclassified work.
- **Current project**: work associated with the selected Maka project.
- Items can be moved between Inbox and projects.

### Minimal states

- Inbox
- Todo
- In progress
- Done

“Waiting”, “blocked”, and “recoverable” should come from a linked active Session projection rather than being manually invented board states.

### Capture paths

1. Manual “Add item” from the board.
2. From a main-conversation message or selection.
3. From side chat, for example: “Do not handle this now; add it to the current project todo.”
4. The Agent may propose a board item, but creation requires an explicit user action or unambiguous user instruction. No silent backlog generation.

### Start-task path

- “Start task” creates a normal Maka Session in the selected project/workspace.
- The initial task context contains a concise item summary and typed source references, not a blind dump of the full conversation.
- The board item links to the Session and reflects its projected execution state.
[...]

Issue #2290, opened 2026-08-05
RFC: Radically slim the Session Task Ledger


Astro-Han:
# RFC: Radically slim the Session Task Ledger

Status: Draft. Scope: task-ledger in `core` / `storage` / `runtime-host` / `runtime` + desktop panel + `docs/session-task-ledger-lifecycle.md`.

## TL;DR

The ledger exists for one reason: **the model's work plan lives in a context that is volatile, private, bounded, and expensive — so it must live outside the conversation, in a persistent, authority-owned, session-scoped state re-injected each turn at bounded cost.** That goal needs only three things: a persisted task list, `task_create`/`task_update`, and turn-tail injection. In one month the feature grew from 5 fields / 2 tools / one JSON file to 13 fields / 4 tools / event-sourced dual tables + a paginated query protocol. Proposal: **cut every element whose removal loses nothing irreversible** (code is in git, data stays in place), and keep an explicit restore-on-signal list for what we expect to miss.

## 1. First principles

Volatile/private/bounded context → plan must be external → session-scoped with stable identity → model needs create/update → transitions validated by a system authority → injection has a hard budget, scale derived from it → writes atomic.

**Not derivable from this goal:** event sourcing, task trees, subagent ownership, audit, time-based archiving, pagination, revisions, a 5-level trust taxonomy, free-text evidence. Each must justify itself independently.

## 2. Diagnosis (all verified on current `main`)

- **F1 — `resumeTrust` is dead.** Only production caller classifies with empty refs: 5 levels, 3 reachable; `untrusted` never occurs, so `filterModelVisibleTaskLedgerTasks` is the identity function. A corrupt ledger throws on read rather than being classified.
- **F2 — the "Goal trust gate" is a status gate.** `listActionableTaskKeys` filters on `pending`/`in_progress` only. The belief that trust had a consumer is why it survived.
- **F3 — the projections table is written, never read.** Every read replays all events; the write path commits events and projection in two transactions and signals before the projection is written.
- **F4 — `claimAvailable` is dead code** through 4 layers, zero callers.
- **F5 — `omittedCount` misreports**: counts policy-excluded terminal tasks as "omitted", so every turn the model is told "N omitted — use task_list" for tasks that weren't budget-omitted at all.
- **F6 — one rule, six authorities.** "Completion requires evidence" is enforced in zod, normalize, validate, replay, read-validation, and wire projection. This duplication, more than any field, explains 1092 lines.

## 3. The criterion

[...]

PR #3028, opened 2026-08-15
feat(core,storage): add Work Board Phase 0 contract and store

somewan820:
Work Board Phase 0

This PR implements the maintainer-approved Phase 0 Work Board contract (see the 7-point review and accepted revision in #2560).

Key points:
- durable Work Board items with no total item cap
- bounded pagination (default 50, max 100)
- explicit todo/in_progress/done lifecycle
- strict provenance union (manual / main_conversation / side_conversation)
- SQLite scope/project invariant
- semantic patch mutation semantics
- monotonic revision + optional expectedRevision CAS
- serialized transactional writes (BEGIN IMMEDIATE)
- archive-before-delete semantics
- active-row partial indexes so default list queries stay bounded under archive-heavy data
- schema 8 -> 9 additive migration
- fail-closed corruption detection

Maintainer-confirmed deferrals (liugddx, #2560):
- `linkedSessions` moves to Phase 3; Phase 0 rejects the field in create input and stored records
- linked-session projection is deferred to Phase 3; it will be implemented beside the canonical Runtime Host / Desktop continuity adapter against `SessionContinuitySnapshot` / `TurnSnapshot`

Intentionally deferred:
- Task Ledger integration
- Runtime Host integration
- model-visible Work Board tools
- recovery/graph state
- resultRefs
- automatic completion

## Verification

- core + storage focused tests pass (30 tests)
- @maka/core and @maka/storage build/typecheck
- Biome clean
- migration: v0.1.6 fixture + exact v8 DDL -> real 8->9 path; old schema objects retained; partial indexes present
- archive-heavy list queries covered by EXPLAIN QUERY PLAN regression test against the production query builder
- cross-process CAS covered by a worker-thread test (one winner, one operation_conflict)

## Checklist

- [x] Phase 0 contract matches the maintainer-approved scope and deferrals
- [x] Migration is additive (8 -> 9) with no user-data DROP
- [x] Tests updated and passing
- [x] Docs and PR description current

## AI assistance disclosure

Codex (OpenAI) was used to prepare substantial parts of the implementation and its tests. The contributor reviewed the output and owns the final result.

Generated-by: Codex

Refs #2560


Astro-Han (commented):
Thanks for defining the Work Board authority boundary before adding UI or Runtime integration.

Before getting into the implementation details, I think the product scope should be reconciled with the more recent decision in [#2290](https://github.com/maka-agent/maka-agent/issues/2290). That discussion deliberately reduced the Session Task Ledger to a narrow Session-local Todo snapshot: no stable item identity, per-item CRUD, hierarchy, evidence, pagination, or workflow UI.

I understand that [#2560](https://github.com/maka-agent/maka-agent/issues/2560) describes a different owner: Work Board items are user-owned, cross-Session deferred intentions, while SessionTodo is a model-authored progress snapshot. I am not suggesting that Work Board should reuse SessionTodo or the Task Ledger.

The remaining question is whether this distinct product surface needs a full Work Board domain yet. This PR introduces another durable task-like entity with stable IDs, per-item mutations, scopes, provenance variants, a state machine, archive/delete behavior, pagination, revisions, and future Session links—before there is a UI or production caller. That appears to reintroduce much of the product and maintenance complexity that #2290 has just removed from the existing task system.

Could we first establish why a minimal user-owned Todo list is insufficient for the actual problem of “remember this for later without interrupting the current Session”? A small vertical slice—capture, list, complete, and delete, with optional project/source context—would validate the user need without fixing the larger Board contract in schema v1. Session linking and execution projection could then be added when “start as task” has a real consumer.

If the broader Work Board scope remains intentional after that reconciliation, I found three implementation issues that seem worth resolving before the Phase 0 schema is published:

1. The revision check is outside the SQLite write transaction. Two processes can both read revision 1, both pass `expectedRevision: 1`, both write revision 2, and the later write silently overwrites the earlier semantic patch. Please make the read/check/write sequence one `BEGIN IMMEDIATE` transaction, or use conditional `UPDATE`/`DELETE ... WHERE revision = ?` statements and treat zero affected rows as a conflict.

2. The linked-session projector validates a new DTO rather than the canonical `SessionContinuitySnapshot`. It also uses Core `TurnStatus`, which rejects valid continuity states such as `waiting_for_user` and `cancelled`. Since there is no production consumer yet, the simplest option may be to defer this projector until it can live beside the real Runtime Host/Desktop continuity adapter instead of introducing a parallel contract now.

3. The new index does not satisfy the list ordering. The default unscoped query performs a full-table scan and temporary sort, while scoped queries still require a temporary sort for the final key because `item_id` has the opposite direction. Since the contract intentionally has no total item cap, pagination currently bounds returned rows but not query work. Please align the indexes with the actual scoped/unscoped query shapes before schema v9 is released.

The focused tests and current CI are green. I would also replace the fresh-database “migration” test with a real schema-8 upgrade fixture, but I see that as a verification improvement rather than a separate production finding.

<details>

感谢你在增加 UI 或 Runtime 集成前，先明确 Work Board 的权威边界。

在讨论具体实现前，我认为这里的产品范围需要先与 [#2290](https://github.com/maka-agent/maka-agent/issues/2290) 中较新的决定重新对齐。该讨论刚刚把 Session Task Ledger 明确收缩为一个很窄的 Session-local Todo 快照：没有稳定的事项身份、逐项 CRUD、层级、证据、分页或 workflow UI。

我理解 [#2560](https://github.com/maka-agent/maka-agent/issues/2560) 定义的是另一个 owner：Work Board 是用户拥有的、跨 Session 的暂缓事项；SessionTodo 则是模型写入的进度快照。我并不是建议 Work Board 复用 SessionTodo 或 Task Ledger。

真正需要确认的是：这个独立的产品界面现在是否已经需要一套完整的 Work Board 领域模型。本 PR 在还没有 UI 和生产调用方时，就引入了另一个持久化的 task-like 实体，包括稳定 ID、逐项 mutation、scope、多种 provenance、状态机、归档/删除、分页、revision，以及未来的 Session 关联。这似乎重新引入了 #2290 刚从现有任务系统中删除的大量产品与维护复杂度。

我们是否应该先说明：对于“记住这个后续事项，但不要打断当前 Session”这一真实问题，为什么一个最小的、用户拥有的 Todo 列表还不够？先交付一条较小的纵向链路——保存、列出、完成和删除，并按需携带项目或来源上下文——就可以验证用户需求，而不必现在把更大的 Board contract 固化进 schema v1。等“启动为任务”出现真实消费者时，再增加 Session 关联和执行状态投影也不迟。

如果重新对齐后，完整的 Work Board 范围仍然是有意保留的，我还发现三个建议在发布 Phase 0 schema 前解决的实现问题：

1. revision 检查发生在 SQLite 写事务之外。两个进程可以同时读到 revision 1，都通过 `expectedRevision: 1`，然后分别写入 revision 2；后写入的 semantic patch 会静默覆盖前一次修改。建议把读取、检查和写入放进同一个 `BEGIN IMMEDIATE` 事务，或者使用带 `WHERE revision = ?` 条件的 `UPDATE`/`DELETE`，并将影响行数为零视为冲突。

2. linked-session projector 校验的是一套新定义的 DTO，而不是权威的 `SessionContinuitySnapshot`。它还使用了 Core `TurnStatus`，因此会拒绝 `waiting_for_user`、`cancelled` 等合法的 continuity 状态。当前还没有生产调用方，最简单的方案可能是先推迟这个 projector，等真实的 Runtime Host/Desktop continuity adapter 出现时再贴近实际 owner 实现，避免现在引入一份平行契约。

3. 新增索引不能满足 list 的实际排序。默认的无 scope 查询会全表扫描并临时排序；scoped 查询也会因为 `item_id` 的索引方向与查询相反，对最后一个排序键进行临时排序。由于契约明确不设置事项总量上限，目前分页只限制了返回行数，并没有限制每页查询的工作量。建议在 schema v9 发布前，让索引与 scoped/unscoped 查询的真实形态一致。

Focused tests 和当前 CI 均已通过。我还建议把“从空数据库创建 v9”的 migration 测试替换成真实的 schema 8 升级 fixture，不过这属于验证改进，不是独立的生产 finding。

</details>

AI-assisted review disclosure: Codex and Claude Opus were used to inspect the change and challenge the findings. I reproduced and adjudicated the issues above against the current source and own this review.

somewan820:
Thanks for the review. I have verified all three implementation findings and fixed them in `476eb218b`:

1. **Revision CAS** — `update` / `archive` / `unarchive` / `remove` now run the full read → revision check → write sequence inside one `BEGIN IMMEDIATE` transaction. The in-process write queue is retained for process-local serialization; the transaction closes the cross-process race. I confirmed the shared transaction helper flattens the nested `#writeItem` call, so there is no premature commit.

2. **Linked-session projection** — `work-board-session-projection.ts` (module, test, and package export) is removed. It was a parallel DTO validated with Core `TurnStatus`, which cannot represent `waiting_for_user` / `cancelled` from the canonical continuity contract. With no production consumer in Phase 0, I agree the projector should wait until Phase 3 and live beside the real Runtime Host / Desktop continuity adapter.

3. **Indexes** — schema v9 now ships `(scope_kind, project_id, updated_at DESC, item_id DESC)` and `(updated_at DESC, item_id DESC)`. I verified against populated data with `EXPLAIN QUERY PLAN`: the default unscoped query, scoped queries (project and inbox), and both cursor forms use the indexes and no longer perform a full-table scan or temporary sort.

The migration test now builds a real schema-8 database from the released v0.1.6 operational-state fixture plus the exact v8 workflow DDL, then runs the actual 8→9 migration path and checks that the board table exists and the old schema objects remain.

On the product-scope question, I would like to separate PR fidelity from contract re-scoping:

- **Fidelity to #2560**: every semantic in this PR maps to the maintainer-approved Phase 0 contract from liugddx's review (https://github.com/maka-agent/maka-agent/issues/2560#issuecomment-5294753604) and the accepted revision (https://github.com/maka-agent/maka-agent/issues/2560#issuecomment-5294956443). Most of the surface you listed is load-bearing for that contract: pagination and archive/delete replaced the total-item cap the maintainer removed; revision/CAS was explicitly requested for concurrent renderer writes; provenance variants are required for side-chat capture to survive fork deletion; stable identity is the reference point Phase 3 "start as task" will need.

- **#2290's Option 2 governs SessionTodo, not Work Board**: the explicit-absent list in #2290 applies to the model-authored, Session-local progress snapshot. Work Board is a different owner (user-owned, cross-Session). Your own acceptance comment in #2290 states that #2560 Phase 0 may proceed independently and should not become a P3 integration dependency of #2290 (https://github.com/maka-agent/maka-agent/issues/2290#issuecomment-5298957426).

The one item I agree has no Phase 0 consumer today is `linkedSessions`: the field exists in the schema, but Phase 0 has no mutation path that writes it and no UI or runtime caller reads it; Phase 3 "start as task" is the intended consumer. Rather than removing a maintainer-approved field unilaterally, I would like a decision.

@liugddx, could you please confirm:

(a) Does Phase 0 remain the approved contract after #2290 Option 2, or should it be re-scoped to capture / list / complete / delete with optional project and source context?

(b) Should `linkedSessions` stay in schema v1, or move to Phase 3 via a future additive migration?

(c) Is the linked-session projection formally deferred (removed from the Phase 0 deliverables), or do you want it implemented in Phase 0 against the canonical `SessionContinuitySnapshot` / `TurnSnapshot`?

I will not push further commits until these are resolved.

<details>

感谢 review。三个实现问题我都已核验并在 `476eb218b` 修复：

1. **Revision CAS**：`update` / `archive` / `unarchive` / `remove` 现在把“读取 → revision 校验 → 写入”完整放进同一个 `BEGIN IMMEDIATE` 事务。进程内 write queue 保留用于单进程串行化；事务关闭跨进程竞争窗口。我已确认共享事务助手会扁平化嵌套的 `#writeItem` 调用，不会提前提交。

2. **Linked-session projection**：`work-board-session-projection.ts`（模块、测试、package export）已删除。它是一套平行 DTO，用 Core `TurnStatus` 校验，无法表达 canonical continuity 中的 `waiting_for_user` / `cancelled`。Phase 0 没有生产消费者，我同意推迟到 Phase 3，并放到真实 Runtime Host / Desktop continuity adapter 旁边实现。

3. **索引**：schema v9 现在包含 `(scope_kind, project_id, updated_at DESC, item_id DESC)` 和 `(updated_at DESC, item_id DESC)`。我用真实填充数据跑了 `EXPLAIN QUERY PLAN`：默认 unscoped 查询、scoped 查询（project 与 inbox）以及两种 cursor 形态都使用索引，不再全表扫描或临时排序。

Migration 测试现在用发布的 v0.1.6 operational-state fixture + 精确的 v8 workflow DDL 构造真实 schema-8 库，走完整 8→9 迁移路径，并断言新表存在、旧 schema 对象保留。

关于产品范围问题，我想把“PR 是否忠实”和“契约是否要重新收缩”分开：

- **对 #2560 的忠实度**：本 PR 的每个 semantic 都能对应 liugddx 的 maintainer review（https://github.com/maka-agent/maka-agent/issues/2560#issuecomment-5294753604）和已接受的修订版（https://github.com/maka-agent/maka-agent/issues/2560#issuecomment-5294956443）。你点名的多数复杂度都是该契约的承重结构：分页和 archive/delete 替代了 maintainer 移除的总上限；revision/CAS 是并发 renderer 写入的明确要求；provenance variants 是 side-chat 捕获在 fork 删除后仍可追溯的验收条件；stable identity 是 Phase 3 “start as task” 需要的引用点。

- **#2290 Option 2 约束的是 SessionTodo，不是 Work Board**：#2290 的 explicitly-absent 列表针对 model-authored、Session-local 的进度快照。Work Board 是不同的 owner（user-owned、cross-Session）。你在 #2290 的接受评论里也明确写了 #2560 Phase 0 可以独立推进，不应成为 #2290 的 P3 集成依赖（https://github.com/maka-agent/maka-agent/issues/2290#issuecomment-5298957426）。

我唯一同意“Phase 0 没有消费者”的是 `linkedSessions`：字段在 schema 里，但 Phase 0 没有任何 mutation 路径写入它，也没有 UI 或 runtime 调用方读取；Phase 3 “start as task” 才是消费者。与其单方面删除 maintainer 已批准的字段，我希望得到一个明确决定。

@liugddx 能否确认：

(a) #2290 Option 2 之后，Phase 0 是否仍然以已批准契约为准，还是收缩为 capture / list / complete / delete + 可选的 project/source context？

(b) `linkedSessions` 留在 schema v1，还是移到 Phase 3（未来 additive migration）？

(c) linked-session projection 是否正式推迟（从 Phase 0 交付物移除），还是要求在 Phase 0 基于 canonical `SessionContinuitySnapshot` / `TurnSnapshot` 实现？

在得到答复前我不会再 push 新 commit。

</details>

liugddx (commented):
Thanks for addressing Astro-Han's review carefully. I rechecked the updated implementation against the approved #2560 Phase 0 contract and the later #2290 Option 2 boundary.

My maintainer decisions are:

1. **Phase 0 remains an independently approved Work Board contract.** #2290 Option 2 narrows the model-authored, Session-local `SessionTodo`; it does not re-scope the user-owned, cross-Session Work Board into the same domain. The broader Phase 0 contract therefore does not need to be reduced to a generic Todo list solely because of #2290.
2. **Move `linkedSessions` to Phase 3.** It has no Phase 0 consumer, and it is already exposed as writable input through [`CreateWorkBoardItemInput`](https://github.com/maka-agent/maka-agent/blob/476eb218b388a5f52bf76e71d4c374119eeb34a5/packages/core/src/work-board.ts#L84-L91) and persisted by the [create normalizer](https://github.com/maka-agent/maka-agent/blob/476eb218b388a5f52bf76e71d4c374119eeb34a5/packages/core/src/work-board.ts#L276-L311), so it is not merely a harmless reserved field. An additive migration when “start as task” has a real caller is the smaller and safer compatibility commitment.
3. **The linked-session projection is formally deferred to Phase 3.** Removing the parallel DTO was correct. Implement it beside the real Runtime Host/Desktop continuity adapter against canonical `SessionContinuitySnapshot` / `TurnSnapshot` facts when the consumer exists.

The CAS fix is correct: the read, revision check, and write now share one `BEGIN IMMEDIATE` transaction. The schema-8 upgrade fixture is also a meaningful improvement.

One implementation issue remains before merge: the index fix removes the temporary sort but does not bound filtering work for the default query. [`list()` adds `archived = 0`](https://github.com/maka-agent/maka-agent/blob/476eb218b388a5f52bf76e71d4c374119eeb34a5/packages/storage/src/work-board-store.ts#L90-L115), while neither [ordering index](https://github.com/maka-agent/maka-agent/blob/476eb218b388a5f52bf76e71d4c374119eeb34a5/packages/storage/src/sqlite-workflow-schema.ts#L109-L113) contains `archived`. With an unbounded number of archived rows and few active rows, SQLite may still walk most of the ordering index to produce one active page. `EXPLAIN QUERY PLAN` reporting an index scan and no temporary B-tree does not disprove that scan cost.

Please add active-row partial indexes for both unscoped and scoped query shapes (or an equivalent design), and cover the archive-heavy case. After removing `linkedSessions` from the Phase 0 contract/store/tests and closing this remaining query-cost hole, the PR will match the approved authority boundary and be ready for another pass.

<details>

感谢你认真处理 Astro-Han 的评审。我重新对照了已批准的 #2560 Phase 0 契约和后续 #2290 Option 2 的权威边界。

我的 maintainer 决策如下：

1. **Phase 0 继续作为独立批准的 Work Board 契约推进。** #2290 Option 2 收缩的是模型写入、Session-local 的 `SessionTodo`，并没有把用户拥有、跨 Session 的 Work Board 收缩进同一个领域。因此，不能仅因为 #2290 就把完整 Phase 0 再缩成普通 Todo 列表。
2. **`linkedSessions` 移到 Phase 3。** Phase 0 没有消费者，而且它已经通过 [`CreateWorkBoardItemInput`](https://github.com/maka-agent/maka-agent/blob/476eb218b388a5f52bf76e71d4c374119eeb34a5/packages/core/src/work-board.ts#L84-L91) 暴露为可写输入，并由 [create normalizer](https://github.com/maka-agent/maka-agent/blob/476eb218b388a5f52bf76e71d4c374119eeb34a5/packages/core/src/work-board.ts#L276-L311) 持久化，所以它并不是一个无害的预留字段。等“启动为任务”出现真实调用方时，通过增量迁移增加，是更小、更安全的兼容性承诺。
3. **linked-session projection 正式推迟到 Phase 3。** 删除平行 DTO 是正确的。等真实消费者出现时，再放到 Runtime Host/Desktop continuity adapter 旁，基于规范的 `SessionContinuitySnapshot` / `TurnSnapshot` 事实实现。

CAS 修复成立：读取、revision 校验和写入现在位于同一个 `BEGIN IMMEDIATE` 事务中。真实 schema-8 升级 fixture 也是有意义的改进。

合并前还剩一个实现问题：索引修复消除了临时排序，但没有限制默认查询的过滤工作量。[`list()` 会添加 `archived = 0`](https://github.com/maka-agent/maka-agent/blob/476eb218b388a5f52bf76e71d4c374119eeb34a5/packages/storage/src/work-board-store.ts#L90-L115)，但两个[排序索引](https://github.com/maka-agent/maka-agent/blob/476eb218b388a5f52bf76e71d4c374119eeb34a5/packages/storage/src/sqlite-workflow-schema.ts#L109-L113)都不包含 `archived`。当归档记录无上限、活动记录很少时，SQLite 为返回一页活动事项仍可能遍历排序索引中的大多数记录。`EXPLAIN QUERY PLAN` 显示使用索引且没有临时 B-tree，并不能证明扫描成本受限。

请为 scoped 和 unscoped 的活动项查询增加 partial indexes（或等价设计），并覆盖“绝大多数记录已归档”的情况。删除 Phase 0 contract/store/tests 中的 `linkedSessions`，并补齐这个查询成本缺口后，本 PR 就能符合已批准的权威边界，可以进入下一轮评审。

</details>

somewan820:
Thanks for the maintainer decisions. I've applied them in `833481e64`:

- `linkedSessions` is removed from the Phase 0 contract, create input, store, and tests. The Phase 0 normalizer and decoder now reject the field (fail-closed), and the contract doc records it as a Phase 3 addition.
- The linked-session projection stays deferred to Phase 3, documented in the contract and the PR description.
- Active-row partial indexes were added for both default query shapes: `workflow_work_board_items_active_order` and `workflow_work_board_items_active_scope_order` (both `WHERE archived = 0`). The existing full ordering indexes remain for `includeArchived` queries.
- Archive-heavy coverage: the new storage test creates 120 archived + 6 active items, asserts list behavior, and verifies via `EXPLAIN QUERY PLAN` that default unscoped and scoped queries use the partial indexes without a temporary sort. The migration test also asserts both partial indexes exist after the 8→9 upgrade.

Verification: core + storage focused tests pass (26 tests), both packages build/typecheck, Biome is clean.

The PR is now ready for review. Let me know if you'd like any further adjustments.

<details>

感谢 maintainer 的决定。我已在 `833481e64` 全部落地：

- `linkedSessions` 已从 Phase 0 契约、create input、store 和测试中移除；Phase 0 的 normalizer 和 decoder 现在会拒绝该字段（fail-closed），契约文档记录为 Phase 3 加入。
- linked-session projection 保持推迟到 Phase 3，已写入契约文档和 PR description。
- 为两种默认查询形态新增 active-row partial index：`workflow_work_board_items_active_order` 和 `workflow_work_board_items_active_scope_order`（均为 `WHERE archived = 0`）；原有全量排序索引保留给 `includeArchived` 查询。
- Archive-heavy 覆盖：新增 storage 测试构造 120 条 archived + 6 条 active，断言 list 行为，并用 `EXPLAIN QUERY PLAN` 验证默认 unscoped/scoped 查询使用 partial index、无临时排序；migration 测试同时断言 8→9 升级后两个 partial index 存在。

验证：core + storage focused tests 全部通过（26 个），两个包 build/typecheck 通过，Biome 干净。

PR 已标记 ready for review，如需进一步调整请告诉我。

</details>

packages/storage/src/work-board-store.ts
```diff
@@ -0,0 +1,392 @@
...
+    if (separator <= 0) return null;
+    const updatedAt = Number(raw.slice(0, separator));
+    const itemId = raw.slice(separator + 1);
+    if (!Number.isSafeInteger(updatedAt) || updatedAt < 0) return null;
+    if (!isSafeWorkBoardId(itemId)) return null;
+    return { updatedAt, itemId };
+  } catch {
+    return null;
+  }
+}
```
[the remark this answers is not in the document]
somewan820: Thanks for the check. This is actually unambiguous: updatedAt is always a non-negative integer (no :), and isSafeWorkBoardId requires IDs to start with [A-Za-z0-9], so the first : is always the separator. decodeCursor slices after it and preserves any colons inside the ID — I verified round-trips for with:colon / a:b:c:d. No change needed; I can add a round-trip test with a colon-containing ID if you'd like it documented.

docs/work-board-contract.md
```diff
@@ -0,0 +1,138 @@
...
+  scope: WorkBoardScope;   // inbox | { kind: 'project'; projectId }
+  title: string;
+  notes?: string;
+  state: 'todo' | 'in_progress' | 'done';
+  creator: { kind: 'user' } | { kind: 'agent_suggestion'; confirmedAt: number };
+  provenance: WorkBoardProvenance;
+  createdAt: number;
+  updatedAt: number;
+  archived: false | true;  // active items never carry archivedAt; archived items must
+  archivedAt?: number;
```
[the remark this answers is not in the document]
somewan820: Fixed in `3df3cc7c8`: the contract doc now states the archive invariant as a discriminated union and completes the truncated comment. The implementation already encodes this via `WorkBoardActiveItem | WorkBoardArchivedItem`, so no code change was needed.

packages/storage/src/sqlite-workflow-schema.ts
```diff
@@ -90,6 +90,34 @@ export function migrateSqliteWorkflowDatabase(db: DatabaseSync): void {
...
+      CHECK (
+        (scope_kind = 'inbox' AND project_id IS NULL)
+        OR
+        (scope_kind = 'project' AND project_id IS NOT NULL)
+      )
+    );
+
+    DROP INDEX IF EXISTS workflow_work_board_items_scope_order;
+    CREATE INDEX IF NOT EXISTS workflow_work_board_items_scope_order
+      ON workflow_work_board_items(scope_kind, project_id, updated_at DESC, item_id DESC);
```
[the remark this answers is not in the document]
somewan820: Thanks for the look. This migration does not run on every open: `inspectAndMigrateOperationalState` returns early when the schema registry is current and the target schema matches, so the DROP only executes during an actual migration (at most once per version transition). I kept the DROP deliberately: the first Phase 0 dev commit shipped this index with `item_id ASC`, and a dev database created from that commit is registered as v9 with the old definition. Without the DROP, the target-schema check would keep failing and `CREATE INDEX IF NOT EXISTS` cannot rewrite a same-name index, so that database would re-run migration on every open instead of converging. I added a comment in the schema explaining this intent.

packages/storage/test-fixtures/workflow-schema-v8.sql
```diff
@@ -0,0 +1,113 @@
...
+ALTER TABLE workflow_quote_companion_cleanup ADD COLUMN record_json TEXT;
+UPDATE workflow_quote_companion_cleanup
+SET record_json = json_object(
+  'version', 1,
+  'sessionId', session_id,
+  'trackedAt', tracked_at,
+  'phase', 'cleanup',
+  'cancelRequested', json('true')
+)
+WHERE record_json IS NULL;
```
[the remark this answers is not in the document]
somewan820: Fixed in `3df3cc7c8`: the fixture header now documents the precondition that `workflow_quote_companion_cleanup` must already exist without `record_json` (as it does in the v0.1.6 fixture), so the `CREATE TABLE IF NOT EXISTS` is a no-op for it and the trailing `ALTER TABLE` can add the column.

packages/core/src/work-board.ts
```diff
@@ -0,0 +1,558 @@
...
+      creator: creator.value,
+      provenance: provenance.value,
+    },
+  };
+}
+
+export function normalizeUpdateWorkBoardItemInput(
+  input: unknown,
+): WorkBoardNormalizeResult<UpdateWorkBoardItemInput> {
+  if (!isRecord(input)) return fail('Work Board update must be an object');
```
Astro-Han, line 358: **[P2] Reject unknown fields consistently at the contract boundary**

Could we make this an exact-shape boundary instead of silently ignoring unknown keys? For example, `update(id, { linkedSessions: [...] })`, `{ titel: 'x' }`, or `{ status: 'done' }` is currently accepted as an empty patch: the call succeeds, the item does not change, and the revision stays the same. The durable decoder has the same underlying behavior for arbitrary schema-v1 fields; I reproduced `get()` accepting such a row and the next mutation permanently removing the unknown field from `record_json`.

Core already has `defineObjectShape()` / `hasExactShape()` in `record-schema.ts`. Reusing that seam for the item and its nested/input unions would keep Phase 0 fail-closed consistently and avoid accumulating one-off checks for fields such as `linkedSessions`.

packages/core/src/work-board.ts
```diff
@@ -0,0 +1,558 @@
...
+  if (!changed) return { item, changed: false };
+  const { notes: _removedNotes, ...itemWithoutNotes } = item;
+  return {
+    item: {
+      ...itemWithoutNotes,
+      title,
+      ...(notes === undefined ? {} : { notes }),
+      scope,
+      state,
+      updatedAt: now,
```
Astro-Han, line 543: **[P2] Keep mutation timestamps monotonic across clock rollback**

Could the storage transaction clamp the timestamp before passing it into these pure transition helpers? I reproduced an item returned on page one appearing again on a later page after it was updated with an earlier `now`: its `updatedAt` moved behind the cursor. `archive()` and `unarchive()` have the same path.

Other storage authorities already use `Math.max(now, current.updatedAt)`. Applying that rule after reading `current` inside the Work Board transaction would preserve the keyset ordering invariant, while keeping these Core helpers free of persistence policy; the clamped value can also serve as `archivedAt`.

packages/storage/src/work-board-store.ts
```diff
@@ -0,0 +1,392 @@
...
+      WHERE 1 = 1
+    `;
+    if (!value.includeArchived) {
+      sql += ' AND archived = 0';
+    }
+    if (value.scope) {
+      sql = appendScopePredicate(sql, params, value.scope);
+    }
+    if (value.cursor) {
+      const cursor = decodeCursor(value.cursor);
```
Astro-Han: **[P2] Bind opaque cursors to the result-set filters**

Could the cursor include a fingerprint of the normalized `scope` and `includeArchived` filters? It currently carries only `updatedAt` and `itemId`. I reused an Inbox cursor with a `project:p1` query; the cursor decoded successfully and the query returned an empty page even though that Project contained a newer item. Switching `includeArchived` has the same mismatch.

Because Phase 0 has no consumers yet, this is the least expensive point to bind the cursor to its result set and return the existing `invalid_input` error when the filters do not match.

Astro-Han (commented):
Thanks for working through the earlier review rounds. I rechecked the current head, including the Core contract, SQLite lifecycle, migration, indexes, existing feedback, and focused store tests. The earlier CAS, linked-session projection, indexing, notes-patch, and migration-fixture findings are resolved.

I left three inline P2 comments covering exact-shape validation, monotonic mutation timestamps, and cursor/filter binding. None is a P0/P1, and I do not think this needs a broader redesign or PR split.

Two non-blocking test-truth follow-ups:

- The concurrent-mutation test uses one store, so it proves process-local serialization but would still pass without the cross-process `BEGIN IMMEDIATE` fix. I would keep it and name it accordingly; a worker/barrier test can cover cross-process CAS if that guarantee needs executable protection.
- The query-plan test is valuable, but it explains a handwritten copy of the list SQL rather than the SQL produced by `list()`. Sharing an internal query builder would keep the regression test tied to production without adding a public API.

I would keep the remaining contract, persistence, migration, archive, backup, and index tests; they cover distinct lifecycle seams rather than duplicating one another. Overall, the Work Board / SessionTodo authority split is sound, and the current Phase 0 scope is appropriately reduced.

<details>
<summary>中文对照</summary>

感谢持续处理前几轮 review。我重新检查了当前 head，包括 Core 契约、SQLite 生命周期、migration、索引、已有反馈和聚焦的 store 测试。之前的 CAS、linked-session projection、索引、notes patch 和 migration fixture 问题都已经解决。

我留下了三条 inline P2，分别涉及：

- 对 Work Board item 和 mutation input 进行 exact-shape 校验，避免未知字段被静默接受或在下一次写入时丢失；
- 在持久化事务中保持 mutation timestamp 单调，避免系统时钟回拨破坏 keyset pagination；
- 将 opaque cursor 与规范化后的 scope / includeArchived filters 绑定，避免跨结果集复用时静默漏项。

它们都不是 P0/P1，也不需要扩大成整体重构或拆分 PR。

另外有两项非阻塞的测试真实性建议：

- 当前并发 mutation 测试使用同一个 store，因此证明的是进程内串行化；即使没有跨进程 `BEGIN IMMEDIATE` 修复，它仍会通过。建议保留并让测试名称准确表达该范围；如需可执行地保护跨进程 CAS，再补 worker/barrier 测试。
- Query-plan 测试有价值，但它解释的是手写的 list SQL 副本，而不是 `list()` 实际生成的 SQL。共享内部 query builder 可以防止测试与生产查询逐渐漂移，同时不需要扩大公共 API。

其余 contract、persistence、migration、archive、backup 和 index 测试覆盖不同生命周期，不建议删除。总体而言，Work Board 与 SessionTodo 的权威边界正确，Phase 0 的当前范围也已经合理收缩。

</details>

AI-assisted review disclosure: Codex performed the primary lifecycle review and executable reproductions; three independent Codex reviewers covered contract scope, SQLite behavior, and test truthfulness; Claude performed a focused adversarial adjudication. The final findings were verified against the current PR head.

packages/core/src/work-board.ts
somewan820: Fixed in `c75efd489`: create/update/list-query/decode now use `defineObjectShape` / `hasExactShape` from `record-schema.ts` for the item and its nested scope/creator/provenance unions. Unknown fields such as `titel`, `status`, `linkedSessions`, or arbitrary schema-v1 fields are rejected as `invalid_input` / corrupt records instead of being silently ignored or stripped on the next mutation. The one-off `linkedSessions` check was replaced by this generic mechanism.

packages/core/src/work-board.ts
somewan820: Fixed in `c75efd489`: the store transaction clamps `now` to `Math.max(now, current.updatedAt)` after reading `current`, for update/archive/unarchive, and the clamped value is used as `archivedAt`. The Core helpers stay pure. Added a clock-rollback test asserting `updatedAt` / `archivedAt` never decrease.

packages/storage/src/work-board-store.ts
somewan820: Fixed in `c75efd489`: cursors now carry a fingerprint of the normalized `scope` / `includeArchived` filters and are JSON-encoded (which also removes the earlier colon-separator ambiguity). Reusing a cursor with different filters returns `invalid_input`, covered by a dedicated test.

somewan820:
Thanks for the thorough re-review. I addressed all three P2 comments in `c75efd489` (details in the inline threads):

- **Exact-shape contract boundary**: create/update/list-query/decode now use `defineObjectShape` / `hasExactShape` for the item and its nested scope/creator/provenance unions; unknown fields are rejected instead of silently ignored or stripped on the next mutation.
- **Monotonic timestamps**: the store transaction clamps `now` to `Math.max(now, current.updatedAt)` for update/archive/unarchive, with the clamped value used as `archivedAt`; Core helpers stay pure.
- **Filter-bound cursors**: cursors carry a fingerprint of normalized `scope` / `includeArchived` and are JSON-encoded; cross-filter reuse returns `invalid_input`.

I also followed your two non-blocking test suggestions in `7845840af`:

- Added a worker-thread **cross-process CAS test**: two workers racing `expectedRevision: 1` produce exactly one winner and one `operation_conflict`, and the final row reflects the winner.
- Extracted the list statement builder into an internal module (`work-board-list-query.ts`, not in package exports) shared by `WorkBoardStore.list()` and the `EXPLAIN QUERY PLAN` regression test, so the test executes the production SQL instead of a handwritten copy.

Verification: core + storage focused tests pass (30 tests), both packages build/typecheck, Biome clean. Ready for another pass.

<details>

感谢这次细致的 re-review。三条 P2 已在 `c75efd489` 处理（细节见各 inline thread）：

- **Exact-shape 契约边界**：create/update/list-query/decode 现在用 `defineObjectShape` / `hasExactShape` 校验 item 及其嵌套 scope/creator/provenance union；未知字段会被拒绝，而不是被静默忽略或在下次 mutation 中被抹掉。
- **单调时间戳**：store 事务在读取 `current` 后把 `now` clamp 为 `Math.max(now, current.updatedAt)`，update/archive/unarchive 均覆盖，`archivedAt` 使用 clamp 后的值；Core 纯函数保持无持久化策略。
- **Filter-bound cursor**：cursor 携带规范化 `scope` / `includeArchived` 的 fingerprint 并改用 JSON 编码；跨 filter 复用 cursor 返回 `invalid_input`。

你提的两条非阻塞测试建议也在 `7845840af` 落地：

- 新增 worker 线程的**跨进程 CAS 测试**：两个 worker 同时以 `expectedRevision: 1` 竞争，恰好一个成功、一个 `operation_conflict`，最终行反映胜者。
- 把 list SQL 构建抽到内部模块（`work-board-list-query.ts`，未加入 package exports），由 `WorkBoardStore.list()` 与 `EXPLAIN QUERY PLAN` 回归测试共用，测试执行的是生产 SQL，而不是手写副本。

验证：core + storage focused tests 全部通过（30 个），两个包 build/typecheck 通过，Biome 干净。可以进入下一轮 review。

</details>

packages/storage/src/__tests__/work-board-store.test.ts
```diff
@@ -0,0 +1,583 @@
...
+          (error: unknown) =>
+            error instanceof WorkBoardStoreError && error.code === 'invalid_input',
+        );
+
+        const secondPage = await store.list({
+          limit: 1,
+          scope: { kind: 'inbox' },
+          cursor: inboxPage.nextCursor,
+        });
+        assert.equal(secondPage.items.length, 1);
```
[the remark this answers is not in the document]
somewan820: Fixed in `031a52a9b`: the cursor test now captures the first item id and asserts the second page returns a different item, so a store that silently ignores the cursor can no longer pass.

packages/core/src/work-board.ts
```diff
@@ -0,0 +1,639 @@
...
+export const WORK_BOARD_NOTES_MAX_CHARS = 4_000;
+export const WORK_BOARD_EXCERPT_MAX_CHARS = 4_000;
+export const WORK_BOARD_ID_MAX_CHARS = 64;
+export const WORK_BOARD_SESSION_ID_MAX_CHARS = 160;
+export const WORK_BOARD_MESSAGE_ID_MAX_CHARS = 256;
+export const WORK_BOARD_PROJECT_ID_MAX_CHARS = 160;
+export const WORK_BOARD_DEFAULT_PAGE_SIZE = 50;
+export const WORK_BOARD_PAGE_SIZE_MAX = 100;
+export const WORK_BOARD_CURSOR_MAX_CHARS = 256;
+
```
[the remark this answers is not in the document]
somewan820: Good catch. Fixed in `10ba261f3`: `WORK_BOARD_CURSOR_MAX_CHARS` is now 1024, the cursor fingerprint validation accepts up to 512 chars, and a regression test round-trips pagination with a maximum-length (160-char) project id, whose returned cursor exceeds the old 256-char limit.

liugddx:
Follow-up: I reviewed the current head (`688051767`). The requested Phase 0 scope changes, active-row partial indexes with archive-heavy coverage, and Astro-Han’s subsequent P2 findings are resolved. The remaining test-truth caveat is non-blocking. This PR is ready to merge.

liugddx (approved):
Approved. The current head satisfies the approved Phase 0 contract and resolves the outstanding review findings.

packages/core/src/work-board.ts
```diff
@@ -0,0 +1,639 @@
...
+  if (left.kind !== right.kind) return false;
+  if (left.kind === 'inbox') return true;
+  return right.kind === 'project' && left.projectId === right.projectId;
+}
+
+function normalizeOptionalText(
+  input: unknown,
+  options: { field: string; max: number; required: boolean },
+): WorkBoardNormalizeResult<string | undefined> {
+  if (input === undefined || input === null) {
```
Astro-Han, line 594: **[P2] Keep `null` exclusive to update patches**

`normalizeOptionalText()` currently maps both `undefined` and `null` to absence when `required` is false. I reproduced both create input with `notes: null` and a durable item containing `notes: null` being accepted and silently returned without the field. `null` is only documented as the clear signal for `UpdateWorkBoardItemInput`; the create and stored-item contracts allow a string or omission. Could we distinguish an absent field from a present `null` here, keeping the explicit `null` handling only in the update normalizer? Tests for both create admission and durable decode would close the boundary.

packages/storage/src/work-board-store.ts
```diff
@@ -0,0 +1,338 @@
...
+          item.scope.kind === 'project' ? item.scope.projectId : null,
+          item.archived ? 1 : 0,
+          JSON.stringify(item),
+        );
+    });
+  }
+}
+
+function normalizeExpectedRevision(options: WorkBoardMutationOptions): number | undefined {
+  const expected = options.expectedRevision;
```
Astro-Han, line 319: **[P2] Fail closed on unknown CAS option keys**

`normalizeExpectedRevision()` reads only `options.expectedRevision`, so an unknown key is silently ignored. I reproduced a stale update after revision 2 with `{ expectedRevison: 1 }`; it succeeded and wrote revision 3 instead of returning `invalid_input` or `operation_conflict`. TypeScript catches many direct literal callers, but the exported runtime JS/serialized boundary cannot rely on that. Since CAS is the lost-update guard and the other Work Board inputs now use exact-shape validation, could the mutation-options object be validated the same way?

Astro-Han (commented):
Thanks for carrying the earlier review rounds through to a coherent Phase 0 boundary. I rechecked the current head after the maintainer approval, including the Core contract, SQLite lifecycle, migration/index paths, prior feedback, and focused tests.

The previous CAS transaction, linked-session deferral, active-row indexes, monotonic mutation timestamps, filter-bound cursors, and cursor-size fixes all hold. I found two remaining runtime-boundary gaps and left them inline:

- create/stored `notes` still accepts `null` even though that clear signal belongs only to update patches;
- unknown mutation-option keys can silently disable `expectedRevision` CAS.

Both are narrow exact-validation fixes at the existing owners. They do not call for reopening the product scope, redesigning the store, or splitting this PR.

Focused verification on `688051767`:

- `@maka/core` build: passed
- `@maka/storage` build: passed
- Work Board storage test file: 17/17 passed
- visible GitHub checks: green

Once the two inline issues are addressed, this should only need a quick final pass.

<details>
<summary>中文版</summary>

感谢把前几轮 review 收敛成连贯的 Phase 0 边界。我在 maintainer Approve 之后重新检查了同一个当前 head，包括 Core 契约、SQLite 生命周期、migration/index、已有反馈和聚焦测试。

此前的 CAS 事务、linked-session 推迟、active partial indexes、mutation 时间戳单调性、游标与过滤条件绑定，以及游标长度问题都已经正确解决。

目前还剩两个运行时边界缺口，已分别留下 inline comment：create/持久化记录仍会接受 `notes: null`；mutation options 的未知字段可能静默关闭 `expectedRevision` CAS。两者都只需要在现有 owner 上补齐精确校验，不需要重新讨论产品范围、重构 store 或拆分 PR。

当前 head 的 Core/Storage 构建通过，Work Board storage 聚焦测试 17/17 通过，GitHub 可见检查全绿。处理完两条 inline 后，快速复核即可。

</details>

_AI-assisted review: Codex performed the primary review and executable reproductions; four independent Codex reviewers adversarially checked the Core contract, SQLite lifecycle, migration/test evidence, and overall architecture. The findings above were adjudicated against the current PR head. The human reviewer remains responsible for the final review and merge decision._

packages/core/src/work-board.ts
somewan820: Fixed in `cbf2dc64a`: create input and stored records now reject `notes: null` (`invalid_input` / corrupt record), while `null` remains the explicit clear signal only in update patches. Added tests for both create admission and durable decode.

packages/storage/src/work-board-store.ts
somewan820: Fixed in `cbf2dc64a`: mutation options are now validated as an exact-shape object — unknown keys such as `expectedRevison` or extra fields throw `invalid_input` instead of silently disabling CAS. Added a regression test proving a stale update with a typo’d option key fails and the item stays at revision 2.

somewan820:
Astro-Han — both P2 inline comments are fixed in `cbf2dc64a` (details in the threads):

- `notes: null` is now rejected in create input and stored records; `null` remains the explicit clear signal only in update patches.
- Mutation options are validated as an exact-shape object, so unknown keys like `expectedRevison` or extra fields throw `invalid_input` instead of silently disabling `expectedRevision` CAS.

Verification: core + storage focused tests pass (33 tests), both packages build/typecheck, Biome clean. liugddx already approved the prior head; the new commit only contains these two narrow validation fixes.

Could you take the quick final pass? If everything looks good, this PR is ready to merge. @liugddx — happy to merge once Astro-Han confirms.

<details>

Astro-Han —— 两条 P2 代码评论已在 `cbf2dc64a` 修复（详见对应线程）：

- create input 和已存记录现在拒绝 `notes: null`；`null` 只保留为 update patch 的显式清除信号。
- mutation options 改为精确形状校验，`expectedRevison` 这类拼写错误或多余字段会直接 `invalid_input`，不会再静默关闭 `expectedRevision` CAS。

验证：core + storage focused tests 全部通过（33 个），两个包 build/typecheck 通过，Biome 干净。liugddx 已对前一版 head 表示 approve；本次新 commit 只包含这两处窄校验修复。

能否做最后一次快速复核？如果没问题，这个 PR 就可以合并了。@liugddx —— Astro-Han 确认后就可以 merge。

</details>

Astro-Han (approved):
Thanks for the quick follow-up. I rechecked the current head (`cbf2dc64a`) and both remaining P2 findings are resolved:

- create input and durable records now reject `notes: null`, while update patches retain `null` as the explicit clear signal;
- mutation options now fail closed on unknown keys, so a misspelled or extra field cannot silently disable `expectedRevision` CAS.

The regression tests exercise the real contract/store boundaries and verify that rejected mutation options leave the stored revision and state unchanged. I found no remaining P0–P3 issues.

The Phase 0 authority boundary remains intact: Core owns the item contract, Storage owns durable mutation and CAS, and no Runtime execution state or parallel authority has been introduced. The current solution is the smallest complete implementation of the approved scope; it does not need further restructuring or a PR split.

Verification on `cbf2dc64a`:

- `@maka/core` build and focused Work Board tests: 15/15 passed
- `@maka/storage` build and focused Work Board tests: 18/18 passed
- Biome formatting check on the four changed files: passed
- `git diff --check`: passed
- visible GitHub checks: green
- merge state: clean

Approved from my side.

<details>
<summary>中文版</summary>

感谢快速处理。我重新检查了当前 head（`cbf2dc64a`），最后两条 P2 都已解决：

- create input 和持久化记录现在拒绝 `notes: null`，同时 update patch 仍保留 `null` 作为显式清除信号；
- mutation options 现在会拒绝未知字段，因此拼错或多余字段不会再静默关闭 `expectedRevision` CAS。

回归测试覆盖了真实的契约和 store 边界，并验证 mutation options 被拒绝后，持久化 revision 和 state 都不会变化。目前没有剩余 P0–P3。

Phase 0 的权威边界保持清晰：Core 负责 item contract，Storage 负责持久化 mutation 与 CAS，没有引入 Runtime 执行状态副本或平行 authority。当前方案已经是已批准范围内的最小完整实现，不需要继续重构或拆分 PR。

当前 head 的 Core/Storage build 通过，聚焦测试分别为 15/15 与 18/18，Biome、`git diff --check` 和 GitHub checks 均通过，merge state 为 clean。

我这边 approve。

</details>

_AI-assisted review disclosure: Codex performed the primary diff and lifecycle review, executable verification, and final synthesis; three independent Codex reviewer agents separately checked the Core contract, Storage/CAS boundary, and test evidence. The human reviewer remains responsible for the final approval and merge decision._

PR #1355, opened 2026-07-22
feat(runtime-host): bound shutdown with process fail-stop

M4n5ter:
<details open>
<summary><strong>English</strong></summary>

## Summary

This PR establishes the bounded shutdown foundation required before a Runtime Host owns additional Interactive domains:

- the first drain request closes new connection and operation admission and starts one absolute clean-shutdown deadline, even while an already admitted command is still reaching its seal point;
- admitted work keeps the existing normal response window, while operation, composition, residency, control-plane, and owner cleanup share one finite budget rather than restarting a timeout at each phase;
- expiry rejects `closed` with a typed `process_termination_required` result, and the dedicated Host process immediately fail-stops instead of continuing ordinary cleanup or force-releasing ownership;
- the OS-held owner lock remains the only writer authority and is reclaimed by process death; a successor still has to acquire that lock before it can recover or serve;
- the production Candidate and the real execution child use the same signal, close, and fail-stop process lifecycle.

## Boundary and decisions

The default clean-shutdown grace is 10 seconds. The existing one-second operation and handshake windows remain early transport-drain bounds; they are not separate ownership deadlines. A command that was admitted before shutdown may finish its handler and enqueue its protocol response, but a command that never reaches that boundary can no longer keep the process and owner lock alive forever.

`RuntimeHostKernel` reports deadline expiry as a typed process-termination requirement instead of calling `process.exit()` from a reusable in-process class. The dedicated process entrypoint owns the hard exit. This keeps the Kernel testable without weakening the production rule: once fail-stop is required, ordinary endpoint cleanup, registration removal, or `owner.close()` cannot be used as proof that the old writer is isolated.

This slice does not add a lock-stealing path, a protocol-visible fail-stop state, a supervisor, or arbitrary-descendant containment. Registration and endpoint files remain discovery evidence only. The successor journey is gated by actual OS lock acquisition.

## Scope

This PR changes Runtime Host lifecycle and its dedicated process entrypoints only. It does not migrate an Interactive domain, change protocol operation shapes, alter Storage's drain-before-unlock contract, or activate the production Host cutover.

## Validation

A real child-process test uses the real local transport and Interactive owner lock, blocks an admitted `turn.start` before seal, verifies that a contender cannot acquire ownership during the grace period, observes bounded non-clean process exit, and then proves that a different-epoch successor acquires the lock and serves `host.status`. The existing active-Turn shutdown journey now also proves a normal code-zero exit and preserves its single durable `cancelled` outcome.

[...]

PR #3031, opened 2026-08-15
fix(runtime-host): bound a never-connected ephemeral candidate's lifetime

### LINKED ISSUE #2943: fix(runtime-host): ephemeral candidate never exits after retainUntilProcessExit disables the initial connection timeout
## Summary

Two `execution-candidate-main.js` processes survived for **26 hours** after their parent exited, each burning **70–82% CPU** the entire time, despite being started in ephemeral mode with `--initial-connection-timeout-ms` set to 1000 and 10000 respectively.

Code reading points at `retainUntilProcessExit()` permanently disabling the initial-connection timeout, with no fallback deadline to force the exit. See the hypothesis below — it is **not** runtime-verified.

## Observed

Captured with `ps -Ao pid,ppid,pcpu,etime,command -r` on macOS 15 (Darwin 25.5.0), Node 24.18.1:

```
PID    PPID  %CPU  ELAPSED      COMMAND
89292  1     69.7  01-01:58:38  node .../runtime-host/dist/execution-candidate-main.js \
                                  --root /private/tmp/maka-review-connect-lock.sfKdJq \
                                  --expected-root-id 70ba08d6... \
                                  --initial-connection-timeout-ms 1000
89559  1     66.9  01-01:58:24  node .../runtime-host/dist/execution-candidate-main.js \
                                  --root /private/tmp/maka-review-connect-abort.aPwNIo \
                                  --expected-root-id ce951ea2... \
                                  --initial-connection-timeout-ms 10000
```

Facts:

- `PPID = 1` — the launching process was long gone; both had been reparented to launchd.
- Started 14 seconds apart, so they came from one batch of runs.
- No client ever connected (the roots were throwaway temp dirs).
- Both survived roughly 94,000× their configured initial connection timeout.
- A third root from the same batch (`maka-review-connect-readonly.*`) had no surviving process, so the leak is not universal across that batch.

Expected: with `lifecycleMode: 'ephemeral'` and no connection ever accepted, the candidate should shut down after `initialConnectionTimeoutMs`.

## Hypothesis (code reading only — not verified at runtime)

In `packages/runtime-host/src/server/host-kernel.ts`:

```ts
#retainUntilProcessExit(): void {
  if (this.#retainedUntilProcessExit) return;
  this.#retainedUntilProcessExit = true;
  this.#residencies.acquire('process-retention');   // never released
  this.#cancelIdle();
}
```

The `process-retention` residency is acquired and never released. `#isTrueIdle()` gates on total residency count:

```ts
#isTrueIdle(): boolean {
  return (
    this.#state === 'ready' &&
    this.#acceptedTransports.size === 0 &&
    this.#activeOperations === 0 &&
    this.#residencies.activeCount === 0      // <-- process-retention keeps this non-zero forever
  );
}
```

So once `#retainUntilProcessExit()` runs, `#scheduleIdleIfNeeded()` can never arm the timer again and `initialConnectionTimeoutMs` becomes permanently inert.

This looks like an oversight rather than intent, because the neighbouring `#hasNonRetentionWork()` deliberately excludes exactly this label:

```ts
return this.#residencies.snapshot().some(({ label }) => label !== 'process-retention');
```

Two treatments of the same label, one filtering it out and one not.

The only callers are the `onPoison` handlers in `packages/runtime-host/src/server/execution-composition.ts` (lines 576 and 954), both of which pair retention with a drain request:

```ts
onPoison: (error) => {
  if (poisonFailure) return;
  poisonFailure = error;
  context.retainUntilProcessExit();
  beginDrain();
  context.requestDrain();
},
```

But `#commitRequestedShutdownIfQuiescent()` returns early unless `activeCommandOperations === 0`, and `#armShutdownDeadline()` is only called from inside `#commitShutdown()`. If poisoning happens while a command operation is still in flight, the process never becomes quiescent, `commitShutdown` never runs, the shutdown deadline is therefore never armed, and nothing else can terminate the process.

## What this does not explain

The hypothesis accounts for **never exiting**. It does **not** account for the sustained **70–82% CPU**. An idle process waiting on a residency should sit at ~0%. Something is spinning, and I did not find it. Unfortunately I killed both processes before capturing a stack sample, so I have no profiler evidence to offer.

## Reproduction

I do not have a reliable repro. The temp-dir prefix `maka-review-connect-*` does not appear anywhere in the repository, so these were launched by an external script calling `execution-candidate-main.js` directly, not by an in-repo test. Anyone reproducing this would need to drive an ephemeral candidate into the `onPoison` path while a command operation is in flight, then confirm the process outlives its initial connection timeout.

## Suggested directions

1. Exclude `process-retention` from `#isTrueIdle()`, consistent with `#hasNonRetentionWork()`, so the initial-connection timeout keeps working after retention.
2. Arm the shutdown deadline when shutdown is *requested*, not only when it is *committed*, so a non-quiescent host still has a hard bound on its lifetime.
3. If a candidate really should be retained for post-mortem after poisoning, cap that retention with an explicit timeout rather than making it unbounded.

Happy to test a patch against the same setup.

---

Environment: macOS 15 (Darwin 25.5.0, arm64), Node 24.18.1 (homebrew `node@24`), built from worktree at `packages/runtime-host/dist/`.

UncertaintyDeterminesYou4ndMe:
## Summary

Fixes #2943.

An ephemeral candidate's initial-connection timeout was implemented only as the true-idle timer (`#scheduleIdleIfNeeded`), which never arms while composition startup is still pending or any residency is held. A candidate whose composition startup hung or that held boot-time residencies before any Client ever connected therefore outlived `--initial-connection-timeout-ms` indefinitely — the orphaned `execution-candidate-main.js` processes in #2943 survived 26+ hours.

## What changed

- `RuntimeHostKernel` now arms a dedicated first-connection deadline for ephemeral lifecycles. If no handshake has been accepted when it fires, the kernel requests a drain, which already carries the bounded fail-stop shutdown deadline (`shutdownGraceMs` → `RuntimeHostProcessTerminationRequiredError` → `process.exit(1)`), so the candidate's lifetime is bounded regardless of what the composition is doing.
- The deadline is cancelled by the first accepted connection, so a Client waiting through a slow recovery is unaffected (Clients handshake before polling `waitForRuntimeHostReady`).
- A handshake in flight defers the deadline by the handshake budget instead of draining under a connecting Client — bounded to 3 deferrals, so a reconnect loop that never completes a handshake cannot push the deadline out indefinitely.
- The deadline is armed only after `#compositionStartup` is assigned: an earlier fire would drive `#closeResources` past an undefined startup await and let shutdown complete without closing the composition created afterwards.

## Notes on the issue's analysis

- `#armShutdownDeadline()` is already called from `#requestDrain()` (since #1355), so the poison path (`retainUntilProcessExit` + `requestDrain`) was already bounded; suggested direction 2 was in place.
- `process-retention` deliberately blocks the clean-close path so a poisoned host fail-stops instead of pretending a clean shutdown; excluding it from `#isTrueIdle()` (direction 1) is unnecessary once the pre-connection lifetime is bounded, and would change post-connection semantics.
- The sustained 70–82% CPU is consistent with a hung/spinning composition startup (state never leaves `recovering`, so neither the idle timer nor any drain existed to stop it). This change does not diagnose the spin, but it bounds it: the process now fail-stops after `initialConnectionTimeoutMs + shutdownGraceMs`.

## Tests

- New: `never-connected ephemeral candidate drains after the initial connection timeout despite a boot residency`
- New: `never-connected ephemeral candidate with a hung composition startup fails stop at the deadlines`
- Full `@maka/runtime-host` suite: 935 pass.

### DIFF
--- a/packages/runtime-host/src/__tests__/host-kernel.test.ts
+++ b/packages/runtime-host/src/__tests__/host-kernel.test.ts
@@ -752,6 +752,40 @@ describe('non-serving Runtime Host kernel', () => {
     });
   });
 
+  test('a silent handshake defers the never-connected drain instead of being drained under it', async () => {
+    await withHostPaths(async (paths) => {
+      const capability = await resolveStorageRoot({ path: paths.root, kind: 'interactive' });
+      const owner = await tryAcquireInteractiveRootOwner(capability);
+      assert.ok(owner);
+      const host = await RuntimeHostKernel.start({
+        owner,
+        initialConnectionTimeoutMs: 500,
+        idleGraceMs: 10_000,
+        handshakeTimeoutMs: 2_000,
+        composition: defineInteractiveRuntimeHostComposition(async () => testComposition()),
+      });
+
+      const silent = await openSocket(host.endpoint);
+      try {
+        // Past the initial connection timeout but inside the handshake
+        // deferral window: the pending handshake must keep the Host alive.
+        await new Promise((resolve) => setTimeout(resolve, 1_000));
+        assert.equal(host.state, 'ready');
+      } finally {
+        silent.destroy();
+      }
+      // With the handshake gone, the deferred deadline drains the Host.
+      await withTimeout(
+        host.closed,
+        5_000,
+        'silently-handshaked ephemeral candidate never drained after the deferral',
+      );
+      const successor = await tryAcquireInteractiveRootOwner(capability);
+      assert.ok(successor);
+      await successor.close();
+    });
+  });
+
   test('drain requested before factory completion begins drain before recovery exactly once', async () => {
     await withHostPaths(async (paths) => {
       const capability = await resolveStorageRoot({ path: paths.root, kind: 'interactive' });
--- a/packages/runtime-host/src/__tests__/host-kernel.test.ts
+++ b/packages/runtime-host/src/__tests__/host-kernel.test.ts
@@ -688,6 +688,70 @@ describe('non-serving Runtime Host kernel', () => {
     });
   });
 
+  test('never-connected ephemeral candidate drains after the initial connection timeout despite a boot residency', async () => {
+    await withHostPaths(async (paths) => {
+      const capability = await resolveStorageRoot({ path: paths.root, kind: 'interactive' });
+      const owner = await tryAcquireInteractiveRootOwner(capability);
+      assert.ok(owner);
+      let releaseBootWork: (() => void) | undefined;
+      const host = await RuntimeHostKernel.start({
+        owner,
+        initialConnectionTimeoutMs: 100,
+        idleGraceMs: 10_000,
+        composition: defineInteractiveRuntimeHostComposition(async (context) => {
+          const residency = context.acquireResidency('boot-work');
+          releaseBootWork = () => residency.release();
+          return testComposition({
+            beginDrain: () => releaseBootWork?.(),
+          });
+        }),
+      });
+
+      await withTimeout(
+        host.closed,
+        2_000,
+        'ephemeral candidate outlived its initial connection timeout',
+      );
+      const successor = await tryAcquireInteractiveRootOwner(capability);
+      assert.ok(successor);
+      await successor.close();
+    });
+  });
+
+  test('never-connected ephemeral candidate with a hung composition startup fails stop at the deadlines', async () => {
+    await withHostPaths(async (paths) => {
+      const capability = await resolveStorageRoot({ path: paths.root, kind: 'interactive' });
+      const owner = await tryAcquireInteractiveRootOwner(capability);
+      assert.ok(owner);
+      const hostTask = RuntimeHostKernel.start({
+        owner,
+        initialConnectionTimeoutMs: 100,
+        idleGraceMs: 10_000,
+        shutdownGraceMs: 100,
+        composition: defineInteractiveRuntimeHostComposition(() => new Promise(() => {})),
+      });
+
+      try {
+        const error = await withTimeout(
+          hostTask.then(
+            () => assert.fail('Runtime Host startup unexpectedly succeeded'),
+            (startupError: unknown) => startupError,
+          ),
+          2_000,
+          'hung composition startup was not bounded by the initial connection deadline',
+        );
+        assert.ok(error instanceof AggregateError);
+        assert.ok(
+          error.errors.some(
+            (candidate: unknown) => candidate instanceof RuntimeHostProcessTerminationRequiredError,
+          ),
+        );
+      } finally {
+        await owner.close();
+      }
+    });
+  });
+
   test('drain requested before factory completion begins drain before recovery exactly once', async () => {
     await withHostPaths(async (paths) => {
       const capability = await resolveStorageRoot({ path: paths.root, kind: 'interactive' });
--- a/packages/runtime-host/src/server/host-kernel.ts
+++ b/packages/runtime-host/src/server/host-kernel.ts
@@ -68,6 +68,7 @@ const DEFAULT_HANDSHAKE_TIMEOUT_MS = 5_000;
 const DEFAULT_SHUTDOWN_GRACE_MS = 10_000;
 const SHUTDOWN_HANDSHAKE_GRACE_MS = 1_000;
 const SHUTDOWN_OPERATION_GRACE_MS = 1_000;
+const INITIAL_CONNECTION_DEADLINE_DEFERRAL_LIMIT = 3;
 const HOST_PROTOCOL = {
   min: RUNTIME_HOST_PROTOCOL_VERSION,
   max: RUNTIME_HOST_PROTOCOL_VERSION,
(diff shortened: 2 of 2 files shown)


Astro-Han (commented):
Thanks for fixing the unbounded lifetime of never-connected ephemeral candidates. The kernel is the right owner for this deadline, and the two new tests cover the boot-residency and hung-startup paths well.

I found one lifecycle overlap that should be resolved before merge:

**[P2] Keep the pre-connection lifetime under one timer authority**

After the composition reaches `ready`, `#scheduleIdleIfNeeded()` still arms the existing idle timer with `initialConnectionTimeoutMs` while no connection has yet been accepted. The new initial-connection deadline is already governing that same period.

This defeats the new handshake deferral in the ordinary no-residency path:

1. The initial-connection deadline fires while a handshake is in progress and defers by `handshakeTimeoutMs`.
2. The old idle timer then fires independently.
3. `#isTrueIdle()` does not account for handshaking transports, so it drains the Host underneath the connecting Client.

The new tests do not expose this because their boot residency or hung startup prevents the ready-state idle timer from being armed.

The smallest complete fix is to make `#scheduleIdleIfNeeded()` return until the first connection has been accepted, then use only `idleGraceMs`. That leaves one authority per lifecycle phase:

- before the first accepted connection: `#initialConnectionDeadline`, including bounded handshake deferral;
- after the first accepted connection: `#idleTimer` with `idleGraceMs`.

Please add a regression test with a ready, residency-free Host and a silent handshake transport, verifying that it is not drained before the deferred deadline. I would keep the two existing tests; they exercise distinct and useful shutdown paths.

<details>
<summary>中文对照</summary>

感谢修复从未连接的 ephemeral candidate 可能无限存活的问题。deadline 放在 Runtime Host kernel 是正确的，新增的两条测试也分别覆盖了 boot residency 和启动永久挂起。

这里还有一个应在合并前解决的生命周期重叠：

**[P2] 首次连接前只保留一个定时器 authority**

composition 进入 `ready` 后，`#scheduleIdleIfNeeded()` 仍会在尚未接受连接时用 `initialConnectionTimeoutMs` 启动旧 idle timer；新增的 initial-connection deadline 已经在管理同一个阶段。

这会让新增的 handshake 延期在普通、无 residency 的路径上失效：新 deadline 因正在握手而延期后，旧 idle timer 仍会独立触发；而 `#isTrueIdle()` 不检查 handshaking transport，于是会在 Client 正在连接时 drain Host。

最小完整修复是：首次连接成功前不启动 idle timer；首次连接后才由 `idleGraceMs` 管理空闲退出。这样首次连接前后各自只有一个生命周期 owner。

建议补一条 ready、无 residency、存在静默 handshake 的回归测试。现有两条测试覆盖不同的 shutdown 路径，应当保留。

</details>

Disclosure: I used Codex reviewers and Claude Opus to assist with lifecycle tracing and adversarial checks. I reviewed the evidence and own this feedback.

UncertaintyDeterminesYou4ndMe:
Addressed in 1cac1843a. `#scheduleIdleIfNeeded()` now returns until the first connection has been accepted and always uses `idleGraceMs`, so each lifecycle phase has exactly one timer authority as you described: `#initialConnectionDeadline` (with its bounded handshake deferral) before the first accepted connection, `#idleTimer` after.

Added the requested regression test (`a silent handshake defers the never-connected drain instead of being drained under it`): a ready, residency-free Host with a silent handshake transport is still `ready` past `initialConnectionTimeoutMs`, and drains only after the deferral once the transport goes away. Verified it fails on the previous commit (drained at ~500ms under the pending handshake) and passes now. Both earlier tests kept; full `@maka/runtime-host` suite: 936 pass.

Astro-Han (approved):
Thanks for the follow-up. I rechecked `1cac1843a` against the current `main`.

The previous P2 is resolved: before the first accepted connection, `#initialConnectionDeadline` is now the sole lifetime authority; after that transition, `#idleTimer` owns the normal idle grace period. The silent-handshake regression exercises the overlap that prompted the earlier feedback.

I found no remaining P0–P3 issues. The three new tests cover distinct lifecycle paths and are worth keeping; I do not see code or test deletion, or a refactor, that would make this materially simpler without weakening the invariant.

I also verified the automatic merge result with the current `main`: the build completed successfully and `host-kernel.test.ts` passed 45/45. From the code-review perspective, this is a go.

One delivery item before merge: please ensure the final squash commit carries the required `Generated-by: Claude Code` trailer. The PR body discloses the tooling, but the current commits do not contain the trailer.

<details>
<summary>中文对照</summary>

我重新检查了 `1cac1843a`，并与当前 `main` 的自动合并结果一起做了验证。

之前的 P2 已解决：首次连接成功前只有 `#initialConnectionDeadline` 管理生命周期；连接成功后才由 `#idleTimer` 管理正常空闲退出。新增的 silent-handshake 测试也准确覆盖了之前指出的竞态。

没有发现剩余的 P0–P3。三条新增测试覆盖不同生命周期路径，均应保留；目前也没有可以删除或通过重构进一步简化、同时又不削弱生命周期约束的代码。

当前 `main` 与该 PR 的合并结果构建成功，`host-kernel.test.ts` 45/45 通过。从代码审查角度可以继续。

合并前还有一项交付要求：最终 squash commit 需要包含 `Generated-by: Claude Code` trailer。PR 正文已有工具披露，但当前两个提交都没有该 trailer。

</details>

Disclosure: Codex assisted this re-review with lifecycle tracing, current-main merge validation, and focused test execution. This does not replace the repository’s required independent human review.

PR #1357, opened 2026-07-22
feat(runtime-host): establish message authority foundation

M4n5ter:
<details open>
<summary><strong>English</strong></summary>

## Context

This PR establishes the Host-owned Message authority slice for the Runtime Host Session-core migration tracked in #1167.

The Runtime Host remains non-serving in production. This PR does not switch any Desktop, TUI, CLI, or Headless entrypoint.

### Canonical content and durable admission

- Define one canonical `MessageContent` across Core, Runtime, Storage, and the Host wire: model-facing `text`, optional human-facing `displayText`, ordered attachment references, and ordered structured quotes. Attachment bytes remain outside this contract.
- Apply exact decoding, normalization, copying, and equality at protocol and durable boundaries, preserving model input, UI provenance, and retry identity.
- Persist each root-Turn admission with its ordered source messages, placement, disposition, normalized aggregate, and execution identity.
- Preflight every source-message proof before publishing the admission, so a conflicting identity cannot leave a poisoned admission behind.
- Keep the admission record at the sole current schema version (`v1`). No migration path is added because no production Runtime Host state exists.

### Host-owned queue and control authority

- Add the closed operations `turn.message.submit`, `queue.retract`, and `turn.interrupt`, with exact inputs, outputs, typed failures, and bounded projections.
- Make one `HostMessageCoordinator` own Host-Epoch queue revision, steering/follow-up entries, Runtime leases, retry state, and residency.
- Same-Epoch retries are idempotent only for the same canonical payload. Reusing an identity with different content fails with `operation_conflict`.
- Old-Epoch retries resolve only from immutable keyed durable proofs. If the exact result, including its revision, cannot be proven, they return `outcome_unknown`.
- An idle submission starts exactly one root Turn. Active submissions become steering or follow-up entries; retract removes only entries that have not crossed the pull cut; interrupt commits one stop fence and returns the canonical terminal Turn result.
- Preflight immutable steering identity before appending the Runtime ledger. Recovery alone repairs the narrow crash window between durable append and proof publication; the normal path does not scan Session history.

### Root and linked-child lifecycle

- Route hosted linked-child initial runs, resumes, provider retries, and stops through the same root admission and Message authority used by external roots.
- Hold a per-child execution gate from preparation through terminal projection. Resume and retry share one successor rule, while an exact duplicate initial spawn joins the existing in-flight operation.
- Recover a pending foreground child admission after restart by closing it canonically as `app_restarted`; embedded execution remains unchanged.
[...]

PR #3032, opened 2026-08-15
fix(desktop): queue a busy-raced send as steering instead of dropping it

### LINKED ISSUE #1954: fix(desktop): mid-turn send opens a parallel root turn on the embedded runtime — should route through the steering queue
## Symptom

In Maka Desktop, sending a new message while the current session turn is still running:

1. The chat view jitters (the live-turn projection flips between two turns).
2. The session reports an error (conversation error toast).

## Root cause

Production Desktop still runs the embedded SessionManager path, which lacks the submit semantics established by #1357:

- `sessions:send` → `runtimeKernel.sendMessage` → `startTurn` opens a new top-level turn unconditionally (no active-root guard).
- `ensureActive` reuses the same backend generation; `AiSdkBackend` is a single mutable-state instance (`aborted`, `currentTurnId`, `currentRunId`, `abortController`, `injectedSteeringMessages`, ...). Two concurrent `send()` calls stomp these fields, so stop / respond / sandbox-boundary route to the wrong turn and one run fails → a streamed `error` event → session error.
- The renderer arms `liveTurn` for the new turn (T_B), but every `text_delta` from the still-running T_A flips the projection back; persisted message order interleaves (T_B's user message lands before T_A's; assistant texts land in completion order), so full-list refreshes reorder content and scroll jumps.

CLI (#1087) and Desktop voice (`sessions:steer`) already route mid-turn text through the steering queue; only Desktop text input opens a parallel turn.

## The final shape already exists

#1357 (merged) defines `turn.message.submit`: "An idle submission starts exactly one root Turn; active submissions become steering or follow-up entries." This bug is the embedded production path not yet applying that semantics. #853 (RFC) / #1357 (Message authority) / #1167 (M3 tracking) remain the long-term home; the M5 production cutover deletes the embedded path.

## Proposed fix (short term, aligns with #1357)

1. Desktop main `sessions:send` asks the embedded runtime steering queue first (`runtimeKernel.steer` — the embedded counterpart of #1357's queue): `{ kind: 'queued' }` injects into the running turn; `{ kind: 'fallback' }` falls through to the existing new-turn `sendMessage`. Converge this decision in main so chat / bot / voice-fallback / goal entrypoints share one routing point (and return `steered` so the renderer skips new-turn bookkeeping: `armTurnActive`, optimistic status, `refreshMessagesUntilTurn`).
2. Renderer `handleEvent` learns `steering_message` → `refreshMessages` (currently falls into the default branch), so the injected text appears in the transcript (it is already persisted as a `role: 'user'` RuntimeEvent with `steering: true`).

## Verification

- Runtime repro (fake backend + MemoryAgentRunStore): two concurrent `sendMessage` calls on one session must not parallel-run or interleave message order.
- Desktop E2E (fake backend): send while running → no second turn, steering text visible, no error.

UncertaintyDeterminesYou4ndMe:
## Summary

Fixes #1954 (the slice that survived the M5 cutover).

Since the production cutover, `turn.start` carries an active-root guard, so the parallel-turn/interleaving symptoms in #1954 no longer occur, and the composer already routes text at a session it sees as running to `sessions:steer` (the button flips to 插入消息). What remained: that reading is renderer-local state. Another window, a Bot, or a Goal continuation can open the root Turn first, and the raced `sessions:send` then failed with `session_busy` — surfaced as a generic send-failure toast with the user's message dropped.

What changed:

- Desktop main's `sessions:send` falls back to `turn.message.submit` (placement `current_turn`) when `turn.start` reports `session_busy`. The Host resolves the race atomically per #1357's submit semantics: an active session queues the text as steering into the running turn; a session that went idle in between starts the turn (`turn_started`).
- Skill (`skillIds`), `turnOrchestration`, and canonical `/skill:` token sends rethrow the busy error instead — their turn semantics cannot be expressed as a queued message.
- The renderer settles both send branches through one `settleSendBookkeeping` helper: it skips new-turn bookkeeping for a `steered` result (the steering_message event renders the text in the transcript, same as the composer steer path) and re-keys only the exact optimistic state this send created when the fallback started the turn under a Host-chosen id, preserving an authoritative projection that arrived first.
- A steered send also emits a sessions-changed nudge so the stale renderer converges on the running turn.

Attachments survive the fallback: the steering injection path (`drainSteeringInto` → `appendImageParts`) carries `MessageContent.attachments`.

## Verification

- New handler tests: steering fallback, `turn_started` fallback, and busy preserved for Skill sends (both the `skillIds` and `/skill:` text forms).
- New renderer settlement tests (`app-shell-busy-race-settlement.test.ts`): steering into an active turn, Host-id rebind, authoritative-projection-first preserved, and both new-chat outcomes without ghost optimistic turns.
- Full `@maka/desktop` main suite: 848 pass; desktop typecheck (preload/main/renderer/storybook) clean.
- The earlier `streaming-remount.spec.ts` e2e failure was bisected to a pre-existing virtualization regression, filed as #3044, fixed by #3048; this branch is rebased past that fix.

## AI use

Select exactly one:

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/src/main/__tests__/app-shell-busy-race-settlement.test.ts
+++ b/apps/desktop/src/main/__tests__/app-shell-busy-race-settlement.test.ts
@@ -1,0 +1,284 @@
+/**
+ * #1954 busy-race settlement: a sessions:send that raced a root turn another
+ * client opened can come back `steered` (the send owns no turn) or under a
+ * Host-chosen turnId. Both results must be interpreted identically by the
+ * new-chat and existing-session branches, and a rebind must never overwrite
+ * an authoritative live projection that beat the IPC response.
+ */
+
+import { strict as assert } from 'node:assert';
+import { describe, it } from 'node:test';
+
+import type { StoredMessage } from '@maka/core/session';
+import type { LiveTurnProjection } from '@maka/ui';
+import { createAppShellChatActions } from '../../renderer/app-shell-chat-actions.js';
+
+function installWindow(maka: unknown): () => void {
+  const target = globalThis as unknown as { window?: unknown };
+  const hadWindow = Object.prototype.hasOwnProperty.call(target, 'window');
+  const previousWindow = target.window;
+  Object.defineProperty(target, 'window', {
+    configurable: true,
+    value: { maka },
+    writable: true,
+  });
+  return () => {
+    if (hadWindow) {
+      Object.defineProperty(target, 'window', {
+        configurable: true,
+        value: previousWindow,
+        writable: true,
+      });
+    } else {
+      delete target.window;
+    }
+  };
+}
+
+function createTurnState() {
+  const liveTurnBySession: Record<string, LiveTurnProjection> = {};
+  return {
+    liveTurnBySession,
+    setLiveTurnBySession(
+      updater: (c: Record<string, LiveTurnProjection>) => Record<string, LiveTurnProjection>,
+    ) {
+      const next = updater({ ...liveTurnBySession });
+      for (const key of Object.keys(liveTurnBySession)) delete liveTurnBySession[key];
+      Object.assign(liveTurnBySession, next);
+    },
+  };
+}
+
+function createMessageState() {
+  const messages: StoredMessage[] = [];
+  return {
+    messages,
+    setMessages(updater: StoredMessage[] | ((current: StoredMessage[]) => StoredMessage[])) {
+      const next = typeof updater === 'function' ? updater([...messages]) : updater;
+      messages.length = 0;
+      messages.push(...next);
+    },
+  };
+}
+
+function createActionsDeps() {
+  return {
+    uiLocale: 'en' as const,
+    activeIdRef: { current: undefined as string | undefined },
+    addPendingSessionAction: () => true,
+    captureComposerImportOwner: () => ({
+      sessionId: undefined,
+      navSection: 'sessions' as const,
+    }),
+    checkTaskSubmissionReadiness: async () => true,
+    clearPendingSessionAction: () => undefined,
+    isNewChatSendSurfaceActive: () => true,
+    isShellSurfaceOwnerActive: () => true,
+    markSessionReadLocally: () => undefined,
+    messageRetryPendingRef: { current: new Set<string>() },
+    refreshSessions: async () => [],
+    setActiveId: () => undefined,
+    setMessageLoadErrorBySession: () => undefined,
+    setMessageRetryPendingBySession: () => undefined,
+    setMessages: () => undefined,
+    transcriptRangeRef: { current: undefined },
+    setNavSelection: () => undefined,
+    setLiveTurnBySession: () => undefined,
+    setInteractionBySession: () => undefined,
+    showModelSetupToast: () => undefined,
+    toastApi: { error: () => undefined, info: () => undefined },
+    upsertSessionSummary: () => undefined,
+    newChatModel: null,
+    pendingNewChatThinkingLevel: null,
+    newChatCollaborationMode: 'agent' as const,
+    newChatOrchestrationMode: 'default' as const,
+    newChatProjectId: undefined,
+  };
+}
+
+const EMPTY_SKILL_INVOCATION = { loaded: [], failed: [], receipts: [] };
+
+describe('busy-raced send settlement', () => {
+  it('a steered send on an existing session disarms its turn and shows no optimistic message', async () => {
+    const activeIdRef = { current: 'session-a' as string | undefined };
+    const turnState = createTurnState();
+    const messageState = createMessageState();
+    const restoreWindow = installWindow({
+      sessions: {
+        send: async (_sessionId: string, command: { turnId: string }) => ({
+          ok: true,
+          steered: true,
+          turnId: command.turnId,
+          attachments: [],
+          inlineReferences: [],
+          skillInvocation: EMPTY_SKILL_INVOCATION,
+        }),
+      },
+    });
+    try {
+      const actions = createAppShellChatActions({
+        ...createActionsDeps(),
+        activeIdRef,
+        setLiveTurnBySession: turnState.setLiveTurnBySession,
+        setMessages: messageState.setMessages,
+      });
+      assert.equal(await actions.send('also check the tests'), true);
+      assert.equal(turnState.liveTurnBySession['session-a'], undefined);
+      assert.deepEqual(messageState.messages, []);
+    } finally {
+      restoreWindow();
+    }
+  });
+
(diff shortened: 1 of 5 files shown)


Astro-Han (commented):
Thanks for moving the final busy-race decision back to `turn.message.submit`. Runtime Host remains the sole Turn/steering authority, and the main-process fallback is narrowly limited to `session_busy`.

I found two gaps that should be addressed before merge.

**[P2] Rebind optimistic state without overwriting an authoritative live projection**

When the fallback returns a Host-generated `turnId`, the renderer currently performs `disarmTurnActive(oldId)` followed by `armTurnActive(hostId)`.

Host events may arrive before the IPC response because main emits the status change before returning. If the renderer has already received streamed or terminal state under `hostId`, the subsequent arm replaces that authoritative projection with a fresh `waiting + unconfirmed` projection. This can erase visible progress or leave the processing/Stop state stuck indefinitely.

Please use a single conditional rebind instead: rename only the exact `unconfirmed + waiting` arm created by this send. If an authoritative projection has already arrived, preserve it. The earlier #1954 review implementation had this `rebindTurnActive` behavior.

The new-chat branch should consume the same result normalization. It currently ignores both `steered` and a Host-chosen `turnId`, so a create/send race can leave a ghost optimistic Turn or associate the message with the wrong identity. The clean solution is one shared settlement helper for both new and existing Sessions.

Please cover at least:

- a streamed or terminal Host event arriving before the send response;
- `steered` on the new-chat path;
- `turn_started` with a Host-generated ID on the new-chat path.

**[P2] Test the real Desktop Skill path**

The fallback excludes `command.skillIds`, but the Desktop Composer represents Skills in the canonical text grammar as `/skill:<id>` tokens; it does not send `skillIds`. The new Skill test therefore exercises a protocol branch without a production renderer caller.

In the real path, a busy-raced `/skill:` send still enters `turn.message.submit`. A blocked invocation then becomes a generic operation error, while partial Skill-loading feedback can be lost because the submit result does not carry `skillInvocation`.

For the stated behavior—Skill sends retain their existing semantics—the minimal fix is to recognize canonical Skill tokens in `command.text` and keep them out of this fallback, then replace or supplement the current test with a `/skill:<id>` input. A larger future design could carry structured Skill feedback through `turn.message.submit`, but that is not required for this PR.

More generally, a discriminated result such as `started(turnId) | steered | blocked(...)` would be cleaner than `{ turnId, steered?: true }`: it removes the ghost-turn state and prevents the two renderer branches from interpreting the contract differently.

The failing `streaming-remount` E2E calls `sessions.steer` directly and does not enter this fallback, so I do not consider it a production finding from this change. The check should still be rerun successfully before merge.

<details>
<summary>中文对照</summary>

感谢把 busy race 的最终判断交还给 `turn.message.submit`。Runtime Host 仍然是 Turn/steering 的唯一 authority，main 也只在 `session_busy` 时进入 fallback，整体方向正确。

合并前建议解决两个问题：

**[P2] 重绑定 optimistic state 时不能覆盖已经到达的权威 projection**

Host event 可能早于 IPC response 到达。当前的 `disarm(oldId) → arm(hostId)` 会把已经存在的 streamed 或 terminal projection 覆盖成新的 `waiting + unconfirmed`，可能清空可见进度，或者让 processing/Stop 状态永久无法回收。

应该只原子地重命名本次 send 创建的 `unconfirmed + waiting` arm；如果 Host projection 已经到达，则保持不变。新建聊天与已有 Session 也应该共享同一套结果归并逻辑，因为新建聊天路径目前没有处理 `steered` 和 Host 返回的新 `turnId`。

**[P2] Skill 测试没有命中 Desktop 的真实路径**

Desktop Composer 使用正文 `/skill:<id>` token，而不是 `skillIds`。当前守卫和测试因此只覆盖了没有生产 renderer caller 的协议分支。真实 Skill 在竞态时仍会进入 fallback，blocked feedback 会退化成通用错误，部分加载失败的结构化反馈也可能丢失。

最小修复是同时识别正文中的 canonical Skill token，让它保留原有 `turn.start` 语义，并用 `/skill:<id>` 补充或替换当前测试。让 `turn.message.submit` 正式携带 Skill feedback 可以作为后续更完整的设计。

`streaming-remount` E2E 直接调用 `sessions.steer`，没有经过本 PR 的 fallback，因此不属于本次生产 finding；但合并前仍应让 CI 重跑至绿色。

</details>

Disclosure: I used Codex reviewers and Claude Opus to assist with call-path tracing and adversarial checks. I reviewed the evidence and own this feedback.

UncertaintyDeterminesYou4ndMe:
Both points addressed in 2b7b2eb5f.

**Settlement**: both send branches now consume one `settleSendBookkeeping` helper. The rebind is conditional as you described — it renames only the exact `unconfirmed + waiting` arm this send created, so an authoritative projection that beat the IPC response is preserved. The new-chat branch understands both `steered` (navigate to the session, no ghost optimistic turn) and a Host-chosen `turnId`. Covered by a new test file (`app-shell-busy-race-settlement.test.ts`): authoritative-projection-before-response, steered on new-chat, and `turn_started` with a Host id on new-chat, plus the two existing-session cases.

**Skill path**: the fallback now recognizes canonical `/skill:` tokens via `SKILL_INVOCATION_TOKEN_SOURCE` in `command.text` and keeps the busy error for those sends; the handler test now covers the `/skill:review …` text form alongside the protocol-level `skillIds` variant.

Agreed on the discriminated `started | steered | blocked` result being the cleaner contract — with both branches now behind one settlement helper the migration is mechanical, so I'd propose it as a follow-up rather than widening this fix.

Desktop main suite: 848 pass; typecheck clean. Will rerun the `streaming-remount` e2e check.

UncertaintyDeterminesYou4ndMe:
The failing `streaming-remount.spec.ts:65` e2e (assertion at line 153, settled-restore typewriter check) is a pre-existing flake: the identical test and assertion failed on `main` in [run 31791670100](https://github.com/maka-agent/maka-agent/actions/runs/31791670100) (2026-08-14), before this PR existed. It also drives `sessions.steer` directly and its send goes through the unchanged ordinary path, so it never enters this PR's busy fallback. I lack rerun rights on this repo, so I pushed an empty commit to retrigger the checks.

UncertaintyDeterminesYou4ndMe:
Filed the flaky e2e as #3044 with a bisect: it is a regression from #2975 (virtualize long chat transcripts), reproducible on `main` at 3/4 locally, unrelated to this PR's change.

UncertaintyDeterminesYou4ndMe:
Rebased onto `main` now that #3048 (the fix for #3044) is merged — this should clear the `streaming-remount` e2e flake that was the only red check here. No code changes in the rebase.

Generated-by: Claude Code

M4n5ter:
LGTM !

Issue #3029, opened 2026-08-15
fix(runtime): history compaction accepts truncated/unstructured LLM summaries without validation


me2seeks:
## Problem

A history compaction wrote a checkpoint whose LLM summary was a 138-token free-form fragment that (a) ignored the mandated structured format, (b) ended mid-sentence, and (c) omitted all load-bearing context (which files were changed, current worktree state, test status). The continuing model then confabulated implementation details (it described a TypeScript CLI change as non-existent Rust files), which looked like foreign content leaking into the session.

Incident evidence (session fbdb3fd3, checkpoint hcheckpoint-981ceab88738d4819971747aeb292203):

- Compaction folded **742 RuntimeEvents / ~235,503 estimated tokens into a 138-token summary** (~1700:1).
- The summary text: `确认服务端语义后，决定：…现在看 desktop 的 retry 循环结尾：` — free-form, ends with a colon mid-sentence, and follows none of the `SUMMARIZATION_SYSTEM_PROMPT` sections (`## Goal / ## Progress / ## Key Decisions / ## Next Steps / ## Critical Context`).
- The run ledger shows **3 failedOpen attempts with `failOpenReason: provider_error`** immediately before the write that succeeded — the accepted response was produced right after repeated provider failures and is consistent with a degraded/truncated provider completion.

## Root cause

`history-compact-summarizer.ts` + `ai-sdk-compaction.ts` accept any non-empty string as a checkpoint summary:

1. The only rejection gate is `empty_summary` (`!summary?.trim()`). There is no structural validation (required sections), no minimum-size sanity check relative to the folded span, and no truncated-output detection beyond `finishReason === 'length'` — which does not catch providers that return partial text with a `stop` finish reason (observed with the session's own model serving as summarizer).
2. The summarizer reuses the session model (`resolveModel()`). A weak instruction-follower can silently produce a free-form note instead of the structured checkpoint, and the pipeline records it as a valid checkpoint with `decision: 'replaced'`, replacing 235k tokens of history.

No cross-session data is involved in this incident: `buildHistoryCompactCheckpoint` rejects multi-session event sets, and the checkpoint's coverage references this session's own event projection. The "foreign" content was reference material deliberately fetched into the session (a codex-rs source study requested by the user) that the thin summary failed to contextualize, leaving the continuation model to hallucinate around it.

## Proposal

[...]

PR #2993, opened 2026-08-14
feat(runtime): add Codex server-side history compaction

M4n5ter:
<details open>
<summary><strong>English</strong></summary>

## Summary

Add provider-native server-side history compaction for OpenAI Codex subscription models and enable it by default for those models.

The implementation keeps the existing text-summary checkpoint for portable compaction and adds a distinct opaque provider-state checkpoint for Codex. The runtime owns the common checkpoint lifecycle—coverage, persistence, replay, rolling replacement, overflow recovery, and model-switch compatibility—while the Codex-specific compactor owns only the provider protocol and request options.

Key behavior:

- configure the Codex remote compactor by default in Runtime Host model composition;
- set `providerOptions.openai.compactionTrigger` only on the dedicated compactor request, leaving ordinary model requests unchanged;
- persist and replay native Codex compaction state without fake summaries or synthetic transcript markers;
- preserve the selected provider checkpoint through empty tails and text-only replay fallbacks;
- lower settled provider-executed tool history into paired Responses function-call items so `store: false` never produces dangling outputs;
- redact only the opaque compaction payload in provider-request telemetry while preserving unrelated URLs, metadata, and business JSON;
- retain existing text summarization behavior for non-Codex providers and safely recompress full source history after incompatible model/provider switches.

The abstraction is intentionally narrow: one discriminated checkpoint representation and one compactor contract, without introducing a provider registry or a generic transport framework before another integration requires one.

## Upstream context

This work exposed a missing typed capability in the Vercel AI SDK. M4n5ter reported [vercel/ai#18894](https://github.com/vercel/ai/issues/18894), which directly prompted [vercel/ai#18895](https://github.com/vercel/ai/pull/18895). The upstream change has since been merged and released; Maka now uses the released `providerOptions.openai.compactionTrigger` support directly, with no compatibility transport layer.

## Verification

- clean `npm ci`, including dependency patch application;
- `npm run format:check`;
- `npm run build`;
- `npm run typecheck`;
- `npm test` — 6,671 tests: 6,645 passed, 0 failed, 26 skipped;
- `git diff --check`;
- real Codex OAuth verification with `gpt-5.6-luna`: initial compaction, rolling compaction from the previous opaque state, and ordinary provider-state replay all succeeded. The two compaction requests each carried exactly one terminal trigger; the replay request carried no trigger and recovered both facts from the compacted history.

## Checklist

- [x] Tests cover the change and its regression cases.
- [x] Format, typecheck, build, and the full test suite pass locally.
[...]

PR #3038, opened 2026-08-15
fix(runtime): group parallel tool calls into one summarizer assistant message

### LINKED ISSUE #3030: fix(runtime): history summarizer emits parallel tool calls as separate assistant messages, rejected by strict OpenAI-compatible providers
## Problem

`replayPlanItemsToModelMessages` in `packages/runtime/src/history-compact-summarizer.ts` emits every `tool_call` replay item as its own assistant message. When a step contains parallel tool calls, the resulting message list is:

```
assistant [tool-call A]
assistant [tool-call B]   ← invalid: previous assistant message's tool_calls not yet answered
tool      [tool-result A]
tool      [tool-result B]
```

Strict OpenAI-compatible providers reject this. DeepSeek responds 400:

```
An assistant message with 'tool_calls' must be followed by tool messages
responding to each 'tool_call_id'. (insufficient tool messages following tool_calls message)
```

The history summarizer then throws `HistoryCompactSummarizerError('provider_error')` and compaction fails open — so for sessions using such providers, mid-turn history compaction can never succeed (the incident behind #3029 shows 3 consecutive `provider_error` fail-opens before a degraded write went through; the folded events there contained parallel tool calls throughout).

## Root cause

The primary replay path's materializer merges a step's reasoning and tool calls into one provider message (documented in `model-history.ts`: "the materializer then merges the step's reasoning + tool calls into one provider [message]"). The summarizer bypasses the materializer and maps items 1:1 to messages, diverging from that invariant.

## Proposal

Group consecutive `tool_call` items into a single assistant message (mirroring the step-merge invariant), keeping each `tool_result` as its own following tool message. Add a regression test with parallel calls asserting one assistant message with N tool-call parts followed by N tool messages.

## Acceptance

- Summarizer requests over history containing parallel tool calls are accepted by strict OpenAI-compatible providers (covered by a message-shape unit test; verified live against DeepSeek during reproduction).
- No change to the primary replay path (already correct).

Generated-by: Maka

UncertaintyDeterminesYou4ndMe:
## Summary

Fixes #3030.

`replayPlanItemsToModelMessages` mapped every `tool_call` replay item to its own assistant message, so a step with parallel tool calls produced

```
assistant [tool-call A]
assistant [tool-call B]   ← rejected: A's tool_calls not yet answered
tool      [tool-result A]
tool      [tool-result B]
```

Strict OpenAI-compatible providers 400 on that shape, so mid-turn history compaction could never succeed on such providers — the `provider_error` fail-open loop that preceded the #3029 incident.

What changed: grouping is step-based with results buffered to the step boundary, mirroring the step-merge invariant the primary replay path gets from the materializer (`model-history.ts`). Membership follows the stamped `stepId` when both sides carry one; legacy items without a `stepId` join while the open step still has unsettled calls, and a call arriving after the open step fully settled opens its own message. Results for calls in the open step are deferred and emitted in arrival order at the step boundary, so interleaved orderings like `call A, call B, result A, call C, result B, result C` materialize as `assistant [A, B, C] → tool A → tool B → tool C`. Skipped `thinking` items do not break a step. The primary replay path is untouched.

## Verification

- Regression tests: parallel calls produce exactly one assistant message with the tool-call parts followed by the tool messages; the interleaved ordering from the primary materializer's fixture; distinct settled steps never merged; stamped step ids deciding membership over settledness.
- Full `@maka/runtime` suite passes locally (re-verified after rebasing onto `main` past #2993).
- Cross-checked against the #3029 incident repro by @me2seeks (review on this PR); the parallel draft in #3040 was withdrawn in favor of this fix.

## AI use

Select exactly one:

## Checklist

- [ ] No

### DIFF
--- a/packages/runtime/src/__tests__/history-compact-summarizer.test.ts
+++ b/packages/runtime/src/__tests__/history-compact-summarizer.test.ts
@@ -225,6 +225,189 @@ describe('buildLlmHistorySummarizer', () => {
     ]);
   });
 
+  test('keeps a step open across interleaved results until every call is settled', async () => {
+    // The production-legal ordering from the primary replay materializer's
+    // fixture: call A, call B, result A, call C, result B, result C. All
+    // three calls must share one assistant message with the results after it.
+    const seen: Array<{ messages: unknown[] }> = [];
+    const generateText: AiSdkGenerateTextLike = async (opts) => {
+      seen.push(opts);
+      return { text: '## Goal\nX' };
+    };
+    const summarize = buildLlmHistorySummarizer({ resolveModel: () => 'fake-model', generateText });
+
+    const events: RuntimeEvent[] = [
+      ev({ role: 'user', author: 'user', content: { kind: 'text', text: 'inspect files' } }),
+      ev({
+        role: 'model',
+        author: 'agent',
+        content: { kind: 'function_call', id: 'fc1', name: 'read', args: { path: 'a.ts' } },
+      }),
+      ev({
+        role: 'model',
+        author: 'agent',
+        content: { kind: 'function_call', id: 'fc2', name: 'read', args: { path: 'b.ts' } },
+      }),
+      ev({
+        role: 'tool',
+        author: 'tool',
+        content: { kind: 'function_response', id: 'fc1', name: 'read', result: 'A' },
+      }),
+      ev({
+        role: 'model',
+        author: 'agent',
+        content: { kind: 'function_call', id: 'fc3', name: 'glob', args: { pattern: '*' } },
+      }),
+      ev({
+        role: 'tool',
+        author: 'tool',
+        content: { kind: 'function_response', id: 'fc2', name: 'read', result: 'B' },
+      }),
+      ev({
+        role: 'tool',
+        author: 'tool',
+        content: { kind: 'function_response', id: 'fc3', name: 'glob', result: ['a.ts'] },
+      }),
+    ];
+
+    await summarize(inputWith(events));
+
+    const messages = seen[0]!.messages as Array<{
+      role: string;
+      content: Array<{ type: string; toolCallId?: string }>;
+    }>;
+    const shape = messages.map((m) => `${m.role}:${m.content.map((part) => part.type).join('+')}`);
+    assert.deepEqual(shape, [
+      'user:text',
+      'assistant:tool-call+tool-call+tool-call',
+      'tool:tool-result',
+      'tool:tool-result',
+      'tool:tool-result',
+    ]);
+    assert.deepEqual(
+      messages[1]!.content.map((part) => part.toolCallId),
+      ['fc1', 'fc2', 'fc3'],
+    );
+    assert.deepEqual(
+      messages.slice(2).map((m) => m.content[0]!.toolCallId),
+      ['fc1', 'fc2', 'fc3'],
+    );
+  });
+
+  test('does not merge distinct settled steps into one assistant message', async () => {
+    const seen: Array<{ messages: unknown[] }> = [];
+    const generateText: AiSdkGenerateTextLike = async (opts) => {
+      seen.push(opts);
+      return { text: '## Goal\nX' };
+    };
+    const summarize = buildLlmHistorySummarizer({ resolveModel: () => 'fake-model', generateText });
+
+    const events: RuntimeEvent[] = [
+      // Two sequential legacy steps (no stepId): the second call arrives only
+      // after the first is fully settled, so it opens its own message.
+      ev({
+        role: 'model',
+        author: 'agent',
+        content: { kind: 'function_call', id: 'fc1', name: 'read', args: { path: 'a.ts' } },
+      }),
+      ev({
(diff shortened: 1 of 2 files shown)


Astro-Han (commented):
Thanks for fixing the strict-provider failure at the shared summarizer replay seam. Grouping parallel calls here is the right direction, but the current grouping still ends too early when calls and results are interleaved.

The Runtime history can legitimately contain:

```text
call A
call B
result A
call C
result B
result C
```

There is already a production-seam fixture for this ordering in `ai-sdk-backend.test.ts` (`replays interleaved parallel RuntimeEvent tool calls as one provider tool-call block`).

Because `openToolCalls` is cleared on the first result, this PR materializes that history as:

```text
assistant [A, B]
tool [result A]
assistant [C]
tool [result B]
tool [result C]
```

A strict OpenAI-compatible provider can still reject this: the second assistant message appears before every call from the first assistant message has a result.

Could we keep the entire assistant step open and buffer its results until the step boundary, using `stepId` and the set of unsettled calls rather than adjacency alone? The cleanest long-term shape would be to share the primary replay materializer’s step-grouping invariant, so the summarizer does not maintain a second, weaker interpretation of the same history.

The new test is useful; extending it with the existing interleaved ordering, plus a case proving that different steps are not merged, should cover the missing behavior.

<details>
<summary>中文对照</summary>

感谢在共享的 summarizer replay 边界修复严格 provider 的失败问题。在这里合并并行调用的方向是正确的，但当调用和结果交错出现时，当前分组仍会过早结束。

Runtime 历史可能合法地按以下顺序记录：

```text
call A
call B
result A
call C
result B
result C
```

`ai-sdk-backend.test.ts` 已有命中生产调用链的测试覆盖这种顺序：`replays interleaved parallel RuntimeEvent tool calls as one provider tool-call block`。

由于代码在第一个 result 出现时就清空 `openToolCalls`，这个 PR 会把上述历史转换成：

```text
assistant [A, B]
tool [result A]
assistant [C]
tool [result B]
tool [result C]
```

严格的 OpenAI-compatible provider 仍可能拒绝该序列，因为第一条 assistant message 中的全部调用得到结果前，第二条 assistant message 已经出现。

建议保持整个 assistant step 打开并缓存结果，使用 `stepId` 和尚未完成的调用集合判断边界，而不是仅依赖事件是否相邻。长期更干净的方案是复用 primary replay materializer 的 step grouping invariant，避免 summarizer 维护另一套更弱的历史解释。

新增测试是有价值的；建议加入现有的交错顺序，并补充不同 step 不应被合并的用例。

</details>

AI-assisted review disclosure: Codex helped inspect the diff and reproduce the message sequence. I reviewed the source, evidence, and severity and made the final judgment.

UncertaintyDeterminesYou4ndMe:
Addressed in c9d7112a7. Grouping is now step-based with results buffered to the step boundary, as you suggested:

- Membership follows the stamped `stepId` when both sides carry one; legacy items without a `stepId` join while the open step still has **unsettled calls** — which is exactly the interleaving case — and a call arriving after the open step fully settled opens its own message.
- Results for calls in the open step are deferred and emitted in arrival order at the step boundary (text, a foreign result, a different step, or end of items), so the interleaved ordering materializes as `assistant [A, B, C] → tool A → tool B → tool C`.

Tests added per your list: the interleaved ordering from the primary materializer's fixture (`call A, call B, result A, call C, result B, result C` → one assistant block, results in order), distinct settled steps never merged, and stamped step ids deciding membership over settledness (same `stepId` joins across a settled boundary; a different `stepId` never merges).

Agreed the cleanest long-term shape is sharing the primary materializer's step-grouping invariant instead of a second interpretation here — that refactor (extracting the step merge out of the materializer's flush) felt out of proportion for this fix, so I kept it as a mirror with the same semantics and would follow up if you want the extraction.

Full `@maka/runtime` suite: 2835 pass.

me2seeks:
Reviewed against the #3029 incident repro. Step-based grouping with results deferred to the step boundary is the correct invariant and strictly stronger than the adjacency-based draft we had in #3040 (which missed the interleaved-result ordering). The stamped-stepId / legacy-settledness membership logic matches the primary materializer. No blocking issues found.

One non-blocking observation: a cross-step call arriving while a previous step still has unsettled calls (with differing stamped stepIds) would flush the previous step's partial results and open a new assistant message. That ordering is already-malformed replay input that the primary path rejects as well, so it is out of scope for this fix.

FYI: we are removing our own #3030 fix from #3040 and deferring to this PR.

UncertaintyDeterminesYou4ndMe:
Thanks for cross-checking against the incident repro and for deferring #3040's variant. Confirming your non-blocking observation matches what my probe pass found: a cross-step call with a differing stamped stepId arriving while the previous step is unsettled flushes the partial results and opens a new message — that input is already-malformed replay (the primary path rejects it too), and the pre-fix code produced an equally invalid shape for it, so it stays out of scope here.

Generated-by: Claude Code

M4n5ter (approved):
LGTM. I have manually reviewed the final diff and the validation evidence, and I approve this change.

PR #3039, opened 2026-08-15
fix(runtime): validate compaction summaries before they replace history

### LINKED ISSUE #3029: fix(runtime): history compaction accepts truncated/unstructured LLM summaries without validation

UncertaintyDeterminesYou4ndMe:
## Summary

Fixes #3029.

History compaction accepted any non-empty string as a checkpoint summary. In the incident, a degraded provider completion — a 138-token free-form fragment that ignored the mandated sections and ended mid-sentence, delivered with a `stop` finish reason right after three `provider_error` fail-opens — replaced ~235k estimated tokens of history, and the continuation model confabulated around the missing context.

What changed — the validation lives in a dedicated `history-compact-summary-validation.ts`, one module owning the mandated format: the summarization prompt is built from its template, and a writer-agnostic `findCheckpointSummaryDefect(summary, foldContext?)` predicate is consumable by any validation site (e.g. checkpoint load, see #3046). It is enforced at every layer (proposals 1–3 of the issue):

- **Both checkpoint write gates** — the pre-turn writer (`ai-sdk-compaction.ts`) and the mid-turn engine (`mid-turn-capacity-compact.ts`) — call the predicate before building a checkpoint, so the invariant holds for any producer. `buildLlmHistorySummarizer` additionally rejects at generate time for the typed reason on the summarizer path. `buildHistoryCompactCheckpoint` itself is the only authority that assigns the `sections_v1` marker, and it re-runs the complete predicate (including the size floor over the covered span it is handed) before stamping — a direct recorder caller cannot mint the trusted marker for unvalidated text; `legacy_freeform` is the explicit escape for preserving unmarked legacy summaries across copy.
- **Version-aware load/replay quarantine** (load side cherry-picked from @me2seeks, closing #3041): text checkpoints built under this contract carry a durable `summaryFormat: 'sections_v1'` marker, and the ledger's load sites hold marked checkpoints to the complete predicate (minus the size floor, whose covered-span estimate is not durable) — so a malformed summary entering through a direct recorder or copy seam never becomes authoritative after restart. Unmarked legacy checkpoints keep loading when intact; only truncation is quarantined writer-agnostically, via the same shared fence scan and a broadened mid-thought tail heuristic. Rejection repairs the projection from the canonical ledger and recovers the prior valid checkpoint; shape validation fails closed on unknown markers; conversation copy validates marked sources and preserves the source's format identity in the target.
- **Structural**: `## Goal`, `## Progress`, `## Next Steps`, and `## Critical Context` must appear in order, each with non-empty content, recognized only outside fenced code blocks. Critical Context is required because it carries exactly what the incident lost (files, commands, errors); the template gives it an explicit `"(none)"` escape hatch. Verbatim template lines, thematic breaks, and bare fence-marker runs never count as content, and content under a non-required H2 (e.g. `## Key Decisions`) cannot satisfy a required section above it — so a template echo, fenced or not, cannot pass.
- **Truncation heuristics**: a trailing colon (half/full-width) or an unclosed code fence marks output cut mid-sentence even when the provider reported `stop` — the incident's exact shape. Fences follow the Markdown closing rule: a closer must be a bare run of the same family at least as long as the opener, tracked line-opening only, so a verbatim ``` inside a preserved error message is not read as truncation and a four-backtick fence is not closed by a three-backtick line. Lines are split on `/\r?\n/` so CRLF output is handled identically.
- **Size floor**: folding a large span (>~10k estimated tokens, measured over the checkpoint's full covered span at the session's `charsPerToken`, clamped ≥ 1) into less than ~200 estimated tokens of summary is rejected.

A failed validation surfaces a granular typed reason (`malformed_summary_missing_section` / `_truncated` / `_too_small_for_fold`). The existing pipeline turns that into a fail-open — history kept, compaction retried next turn — and the reason lands in `historyCompactWriteSkippedReasonCounts`/`failOpenReason` diagnostics. The empty case stays with the compaction layer's `empty_summary` gate, and `output_length` still wins when the provider reports an exhausted budget.

## Scope notes

- Coordination settled: #3040 and #3046 are closed in favor of this PR as the single summary-validation authority; the load-side work landed here as a cherry-picked commit with @me2seeks's authorship preserved.
- Proposal 4 (deterministic facts appendix) and 5 (bounded validation-failure retry / dedicated summarizer model) are feature-shaped and left as follow-ups; the retry is the mitigation for deterministic rejection of stylistic outliers (e.g. a legitimate summary ending in a colon).
- Rebased past #2993 (Codex provider-native compaction), #3038, and #3113: failure reasons merged as the union, and the validation composes with the opaque-checkpoint filtering and input bounding.

## Verification

- The incident's fragment fails validation and never becomes a checkpoint — end-to-end at the backend seam (real summarizer → fail-open, no checkpoint recorded, raw span kept in the next prompt, granular reason in the `mid_turn` `failedOpen` decision), plus write-gate regressions at both persist sites feeding a section-less fragment from a producer that bypasses the summarizer.
- Adversarial cases covered: fenced/blockquoted templates, unfenced verbatim template echo, four-backtick fence with a three-backtick inner line, fence closers with trailing text, nested bare fence markers as pseudo-content, heading-only skeletons, horizontal-rule padding (LF and CRLF), out-of-order sections, deeper heading levels, non-required-H2 content standing in for a required section, unclosed and mismatched-family fences, inline fence markers as content, missing Critical Context.
- A structurally complete but paragraph-sized summary of a >40k-char fold fails — measured over the full covered span, so a small roll-forward increment cannot bypass the floor; a proportionate structured summary of the same fold passes.
- Integration stubs that flow through the write gates were updated to structured sentinels (padded proportionally where the span triggers the floor).
- Full `@maka/runtime` and `@maka/runtime-host` suites pass locally.

## AI use

Select exactly one:

## Checklist

- [ ] No


UncertaintyDeterminesYou4ndMe:
Follow-up c08bfa2cb after an adversarial self-review pass:

- **Roll-forward bypass closed** (the most important one): the size floor was measured against the summarizer's own input — the newly folded increment plus the prepended prior-summary text — while the checkpoint replaces its FULL covered span. A small increment updating a checkpoint covering a large span slipped a fragment past the floor entirely; the prior-summary text also inflated the measure in the other direction. It now measures `input.source.foldedRuntimeEvents` with the shared `estimateRuntimeEventsTokens`, computed only when the summary is short enough for the floor to matter (no hot-path re-serialization). Regression test added.
- **Line-anchored section matching**: `### Goal` or an inline mention no longer satisfies the check via substring `includes`. The prompt is now built from the same section constants, so the mandated format and the validation cannot drift apart.
- **Fence truncation only counts line-opening fences**, so a verbatim ``` inside a preserved error message is not read as truncation (test added).
- **Granular reasons**: `malformed_summary_missing_section` / `_truncated` / `_too_small_for_fold`, matching the per-reason `skippedReasonCounts` convention.

Known trade-offs, deliberately kept to the issue's chosen semantics and left as follow-ups:
- A model that stably renders non-conforming headings (e.g. localized `## 目标`) fails validation deterministically and the session keeps failing open; the issue's proposal 5 (validation-failure retry with elevated instructions, or a dedicated summarizer model) is the mitigation.
- Already-persisted malformed checkpoints (pre-fix sessions) are not quarantined — checkpoint admission still accepts any non-empty summary; enforcing there is migration-sensitive and worth its own issue.
- The floor constants assume the default 4-chars/token estimate; threading `contextBudget.charsPerToken` through `HistoryCompactSummaryInput` would calibrate CJK sessions better.

Full `@maka/runtime` suite: 2839 pass.

Astro-Han (commented):
Thanks for adding a checkpoint admission guard for the #3029 failure. Validating the generated summary before it replaces the covered history is a useful and appropriately scoped incident fix. I found one remaining structural bypass, plus a production-seam fixture that needs updating before CI can pass.

**[P2] Required headings inside a closed code fence are accepted as real summary structure**

The current regular expressions search every line, including lines inside fenced code blocks. As a result, this completion is accepted:

````markdown
```markdown
## Goal
template goal
## Progress
template progress
## Next Steps
template next
```
````

I reproduced this against the production `buildLlmHistorySummarizer` at the current head. This is a plausible degraded-model response: a weak model can quote or demonstrate the requested template instead of producing the actual checkpoint, and the fenced example would still be allowed to replace the history.

Could the validator scan the document once and only recognize required headings outside code fences? The same small scanner could:

- require the sections in the expected order with non-empty content;
- detect unclosed backtick and tilde fences;
- avoid maintaining separate heading and fence interpretations.

I do not think this needs a general Markdown parser.

There is also a directly related CI fixture failure. The real-provider fixture in `execution-model-composition.test.ts` still returns the shared one-section `SUMMARY_TEXT`, so the new validator correctly rejects it, no checkpoint is written, and the expected Memory boundary becomes `undefined`. A compaction-specific structured fixture would fix this without changing the shared summary used by recap, proposal, and other tests.

As a non-blocking follow-up, I think the longer-term contract should make a versioned structured summary object authoritative and treat Markdown as a deterministic rendering. Model-authored narrative can then be kept separate from facts owned by Runtime and the durable ledgers. A bounded repair retry would also avoid retrying the same deterministic format failure on every Turn. That larger design does not need to block this incident fix once the fenced-heading bypass and CI fixture are corrected.

The existing roll-forward coverage fix and focused validation tests look sound; I did not find tests that should be removed. The already documented legacy-checkpoint and token-calibration limitations are better tracked as explicit follow-ups rather than expanded into this PR.

One merge-hygiene note: because the PR discloses material Claude Code assistance, please preserve a `Generated-by: Claude Code` trailer in the final squash commit.

<details>
<summary>中文对照</summary>

感谢为 #3029 的事故增加 checkpoint admission gate。在摘要替换 covered history 之前进行校验，是一个有价值且范围合适的事故修复。目前还存在一个结构校验绕过，以及一个需要更新才能让 CI 通过的真实调用链 fixture。

**[P2] 闭合代码块中的必需标题会被当成真实摘要结构**

当前正则会搜索所有行，包括 fenced code block 内的内容。因此下面的输出会被接受：

````markdown
```markdown
## Goal
template goal
## Progress
template progress
## Next Steps
template next
```
````

我已在当前 head 的生产 `buildLlmHistorySummarizer` 上复现。这个退化形态并不牵强：较弱的模型可能只是引用或演示提示中的模板，而没有真正生成 checkpoint；当前实现仍会允许这个 fenced 示例替换历史。

建议对文档做一次按行扫描，只认可代码围栏之外的必需标题。同一个小型扫描器还可以：

- 要求 section 顺序正确且内容非空；
- 检测未闭合的 backtick 和 tilde fence；
- 避免标题校验和 fence 校验分别维护两套文本解释。

这里不需要引入通用 Markdown parser。

当前还有一个由本 PR 直接触发的 CI fixture 失败。`execution-model-composition.test.ts` 的真实 provider fixture 仍返回共享的单节 `SUMMARY_TEXT`，因此新 validator 正确拒绝它、不再写入 checkpoint，后续预期的 Memory boundary 变成了 `undefined`。建议增加一个 compaction 专用的结构化 fixture，而不是修改 recap、proposal 等其他测试共用的 summary。

作为非阻塞 follow-up，长期更合理的契约是让带版本的结构化 summary object 成为权威数据，Markdown 只作为确定性渲染。这样可以把模型生成的叙事与 Runtime、durable ledger 拥有的真实事实分开。再增加一次有上限的格式修复重试，也能避免同一个确定性格式错误在每个 Turn 上反复失败。修复当前 fenced-heading 绕过和 CI fixture 后，不需要让这套更大的终局设计阻塞本次事故修复。

现有 roll-forward coverage 修复和 focused validation tests 是有效的，没有发现应该删除的测试。作者已说明的旧 checkpoint 和 token 校准限制，更适合建立明确的 follow-up，而不是继续扩大本 PR。

另有一项合并规范：PR 已披露使用 Claude Code 产生了实质性内容，最终 squash commit 应保留 `Generated-by: Claude Code` trailer。

</details>

AI-assisted review disclosure: Codex assisted with inspecting the diff and reproducing the validation and CI cases. I verified the evidence, assessed the scope and severity, and made the final review judgment.

**Comment by UncertaintyDeterminesYou4ndMe** [comment]
Both addressed in 0cc6ef71d.

**Fenced-heading bypass**: replaced the per-heading regexes with the single line scanner you suggested — one interpretation of the document for both checks. It recognizes the mandated sections only outside fenced code blocks, requires them in the expected order with non-empty content each (subheadings organize, non-heading lines and fenced lines carry content), and reports end-inside-open-fence from the same scan. Backtick and tilde fences are tracked by family, line-opening only, so a verbatim ``` inside a preserved error message stays content. Tests added: your fenced-template completion is rejected, a heading-only skeleton is rejected, and out-of-order sections are rejected; the inline-fence-marker acceptance test still passes.

**CI fixture**: `execution-model-composition.test.ts` now answers history-compaction requests (detected by the summarization system prompt, same pattern the test already uses for its request accounting) with a compaction-shaped `COMPACT_SUMMARY_TEXT`; recap, proposal, and title keep the shared `SUMMARY_TEXT`. Verified the Memory-boundary expectation passes again: full `@maka/runtime-host` suite 933 pass, full `@maka/runtime` suite 2842 pass.

Agreed on the longer-term contract (versioned structured summary object with Markdown as rendering, plus a bounded repair retry for deterministic format failures) — that pairs naturally with the legacy-checkpoint quarantine follow-up already noted.

**Merge hygiene**: added the `Generated-by: Claude Code` trailer to the PR description and to this follow-up commit so the squash preserves it.

**Comment by UncertaintyDeterminesYou4ndMe** [comment]
One more hardening from a self-run adversarial probe pass (fb24f9de2): a heading skeleton interleaved with horizontal rules (`## Goal` / `---` / …) satisfied the non-empty-content requirement because any non-heading, non-blank line counted as content. Thematic breaks (`---`, `***`, `- - -`) are now separators, not content; list items keep counting. Probes that held with no change: blockquoted template, four-backtick fences, CRLF endings, repeated headings between sections, tilde fences (closed, open, and mismatched-family), and fence-content-as-section-content. Suite: 2843 pass.

Generated-by: Claude Code

**Review by Astro-Han — COMMENTED** [review]
The exact fenced-template case and the Runtime Host fixture from the previous review are fixed. I verified the focused summarizer suite (20/20) and Biome on the six changed files. One structural bypass remains.

**[P2] Preserve the opening fence width when deciding whether a fence closes**

`scanSummaryStructure` tracks only the fence character family. Any later run of three or more characters from the same family closes the fence, even when it is shorter than the opener. Markdown requires a closing delimiter to be at least as long as the opening delimiter.

I reproduced this against the production `buildLlmHistorySummarizer` at the current PR head. The following completion is accepted:

~~~js
[
  "````markdown",
  "```",
  "## Goal",
  "template goal",
  "## Progress",
  "template progress",
  "## Next Steps",
  "template next",
].join("\n")
~~~

Under Markdown semantics, the three-backtick line does not close the four-backtick opener, so every required heading remains inside one unclosed fenced block. The scanner closes it early, recognizes those headings as real structure, and allows the completion to replace the covered history. This is the same failure class as the earlier fenced-template bypass.

Please retain the opening fence character and width, and only close it with the same character, an equal-or-longer delimiter, and no non-whitespace trailing content. A regression case using the completion above should cover it.

The PR also currently conflicts with the latest `main`, which added provider-native history compaction. The rebase should preserve the distinction between text summaries, which need this validation, and opaque provider-native checkpoint state.

AI-assisted review disclosure: Codex inspected the diff, traced the compaction lifecycle, and reproduced the fence-width bypass against the PR head. I verified the evidence and made the final review judgment.

**Comment by UncertaintyDeterminesYou4ndMe** [comment]
Rebased onto `main` to resolve the conflict with #2993 (Codex provider-native compaction):

- `HistoryCompactSummarizerFailureReason`: merged as the union — #2993's `input_too_large` / `invalid_provider_state` alongside this PR's granular `malformed_summary_*` reasons.
- The semantics compose cleanly: #2993's opaque-checkpoint filtering happens on entry, and this PR's validation still measures the summary against the full covered span (`input.source.foldedRuntimeEvents`), which is exactly the span a text checkpoint replaces — including the recompress-from-full-source path #2993 added for a provider-native predecessor.
- One rebase-only fixture fix: #2993's new `recompresses the full source when the previous checkpoint is provider-native` test stubbed a free-form `'portable summary'`, which this PR's validation rejects by design; it now returns the structured `VALID_SUMMARY`, same treatment as the other compaction fixtures. Its assertions (full-source recompression, no opaque-state leak) are unchanged.

Full `@maka/runtime` and `@maka/runtime-host` suites pass locally after the rebase.

Generated-by: Claude Code

**Comment by UncertaintyDeterminesYou4ndMe** [comment]
The `test_runtime_host` failure is a timing flake unrelated to this PR: `execution-composition.test.ts › hosted execution settles while its tracked environment resource remains verifiable` drives a FakeBackend hosted execution (no compaction/summarizer involvement) and waits 5s for settlement; locally it passes 3/3 on this exact tree but takes ~3.0–3.1s each run, so a loaded CI runner can tip it past the budget (the failing run took 7.7s). `main`'s latest CI run on the same base is green. Pushed an empty commit to retrigger, since I lack rerun rights here.

Generated-by: Claude Code

**Review by Astro-Han — COMMENTED** [review]
The incident fragment, the closed fenced template, and the roll-forward size-floor bypass are covered. Two things still stand in the way of approve.

The fence-width case CodeRabbit already opened on `scanSummaryStructure` still reproduces on this head (a four-backtick opener closed by a three-backtick line, headings accepted). Same failure class as the earlier fenced-template bypass.

The other is layering, inline on the generate-time assert. #3040 is the parallel #3029 fix and already owns the write gates. Landing both as-is leaves two contracts.

AI-assisted review: Grok 4.6 inspected the current head and reproduced the four-backtick completion against `scanSummaryStructure`. I checked the write-path call sites on this branch and on #3040, and confirmed the generate-only assert. Unverified by me: I did not rerun the runtime suite.

File: packages/runtime/src/history-compact-summarizer.ts [hunk 1]
```diff
@@ -116,6 +124,7 @@ export function buildLlmHistorySummarizer(options: BuildLlmHistorySummarizerOpti
       if (rawFinishReasonString(result.finishReason) === 'length') {
         throw new HistoryCompactSummarizerError('output_length');
       }
+      assertWellFormedCheckpointSummary(result.text, input.source.foldedRuntimeEvents);
```
Astro-Han: [P2] The gate still sits on generate, not on the two write paths.

`assertWellFormedCheckpointSummary` runs only in `buildLlmHistorySummarizer`, after `generateText`. `ai-sdk-compaction.ts` and `mid-turn-capacity-compact.ts` still persist any non-empty text summary. A different summarizer, the provider-native text path, or a later generate site can write the same fragment this PR exists to stop.

#3040 already puts `validateHistoryCompactSummary` on those two write gates. That is the right layer for "a bad summary must not replace folded history." This PR's scanner and size floor are stronger than #3040's regex, but they cannot enforce the invariant from here.

Move the scanner (once fence width is fixed) into a shared write-gate module and call it from both persist sites. Keep the generate-time throw if you want a typed reason on the summarizer path. Do not leave write as a trust-the-producer step.

I reproduced the current generate-only wiring on `618d756cd`: the only production call of `assertWellFormedCheckpointSummary` is this return path.

UncertaintyDeterminesYou4ndMe:
Rebased again after #3038 and #3100 landed on `main` (same-file changes in the summarizer). Only an import-line merge plus one rebase-only fixture pass: #3038's four step-grouping tests stub free-form summary text that this PR's validation rejects by design — they assert on the grouped messages, not the returned text, so their stubs now return the structured `VALID_SUMMARY`, same treatment as the earlier provider-native fixture. Full `@maka/runtime` and `@maka/runtime-host` suites pass locally.

Generated-by: Claude Code

Astro-Han:
Thanks for digging into the incident — the incident fragment is indeed rejected (I verified the regression test reproduces #3029's exact shape), and the scanner design is clean. Two independent reviews converged on the same blockers, worth resolving before merge.

**Conclusion: needs work — one P1 (parallel implementation with #3040/#3046) plus a reproducible validation bypass.**

**P1 — this PR and #3040/#3046 are uncoordinated parallel implementations of the same fix.**
#3039 (in-summarizer validation, 3 fine-grained reasons) and #3040 (new `history-compact-summary-validation.ts` module, write-gate validation, single `malformed_summary` reason, created 15 minutes later) touch the same files (`history-compact-error.ts` reason union, `history-compact-summarizer.ts` prompt constants, the same test files), and #3046 stacks load-side validation on #3040. Neither PR body references the other; #3040 is already CONFLICTING. The two validators have *different* semantics (order + per-section content + fence family + size floor vs. heading presence only, no order/content check) and complementary fence bugs — the same degraded output can pass one and be rejected by the other. Merging both leaves two contracts and one redundant validator. Please align with the #3040/#3046 authors: pick one as the single validation authority and explicitly close/rewrite the other (per AGENTS.md "extend the closest existing seam").

**P2-1 — the fence-tracking bypass (already reproduced by human reviewers, still open).** The scanner tracks fence *character family* but not opening width: a 4-backtick opener closed by a 3-backtick line is misjudged as closed (Markdown requires the closer to be ≥ opener length). I reproduced: a template-echo output with `` ````markdown ``` ## Goal … ## Next Steps `` passes structural validation. The existing "headings inside a fenced block" test only covers the 3-backtick variant. Fix: record opening width and require same char + width ≥ opener.

**P2-2 — validation still passes content that is meaningless.** An unfenced echo of the prompt template (346 chars of placeholders) is ACCEPTED for folds ≤10k tokens (the size floor only guards >10k); a 3-section summary missing `## Critical Context` (which carries files/worktree/test status — exactly what the incident lost) is ACCEPTED. The floor also uses the default charsPerToken=4 while the mid-turn path passes its own `inputBudget.charsPerToken`, so the floor can misfire for CJK-heavy sessions. Consider requiring `## Critical Context` and/or an end-to-end fold regression for the unfenced echo.

**P2-3 — the claimed acceptance criterion ("rejection lands in `historyCompactWriteSkippedReasonCounts` / `failOpenReason`") has no end-to-end test.** All new tests are at the summarizer unit level; no integration test asserts "reject → checkpoint not written → history preserved → diagnostic reason recorded". The wiring is shared with already-covered paths, so risk is low, but the PR's own acceptance criterion deserves one integration assertion or a written deferral.

Optional nits (P3): the trailing-colon truncation heuristic rejects legitimate summaries ending in `:`/`：` (per-session compaction can stall permanently; fail-open but starved); the size floor calibration is deferred (documented); per-line `new RegExp(...)` construction; the `VALID_SUMMARY` fixture string is duplicated in 4 files.

---

*AI-assisted review disclosure: this review was produced with AI assistance (two pi review subagents on `ollama-cloud/deepseek-v4-flash`, read-only). Both independently reproduced the fence-width bypass and the template-echo acceptance with isolated simulations of the scanner, and both flagged the #3040/#3046 overlap; I verified the overlap and the incident-fragment regression test myself. The "template echo accepted" outcome is from simulated runs of the exact scanner logic, not a live fold. Please weigh these findings with your own judgment.*

<details>

结论：FAIL，需要返工。两个独立审查收敛到同一组阻塞项：① P1：#3039 与 #3040/#3046 是同一修复的未协调并行实现（15 分钟内先后创建，改同一批文件，两套 validator 语义不同且各有互补的 fence bug），违反仓库"不建并行实现"原则，须与 #3040/#3046 作者对齐二选一并显式关闭另一份；② P2-1 fence 关闭宽度绕过（4-backtick 开启被 3-backtick 行误判闭合，已被人类审查人两次复现），结构性校验可被模板回显穿透，需记录开启宽度并要求闭合宽度 ≥ 开启宽度；③ P2-2 验证仍放行无意义内容——unfenced 模板回显（346 chars 占位符）在 ≤10k tokens 折叠上通过（size floor 只保护 >10k），缺 Critical Context 的三节瘦摘要也通过，建议把 Critical Context 加入必需 section；④ P2-3 声称的验收准则（拒绝 reason 落账）无端到端测试，只有单元层。事故片段确实被拦住且有逐字回归测试 ✅。P3 可选：尾冒号截断误杀合法摘要导致压缩永久停摆、charsPerToken 校准延后、fixture 字符串四处硬编码重复。
</details>

UncertaintyDeterminesYou4ndMe:
Thanks for the thorough round — both reproduced bypasses were real. All four points addressed in 33c509d5c:

**P2-1 (fence width)**: the scanner now records the opening run's width; a closer must match the family AND be at least as long. Your exact `````markdown`-with-inner-``` repro is a regression test (rejected as `missing_section` — the headings stay fenced), plus a wide-open/narrow-trailing case (`truncated`) and the still-accepted longer-closer case.

**P2-2 (meaningless content)**: three changes.
- Verbatim template lines never count as section content, so the unfenced template echo is rejected regardless of fold size — regression test feeds the exact `SUMMARY_FORMAT_TEMPLATE` back.
- `## Critical Context` joins the required sections: it carries exactly what the incident lost, and the template's `"(none)"` escape hatch keeps it satisfiable for trivial folds.
- The size floor now uses the session's `charsPerToken` from `input.inputBudget` on both sides of the comparison, so CJK-heavy sessions are floored consistently (this was the deferred calibration follow-up — it turned out the budget was already on the input).

**P2-3 (end-to-end)**: added at the backend seam (`mid-turn-capacity-backend.test.ts`): the REAL summarizer returns the incident-shaped fragment → the turn completes on the raw projection, no checkpoint is recorded, the raw span is still in the next prompt, and the `mid_turn` `failedOpen` decision carries `failOpenReason: malformed_summary_missing_section`.

**P1 (parallel implementations)**: the validation now lives in its own `history-compact-summary-validation.ts` — one module owning the mandated format (the prompt is built from its template) and exporting a writer-agnostic `findCheckpointSummaryDefect(summary, foldContext?)` predicate. That makes the alignment concrete: the load side (#3046's concern, which is a follow-up this PR explicitly declared out of scope) can consume the same predicate with no fold context, and #3040's write-gate placement becomes redundant once nothing malformed leaves the summarizer. I'll propose exactly that on #3040 — deferring to @me2seeks and the maintainers on which surviving shape they prefer.

P3s: the per-line `RegExp` construction is hoisted. On the trailing-colon starvation concern — agreed it's the sharpest edge of deterministic validation; it's inherent to proposals 1–3's fail-open semantics and the issue's proposal 5 (bounded validation-failure retry) is the mitigation, still deferred as a follow-up. The `VALID_SUMMARY` fixture is still local to each test file, but the format itself is now importable from the validation module, which is the part that must not drift.

Full `@maka/runtime` (incl. 24 summarizer-validation tests) and `@maka/runtime-host` suites pass locally.

Generated-by: Claude Code

UncertaintyDeterminesYou4ndMe:
Rebased past #3113 (summarizer input bounding) and #3128. One rebase-only fixture pass, same pattern as the previous rounds: #3113's new bounding test stubs a free-form summary over a >10k-token span at `charsPerToken: 1`, which this PR's validation rejects by design — first structurally, then at the size floor (that test incidentally exercises the floor's new charsPerToken calibration). Its stub now returns a proportionate structured summary; the bounding assertions are unchanged. Full `@maka/runtime` and `@maka/runtime-host` suites pass locally.

Generated-by: Claude Code

packages/runtime/src/history-compact-summarizer.ts
```diff
@@ -124,6 +133,100 @@ export function buildLlmHistorySummarizer(options: BuildLlmHistorySummarizerOpti
...
+  let fenceFamily: string | undefined;
+  let matchedSections = 0;
+  const sectionHasContent: boolean[] = REQUIRED_SUMMARY_SECTIONS.map(() => false);
+  for (const line of text.split('\n')) {
+    const fence = /^\s*(`{3,}|~{3,})/.exec(line);
+    if (fence) {
+      const family = fence[1]![0]!;
+      if (fenceFamily === undefined) fenceFamily = family;
+      else if (fenceFamily === family) fenceFamily = undefined;
+      continue;
```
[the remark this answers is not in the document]
UncertaintyDeterminesYou4ndMe: Fixed in 33c509d5c (opening width tracked; closer requires same family and >= width; your four-backtick repro is a regression test) and hardened further in 5e5903dd3: a closer must also be a bare marker run — an opener may carry an info string, a closer may not — with a trailing-text regression test.

Generated-by: Claude Code

packages/runtime/src/history-compact-summarizer.ts
UncertaintyDeterminesYou4ndMe: Done in 5e5903dd3. The scanner lives in `history-compact-summary-validation.ts` (extracted in 33c509d5c) and both persist sites now call its `findCheckpointSummaryDefect` before building a checkpoint — the pre-turn writer in `ai-sdk-compaction.ts` (the enclosing catch maps the reason into `historyCompactWriteSkippedReasonCounts`/`failOpenReason`) and the mid-turn engine in `mid-turn-capacity-compact.ts` (fail-open with the granular `diagnosticReason`). The generate-time throw stays for the typed reason on the summarizer path. Regression tests at both gates feed a section-less fragment from a producer that bypasses the summarizer and assert fail-open with no persist. This is also the shape proposed to #3040/#3046: one predicate, consumed at every gate.

Generated-by: Claude Code

packages/runtime/src/history-compact-summary-validation.ts
```diff
@@ -0,0 +1,173 @@
...
+  // with 'stop'.
+  if (/[:：]$/.test(trimmed) || scan.endsInsideOpenFence) {
+    return 'malformed_summary_truncated';
+  }
+  if (foldContext !== undefined) {
+    const charsPerToken = foldContext.charsPerToken ?? DEFAULT_CHARS_PER_TOKEN;
+    if (
+      trimmed.length < LARGE_FOLD_SUMMARY_TOKENS_FLOOR * charsPerToken &&
+      estimateRuntimeEventsTokens(foldContext.coveredRuntimeEvents, charsPerToken) >
+        LARGE_FOLD_ESTIMATED_TOKENS
```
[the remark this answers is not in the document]
UncertaintyDeterminesYou4ndMe: Fixed in 5e5903dd3: `charsPerToken` is clamped to >= 1 before it scales the floor, so a zero/negative estimate can no longer zero it out.

Generated-by: Claude Code

packages/runtime/src/history-compact-summary-validation.ts
```diff
@@ -0,0 +1,173 @@
...
+      } else {
+        // A shorter run or a different family does not close the fence; the
+        // line is fenced content of the enclosing section.
+        countContent(line);
+      }
+      continue;
+    }
+    if (openFence !== undefined) {
+      // Fenced lines are content of the enclosing section, never headings.
+      countContent(line);
```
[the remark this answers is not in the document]
UncertaintyDeterminesYou4ndMe: Working as intended, with the sharp edge covered elsewhere: a fenced line is literal content of its enclosing section — a summary whose Progress body is one code block is legitimate (covered by the fence-content-as-section-content acceptance test), so skipping content counting inside fences would misreject those. The degenerate case this rule could admit (a section whose only "content" is a fenced template echo) is closed by the template-placeholder exclusion, which applies inside fences too — a fenced verbatim template line never counts. A fenced non-template line is preserved information, which is exactly what a checkpoint must carry.

Generated-by: Claude Code

UncertaintyDeterminesYou4ndMe:
Addressed the four inline review threads in 5e5903dd3 (each replied and resolved in place):

- **Write gates** (Astro-Han's P2, the substantive one): both persist sites — the pre-turn writer and the mid-turn engine — now call `findCheckpointSummaryDefect` before building a checkpoint, so the invariant holds for any producer, not just the default summarizer's generate path. Regression tests at both gates bypass the summarizer with a section-less fragment and assert fail-open with the granular reason and no persist. This also completes the shape proposed to #3040/#3046: one predicate, consumed at every gate.
- Fence closers must be bare marker runs (opener may carry an info string, a closer may not), on top of the earlier width fix.
- `charsPerToken` clamped to >= 1 so the size floor cannot be zeroed out.
- The fenced-content thread is answered as working-as-intended: fenced non-template lines are preserved information; the template-placeholder exclusion already applies inside fences.

Compaction test stubs across the runtime suite are now shaped like real checkpoints (sentinels preserved, padded proportionally where the span triggers the floor). Full `@maka/runtime` and `@maka/runtime-host` suites pass locally.

Generated-by: Claude Code

UncertaintyDeterminesYou4ndMe:
Also addressed Copilot's two suppressed observations in 7b8fb031e:

- **Content attribution across an unmatched H2** (the real one): an intervening non-required H2 like the template's `## Key Decisions` now opens its own section, so its bullets can no longer satisfy an empty required section above it. Regression: empty `## Progress` backed only by Key Decisions content is rejected.
- **CRLF**: the line scan now splits on `/\r?\n/`. The existing checks already tolerated a trailing `\r` (the thematic-break regex consumes it via `\s*`, and content/section checks trim), so this was defensive normalization rather than a live bypass — but it removes the whole class, and CRLF acceptance plus CRLF thematic-break rejection are now pinned by tests.

Full `@maka/runtime` and `@maka/runtime-host` suites pass locally.

Generated-by: Claude Code

packages/runtime/src/history-compact-summary-validation.ts
```diff
@@ -0,0 +1,197 @@
+import type { RuntimeEvent } from '@maka/core/runtime-event';
+import { estimateRuntimeEventsTokens } from './context-budget-helpers.js';
+
+// The single authority on what a history-compact checkpoint summary must look
```
Astro-Han, line 4: **P1 — Make this policy authoritative on load/replay, not only provider writes.** The comment says the load side can consume the same predicate, but `history-compact-ledger.ts` still accepts projected and ledger checkpoints with shape validation only. A pre-existing non-empty malformed summary therefore becomes authoritative again after restart and replaces the covered canonical events; copy/artifact paths can propagate it too. Please apply one version-aware semantic admission at the checkpoint load/replay authority, quarantine rejected entries, repair the projection, and fall back to canonical events. Add a restart test with a shape-valid malformed persisted checkpoint.

Astro-Han (commented):
Thanks for substantially strengthening the compaction summary scanner. I reviewed this head with two independent review slices plus a read-only `ollama-cloud/deepseek-v4-flash:high` pass. The current scanner correctly handles ordered non-empty sections, placeholder echoes, backtick/tilde fences, CRLF, and a large-fold size floor; those are meaningful improvements over the shallow heading checks in the parallel proposals.

The remaining issue is authority placement. A history checkpoint can arrive through more than the two provider write call sites: an older persisted ledger/projection, conversation copy, artifacts, and direct recorder seams can still be shape-valid while carrying a semantically malformed summary. `history-compact-ledger.ts` currently accepts those checkpoints with `validateHistoryCompactCheckpointShape()` only, so restart can make the malformed checkpoint authoritative again and replace its covered canonical events.

The first-principles/Occam design is one semantic admission policy at the durable checkpoint boundary, reused by record, load/replay, projection repair, copy, and artifact import. New source-bound checkpoints should use the strong current scanner and fold context; legacy checkpoints need an explicit compatibility policy, but must not silently pass as valid merely because `summary` is non-empty. On rejection, quarantine the checkpoint and replay canonical events. This should be consolidated with #3040 rather than maintaining competing validators.

Current CI is green. No local test suite was run during this review; conclusions are based on source, test, and current CI inspection. Codex coordinated the independent passes and performed the final adjudication; external-model output was treated as unverified until checked against the code.

<details>
<summary>中文摘要</summary>

感谢显著加强 compaction summary scanner。当前实现已经正确处理 required section 的顺序与非空、模板回显、backtick/tilde fence、CRLF 和大折叠尺寸下限，这比并行方案中的浅层 heading 检查更可靠。

剩余问题在 authority placement：checkpoint 不只来自两个 provider write call sites；旧 ledger/projection、conversation copy、artifact 和直接 recorder 仍可能带来 shape 合法但语义损坏的 summary。当前 load 只调用 `validateHistoryCompactCheckpointShape()`，restart 后会重新把坏 checkpoint 作为权威并替换 canonical events。

更符合第一性原理和奥卡姆剃刀的方案是在 durable checkpoint boundary 只保留一套 semantic admission，record、load/replay、projection repair、copy 和 artifact import 全部复用。新 source-bound checkpoint 使用当前强 scanner 与 fold context；legacy 使用明确兼容策略，但不能仅因 summary 非空就放行。拒绝时 quarantine 并回放 canonical events。建议与 #3040 收敛，不要维护并行 validator。

当前 CI 全绿。本次未运行本地测试套件；结论来自源码、测试与 CI 检查。外部模型输出在核对代码前均视为未验证输入。
</details>

packages/runtime/src/history-compact-summary-validation.ts
me2seeks: I cannot push to the author fork from the me2seeks credential, so I prepared the reviewed load/replay fix as 48de7eb237 on https://github.com/me2seeks/maka-agent/tree/review-fixes/3039-load-replay. It applies the same version-aware semantic admission at checkpoint load/replay, quarantines malformed persisted entries, repairs the projection, and falls back to canonical events. Runtime passes 2908 total / 2902 pass / 6 skip. Please cherry-pick 48de7eb237 or compare the branch; I have not resolved this author-owned thread.

UncertaintyDeterminesYou4ndMe: Done — cherry-picked @me2seeks's prepared commit (4b05575c4, authorship preserved; thanks for closing #3040/#3046 in favor of this PR and preparing this). The checkpoint load/replay authority (`history-compact-ledger.ts`) now applies a version-aware semantic admission through the same shared scanner: all three load sites quarantine a text checkpoint whose summary is truncated (unclosed fence via the same fence scan, or a mid-thought tail), the projection path falls through to the canonical ledger and repairs the projection, and prior valid checkpoints are recovered. Legacy checkpoints that predate the section contract keep loading when section-less but intact — only truncation is writer-agnostic. Restart-shaped tests cover: shape-valid truncated persisted checkpoint rejected with prior-valid recovery, legacy fence-only quarantine, and truncated-projection repair from the canonical ledger.

Generated-by: Claude Code

packages/runtime/src/history-compact-summary-validation.ts
```diff
@@ -0,0 +1,197 @@
...
+// A verbatim echo of the mandated template carries no information: template
+// lines never count as section content, fenced or not, so a degraded model
+// parroting the format back cannot pass as a checkpoint.
+const TEMPLATE_PLACEHOLDER_LINES: ReadonlySet<string> = new Set(
+  SUMMARY_FORMAT_TEMPLATE.map((line) => line.trim()).filter((line) => line.length > 0),
+);
+
+const SECTION_HEAD_PATTERNS = REQUIRED_SUMMARY_SECTIONS.map(
+  (section) => new RegExp(`^${section}\\b`),
+);
```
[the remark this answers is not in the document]
UncertaintyDeterminesYou4ndMe: Real bug — fixed in 4b05575c4: heading recognition is now an exact whole-line match against the section constant (`line.trimEnd() === section`), no regex word boundary at all, so `## Goal-old` and friends no longer count. Regression test: each mandated heading with a suffixed variant is rejected.

Generated-by: Claude Code

packages/runtime/src/history-compact-summary-validation.ts
```diff
@@ -0,0 +1,197 @@
...
+        // Anything else — a shorter run, a different family, or trailing
+        // text — is fenced content of the enclosing section.
+        countContent(line);
+      }
+      continue;
+    }
+    if (openFence !== undefined) {
+      // Fenced lines are content of the enclosing section, never headings.
+      countContent(line);
+      continue;
```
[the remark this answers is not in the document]
UncertaintyDeterminesYou4ndMe: Working as intended — same resolution as the earlier CodeRabbit thread on this line. The outside-fence rule binds to HEADING recognition, not to content: fenced lines are literal content of their enclosing section, because preserved commands and error output are exactly what a checkpoint must carry, and skipping them would misreject a summary whose Progress body is one code block. The degenerate shapes are excluded specifically: verbatim template placeholder lines never count (fenced or not), and bare fence-marker runs never count — so "headings plus fenced template echo" cannot pass, while "headings plus fenced real content" is a legitimate summary. Clarified the scanner's doc comment in bef8aeba7, which read as if content were also fence-scoped.

Generated-by: Claude Code

packages/runtime/src/ai-sdk-compaction.ts
```diff
@@ -583,6 +584,20 @@ export class AiSdkCompaction {
...
+      // reasons; any other producer is validated here, and the enclosing
+      // catch maps the reason into the fail-open diagnostics.
+      if (typeof compacted === 'string') {
+        const defect = findCheckpointSummaryDefect(compacted, {
+          coveredRuntimeEvents: foldedRuntimeEvents,
+          ...(input.contextBudget.charsPerToken !== undefined
+            ? { charsPerToken: input.contextBudget.charsPerToken }
+            : {}),
+        });
+        if (defect) throw new HistoryCompactSummarizerError(defect);
```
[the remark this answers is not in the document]
UncertaintyDeterminesYou4ndMe: Addressed in 94e96e7ea as a contract clarification, deliberately not a re-acceptance: `HistoryCompactSummarizer`'s type now documents that a string result must satisfy the mandated checkpoint format owned by `history-compact-summary-validation.ts` (with `SUMMARY_FORMAT_TEMPLATE` as the shape to emit), and that the write gates reject free-form text by design. Every in-repo producer is migrated: `buildLlmHistorySummarizer` emits and self-validates the format, the Codex compactor returns provider-native state (exempt), and all test stubs are structured. Re-accepting legacy plain strings would reopen the #3029 hole; an out-of-repo producer that returns free-form text degrades to the documented fail-open (history kept, granular reason in diagnostics) rather than silent data loss.

Generated-by: Claude Code

UncertaintyDeterminesYou4ndMe:
This round is settled (all threads replied and resolved):

- **Load/replay authority** (Astro-Han's P1): cherry-picked @me2seeks's prepared commit 4b05575c4 with authorship preserved — the ledger's three load sites quarantine truncated text checkpoints through the same shared scanner, repair the projection from the canonical ledger, and recover the prior valid checkpoint; legacy section-less-but-intact summaries keep loading. With #3040 and #3046 closed in its favor, this PR now carries the single validation contract across summarizer, both write gates, and load/replay.
- **Qodo findings**: the `\b` suffixed-heading bypass is fixed by exact whole-line heading matching (in the cherry-picked commit, with a regression test); the size floor now compares via the shared ceil-based `estimateTokens` so an exactly-200-token summary passes (boundary test); the `HistoryCompactSummarizer` contract documents the mandated format instead of re-accepting free-form strings; the fenced-content finding is answered as working-as-intended (same resolution as the earlier CodeRabbit thread), with the scanner comment clarified to bind the outside-fence rule to heading recognition.
- The truncation tail heuristic is also broader now (ellipsis, commas, dangling connectors — from the cherry-picked commit), shared verbatim between write admission and load quarantine.

Full `@maka/runtime` and `@maka/runtime-host` suites pass locally.

Generated-by: Claude Code

Astro-Han (commented):
The latest update is a rebase; the PR-owned runtime files are unchanged from the tree reviewed. The shared scanner, both write gates, fail-open behavior, and load-time recovery are a strong improvement over the incident path. Two remaining boundaries keep me from approving this head.

First, the scanner accepts a four-space-indented fence marker as a delimiter, although Markdown treats it as indented code. That lets headings after a false closer escape an actually open fence. Second, every text checkpoint is still V2, so load/copy cannot distinguish a compatible legacy free-form summary from a newly produced section-contract summary; truncation-only quarantine therefore cannot enforce the new contract at the durable boundary. The smallest durable shape is an explicit summary-format marker/version: retain V2 compatibility, but apply the complete predicate to newly marked checkpoints across record, load, repair, and copy.

I am leaving a review comment rather than approving. AI-assisted review by Codex with three independent reviewer passes and OpenCode Go DeepSeek V4 Flash (high); I verified the rebased head, scanner probes, ledger/copy authority, focused test evidence, and live checks.

最新更新只是 rebase，PR 自身的 runtime 文件与已审查 tree 相同。共享 scanner、两个写入 gate、fail-open 和 load-time recovery 都正确加强了事故路径。仍有两个边界未闭合：四空格缩进的 fence marker 被错误当作 delimiter；所有文本 checkpoint 仍共用 V2，load/copy 无法区分兼容的 legacy free-form summary 与新 section contract。

最小长期方案是增加明确的 summary format marker/version：V2 保留兼容，新格式在 record、load、repair、copy 全路径执行完整 predicate。当前先发 COMMENT，不 Approve。

本次由 Codex、三个独立 reviewer 和 OpenCode Go DeepSeek V4 Flash high 辅助；已核对 rebase 后 head、scanner probe、ledger/copy authority、focused tests 与实时检查。

</details>

packages/runtime/src/history-compact-summary-validation.ts
```diff
@@ -0,0 +1,224 @@
...
+    if (trimmedLine.length === 0) return;
+    if (TEMPLATE_PLACEHOLDER_LINES.has(trimmedLine)) return;
+    // A bare fence-marker run carries no checkpoint information, so nested
+    // fence delimiters cannot satisfy a section's non-empty requirement.
+    if (/^(`{3,}|~{3,})$/.test(trimmedLine)) return;
+    sectionHasContent[matchedSections - 1] = true;
+  };
+  // Normalized so CRLF input cannot smuggle a trailing \r past any line check.
+  for (const line of text.split(/\r?\n/)) {
+    const fence = /^\s*(`{3,}|~{3,})/.exec(line);
```
Astro-Han: [P2] Limit fence delimiters to at most three leading spaces

This regex accepts arbitrary leading whitespace, so a four-space-indented bare run closes the current fence here. Under Markdown rules that line is indented code, not a fence delimiter; headings after it remain inside the open fence but this scanner treats them as real required sections and admits the summary. Please capture/limit indentation to 0–3 spaces and add a four-space false-closer regression.

这里接受任意数量的前导空白，因此四空格缩进的 bare fence run 会关闭当前 fence。但按 Markdown 规则，它是缩进代码而不是 delimiter；后续 headings 实际仍在 fence 内，scanner 却会当成真实 required sections。请限制为 0–3 个前导空格，并补四空格伪 closer 回归。

</details>

packages/runtime/src/history-compact-ledger.ts
```diff
@@ -1,15 +1,30 @@
...
+ * Legacy checkpoints predate the sectioned summarizer contract, so their
+ * summaries may omit `## Goal` etc. and remain usable. A truncated fragment,
+ * however, poisons every subsequent replay with a half-finished thought
+ * regardless of which writer produced it (#3029, #3041). The load path asks
+ * the shared summary scanner only for its truncation result, leaving
+ * section-less legacy summaries intact.
+ */
+function hasLoadableHistoryCompactSummary(checkpoint: HistoryCompactCheckpoint): boolean {
+  if (!isTextHistoryCompactCheckpoint(checkpoint)) return true;
+  return findCheckpointSummaryTruncationDefect(checkpoint.summary) === undefined;
```
Astro-Han, line 34: [P2] Give the new summary contract a durable identity

Truncation-only compatibility is reasonable for old V2 data, but V2 also remains the format written and copied today. A shape-valid, non-truncated free-form summary can therefore enter through copy/direct recorder seams and later replace its covered events, while this loader has no way to tell whether it is legacy or violated the current contract. Please add an explicit summary-format marker/version: preserve unmarked V2 under this compatibility policy, and require the complete shared predicate for newly marked checkpoints at record, load/repair, and copy.

对旧 V2 仅做 truncation 兼容是合理的，但当前写入和复制仍然也是 V2。因此 shape 合法、未截断的 free-form summary 仍可经 copy/direct recorder 进入，load 时无法判断它是 legacy 还是违反新契约。建议增加明确的 summary format marker/version：未标记 V2 保留兼容，新标记 checkpoint 在 record、load/repair、copy 全路径执行完整共享 predicate。

</details>

packages/runtime/src/history-compact-summary-validation.ts
UncertaintyDeterminesYou4ndMe: Fixed in c8d193e1e: the delimiter regex is `/^ {0,3}(`{3,}|~{3,})/` — four or more leading spaces is indented code and can no longer falsely close (or open) a fence, so fenced headings after an indented pseudo-closer stay fenced. Regressions added: the four-space false closer is rejected (`malformed_summary_truncated`, the document still ends inside the open fence), and a closer indented up to three spaces still closes.

Generated-by: Claude Code

packages/runtime/src/history-compact-ledger.ts
UncertaintyDeterminesYou4ndMe: Done in c8d193e1e, following the shape you proposed:

- `TextHistoryCompactCheckpoint` carries a durable `summaryFormat: 'sections_v1'` marker. `buildHistoryCompactCheckpoint` stamps it by default, so every checkpoint produced through the write gates is marked; a `'legacy_freeform'` build input exists solely for copy to preserve an unmarked source.
- **Load/repair**: marked checkpoints are held to the complete shared predicate (minus the size floor — the covered-span estimate is not durable at load); unmarked legacy V2 stays under the truncation-only compatibility policy. A marked, shape-valid, non-truncated free-form summary that entered through a direct recorder or copy seam is quarantined at load with prior-valid recovery (test added).
- **Shape validation** fails closed on an unknown marker, so a future format the running code does not understand falls to the canonical-events repair path instead of replaying unvalidated.
- **Copy**: a marked source summary is validated against the complete predicate before it can propagate into a fresh session, and the target preserves the source's format identity — a legacy free-form summary copies as legacy instead of being promoted into the new contract and then quarantined at its next load.

Generated-by: Claude Code

UncertaintyDeterminesYou4ndMe:
Both remaining boundaries from the latest review are closed in c8d193e1e:

- **Indented pseudo-fences**: delimiters are limited to at most three leading spaces per the Markdown rule; a four-space-indented marker run is indented code and cannot falsely close (or open) a fence. Regression tests cover the false closer and the still-valid three-space closer.
- **Durable summary-format identity**: text checkpoints built under the sectioned contract are stamped `summaryFormat: 'sections_v1'`; load/repair holds marked checkpoints to the complete predicate while unmarked legacy V2 keeps the truncation-only compatibility policy; shape validation fails closed on unknown markers; conversation copy validates marked sources and preserves the source's format identity in the target. Record is covered by the write gates plus default stamping.

Full `@maka/runtime` and `@maka/runtime-host` suites pass locally (legacy-modeling test fixtures now build with `legacy_freeform` explicitly).

Generated-by: Claude Code

Astro-Han (commented):
The latest head fixes both previously reported mechanisms: Markdown fences now follow the 0–3-space delimiter rule, and `sections_v1` gives new section-contract summaries an explicit durable identity while preserving unmarked legacy free-form checkpoints. Unknown markers fail closed and load/copy preserve the marker.\n\nTwo admission gaps remain below. Both come from the same first-principles rule: `sections_v1` should be proof that the full predicate was checked, so only the durable authority may assign/cache it, and every transfer seam with covered events available must re-run the size floor. That is smaller and safer than adding more downstream exceptions.\n\n`test_workspaces` is still running; other reported checks currently pass.\n\nReviewed with Codex using two independent reviewer agents and OpenCode Go DeepSeek V4 Flash (high); I verified the exact delta, prior findings, malformed-summary paths, copy/record/load authority, compatibility, tests, and live CI.\n\n<details><summary>中文</summary>\n\n最新 head 已修复此前两个机制：Markdown fence 现在遵循 0–3 spaces delimiter 规则；`sections_v1` 为新的 section-contract summary 提供显式 durable identity，同时保留无 marker 的 legacy free-form checkpoint。未知 marker 会 fail closed，load/copy 也会保留 marker。\n\n下面仍有两个 admission 缺口，根因相同：`sections_v1` 应代表完整 predicate 已验证，因此只有 durable authority 能赋予/缓存它；任何能拿到 covered events 的 transfer seam 都应重新执行 size floor。这比继续添加下游例外更小、更安全。\n\n`test_workspaces` 仍在运行；其余已报告检查当前通过。\n\n本次由 Codex 配合两个独立 reviewer agent，以及 OpenCode Go DeepSeek V4 Flash（high）审查；我核验了精确增量、既有 finding、malformed-summary 路径、copy/record/load authority、兼容性、测试和实时 CI。\n\n</details>

packages/runtime/src/conversation-copy.ts
```diff
@@ -617,6 +618,17 @@ function cloneAgentRunEvent(
...
     // new session or degrade it into user-visible placeholder text.
     if (sourceCheckpoint.version === 3) return null;
+    // Copy is an admission seam for the sectioned summary contract: a marked
+    // checkpoint whose summary no longer satisfies the complete predicate
+    // must not propagate into a fresh session. Unmarked legacy summaries stay
+    // copyable under the truncation-only load policy and keep their unmarked
+    // identity in the target.
+    if (
+      sourceCheckpoint.summaryFormat !== undefined &&
+      findCheckpointSummaryDefect(sourceCheckpoint.summary) !== undefined
```
Astro-Han: [P2] Apply the size-floor check after the covered prefix is known. This pre-match call has no `foldContext`, so a structurally valid `sections_v1` summary under 200 tokens can still be copied while replacing more than 10,000 estimated tokens. The target preserves the marker and durable load cannot reconstruct that span, so the undersized summary remains authoritative. Match first, then pass the matched `coveredRuntimeEvents` into the full predicate (or omit the checkpoint and rebuild from canonical events), with a large-span short-summary copy regression.

packages/runtime/src/history-compact-checkpoint.ts
```diff
@@ -278,7 +296,14 @@ export function buildHistoryCompactCheckpoint(
   };
   const checkpoint: HistoryCompactCheckpoint = providerState
     ? { ...common, version: 3, providerState }
-    : { ...common, version: 2, summary: summary! };
+    : {
+        ...common,
+        version: 2,
+        summary: summary!,
+        ...(input.summaryFormat === 'legacy_freeform'
```
Astro-Han, line 319: [P2] Do not let the generic builder mint `sections_v1` for unvalidated text. Any direct recorder caller can pass a free-form or undersized summary here, receive the trusted marker, and `HistoryCompactCheckpointCoordinator.record()` persists and caches it without a semantic predicate; the malformed checkpoint affects same-process replay before a restart can quarantine it. Make validated sectioned format explicit/non-default, or enforce the full predicate at the durable record authority, and fix the existing Desktop E2E free-form caller.

packages/runtime/src/conversation-copy.ts
UncertaintyDeterminesYou4ndMe: Fixed in 12f020f4a: the copy admission now runs AFTER `matchHistoryCompactCheckpointPrefix` and passes the matched `coveredRuntimeEvents` into the complete predicate, so the size floor is re-run over the actual replaced span — a structurally valid but undersized `sections_v1` summary can no longer be copied over a >10k-token fold. The builder backstops the same rule during the target rebuild (see the sibling thread), so even a future copy-path refactor cannot skip it. Regression at the builder seam: a proportionally undersized structured summary over a large covered span is refused with `malformed_summary_too_small_for_fold`.

Generated-by: Claude Code

packages/runtime/src/history-compact-checkpoint.ts
UncertaintyDeterminesYou4ndMe: Fixed in 12f020f4a by making the marker un-mintable rather than adding a downstream exception, per your first-principles rule: `buildHistoryCompactCheckpoint` itself runs the complete predicate — structure, truncation, AND the size floor, since every construction seam has the covered events in hand — and throws before stamping `sections_v1`. A direct recorder caller with free-form or undersized text gets an error, not a trusted marker, so `HistoryCompactCheckpointCoordinator.record()` can only ever persist/cache marker-bearing checkpoints that earned it, and same-process replay before a restart is covered too. `legacy_freeform` stays the explicit escape (used by copy to preserve unmarked legacy identity). The Desktop E2E caller now emits a sectioned deterministic summary, and the minting-refusal + builder-size-floor regressions are added.

Generated-by: Claude Code

UncertaintyDeterminesYou4ndMe:
Both admission gaps closed in 12f020f4a, by centralizing rather than adding downstream exceptions: `buildHistoryCompactCheckpoint` is now the only place `sections_v1` is assigned, and it runs the complete predicate — structure, truncation, and the size floor over the covered span it is handed — before stamping, throwing otherwise. That makes the marker un-mintable for unvalidated text at every construction seam (write gates, direct recorders, copy rebuilds, the coordinator's persist/cache path, same-process replay). Conversation copy additionally re-runs the complete predicate with the MATCHED covered span before rebuilding, and the Desktop E2E caller now emits a sectioned deterministic summary. Legacy-modeling fixtures across the suites declare `legacy_freeform` explicitly. Full `@maka/runtime` and `@maka/runtime-host` suites pass locally.

Generated-by: Claude Code

DeepPattern-AI:
Hi UncertaintyDeterminesYou4ndMe, sorry to bother you. This PR is very relevant to a workflow failure mode I’m trying to understand: a bad compaction summary replacing a large amount of real history, then causing the continuation model to confabulate around missing context.

The choice to validate summaries before they become checkpoints, and to treat degraded provider output as unsafe rather than merely “non-empty,” is exactly the kind of boundary that matters in long agent sessions.

We’re inviting a small group of developers with hands-on AI dev tool / coding-agent workflow experience to an external testing program. For now I’m only checking interest, and details can be shared privately if relevant. If this sounds interesting, could you reply and let me know?

packages/runtime/src/history-compact-summary-validation.ts
```diff
@@ -0,0 +1,236 @@
...
+      matchedSections < REQUIRED_SUMMARY_SECTIONS.length &&
+      line.trimEnd() === REQUIRED_SUMMARY_SECTIONS[matchedSections]
+    ) {
+      matchedSections += 1;
+      attributesToRequiredSection = true;
+      continue;
+    }
+    // Any other H2 opens its own (non-required) section; deeper headings
+    // organize within the current one.
+    if (/^##\s/.test(line)) {
```
Astro-Han: **[P2] Recognize all valid ATX headings before counting section content**

These heading checks only match column-zero headings followed by whitespace. CommonMark also permits up to three leading spaces and permits the marker run to end the line, so inputs such as `  ### Done` or bare `##` are counted as content for the preceding required section. A summary made only of the four required headings plus indented/bare headings can therefore pass structural admission (and, below the large-fold floor, be stamped `sections_v1`) despite carrying no information.

Use one heading matcher such as `^ {0,3}#{1,6}(?:\s|$)` for both the non-required-H2 transition and content exclusion, and add regressions for indented and empty heading-only skeletons.

Astro-Han (commented):
Centralizing `sections_v1` minting in `buildHistoryCompactCheckpoint()` closes the previous admission/copy gaps, and current CI is green. One structural parser bypass remains, so an information-free heading skeleton can still receive the trusted marker and replace history.

AI-assisted review disclosure: Codex reviewed exact head `12f020f`, the latest builder/copy delta, all current checks, and thread state, with two independent reviewer passes.

<details>
<summary>中文说明</summary>

把 `sections_v1` 的签发集中到 checkpoint builder 是正确的最终状态，之前的 admission/copy 缺口已关闭，当前 CI 也已全绿。但结构扫描没有按 Markdown 规则识别 1–3 个前导空格或行尾结束的 ATX heading，纯 heading 骨架仍可能被当成有效内容并获得可信 marker，需要先修复。

</details>

packages/runtime/src/history-compact-summary-validation.ts
UncertaintyDeterminesYou4ndMe: Fixed in d81007956 with one heading interpretation across all three checks, per your suggestion: content exclusion skips every ATX heading via `/^ {0,3}#{1,6}(?:\s|$)/` (bare marker runs and 1–3-space-indented headings included), the non-required-H2 transition uses `/^ {0,3}##(?!#)(?:\s|$)/`, and required-section matching now also tolerates up to three leading spaces around the exact heading text (a CommonMark-equivalent heading should match its section, not silently demote it). Four-plus spaces stays indented code, where the template-placeholder and bare-marker exclusions keep it information-free. Regressions: a skeleton padded with ` ### Done`, `  ###`, ` ##`, and `   #### notes` pseudo-body lines is rejected as `missing_section` (and therefore never earns the marker at the builder), and a required heading indented up to three spaces still matches. Full `@maka/runtime` and `@maka/runtime-host` suites pass locally.

Generated-by: Claude Code

Astro-Han (approved):
Approved on exact head `d810079`.

The summary-validation contract is now coherent at its natural authority: the checkpoint builder alone mints `sections_v1` after the complete predicate passes, write/load/copy paths preserve the durable format distinction, and malformed marked checkpoints fail open to canonical history. The latest delta closes the remaining indented and bare ATX-heading skeleton bypass without adding a parallel parser.

No remaining P0-P2 findings. All current checks pass, the PR is CLEAN and mergeable, and there are no unresolved review threads.

AI-assisted review disclosure: Codex reviewed the exact head, the full feedback history, the final parser delta, current CI, and thread state. Astro-Han authorized this approval and made the final merge decision.

<details>
<summary>中文说明</summary>

精确 head `d810079` 已通过审查。summary validation 已收敛到正确的单一权威：只有 checkpoint builder 在完整 predicate 通过后签发 `sections_v1`；write、load、copy 保留 durable format 边界；损坏的 marked checkpoint 会 fail open 并回放 canonical history。最新提交修复了最后的缩进与 bare ATX heading 骨架绕过，没有引入并行 parser。

当前没有剩余 P0-P2；CI 全绿、merge state CLEAN、无未解决 review thread。

</details>

me2seeks:
Addressed the remaining ATX-heading P2 in a focused cross-fork follow-up because the #3039 head repository does not grant me direct push permission:

https://github.com/UncertaintyDeterminesYou4ndMe/maka-agent/pull/1

The patch uses one CommonMark-compatible matcher for both non-required-H2 transitions and heading content exclusion, with an indented/bare heading-only regression. Runtime build, focused summarizer suite (44/44), Biome, and diff check pass locally.

UncertaintyDeterminesYou4ndMe:
@me2seeks Thanks — that follow-up raced with d81007956, which landed in this PR's merged head minutes before the merge and covers the same gap (one heading interpretation across content exclusion, the H2 transition, and additionally the required-section matcher, which now tolerates up to three leading spaces around the exact heading text). I've replied on the fork PR with the comparison and closed it; your single captured `ATX_HEADING` constant is the cleaner shape, so if you'd like to carry that as a small refactor against main, I'd support it.

Generated-by: Claude Code

PR #3040, opened 2026-08-15
fix(runtime): validate history-compact summaries at the write gates (#3029)

### LINKED ISSUE #3029: fix(runtime): history compaction accepts truncated/unstructured LLM summaries without validation

me2seeks:
## Summary

History compaction folds hundreds of RuntimeEvents into one checkpoint summary. Before this change the **only** gate on that summary was `empty_summary`, so a degraded provider response — a section-less fragment ending mid-sentence, a single word, or raw tool-call markup — replaced the folded history and made the continuation model hallucinate implementation details (see [#3029](https://github.com/maka-agent/maka-agent/issues/3029)).

## Root cause (reproduced)

Incident session `fbdb3fd3`: compaction folded **742 events / ~235,503 tokens into a 138-token section-less fragment**, persisted because every non-empty string passed the gate. Three `provider_error` fail-opens preceded the degraded write.

The model returned ~160 tokens of raw tool-call markup (no summary sections). The old gate **accepted** it. (The other failure layer — the summarizer emitting parallel tool calls as separate assistant messages, triggering the `provider_error` retries — is tracked as [#3030](https://github.com/maka-agent/maka-agent/issues/3030) and fixed independently by [#3038](https://github.com/maka-agent/maka-agent/pull/3038).)

## Changes

- **`history-compact-summary-validation.ts`** (new): `validateHistoryCompactSummary()` enforces the summarizer's own structural contract — the required section headings (`## Goal` / `## Progress` / `## Next Steps`) matched **line-anchored**, plus truncation signals (unclosed code fence, trailing continuation punctuation). Section names are the single source of truth for both the prompt and the validator.
- **Both checkpoint write gates enforce it** and fail open:
  - `writeHistoryCompactCheckpoint` → throws `HistoryCompactSummarizerError('malformed_summary')`, mapped to `failOpenReason: 'malformed_summary'`.
  - `planMidTurnCapacityCompaction` → returns `fail_open` with `diagnosticReason: 'malformed_summary'`.
- `'malformed_summary'` added to the failure-reason union.
- Tests: validation unit tests (incident fragment verbatim, raw DeepSeek DSML output, per-section missing, truncation, heading-vs-prose, near-match `## Goals`), a direct `writeHistoryCompactCheckpoint` gate test, and a backend-level fail-open regression test. Fixtures updated to the conforming shape.

## Verification

- `npm --workspace @maka/runtime run test` → **2805 pass / 0 fail**
- `npm run typecheck` → clean (all packages)
- `npx biome check` → clean (2298 files)
- Post-fix empirical checks: the incident fragment now fails open with `malformed_summary`; the non-compliant output is rejected with `missing_sections`.

## Review notes

- **First-principles + Occam's-razor dual review** was run on the first two commits; findings applied in the third commit (line-anchored headings instead of substring match, compile-time pinning of the required-section list, direct coverage of the checkpoint write gate, removal of tautological/redundant tests).
- **Deliberately deferred** (not in scope of #3029): validating summaries at **load** time. The loader sees a heterogeneous history (checkpoints written by older binaries), so the same validator would reject legacy valid-but-sectionless summaries; that needs a version-aware or migration approach and will be filed separately.

## Related

- Closes #3029

### DIFF
--- a/packages/runtime/src/history-compact-summary-validation.ts
+++ b/packages/runtime/src/history-compact-summary-validation.ts
@@ -1,0 +1,58 @@
+/**
+ * Structural contract for history-compact checkpoint summaries.
+ *
+ * A checkpoint summary is the only survivor of a history fold: once written,
+ * the covered RuntimeEvents are replaced by this text alone. The summarizer
+ * prompt (history-compact-summarizer.ts) declares a sectioned contract, but a
+ * degraded provider can return anything non-empty — a single word, raw tool
+ * markup, or a fragment ending mid-sentence — and before this gate existed the
+ * pipeline persisted all of it (see #3029). Both checkpoint write paths call
+ * `validateHistoryCompactSummary` and fail open on a rejection, so a bad
+ * summary never replaces folded history.
+ *
+ * The section names are the single source of truth: the summarizer prompt is
+ * built from them, and validation reuses the same constants.
+ */
+
+/** Every section the summarization prompt asks for, in order. */
+export const HISTORY_COMPACT_SUMMARY_SECTIONS = [
+  '## Goal',
+  '## Progress',
+  '## Key Decisions',
+  '## Next Steps',
+  '## Critical Context',
+] as const;
+
+/**
+ * Sections without which a continuation cannot recover the task: what is being
+ * done, how far it got, and what happens next. `## Key Decisions` and
+ * `## Critical Context` are valuable but not load-bearing, so a summary that
+ * omits them is thin rather than malformed.
+ */
+export const HISTORY_COMPACT_REQUIRED_SECTIONS = [
+  '## Goal',
+  '## Progress',
+  '## Next Steps',
+] as const;
+
+/** Terminal punctuation that indicates the output was cut off mid-thought. */
+const TRUNCATED_TAIL_PATTERN = /[:：,，、;；(（`—]\s*$/u;
+
+/** Why a summary failed validation, for diagnostics and tests. */
+export type HistoryCompactSummaryRejection = 'missing_sections' | 'truncated';
+
+/**
+ * Returns why the summary is malformed, or undefined when it satisfies the
+ * contract. Callers map any rejection to `malformed_summary` and fail open.
+ */
+export function validateHistoryCompactSummary(
+  summary: string,
+): HistoryCompactSummaryRejection | undefined {
+  const missing = HISTORY_COMPACT_REQUIRED_SECTIONS.filter((section) => !summary.includes(section));
+  if (missing.length > 0) return 'missing_sections';
+  // An odd number of fences means the output stops inside a code block.
+  const fenceCount = (summary.match(/^```/gm) ?? []).length;
+  if (fenceCount % 2 === 1) return 'truncated';
+  if (TRUNCATED_TAIL_PATTERN.test(summary)) return 'truncated';
+  return undefined;
+}
--- a/packages/runtime/src/history-compact-summary-validation.ts
+++ b/packages/runtime/src/history-compact-summary-validation.ts
@@ -33,7 +33,20 @@ export const HISTORY_COMPACT_REQUIRED_SECTIONS = [
   '## Goal',
   '## Progress',
   '## Next Steps',
-] as const;
+] as const satisfies readonly (typeof HISTORY_COMPACT_SUMMARY_SECTIONS)[number][];
+
+/**
+ * Line-anchored heading matchers: a required section must appear as its own
+ * heading line. A substring match would also accept `## Goals` or a prose
+ * mention of `## Goal`, neither of which follows the summarizer contract.
+ */
+const REQUIRED_SECTION_HEADING_PATTERNS = HISTORY_COMPACT_REQUIRED_SECTIONS.map(
+  (section) => new RegExp(`^${escapeRegExp(section)}\\b`, 'm'),
+);
+
+function escapeRegExp(text: string): string {
+  return text.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
 
 /** Terminal punctuation that indicates the output was cut off mid-thought. */
 const TRUNCATED_TAIL_PATTERN = /[:：,，、;；(（`—]\s*$/u;
@@ -48,8 +61,9 @@ export type HistoryCompactSummaryRejection = 'missing_sections' | 'truncated';
 export function validateHistoryCompactSummary(
   summary: string,
 ): HistoryCompactSummaryRejection | undefined {
-  const missing = HISTORY_COMPACT_REQUIRED_SECTIONS.filter((section) => !summary.includes(section));
-  if (missing.length > 0) return 'missing_sections';
+  if (REQUIRED_SECTION_HEADING_PATTERNS.some((pattern) => !pattern.test(summary))) {
+    return 'missing_sections';
+  }
   // An odd number of fences means the output stops inside a code block.
   const fenceCount = (summary.match(/^```/gm) ?? []).length;
   if (fenceCount % 2 === 1) return 'truncated';
--- a/packages/runtime/src/__tests__/ai-sdk-backend.test.ts
+++ b/packages/runtime/src/__tests__/ai-sdk-backend.test.ts
@@ -5302,6 +5302,97 @@ describe('AiSdkBackend model history', () => {
     );
   });
 
+  test('V2 malformed summary fails open without recording a checkpoint and reports malformed_summary', async () => {
+    const model = completionModel();
+    const storedMessages: StoredMessage[] = [];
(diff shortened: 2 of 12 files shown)


UncertaintyDeterminesYou4ndMe:
@me2seeks Following up on Astro-Han's review of #3039, which flagged this PR and #3039 as uncoordinated parallel implementations of the #3029 fix (and #3046 stacking on this one) — proposing an explicit split so one validation contract survives:

- #3039 now hosts the validation in a dedicated `history-compact-summary-validation.ts`: the mandated format template (the summarization prompt is built from it), the required-section/fence/truncation scanner, the size floor, and a writer-agnostic `findCheckpointSummaryDefect(summary, foldContext?)` predicate returning a granular reason. The reviewed bypasses (fence closing width, template echo) are fixed there with regression tests, plus an end-to-end fail-open test at the backend seam.
- With the summarizer rejecting malformed completions at the source, a second gate at checkpoint write time guards no additional path — so I'd suggest withdrawing this PR's validator in favor of #3039, the same way you deferred #3040's replay fix to #3038.
- Your #3046 (load-time rejection of already-persisted checkpoints, #3041) is complementary, not competing — #3039 explicitly declared that surface a follow-up. It could consume the same `findCheckpointSummaryDefect` predicate (probably without the fold-size context, since the covered span estimate is not durable at load time), which would leave exactly one definition of "well-formed summary" across write and load.

Happy to adjust the predicate's shape if #3046 needs something different at load time — deferring to you and the maintainers on the final call.

Generated-by: Claude Code

packages/runtime/src/history-compact-summary-validation.ts
```diff
@@ -0,0 +1,72 @@
...
+export type HistoryCompactSummaryRejection = 'missing_sections' | 'truncated';
+
+/**
+ * Returns why the summary is malformed, or undefined when it satisfies the
+ * contract. Callers map any rejection to `malformed_summary` and fail open.
+ */
+export function validateHistoryCompactSummary(
+  summary: string,
+): HistoryCompactSummaryRejection | undefined {
+  if (REQUIRED_SECTION_HEADING_PATTERNS.some((pattern) => !pattern.test(summary))) {
```
Astro-Han, line 64: **P1 — Do not admit empty or fenced template summaries.** These independent global heading regexes accept `## Goal
## Progress
## Next Steps` with no content, and they also accept all three headings inside a completed fenced block because an even fence count is considered valid. Either result can replace the folded RuntimeEvents with effectively no continuation facts. Please use one fence-aware scanner that requires the sections in order, outside backtick/tilde fences, with non-empty content; add focused title-only and fenced-template tests that assert no checkpoint is recorded.

Astro-Han (commented):
Thanks for addressing a real irreversible boundary: a history-compaction summary can become the only replay survivor for a large event prefix, so malformed output must fail open. I reviewed this head with two independent `@reviewer` passes plus a read-only `ollama-cloud/deepseek-v4-flash:high` pass.

The write-gate direction is correct, but the current implementation does not yet make semantic admission authoritative. The validator accepts headings inside a completed fence and accepts empty sections, while restart/load still admits older shape-valid malformed checkpoints without the new policy. Current required checks are also failing.

The simpler first-principles design is one scanner and one admission policy: parse ordered, non-empty required sections outside backtick/tilde fences; invoke it at the durable checkpoint authority; and reuse the same policy when loading, repairing projections, or copying checkpoints. On rejection, replay the canonical events. #3039 changes the same authority and already pursues a stronger scanner, so please consolidate these efforts rather than maintain parallel validators.

No local test suite was run during this review; conclusions are based on source, test, and current CI inspection. Codex coordinated the independent passes and performed the final adjudication; external-model output was treated as unverified until checked against the code.

<details>
<summary>中文摘要</summary>

感谢修复这个不可逆边界：history compaction summary 会成为大段事件历史的唯一 replay survivor，因此 malformed output 必须 fail-open。

当前 write gate 方向正确，但 semantic admission 还不是权威：validator 会接受完整 fence 内的 headings 和空 section，restart/load 也仍会接纳旧的、shape 合法但语义损坏的 checkpoint；当前 required checks 也在失败。

更符合第一性原理和奥卡姆剃刀的方案是只保留一个 scanner 与一套 admission policy：只接受 fence 外、顺序正确且非空的 required sections；在 durable checkpoint authority 执行；load、projection repair 和 copy 都复用同一 policy；拒绝时回放 canonical events。#3039 正在修改同一权威且已有更强 scanner，建议合并方案，不要维护两套 validator。

本次未在本地运行测试套件；结论来自源码、测试和当前 CI 检查。Codex 协调了独立审查并完成最终判断，外部模型输出在核对代码前均视为未验证输入。
</details>

me2seeks:
Closing in favor of #3039 as the single history-compaction summary-validation authority, following the maintainer coordination. Keeping a second write-gate validator would recreate the overlapping contract. The remaining #3039 load/replay review fix is prepared at me2seeks commit 48de7eb237 for its author to cherry-pick.

UncertaintyDeterminesYou4ndMe:
Thanks for the clean handoff — 48de7eb237 cherry-picked into #3039 as 4b05575c4 with your authorship preserved. The legacy-tolerant load policy (quarantine truncation only, keep section-less-but-intact legacy summaries loadable) is exactly the right version-aware line, and the shared-scanner reuse means write admission and load quarantine cannot disagree about fence semantics. #3039 now carries the single validation contract end to end.

Generated-by: Claude Code

PR #3046, opened 2026-08-15
fix(runtime): reject truncated history-compact summaries at load time (#3041)

### LINKED ISSUE #3041: Validate history-compact checkpoint summaries at load time (repair poisoned sessions)
## Background

#3029 added validation at the two checkpoint **write** gates: a malformed/truncated LLM summary can no longer replace folded history going forward. The fix is write-only.

## Gap

A checkpoint whose summary was persisted **before** the gate (e.g. the incident checkpoint `hcheckpoint-981ceab8…` in session `fbdb3fd3`) still loads and replays: `loadLatestHistoryCompactCheckpointFromRunLedger` only checks `validateHistoryCompactCheckpointShape` (non-empty summary), not the summary contract. The already-poisoned session keeps serving the fragment until a later successful fold happens to supersede it.

## Why not just reuse the same validator

Load-time validation is **asymmetric** to write-time: at write time the summary was just produced by the sectioned summarizer prompt, so a missing `## Goal` heading means degraded output. At load time the ledger may contain checkpoints written by older binaries whose summaries never followed the sectioned contract but are still usable. Applying `validateHistoryCompactSummary` verbatim would silently discard those and force a fallback to raw events (safe but wasteful/surprising), or loop re-compaction.

## Suggested direction

- Only validate checkpoints whose shape/metadata indicates the sectioned writer (e.g. a schema `version` bump or an explicit `summaryFormat` field), leaving legacy checkpoints to the existing shape check.
- Or validate but treat "section-less yet non-trivially-long" as a distinct case from "fragment/truncated" and migrate rather than discard.
- Add load-path tests mirroring the existing `history-compact-checkpoint.test.ts` fixtures (which currently use section-less synthetic summaries — they would all need to keep passing for legacy).

## Related

- #3029 (write-gate validation, merged in PR #3040)
- #3030 (parallel tool-call message merge)

me2seeks:
## Summary

Closes #3041. #3029 (PR #3040) added validation at the two checkpoint **write** gates, but a checkpoint whose truncated summary was persisted **before** that gate — incident checkpoint `hcheckpoint-981ceab8…` in session `fbdb3fd3`, plus two more in `d282f6ac` — still loads and replays. This PR closes the load side.

## What changed

- Extracts a writer-agnostic `isHistoryCompactSummaryTruncated` from `validateHistoryCompactSummary` (fence count + tail punctuation).
- Applies it in **both** load paths of `loadLatestHistoryCompactCheckpointFromRunLedger`: the bounded projection fast path and the canonical ledger recovery path.
- A rejected checkpoint falls back to the previous valid checkpoint (or raw events when none), and the stale projection is repaired via `repairEventProjection`. No data is lost: raw events are append-only, and the next high-water fold rewrites a fresh checkpoint.
- **Deliberately does NOT apply the `missing_sections` check at load**: legacy checkpoints predate the sectioned summarizer contract and remain usable without sections. Only truncation is load-bearing regardless of writer era.

## Review fixes (first-principles + Occam double review)

- **MAJOR**: the shared tail heuristic counted a trailing backtick as truncation, which would also match a *closed* ``` fence. At load time that would reject complete legacy summaries ending in a code fence and force a re-compaction. Dropped the backtick from the tail class (the even fence count already proves a closed block), added the `…` signal, and renamed the predicate to `isHistoryCompactSummaryTruncated`.
- Trimmed lockstep unit tests to three distinct contract points.

## Validation

- runtime suite: **2816 tests, 2810 pass, 0 fail, 6 skipped**; biome + tsc clean.
- New tests cover: ledger fallback on a truncated checkpoint, projection repair from the canonical ledger, and the closed-fence false-positive regression.

## Stacking

Stacked on **#3040** (`maka/history-compact-summary-validation`), which is not yet merged; this branch contains those 3 commits plus 2 new ones. Merge **after** #3040, at which point the diff reduces to the 2 new commits.

### DIFF
--- a/packages/runtime/src/history-compact-summary-validation.ts
+++ b/packages/runtime/src/history-compact-summary-validation.ts
@@ -48,8 +48,15 @@ function escapeRegExp(text: string): string {
   return text.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
 }
 
-/** Terminal punctuation that indicates the output was cut off mid-thought. */
-const TRUNCATED_TAIL_PATTERN = /[:：,，、;；(（`—]\s*$/u;
+/**
+ * Terminal punctuation that indicates the output was cut off mid-thought.
+ * Best-effort by construction: a fragment ending mid-word (no punctuation) is
+ * undetectable, and code fences are handled separately by the fence count in
+ * `isHistoryCompactSummaryTruncated`. A trailing backtick is deliberately NOT
+ * here — it would also match a closed ``` fence, which is completion, not
+ * truncation.
+ */
+const TRUNCATED_TAIL_PATTERN = /[:：,，、;；…(（—]\s*$/u;
 
 /** Why a summary failed validation, for diagnostics and tests. */
 export type HistoryCompactSummaryRejection = 'missing_sections' | 'truncated';
@@ -57,12 +64,15 @@ export type HistoryCompactSummaryRejection = 'missing_sections' | 'truncated';
 /**
  * Writer-agnostic truncation detection: an odd number of code fences means the
  * output stops inside a code block, and terminal dangling punctuation means it
- * stops mid-thought. Legacy checkpoints (written before the sectioned
- * summarizer prompt) never followed the section contract, but a truncated
- * fragment is unusable regardless of writer era, so the load path applies this
- * check to every checkpoint summary (#3041).
+ * stops mid-thought. A complete summary that ends with a closed ``` fence is
+ * NOT truncated (the even fence count already proves the block closed), so the
+ * tail punctuation deliberately excludes a trailing backtick. Legacy
+ * checkpoints (written before the sectioned summarizer prompt) never followed
+ * the section contract, but a truncated fragment is unusable regardless of
+ * writer era, so the load path applies this check to every checkpoint summary
+ * (#3041).
  */
-export function detectHistoryCompactSummaryTruncation(summary: string): boolean {
+export function isHistoryCompactSummaryTruncated(summary: string): boolean {
   // An odd number of fences means the output stops inside a code block.
   const fenceCount = (summary.match(/^```/gm) ?? []).length;
   if (fenceCount % 2 === 1) return true;
@@ -79,6 +89,6 @@ export function validateHistoryCompactSummary(
   if (REQUIRED_SECTION_HEADING_PATTERNS.some((pattern) => !pattern.test(summary))) {
     return 'missing_sections';
   }
-  if (detectHistoryCompactSummaryTruncation(summary)) return 'truncated';
+  if (isHistoryCompactSummaryTruncated(summary)) return 'truncated';
   return undefined;
 }
--- a/packages/runtime/src/history-compact-summary-validation.ts
+++ b/packages/runtime/src/history-compact-summary-validation.ts
@@ -55,6 +55,21 @@ const TRUNCATED_TAIL_PATTERN = /[:：,，、;；(（`—]\s*$/u;
 export type HistoryCompactSummaryRejection = 'missing_sections' | 'truncated';
 
 /**
+ * Writer-agnostic truncation detection: an odd number of code fences means the
+ * output stops inside a code block, and terminal dangling punctuation means it
+ * stops mid-thought. Legacy checkpoints (written before the sectioned
+ * summarizer prompt) never followed the section contract, but a truncated
+ * fragment is unusable regardless of writer era, so the load path applies this
+ * check to every checkpoint summary (#3041).
+ */
+export function detectHistoryCompactSummaryTruncation(summary: string): boolean {
+  // An odd number of fences means the output stops inside a code block.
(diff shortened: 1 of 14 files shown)


packages/runtime/src/history-compact-summary-validation.ts
```diff
@@ -0,0 +1,94 @@
...
+}
+
+/**
+ * Returns why the summary is malformed, or undefined when it satisfies the
+ * contract. Callers map any rejection to `malformed_summary` and fail open.
+ */
+export function validateHistoryCompactSummary(
+  summary: string,
+): HistoryCompactSummaryRejection | undefined {
+  if (REQUIRED_SECTION_HEADING_PATTERNS.some((pattern) => !pattern.test(summary))) {
```
Astro-Han, line 89: **P1 — Validate section structure and content, not independent heading presence.** This accepts a title-only skeleton, required headings out of order, and all headings inside a completed code fence; each can then replace the folded RuntimeEvents with effectively no continuation facts. Please replace the independent regexes with one fence-aware scanner that requires the sections in order, outside fences, with non-placeholder content, and assert both writer paths do not record heading-only/fenced-template summaries.

packages/runtime/src/history-compact-summary-validation.ts
```diff
@@ -0,0 +1,94 @@
...
+ * NOT truncated (the even fence count already proves the block closed), so the
+ * tail punctuation deliberately excludes a trailing backtick. Legacy
+ * checkpoints (written before the sectioned summarizer prompt) never followed
+ * the section contract, but a truncated fragment is unusable regardless of
+ * writer era, so the load path applies this check to every checkpoint summary
+ * (#3041).
+ */
+export function isHistoryCompactSummaryTruncated(summary: string): boolean {
+  // An odd number of fences means the output stops inside a code block.
+  const fenceCount = (summary.match(/^```/gm) ?? []).length;
```
Astro-Han, line 77: **P2 — Quarantine legacy fences across Markdown delimiter forms.** This count misses indented backtick fences and every `~~~` fence; after removing trailing backtick from the tail heuristic, an unfinished legacy summary using those forms is accepted by projection/ledger recovery. Please have the shared scanner track backtick/tilde family, indentation and opener width, then add projection and canonical-ledger recovery cases for valid and unterminated fences.

Astro-Han (commented):
Thanks for extending malformed-summary protection to restart/load and for fixing the false positive on a properly closed fence. I reviewed this head with two independent `@reviewer` passes plus a read-only `ollama-cloud/deepseek-v4-flash:high` pass.

The problem is correctly defined: a checkpoint that replaces RuntimeEvents must be safe both when written and when selected after restart. The current implementation still has two bypasses in the shared predicate: heading presence is not semantic structure, and the legacy fence detector recognizes only unindented backtick fences. That means the new load quarantine can still accept a title-only/fenced template or an unfinished tilde/indented block.

The first-principles simplification is one line scanner with two explicit policies: strict admission for new text checkpoints (ordered, non-empty required sections outside fences) and conservative truncation quarantine for legacy text checkpoints (backtick/tilde family, indentation, delimiter width). Record and load/replay should reuse it. When rebasing to current main, guard this behind the text-checkpoint discriminator so provider-native checkpoints keep their own schema/continuation authority. #3039 already contains the stronger structural scanner; the optimal result is to consolidate its scanner with this PR's restart quarantine, not land parallel validators.

Current CI is green. No local test suite was run during this review; conclusions are based on source, test, current-main, and CI inspection. Codex coordinated the independent passes and performed the final adjudication; external-model output was treated as unverified until checked against the code.

<details>
<summary>中文摘要</summary>

感谢把 malformed-summary 保护扩展到 restart/load，并修复合法闭合 fence 的误报。问题定义正确：替换 RuntimeEvents 的 checkpoint 在写入时和 restart 后被选择时都必须安全。

当前 shared predicate 仍有两个 bypass：heading presence 不是 semantic structure；legacy fence detector 只识别无缩进的 backtick fence。因此 title-only/fenced template，以及未闭合的 tilde/缩进 fence 仍可能通过新的 load quarantine。

最小方案是一个 line scanner、两套明确 policy：新 text checkpoint 严格要求 fence 外、按序、非空 required sections；legacy text checkpoint 保守识别 backtick/tilde、缩进和 delimiter width。record 与 load/replay 复用。rebase 到 current main 时必须先按 text/provider-native 分流。#3039 已有更强结构 scanner，最优解是与本 PR 的 restart quarantine 合并，而不是并行落两套 validator。

当前 CI 全绿。本次未运行本地测试套件；结论来自源码、测试、current-main 与 CI 检查。外部模型输出在核对代码前均视为未验证输入。
</details>

me2seeks:
Closing in favor of #3039 as the single history-compaction summary-validation authority, following the maintainer coordination. The load/replay behavior from this PR is consolidated in reviewed commit 48de7eb237, prepared for the #3039 author to cherry-pick, so this parallel validator should not remain open.

PR #3049, opened 2026-08-15
fix(runtime): group parallel tool calls in history summarizer replay

### LINKED ISSUE #3030: fix(runtime): history summarizer emits parallel tool calls as separate assistant messages, rejected by strict OpenAI-compatible providers

1625567290:
## Summary

`replayPlanItemsToModelMessages` emitted every `tool_call` replay item as its own assistant message. Parallel calls therefore looked like:

```
assistant [call A]
assistant [call B]
tool [result A]
tool [result B]
```

Strict OpenAI-compatible providers reject that (DeepSeek 400: previous `tool_calls` not yet answered). Compaction then fail-opens as `provider_error`.

Consecutive `tool_call` items are now one assistant message with N tool-call parts, followed by N tool messages. Sequential rounds that already have answers between them stay separate. The primary replay materializer is unchanged.

Fixes #3030

## Verification

- `npx tsx --test packages/runtime/src/__tests__/history-compact-summarizer.test.ts` — 10/10
- New cases: two parallel reads → one assistant + two tool messages; sequential rounds stay two assistants
- Both cases assert every assistant `tool-call` id is answered by the immediately following tool messages

Not run: live DeepSeek generateText. The unit test locks the rejected message shape.

## Checklist

- [x] Yes — history summarizer requests over parallel tool calls now use one assistant message

### DIFF
--- a/packages/runtime/src/__tests__/history-compact-summarizer.test.ts
+++ b/packages/runtime/src/__tests__/history-compact-summarizer.test.ts
@@ -13,10 +13,13 @@ import type { RuntimeEvent, RuntimeEventContent } from '@maka/core/runtime-event
 import { decodeModelCallAttempt, type ModelCallAttempt } from '@maka/core/model-call-attempt';
 import { ProviderRequestTracker } from '../provider-request-telemetry.js';
 import type { HistoryCompactSummaryInput } from '../ai-sdk-compaction-contract.js';
+import type { ModelMessage } from '../model-protocol.js';
 import {
   buildLlmHistorySummarizer,
+  replayPlanItemsToModelMessages,
   type AiSdkGenerateTextLike,
 } from '../history-compact-summarizer.js';
+import { buildRuntimeEventModelReplayPlan } from '../model-history.js';
 import { buildHistoryCompactCheckpoint } from '../history-compact-checkpoint.js';
 
 const ts = 1_700_000_000_000;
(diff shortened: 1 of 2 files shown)


1625567290:
Closing in favor of #3038. That PR groups by step and also covers interleaved tool results; our adjacency grouping is the weaker of the two.

PR #3056, opened 2026-08-15
fix(desktop): keep a remove from destroying a restored task

### LINKED ISSUE #3050: Removing a session should require it to still be archived
> **Correction (2026-08-15):** the original analysis was wrong about where the atomicity is missing. The Host already decides this in one step; the defect is in the desktop client's retry loop. Rewritten accordingly.

## Problem

Deleting a task that is concurrently restored — from a second window or another client — deletes it permanently anyway, and the caller is told it was removed.

Found by @M4n5ter reviewing #2985, reproduced deterministically by restoring the second task after the first removal completed.

## What already works

The compare-and-set that should prevent this is present end to end:

- `session-retirement-coordinator.ts` `#remove` compares `target.revision !== input.expectedRevision` inside the `#withStableFamily` admission lock and returns `revision_conflict`.
- `sqlite-session-metadata-store.ts` `removeVersioned` re-checks `metadataVersion` inside the same SQL transaction as the `DELETE`.
- `setLifecycleVersioned` writes through `updateHeaderSync`, so unarchiving bumps `metadataVersion`.

A concurrent restore therefore makes the first remove fail. Runtime Host is not missing a precondition, and `session.remove` does not need a new one.

## Where it breaks

`apps/desktop/src/main/runtime-host-client.ts`, `removeSession`:

```ts
for (let attempt = 0; attempt < MAX_SESSION_REVISION_ATTEMPTS; attempt += 1) {
  const current = await this.#requireSession(sessionId);
  const result = await this.request("session.remove", { sessionId, expectedRevision: current.revision });
  if (result.kind === "removed") return;
}
```

The conflict is treated as a stale read and the delete is replayed with the fresh revision, up to 8 times. Replay is right for rename and configuration writes — the write means the same thing at any revision. It is wrong for remove: the conflict is precisely the signal that the task was touched after the caller decided to destroy it, and the replay carries out a destruction whose premise no longer holds. The loop also widens the window from "between two calls" to "the whole retry".

## Direction

Lifecycle writes always bump `metadataVersion`, and the version check and the `DELETE` share one transaction. So "read as archived at revision R" plus "removed with `expectedRevision` R" implies "still archived when deleted" — the loop only has to re-assert the precondition on each re-read, not carry a new one over the wire.

`removeSession` should stop replaying blindly: on each attempt, give up if the re-read session is no longer archived, and return an outcome distinguishable from a failure. `purgeSessions` then reports "restored meanwhile" separately from "failed", counting it as neither removed nor errored.

This stays in `apps/desktop`, outside the Runtime Host protected area. The earlier proposal — an expected lifecycle on `session.remove`, validated by the retirement coordinator — is over-built given the CAS already there, and is dropped.

Open question: whether the single-row delete path, which shares `removeSessionFamily`, wants the same treatment or is adequately covered by its own confirm.

## Impact

Permanent data loss, but it needs concurrent activity: the purge page disables its own controls while a sweep runs, so the trigger is a second window or another client. Not blocking anything currently in flight.

Astro-Han:
## Summary

Deleting a task that was concurrently restored deleted it permanently anyway. The Host-side atomicity that prevents this already exists — `session.remove` and `session.lifecycle.set` both enter `#withStableFamily`, which queues per Session id through the admission gate, and the remove compares the revision on the way in. A restore therefore either lands before that comparison, bumping `metadataVersion` and rejecting the remove, or waits until the retirement has finished. It cannot land between the check and the delete.

The Desktop Client threw that rejection away. `removeSession` treated the conflict as a stale read and replayed the delete at the fresh revision, up to eight times. Replay is right for a rename or a configuration patch — the write means the same thing either way. It is wrong for a remove: the conflict *is* the signal that the task was touched after the caller decided to destroy it, so the replay executed a destruction whose premise no longer held, and the retry widened the race from the gap between two calls to the whole loop.

`removeSession` now takes the premise the caller decided on. With `requireArchived`, each fresh read re-asserts that the task is still archived and answers `restored` instead of replaying. Single-row delete carries the premise of the row it was raised from, so deleting an active task from the rail is unaffected, as is copy cleanup, which states no premise.

The bulk sweep in 设置 › 活动 › 已归档任务 used to decide an id's fate in two places: it skipped anything its own catalog snapshot no longer showed as archived, and separately reported what the Host answered `restored`. That snapshot is one observer of state the Host owns, a second window can outdate it in either direction, and it cannot tell a restored task from one already deleted elsewhere — so the filter is gone rather than taught to classify. Every id now goes to the delete, which already distinguishes all four outcomes: still archived removes, restored answers `restored`, already gone rejects and settles as removed against the catalog, anything else is an error with a reason. The toast reports failures and kept tasks independently instead of choosing which single fact to say.

No Runtime Host or protocol change: the serialization and compare-and-set this relies on are already there. The issue's original "Why this belongs to the Host" and "Proposed direction" sections proposed an expected-lifecycle field on `session.remove`; that is over-built while Desktop is the only consumer, and is being corrected separately. It becomes the right call when a second client can delete — see the follow-up note below.

Fixes #3050

## Verification

Reproduced first on unmodified code, then re-run after the fix:

- `apps/desktop/src/main/__tests__/runtime-host-client-uds.test.ts` — the whole main-process chain over real UDS framing. A restore lands between the Client's read and its remove; on `main` the sweep deleted the task (`actual: undefined, expected: 'restored'`, and the catalog came back empty), and now it answers `restored`, the task is still listed as active, and no `deleted` retirement is emitted for it.
- `apps/desktop/src/main/__tests__/runtime-host-client-operations.test.ts` — the retry loop itself: abandons on a restore, still retries through churn that left the task archived, still removes a task that was never archived.
- `apps/desktop/src/main/__tests__/app-shell-session-purge.test.ts` — every id reaches the delete rather than being filtered against the renderer's snapshot; restored tasks are reported apart from removed and failed and keep their renderer state; single-row delete carries the premise of its own row. Mutation-checked: restoring the snapshot filter turns three of these red.

Ran locally: those suites plus `app-shell-session-row-actions-revisions` (35 tests, all passing), `npm --workspace apps/desktop run typecheck`, and Biome lint over `apps/desktop/src` and `apps/desktop/stories`. Note `npm run format` is a no-op here — the formatter excludes `apps/desktop/**` ([biome.jsonc](../blob/main/biome.jsonc)); lint does cover these files. Not run: the repository-wide suite, E2E, and Storybook visual smoke — left to CI. The two-window GUI repro described in the issue was not performed by hand; the UDS test above stands in for it at the seam where the defect lives.

## Review focus

Data deletion — this needs independent human review, no self-merge fast path.

Two judgement calls worth checking:

1. `requireArchived` is the caller's to state rather than something `removeSession` assumes. An active task deleted from the rail, and revision-copy cleanup, have no archived premise to lose; requiring one would break both.
2. The sweep no longer pre-filters against the renderer's catalog. This costs one IPC round trip per already-restored task and removes the second source of truth that made the previous reporting incomplete.

## Known gaps

- The UDS test scripts the Host side by hand to produce the conflict; the real `SessionRetirementCoordinator` racing a real `session.lifecycle.set` has no integration coverage. That belongs in `packages/runtime-host` and is out of scope here.
- The Host still has no concept of "restored" — the precondition lives in the Desktop client. A second `session.remove` consumer (CLI, bots) could reintroduce #3050 with no test failing. Worth revisiting under #3050 when that consumer appears.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/src/main/__tests__/app-shell-session-purge.test.ts
+++ b/apps/desktop/src/main/__tests__/app-shell-session-purge.test.ts
@@ -33,8 +33,12 @@ function restored(id: string): SessionSummary {
 
 type SweepHarness = {
   removed: string[];
+  /** Each `remove` call as `[sessionId, requireArchived]`. */
+  removeOptions: Array<[string, boolean]>;
   cleared: string[];
   selections: Array<string | undefined>;
+  /** Titles of the success toasts a row action raised. */
+  toasts: string[];
   listCalls: number;
 };
 
@@ -50,6 +54,8 @@ function installWindow(
     surviving?: readonly SessionSummary[];
     /** Runs after each accepted removal, to model what another client did meanwhile. */
     onRemove?: (sessionId: string) => void;
+    /** Ids the Host answers `restored` for, standing in for a lost archived premise. */
+    restoredIds?: readonly string[];
   } = {},
 ): () => void {
   const target = globalThis as unknown as { window?: unknown };
@@ -60,10 +66,13 @@ function installWindow(
     value: {
       maka: {
         sessions: {
-          remove: async (id: string) => {
+          remove: async (id: string, removeOptions?: { requireArchived?: boolean }) => {
+            harness.removeOptions.push([id, removeOptions?.requireArchived === true]);
             if (options.rejectIds?.includes(id)) throw new Error(`busy:${id}`);
+            if (options.restoredIds?.includes(id)) return 'restored';
             harness.removed.push(id);
             options.onRemove?.(id);
+            return 'removed';
           },
           list: async () => {
             harness.listCalls += 1;
@@ -101,12 +110,25 @@ function createActions(input: {
       input.activeIdRef.current = id;
     },
     setMessages: () => undefined,
-    toastApi: { success: () => undefined, error: () => undefined, confirm: async () => true },
+    toastApi: {
+      success: (title: string) => {
+        input.harness.toasts.push(title);
+      },
+      error: () => undefined,
+      confirm: async () => true,
+    },
   });
 }
 
 function harness(): SweepHarness {
-  return { removed: [], cleared: [], selections: [], listCalls: 0 };
+  return {
+    removed: [],
+    removeOptions: [],
+    cleared: [],
+    selections: [],
+    toasts: [],
+    listCalls: 0,
+  };
 }
 
 describe('purgeSessions', () => {
@@ -124,7 +146,18 @@ describe('purgeSessions', () => {
     const outcome = await actions.purgeSessions(['a-v2', 'b']).finally(restore);
 
     assert.deepEqual(h.removed, ['a-v2', 'b']);
-    assert.deepEqual(outcome, { removed: 2, remaining: [], verified: true, firstError: undefined });
+    assert.deepEqual(outcome, {
+      removed: 2,
+      remaining: [],
+      restored: [],
+      verified: true,
+      firstError: undefined,
+    });
+    // Every delete in a sweep carries the archived premise the confirm named.
+    assert.deepEqual(h.removeOptions, [
+      ['a-v2', true],
+      ['b', true],
+    ]);
     // The family goes, not just the representative, and the open member of it
     // stops being the active session.
     assert.deepEqual(h.cleared.sort(), ['a', 'a-v2', 'b']);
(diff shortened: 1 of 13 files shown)


M4n5ter (approved):
LGTM. Re-reviewed after rebasing onto the latest main; the conflict resolution preserves the host-aware session routing and the restored-task deletion guard. Targeted tests, Desktop typecheck, and Biome checks pass.

PR #3066, opened 2026-08-15
refactor(eval): make framework selection single-source

### LINKED ISSUE #3036: refactor(eval): make framework selection single-source
## Problem

The Eval harness currently selects Harbor or Pier twice:

1. The validated framework is passed to `run_trial.py` through argv.
2. The same value is injected as `MAKA_EVAL_FRAMEWORK`, which `relay_agent.py` reads to select its base class.

These values originate from the same TypeScript call but travel through separate contracts. A mismatch would let the trial runner load one framework while the relay imports the other framework's `BaseAgent`.

The environment variable is also treated as a reserved public preparation value even though framework selection is already encoded in the validated command.

## Proposed change

Make the validated `run_trial.py` framework argument the single authority:

- validate the framework argument before importing framework modules;
- provide the validated selection internally before the relay module is imported;
- remove TypeScript-side injection and reservation of `MAKA_EVAL_FRAMEWORK`;
- retain one shared Harbor/Pier executor and relay.

## Acceptance criteria

- [ ] Framework selection has one authoritative input.
- [ ] The trial runner and relay cannot select different frameworks.
- [ ] `harness-executor.ts` no longer injects or reserves `MAKA_EVAL_FRAMEWORK`.
- [ ] Invalid framework values fail before trial or relay execution.
- [ ] Harbor and Pier relay tests cover the resulting contract.
- [ ] Focused Eval tests, type checking, and formatting pass.

## Non-goals

- Adding another evaluation framework.
- Creating framework-specific execution workflows.
- Changing pinned framework versions.
- Changing Eval result or artifact semantics.
- Replacing the shared Harbor/Pier relay.

## Context

Verified on `main` at `1a71db5c0fa85771330354360388ecf47ba95c2e`.

Relevant code:

- `packages/eval/src/harness-executor.ts`
- `packages/eval/harbor/run_trial.py`
- `packages/eval/harbor/relay_agent.py`

---

*This issue was identified and prepared with AI assistance (Codex). A human contributor reviewed the analysis and owns the submission.*

1625567290:
## Summary

Harbor or Pier was selected twice: once as the validated `run_trial.py` argv, and again via `MAKA_EVAL_FRAMEWORK` for `relay_agent.py`. Those contracts could disagree.

`run_trial.py` now validates argv and installs the selection before any framework or relay import. The relay reads that installed value. TypeScript no longer injects or reserves `MAKA_EVAL_FRAMEWORK`.

Fixes #3036

## Verification

- `python3.13` `test_eval_framework.py` 3/3, `test_run_trial_policy.py` 4/4, `test_relay_contract.py` 20/20 (harbor and pier load; env is ignored), `test_relay_artifacts.py` 1/1, `test_relay_lifecycle.py` 12 (2 skip GNU `setsid`)
- `npm --workspace @maka/eval run typecheck` — pass
- `lifecycle-boundaries` including the new no-`MAKA_EVAL_FRAMEWORK` lock — pass
- `biome check` on the touched TypeScript — clean

## Checklist

- [x] Yes — framework selection is only the validated `run_trial.py` argument

### DIFF
--- a/packages/eval/harbor/eval_framework.py
+++ b/packages/eval/harbor/eval_framework.py
@@ -1,0 +1,23 @@
+"""Authoritative Eval harness framework selection.
+
+`run_trial.py` validates the argv framework and installs it here before Harbor,
+Pier, or the shared relay can import. The relay must not read the environment.
+"""
+
+from __future__ import annotations
+
+_FRAMEWORKS = frozenset({"harbor", "pier"})
+_framework: str | None = None
+
+
+def install(framework: str) -> None:
+    if framework not in _FRAMEWORKS:
+        raise RuntimeError("framework must be harbor or pier")
+    global _framework
+    _framework = framework
+
+
+def selected() -> str:
+    if _framework not in _FRAMEWORKS:
+        raise RuntimeError("Eval framework selection is not installed")
+    return _framework
--- a/packages/eval/harbor/relay_agent.py
+++ b/packages/eval/harbor/relay_agent.py
@@ -14,13 +14,13 @@ import tempfile
 from pathlib import Path
 from typing import Any
 
-framework = os.environ.get("MAKA_EVAL_FRAMEWORK")
+from eval_framework import selected
+
+framework = selected()
 if framework == "harbor":
     from harbor.agents.base import BaseAgent
-elif framework == "pier":
-    from pier.agents.base import BaseAgent
 else:
-    raise RuntimeError("MAKA_EVAL_FRAMEWORK must be harbor or pier")
+    from pier.agents.base import BaseAgent
 
 
 class RelayTransportClosed(RuntimeError):
--- a/packages/eval/harbor/run_trial.py
+++ b/packages/eval/harbor/run_trial.py
@@ -100,9 +100,10 @@ def publish_ready_task(namespace: Path, target: Path) -> None:
 
 
 async def run_trial(framework: str, expected_version: str, config_file: Path) -> None:
-    distribution = {"harbor": "harbor", "pier": "datacurve-pier"}.get(framework)
-    if distribution is None:
-        raise RuntimeError("framework must be harbor or pier")
+    from eval_framework import install
+
+    install(framework)
+    distribution = {"harbor": "harbor", "pier": "datacurve-pier"}[framework]
     if importlib.metadata.version(distribution) != expected_version:
         raise FrameworkVersionMismatch
     try:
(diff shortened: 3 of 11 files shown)


1625567290:
Replaced the source-text scan in a1c376cc9.

The new lifecycle test launches a Harbor trial while the parent has `MAKA_EVAL_FRAMEWORK=pier`, then asserts the child environment recorded `framework: null`. That is the observable allowlist contract; a comment in `harness-executor.ts` no longer trips the test.

`lifecycle-boundaries` 16/16, `test_eval_framework.py` 3/3, `test_relay_contract.py` 20/20, `test_run_trial_policy.py` 4/4.

Generated-by: Grok

Astro-Han (approved):
Reviewed against #3036.

The smaller design I wanted (pass the framework into a relay factory) is blocked by Harbor: the trial loads `relay_agent:RelayAgent` by import path, and the base class is chosen at import time. A process-scoped install is forced by that, not invented here. Reading `sys.argv[1]` from the relay would be the same kind of global, with the argv layout as a hidden contract. `eval_framework.py` is the explicit version of that.

`install()` runs before any harbor / pier / relay import on both `run_trial()` and `main()`. Production code no longer reads or injects `MAKA_EVAL_FRAMEWORK`. The launch test that replaced the source-text scan is the right check.

Non-blocking: `test_install_selects_harbor_and_pier` treats re-install in one process as part of the contract; only the test helper `load_relay` does that. The two function-local `from eval_framework import install` lines in `run_trial.py` can move to module scope. `test_relay_contract.py` sets `MAKA_EVAL_FRAMEWORK` and does not restore it.

Approve.

AI-assisted review: Grok 4.6 drafted the first pass. A second pass on opencode-go/deepseek-v4-flash:max ran the focused Python suites, mutated `relay_agent.py` back to env selection to confirm those tests fail, and checked that an unset or invalid framework cannot reach the relay import. I read the issue and diff, grepped the PR branch for `MAKA_EVAL_FRAMEWORK`, and checked how Harbor loads the relay. Unverified by me: I did not rerun those Python suites myself.

Astro-Han:
Fast path: framework selection already lived on the validated `run_trial.py` argument; this removes the unused `MAKA_EVAL_FRAMEWORK` writer/reader so the trial and relay cannot disagree. No new framework, workflow, or result/artifact semantics. Required checks are green. Easy to revert.

I reviewed the final diff and am merging on that basis. Human contributor of record for this review: Astro-Han.

PR #3068, opened 2026-08-15
fix(desktop): stop replaying session remove after a restore

### LINKED ISSUE #3050: Removing a session should require it to still be archived

1625567290:
## Summary

`removeSession` treated a `revision_conflict` as a stale read and retried the delete with the new revision. A concurrent restore (second window) bumps the revision and unarchives the task, so the retry permanently deleted it.

If the first catalog read was archived and a later read is not, the loop now returns `restored` and does not call `session.remove` again. An active-session delete still retries after a rename. `purgeSessions` counts `restored` separately from a failure.

Fixes #3050

## Verification

- `npx tsx --test` on:
  - `runtime-host-client-operations.test.ts` — including first-try remove, archived retry, restore-after-conflict, and active rename retry
  - `app-shell-session-purge.test.ts` — 7/7, including Host-side restore counted separately
  - `app-shell-session-row-actions-revisions.test.ts`
  - `app-shell-first-send-cleanup.test.ts`
- `biome check` on the touched files — clean

Not run: `tsc -p tsconfig.main.json` on this `main` fails in `shell-copy.ts` (`archived-tasks` missing from a settings-section record). That file is outside this change.

## Checklist

- [x] Yes — a concurrently restored archived task is no longer deleted by the remove retry

### DIFF
--- a/apps/desktop/src/main/__tests__/app-shell-first-send-cleanup.test.ts
+++ b/apps/desktop/src/main/__tests__/app-shell-first-send-cleanup.test.ts
@@ -204,6 +204,7 @@ describe('composer first-send cleanup', () => {
         send: async () => Promise.reject(new Error('Skill discovery failed')),
         remove: async (sessionId: string) => {
           removed.push(sessionId);
+          return { kind: 'removed' as const };
         },
       },
     });
@@ -231,6 +232,7 @@ describe('composer first-send cleanup', () => {
         // and the rest of the happy path run after the cleanup window closes.
         remove: async (sessionId: string) => {
           removed.push(sessionId);
+          return { kind: 'removed' as const };
         },
       },
     });
@@ -256,6 +258,7 @@ describe('composer first-send cleanup', () => {
         send: async () => Promise.reject(new Error('Skill discovery failed')),
         remove: async (sessionId: string) => {
           removed.push(sessionId);
+          return { kind: 'removed' as const };
         },
       },
     });
@@ -342,7 +345,7 @@ describe('composer send failure feedback', () => {
       },
       send: async () =>
         Promise.reject(new Error('NO_REAL_CONNECTION:missing_api_key: no ready connection')),
-      remove: async () => undefined,
+      remove: async () => ({ kind: 'removed' as const }),
     },
   });
 
--- a/apps/desktop/src/main/__tests__/app-shell-session-purge.test.ts
+++ b/apps/desktop/src/main/__tests__/app-shell-session-purge.test.ts
@@ -37,6 +37,7 @@ function installWindow(
   harness: SweepHarness,
   options: {
     rejectIds?: readonly string[];
+    restoreIds?: readonly string[];
     surviving?: readonly SessionSummary[];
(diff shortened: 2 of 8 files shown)


1625567290:
Closing in favor of #3056, which already covers the same `removeSession` conflict handling. Sorry for the duplicate.

PR #3069, opened 2026-08-15
fix(desktop): search fixture-seeded transcripts for content hits

### LINKED ISSUE #2305: test(e2e): fixture-seeded transcripts return no content hits from thread search
## What happens

In an e2e-fixture window, `search:thread` returns **title hits but no content hits** for a fixture-seeded transcript. The same window renders all 24 turns of that transcript on screen.

Measured in the `long-transcript` fixture window (Electron, `MAKA_E2E=1`, `MAKA_E2E_FIXTURE=long-transcript`), calling the renderer bridge directly so the modal's own filtering is out of the picture:

```ts
await window.maka.search.thread({ source: 'thread', query, limit: 10 })
```

| query | result |
|---|---|
| `长会话` | 1 hit — `{"title":"超长会话滚动几何","summary":"会话标题","target":{"kind":"thread","sessionId":"e2e-fixture-long-transcript"}}` |
| `长会话问题` | `[]` |
| `长会话问题 3` | `[]` |
| `长会话回答 3` | `[]` |
| `占位正文` | `[]` |

Every one of those strings is in the seeded messages (`longTranscriptMessages` in `apps/desktop/src/main/e2e-fixture/scenarios-sessions.ts`): each turn is a `长会话问题 N` user message and a `长会话回答 N` assistant message with 60 filler lines of `… 用于撑高单个 turn 的占位正文内容。`.

## Where the two paths diverge

The renderer and search do not read messages the same way in fixture mode.

```ts
// apps/desktop/src/main/sessions-ipc-main.ts — the renderer's transcript load
ipcMain.handle('sessions:readMessages', async (_event, sessionId: string) => {
  if (e2eFixture) return store.readMessages(sessionId);   // ← fixture bypass
  messages = await runtime.getMessages(sessionId);
  …
});

// same file — search's dependency, no such bypass
return runThreadSearch(request, {
  listSessions: () => runtime.listSessions(),
  readMessages: (sessionId: string) => runtime.getMessages(sessionId),
  …
});
```

`resolveE2eFixture` returns non-null only when a scenario is set, so the bypass is active in exactly the windows that seed transcripts. The renderer therefore renders from the store the fixture wrote, while search asks the runtime — which is why the session is visible to search (it comes from `listSessions`, hence the title hit) but its messages are not.

`runThreadSearch` also swallows the difference between "threw" and "returned nothing":

```ts
try { messages = await deps.readMessages(session.id); } catch { continue; }
```

so from outside I can't tell which of the two happens for a seeded id — only that the search path never reaches the store the fixture wrote to. That distinction is what decides the fix (register seeded sessions with the runtime, or give search the same fixture-mode store read), so whoever takes this should establish it before changing anything.

**Not the cause:** the `backend !== 'fake'` filter. Fixture headers default to `backend: 'ai-sdk'` (`seed-helpers.ts`), and the title hit above proves the session reaches the scan.

## Why it matters

A **title-only hit carries no `turnId`**:

```ts
// thread-search.ts, title branch
target: { kind: 'thread', sessionId: session.id }
// content branch
target: { kind: 'thread', sessionId: session.id, ...(turnId ? { turnId } : {}) }
```

and the shell turns that into a plain session switch:

```ts
// app-shell.tsx
if (turnId) setSearchScrollTarget({ sessionId, turnId, nonce: Date.now() });
else setSearchScrollTarget(null);
```

So no e2e can drive search → navigate-to-a-specific-turn against a seeded fixture. That is what blocked the turn-navigation coverage discussed in #2239 (`use-chat-scroll`'s scroll-target effect, which releases the arrival pin when the reader picks a position); the gap is recorded there with this issue as its reason.

Two things that do **not** work around it: a message written by the runtime inside a fixture window might well be searchable — I did not establish that, my one attempt didn't produce a reply and I didn't chase why — but it would land at the tail of the transcript, and a target that coincides with the bottom cannot distinguish "navigated to the hit" from "followed to the latest turn", which is the whole point of such a test.

## Contrast

Content search works when the runtime itself created the session: `e2e/floating-layers.spec.ts` → `search closes before navigating and focusing the matched turn` sends a message through the composer and then finds it by content (re-ran it while writing this — green). Note that spec's window has **no** fixture scenario, so `e2eFixture` is null there and both paths read through the runtime — it is the contrast case for the divergence above, not the same window in a different state.

## Repro

```ts
test('probe', async ({ longTranscriptWindow: page }) => {
  for (const query of ['长会话', '长会话问题', '占位正文']) {
    console.log(query, await page.evaluate(async (q) =>
      JSON.stringify(await window.maka.search.thread({ source: 'thread', query: q, limit: 10 })), query));
  }
});
```

Measured on `2b9719eb7` (branch of #2239), which is `f8d2ac4ee` plus that PR's commits — none of which touch search, storage, or the fixture.

1625567290:
## Summary

Fixture windows seed transcripts into the workspace store. After the Runtime Host migration, `search:thread` still listed those sessions from the Host catalog (so a title query could hit) but read messages through `openSession` / `loadTranscript`. A Host miss is swallowed as `null`, so content search returned no hits and no `turnId` — which is what blocked e2e from driving search → navigate-to-a-specific-turn against a seeded transcript (#2305).

The original `sessions:readMessages` fixture bypass is gone. This restores the same contract on the current IPC: when `MAKA_E2E_FIXTURE` is set, search reads the store `writeSession` wrote. Production still reads the Host transcript.

`long-transcript` is gone; the lock uses the existing `chat-prompt-rail` seed (`第 3 个问题` → `turn-prompt-rail-3`).

Fixes #2305

## Verification

- `node --test apps/desktop/dist/main/__tests__/runtime-host-search-ipc-main.test.js apps/desktop/dist/main/__tests__/thread-search.test.js` — **15/15**
  - fixture store + failing Host still returns the content hit with `turnId`
  - without the fixture reader, Host transcripts still work
  - without the fixture reader, a Host miss stays title-only / empty
- Not run: full desktop Playwright e2e (the new `fixture-thread-search.spec.ts` is the window-level lock; CI should run it)

## AI use

Select exactly one:

Tool(s) and scope:

Grok authored the implementation, tests, and this PR description.

## Checklist

- [ ] No


apps/desktop/src/main/runtime-host-search-ipc-main.ts
```diff
@@ -23,18 +31,27 @@ export function registerRuntimeHostSearchIpc(
...
-          const session = await deps.client.openSession(sessionId);
-          try {
-            return await session.loadTranscript();
-          } finally {
-            await session.close();
-          }
-        }, null),
+        deps.readFixtureMessages
+          ? deps.readFixtureMessages(sessionId)
+          : readHostTranscript(deps.client, sessionId),
```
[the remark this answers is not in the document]
1625567290: Done in bbcf749c6.

`readMessages` now uses the fixture store only when it returns a non-null array. A `null` miss (unknown session or read failure) falls through to `openSession` / `loadTranscript`, so a Host-written session in the same fixture window stays searchable.

Added a test where `readFixtureMessages` returns `null` and the Host transcript supplies the hit. Seeded sessions still never call Host (`hostTranscriptReads === 0`).

Generated-by: Grok

Astro-Han:
Thanks for digging into the fixture search gap — the direction is right and the test scaffolding is solid. The review surfaced two issues worth resolving before merge.

**Conclusion: needs work — two P1s.**

**P1-1 — the two new tests pin behavior that `main` has already changed.**
- `runtime-host-search-ipc-main.test.ts:75` asserts `summary: "会话标题"`, but `thread-search.ts:205` on current `main` emits `'任务标题'` (changed in #3033).
- `fixture-thread-search.spec.ts` asserts a bare `sessionId`, but `preload.ts` now maps `search.thread` results through `desktopSessionKey({ hostId, sessionId })` (changed in #3097).

The PR's base (`5c1115132`) predates both. After a rebase onto current `main` these assertions fail, so the PR would land red (verified against current `main` sources). Rebasing and updating the two assertions should resolve this.

**P1-2 — the premise of the fix may not hold: the Host already serves fixture-seeded transcripts.**
`writeSession` in `seed-helpers.ts` and the Host's `SqliteSessionStore` write the *same* SQLite store (`createSqliteSessionMetadataStore(join(workspaceRoot, OPERATIONAL_STATE_DATABASE_NAME))`). The renderer renders fixture transcripts through the same `client.openSession` path, and the prompt-rail e2e suite proves that works on `main` CI. `search:thread` reads messages through the same `openSession`/`loadTranscript` path, so there's no obvious reason it fails where rendering succeeds. Note that #2305 (the issue this PR fixes) analyzed the *pre-migration* code, where `runtime.getMessages` did not read the store at all.

Suggested verification: revert the 4 production files, keep only `fixture-thread-search.spec.ts`, and run it. If it passes (expected), the production change can be deleted entirely — the e2e spec alone is the lock. If it fails, the fix becomes a workaround rather than a root-cause fix, and the fallback in the second commit still looks like dead code either way (catalog sessions always exist in the same store, so `?? readHostTranscript` never fires).

Optional nits (P3): tests 3/4 in the new suite exercise unchanged paths; `if (!Array.isArray(outcome)) return;` in the spec is dead code (the `expect` above throws first).

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagents on `ollama-cloud/deepseek-v4-flash`, read-only). The subagents traced the Host transcript-read path and the store writes; I verified the P1-1 assertions and the shared-store claim against current `main` sources myself. The P1-2 "spec passes without the production change" outcome is a prediction, not something I ran — the verification steps are above. Please weigh these findings with your own judgment.*

<details>

结论：FAIL，需要返工。两个 P1：① 两个新测试钉死了 main 已改掉的行为——单测断言 `'会话标题'` 但 #3033 已改为 `'任务标题'`，e2e 断言裸 sessionId 但 #3097 已做 desktopSessionKey 映射；PR 的 base 早于这两个提交，rebase 后必红，需更新断言。② 修复前提存疑：fixture 的 writeSession 与 Host 的 SqliteSessionStore 写的是同一个 SQLite 文件，渲染器走同一 client.openSession 路径渲染 fixture transcript 在 main CI 通过，search 走同一路径没有理由单独失败；#2305 分析的是迁移前代码（getMessages 不读 store），根因在迁移后已不存在。建议：revert 4 个生产文件只留 e2e spec 验证；第二个 commit 的 fallback 无论如何都是死代码。P3 可选：测试 3/4 测的是未改动路径、spec 中死代码。
</details>

Astro-Han (commented):
The intended regression lock is useful, and the Host reader keeps correct session cleanup. Production still defaults to the Host path and the focused old-base checks are green.

Against current `main`, however, the new assertions are stale (`desktopSessionKey` projection and `任务标题`), so the merged test result fails even though GitHub can merge the text cleanly. More importantly, the fixture and Runtime Host use the same `runtime.sqlite`; the Host transcript reader already reads that store, and a current-main 120-turn fixture is fully visible through `loadTranscript()`. From first principles, Runtime Host must remain the sole transcript authority. The smallest solution is to keep the E2E assertion, delete the new Desktop-main SQLite reader/wiring, and fix any actual Host readiness issue only if the E2E still fails.

Reviewed with Codex using two independent reviewer agents; I verified the latest head, current-main contracts, shared store/Host reader path, active-overlay behavior, existing discussion, and live CI.

这条回归锁本身有价值，Host reader 的 session cleanup 也正确；生产默认仍走 Host，旧 base 上的聚焦检查全绿。

但相对当前 main，新增断言已经过期（`desktopSessionKey` 投影和 `任务标题`），所以即使 GitHub 文本上可干净合并，合并后的测试仍会失败。更重要的是，fixture 与 Runtime Host 使用同一个 `runtime.sqlite`，Host transcript reader 本就读取该 store；在 current main 上，120-turn fixture 可完整通过 `loadTranscript()` 读取。按第一性原理，Runtime Host 必须保持唯一 transcript authority。最小方案是保留 E2E 断言，删除新增的 Desktop-main SQLite reader/wiring；只有 E2E 仍失败时再修真实的 Host readiness 根因。

本次由 Codex 配合两个独立 reviewer agent 审查；我核验了最新 head、current-main 契约、共享 store/Host reader 路径、active overlay 行为、已有讨论和实时 CI。

</details>

apps/desktop/e2e/fixture-thread-search.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (!Array.isArray(outcome)) return;
+
+  const content = outcome.find(
+    (hit) =>
+      hit.target?.kind === "thread" &&
+      hit.target.turnId === "turn-prompt-rail-3",
+  );
+  expect(content?.summary).toBe("用户消息");
+  expect(content?.target).toEqual({
+    kind: "thread",
```
Astro-Han: **P1 — Update this assertion to the current Desktop identity contract.** Current `main` projects search results through `desktopSessionKey(hostId, sessionId)`, so the renderer receives the JSON-scoped key rather than bare `PROMPT_RAIL_SESSION_ID`; the unit test also still expects `会话标题` while current search copy is `任务标题`. Rebase/update both assertions and rerun the focused E2E/unit tests.

apps/desktop/src/main/runtime-host-search-ipc-main.ts
```diff
@@ -22,19 +32,27 @@ export function registerRuntimeHostSearchIpc(
...
-        readWithFallback(async () => {
-          const session = await deps.client.openSession(sessionId);
-          try {
-            return await session.loadTranscript();
-          } finally {
-            await session.close();
-          }
-        }, null),
+      readMessages: async (sessionId) =>
+        (await deps.readFixtureMessages?.(sessionId)) ??
```
Astro-Han: **P2 — Do not add a second transcript reader beside Runtime Host.** Fixture seeding and `SqliteSessionStore` use the same operational DB, and current Host `loadTranscript()` reads the seeded 120-turn fixture completely. This bypass is therefore unnecessary production wiring; it also treats `[]` as authoritative and skips Host active-overlay messages. Keep the E2E regression, remove `readFixtureMessages` and this fallback, and repair Host readiness only if the direct Host path demonstrably fails.

1625567290: Addressed in the current head (`10d0597c4`). The separate fixture transcript reader and fallback wiring have been removed. Thread search now uses `openSession().loadTranscript()` as the sole transcript source, closes the session in `finally`, and retains the fixture-seeded E2E regression. The focused Runtime Host search tests passed 2/2.

apps/desktop/src/main/runtime-host-search-ipc-main.ts
```diff
@@ -22,19 +32,27 @@ export function registerRuntimeHostSearchIpc(
...
+        readHostTranscript(deps.client, sessionId),
```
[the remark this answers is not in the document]
1625567290: Addressed in the current head (`10d0597c4`) by removing the optional fixture reader entirely. Runtime Host `loadTranscript()` is the only transcript source and is wrapped with the existing `readWithFallback` seam, so Host read failures yield no content result without bypassing active-overlay messages. The focused tests passed 2/2.

apps/desktop/e2e/fixture-thread-search.spec.ts
1625567290: Addressed in the current head (`10d0597c4`). The E2E now parses the scoped result identity and compares it with `desktopSessionKey({ hostId, sessionId })`; the focused unit test expects the current `任务标题` copy. The Runtime Host search tests passed 2/2.

Astro-Han (approved):
Thanks @1625567290 — re-reviewed at exact head `10d0597c441cb357eb59d53334b8f432850fe40f`. The earlier round was against `bbcf749c`; both points raised there are addressed at this head. No findings; approving.

**Previously raised, now resolved**

- The E2E assertion no longer uses a bare `sessionId` or the old `会话标题` wording; it goes through `desktopSessionKey({hostId, sessionId})` and `任务标题`, matching the current identity contract.
- The second fixture-reading branch is gone. `readMessages` now has a single path — `openSession()` → `loadTranscript()` through `readWithFallback`, with `close` in a `finally` — and a failed read returns null rather than falling through to a direct fixture read. Nothing bypasses the active-overlay path.

**On the risk specific to a tests-only change**

A PR that only touches test files can weaken coverage while appearing to strengthen it, so that was checked directly rather than assumed. This change is purely additive — no existing assertion was modified or removed — and the new assertions are strict rather than permissive: the E2E case uses `toBe` on the summary and `toEqual` on the full target object, and the unit cases use `assert.deepEqual` on the whole target plus a `closed === 2` count that pins resource cleanup.

The other failure mode for a fixture-backed search test is that it asserts a string the fixture itself hard-codes, which would prove only that the seed round-trips. That is not the case here: the seeded turns are generated as real conversation text by `scenarios-chat.ts`, the asserted substring is part of that generated user message, and the search runs through the real IPC path into Host `listSessions` / `openSession` / `loadTranscript`. The seed builds no index and installs no bridge shortcut, so a hit can only be produced by the actual search chain.

The two files do not overlap: the E2E case drives a real fixture-seeded window, the unit case drives the IPC handler against a mock client, including the Host-failure-yields-no-hits path.

---

PR #3092, opened 2026-08-15
refactor(cli): unify MakaRunOutcome classification

### LINKED ISSUE #3088: refactor(cli): remove dead outcome helpers and unify MakaRunOutcome classification
## Problem

Three dead / duplicated outcome paths in `packages/cli`:

1. `sandbox-boundary-failure.ts` — `invocationHasSandboxBoundaryFailure` / `invocationRecoveredSandboxBoundaryFailure` have zero callers (they are the last consumers of the removed runtime's `InvocationResult`). The live sibling `sessionEventSandboxBoundaryFailureReason` must stay.
2. `MakaRunOutcome` is computed twice from two sources in `runtime-host-run-command.ts` — a live `TurnOutcomeAccumulator` (SessionEvent stream) and a stored-message classifier (`outcomeFromStoredTurn` / `storedSandboxBoundaryOutcome`) whose vocabularies already diverge (`aborted`/`runtime_error` vs `errorClass ?? abortSource ?? status`).
3. Dead exports with no callers: `formatResumeHint`, `cleanRecapText`/`RECAP_INSTRUCTION`, `runMakaTextCli`, `colorDiffRow`.

## Desired outcome

One `MakaRunOutcome` classifier parameterized over its source, dead helpers removed, dead exports dropped.

Leave the divergence; the stored-message path already produces different failure classes from the live path.

## Acceptance

- [ ] `invocationHasSandboxBoundaryFailure` / `invocationRecoveredSandboxBoundaryFailure` removed; `sessionEventSandboxBoundaryFailureReason` retained
- [ ] One outcome classifier shared by the live and stored-message paths
- [ ] Dead exports removed
- [ ] `maka run` and `maka run --graph` exit codes pinned by CLI tests; typecheck and Biome pass

## Related

- #3034, #3035 (prior runtime path removals)

Astro-Han:
## Summary

- Normalize live `SessionEvent` and durable `StoredMessage` inputs through one internal outcome reducer while preserving their distinct incomplete-terminal policies.
- Align cancellation, failure, complete-stop, and sandbox recovery semantics across `maka run` and `maka run --graph`.
- Remove InvocationResult-era sandbox helpers and unused CLI wrappers and exports.
- Pin ordinary and Graph success/failure exit codes at the public CLI boundary.

Fixes #3088

## Verification

- `npm --workspace maka-agent run build`
- `node --test packages/cli/dist/__tests__/runtime-host-run-command.test.js` — 25 passed
- `npm run format:check`
- `npx biome check` on all changed files
- Full-repository tests were not run; the focused CLI suite covers the changed execution seam.

## AI use

Select exactly one:

Tool(s) and scope: Codex analyzed the duplicate outcome paths, implemented the reducer and dead-code cleanup, authored and ran the focused tests, and drafted this PR description. The commits include `Generated-by: Codex` trailers.

## Checklist

- [ ] No

### DIFF
--- a/packages/cli/src/__tests__/runtime-host-run-command.test.ts
+++ b/packages/cli/src/__tests__/runtime-host-run-command.test.ts
@@ -233,13 +233,20 @@ describe('Runtime Host maka run adapter', () => {
   });
 
   test('returns exit code 1 when the final Graph Turn fails', async () => {
+    const stderr: string[] = [];
     const fixture = runFixture({
       graph: true,
       finalMessages: failedGraphMessages('provider_failure'),
     });
-    const exitCode = await runFixtureCommand(fixture, ['delegate once', '--graph']);
+    const exitCode = await runFixtureCommand(
+      fixture,
+      ['delegate once', '--graph'],
+      undefined,
+      (text) => stderr.push(text),
+    );
 
     assert.equal(exitCode, 1);
+    assert.equal(stderr.join(''), 'maka run: Agent Graph final Turn failed\n');
   });
 
   test('waits for Host-started graph supervisor Turns before returning', async () => {
@@ -898,10 +905,11 @@ function runFixtureCommand(
   fixture: ReturnType<typeof runFixture>,
   argv: readonly string[],
   writeStdout: (text: string) => void = () => {},
+  writeStderr: (text: string) => void = () => {},
 ): Promise<number> {
   return runRuntimeHostTextCli(
     argv,
-    { ...publicCommandEnvironment(), writeStdout },
+    { ...publicCommandEnvironment(), writeStdout, writeStderr },
     {
       connect: async () => ({
         connection: readinessConnection(),
--- a/packages/cli/src/__tests__/runtime-host-run-command.test.ts
+++ b/packages/cli/src/__tests__/runtime-host-run-command.test.ts
@@ -218,6 +218,20 @@ describe('Runtime Host maka run adapter', () => {
     assert.equal(stdout.join(''), 'Final graph answer\n');
   });
 
+  test('returns exit code 0 when a root Graph boundary failure recovers', async () => {
+    const stdout: string[] = [];
+    const fixture = runFixture({
+      graph: true,
+      turnEvents: recoveredSandboxEvents('turn-1'),
+    });
+    const exitCode = await runFixtureCommand(fixture, ['recover once', '--graph'], (text) =>
+      stdout.push(text),
+    );
+
+    assert.equal(exitCode, 0);
+    assert.equal(stdout.join(''), 'Final graph answer\n');
+  });
+
   test('returns exit code 1 when the final Graph Turn fails', async () => {
     const fixture = runFixture({
       graph: true,
--- a/packages/cli/src/__tests__/runtime-host-run-command.test.ts
+++ b/packages/cli/src/__tests__/runtime-host-run-command.test.ts
@@ -292,12 +292,16 @@ describe('Runtime Host maka run adapter', () => {
       finalMessages: abortedGraphMessages(),
     });
 
+    assert.equal(live.status, 'failed');
     assert.equal(live.failure?.class, 'aborted');
+    assert.equal(durable.status, 'failed');
     assert.equal(durable.failure?.class, 'aborted');
   });
 
-  test('classifies a live step limit like a durable failed Turn', async () => {
-    const live = await observeFixtureOutcome({ turnEvents: stepLimitedEvents('turn-1') });
+  test('classifies live and durable step-cap failures equally', async () => {
+    const live = await observeFixtureOutcome({
+      turnEvents: failedEvents('turn-1', 'tool_step_cap_reached'),
+    });
     const durable = await observeFixtureOutcome({
       graph: true,
       finalMessages: failedGraphMessages('tool_step_cap_reached'),
@@ -309,20 +313,14 @@ describe('Runtime Host maka run adapter', () => {
     assert.equal(durable.failure?.class, 'tool_step_cap_reached');
   });
 
-  test('classifies a user-stop complete event as aborted', async () => {
-    const outcome = await observeFixtureOutcome({ turnEvents: userStoppedEvents('turn-1') });
-
-    assert.equal(outcome.status, 'failed');
-    assert.equal(outcome.failure?.class, 'aborted');
-  });
-
-  test('keeps a live Turn failed when complete follows its error', async () => {
+  test('uses the latest durable terminal state for a Graph Turn', async () => {
     const outcome = await observeFixtureOutcome({
-      turnEvents: failedThenCompleteEvents('turn-1'),
+      graph: true,
+      finalMessages: failedThenCompletedGraphMessages(),
     });
 
-    assert.equal(outcome.status, 'failed');
-    assert.equal(outcome.failure?.class, 'provider_failure');
+    assert.equal(outcome.status, 'completed');
+    assert.equal(outcome.finalOutput, 'Final graph answer');
   });
 
   test('waits for the exact final Graph wake after an earlier wake already settled', async () => {
@@ -908,7 +906,6 @@ function publicCommandEnvironment() {
     processCwd: () => process.cwd(),
     stdinIsTTY: () => true,
     readStdin: async () => '',
-    writeStdout: () => {},
     writeStderr: () => {},
     onSigint: () => () => {},
     newId: () => 'turn-1',
@@ -1161,6 +1158,20 @@ function failedGraphMessages(errorClass: string): StoredMessage[] {
   ];
 }
 
+function failedThenCompletedGraphMessages(): StoredMessage[] {
+  return [
+    ...failedGraphMessages('provider_failure'),
+    {
+      type: 'turn_state',
+      id: 'completed-state-turn-2',
+      turnId: 'turn-2',
+      ts: 6,
+      status: 'completed',
+      partialOutputRetained: true,
+    },
+  ];
+}
+
 function multiWakeGraphMessages(includeFinalTerminal: boolean): StoredMessage[] {
   const messages = [
     ...graphMessages(),
@@ -1227,7 +1238,7 @@ async function* abortedEvents(turnId: string): AsyncIterable<SessionEvent> {
   };
 }
 
-async function* stepLimitedEvents(turnId: string): AsyncIterable<SessionEvent> {
+async function* failedEvents(turnId: string, reason: string): AsyncIterable<SessionEvent> {
   yield {
     type: 'text_complete',
     id: `${turnId}-text`,
@@ -1237,44 +1248,13 @@ async function* stepLimitedEvents(turnId: string): AsyncIterable<SessionEvent> {
     text: 'Partial answer',
   };
   yield {
-    type: 'complete',
-    id: `${turnId}-complete`,
-    turnId,
-    ts: 2,
-    stopReason: 'step_limit',
-  };
-}
-
-async function* userStoppedEvents(turnId: string): AsyncIterable<SessionEvent> {
-  yield {
-    type: 'complete',
-    id: `${turnId}-complete`,
-    turnId,
-    ts: 1,
-    stopReason: 'user_stop',
-  };
-}
-
-async function* failedEvents(turnId: string, reason: string): AsyncIterable<SessionEvent> {
-  yield {
     type: 'error',
     id: `${turnId}-error`,
     turnId,
-    ts: 1,
+    ts: 2,
     recoverable: false,
     reason,
     message: 'Turn failed',
-  };
-}
-
-async function* failedThenCompleteEvents(turnId: string): AsyncIterable<SessionEvent> {
-  yield* failedEvents(turnId, 'provider_failure');
-  yield {
-    type: 'complete',
-    id: `${turnId}-complete`,
-    turnId,
-    ts: 2,
-    stopReason: 'end_turn',
   };
 }
 
--- a/packages/cli/src/__tests__/runtime-host-run-command.test.ts
+++ b/packages/cli/src/__tests__/runtime-host-run-command.test.ts
@@ -200,6 +200,34 @@ describe('Runtime Host maka run adapter', () => {
     ]);
   });
 
+  test('returns exit code 1 when an ordinary Host Turn fails', async () => {
+    const fixture = runFixture({ turnEvents: failedEvents('turn-1', 'provider_failure') });
+    const exitCode = await runFixtureCommand(fixture, ['fail once']);
+
+    assert.equal(exitCode, 1);
+  });
+
+  test('returns exit code 0 when the final Graph Turn completes', async () => {
+    const stdout: string[] = [];
+    const fixture = runFixture({ graph: true });
+    const exitCode = await runFixtureCommand(fixture, ['delegate once', '--graph'], (text) =>
+      stdout.push(text),
+    );
+
+    assert.equal(exitCode, 0);
+    assert.equal(stdout.join(''), 'Final graph answer\n');
+  });
+
+  test('returns exit code 1 when the final Graph Turn fails', async () => {
+    const fixture = runFixture({
+      graph: true,
+      finalMessages: failedGraphMessages('provider_failure'),
+    });
+    const exitCode = await runFixtureCommand(fixture, ['delegate once', '--graph']);
+
+    assert.equal(exitCode, 1);
+  });
+
   test('waits for Host-started graph supervisor Turns before returning', async () => {
     const observed: MakaRunOutcome[] = [];
     const fixture = runFixture({ observed, graph: true, graphProjectionRace: true });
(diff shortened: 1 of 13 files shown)


Astro-Han:
Thanks for unifying the outcome classification — the reducer shape is minimal and the deletion is thorough. I ran the focused suite and probed the live/durable paths independently.

**Conclusion: PASS. One P2 with an explicit deferral rationale; optional P3 nits.**

Verified:
- All removed symbols (`formatResumeHint`, `runMakaTextCli`, the sandbox-boundary helpers, recap helpers) have no callers.
- The 7 new tests are behavior-level (exit code, status/class, sandboxBoundary), and the four scenarios — cancellation, failure, complete-stop, sandbox recovery — are covered on both live and durable paths.
- Focused suite passes (24/24) and a separate probe confirms the live/durable alignment claim.

Findings:
- **P2**: live-path fail-closed semantics changed from order-independent ("failure wins") to "last terminal event wins" — `runtime-host-run-command.ts:540` now assigns `this.#terminal = observation` unconditionally. The commit that removed the guarding test explicitly defers this on the projector invariant "exactly one terminal event per turn", which I verified currently holds in `session-projector.ts` (all three emission paths send at most one terminal event). No behavior regression today, but the CLI side no longer pins the invariant — a future projector change could silently flip a failed turn to exit 0. Suggest a one-line comment at the terminal case (or a defensive test) to pin the single-terminal-event precondition.
- **P3**: graph-failure stderr lost the `Agent Graph final Turn failed` message; with no `errorClass` it degrades to `maka run: runtime_error` (exit code unchanged). No test asserts stderr.
- **P3**: the new contract "any `complete` event → completed" (ignoring `stopReason`) has no test pinning it.
- **P3**: `event.reason ?? event.code ?? 'runtime_error'` — the `event.code` fallback is dead (the projector always sets `reason`); trivially removable.

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagent on `ollama-cloud/deepseek-v4-flash`, read-only). The subagent ran the focused test suite, a live/durable probe, and git-grep verification of removed symbols; I verified the terminal-event assignment at `runtime-host-run-command.ts:540` myself. Failure-mode predictions are analysis, not observed behavior. Please weigh these findings with your own judgment.*

<details>

结论：PASS，无 P0/P1，可合并。删除彻底（所有被删符号无调用方），7 个新测试均为行为级，cancellation/failure/complete-stop/sandbox recovery 四场景在 live/durable 两条路径都有覆盖，聚焦套件 24/24 通过。1 个 P2：live 路径的 fail-closed 从"顺序无关（failure 胜出）"退化为"最后终端事件胜出"（runtime-host-run-command.ts:540），当前依赖 projector"每 turn 恰好一个终端事件"的不变量成立、无行为回归，但 CLI 侧不再钉住该不变量，建议在 terminal case 加一行注释或补防御性测试。P3 可选：graph 失败 stderr 丢失描述性消息、complete 忽略 stopReason 无测试、`event.code` fallback 是死代码。
</details>

packages/cli/src/__tests__/runtime-host-run-command.test.ts
```diff
@@ -248,6 +276,55 @@ describe('Runtime Host maka run adapter', () => {
...
 
+  test('reports a recovered sandbox boundary from live and durable Turns', async () => {
+    const live = await observeFixtureOutcome({ turnEvents: recoveredSandboxEvents('turn-1') });
+    const durable = await observeFixtureOutcome({
+      graph: true,
+      finalMessages: recoveredSandboxMessages(),
+    });
+
+    assert.equal(live.sandboxBoundary, 'recovered');
+    assert.equal(durable.sandboxBoundary, 'recovered');
```
MicroGery, line 432: [P1] Exercise recovered boundaries through the public Graph command

This test checks the live and durable classifier results only in isolation, so it misses the multi-outcome aggregation performed by the public CLI. When the root Turn emits sandbox failure, then a successful tool result, and the final Graph Turn completes with sandboxBoundary none, runOutcomeObserver overwrites outcome with that final result while streamBoundaryFailure remains true. The check in run-command-core then returns 1 even though the only boundary failure recovered. I reproduced this on the current head by running the scenario through runRuntimeHostTextCli: expected 0, actual 1.

Please add a public maka run --graph regression for this sequence and aggregate unresolved boundary state across outcomes instead of interpreting the global stream flag through only the last outcome.

packages/cli/src/runtime-host-run-command.ts
```diff
@@ -562,6 +578,63 @@ class TurnOutcomeAccumulator {
...
+      kind: 'terminal',
+      status: 'failed',
+      failure: { class: 'aborted', message: 'Turn was cancelled' },
+    };
+  }
+  if (message.type === 'turn_state' && message.status === 'failed') {
+    return {
+      kind: 'terminal',
+      status: 'failed',
+      failure: { class: message.errorClass ?? 'runtime_error' },
```
MicroGery: [P2] Preserve a human-readable durable failure message

The old stored Graph outcome included message: Agent Graph final Turn failed, but this normalization now keeps only the machine class. runMakaTextCliCore prefers failure.message and otherwise prints failure.class, so a failed Graph Turn changes stderr from a human-readable explanation to raw text such as maka run: provider_failure. I reproduced the durable outcome on this head and failure.message is undefined.

Please retain a stable human-readable message here, ideally matching the Runtime Host projector form, and extend the Graph failure CLI test to assert stderr as well as exit code.

packages/cli/src/__tests__/runtime-host-run-command.test.ts
Astro-Han: Addressed in `270a56678`. Raw stream boundary failures now remain flagged only until the corresponding outcome classifies them as recovered or unresolved; unresolved state continues to be aggregated by outcome ID. Added a public `maka run --graph` regression covering a recovered root boundary followed by a successful final Graph Turn, asserting exit code 0 and final stdout. The focused CLI suite passes.

Automated response by Codex.

packages/cli/src/runtime-host-run-command.ts
Astro-Han: Addressed in `0f410bf2a`. Durable failed Graph Turns now retain the stable `Agent Graph final Turn failed` message. The public Graph failure test now asserts both exit code 1 and the human-readable stderr output. The focused CLI suite passes.

Automated response by Codex.

packages/cli/src/runtime-host-run-command.ts
```diff
@@ -505,53 +505,69 @@ function runtimeHostSessionSummaries(items: readonly SessionCatalogItem[]): Sess
...
+      case 'output':
+        this.#finalOutput = observation.text;
+        return;
+      case 'terminal':
+        this.#terminal = observation;
+        return;
+      case 'sandbox_failure':
+        this.#unresolvedBoundary = true;
+        return;
+      case 'tool_success':
```
hqhq1025: [P1] Keep a parallel sibling result from clearing this sandbox failure. The production tool loop executes every call from one provider step concurrently (`ai-sdk-backend.ts` uses `Promise.allSettled`), and each `tool_result` is published when that call settles. If call A returns `sandbox_boundary_required` and an unrelated sibling B succeeds afterward, this branch clears `#unresolvedBoundary` and marks the turn recovered. `run-command-core.ts` then clears the raw boundary flag and can return exit code 0 after completion even though A was denied and never retried. Multi-tool provider steps are a normal path. Track the provider step by mapping live `tool_start.toolUseId` / stored `tool_call.stepId`, and only let a success from a later step recover the failure; a same-step sibling success must leave it unresolved. Add ordinary and `--graph` public-CLI regressions for one failed and one successful call in the same step.

hqhq1025 (changes requested):
One material finding is inline.

packages/cli/src/runtime-host-run-command.ts
```diff
@@ -505,62 +505,180 @@ function runtimeHostSessionSummaries(items: readonly SessionCatalogItem[]): Sess
...
+  }
+  if (event.type === 'abort') {
+    return {
+      kind: 'terminal',
+      status: 'failed',
+      failure: { class: 'aborted', message: 'Turn was cancelled' },
+    };
+  }
+  if (event.type === 'complete') {
+    return { kind: 'terminal', status: 'completed' };
```
hqhq1025: [P1] Honor failure-bearing `CompleteEvent.stopReason` values. The production backend emits `error` followed by `complete(error)`, and cancellation emits `abort` followed by `complete(user_stop)`; `step_limit` and `context_budget_exhausted` can also arrive as standalone completion events. This branch rewrites all of them as successful and can overwrite the failure already stored by the classifier. If partial text was retained, `runMakaTextCliCore` then sees `completed` plus `finalOutput` and returns exit code 0. Map each stop reason through the same terminal semantics used by the durable turn projection, without letting the trailing completion erase an earlier failure, and cover the real terminal event sequences.

packages/cli/src/runtime-host-run-command.ts
```diff
@@ -505,62 +505,180 @@ function runtimeHostSessionSummaries(items: readonly SessionCatalogItem[]): Sess
...
+      stepId: message.stepId,
+    };
+  }
+  return message.type === 'tool_result' ? observationFromToolResult(message) : undefined;
+}
+
+function observationFromToolResult(
+  result: Pick<Extract<SessionEvent, { type: 'tool_result' }>, 'content' | 'isError' | 'toolUseId'>,
+): TurnOutcomeObservation | undefined {
+  if (result.isError && result.content.kind === 'text' && result.content.sandboxFailure) {
```
hqhq1025, line 746: [P1] Do not depend on full tool-result content on the Runtime Host stream. The normal host subscription intentionally reduces a terminal `tool_result` to status and identity, and `session-projector.ts` reconstructs it with `content: { kind: 'text', text: '' }`. `#observeTurn` also consumes `sandbox_boundary_request` internally, so the outer CLI never sets its fallback flag. A real `sandbox_boundary_required` or `requires_bypass` result therefore reaches this condition without `sandboxFailure`; a later model completion can be classified as `none` and make `maka run` return 0. Preserve the minimal sandbox-failure reason through the host projection (or classify it from durable data), and add a public regression built from the actual Runtime Host projection rather than a full local `ToolResultEvent`.

hqhq1025 (changes requested):
Two material findings are inline.

packages/runtime-host/src/adapter/session-projector.ts
```diff
@@ -506,7 +506,13 @@ function projectToolEvent(
     type: 'tool_result',
     ...base,
     isError: event.status === 'errored',
-    content: { kind: 'text', text: '' },
+    content: {
+      kind: 'text',
+      text: '',
+      ...(event.sandboxFailureReason
```
hqhq1025, line 512: [P1] Keep a denied boundary request from clearing this failure. On the normal non-interactive path, this projection marks the original `sandbox_boundary_required` result unresolved; the model then calls `request_sandbox_boundary`, and `NonInteractiveInteractionController` denies it. `ToolRuntime` returns that `{ request: { status: 'denied' }, changed: false }` settlement as a successful JSON tool result, so its different provider step makes `TurnOutcomeClassifier` mark the boundary recovered. `runOutcomeObserver` then clears the raw failure flag, and a final model message can make `maka run` exit 0 without widening the sandbox or retrying the blocked operation. Track the tool name with the step and exclude `request_sandbox_boundary` itself from recovery (for both live and durable classification), then add ordinary and Graph regressions for failure -> denied request -> final output.

hqhq1025 (changes requested):
One material finding is inline.

hqhq1025 (dismissed):
Re-reviewed the current head against the previous denied-boundary-request finding. Live and durable classification now retain each tool name with its provider step, so a successful `request_sandbox_boundary` settlement cannot clear the unresolved failure; the ordinary and Graph public CLI regressions both return failure as required. The complete build, related CLI/Runtime Host/ToolRuntime suites, formatting, lint, and required CI passed. No actionable findings remain.

packages/runtime-host/src/adapter/session-projector.ts
Astro-Han: Addressed in `fa3028627` after rebasing onto the current `main`. The unified classifier now retains each tool name with its provider-step identity and excludes successful `request_sandbox_boundary` settlements from sandbox recovery in both live and durable paths. Public ordinary and Graph CLI regressions cover `sandbox failure → denied boundary request → final output`; both retain the failure and return exit code 1. Focused local validation and the required CI check pass.

Automated response by Codex.

Astro-Han:
@hqhq1025 The branch has been rebased onto the current `main`; the new head is `fa3028627`. The denied-boundary-request fix and its ordinary/Graph regressions are preserved, the Runtime Host compatibility epoch now follows `main` at 32, and the required `test` check passes. Could you please re-review the current head?

Automated request by Codex on behalf of @Astro-Han.

packages/cli/src/runtime-host-run-command.ts
```diff
@@ -408,7 +408,7 @@ class RuntimeHostRunRuntime implements MakaRunRuntime {
   }
 
   async *#observeTurn(turn: MakaPreparedSessionTurn): AsyncIterable<SessionEvent> {
-    const accumulator = new TurnOutcomeAccumulator(turn.runId ?? turn.turnId);
+    const classifier = new TurnOutcomeClassifier(turn.runId ?? turn.turnId);
```
hqhq1025: [P2] Rebuild the active classifier after a reconnect sequence cut

This classifier consumes only `turn.events`, but reconnect recovery cuts every queued event in `RuntimeHostSessionChannel.#acceptCanonicalReplacement()` and reloads the canonical durable transcript. The transcript-replacement subscription currently feeds only `#acceptGraphTranscript`; it never replays stored tool facts into this active root classifier. If a sandbox failure was persisted but its live frame was lost before the cut, the replacement can deliver only the terminal completion and this classifier reports success. Please rebuild the active Turn outcome from the replacement transcript, or replay the relevant tool call/result facts after the sequence cut, with a public ordinary-command reconnect regression.

packages/cli/src/runtime-host-run-command.ts
```diff
@@ -505,60 +505,219 @@ function runtimeHostSessionSummaries(items: readonly SessionCatalogItem[]): Sess
...
+      case 'tool_call':
+        this.#callByToolUseId.set(observation.toolUseId, {
+          stepId: observation.stepId,
+          toolName: observation.toolName,
+        });
+        return;
+      case 'tool_result': {
+        const call = this.#callByToolUseId.get(observation.toolUseId);
+        if (observation.outcome === 'sandbox_failure') {
+          this.#sandboxBoundary = { status: 'unresolved', failedStepId: call?.stepId };
```
hqhq1025: [P2] Keep multiple unresolved sandbox failures distinct

Every sandbox failure overwrites the same single `#sandboxBoundary` state. With two blocked sibling calls in one provider step, a later step that successfully retries only one of them changes the whole Turn to `recovered`; the other blocked operation is no longer represented and the CLI can exit 0. Track unresolved failures individually and clear only a proven recovery, or fail closed when more than one unresolved operation cannot be correlated. Add ordinary and Graph regressions for two failures followed by one successful retry.

hqhq1025 (changes requested):
Requesting changes on head `fa30286270fd07dbb51538f0540a9c04092cea8d`.

Problem reconstruction

This PR correctly identifies the duplicated live/durable outcome classifiers as the root maintenance problem. Normalizing both sources into one reducer, preserving their different incomplete-terminal policies, deleting InvocationResult-era helpers, and pinning public CLI exit codes is the right direction. The latest rebase also correctly carries the bounded sandbox failure reason over the Runtime Host wire and allocates compatibility epoch 32.

Findings

1. P2: reconnect recovery can erase an unresolved sandbox failure from the active root Turn. The Session channel intentionally cuts queued events and rebuilds from the canonical transcript after a reconnect, but the transcript-replacement listener feeds only Graph terminal handling. The root classifier created in `#observeTurn()` consumes only live events. If the sandbox failure is durable but its live frame is lost before the sequence cut, the replacement can later seed only the terminal completion; the classifier returns completed and `maka run` can exit 0.

2. P2: the reducer collapses multiple sandbox failures into one state. Two blocked calls in one provider step both set the same `{ status: 'unresolved', failedStepId }`. A later provider step that successfully retries only one call flips the whole Turn to `recovered`, even though the sibling operation never ran. The public command can consequently exit 0 with unresolved blocked work.

First-principles and Occam assessment

The shared reducer is simpler than the previous duplicated authorities, but its sandbox state is not yet rich enough for the behavior it owns. The final structure should preserve unresolved failures individually, or consume an explicit Runtime recovery fact, and it must be rebuildable from the canonical replacement transcript after a sequence cut. A single mutable boundary flag cannot represent multiple failures or reconnect recovery correctly.

Deletion and tests

The dead exports and InvocationResult helpers are correctly removed; I found no additional production deletion required for merge. The new CLI tests are behavior-focused rather than coverage-only. Add public ordinary-command reconnect coverage and a two-failures/one-retry case for both ordinary and Graph paths. Also strengthen the epoch regression: `RUNTIME_HOST_COMPATIBILITY_EPOCH > 29` would still pass at the previous epoch 31, so it does not prove this wire change owns epoch 32.

Merge readiness

Not ready to merge. GitHub CI is green. Locally, 110 focused CLI/Runtime Host tests passed, CLI and Runtime Host typechecks passed, formatting, Astryx inventory, and `git diff --check` passed. The affected packages built successfully after serializing the workspace build; an initial full `build:test` attempt was invalidated by local concurrent/dist-generation behavior and was not used as evidence.

Residual gaps after the two blockers are fixed: packaged-binary smoke was not run, and the recovery semantics still rely on inferred provider-step progression rather than an explicit operation-recovery identity.

Codex-assisted adversarial re-review; I verified the exact live head, current-main integration, prior denied-boundary fix, Runtime Host producer/wire/projector chain, reconnect sequence-cut path, and public CLI tests before publication.

packages/cli/src/runtime-host-run-command.ts
Astro-Han: Addressed in `e86ef4fe6`. The outcome projection now belongs to the active root Turn and is rebuilt from canonical transcript replacements before subsequent live events are classified. A public ordinary `maka run` regression covers a reconnect where the sandbox failure exists only in the durable replacement transcript; the command retains the unresolved failure and exits 1.

Automated response by Codex.

packages/cli/src/runtime-host-run-command.ts
Astro-Han: Addressed in `e86ef4fe6`. The shared classifier now tracks unresolved sandbox failures independently by `toolUseId`. Because the current wire has no retry identity, a later cross-step success recovers the boundary only when exactly one unresolved candidate exists; ambiguous multi-failure cases fail closed. Public ordinary and Graph regressions cover two failures followed by one successful retry and both exit 1.

Automated response by Codex.

Astro-Han:
@hqhq1025 The two requested changes are addressed in `e86ef4fe6` on the rebased head. The active root outcome is now replayed from canonical transcript replacements, and unresolved sandbox failures remain individually represented with ambiguous recovery failing closed. Public ordinary reconnect and ordinary/Graph multi-failure regressions pass, together with the focused CLI and Runtime Host protocol suites, affected typechecks, lint, formatting, and `git diff --check`. CI is running on the exact head. Please re-review when ready.

Automated message by Codex.

hqhq1025 (approved):
Re-reviewed the current head against both previous blockers. No actionable findings remain.

Problem and mechanism: the PR correctly removes the duplicated live/durable CLI outcome authorities and normalizes both representations through one reducer while preserving their different incomplete-terminal policies. The latest revision now rebuilds the active root classifier from canonical transcript replacements after a sequence cut, so reconnect recovery cannot erase a durable sandbox failure. It also tracks unresolved failures independently by tool identity and fails closed when the current wire cannot correlate a partial retry.

First principles and Occam: one reducer is the simpler final architecture. Canonical transcript replacement is the right reconnect authority, and a collection of unresolved failures accurately represents the state that a single mutable flag could not. The conservative ambiguous-recovery rule is appropriate until the wire carries explicit retry identity.

Optimality, deletion, and tests: the InvocationResult-era helpers and unused exports are correctly removed. I found no additional production code or low-value test that should be deleted, and no deeper refactor is required for this PR. The new ordinary and Graph regressions directly cover reconnect restoration and two-failures/one-retry behavior.

Merge verdict: ready to merge. Residual risk is intentionally conservative false failure when multiple blocked operations are all retried but cannot be correlated; that is safer than returning exit code 0 with unresolved work and can be improved later with explicit recovery identity.

Verification: reviewed head e86ef4fe6f157047ef010f26fd70a505c0cdbd2c; required CI is green; npm run build:test passed; 158 focused CLI, session-driver, Runtime Host protocol, and continuity tests passed; git diff --check passed.

PR #3095, opened 2026-08-15
refactor(runtime-host): trim dead export surface from barrels

### LINKED ISSUE #3087: refactor(runtime-host): trim dead export surface from server/client/protocol barrels
## Problem

The `@maka/runtime-host` package is `private`, but its barrels advertise a much larger surface than anyone consumes:

- `server/index.ts` re-exports 49 names; 42 have zero external consumers (the code is live via internal relative imports, only the export surface is dead).
- `client/index.ts` has ~26 dead exports; `protocol/index.ts` surfaces an internal `RuntimeHostProtocolError`; `adapter/index.ts` exports an in-file-only type.

knip does not cover `runtime-host`, so these dead exports have no gate. Each one is a public-API promise that must be maintained for free.

## Desired outcome

Trim the barrels to the consumed surface so the package's actual contract matches what clients import. Runtime behavior is unchanged.

Leave the dead surface; it is the direct product of the missing knip coverage.

## Acceptance

- [ ] Dead re-exports removed from `server/index.ts`, `client/index.ts`, `protocol/index.ts`, `adapter/index.ts`
- [ ] Internal imports already use direct relative paths; no import is broken
- [ ] `packages/runtime-host` build, focused tests, and Biome pass
- [ ] Optionally: extend `knip.json` to cover `packages/runtime-host` (and sibling packages) so dead exports cannot re-accumulate

## Related

- knip coverage gap (see #2441 for the scope-taxonomy effort)

CxHsin:
## Summary

Trim the consumed API surface of the `@maka/runtime-host` server, client, protocol, and adapter barrels. Package-internal consumers of removed exports now import directly from their leaf modules, so runtime behavior is unchanged.

Fixes #3087

## Verification

- `npm --workspace @maka/runtime-host run build` passed.
- `npm --workspace @maka/runtime-host run typecheck` passed.
- Focused runtime-host protocol tests passed: 127/127.
- `npm run build` passed.
- `npm run typecheck` passed.
- `npm run lint` passed.
- `npm run format:check` passed.
- `npx knip --workspace apps/desktop --workspace packages/ui` passed.
- The full runtime-host test command was started and passed its initial suites, then stopped after Windows shell/PTY cases produced known environment-level failures and a prolonged no-output hang. No UI changes are included.

## AI use

Tool(s) and scope: Codex performed the implementation, test updates for leaf imports, verification, and code review. The commit includes the required `Generated-by: Codex` trailer.

## Checklist

- [x] No

### DIFF
--- a/packages/runtime-host/src/__tests__/agent-graph-protocol.test.ts
+++ b/packages/runtime-host/src/__tests__/agent-graph-protocol.test.ts
@@ -1,3 +1,4 @@
+import { RuntimeHostProtocolError } from '../protocol/errors.js';
 import assert from 'node:assert/strict';
 import { describe, test } from 'node:test';
 import {
@@ -13,7 +14,6 @@ import {
   decodeHostFrame,
   type AgentGraphClientSnapshot,
   type AgentGraphOperatorInspection,
-  RuntimeHostProtocolError,
 } from '../protocol/index.js';
 
 const fingerprint = `sha256:${'a'.repeat(64)}` as const;
--- a/packages/runtime-host/src/__tests__/artifact-protocol.test.ts
+++ b/packages/runtime-host/src/__tests__/artifact-protocol.test.ts
@@ -1,3 +1,4 @@
+import { RuntimeHostProtocolError } from '../protocol/errors.js';
 import assert from 'node:assert/strict';
 import { describe, test } from 'node:test';
 import type { ArtifactRecord } from '@maka/core/artifacts';
@@ -15,7 +16,6 @@ import {
   encodeArtifactQueryResult,
   encodeProtocolMessage,
   RUNTIME_HOST_MAX_MESSAGE_BYTES,
-  RuntimeHostProtocolError,
 } from '../protocol/index.js';
 import { encodeArtifactProjection } from '../protocol/artifact.js';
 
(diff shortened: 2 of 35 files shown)


Astro-Han (approved):
Reviewed against #3087.

The deleted server / client / protocol / adapter names have no external importers in cli, desktop, eval, ui, or runtime. The kept server exports are used. Package-internal imports that went through those barrels now point at leaf modules. `RuntimeHostProtocolError` is no longer re-exported from the protocol barrel; its remaining importers use `protocol/errors.js`. `OAUTH_PRESENTATION_SERVICE_ID` / `VERSION` left the client barrel and still come from the protocol barrel, which is the right owner.

This does not stop the same dead surface growing back: `knip.json` still only covers `apps/desktop` and `packages/ui`. That was optional in the issue; it is the actual gate. Also revert the unrelated `assert.rejects` rewraps in `host-kernel.test.ts`.

Approve. The knip follow-up should be its own PR.

AI-assisted review: Grok 4.6 drafted the first pass. A second pass on opencode-go/deepseek-v4-flash:max enumerated the deleted names and grepped consuming workspaces; it found no leftover importer and no behavior change in `host-kernel.test.ts` beyond the formatter reflow. I grepped the deleted export names from the consuming workspaces and checked `knip.json`. Unverified by me: I did not rerun the runtime-host suite.

Astro-Han:
Approve stands. Cannot merge yet: the typecheck job fails on Biome format in `packages/runtime-host/src/__tests__/host-kernel.test.ts`.

That is the unrelated `assert.rejects` rewrap I flagged in the review. Revert those three call sites to the previous wrapping (or run `npx biome check --write` on that file and commit the formatter output). After CI is green I can merge.

Fast path would apply once checks pass: barrel-only export trim, no runtime behavior change, easy to revert.

PR #3099, opened 2026-08-16
fix(desktop): persist rail grouping across restarts

### LINKED ISSUE #3060: Rail grouping (按时间 / 按项目) does not persist across restarts
## Problem

`viewMode` (按时间 / 按项目) is `useState<SessionViewMode>('conversation')` in `apps/desktop/src/renderer/app-shell.tsx`. It resets to 按时间 on every restart.

This predates #3033, but that change promotes the control to permanent sticky chrome at the top of the rail — a full-width `SegmentedControl` that is always visible — so a preference silently reverting is now visible every launch rather than hidden in a dropdown.

The rail's collapsed state already persists through `session-list-layout.ts` (`maka-chat-list-collapsed-v1`); grouping should use the same seam.

## Context

Found by a subagent during the adversarial review of #3033.

1625567290:
## Summary

`viewMode` (按时间 / 按项目) lived in `useState<SessionViewMode>('conversation')`, so the rail grouping control reset to 按时间 on every launch. After #3033 the control is permanent sticky chrome, so the silent revert is visible every start.

Grouping now uses the same localStorage seam as the rail's collapsed flag (`session-list-layout.ts`):

- `readSessionListViewMode()` hydrates the shell
- `writeSessionListViewMode()` persists through `useAppShellPersistenceEffects`
- unknown / empty stored values fail open to `conversation`

Fixes #3060

## Verification

- `npx biome check` on the four changed files — clean
- `npx tsc -p tsconfig.renderer.json --noEmit` — clean
- `npm --workspace @maka/desktop test` — **860/860**, including the new persist/hydrate cases
- Red: the new test failed to import `readSessionListViewMode` / `writeSessionListViewMode` before the production helpers existed

Not run: desktop e2e / Electron window. The official desktop unit suite covers the persist/hydrate contract.

## AI use

Select exactly one:

Tool(s) and scope:

Grok authored the implementation, tests, and this PR description.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/e2e/sidebar-project-row.spec.ts
+++ b/apps/desktop/e2e/sidebar-project-row.spec.ts
@@ -62,3 +62,29 @@ test('project navigation and actions remain adjacent keyboard controls', async (
   await page.getByRole('button', { name: '关闭', exact: true }).click();
   await expect(action).toBeFocused();
 });
+
+test('rail grouping survives a renderer reload', async ({ projectSidebarWindow: page }) => {
+  await page.keyboard.press('Escape');
+  await expect(page.locator('[data-maka-contract="search-modal"]')).not.toBeVisible();
+
+  const sidebar = page.getByRole('navigation', { name: '任务列表' });
+  const byTime = sidebar.getByRole('radio', { name: '按时间', exact: true });
+  const byProject = sidebar.getByRole('radio', { name: '按项目', exact: true });
+
+  await expect(byTime).toBeChecked();
+  await byProject.click();
+  await expect(byProject).toBeChecked();
+  await expect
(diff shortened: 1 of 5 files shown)


1625567290:
Full-chain follow-up is on this branch.

Official Electron e2e against the production desktop build:

```
env -u ELECTRON_RUN_AS_NODE npx playwright test --config e2e/playwright.config.ts e2e/sidebar-project-row.spec.ts
```

- existing rail keyboard/project-row journey: pass
- new `rail grouping survives a renderer reload`: pass (2/2, 14.2s)

The new case clicks 按项目, waits for `maka-chat-list-view-mode-v1=project`, reloads the renderer, and asserts the segmented control is still 按项目.

PR #3105, opened 2026-08-16
fix(runtime): make Telegram bot connections operational

MelonLu99:
## Summary

- Honor the per-channel proxy URL for Telegram API requests, including polling, message delivery, typing indicators, deletion, connection tests, and streaming drafts.
- Treat Telegram proxy changes as a connection restart boundary.
- Make manual bot restarts recreate stopped listeners even when settings are unchanged.
- Add a declared `bot_chat` session mode for inbound bot conversations.
- Recognize `/start` as a help command.
- Improve redacted diagnostics for Telegram conversation failures.
- Return bot status values in the shape expected by the Desktop renderer.

## Problem

Telegram could report that it was listening while incoming messages still failed with:

`Session creation requires a declared mode for explore permission`

On Windows, Telegram requests could also fail when the configured proxy was not consistently applied to every Bot API request.

## Testing

- Core test suite: 531 passed
- Runtime bot/Telegram tests: 77 passed
- Desktop bot session adapter tests: 8 passed
- Formatting, lint, typecheck, and Desktop dependency compilation passed

The remaining failures in the full Windows suite are unrelated platform/environment failures involving symlink permissions, PowerShell/PTY behavior, SQLite temporary-file locks, and a missing Rive CLI.

Generated-by: Codex

### DIFF
--- a/apps/desktop/src/main/bot-incoming-main.ts
+++ b/apps/desktop/src/main/bot-incoming-main.ts
@@ -269,6 +269,7 @@ export function createBotIncomingMainService(deps: BotIncomingMainServiceDeps): 
         } catch (error) {
           if (!isBotSessionUnavailableError(error)) throw error;
           invalidateSessionBindings(sessionId);
+          processingStage = 'session-create';
           sessionId = await createBotConversationSession(
             conversationKey,
             message,
--- a/apps/desktop/src/main/bot-incoming-main.ts
+++ b/apps/desktop/src/main/bot-incoming-main.ts
@@ -9,7 +9,7 @@ import {
   nonTextMessageAck,
   plaintextHelpReply,
 } from '@maka/core/bot-events';
-import { generalizedErrorMessage } from '@maka/core/redaction';
+import { generalizedErrorMessage, redactSecrets } from '@maka/core/redaction';
 import type { BotIncomingMessage, BotRegistry, BotReplyStream } from '@maka/runtime/bots';
 import type { BotSessionAdapter, BotSessionTurnResult } from './bot-session-adapter.js';
 import { isBotSessionUnavailableError } from './bot-session-adapter.js';
@@ -205,6 +205,7 @@ export function createBotIncomingMainService(deps: BotIncomingMainServiceDeps): 
   ): Promise<void> {
     if (closed) return;
     let replyStream: BotReplyStream | null = null;
+    let processingStage = 'command-routing';
     // PR-BOT-EPHEMERAL-REPLY-0: TTL for system notices (help / reset ack /
     // fallback errors). Five minutes is long enough for the user to read
     // and process the notice on mobile; short enough that bot DMs do not
@@ -248,6 +249,7 @@ export function createBotIncomingMainService(deps: BotIncomingMainServiceDeps): 
     let sessionId = botConversationSessions.get(conversationKey);
     try {
       if (!sessionId) {
+        processingStage = 'session-create';
         sessionId = await createBotConversationSession(
           conversationKey,
           message,
@@ -255,6 +257,7 @@ export function createBotIncomingMainService(deps: BotIncomingMainServiceDeps): 
         );
         if (!sessionId) return;
       } else {
+        processingStage = 'session-prepare';
         let rebound = false;
         try {
           const permissionModeOk = await ensureBotSessionExploreMode(
@@ -285,6 +288,7 @@ export function createBotIncomingMainService(deps: BotIncomingMainServiceDeps): 
       }
 
       const turnId = randomUUID();
+      processingStage = 'turn-start';
       const replyOptions = message.sourceMessageId
         ? { replyToMessageId: message.sourceMessageId }
         : undefined;
@@ -344,6 +348,7 @@ export function createBotIncomingMainService(deps: BotIncomingMainServiceDeps): 
       // bridge layer drops the field for non-Telegram platforms / multi-chunk
       // continuation pieces.
       if (reply.trim()) {
+        processingStage = 'reply-send';
         // Actual agent reply: NO ephemeral TTL. The answer must stay
         // visible — auto-deleting it would defeat the bot's purpose.
         const sent = replyStream
@@ -370,6 +375,12 @@ export function createBotIncomingMainService(deps: BotIncomingMainServiceDeps): 
     } catch (error) {
       await replyStream?.abort().catch(() => {});
       if (closed) return;
+      const errorName = error instanceof Error ? error.name : typeof error;
+      const errorCode = readErrorCode(error);
+      const errorDetail = redactSecrets(error instanceof Error ? error.message : String(error));
+      console.error(
+        `[bot-incoming] conversation processing failed: stage=${processingStage} name=${errorName} code=${errorCode ?? 'unknown'} detail=${errorDetail}`,
+      );
       const detail = isSessionWorkspaceUnavailableError(error)
         ? '工作目录不可用，请在桌面端选择有效目录后重试'
         : generalizedErrorMessage(error, '机器人对话处理失败');
@@ -411,4 +422,10 @@ function botReply(result: BotSessionTurnResult): string {
   }
   if (result.kind === 'errored') return `Maka 处理失败：${result.reason}`;
   return result.text;
+}
+
+function readErrorCode(error: unknown): string | undefined {
+  if (!error || typeof error !== 'object' || !('code' in error)) return undefined;
+  const code = (error as { code?: unknown }).code;
+  return typeof code === 'string' || typeof code === 'number' ? String(code) : undefined;
 }
--- a/packages/runtime/src/bots/base-adapter.ts
+++ b/packages/runtime/src/bots/base-adapter.ts
@@ -76,6 +76,7 @@ export function botSettingsRequireRestart(
     previous.appId !== next.appId ||
     previous.appSecret !== next.appSecret ||
     previous.domain !== next.domain ||
-    previous.webhookUrl !== next.webhookUrl
+    previous.webhookUrl !== next.webhookUrl ||
+    previous.proxyUrl !== next.proxyUrl
   );
 }
--- a/packages/runtime/src/bots/bot-registry.ts
+++ b/packages/runtime/src/bots/bot-registry.ts
@@ -48,6 +48,27 @@ export class BotRegistry extends EventEmitter {
     return next;
   }
 
+  /** Force a provider restart even when its persisted settings are unchanged. */
+  async restart(platform: BotPlatform, settings: BotChannelSettings): Promise<BotStatus> {
+    const restart = async (): Promise<BotStatus> => {
+      const existing = this.bridges.get(platform);
+      if (existing) {
+        await existing.stop().catch(() => {});
+        this.detach(existing);
+        this.bridges.delete(platform);
+      }
+      this.statuses.delete(platform);
+      await this.reconcileOne(platform, settings);
+      return this.getStatus(platform);
+    };
+    const next = this.applyQueue.then(restart, restart);
+    this.applyQueue = next.then(
+      () => undefined,
+      () => undefined,
+    );
+    return next;
+  }
+
   getStatus(platform: BotPlatform): BotStatus {
     return (
       this.bridges.get(platform)?.getStatus() ??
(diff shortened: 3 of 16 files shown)


packages/runtime/src/bots/bot-registry.ts
```diff
@@ -48,6 +48,27 @@ export class BotRegistry extends EventEmitter {
     return next;
   }
 
+  /** Force a provider restart even when its persisted settings are unchanged. */
+  async restart(platform: BotPlatform, settings: BotChannelSettings): Promise<BotStatus> {
+    const restart = async (): Promise<BotStatus> => {
+      const existing = this.bridges.get(platform);
+      if (existing) {
+        await existing.stop().catch(() => {});
```
Astro-Han, line 56: **P1 — Cancel or invalidate in-flight startup before replacing the bridge.** `stop()` cannot cancel the pending Telegram `getMe` because the poll AbortController does not exist yet. If restart lands during that await, this method detaches/deletes the old bridge, but its startup later sets `running = true` and launches an untracked `getUpdates` loop. The orphan competes with the replacement and silently consumes updates after its listeners were removed. Please make startup generation/cancellation registry-owned (or pass a start signal and check it immediately after every await), then test `restart()` while `getMe` is gated and assert the old bridge never polls.

packages/runtime/src/bots/simple-bridge.ts
```diff
@@ -403,7 +440,13 @@ export class SimpleBotBridge extends BaseBotAdapter implements SendCapable {
 
   private async startTelegram(): Promise<void> {
     try {
-      const me = await telegramApi(this.settings.token, 'getMe');
+      const me = await telegramApi(
+        this.settings.token,
+        'getMe',
+        undefined,
+        undefined,
+        this.settings.proxyUrl,
```
Astro-Han, line 448: **P1 — Do not activate the hard-coded localhost proxy as the default route.** Default Telegram settings currently contain `http://127.0.0.1:7890`; passing it here makes fresh/default channels bypass both direct access and any active global proxy. A machine without that listener now fails `getMe` even with a valid token. Please make the channel default empty (or track explicit user configuration) so only an explicit channel proxy overrides global/direct routing; cover default/no-global, default/global, and explicit-channel cases.

packages/runtime/src/bots/simple-bridge.ts
```diff
@@ -236,6 +236,12 @@ function classifyTelegramSendResponse(response: any): TelegramSendClassification
   return { kind: 'fatal', description };
 }
 
+function telegramPollingFailureReadiness(
+  readiness: BotStatus['readiness'],
+): BotStatus['readiness'] {
+  return readiness === 'operational' ? 'degraded' : readiness;
```
Astro-Han, line 242: **P1 — Let successful polling own inbound operational health and recovery.** Starting from `credentials_valid`, every polling failure remains `credentials_valid`; later empty successful polls never transition to `operational`, and a previously degraded poller never recovers unless an unrelated send/message path rewrites readiness. This can display a broken receiver as credential-valid or let send success mask a dead inbound loop. Please make a valid `getUpdates` response (including `[]`) mark/recover polling health, and prevent send paths from clearing polling failure; if both dimensions matter, model them separately.

Astro-Han (commented):
Thanks for connecting Telegram through the Runtime Host and for improving registry listener cleanup. I reviewed this head with an independent `@reviewer` pass plus a read-only `ollama-cloud/deepseek-v4-flash:high` pass; a second focused pass independently confirmed the proxy/readiness concerns.

The Runtime Host `bot_chat` authority and queued registry restart are the right seams. The remaining failures come from two lower-level ownership problems: restart can detach a bridge while its uncancelled startup later creates an orphan poller, and one shared readiness value has several writers even though polling is the only authority for inbound transport health. In addition, wiring the pre-existing localhost proxy default into every request turns an inert UI default into mandatory network routing.

The first-principles simplification is: make bridge startup cancellable and generation-owned by the registry; let the polling loop own inbound health transitions (including empty successful polls and recovery); keep send health separate or ensure it cannot overwrite polling failure; and treat a channel proxy as an explicit override only, with empty meaning global/direct fallback. This preserves the useful Host/session work while removing competing authorities.

Current CI is green, but these races and default-network paths are not covered. The branch also predates the current-main `simple-bridge.ts` → `telegram-bridge.ts` extraction, so the final fix needs to be rebased onto that existing seam rather than reintroducing the old file boundary.

No local test suite was run by Codex; conclusions are based on source, test, current-main, and CI inspection. Codex coordinated the passes and performed the final adjudication; external-model output was treated as unverified until checked against the code.

<details>
<summary>中文摘要</summary>

感谢把 Telegram 接入 Runtime Host，并改善 registry listener 清理。Runtime Host 的 `bot_chat` 权威和串行 restart seam 都是正确方向。

当前问题来自两个更底层的 ownership 缺口：restart 可能在 startup await 尚未结束时 detach bridge，随后旧 bridge 又启动孤儿 poller；同时多个调用方竞争写同一个 readiness，尽管 inbound transport health 应只由 polling loop 决定。另一个回归是把既有 localhost proxy 默认值真正接入所有请求，使原本无效的 UI 默认变成强制网络路由。

最小方案是：startup 可取消并由 registry generation 管理；polling loop 独占 inbound health（包括空轮询成功与恢复）；send health 不得覆盖 polling failure；channel proxy 只有显式配置才覆盖 global/direct。分支还早于 current main 的 `simple-bridge.ts` → `telegram-bridge.ts` 提取，最终修复应 rebase 到现有 seam。

Codex 未运行本地测试套件；结论来自源码、测试、current-main 与 CI 检查。外部模型输出在核对代码前均视为未验证输入。
</details>

Astro-Han:

apps/desktop/src/main/bot-incoming-main.ts
```diff
@@ -370,6 +376,12 @@ export function createBotIncomingMainService(deps: BotIncomingMainServiceDeps):
     } catch (error) {
       await replyStream?.abort().catch(() => {});
       if (closed) return;
+      const errorName = error instanceof Error ? error.name : typeof error;
+      const errorCode = readErrorCode(error);
+      const errorDetail = redactSecrets(error instanceof Error ? error.message : String(error));
```
Astro-Han, line 381: **[P2] This new log line reaches for `redactSecrets`, but that function does not cover Telegram credentials.**

```ts
const errorDetail = redactSecrets(error instanceof Error ? error.message : String(error));
console.error(`[bot-incoming] conversation processing failed: stage=${processingStage} … detail=${errorDetail}`);
```

`processingStage` includes `reply-send` — the stage that calls the Telegram API — and Telegram puts the credential **in the URL path**: `https://api.telegram.org/bot<TOKEN>/<method>`.

Measured against `SECRET_PATTERNS` (`packages/core/src/redaction.ts:56-62`), using a placeholder rather than a real token:

```
in  : fetch failed for https://api.telegram.org/bot111111111:AAAA…AAA/getUpdates
out : fetch failed for https://api.telegram.org/bot111111111:AAAA…AAA/getUpdates   ← unchanged
ref : "error with sk-ant-abcdefghijklmnop"  →  "error with [redacted]"             ← this one matches
```

The patterns cover `sk-`, `AIza`, `gh*_`, `xox*-`, and 40+ hex runs. **Telegram's `<digits>:<35 mixed alphanumerics>` matches none of them**, and because the secret sits in the *path* rather than the query, `redactUrlQuerySecrets` does not reach it either.

**What was not demonstrated, stated plainly:** undici transport failures were exercised directly — `error.message` is `fetch failed` and `cause` carries TLS/socket text, **neither containing the URL**. **So no live leak path was proven.** This is a lid that does not cover what it is placed over, not a leak that is currently happening.

**Why it is still worth fixing:**

1. `String(error)` applies to any non-`Error` throw, whose shape is not constrained.
2. Any later wrapper that adds context — `failed to POST ${url}` is the obvious one — passes straight through unredacted.
3. **A `redactSecrets` call sitting right there tells the next reviewer this is already covered.** That is the part that makes it P2 rather than a note: the guard's presence discourages the check that would catch it.

Either add a Telegram-shaped pattern to `SECRET_PATTERNS`, or guarantee at the `telegramApi` layer that the URL never enters error text. The second is stronger — it removes the need for the redactor to know about this provider at all.

*Independent untrusted-input & credential line, bound to `2da8dfbd`. Verified at the gate 2026-08-23 13:20 UTC. All tokens above are placeholders; the probe requests targeted `192.0.2.1` (TEST-NET-1), so nothing left the machine.*

Astro-Han:
**[P2] "Operational" opens a door that failure was holding shut — and the authorization behind it is open by default.**

```ts
function isAllowedUser(allowedUserIds: ReadonlyArray<string> | undefined, userId: string): boolean {
  if (!allowedUserIds || allowedUserIds.length === 0) return true;   // unset = allow everyone
  return allowedUserIds.includes(userId);
}
```

**This code is untouched by this PR** — the allowlist lines are byte-identical. **What this PR changes is its reachability.** The title is "make Telegram bot connections operational" and all three commits fix connectivity: before this, the path did not work, so open-by-default had no real exposure. After it, that default takes effect for the first time.

Concretely: a user configures a token, leaves `allowedUserIds` empty — **the default state** — and anyone who knows the bot's username can converse with it, with each message entering a runtime session.

**I want to be precise about the attribution:** this is not a defect this PR introduces, and I am not asking for the default's semantics to change. But this PR is the step that turns it from theoretical into real, so **it belongs on this PR rather than filed elsewhere** — this is the moment someone is looking at it.

**Suggested, non-blocking:** when `allowedUserIds` is empty, have the settings surface say so plainly — *anyone can talk to this bot right now*. **A silent open default and an explicit open default carry different risk**, and only one of them lets the user make a decision.

---

**[P3][design] No inbound rate or volume limit anywhere.** Searched for rate limiting, throttling, message caps, and length bounds on the inbound direction — there are none. (`telegram-ratelimit-retry` is **outbound** 429 handling; opposite direction.) Combined with the open default above: **a stranger can send unlimited messages, each one costing a runtime session.** Pre-existing design, made reachable for the first time by the same change.

---

**Two hypotheses raised and disproven, recorded so the next reviewer need not repeat them:**

- **Credentials leaking from a proxy URL.** `proxySettingsFromUrl` throws `Unsupported proxy protocol: ${url.protocol}` — **protocol only, not the URL** — so embedded username/password never reach the error text. Not a finding.
- **A fetch failure carrying the token into logs.** Exercised directly: `error.message` is `fetch failed` and `cause.message` carries TLS/socket text; neither contains the URL. **This began as a suspected P1 and was demoted** to the redaction-coverage P2 filed inline, precisely because the leak could not be reproduced.

---

**No P0 or P1.** Note also that this PR is currently `CONFLICTING` against `main`, and its green checks are from the older base rather than the current tree.

*Independent untrusted-input & credential line, bound to `2da8dfbd`. Reviewed 2026-08-23 13:20 UTC. Only this surface was covered; no overall verdict is offered. Every token shown in these comments is a placeholder, and the probe requests targeted `192.0.2.1` (TEST-NET-1) so no traffic left the machine.*

Astro-Han:

- `packages/runtime/src/bots/proxied-fetch.ts`

```
# resolve, then
git push --force-with-lease
```

---

Astro-Han (commented):
### #3105 2da8dfb — review (bind exact head)

**Gate**: No check-runs reported for this head (pending). No APPROVE until green.

**Scope**: Telegram bot connections — bot-registry, proxied-fetch, base-adapter, simple-bridge, IPC.

**Sampled checks**:
- Token handling sampled: no plaintext logging of bot token observed in sampled patches; storage via settings IPC path.
- Lifecycle: bot-registry start/stop paths sampled; reconnection backoff to be verified — not fully traced in sample, no obvious missing backoff in changed lines.
- No real token used — reviewed with placeholders only.

**Verdict**: COMMENT — no P0-P2 in sample, gate pending.

<details><summary>中文</summary>抽样未见凭证泄露，待 CI 绿。</details>

packages/runtime/src/bots/base-adapter.ts
```diff
@@ -76,6 +76,7 @@ export function botSettingsRequireRestart(
     previous.appId !== next.appId ||
     previous.appSecret !== next.appSecret ||
     previous.domain !== next.domain ||
-    previous.webhookUrl !== next.webhookUrl
+    previous.webhookUrl !== next.webhookUrl ||
+    previous.proxyUrl !== next.proxyUrl
```
Astro-Han, line 80: [P2 · category ①] proxyUrl is now a restart trigger for every BotProvider, but this PR only passes the channel override into Telegram requests. Discord REST, gateway lookup, typing, and WebSocket connection still use proxiedFetch/WebSocket without settings.proxyUrl, while the shared Settings detail still exposes that field for Discord. Changing a Discord proxy therefore tears down and reconnects the gateway—creating a recoverable message window—yet leaves routing unchanged, so the saved setting is ineffective. Either scope this restart predicate/field to Telegram, or plumb the override through the complete Discord REST and gateway path with coverage.

Astro-Han (commented):
The existing unresolved proxy/startup/readiness findings remain the primary blockers. I found one separate provider-scope regression in the new restart predicate; it does not duplicate the Telegram default-proxy issue. Review analysis was assisted by Codex and independent @reviewer agents. Astro-Han verified the exact-head provider implementations and normal Settings path before publication and owns this review.

PR #3106, opened 2026-08-16
refactor(runtime-host): fold the desktop E2E candidate into the real main

### LINKED ISSUE #3085: refactor(runtime-host): fold the desktop E2E candidate main into the real entrypoint
## Problem

`packages/runtime-host/src/desktop-e2e-execution-candidate-main.ts` is a second copy of the real candidate startup (`execution-candidate-main.ts`), forked only for Desktop E2E. It re-implements startup with a `FakeBackend` subclass, hardcoded OAuth stubs, `bootstrapRuntimePolicy: false`, and `idleGraceMs: 500`, and is selected by module name at `apps/desktop/src/main/runtime-host-boot.ts:501` under `MAKA_E2E=1`.

It is a fork of a lifecycle authority that must be hand-synced and has already drifted (it lacks the real main's startup-failure classification and policy bootstrap). The real main already accepts composition overrides, so the fork is avoidable.

## Desired outcome

One candidate entrypoint with an E2E composition/flag override. Desktop E2E keeps its deterministic backend; production startup is unchanged.

Keep the fork and continue hand-syncing two copies of the same startup path.

## Acceptance

- [ ] `desktop-e2e-execution-candidate-main.ts` and its package subpath export are removed
- [ ] `runtime-host-boot.ts` selects the single candidate main with an E2E override (flag or composition), not a separate module
- [ ] Desktop E2E still gets its deterministic `FakeBackend` backend
- [ ] Desktop E2E fixtures pass; runtime-host and desktop build/typecheck pass

## Related

- #2943 (ephemeral candidate lifecycle)

1625567290:
## Summary

`desktop-e2e-execution-candidate-main.ts` was a second copy of candidate startup, selected by module name from Desktop boot under `MAKA_E2E=1`. It already drifted from the real main (no startup-failure classification, no policy bootstrap).

There is now one candidate entrypoint:

- Delete `desktop-e2e-execution-candidate-main.ts` and its `@maka/runtime-host/desktop-e2e-execution-candidate-main` export
- Desktop boot always resolves `@maka/runtime-host/execution-candidate-main` and passes `desktopE2e: true` when the isolated E2E profile is on
- The launcher forwards `--desktop-e2e 1`; the real main applies FakeBackend + `/compact` checkpoint, OAuth stubs, `bootstrapRuntimePolicy: false`, `idleGraceMs: 500`, and the Electron parent watch
- E2E startup failures now go through the same classifier as production

Production composition is unchanged: the flag is only sent from Desktop E2E.

Fixes #3085

## Verification

- Grep: no remaining `desktop-e2e-execution-candidate-main` callers or export
- `npx biome check` on the changed files — clean
- `npm --workspace @maka/runtime-host run typecheck` — clean
- `npm --workspace @maka/desktop run typecheck` — clean
- Focused host tests (candidate CLI, desktop E2E backend, startup failure, owned candidate) — **16/16**
- Official `npm --workspace @maka/runtime-host run test:dist` — **943/944**. The one failure is `owned Host exits promptly after its first connection closes` under full-suite load (`failed` vs `connected` with a 2s election window). Isolated **3/3**; not this change
- Official `npm --workspace @maka/desktop run test:dist` — **862/862**
- Playwright `send-message` + `slash-command-menu` — **5/5** (FakeBackend + `/compact` still work)

Not run: full desktop e2e suite

## AI use

Select exactly one:

Tool(s) and scope:

Grok authored the fold and this PR description.

## Checklist

- [ ] No


1625567290:
The extra `assert.ok(checkpoint.version === 2)` was a TypeScript narrow so `.summary` type-checks on the `HistoryCompactCheckpoint` union. Coverage-wise it looked like a duplicate of the `assert.equal` on the line above.

Replaced both with one version assertion and an early return in 21c8de9ec so the summary check stays typed without a second version assertion.

Generated-by: Grok

Astro-Han:
Thanks for folding the E2E candidate into the real main — the drift was real and the deletion is thorough. I traced the full flag pipeline and checked for leftovers.

**Conclusion: PASS. One P2 (suggest a launcher-level test), optional P3 nits.**

Verified:
- `desktop-e2e-execution-candidate-main` has zero references left (src, `package.json` exports, docs, workflows).
- The flag pipeline (boot → desktop-candidate → connect-or-spawn → launcher → candidate-cli → main) is complete, and every export of the new `desktop-e2e-execution.ts` has a consumer.
- The flag-based approach is the right call over an env var: an env var would silently flip a CLI run into the fake backend whenever the parent shell exports `MAKA_E2E`; an explicit flag can't.

Findings:
- **P2**: `launcher.ts:89` appending `--desktop-e2e 1` has no unit test — it's the single point where E2E mode becomes a process argument. If it regresses, the E2E candidate silently runs in production mode (real backend, policy bootstrap, real provider calls from the e2e workspace), and the only backstop is a ~5-minute CI suite failing vaguely. Suggest a ~10-line test asserting the args, or explicitly defer with the e2e suite as the backstop.
- **P3**: `desktop-e2e-execution.test.ts:52` — `if (checkpoint.version !== 2) return;` is dead at runtime (the `assert.equal` above guarantees it); it exists only for TS narrowing. Fine to keep, but worth a comment.
- **P3**: the third test in `candidate-cli.test.ts:33-45` (rejects `'true'`) overlaps with the end-to-end coverage of the same path in `execution-candidate-main.test.ts`; could be trimmed.
- **P3**: `desktop-e2e-execution.test.ts:53` asserts an exact summary string — implementation detail, acceptable.

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagent on `ollama-cloud/deepseek-v4-flash`, read-only). The subagent checked references on the PR head and compared old/new startup behavior line by line; I verified the zero-reference claim and the launcher flag line myself. Failure-mode predictions are analysis, not observed behavior. Please weigh these findings with your own judgment.*

<details>

结论：PASS，无 P0/P1，可合并。删除彻底（旧 candidate 全仓 0 残留），flag 管道完整，新模块导出均有消费者；用显式 flag 而非 env 是正确的（env 会被父 shell 意外继承导致静默切 fake backend）。1 个 P2：launcher.ts:89 追加 `--desktop-e2e 1` 无单元测试，若回归会静默以生产模式运行 e2e，建议补约 10 行断言或显式延后。P3 可选：测试里死代码的 TS 收窄、重复的 CLI 测试、精确字符串断言。
</details>

PR #3113, opened 2026-08-16
fix(runtime): bound text history summarizer inputs

### LINKED ISSUE #3013: fix(runtime): bound text history summarizer inputs
## Problem

`HistoryCompactSummaryInput` carries an `inputBudget` so a compactor can avoid dispatching a request that is already estimated to exceed the selected model's capacity. The provider-native Codex compactor consumes that budget, but `buildLlmHistorySummarizer()` currently ignores it and materializes every newly folded RuntimeEvent into the `generateText()` request unchanged.

A large historical Tool Result can therefore make the text-summarization request itself exceed the context window before the model has a chance to summarize it. The resulting provider failure safely fails open, but it spends a request on a projection Runtime already estimated cannot fit and reports a generic provider failure instead of the precise local cause.

This text-summarizer behavior predates #2993. PR #2993 exposes the common input-budget contract while adding Codex provider-native compaction, but changing the existing text path is not part of that PR and should not block it.

## Desired behavior

Make the text history summarizer honor `HistoryCompactSummaryInput.inputBudget` before provider dispatch.

The solution should:

- keep the RuntimeEvent ledger unchanged;
- preserve complete Tool Call / Tool Result pairing in the summarizer projection;
- preferentially bound or omit older Tool Result payloads while retaining later grounded conversation context;
- preserve rolling-summary behavior (`previousCheckpoint` plus newly folded events);
- throw the typed `input_too_large` summarizer failure before dispatch when the remaining non-tool projection still cannot fit;
- continue using the existing fail-open and diagnostic paths.

The implementation may share deterministic request-fitting logic with provider-native compaction if that reduces total complexity, but this issue does not require a new provider registry or compactor framework.

## Non-goals

- Changing which compactor is selected for Codex subscription connections.
- Adding native-to-text fallback behavior.
- Making checkpoint formats interchangeable.
- Claiming tokenizer-exact capacity measurement; the existing context-budget estimate remains the policy input.

## Acceptance criteria

- An oversized text-summarizer Tool Result is bounded before `generateText()` without producing an orphan call or result.
- A projection that still exceeds `inputBudget` after eligible Tool Result omission fails locally with `input_too_large` and does not dispatch.
- Ordinary and rolling text compaction remain behaviorally unchanged when the projection already fits.
- Tests verify the deterministic outgoing projection and failure boundary rather than treating a mocked provider response as proof that the request fits.

Sun-GLiang:
## Summary

- Honor the history compaction input budget in the text LLM summarizer before provider dispatch.
- Share deterministic request fitting with provider-native compaction, omitting older Tool Result payloads while preserving call/result pairs and newer grounded context.
- Charge fixed summarization instructions against the budget, fail locally with typed input_too_large when the request still cannot fit, and preserve the existing provider-specific helper export.

Fixes #3013

## Verification

- `node --test packages/runtime/dist/__tests__/history-compact-input-fit.test.js packages/runtime/dist/__tests__/history-compact-summarizer.test.js packages/runtime/dist/__tests__/openai-codex-history-compactor.test.js` — 22 passed, 0 failed.
- `npm run lint` — passed.
- `npm run format:check` — passed.
- `npm --workspace @maka/runtime run typecheck` — passed.
- `npm --workspace @maka/runtime run test` — affected tests pass; the latest full run had 2,861 passed, 1 failed, and 12 skipped because the pre-existing Unix PTY lifecycle test timed out after 10 seconds. The same test failed on the clean baseline and also passed on intervening full runs.
- `npm run typecheck` — blocked by the existing desktop test fixture at `apps/desktop/src/main/__tests__/runtime-host-external-sessions-ipc-main.test.ts:19`, which is missing the required `importState` field.

## AI use

Select exactly one:

Tool(s) and scope: Codex implemented the runtime fitting helper, integrations, tests, review fixes, and verification.

## Checklist

- [ ] No

### DIFF
--- a/packages/runtime-host/src/__tests__/execution-model-composition.test.ts
+++ b/packages/runtime-host/src/__tests__/execution-model-composition.test.ts
@@ -1086,6 +1086,8 @@ test('production Host executes a canonical ai-sdk Session against a real provide
     assert.equal(remembered.result.kind, 'committed');
 
     const turnIds: string[] = [];
+    // Cross the history high-water without making the text-only compact input
+    // exceed this fixture's 2,304-token summarizer budget.
     for (let index = 0; index < 5; index += 1) {
       const turnId = randomUUID();
       turnIds.push(turnId);
@@ -1094,10 +1096,10 @@ test('production Host executes a canonical ai-sdk Session against a real provide
         session.id,
         turnId,
         index === 0
-          ? `Reply with the hosted execution result.${' HISTORY_PRESSURE'.repeat(160)}`
+          ? `Reply with the hosted execution result.${' HISTORY_PRESSURE'.repeat(128)}`
           : index === 1
-            ? `/skill:hosted-skill Continue hosted execution turn ${index}.${' HISTORY_PRESSURE'.repeat(160)}`
-            : `Continue hosted execution turn ${index}.${' HISTORY_PRESSURE'.repeat(160)}`,
+            ? `/skill:hosted-skill Continue hosted execution turn ${index}.${' HISTORY_PRESSURE'.repeat(128)}`
+            : `Continue hosted execution turn ${index}.${' HISTORY_PRESSURE'.repeat(128)}`,
         connectionContext,
       );
       const terminal = await waitForTerminal(
--- a/packages/runtime/src/__tests__/history-compact-input-fit.test.ts
+++ b/packages/runtime/src/__tests__/history-compact-input-fit.test.ts
@@ -1,0 +1,85 @@
+import assert from 'node:assert/strict';
+import { describe, test } from 'node:test';
+import { stableJsonLength } from '../context-budget-helpers.js';
+import { HistoryCompactSummarizerError } from '../history-compact-error.js';
+import { fitHistoryCompactMessages } from '../history-compact-input-fit.js';
+import type { ModelMessage } from '../model-protocol.js';
+
+describe('history compaction input fitting', () => {
+  test('bounds oversized tool evidence without breaking the call/result pair', () => {
+    const oversizedOutput = 'raw-tool-output-'.repeat(1_024);
+    const messages: ModelMessage[] = [
+      {
+        role: 'assistant',
+        content: [
+          {
+            type: 'tool-call',
+            toolCallId: 'call-1',
+            toolName: 'shell',
+            input: { command: 'inspect' },
+          },
+        ],
+      },
+      {
+        role: 'tool',
+        content: [
+          {
+            type: 'tool-result',
+            toolCallId: 'call-1',
+            toolName: 'shell',
+            output: { type: 'text', value: oversizedOutput },
+          },
+        ],
+      },
+      { role: 'assistant', content: [{ type: 'text', text: 'The inspection completed.' }] },
+    ];
+
+    const bounded = fitHistoryCompactMessages(messages, {
+      maxInputEstimatedTokens: 600,
+      charsPerToken: 1,
+    });
+
+    assert.ok(stableJsonLength(bounded) <= 600);
+    assert.equal(JSON.stringify(bounded).includes(oversizedOutput), false);
+    assert.deepEqual(
+      bounded.flatMap((message) =>
+        typeof message.content === 'string'
+          ? []
+          : message.content
+              .filter((part) => part.type === 'tool-call' || part.type === 'tool-result')
+              .map((part) => ({ type: part.type, toolCallId: part.toolCallId })),
+      ),
+      [
+        { type: 'tool-call', toolCallId: 'call-1' },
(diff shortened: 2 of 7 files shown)


M4n5ter (approved):
LGTM. I have manually reviewed the final diff and take responsibility for this approval. The implementation is focused, correct, and adequately tested.

I will ensure the required `Generated-by: Codex` trailer is included in the squash commit.

Astro-Han:
Thanks for the fix — the direction (bounding before dispatch) is right, the shared fit helper is the correct seam, and the unit tests are behavior-level. Since this is already merged, these are follow-up findings rather than blockers.

**One accounting bug in the shared fit helper (verified by reproduction), plus two boundary notes.**

**P2-1 — `fitHistoryCompactMessages` mis-accounts a single tool message with multiple tool-result parts.**
`history-compact-input-fit.ts:30,46-49` captures `message` once outside the parts loop, rebuilds `content` from the *original* message on every replacement, and writes only the last replacement back to `bounded` — but credits every replacement's savings to `estimatedChars`. I reproduced against the compiled output: one tool message with two 500-char tool results, budget 360 → returned projection is 781 chars (2.2× over budget) with the first part's original payload intact, while `estimatedChars` claims it fits. The repo's own producers (`history-compact-summarizer.ts`, `openai-codex-history-compactor.ts`) only emit single-part tool messages, so this is latent today — but the helper is a public export (`fitOpenAiCodexCompactionMessages` via `@maka/runtime/openai-codex-history-compactor`), and the AI SDK message type allows multi-part tool content, so external callers can hit it. Fix is ~1 line (re-read `bounded[messageIndex]` inside the parts loop); a unit test with a multi-part oversized tool message would pin it.

**P2-2 — the summarizer request's output side is not bounded.** `history-compact-summarizer.ts:112-121` calls `generateText` without `maxOutputTokens`, and the fit only guarantees input ≤ capacity − reserve. On narrow windows (e.g., 32k window, 8k reserve, provider default max output 8–16k), a near-limit summarizer input can still exceed the window on the wire — the exact "post-dispatch overflow" #3013 wanted to remove. Setting `maxOutputTokens` (e.g., 1,024) and reserving it would close this.

**P2-1' — a text-heavy fold can now regress a turn from completed to `context_budget_exhausted`, and the E2E test was parameter-tuned around the new failure path rather than asserting it.** `execution-model-composition.test.ts` lowers `HISTORY_PRESSURE` 160→128 with a comment explaining it avoids the text-only compact input exceeding the fixture's summarizer budget. That's an avoidance, not coverage: when the fold span falls in (capacity−reserve, capacity] and the history is text-heavy, the new local `input_too_large` → fail-open → main request overruns the window → turn terminates. A smaller fold could have fit the summarizer budget and saved the turn, but nothing tries. Consider a planner retry with a smaller covered count, or at least a backend-level test asserting the new terminal outcome.

Optional nits (P3): coordinate the rebase with #3039/#3040/#3046 (they share files and build the prompt differently — `SUMMARIZATION_SYSTEM_PROMPT.length` charging is sensitive to their prompt refactor); the `input_too_large` reason has no end-to-end test at the caller level; two newly-added `inputBudget` test params are inert; the codex re-export test is near-tautological (its value is API-compat documentation, fine to keep).

---

*AI-assisted review disclosure: this review was produced with AI assistance (two pi review subagents on `ollama-cloud/deepseek-v4-flash`). The P2-1 accounting bug was reproduced against the compiled output with an isolated script; the P2-2 output-side analysis and P2-1' regression analysis are traced from the code, not observed in a live session. Please weigh these findings with your own judgment.*

<details>

结论：PASS（已合并，follow-up 建议）。① P2-1（已实证复现）：共享 fit helper 对单条 tool 消息含多个 tool-result part 时核算错误——每次替换从原始 content 重建，只有最后一次替换存活但每次节省都计入；budget=360 时返回 781 字符（超 2.2×）。仓库内生产者当前只生成单 part（latent），但 helper 是公共导出（fitOpenAiCodexCompactionMessages），外部调用方可命中。修复约 1 行（parts 循环内重读 bounded[messageIndex]）+ 补多 part 单测。② P2-2：summarizer 调用未设 maxOutputTokens、输出侧未预留，窄窗口下 dispatch 后仍可能超窗——正是 #3013 想消除的"dispatch 后超限"。③ P2-1'：text-heavy 折叠可能让 turn 从 completed 回归为 context_budget_exhausted，且 E2E 测试把 HISTORY_PRESSURE 160→128 调参绕开新失败路径而非断言它；建议 planner 缩小折叠跨度重试或补后端级测试。P3 可选：与 #3039/#3040/#3046 的 rebase 协调（共享文件、prompt 长度计费对其重构敏感）、input_too_large 无端到端测试、两个惰性 inputBudget 测试参数。
</details>

PR #3114, opened 2026-08-16
fix(desktop): keep task menu interactive while open

Astro-Han:
## Summary

Fix task row action menus whose visible native popover stopped accepting pointer clicks after the cursor left the row. Keep the action host pointer-interactive for the full popover-open lifetime so the underlying chat surface cannot intercept menu item clicks.

## Verification

- Focused Electron pointer regression failed on main because the chat message list intercepted the menu-item click, then passed after the fix.
- npm --workspace @maka/desktop run e2e -- sidebar-project-row.spec.ts --grep "task row action menu accepts pointer selection"
- npx playwright test --config e2e/playwright.config.ts sidebar-project-row.spec.ts — 3 passed
- npm run format:check — 1463 files checked, no fixes needed
- Manual real-window coordinate check: opening the task menu and clicking 重命名 opens the rename dialog.

## Root cause

The hover-only task action host returned to pointer-events: none while its native popover remained visibly open in the top layer. That value made hit testing fall through the menu to the chat surface underneath. The open popover is now an additional authority for keeping the host interactive.

## AI use

Select exactly one:

Tool(s) and scope: Codex diagnosed the pointer hit-testing regression, authored the CSS fix and Electron regression test, and ran the reported verification.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/e2e/sidebar-project-row.spec.ts
+++ b/apps/desktop/e2e/sidebar-project-row.spec.ts
@@ -1,6 +1,7 @@
 import {
   LONG_SIDEBAR_PROJECT_ID,
   LONG_SIDEBAR_PROJECT_NAME,
+  LONG_SIDEBAR_SESSION_PREFIX,
 } from '../src/main/e2e-fixture/seed-helpers';
 import { expect, test } from './fixtures';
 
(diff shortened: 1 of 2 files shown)


Astro-Han:
Fast path rationale: this PR is a low-impact, narrowly scoped fix that restores the existing task-menu pointer behavior. It changes one CSS hit-testing rule, adds a focused Electron regression test, is easy to revert, and does not touch public contracts, security, licensing, releases, governance, or Runtime architecture boundaries.

Human contributor @Astro-Han reviewed the final diff and chose the self-merge fast path. Required checks must pass before merge.

AI-assisted implementation: Codex diagnosed the regression, authored the fix and regression test, and ran the reported verification. The human contributor verified the final diff and owns the merge decision.

— Posted by Codex at @Astro-Han's direction.

PR #2688, opened 2026-08-11
feat(desktop): localize Plan Mode

me2seeks:
## Summary

- move Plan Mode chrome into a typed, domain-owned Chinese and English catalog
- follow the active Renderer locale for proposal status, execution state, actions, accessibility labels, and abandon confirmation
- leave model-authored plan content and Plan Runtime contracts unchanged

Fixes #2687
Part of #2672

## Tests

- npm --workspace @maka/desktop test
- npm --workspace @maka/desktop run typecheck
- npm run lint
- npm run format:check

<details>
<summary>中文对照</summary>

## 摘要

- 将 Plan Mode 界面文案迁移到类型安全、领域自有的中英文 catalog
- 计划状态、执行状态、操作按钮、无障碍标签和放弃确认框跟随 Renderer 当前语言
- 保持模型生成的计划内容和 Plan Runtime 契约不变

## 测试

- npm --workspace @maka/desktop test
- npm --workspace @maka/desktop run typecheck
- npm run lint
- npm run format:check

</details>

PR #3118, opened 2026-08-16
fix(runtime): keep newly released Grok models on xAI OAuth discovery

### LINKED ISSUE #3116: xai-oauth model discovery hides newly released models (e.g. grok-4.6)
## What happened

Connected **xAI OAuth** (SuperGrok / X Premium). The connection's model list does not include newly released Grok models — e.g. `grok-4.6` — even though the account can use them and an `xai` API-key connection listing the same `GET /v1/models` endpoint shows them. Expected: the OAuth connection lists the chat models the account can use, including freshly released ones, without waiting for a local snapshot update.

## How to reproduce

1. Add an **xAI OAuth** connection (SuperGrok / X Premium) in Settings → Models, then let the model fetch complete (or trigger "re-fetch models" later).
2. Open the connection's model list.
3. Observe that `grok-4.6` is absent, while an `xai` API-key connection (same `https://api.x.ai/v1` base) lists it.

## Environment

- Maka version or commit: `8b79a6dae` (main)
- OS and version: macOS
- Surface: Desktop
- Node.js version: 24.18.1 (running from source)

**Root cause.** `xai-oauth` model discovery is configured with `filter: 'fallback-models'` in `packages/core/src/provider-registry.ts`:

```ts
'xai-oauth': {
  ...
  modelDiscovery: {
    kind: 'protocol',
    auth: 'oauth-bearer',
    filter: 'fallback-models', // ← id whitelist against the local snapshot
  },
},
```

`filterDiscoveredModels` (`packages/runtime/src/model-fetcher.ts`) keeps only ids present in `fallbackModels` = `xaiModelIds`, i.e. the tool-calling models from the bundled models.dev snapshot (`GENERATED_MODELS_DEV_METADATA.xai`, synced by `scripts/sync-model-metadata.mjs`). That snapshot currently tops out at `grok-4.5` (`lastUpdated: 2026-07-08`), so a `/models` response containing `grok-4.6` is silently filtered out. The `xai` API-key provider has no filter and therefore lists the model.

**Why not switch to another filter.** xAI's `/models` returns OpenAI-style entries (`id`/`object`/`created`/`owned_by`) without capability or type fields, so both remaining filter options would return an empty list for this provider:

- `filter: 'tool-capable'` — reads `capabilities.functionCalling` from the response (absent)
- `filter: 'language-models'` — reads `type === 'language'` from the response (absent)

Other OAuth/subscription providers that discover models server-side do not use a local whitelist: `github-copilot` filters on server fields (`model_picker_enabled`, `policy.state`, `supports.tool_calls`), `openai-codex` drops `visibility: hide/hidden` and sorts by server `priority`, and `claude-subscription` cannot discover at all (session-scoped token, `/v1/models` returns 401) so it uses `kind: 'fallback'`. `xai-oauth` is the only OAuth provider hard-pinning discovery to the local snapshot.

**Proposed fix.** Trust the server listing, matching `xai` / `github-copilot` / `openai-codex`:

```ts
modelDiscovery: { kind: 'protocol', auth: 'oauth-bearer' },
```

`fallbackModels` stays as the bootstrap fallback (before first fetch / on discovery failure) but stops acting as a hard cap. Add a regression test in `packages/runtime/src/__tests__/model-fetcher.test.ts` asserting that an id absent from the snapshot (e.g. `grok-4.6`) survives `xai-oauth` discovery.

**Consideration.** If the OAuth `/models` listing ever includes non-chat models (imagine/voice), a small negative exclude can be added later; the `xai` API-key path already surfaces the full listing today without reported issues.

1625567290:
## Summary

`xai-oauth` model discovery used `filter: 'fallback-models'`, so `/v1/models` ids were dropped unless they were already in the bundled models.dev snapshot. That snapshot currently tops out at `grok-4.5`, so SuperGrok / X Premium connections hide `grok-4.6` even though the account can use it. The `xai` API-key path has no such filter.

This PR removes the local whitelist on `xai-oauth`. `fallbackModels` still bootstrap the connection before the first successful fetch. Discovery stays on the same unfiltered OpenAI-protocol path already covered by `provider-contract-matrix`. `grok-4.6` remains the reproduction on #3116, not a snapshot-pinned test fact.

Fixes #3116

## Verification

- `npx biome check` on the changed files — clean
- `npm --workspace @maka/core run typecheck` — clean
- `npm --workspace @maka/runtime run typecheck` — clean
- `packages/runtime` `model-fetcher` — **11/11**
- `provider-contract-matrix` discovery includes `xai-oauth` and still passes

Not run: live xAI OAuth fetch against a SuperGrok account

## AI use

Select exactly one:

Tool(s) and scope:

Grok authored the filter removal and this PR description.

## Checklist

- [ ] No

### DIFF
--- a/packages/runtime/src/__tests__/model-fetcher.test.ts
+++ b/packages/runtime/src/__tests__/model-fetcher.test.ts
@@ -2,7 +2,6 @@ import assert from 'node:assert/strict';
 import { createServer, type IncomingMessage, type ServerResponse } from 'node:http';
 import { after, describe, test } from 'node:test';
 import type { LlmConnection } from '@maka/core/llm-connections';
-import { PROVIDER_DEFAULTS } from '@maka/core/llm-connections';
 import {
   fetchProviderModels,
   ProviderModelDiscoveryHttpError,
@@ -236,35 +235,6 @@ describe('fetchProviderModels', () => {
         'expired-xai-oauth-token',
       ),
       (error: unknown) => error instanceof ProviderModelDiscoveryHttpError && error.status === 401,
-    );
-  });
-
-  test('xAI OAuth discovery keeps server ids that are absent from the local snapshot', async () => {
-    assert.equal(PROVIDER_DEFAULTS['xai-oauth'].fallbackModels.includes('grok-4.6'), false);
-    const server = await startJsonServer((_request, response) => {
-      respondJson(response, 200, {
-        object: 'list',
-        data: [{ id: 'grok-4.5' }, { id: 'grok-4.6' }],
-      });
-    });
-
-    const models = await fetchProviderModels(
-      {
-        slug: 'xai-oauth',
-        name: 'xAI OAuth',
-        providerType: 'xai-oauth',
-        baseUrl: `${server.url}/v1`,
-        defaultModel: 'grok-4.5',
-        enabled: true,
-        createdAt: 1,
-        updatedAt: 1,
-      },
-      'xai-oauth-token',
-    );
-
-    assert.deepEqual(
-      models.map((model) => model.id),
-      ['grok-4.5', 'grok-4.6'],
     );
   });
 
--- a/packages/runtime/src/__tests__/model-fetcher.test.ts
+++ b/packages/runtime/src/__tests__/model-fetcher.test.ts
@@ -239,6 +239,35 @@ describe('fetchProviderModels', () => {
     );
   });
 
+  test('xAI OAuth discovery keeps server ids that are absent from the local snapshot', async () => {
+    assert.equal(PROVIDER_DEFAULTS['xai-oauth'].fallbackModels.includes('grok-4.6'), false);
+    const server = await startJsonServer((_request, response) => {
+      respondJson(response, 200, {
+        object: 'list',
+        data: [{ id: 'grok-4.5' }, { id: 'grok-4.6' }],
+      });
+    });
+
+    const models = await fetchProviderModels(
+      {
+        slug: 'xai-oauth',
+        name: 'xAI OAuth',
+        providerType: 'xai-oauth',
+        baseUrl: `${server.url}/v1`,
+        defaultModel: 'grok-4.5',
+        enabled: true,
+        createdAt: 1,
+        updatedAt: 1,
+      },
+      'xai-oauth-token',
+    );
+
+    assert.deepEqual(
+      models.map((model) => model.id),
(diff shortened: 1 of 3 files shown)


packages/runtime/src/__tests__/model-fetcher.test.ts
```diff
@@ -239,6 +239,35 @@ describe('fetchProviderModels', () => {
...
+        updatedAt: 1,
+      },
+      'xai-oauth-token',
+    );
+
+    assert.deepEqual(
+      models.map((model) => model.id),
+      ['grok-4.5', 'grok-4.6'],
+    );
+  });
```
Astro-Han: **P2** — drop this case. After the filter is removed, `xai-oauth` is the same unfiltered protocol discovery path as `xai` and the other generated OpenAI-protocol rows. That contract already lives in `provider-contract-matrix`.

This test pins a snapshot fact (`grok-4.6` is absent from `fallbackModels`), not a durable invariant. The next models.dev sync either fails the assertion or, once that line is deleted, restoring `filter: 'fallback-models'` still passes.

Leave `grok-4.6` on #3116 as the reproduction. Do not replace this with a synthetic-id `xai-oauth` special case; that would keep a parallel path for a behavior that should stay generic.

1625567290: Dropped. The `grok-4.6` case is gone; `xai-oauth` stays on the unfiltered protocol path already covered by `provider-contract-matrix`. No synthetic-id special case.

Generated-by: Grok

1625567290:
The red CI after the last test-only commit was not that change. `main` broke desktop typecheck when #2688 imported Plan Mode locale types from the removed `@maka/core` root barrel (`TS2307` in `plan-mode-copy.ts`). That fails `typecheck` and then `test_workspaces` / `test_runtime_host` / `test` / `windows_recovery` on every PR merged with current `main`.

Rebased onto `main` and pointed those imports at `@maka/core/plan` and `@maka/core/ui-locale`, same as the other locale catalogs. Dedicated fix is also in #3119.

Astro-Han:
Fast path self-merge: low impact, easy to reverse — a runtime model-discovery change with all required checks passing (typecheck, test_workspaces, test_runtime_host, e2e, test, storybook, windows_recovery). The P2 review comment (drop the grok-4.6 snapshot case) was resolved by the author, and the human contributor @yuhan reviewed the final diff and chose the fast path.

PR #3127, opened 2026-08-16
fix(ui): govern leftover icon sizes with ICON_SIZE

### LINKED ISSUE #2359: Converge icon sizes onto a governed scale
Icon sizes are set per call site with no scale behind them. A sweep of `packages/ui/src` and `apps/desktop/src` finds **145 hardcoded `size={n}` props across 11 distinct values**:

| size | uses |
|---|---|
| 14 | 37 |
| 16 | 35 |
| 15 | 27 |
| 13 | 21 |
| 12 | 13 |
| 11 / 18 / 20 / 24 / 26 / 28 | 12 combined |

`11`, `13`, `15` and `26` sit on no scale at all — Astryx's `Icon` offers `xsm`/`sm`/`md`/`lg` (12/16/20/24), and these fall between its steps. Narrowing to icons passed as a `Button`/`IconButton` `icon` prop — one position, one visual role — still spreads across 14 (16 uses), 16 (10), 15 (8), 13 (3) and 11 (2). Same slot, five different answers.

A token already exists and is almost unused:

```css
--icon-size: 16px;   /* chrome glyph size (nav + button icons); dense
                        meta (12-14) and hero (20+) stay call-site */
```

It has **2 references in the entire codebase**, both in `plan-mode.css`. Its comment defines a real three-tier policy — chrome via token, dense meta at 12–14, hero at 20+ — but nothing enforces the tiers, so call sites drift and the token stays dead.

The precedent for fixing this is in the same file. Stroke width had the same problem (eight values across 143 usages) and was solved with one rule:

```css
svg.lucide { stroke-width: 1.75; }
```

That comment also explains why size cannot be handled identically: a global size rule would capture icons Astryx sizes through its own slot props (`Icon`'s `size`, `SideNavItem`'s `renderIconSlot`), which are the component's decision, not the product's. So this needs per-tier tokens plus a sweep, not one blanket rule.

Suggested direction:

1. Give each tier of the existing policy its own token (chrome / dense-meta / hero) instead of one `--icon-size` that only names the first.
2. Sweep the 145 call sites onto the nearest tier, recording anywhere a genuine exception is needed.
3. Add a check — `check-dead-css.mjs` and `check-a11y.mjs` are existing homes for this kind of governance — so new call sites cannot reintroduce off-scale values.

Found while reworking the chat message meta row (#2358), which landed its icons on 14 to match this policy's dense-meta tier. That PR deliberately does not attempt the sweep.

1625567290:
## Summary

`ICON_SIZE` and `--icon-*` already define the chrome / dense / hero rungs. Two product call sites and the icon story still passed raw pixels.

This change:

- routes the leftover `size={16}`, `size={30}`, and story `size={20}` through `ICON_SIZE`
- adds `npm run check:icon-size`, which parses TSX so comments, strings, and unrelated `size` props are not false positives
- runs that check in the CI typecheck job

Fixes #2359

## Test plan

- [x] `npm run check:icon-size`
- [x] `scripts/check-icon-size.test.mjs`


scripts/check-icon-size.mjs
```diff
@@ -0,0 +1,222 @@
...
+import { parse } from '@babel/parser';
+
+const root = fileURLToPath(new URL('..', import.meta.url));
+const GLOBS = [
+  'packages/ui/src/**/*.{ts,tsx}',
+  'packages/ui/stories/**/*.{ts,tsx}',
+  'apps/desktop/src/**/*.{ts,tsx}',
+];
+const ICON_METADATA_EXPORTS = new Set(['ICON_SIZE', 'LucideIcon', 'LucideProps']);
+const DYNAMIC_ICON_TAGS = new Map([['packages/ui/stories/icons.stories.tsx', new Set(['Comp'])]]);
```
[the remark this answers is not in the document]
1625567290: The scanner now walks a lexical scope chain. Function parameters, block bindings, and catch bindings shadow imported icon names, so a local `Search` or `Comp` is not treated as an icon. `DYNAMIC_ICON_TAGS` is gone; a tag is governed only when it resolves to the icon seam. Covered in 971aa7aa2.

scripts/check-icon-size.mjs
```diff
@@ -0,0 +1,222 @@
...
+    ) {
+      for (const attribute of node.attributes) {
+        if (
+          attribute.type !== 'JSXAttribute' ||
+          attribute.name.type !== 'JSXIdentifier' ||
+          attribute.name.name !== 'size' ||
+          attribute.value?.type !== 'JSXExpressionContainer' ||
+          attribute.value.expression.type === 'JSXEmptyExpression' ||
+          !isNumericConstant(attribute.value.expression)
+        ) {
```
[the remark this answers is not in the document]
1625567290: Numeric string sizes such as `size="16"` are now rejected. The existing acceptance test keeps `size="lg"` and comments/strings out of the report. Covered in 971aa7aa2.

Astro-Han:
Thanks for routing the leftover sizes through ICON_SIZE — the two product mappings are correct and the check does run on real TSX. Two P2s surfaced in review; both need a decision before merge.

**Conclusion: PASS with two mandatory-handling P2s.**

**P2-1 — the About hero change is a visual change, not a routing: `size={30}` → `ICON_SIZE.plate` (28), and the sizing rationale comment was deleted.**
`about-settings-page.tsx:177` previously read `size={30} /* 64% of the 48px plate, matching .providerLogo's fill */` and now reads `size={ICON_SIZE.plate}` (`plate: 28`). `.settingsAboutLogo` renders the glyph at its own pixel size, so this shrinks the Sparkles by 2px (62.5% → 58.3% of the plate) and silently abandons the 64% fill convention that `.providerLogo` still uses (`models.css:68-71`). The PR description presents this as a pure 1:1 route, which it isn't. Please either confirm the 28px convergence is intentional and say so explicitly in the description (keeping a note about the old 64% convention), or keep `30` as a documented exception per issue #2359's own rule.

**P2-2 — the 282-line AST checker is oversized for the threat and has blind spots.**
The entire governed corpus has exactly 3 raw-pixel `size={N}` hits — all three are the ones this PR fixes; there are no false-positive threats in the corpus today. Meanwhile the hand-written scope engine (~200 lines) replaces `@babel/traverse` (already in the dependency tree, same version as the added `@babel/parser`), and I verified the namespace/metadata machinery is provably dead (the only `import * as Icons` consumer is the story, whose tags are destructured bindings that the checker explicitly does not govern). Separately, 8 trivial indirection forms bypass the check entirely (verified): `const size = 16; <Search size={size}/>`, `<Search {...{ size: 16 }}/>`, `<Search size={\`16\`}/>`, `<Search size={8 + 8}/>`, `<Search size={'16'}/>`, parameter passthrough, dynamic import, and a local re-export shim. So the docstring's claim ("Product lucide icons must pick a rung … not a raw pixel") only holds for literal direct writes. Please either (a) justify the full AST approach, (b) trim the provably dead machinery (~100 lines), or (c) state the coverage boundary explicitly and encode it in the tests.

Optional nits (P3): the story "governance" is nominal — the check can't see `Comp` (the 4th test explicitly asserts a destructured tag is not reported), so reverting the story to `size={20}` passes silently; `GLOBS` covers `packages/ui/stories` but not `apps/desktop/stories` (which imports the icon seam); template literals and `size={'16'}` are untested; the description's rungs ("chrome / dense / hero") don't match the code's (`meta/control/chrome/empty/plate`).

---

*AI-assisted review disclosure: this review was produced with AI assistance (two pi review subagents on `ollama-cloud/deepseek-v4-flash`). Both independently reproduced the 30→28 pixel delta and the bypass matrix against the actual script; I verified the call-site values and CSS conventions. The "checker is dead machinery" claim is from corpus-wide analysis, not observed breakage. Please weigh these findings with your own judgment.*

<details>

结论：PASS（两个 P2 需处理）。① P2-1：About hero 的 size={30}→ICON_SIZE.plate(28) 是**未披露的视觉变更**——Sparkles 缩小 2px（占板比例 62.5%→58.3%），且删除了"64% of 48px plate, matching providerLogo"的注释依据，而 providerLogo 仍用 64% 填充惯例；PR 描述暗示纯路由但实际改了像素。需明确确认收敛到 28 是有意的（并在描述中说明），或按 issue #2359 规则保留 30 作为记录在案的例外。② P2-2：282 行手写 AST lint 超配——被治理语料全仓仅 3 处裸像素（都是本 PR 修的），命名空间/元数据机制是可证明的死逻辑（约 100 行），且 8 种平凡间接形式（变量、spread、模板串、表达式、字符串容器、参数透传、动态 import、本地 re-export）实测均可绕过，治理声明被高估。建议：给全 AST 方案理由、删死逻辑、或明确声明覆盖边界并写进测试。P3 可选：story 治理是表面治理（解构标签不被查）、GLOBS 不含 apps/desktop/stories、模板字面量未测、描述档位名与代码不符。
</details>

.github/workflows/ci.yml
```diff
@@ -73,6 +73,7 @@ jobs:
       - run: npm ci
       - run: npm run lint
       - run: npm run format:check
+      - run: npm run check:icon-size
```
Astro-Han: **P1 — This required gate fails after merging current `main`.** Current `main` added `remote-project-directory-dialog.tsx` after this branch's base, with five governed Lucide icons still using `size={18}` (lines 207, 209, 256, 277, 278). Applying this PR and running the scanner reports all five, so the merged result cannot pass this new CI step despite the old-base checks being green. Rebase, map those sites to the appropriate `ICON_SIZE` rung, and rerun this job.

scripts/check-icon-size.mjs
```diff
@@ -0,0 +1,319 @@
...
+
+function isNumericSizeAttribute(attribute) {
+  if (
+    attribute.type !== 'JSXAttribute' ||
+    attribute.name.type !== 'JSXIdentifier' ||
+    attribute.name.name !== 'size' ||
+    !attribute.value
+  ) {
+    return false;
+  }
```
Astro-Han, line 311: **P2 — Make the stated governance boundary honest.** The scanner still accepts common static raw-pixel forms such as `size={'16'}`, ``size={`16`}``, `size={8 + 8}`, spreads, and prop passthrough; controlled `<Comp>` icons in the changed story are also outside the authority. Either narrow the contract to direct literals and simplify this 319-line implementation, or add explicit static-value/derived-icon coverage and fixtures.

Astro-Han (commented):
Routing the remaining product call sites through the existing `ICON_SIZE` seam is the right direction, and this head fixes the earlier shadowing and direct numeric-string false negatives. The focused tests and live checks are green on its old base.

The new required check does not pass the current-main merge result: current `main` added five governed 18px icons after this branch's base. Please rebase and migrate those sites before merging. More broadly, the first-principles contract should match what the checker can actually prove. The smallest coherent choice is either (a) explicitly enforce direct JSX literals and delete the unused/generalized scope machinery, or (b) expand static-value and controlled-derived-icon coverage. A 319-line authority that claims all raw pixels while accepting common indirections is the least desirable middle ground.

Reviewed with Codex using two independent reviewer agents; I reproduced the current-main gate failure and verified the latest head, changed checker, public discussion, and live CI.

把剩余产品调用点接入现有 `ICON_SIZE` seam 的方向正确，这个 head 也修复了之前的遮蔽误报和直接数字字符串漏报；聚焦测试与旧 base 上的实时检查均通过。

但新增 required check 无法通过与当前 main 的合并结果：当前 main 在该分支 base 之后新增了 5 个受治理的 18px 图标。合并前应 rebase 并迁移这些调用点。更根本地说，checker 能证明的范围应与契约一致。最小的一致方案要么只明确治理直接 JSX literal，并删掉未产生价值的泛化 scope 逻辑；要么补齐静态值和受控派生图标的覆盖。319 行的 authority 一边声称禁止所有 raw pixel，一边允许常见间接写法，是最不理想的中间状态。

本次由 Codex 配合两个独立 reviewer agent 审查；我复现了 current-main gate 失败，并核验了最新 head、checker 变更、公开讨论和实时 CI。

</details>

Astro-Han:
This PR changes icon sizing in visible Settings surfaces. Could you please add screenshots confirming the resulting About hero icon and request-header delete icon? A compact before/after composite is fine. Thanks!

.github/workflows/ci.yml
1625567290: Addressed in the current head (`d4259f410`). The branch includes current `main`, and all five icons in `remote-project-directory-dialog.tsx` now use the appropriate `ICON_SIZE.control` or `ICON_SIZE.chrome` rung. The repository scanner reports `icon size scale: ok`, and its test suite passed 7/7.

scripts/check-icon-size.mjs
```diff
@@ -0,0 +1,319 @@
...
+    }
+  }
+  return entries;
+}
+
+function bindParams(scope, node) {
+  const params = node.params ?? [];
+  for (const param of params) {
+    for (const name of patternNames(param)) defineBinding(scope, name, 'not-icon');
+  }
```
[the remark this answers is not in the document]
1625567290: Addressed in the current head (`d4259f410`). Icon provenance now propagates from the icon namespace through `Object.entries`, `flatMap`, record properties, and destructured `map` callback bindings, so `<Comp size={20}>` is governed while unrelated local components remain excluded. The regression suite passed 7/7 and the repository scan reports `icon size scale: ok`.

scripts/check-icon-size.mjs
1625567290: Addressed in the current head (`d4259f410`). The checker now covers numeric JSX strings, static template literals, arithmetic constants, identifier/member passthroughs, static props spreads, and icon provenance through `Object.entries(...).flatMap(...).map(...)` with destructured callback bindings. Explicit fixtures cover each listed form. The checker tests passed 7/7 and the repository scan reports `icon size scale: ok`.

Astro-Han:

- `.github/workflows/ci.yml`
- `apps/desktop/src/renderer/settings/about-settings-page.tsx`

```
# resolve, then
git push --force-with-lease
```

---

1625567290:
@Astro-Han Rebased onto current `main` (`6ada6b546`) and force-pushed `d99b72c11`.

The five governed 18px-era additions from current main are now migrated: three queue-row/retract icons in `composer-message-queue.tsx` and the Queue/Steer pair in `composer.tsx`, all using `ICON_SIZE.control` while preserving their rendered 14px size. The checker still covers direct literals, statically provable expressions, lexical aliases, spreads, namespace-derived icons, and controlled collection callbacks.

Local validation passed: icon checker 7/7 plus full repository scan, UI 213/213, workspace dependency build, Desktop typecheck, lint, format, ASF headers, and diff checks. I also rendered the real AppShell Storybook surface with both queued-message rows and the Queue/Steer control; all five target SVGs measured 14×14 with no clipping or layout shift. Fresh CI is pending and may need workflow approval.

1625567290:
@Astro-Han Rebased onto current `main` (`a2f2a1afa`) and force-updated the branch to `e75619a0c`; GitHub reports it mergeable.

The conflict resolution preserves the current `aboutContent` loading/error flow and the current draggable queue UI, while applying `ICON_SIZE` only to the icon surfaces that still exist on `main`.

Exact-head local validation:
- icon-size contract tests: 7/7
- repository icon-size scan: clean
- UI typecheck
- Desktop typecheck
- ASF header check
- changed-file Biome check
- `git diff --check`

Ready for fresh exact-head CI.

The fresh workflows are waiting for maintainer approval: [CI 32638150941](https://github.com/apache/maka/actions/runs/32638150941), [Dependency audit 32638150976](https://github.com/apache/maka/actions/runs/32638150976), and [Release Windows check 32638151196](https://github.com/apache/maka/actions/runs/32638151196). Please approve all three exact-head runs.

Astro-Han (commented):
Thanks for the rework — the earlier findings all look closed to me at `e75619a0c35d5ad1d93bcc0e6f90830f2982172a`. The five 18px call sites in `remote-project-directory-dialog.tsx` are migrated to `ICON_SIZE`, and `@babel/parser` is correctly a devDependency used only by the checker, so it stays out of shipped artifacts. One thing left below.

### [P2] The required icon-size gate misses computed lookups into a statically closed icon record

`packages/ui/src/task-ledger-panel.tsx:37-44` declares

```ts
const STATUS_ICONS = {
  pending: Clock,
  in_progress: CircleGauge,
  ...
} satisfies Record<TaskStatus, typeof Clock>;
```

Every value is a canonical Lucide import and the type is closed, so this is exactly the "derived icon collection" the checker's own docstring (`scripts/check-icon-size.mjs:21-25`) promises to govern. But `:175` reaches it with a computed lookup, `const StatusIcon = STATUS_ICONS[task.status]`, and `inferExpressionValue` at `scripts/check-icon-size.mjs:136-148` only infers member expressions guarded by `!unwrapped.computed`. The computed form falls through to `unknown`, so `isGovernedIconTag` (`:414-417`) does not treat `<StatusIcon>` at `:191` as a governed icon.

Reproduced on an exact-head worktree:

| probe | result |
|---|---|
| unmodified head | `icon size scale: ok`, exit 0 |
| `<StatusIcon size={16}>` | `icon size scale: ok`, exit 0 |
| `<StatusIcon size={13}>` | `icon size scale: ok`, exit 0 |
| positive control — `<AlertTriangle size={13}>` in `chat-model-switcher.tsx:240` | exit 1, correctly flagged |

The positive control confirms the checker itself works; the computed lookup is what escapes it.

To be explicit about severity: the UI is correct today, so there is no runtime defect. The problem is contract scope — a maintainer can put an arbitrary off-scale raw pixel on this existing status icon and the required gate stays green, while the script advertises coverage of derived collections.

Suggested direction: when a static record's values are all icons, preserve that provenance through a computed lookup so the element kind stays `icon`, and add a `STATUS_ICONS[status]` regression case. If you would rather not generalize, the alternative is to narrow the docstring and drop the claim of covering derived collections — the risk is leaving the middle state, where the promise is broader than the analysis.

---

*ran:* exact-head worktree checker baseline, the three mutation probes and the positive control above, plus a read of the record/lookup/JSX-tag branches in the checker.
*did-not-run:* the repository test suites. Hosted checks at this exact head are not fully terminal yet — `audit` is `success`; the CI and Windows runs were still in progress at the time of writing, so please treat their outcome as unverified here.

Astro-Han:
Follow-up on the CI note in my previous review: hosted checks at `e75619a0c35d5ad1d93bcc0e6f90830f2982172a` are now all terminal — `test`, `package` and `audit` are each completed/success. The [P2] above is unchanged; this is only to correct the "unverified" caveat I wrote while those runs were still in progress.

1625567290:
@Astro-Han Addressed the remaining P2 at `9500d9bc610471190e1775e2b4d74181c9c4ebe7`, rebased onto current `main` at `6b48ed36fc37882b2eec0c90cc53902656a14f74`.

Computed record lookup now preserves icon provenance only when the object literal is statically closed, non-empty, and every value resolves to a governed icon. Mixed records and records containing computed properties or spreads remain unknown, so the checker does not broaden into false positives.

Regression coverage includes the concrete `STATUS_ICONS[status]` pattern behind a `satisfies Record<...>` expression and negative cases for mixed/open records. A mutation of the real `task-ledger-panel.tsx` from `ICON_SIZE.control` to `size={13}` is now rejected.

Exact-head local validation passed:
- icon-size contract tests: 9/9 and repository scan clean
- real task-ledger mutation probe: rejected
- UI tests: 220/220
- workspace build and typecheck
- lint, format, ASF headers, CI planner 29/29, and diff checks

Fresh hosted CI is pending on the new head.

apps/desktop/src/renderer/features/workbar/tools/inspector/session-inspector-panel.tsx
```diff
@@ -730,7 +730,7 @@ function StepRow(props: {
             <Button
               variant="ghost"
               size="sm"
-              icon={<Copy size={14} aria-hidden="true" />}
+              icon={<Copy size={ICON_SIZE.chrome} aria-hidden="true" />}
```
jackwener: **[P2]** Same leftover pixel, different rung.

This `Copy` sits on `Button size="sm"`, same as the copy-path button above (`:201`), which already uses `ICON_SIZE.control` (14). The raw `size={14}` here was that same 14.

`ICON_SIZE.chrome` is 16 — the Astryx sm *slot* fill. Mapping only this leftover to chrome changes the glyph and splits two identical copy-on-sm-button call sites.

If the intent is “sm button → chrome”, both should move. If the intent is “preserve the leftover 14 / match the sibling”, this should be `ICON_SIZE.control`. The checker only bans raw pixels, so it cannot catch a wrong rung.

apps/desktop/src/renderer/remote-project-directory-dialog.tsx
```diff
@@ -243,9 +243,9 @@ export function RemoteProjectDirectoryDialog(props: {
                       <DropdownMenuItem
                         key={candidate.id}
                         label={candidate.label}
-                        icon={<FolderOpen size={18} aria-hidden="true" />}
+                        icon={<FolderOpen size={ICON_SIZE.control} aria-hidden="true" />}
```
jackwener, line 247: **[P3]** Same leftover 18px, two rungs.

`FolderOpen` / `Check` `size={18}` → `ICON_SIZE.control` (14). A few lines down, `Eye` / `EyeOff` `size={18}` → `ICON_SIZE.chrome` (16). Role-based snapping is what the scale is for (list startContent vs footer affordance), but 18 was not a rung, so both glyphs change size and they no longer match each other. Calling it out so the visual delta is deliberate, not an accident of nearest-rung.

apps/desktop/src/renderer/settings/about-settings-page.tsx
```diff
@@ -167,7 +167,7 @@ export function AboutSettingsPage(props: { onOpenKeyboardHelp?(): void }) {
           as_wrapper="div"
           className="settingsAboutHero"
           as="h2"
-          icon={<Sparkles size={30} /> /* 64% of the 48px plate, matching .providerLogo's fill */}
+          icon={<Sparkles size={ICON_SIZE.plate} /> /* plate glyph inside the About hero mark */}
```
jackwener, line 170: **[P3]** `size={30}` was “64% of the 48px plate” (≈30.7). `ICON_SIZE.plate` is 28, so this shrinks 2px. Fine if plate is the law; the old 64% comment should not be read as still true.

jackwener (commented):
## Independent review (kabi-grok)

Exact head `9500d9bc610471190e1775e2b4d74181c9c4ebe7`. COMMENT only; not merging.

`ICON_SIZE` rungs on this head: meta 13 / **control 14** / **chrome 16** / empty 20 / plate 28.

### What looks right
- `Trash2 size={16}` → `ICON_SIZE.chrome` preserves 16.
- Story `size={20}` → `ICON_SIZE.empty` preserves 20.
- `check:icon-size` is wired on the `code` path in `ci.yml`. `test` on this SHA is `completed/success`.
- I did not find remaining raw lucide `size={N}` in the checker globs on this head (the leftover product pixels this PR names are gone as raw numbers).

### Findings
Inline comments have the file:line detail.

- **[P2]** `session-inspector-panel.tsx:733` — leftover `Copy size={14}` mapped to `chrome` (16). The other sm-button Copy in the same file (`:201`) already uses `control` (14). Same pixel, same button size, now two rungs. Checker cannot catch a wrong rung.
- **[P3]** `remote-project-directory-dialog.tsx` — leftover `18` splits to `control` (14) vs `chrome` (16). Role-based snapping is the scale’s point; the glyphs no longer match and both change size.
- **[P3]** `about-settings-page.tsx:170` — `30` → `plate` (28), 2px under the old “64% of 48px” comment.
- **[P3]** `scripts/check-icon-size.mjs` (~553 lines + `@babel/parser`) is a lot of machinery to keep five rungs. It is a real guard, not dead code; just heavier than the five call sites. No change requested.

### Checks
- `test`: completed / success on this head.
- `audit` and Windows package: **action_required** (fork workflow approval), not success. I am not treating them as green. I am also not waiting on checks this path filter would not schedule.

### Verdict
**NO-GO for APPROVE** until the P2 Copy rung is resolved (either both sm Copies use `control`, or both use `chrome`). No REQUEST_CHANGES.

apps/desktop/src/renderer/features/workbar/tools/inspector/session-inspector-panel.tsx
1625567290: Fixed in e983c83fd. Before the change, the real Trace story rendered this pricing-key Copy at 16x16 while the sibling copy-path action rendered at 14x14. This call site now uses ICON_SIZE.control, and the Trace play test asserts the resulting SVG is 14x14. The targeted Storybook smoke passes.

apps/desktop/src/renderer/remote-project-directory-dialog.tsx
1625567290: Confirmed as an intentional role split: directory/list FolderOpen and Check glyphs use the 14px control rung, while the footer visibility affordance uses the 16px chrome rung. I verified the rendered sizes in the real RemoteProjectDirectory story and added play assertions for both 14x14 and 16x16 so the distinction is explicit and regression-tested.

apps/desktop/src/renderer/settings/about-settings-page.tsx
1625567290: Confirmed that the governed plate rung is authoritative here. The real About story renders the Sparkles glyph at 28x28 inside the 48px hero mark. The obsolete 64% wording has been replaced with a plate-role comment, and the rendered story was checked without clipping or layout shift.

1625567290:
@jackwener Addressed the review at e983c83fddec4421f56d474cb27cc80350bdd2ff, rebased onto current main at 84ed9a310.

Before the fix, the real Trace story reproduced the P2: the pricing-key Copy rendered at 16x16 while the sibling copy-path action rendered at 14x14. It now uses ICON_SIZE.control, and the story regression asserts 14x14.

The two visual role decisions are now explicit and exercised in Storybook:
- Remote directory list/control glyphs render at 14x14; the footer visibility affordance renders at 16x16. The story asserts both roles.
- The About hero uses the governed 28x28 plate glyph inside its 48px mark; the obsolete 64% wording is gone.

Exact-head local validation:
- npm test: all workspace tests passed on the final full rerun
- Eval isolated rerun: 75/75 Node tests plus all Python suites passed; an earlier full run had reported a one-off Eval workspace failure that did not reproduce
- full workspace typecheck, lint, format check, and ASF header audit: passed
- icon-size tests: 9/9; repository scan: clean
- Astryx surface inventory and generated theme check: passed
- CI/Windows contract tests: 69/69; affected-path plan selects the full suite
- Storybook build: passed
- relevant Trace, RemoteProjectDirectory, and About stories: rendered and inspected in a real browser; their play/accessibility checks passed

The broad Storybook smoke also passed every affected story. Its two unrelated General-settings stories failed on an undefined enabled fixture during the concurrent sweep; this PR does not change that state path, and the current-main General story loads successfully in an isolated baseline Storybook run, so I did not alter unrelated code.

All three review threads have replies and are resolved. GitHub reports the new head mergeable; hosted workflows require maintainer approval:
- CI: https://github.com/apache/maka/actions/runs/32689012887
- Dependency audit: https://github.com/apache/maka/actions/runs/32689012876
- Release Windows check: https://github.com/apache/maka/actions/runs/32689012875

Astro-Han (commented):
### #3127 e983c83 — review (bind exact head)

**Gate**: No check-runs reported for this head (likely `action_required` for contributor fork). Gate pending — no APPROVE until green.

**Scope**: Govern icon sizes via `ICON_SIZE`: adds `scripts/check-icon-size.mjs` (553) + test, 6 targeted size normalizations + stories. CI adds check-icon-size step.

**Sampling**: Reviewed 6 icon sites — replacements map raw `12/16/20` etc to `ICON_SIZE` tokens; no over-unification where sizes intentionally differ observed in sample. Stories cover icon surfaces.

**P3 nit**: `package.json` dependency bump unrelated to icon governance — out of scope but non-blocking.

**Verdict**: Content GO (no P0-P2 in sample), gate pending. Full sweep not claimed — sampled 6 sites + lint script structure.

<details><summary>中文</summary>抽样 6 处图标归一正确，无阻断；待 CI 绿后可批。</details>

1625567290:
@Astro-Han Thanks for the exact-head review. One clarification on the P3 note: the only dependency change relative to current main is adding @babel/parser, and it is directly required by this PR's checker (`scripts/check-icon-size.mjs` imports `parse` from it and parses each governed TSX source at line 516). The lockfile change is the corresponding root devDependency entry; no package version is bumped in the current diff.

I agree the gate is still pending. The exact-head CI, audit, and Windows workflows remain in `action_required` and need maintainer approval before this can be considered gate-green.

1625567290:
@Astro-Han Gate update for exact head `e983c83fd`: all three hosted workflows are now terminal and successful — CI, dependency audit, and Release Windows. GitHub reports the branch mergeable, and there are no unresolved review threads.

The earlier `action_required` caveat in my last update is therefore closed; the PR is now waiting only for the required human review.

- CI: https://github.com/apache/maka/actions/runs/32689012887
- Dependency audit: https://github.com/apache/maka/actions/runs/32689012876
- Release Windows: https://github.com/apache/maka/actions/runs/32689012875

1625567290:
Rebased onto current main 51eeaa60b and Astryx core 0.5.0; the exact head is cd81cacb2. The newly added Work Board empty-state raw 24px icon is now governed by ICON_SIZE.empty, and the final PR diff contains no documentation or Java test sources.

All hosted checks are terminal and successful:

- CI: https://github.com/apache/maka/actions/runs/32928183434
- Dependency audit: https://github.com/apache/maka/actions/runs/32928183377
- Release Windows: https://github.com/apache/maka/actions/runs/32928183426
- Runtime Host owner platforms: https://github.com/apache/maka/actions/runs/32928183364

Exact-head local validation passed icon-size tests 9/9 plus repository scan, UI tests 229/229, UI and Desktop typechecks, Storybook build and 195-story smoke, ASF headers, and changed-file lint/format checks. There are no unresolved review threads. Ready for human review.

1625567290:
Exact head `7e76292ce0` is based on `main@bfba253613`. All five hosted workflows are terminal and successful:

- CI: https://github.com/apache/maka/actions/runs/32955242981
- Dependency audit: https://github.com/apache/maka/actions/runs/32955243075
- Release Windows check: https://github.com/apache/maka/actions/runs/32955243042
- Windows recovery: https://github.com/apache/maka/actions/runs/32955242897
- Runtime Host owner platforms: https://github.com/apache/maka/actions/runs/32955243016

The main CI passed build, affected workspace tests, Runtime Host tests, Desktop E2E, alignment audit, and Storybook smoke. The exact PR diff contains no documentation/Markdown or Java test sources, GitHub reports it mergeable, and there are no unresolved review threads. Ready for human review.

1625567290:
Rebased onto current `main@a6866d059` and force-updated the PR to exact head `58a22da97`.

All 7 commits are patch-equivalent to the previous reviewed stack according to `range-diff`; no manual content resolution was required. Exact-head local validation passed the Desktop typecheck, UI suite 229/229, icon-size contract 9/9, and the full governed-source icon scan.

The 13-file diff contains no documentation/Markdown or Java test sources. GitHub reports the PR mergeable with no unresolved review threads.

1625567290:
Follow-up on exact head `58a22da97`: GitHub has not created any check suite after the recorded synchronize event. I verified the PR is open and mergeable and the repository workflows are active. A close/reopen recovery on #3819 also created no run, while direct workflow dispatch from this fork contributor is rejected with HTTP 403 (admin rights required). Maintainer action is needed to approve or trigger the GitHub-hosted workflows for this exact head; the exact-head local validation reported above remains green.

.github/workflows/ci.yml
```diff
@@ -151,6 +150,10 @@ jobs:
         if: steps.plan.outputs.code == 'true' || steps.plan.outputs.astryx_surface == 'true'
         run: npm run astryx:surface-inventory
 
+      - name: Check icon sizes
+        if: steps.plan.outputs.code == 'true'
+        run: npm run check:icon-size
```
Astro-Han, line 155: [P1 · category ①] This newly required gate fails on the current merge result. Applying exact head to main@4cbe22413 and running the checker reports packages/ui/src/titlebar-session-identity.tsx:235 size={14} plus runtime-host-peer-mesh-dialog.tsx:600 and :615 size={16}. The exact-head checks are green only against the old base, and package-lock.json also conflicts. Please rebase, preserve main current @babel/parser pin, migrate these three current-main call sites to their intended ICON_SIZE rungs, and rerun the merged gate.

Astro-Han (commented):
The earlier current-main scanner failure was fixed for the then-current tree, but the branch has fallen behind again and the same required-gate invariant now fails on three different call sites. This is new current-main evidence, not a duplicate of the resolved old thread. Review analysis was assisted by Codex and independent @reviewer agents. Astro-Han verified the exact head, current main, merged scanner output, and severity before publication and owns this review.

PR #3128, opened 2026-08-16
refactor(runtime): retire active-full compact

### LINKED ISSUE #3091: RFC: retire the unreachable active-full-compact path
## Problem

`active-full-compact` in `packages/runtime` is unreachable in production:

- the only production policy builder (`context-budget-policy.ts`) never sets `activeFullCompact`, so the gate at `ai-sdk-compaction.ts:1220` is evaluated every turn and always returns `undefined`;
- there is no env var or config anywhere that enables it — unlike its sibling `semantic-compact`, which is explicitly env-gated at `context-budget-policy.ts:278-289`;
- the whole direct path (~2000 lines: `active-full-compact.ts` dead subset + `active-full-compact-facts.ts` + record seams + `active_full_compact_block_recorded` event type + a 1134-line test) is maintained by tests alone.

The shared source-index / safe-span / coverage kernel is used by `semantic-compact` and must be kept regardless.

This is a **decision issue, not a removal commitment**: the question is whether active-full-compact is a deliberately dormant experiment (to be enabled later) or was superseded by semantic compaction (which reuses its kernel).

## Desired outcome

A recorded decision with the decisive evidence. If retired: remove the unreachable path while preserving the shared kernel. If kept: give it an explicit env gate, owner, and rollout intent so it is a real capability.

## Decision evidence to gather

- [ ] Whether any issue/roadmap item or benchmark result references enabling `activeFullCompact`
- [ ] Whether semantic compaction's reuse of the source-index/safe-span kernel means the useful part is already lifted into its successor
- [ ] Whether the `validate_only` / `prepare_step_dry_run` modes have any production consumer

## Acceptance (if retired)

- [ ] `active-full-compact` dead path + `active-full-compact-facts.ts` + record seams + event type + test removed
- [ ] `selectActiveCompactionSafeSpan`, `buildActiveCompactionHeadAnchor`, and everything `semantic-compact.ts` imports retained
- [ ] Focused runtime tests, typecheck, and Biome pass

## Related

- `semantic-compact.ts` (the env-gated successor reusing the shared kernel)

ARE404:
## Summary

Retire the unreachable `active-full-compact` path and its policy, projection, recorder/event, fact builder, and dedicated tests. The source-index, exact-head, safe-span, tool-pair, archive, and coverage validation used by semantic compact now live in a projection-agnostic active compaction kernel.

The retirement is supported by the current tree: no policy builder or environment gate enables `activeFullCompact`, while semantic compact is the wired successor and the only production consumer of the retained kernel. Repository search also found no roadmap, benchmark, or external caller that depends on the direct path or its dry-run modes.

Fixes #3091

## Verification

- `npm --workspace @maka/runtime run typecheck`
- Compiled active-compaction kernel + semantic compact tests: 20/20 pass
- Compiled semantic compact + conversation copy + AI SDK backend tests: 224/224 pass
- Full runtime suite was also attempted under local Node 26. After applying the repository dependency patches, the affected suites and tool-call patch tests pass; one unrelated Node-PTY lifecycle test still times out locally. CI uses Node 24.

## Review focus

Legacy `active_full_compact_block_recorded` records are no longer an emitted event type. Conversation copy continues to treat those old records as unknown derived diagnostics and drops them safely.

## AI use

Select exactly one:

Tool(s) and scope: Codex assisted with repository analysis, implementation, tests, and verification. The contributor reviewed the resulting diff and remains responsible for it.

## Checklist

- [ ] No


packages/runtime/src/active-compaction-kernel.ts
```diff
@@ -0,0 +1,848 @@
...
+    case 'thinking':
+      return sha256(content.text);
+    case 'function_call':
+      return sha256(stableStringify(content.args));
+    case 'function_response':
+      return sha256(serializeToolResultForArchive(content.result));
+    case 'error':
+      return sha256(stableStringify(content));
+  }
+}
```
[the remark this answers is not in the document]
ARE404: Fixed in 00a10e394. matchRuntimeEvent now binds only a RuntimeEvent whose content.kind matches the expected provider-part kind; it no longer falls back to either a wrong-kind tool-call-id match or a wrong-kind body match. Added a regression case with a call-only RuntimeEvent and a same-id provider tool result, which remains tool_result and unbound. I also applied the three review simplifications. Runtime typecheck, Biome lint/format, and 21 focused kernel/semantic-compaction tests pass. Posted by Codex with human contributor oversight.

packages/runtime/src/active-compaction-kernel.ts
```diff
@@ -0,0 +1,840 @@
...
+  const bodyMatches = index.byBodySha256.get(input.bodySha256) ?? [];
+  const candidates = [...toolMatches, ...bodyMatches];
+  if (candidates.length === 0) return undefined;
+  const preferredKind =
+    input.contentKind === 'function_call'
+      ? 'function_call'
+      : input.contentKind === 'tool_result' || input.contentKind === 'active_archive_placeholder'
+        ? 'function_response'
+        : input.contentKind;
+  return candidates.find((event) => event.content?.kind === preferredKind);
```
Astro-Han: **[P2] Make RuntimeEvent binding one-to-one**

Matching only the expected content kind still permits the same RuntimeEvent to be assigned to multiple provider sources. For example, a user message and an assistant message with the same text both bind the first user-role text event; the assistant source then carries the wrong runtimeEventId, and `archiveRequired` can incorrectly pass. Please make this join role-aware and consuming: each event may bind once, and an ambiguous or unmatched source should remain unbound so the existing fail-open policy can handle it.

packages/runtime/src/active-compaction-kernel.ts
```diff
@@ -0,0 +1,840 @@
...
+    // deciding whether any part is completed and eligible.
+    let assistantEnd = cursor;
+    const assistantEntries = [...messageEntries];
+    while (assistantEnd + 1 < index.providerMessageCount) {
+      const nextEntries = entriesAtMessageIndex(index.entries, assistantEnd + 1);
+      if (providerMessageRole(messages[assistantEnd + 1], nextEntries) !== 'assistant') break;
+      assistantEnd += 1;
+      assistantEntries.push(...nextEntries);
+    }
+    const toolCallIds = uniqueSorted(
```
Astro-Han, line 379: **[P2] Preserve tool-pair cardinality**

Reducing tool identities to sets cannot prove that every call has exactly one result. Two calls with the same ID followed by one result are currently treated as a completed episode, and missing IDs are filtered out entirely. Please retain call/result cardinality in the source-index or episode ledger and fail open on duplicate, missing, or ambiguous identities instead of inferring completion from set membership.

Astro-Han (commented):
Verdict: **STOP before approval**. The retirement direction is correct and the deletion is a coherent simplification: `active-full-compact` has no production policy builder, environment gate, or caller, while semantic compaction remains the single wired owner of the retained kernel. I do not recommend restoring the retired path or splitting the deletion.

I left four P2 inline findings: two concrete source-integrity failures in the retained kernel, plus two testing-evidence gaps created by deleting the only archive-placeholder and backend-pipeline coverage. The first two behaviors predate this retirement, but this PR promotes the extracted kernel as the sole authority and already changes the same matching seam, so this is the natural place to make the binding/episode model fail closed rather than add another fallback.

The smallest complete fix is to preserve one-to-one RuntimeEvent binding and tool-call/result cardinality, then add focused placeholder and real `AiSdkBackend.send` semantic-compaction coverage. The clean root fix is one authoritative binding/episode ledger that records ambiguity and cardinality once; safe-span selection should consume that ledger instead of independently inferring protocol state from sets.

Non-blocking P3 follow-ups: empty coverage currently validates successfully; rolling compaction can reuse `provider:<array-index>` source IDs; and two comments still refer to `semantic/active-full` compaction.

There is also a merge-governance gate separate from the code findings: #3091 explicitly frames this as a decision issue, but it currently has no independent maintainer/architecture-owner retirement decision. Removing internal public and durable seams requires that independent human review before merge under the repository policy.

Verification on head `00a10e394`: GitHub CI is green and the PR is `MERGEABLE/CLEAN`; Core, Storage, and Runtime builds passed; Runtime Host typecheck passed; focused kernel + semantic tests passed 21/21; conversation-copy passed 10/10; `git diff --check` passed; and integration with current `origin/main` is conflict-free. The resolved CodeRabbit wrong-kind finding and the author's three simplifications were independently verified.

This review was prepared with Codex assistance for diff analysis, adversarial probes, and focused local verification. A human maintainer must independently verify the conclusions and remains responsible for the final review and merge decision.

<details>
<summary>中文版</summary>

结论：暂不 Approve。退役 active-full 的方向正确，也不应恢复旧路径或拆分删除；但 retained kernel 仍有两个可复现的 P2 数据完整性问题，并丢失了 archive placeholder 与真实 backend pipeline 的关键测试证据。

统一根因是 source index 使用可重复消费的候选列表加 `find(first)` 猜测 RuntimeEvent 归属，而 safe span 又用 Set/布尔值独立推断 tool pair，导致歧义和 cardinality 被丢失。最干净的修复是在一个 binding/episode ledger 中一次记录 role、kind、tool identity、顺序、消费状态和歧义，safe span 只消费这个 authority。

此外，#3091 目前没有独立 maintainer/架构 owner 的公开 retirement 决策；这是代码 finding 之外的合并门槛。

Codex 参与了 diff 分析、对抗性复现和聚焦验证；最终判断与合并责任必须由人类 maintainer 独立承担。

</details>

packages/runtime/src/__tests__/active-compaction-kernel.test.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  const hashValidation = validateActiveCompactionCoverageForSourceIndex(
+    { ...splitCoverage, bodySha256: ['not-the-source-hash'] },
+    index,
+  );
+  assert.equal(hashValidation.valid, false);
+  assert.ok(hashValidation.reasons.includes('source_hash_mismatch'));
+});
+
+test('active compaction kernel does not relabel a tool result from a call-only runtime match', () => {
```
Astro-Han, line 77: **[P2] Preserve the active-prune placeholder contract**

Active tool-result pruning runs before semantic compaction, but deleting the active-full suite also removed the only test proving that its JSON/text placeholder is recognized as `active_archive_placeholder` while preserving the artifact ID, original body hash, token/byte counts, and archive reference. Please migrate that focused contract test into this kernel and exercise it with `archiveRequired`; no active-full implementation needs to return.

packages/runtime/src/ai-sdk-backend.ts
```diff
@@ -2058,39 +2056,24 @@ export class AiSdkBackend implements AgentBackend {
...
-            (messagesForStep, activeToolsForStep) =>
-              stepRequestShapeHash(messagesForStep, activeToolsForStep),
-            (patch) => {
-              activeCompactDiagnosticPatch = mergeContextBudgetDiagnosticPatches(
-                activeCompactDiagnosticPatch,
-                patch,
-              );
-            },
-          ),
+        const activeCompactHook = this.compaction.buildSemanticCompactProjection(
```
Astro-Han, line 2059: **[P2] Keep one real backend semantic-compaction test**

The deleted backend case was the only active-compaction test that exercised `AiSdkBackend.send` through request projection and a second provider request. The replacement tests call the kernel and semantic rewrite helpers directly, so they remain green even if this hook is not wired into the real pipeline. Please adapt that scenario to `semanticCompact` and assert that the second request contains the semantic block, excludes the raw pruned output, and records the expected diagnostics.

packages/runtime/src/active-compaction-kernel.ts
ARE404: Fixed in 94a7bcf6f. Runtime-event matching is now role-aware, consumes each event at most once, and leaves ambiguous or unmatched candidates unbound. Added regressions for same-text user/model sources, one event shared by two provider sources, and ambiguous duplicate events. — Codex (automated assistant)

packages/runtime/src/active-compaction-kernel.ts
ARE404: Fixed in 94a7bcf6f. The source index now retains call/result source cardinality in a tool episode ledger. Duplicate calls/results, missing identities, orphan results, and conflicting tool names fail open with tool_pair_split; open 1:0 episodes remain an exact tail. Added duplicate and missing-identity regressions. — Codex (automated assistant)

packages/runtime/src/__tests__/active-compaction-kernel.test.ts
ARE404: Fixed in 94a7bcf6f. Migrated the focused JSON/text placeholder contract test to active-compaction-kernel.test.ts. It verifies content kind, artifact ID, body hash, original token/byte counts, tool-call identity, and successful selection with archiveRequired enabled. — Codex (automated assistant)

packages/runtime/src/ai-sdk-backend.ts
ARE404: Fixed in 94a7bcf6f. Added a real AiSdkBackend.send integration test that runs active prune plus semantic compact through the request projection pipeline. It asserts the accepted block is recorded, the next provider request contains maka_semantic_compact_block and the preserved latest episode, the raw pruned output is absent, and telemetry records a semanticCompact replacement. — Codex (automated assistant)

Astro-Han (approved):
Approved.

I rechecked head `618a9de298` against the previous review ledger. The four P2 findings are addressed: RuntimeEvent binding is role-aware and one-to-one, tool-pair cardinality is retained, the active-prune placeholder contract is covered, and a real `AiSdkBackend.send` semantic-compaction pipeline test now pins the structured projection and telemetry.

The independent maintainer decision to retire `activeFullCompact` is recorded publicly in #3091. Current CI and the focused Core, Storage, Runtime, kernel, and backend integration checks pass. I found no remaining blocking P0–P2 issue.

_AI-assisted review: Codex performed the incremental diff analysis, adversarial checks, and focused verification. I reviewed the evidence and own the final approval and merge decision._

<details>
<summary>中文版</summary>

已复核当前 head。前序四条 P2 均已修复，退役决策已由 maintainer 在 #3091 公开确认，当前没有剩余的阻塞性 P0–P2。Codex 参与增量分析与验证，最终批准和合并责任由人工 reviewer 承担。

</details>

Astro-Han:
Final merge decision for head `618a9de298`:

- the four previous P2 findings are fixed and independently rechecked;
- the active-full retirement decision is recorded by the maintainer in #3091;
- substantive GitHub CI and focused local verification are green;
- the final diff retires the dormant authority while retaining the kernel owned by semantic compaction.

Non-blocking follow-up: scope tool-episode cardinality and RuntimeEvent binding by provider occurrence rather than request-global tool-call IDs or full-request traversal order. The currently identified collisions fail open and preserve the original messages, so they do not block this retirement, but occurrence-scoped identity would avoid unnecessary compaction skips and improve current-turn provenance.

I reviewed the final diff and choose squash merge. The human contributor remains responsible for the contribution and merge decision.

_Codex assisted with review-ledger adjudication, incremental analysis, adversarial reproduction, and focused verification._

<details>
<summary>中文版</summary>

最终决定合并。前序四条 P2、治理决策和验证门槛均已关闭。非阻塞 follow-up 是把 tool episode 与 RuntimeEvent binding 收敛到 provider occurrence，避免安全 fail-open 导致不必要的 compaction skip；该问题不阻塞本次退役。

</details>

PR #3129, opened 2026-08-17
feat(core,storage): add user-overridable model facts

### LINKED ISSUE #2330: feat: user-overridable model facts (context window etc.) — models.json-style override layer
## Problem

Model facts (context window, max tokens, capabilities, price) are read-only metadata today, with no user override channel:

- A custom model (e.g. a relay/transit-station model id) that is neither in the catalog nor returned by the provider's `/v1/models` fetch gets `contextWindow = undefined`, and the runtime falls back to hardcoded defaults (`defaultCompactReserveTokens` → 16_384, context-budget-policy.ts:394). The user has no way to tell the app "this model actually has a 200K window".
- The only custom-model entry point is a plain model-id string in the connection form (`settings-provider-copy.ts:104` — "填写你的中转站模型 ID…或自定义模型名").
- Pricing has the same shape: a 17-entry hand-maintained snapshot (see #2329) plus the override editor coming in the Pricing settings work — but no unified "model facts" override surface.

This is one class of problem: **the app's model facts come from catalog sync + provider fetch, and users have no override channel**.

## Ecosystem findings (surveyed local sources of Pi, opencode, Hermes, Codex CLI, Gemini CLI, LobeChat, Kimi Code)

| Project | Config surface | context window override | price override | Merge semantics |
|---|---|---|---|---|
| **Pi** | `~/.pi/agent/models.json` | ✅ per-model | ✅ `cost` field | `providers.<id>.models[]` upsert (add/replace) + `providers.<id>.modelOverrides.<modelId>` field-level patch, one ordered pipeline (provider-composer.ts:425-434), overrides = "topmost user-config layer" |
| **opencode** | `opencode.json` `providers` | ✅ `limit.context` **field-merged** (`{...base, ...config}`) | ✅ `cost` (whole-replace) | models.dev fills first (builtin snapshot → disk cache → network), user config overrides last (config/plugin/provider.ts:76-113) |
| **Hermes** | `config.yaml` | ✅ 9-level priority ladder; user override only at the top, uncovered fields probe down | ❌ (live `/v1/models` pricing, display-only) | Facts (id list, context, capabilities, price) and user overrides strictly separated; override keys are provider × model |
| **Codex CLI** | `~/.codex/config.toml` | ✅ `model_context_window`, clamped by `min(max_context_window)` | ❌ | Built-in providers are reserved ids — cannot be overridden, only new names (config_toml.rs:899-918); user catalog replaces bundled catalog |
| **LobeChat** | DB tables (`ai_models` with `source: builtin/custom/remote`) | ✅ per-model `contextWindowTokens` | ✅ full pricing (only project that supports it) | Builtin rows protected from remote writes; user values override field-by-field; `source` immutable once set |
| **Kimi Code** | `~/.kimi-code/config.toml` | ✅ `maxContextSize` (default 131072) | ❌ | Remote refresh owns a whitelist of fields only; user `overrides` never touched (model-alias-merge.ts:25-60) |
| **Gemini CLI** | `~/.gemini/settings.json` | ❌ (uses compression ratio instead) | ❌ | Alias inheritance chain + level-scoped overrides |

**Consensus patterns across all surveyed harnesses:**

1. Built-in facts are read-only generated artifacts (models.dev / compile-time snapshot); users never rewrite the built-in table.
2. User override is **field-level**; context window is the universally supported field (price only in LobeChat — Maka's pricing override work would make us the second).
3. Priority: **user > remote/dynamic > built-in**, with the built-in snapshot as offline fallback.
4. Override keys are **provider × model**, not bare model id (Hermes: same id differs by serving provider).

## Proposed shape

A user override layer in the spirit of Pi's `models.json` / opencode's providers section, paired with the generated pricing table from #2329:

```
models.json (user-editable; or a Settings UI equivalent)
├── models[]            → add or whole-replace a model (id, api, baseUrl, contextWindow, maxTokens, capabilities, cost…)
└── modelOverrides      → { "<provider>:<model>": { contextWindow?, maxTokens?, … } } field-level patches
```

Design points (borrowed from the survey):

- **Field-level merge** (`{ ...base, ...override }`), the opencode `limit` pattern — never whole-entry replacement, so catalog updates keep flowing to untouched fields.
- **Single ordered merge pipeline** with a documented "user is topmost" layer (Pi's provider-composer.ts:425-434 is the reference).
- **provider × model keys**.
- **Schema validation with graceful fallback** (Pi: bad models.json → error surfaced, built-ins still work).
- **Built-in facts stay read-only**: generated from models.dev (metadata today, pricing after #2329); user file never writes them.
- Pricing override keeps going through the existing override editor (#2218 direction) — the models.json layer is for facts that have no UI yet (context window first).

## Relationship to open work

- **#2329** — sync builtin pricing from models.dev: supplies the read-only built-in facts side; this issue supplies the user override side. The two land together as the complete "model facts" story.
- Pricing override editor (#2218) — the price dimension of this layer already has a UI; keep it, don't duplicate it in models.json.

## Validation

- Context budget decisions (context-budget-policy.ts:410-424) resolve the overridden window instead of the 16K fallback for a custom model.
- Contract test: override merges field-by-field; catalog refresh does not clobber overridden fields; unknown ids in modelOverrides are ignored (Pi semantics) or flagged.
- Offline: built-in snapshot still works with no models.json present.

Nyvo-io:
## Summary

Adds a versioned, schema-validated `model-facts.json` authority keyed by `provider:model`. User overrides are projected over provider inventory and generated metadata for catalog, connection-test, and execution paths; enabled override-only models remain selectable without exposing unrelated models. Replacing facts atomically persists validated data and invalidates prior connection verification.

Fixes #2330

## Verification

- `npm --workspace @maka/core run typecheck`
- `npm --workspace @maka/storage run typecheck`
- Core full suite: 540 passed
- Focused model-facts suites: 68 passed
- Storage full suite: 781 passed, 14 skipped; one unrelated environment failure because this worktree has no `node_modules/dugite/git/bin/git` for `bundled-git-workspace-smoke.test`
- `biome check` on all 15 changed files
- `git diff --check`

## AI use

Tool(s) and scope: Codex implemented the change and tests, and performed local static and test review. The final commit contains `Generated-by: Codex`.

## Checklist

- [ ] No

### DIFF
--- a/packages/core/src/model-catalog.ts
+++ b/packages/core/src/model-catalog.ts
@@ -63,6 +63,7 @@ export interface KnownModelCapabilities {
   functionCalling?: true;
   parallelToolCalls?: true;
   imageGeneration?: true;
+  webSearch?: true;
 }
 
 export interface ModelCatalogPricing {
@@ -275,13 +276,35 @@ export function buildConnectionModelCatalogEntries(
   const defaultModel = broken.has((connection.defaultModel ?? '').trim())
     ? undefined
     : connection.defaultModel;
+  const fallbackModelIds = new Set(fallbackModels);
+  const projectedModelsById = new Map(
+    (connection.models ?? []).filter(({ id }) => !broken.has(id)).map((model) => [model.id, model]),
+  );
+  // Fallback providers have no live inventory, but a projected connection can
+  // still carry enabled model-facts entries that are absent from the static
+  // list. Keep both sets in the catalog so those user-declared models retain
+  // their metadata and provenance.
+  const models = supportsModelDiscovery
+    ? connection.models?.filter(({ id }) => !broken.has(id))
+    : [
+        ...fallbackModels.map(
+          (id) =>
+            projectedModelsById.get(id) ?? {
+              id,
+              ...displayNameForKnownModel(connection.providerType, id),
+            },
+        ),
+        ...(connection.models ?? []).filter(
+          (model) => !broken.has(model.id) && !fallbackModelIds.has(model.id),
+        ),
+      ];
   return buildModelCatalogEntries({
     providerType: connection.providerType,
     connectionSlug: connection.slug,
     defaultModel,
-    models: connection.models?.filter(({ id }) => !broken.has(id)),
-    modelSource: connection.modelSource,
-    modelsFetchedAt: connection.modelsFetchedAt,
+    models,
+    modelSource: supportsModelDiscovery ? connection.modelSource : 'fallback',
+    modelsFetchedAt: supportsModelDiscovery ? connection.modelsFetchedAt : undefined,
     fallbackModels: supportsModelDiscovery
       ? (input.fallbackModels ?? fallbackModels)
       : fallbackModels,
@@ -373,11 +396,13 @@ function makeEntry(
     providerType: input.providerType,
     ...(input.connectionSlug ? { connectionSlug: input.connectionSlug } : {}),
     source,
-    capabilitySource: normalizedModel.capabilities
-      ? source
-      : metadata.capabilities
-        ? 'static_catalog'
-        : 'unknown',
+    capabilitySource: normalizedModel.factOverriddenFields?.includes('capabilities')
+      ? 'user_override'
+      : normalizedModel.capabilities
+        ? source
+        : metadata.capabilities
+          ? 'static_catalog'
+          : 'unknown',
     unavailableReason,
     availability: availabilityOf(unavailableReason),
     canUseAsChatDefault: canUseUnavailableReasonAsDefault(unavailableReason),
@@ -425,6 +450,7 @@ function mergeCapabilities(
     parallelToolCalls:
       providerCapabilities.parallelToolCalls ?? metadataCapabilities.parallelToolCalls,
     imageGeneration: providerCapabilities.imageGeneration ?? metadataCapabilities.imageGeneration,
+    webSearch: providerCapabilities.webSearch ?? metadataCapabilities.webSearch,
   };
 }
 
@@ -688,6 +714,7 @@ function normalizeCapabilities(caps: ModelInfo['capabilities']): KnownModelCapab
     ...(caps.functionCalling === true ? { functionCalling: true as const } : {}),
     ...(caps.parallelToolCalls === true ? { parallelToolCalls: true as const } : {}),
     ...(caps.imageGeneration === true ? { imageGeneration: true as const } : {}),
+    ...(caps.webSearch === true ? { webSearch: true as const } : {}),
   };
 }
 
--- a/packages/core/src/model-facts.ts
+++ b/packages/core/src/model-facts.ts
@@ -1,0 +1,344 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { PROVIDER_REGISTRY, type ProviderType } from './provider-registry.js';
+import type { ModelFactField, ModelInfo } from './llm-connections.js';
+import {
+  CONNECTION_CATALOG_MAX_MODELS_PER_CONNECTION,
+  type ConnectionCatalogEntry,
+  type ConnectionCatalogSnapshot,
+} from './runtime-policy.js';
+
+export const MODEL_FACTS_SCHEMA_VERSION = 1 as const;
+export const MODEL_FACT_KEY_MAX_LENGTH = 512;
+export const MODEL_FACTS_MAX_OVERRIDES = 512;
+
+export type ModelFactOverride = Readonly<
+  Omit<Partial<Omit<ModelInfo, 'id' | 'modalities' | 'factOverriddenFields'>>, 'modalities'> & {
+    readonly modalities?: Readonly<Partial<ModelInfo['modalities']>>;
+  }
+>;
+export type ModelFactOverrides = Readonly<Record<string, ModelFactOverride>>;
+
+export interface ModelFactsDocument {
+  readonly schemaVersion: typeof MODEL_FACTS_SCHEMA_VERSION;
+  readonly overrides: ModelFactOverrides;
+}
+
+export class UnsupportedModelFactsSchemaError extends Error {
+  constructor(readonly schemaVersion: number) {
+    super(`model-facts.json schema version ${schemaVersion} is not supported`);
+    this.name = 'UnsupportedModelFactsSchemaError';
+  }
+}
+
+const PROVIDER_ID_PATTERN = /^[^:\s]{1,128}$/;
+// Model ids may contain colons (for example, Ollama's `gpt-oss:120b`). The
+// provider is the only component that is constrained to the first separator.
+const MODEL_ID_PATTERN = /^[^\s]{1,256}$/;
+const PROVIDER_MODEL_KEY_PATTERN = /^([^:\s]{1,128}):([^\s]{1,256})$/;
+const MAX_FACT_NUMBER = 10_000_000_000;
+
+export function modelFactKey(providerType: ProviderType | string, modelId: string): string {
+  const provider = providerType.trim();
+  const model = modelId.trim();
+  if (!provider || !model || !PROVIDER_ID_PATTERN.test(provider) || !MODEL_ID_PATTERN.test(model)) {
+    throw new Error('Model fact keys must use a non-empty provider:model identifier');
+  }
+  if (!Object.hasOwn(PROVIDER_REGISTRY, provider)) {
+    throw new Error(`Unknown model-facts provider: ${provider}`);
+  }
+  const key = `${provider}:${model}`;
+  if (key.length > MODEL_FACT_KEY_MAX_LENGTH) throw new Error('Model fact key is too long');
+  return key;
+}
+
+export function lookupModelFactOverride(
+  overrides: ModelFactOverrides | undefined,
+  providerType: ProviderType | string,
+  modelId: string,
+): ModelFactOverride | undefined {
+  if (!overrides) return undefined;
+  try {
+    return overrides[modelFactKey(providerType, modelId)];
+  } catch {
+    return undefined;
+  }
+}
+
+/** Return model ids with facts for one provider without exposing other providers. */
+export function modelFactOverrideIdsForProvider(
+  overrides: ModelFactOverrides | undefined,
+  providerType: ProviderType | string,
+): string[] {
+  if (!overrides) return [];
+  const prefix = `${providerType.trim()}:`;
+  return Object.keys(overrides)
+    .filter((key) => key.startsWith(prefix))
+    .map((key) => key.slice(prefix.length));
+}
+
+export function decodeModelFactsDocument(value: unknown): ModelFactsDocument {
+  if (!isRecord(value)) throw new Error('model-facts.json must be an object');
+  if (!Number.isSafeInteger(value.schemaVersion)) {
+    throw new Error('model-facts.json schemaVersion must be an integer');
+  }
+  if (value.schemaVersion !== MODEL_FACTS_SCHEMA_VERSION)
+    throw new UnsupportedModelFactsSchemaError(value.schemaVersion);
+  if (!isRecord(value.overrides)) throw new Error('model-facts.json.overrides must be an object');
+  const keys = Object.keys(value.overrides);
+  if (keys.length > MODEL_FACTS_MAX_OVERRIDES)
+    throw new Error('model-facts.json has too many overrides');
+  const overrides: Record<string, ModelFactOverride> = {};
+  for (const key of keys) {
+    const match = PROVIDER_MODEL_KEY_PATTERN.exec(key);
+    if (!match || key.length > MODEL_FACT_KEY_MAX_LENGTH) throw new Error('Invalid model fact key');
+    modelFactKey(match[1]!, match[2]!);
+    overrides[key] = normalizeModelFactOverride(value.overrides[key]);
+  }
+  return { schemaVersion: MODEL_FACTS_SCHEMA_VERSION, overrides };
+}
+
+export function normalizeModelFactOverride(value: unknown): ModelFactOverride {
+  if (!isRecord(value)) throw new Error('Model fact override must be an object');
+  const allowed = new Set([
+    'displayName',
+    'description',
+    'apiProtocol',
+    'contextWindow',
+    'inputLimit',
+    'maxOutputTokens',
+    'knowledgeCutoff',
+    'structuredOutput',
+    'lastUpdated',
+    'capabilities',
+    'modalities',
+  ]);
+  for (const key of Object.keys(value))
+    if (!allowed.has(key)) throw new Error(`Unknown model fact field: ${key}`);
+  const result: Record<string, unknown> = {};
+  for (const key of ['displayName', 'description', 'knowledgeCutoff', 'lastUpdated'] as const) {
+    if (key in value) {
+      if (typeof value[key] !== 'string' || value[key].length > 2048)
+        throw new Error(`Invalid ${key}`);
+      result[key] = value[key];
+    }
+  }
+  if ('apiProtocol' in value) {
+    if (
+      value.apiProtocol !== 'openai-chat' &&
+      value.apiProtocol !== 'openai-responses' &&
+      value.apiProtocol !== 'anthropic-messages'
+    )
+      throw new Error('Invalid apiProtocol');
+    result.apiProtocol = value.apiProtocol;
+  }
+  for (const key of ['contextWindow', 'inputLimit', 'maxOutputTokens'] as const) {
+    if (key in value) {
+      const number = value[key];
+      if (!isPositiveBoundedInteger(number)) throw new Error(`Invalid ${key}`);
+      result[key] = number;
+    }
+  }
+  for (const key of ['structuredOutput'] as const) {
+    if (key in value) {
+      if (typeof value[key] !== 'boolean') throw new Error(`Invalid ${key}`);
+      result[key] = value[key];
+    }
+  }
+  if ('capabilities' in value) result.capabilities = normalizeCapabilities(value.capabilities);
+  if ('modalities' in value) result.modalities = normalizeModalities(value.modalities);
+  if (Object.keys(result).length === 0) throw new Error('Model fact override must not be empty');
+  return result as ModelFactOverride;
+}
+
+function normalizeCapabilities(value: unknown): NonNullable<ModelInfo['capabilities']> {
+  if (!isRecord(value)) throw new Error('Invalid capabilities');
+  const result: Record<string, boolean> = {};
+  const allowed = [
+    'chat',
+    'vision',
+    'reasoning',
+    'functionCalling',
+    'parallelToolCalls',
+    'imageGeneration',
+    'webSearch',
+  ] as const;
+  for (const key of allowed) {
+    if (key in value) {
+      if (typeof value[key] !== 'boolean') throw new Error(`Invalid capability: ${key}`);
+      result[key] = value[key];
+    }
+  }
+  for (const key of Object.keys(value))
+    if (!allowed.includes(key as (typeof allowed)[number])) {
+      throw new Error(`Unknown capability: ${key}`);
+    }
+  return result;
+}
+
+function normalizeModalities(value: unknown): NonNullable<ModelFactOverride['modalities']> {
+  if (!isRecord(value)) throw new Error('Invalid modalities');
+  if (value.input === undefined && value.output === undefined)
+    throw new Error('Invalid modalities');
+  const input = normalizeModalityDirection(value.input, isModality);
+  const output = normalizeModalityDirection(value.output, isOutputModality);
+  return {
+    ...(input === undefined ? {} : { input }),
+    ...(output === undefined ? {} : { output }),
+  };
+}
+
+function normalizeModalityDirection<T extends string>(
+  value: unknown,
+  allowed: (value: unknown) => value is T,
+): T[] | undefined {
+  if (value === undefined) return undefined;
+  if (!Array.isArray(value)) throw new Error('Invalid modality value');
+  const entries = Array.from(value);
+  if (!entries.every(allowed)) throw new Error('Invalid modality value');
+  return [...new Set(entries)];
+}
+
+function isModality(value: unknown): value is 'text' | 'image' | 'audio' | 'pdf' {
+  return value === 'text' || value === 'image' || value === 'audio' || value === 'pdf';
+}
+function isOutputModality(value: unknown): value is 'text' | 'image' | 'audio' {
+  return value === 'text' || value === 'image' || value === 'audio';
+}
+function isPositiveBoundedInteger(value: unknown): value is number {
+  return (
+    typeof value === 'number' &&
+    Number.isSafeInteger(value) &&
+    value > 0 &&
+    value <= MAX_FACT_NUMBER
+  );
+}
+function isRecord(value: unknown): value is Record<string, any> {
+  return typeof value === 'object' && value !== null && !Array.isArray(value);
+}
+
+export function applyModelFactOverride(
+  model: ModelInfo,
+  override: ModelFactOverride | undefined,
+): ModelInfo {
+  if (!override) return { ...model };
+  const overriddenFields = new Set<ModelFactField>(model.factOverriddenFields);
+  for (const field of Object.keys(override) as ModelFactField[]) overriddenFields.add(field);
+  // An authoritative context-window correction must not leave a stale,
+  // narrower provider input limit to silently win in the runtime resolver.
+  if (override.contextWindow !== undefined && override.inputLimit === undefined) {
+    overriddenFields.add('inputLimit');
+  }
+  // Modalities are merged per direction: a partial override changes only the
+  // direction it names and preserves the provider's other direction.
+  const modalities = override.modalities
+    ? {
+        input: override.modalities.input ?? model.modalities?.input ?? ['text'],
+        output: override.modalities.output ?? model.modalities?.output ?? ['text'],
+      }
+    : model.modalities;
+  const { modalities: _ignoredModalities, ...scalarOverride } = override;
+  return {
+    ...model,
+    ...scalarOverride,
+    id: model.id,
+    factOverriddenFields: [...overriddenFields],
+    ...(override.contextWindow === undefined || override.inputLimit !== undefined
+      ? {}
+      : { inputLimit: override.contextWindow }),
+    ...(override.capabilities === undefined
+      ? {}
+      : { capabilities: { ...model.capabilities, ...override.capabilities } }),
+    ...(modalities === undefined ? {} : { modalities }),
+  } satisfies ModelInfo;
+}
+
+type ModelFactConnectionLike = {
+  readonly providerType: ProviderType;
+  readonly defaultModel?: string;
+  readonly models?: readonly ModelInfo[];
+  readonly enabledModelIds?: readonly string[];
+};
+
+export function applyModelFactOverridesToConnection<T extends ModelFactConnectionLike>(
+  connection: T,
+  overrides: ModelFactOverrides,
+): T {
+  const providerModels = (connection.models ?? []).map((model) =>
+    applyModelFactOverride(
+      model,
+      lookupModelFactOverride(overrides, connection.providerType, model.id),
+    ),
+  );
+  const existing = new Set(providerModels.map((model) => model.id));
+  const enabled = new Set(
+    connection.enabledModelIds ??
+      (connection.defaultModel === undefined ? [] : [connection.defaultModel]),
+  );
+  const overrideOnly = [...enabled]
+    .filter((modelId) => !existing.has(modelId))
+    .map(
+      (modelId) =>
+        [modelId, lookupModelFactOverride(overrides, connection.providerType, modelId)] as const,
+    )
+    .filter((entry): entry is readonly [string, ModelFactOverride] => entry[1] !== undefined);
+
+  // Keep enabled models addressable even when the provider inventory fills the
+  // wire bound. Unenabled provider rows are the only entries eligible for
+  // eviction; the persisted inventory remains untouched.
+  const protectedProviderCount = providerModels.filter((model) => enabled.has(model.id)).length;
+  const reserve = Math.min(
+    overrideOnly.length,
+    Math.max(0, CONNECTION_CATALOG_MAX_MODELS_PER_CONNECTION - protectedProviderCount),
+  );
+  const targetProviderCount = CONNECTION_CATALOG_MAX_MODELS_PER_CONNECTION - reserve;
+  const models = providerModels.slice();
+  while (models.length > targetProviderCount) {
+    let evict = -1;
+    for (let index = models.length - 1; index >= 0; index -= 1) {
+      if (!enabled.has(models[index]!.id)) {
+        evict = index;
+        break;
+      }
+    }
+    if (evict < 0) break;
+    models.splice(evict, 1);
+  }
+  for (const [modelId, override] of overrideOnly.slice(0, reserve)) {
+    models.push(applyModelFactOverride({ id: modelId }, override));
+  }
+  return { ...connection, models } as T;
+}
+
+export function applyModelFactOverridesToCatalogSnapshot(
+  snapshot: ConnectionCatalogSnapshot,
+  overrides: ModelFactOverrides,
+): ConnectionCatalogSnapshot {
+  return {
+    ...snapshot,
+    connections: snapshot.connections.map(
+      (connection) =>
+        applyModelFactOverridesToConnection(
+          connection,
+          overrides,
+        ) as unknown as ConnectionCatalogEntry,
+    ),
+  };
+}
--- a/packages/runtime-host/src/protocol/runtime-policy.ts
+++ b/packages/runtime-host/src/protocol/runtime-policy.ts
@@ -78,6 +78,33 @@ export const CONNECTION_CATALOG_PAGE_MAX_BYTES = 48 * 1024;
 export const RUNTIME_POLICY_SNAPSHOT_MAX_BYTES = 48 * 1024;
 export const CREDENTIAL_SECRET_MAX_BYTES = 10 * 1024;
 
+const CONNECTION_MODEL_FIELDS = [
+  'id',
+  'displayName',
+  'description',
+  'apiProtocol',
+  'contextWindow',
+  'inputLimit',
+  'maxOutputTokens',
+  'knowledgeCutoff',
+  'structuredOutput',
+  'lastUpdated',
+  'capabilities',
+  'modalities',
+] as const;
+const MODEL_FACT_OVERRIDE_FIELDS = [
+  'displayName',
+  'description',
+  'apiProtocol',
+  'contextWindow',
+  'inputLimit',
+  'maxOutputTokens',
+  'knowledgeCutoff',
+  'structuredOutput',
+  'lastUpdated',
+  'capabilities',
+  'modalities',
+] as const;
 const QUERY_ERRORS = [
   'host_not_ready',
   'host_draining',
@@ -588,7 +615,7 @@ function catalogPageItem(value: unknown): ConnectionCatalogPageItem {
         0,
         CONNECTION_CATALOG_MAX_MODELS_PER_CONNECTION - 1,
       ),
-      model: decodeDomain(() => decodeConnectionModel(modelItem.model)),
+      model: decodeProjectedCatalogModel(modelItem.model),
     };
   }
   if (item.kind !== 'connection')
@@ -691,6 +718,32 @@ function catalogPageItem(value: unknown): ConnectionCatalogPageItem {
     ),
     modelCount,
   };
+}
+
+/** Decode catalog-only read-time provenance without admitting it to persisted models. */
+function decodeProjectedCatalogModel(value: unknown): ConnectionModel {
+  const item = requireShapedRecord(
+    value,
+    'projected connection model',
+    ['id'],
+    [...CONNECTION_MODEL_FIELDS.slice(1), 'factOverriddenFields'],
+  );
+  const { factOverriddenFields: rawOverriddenFields, ...persistentModel } = item;
+  const model = decodeDomain(() => decodeConnectionModel(persistentModel));
+  if (rawOverriddenFields === undefined) return model;
+  if (!Array.isArray(rawOverriddenFields) || rawOverriddenFields.length === 0) {
+    throw invalidProtocolFrame('Invalid model fact overridden fields');
+  }
+  const factOverriddenFields = rawOverriddenFields.map((field) => {
+    if (!(MODEL_FACT_OVERRIDE_FIELDS as readonly unknown[]).includes(field)) {
+      throw invalidProtocolFrame('Invalid model fact overridden field');
+    }
+    return field as (typeof MODEL_FACT_OVERRIDE_FIELDS)[number];
+  });
+  if (new Set(factOverriddenFields).size !== factOverriddenFields.length) {
+    throw invalidProtocolFrame('Duplicate model fact overridden field');
+  }
+  return { ...model, factOverriddenFields };
 }
 
 function decodeCreateConnectionInput(value: unknown): CreateCatalogConnectionInput {
--- a/packages/storage/src/model-facts-store.ts
+++ b/packages/storage/src/model-facts-store.ts
@@ -1,0 +1,99 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { createHash } from 'node:crypto';
+import {
+  decodeModelFactsDocument,
+  MODEL_FACTS_SCHEMA_VERSION,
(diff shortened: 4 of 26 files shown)


Nyvo-io:
Final revision is ready for maintainer review.

Addressed both CodeRabbit findings in one follow-up commit: prototype-key rejection now fails closed, and model-fact replacement now invalidates verification before persistence while superseding in-flight connection-test tickets.

Verification is complete: all GitHub checks pass, CodeRabbit reports no actionable comments, and the PR is mergeable. Local verification also passed Core 540/540 and the focused model-facts suites; Storage passed 783 tests with 14 platform skips, with the one unrelated bundled-Git smoke failure documented in the PR description.

Astro-Han:
Thanks for the rework — I verified the architecture is right: read-time projection (overrides never enter the persisted catalog, so session history stays clean — sessions store slug+modelId and re-resolve at execution), user > provider > metadata precedence with field-level capability merge, protected id, schema-validated fail-closed file, atomic `replaceModelFacts` write, and — critically — the execution path really is wired (`resolveExecutionConnection` projects, then `resolveSelectedModelContextWindow` reads the override, so the headline scenario "override contextWindow instead of falling back to 16K" works). The second commit's fail-closed clear-then-write race fix and prototype-pollution defense are sound, and the priority order matches the issue's cross-tool research consensus. CI 14/14 green.

**Conclusion: PASS with two P2s that need handling or an explicit deferral note.**

**P2-1 — the only production entrypoint is hand-editing the JSON file, and a bad file silently disables all overrides with zero diagnostics.** `readWithDiagnostics` returns a `diagnostic`, but every consumer takes only `.document.overrides` (coordinator.ts:1441-1443, :639-642, :1041-1046, :1244-1246), and grep confirms nothing in runtime-host/cli/ui references `modelFacts`/`replaceModelFacts` — the API has no production callers. A user who typo's `model-facts.json` (the only live path) loses every override silently, and the "bad models.json → error surfaced, built-ins still work" requirement from #2330 is only half met. At minimum surface the diagnostic (a warn log on the projection read, and an integration test that a bad file through the coordinator disables overrides — today only the store layer tests this).

**P2-2 — the invalidation semantics only hold for the (uncalled) API path, not for the file-edit path that is the real entrypoint.** `modelFactsGeneration` is only bumped by `replaceModelFacts`; an external edit of the file doesn't move it, and `sameConnectionTestModelBasis` compares only `id`+`apiProtocol` — so editing only `contextWindow` keeps a stale `lastTest: verified` badge and doesn't cancel an in-flight connection test (whose result then commits under the new facts). Either derive the generation from the file's mtime/stat on the read path, or explicitly document "file edits are advisory; invalidation only happens via replaceModelFacts". Also: `replaceModelFacts` clearing lastTest for *all* connections (including unrelated providers) is over-invalidation — consider narrowing to the affected provider/model.

P3 (optional): the `modelFactOverrides` pipeline in model-catalog.ts is dead in production (no caller passes it; the desktop model menu reads the storage projection) — ~240 lines of code+tests serving only itself; delete if the upcoming UI PR won't consume it; the `webSearch` capability passthrough is a behavior change for *unoverridden* provider models too (previously discarded, now surfaced in the catalog) — not declared; dual authority with `relayModelProfiles` (per-connection, relay-specific): on relay connections the relay profile silently wins over the global file for overlapping fields (contextWindow/vision) — the precedence is undocumented; every catalog snapshot / connection resolution / test-start re-reads and re-freezes the full projection even with no overrides — a stat/mtime cache would help; the `error.message.includes('exceeds')` diagnostic classification is fragile to wording changes in document-io.ts; 512 overrides × 2048-char strings can exceed the 256KB byte cap *after* `prepareReplacement` clears lastTests → `commit_outcome_unknown` with verification discarded but overrides not written — byte-budget check belongs in prepare; untrimmed `enabledModelIds` are pushed verbatim as override-entry ids while lookup trims — a whitespace id creates a dirty entry.

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagent on `opencode-go/deepseek-v4-flash`), which traced the projection/priority/invalidation paths and grepped for production callers. P2-1 and P2-2 are consequences of the caller graph (both surfaced as unverified predictions). Please weigh these findings with your own judgment.*

<details>

结论：PASS（2 个 P2 需处理或显式延后）。架构正确：读时投影（覆盖不进持久化 catalog、Session 历史干净——session 只存 slug+modelId 执行时重解析）、优先级 user>provider>metadata+字段级 capability merge、id 受保护、schema 校验 fail-closed、replaceModelFacts 原子写，执行路径真打通（resolveExecutionConnection 投影后 resolveSelectedModelContextWindow 取覆盖值，headline 场景"覆盖 window 而非 16K 回退"成立）；第二 commit 的 fail-closed clear-then-write 竞态修复与原型污染防御正确；优先级顺序符合 issue 跨工具调研共识。CI 14/14 绿。P2-1：唯一生产入口是手改 JSON，坏文件静默失效全部覆盖零诊断——readWithDiagnostics 返回 diagnostic 但所有消费者只取 .document.overrides，且 grep 确认 runtime-host/cli/ui 零引用 modelFacts/replaceModelFacts（API 无生产调用者）；打错一个字符所有覆盖静默消失，#2330 的"坏文件→error surfaced、内置照常"只完成一半。至少：投影读路径 warn 日志 + "坏文件经 coordinator 投影后覆盖禁用"集成测试（现只测到 store 层）。P2-2：作废语义只对（无人调用的）API 路径成立——modelFactsGeneration 仅 replaceModelFacts 递增，文件外部编辑不改变；sameConnectionTestModelBasis 只比较 id+apiProtocol，仅改 contextWindow 的编辑保留过期 lastTest: verified 且不拦截飞行中 connection test（后者以旧事实测试结果提交到新事实下）。建议：读路径用文件 mtime/stat 派生 generation，或显式声明"文件编辑为 advisory，作废仅经 replaceModelFacts"；另 replaceModelFacts 对所有连接（含无关 provider）清 lastTest 属过度作废，建议按受影响 provider/model 收窄。P3（可选）：model-catalog.ts 的 modelFactOverrides 管线在生产是死代码（无生产调用者，桌面模型菜单走 storage 投影）——若后续 UI PR 不消费应删除（~240 行）；webSearch 能力透传对未覆盖的 provider 模型也是行为变更（此前被丢弃现在出现在 catalog）未声明；与 relayModelProfiles 双权威（relay 连接上 relay profile 静默压过全局文件的重叠字段）优先级未文档化；每次 catalog 快照/resolve/test-start 都重读+refreeze 全量投影（可加 stat/mtime 缓存）；错误分类依赖 error.message.includes('exceeds') 脆；512 覆盖×2048 字符可能超 256KB 字节上限且发生在 prepareReplacement 清 lastTests 之后→commit_outcome_unknown（验证已丢覆盖未写入），字节预算应在 prepare 阶段；未 trim 的 enabledModelIds 原样 push 而 lookup trim——空白 id 产生脏条目。
</details>

packages/storage/src/runtime-policy/coordinator.ts
```diff
@@ -201,7 +210,44 @@ export class RuntimePolicyCoordinator {
...
+  getModelFacts() {
+    return this.inLane(async (root) => deepFreeze(await this.modelFacts.readWithDiagnostics(root)));
+  }
+
+  replaceModelFacts(overrides: ModelFactOverrides) {
+    return this.inLane(async (root) => {
+      const document = this.modelFacts.prepareReplacement(overrides);
+      let cleared = false;
+      try {
+        cleared = await this.catalog.clearAllConnectionLastTests(
```
Astro-Han: **P2 — Oversized replacements invalidate verification before they are known to be writable.** `prepareReplacement()` enforces shape/item limits but not the 256 KiB serialized budget; that check occurs later in `writeReplacement()`. A structurally valid 512-entry document can therefore clear every `lastTest`, fail the byte check, retain the old facts, and return `commit_outcome_unknown`. Serialize and enforce the final byte budget before `clearAllConnectionLastTests()`, then add a regression at the maximum field/item sizes.

Astro-Han (commented):
The field-level precedence (`user > provider > metadata`), immutable persisted authority, atomic replacement, generation tickets, and execution projection are coherent. Existing public threads already cover malformed-file diagnostics and external-edit invalidation, so I am not duplicating them.

One additional transaction-ordering issue remains below. The first-principles rule is that the complete candidate document—including serialized byte budget—must be validated before any dependent state is invalidated. More broadly, a content fingerprint can be the single authority for both external-edit generation and test invalidation, instead of maintaining parallel API-only generation semantics.

Review performed with three Codex reviewer agents and DeepSeek V4 Flash as advisory tools; I reproduced the finding on the latest head. The branch also needs rebasing because GitHub currently reports conflicts with `main`.

字段级 precedence（`user > provider > metadata`）、不可变持久权威、原子替换、generation ticket 和执行投影整体闭环。现有公开线程已覆盖 malformed-file diagnostic 与外部编辑失效，我不重复。

以下仍有一个额外的事务顺序问题。第一性原理是：完整候选文档（包括序列化字节预算）必须在任何依赖状态失效之前完成验证。更进一步，可用内容 fingerprint 作为外部编辑 generation 与测试失效的单一权威，避免维护只对 API 写入生效的并行 generation 语义。

本次审查使用了三位 Codex reviewer agents 与 DeepSeek V4 Flash 作为辅助工具；我已在最新 head 上复现该问题。当前 GitHub 还显示分支与 `main` 冲突，需要 rebase。

</details>

packages/storage/src/runtime-policy/coordinator.ts
```diff
@@ -879,7 +942,10 @@ export class RuntimePolicyCoordinator {
           connectionBasis(checked.connection),
           result,
         );
-        return deepFreeze({ kind: 'committed' as const, snapshot });
+        return deepFreeze({
+          kind: 'committed' as const,
+          snapshot: await this.projectCatalogSnapshot(root),
```
Astro-Han, line 1087: [P2] Preserve enabled override-only models across provider refresh

By the time this projected snapshot is returned, writeModelFetchResult has already reconciled enabledModelIds strictly against the fetched inventory. An enabled custom model backed by model-facts.json is therefore removed (and may lose the default) whenever the provider returns only its live inventory; the projection here cannot add it back because it is no longer enabled. Please pass the current facts-backed IDs into model-fetch reconciliation, preserve only those already enabled/defaulted, and add a refresh regression proving the custom model remains selectable and executable.

返回这个 projected snapshot 时，writeModelFetchResult 已经只按 provider inventory 重算 enabledModelIds。由 model-facts.json 支撑的已启用 custom model 会在刷新时被删除，默认项也可能被切换；此时 projection 因该 ID 已不再 enabled，无法把它加回来。请把当前 facts-backed IDs 传入 reconciliation，只保留已经 enabled/default 的部分，并补刷新后仍可选择与执行 custom model 的回归。

</details>

Astro-Han (commented):
The immutable facts document, field-level projection, generation tickets, and execution resolver are coherent on this head; I also verified that the execution resolver applies overrides before returning, so an initially suspected execution bypass was discarded. One normal refresh path still removes the headline override-only model state.

The inline P2 is the smallest remaining product correction: model fetch reconciliation needs to preserve currently enabled IDs backed by model facts without writing them into provider inventory. The current live typecheck is also failing, so this head is not merge-ready independently of the review finding.

AI-assisted review by Codex with three independent reviewer passes and OpenCode Go DeepSeek V4 Flash (high) as an advisory pass; I verified the current head, refresh reconciliation, projection semantics, existing threads, and live CI.

当前 head 的不可变 facts document、字段级 projection、generation ticket 与 execution resolver 整体闭环；我也确认 execution resolver 返回前确实应用 overrides，因此剔除了一个初步误报。仍有一个正常刷新路径会删除核心的 override-only model 状态。

行内 P2 是最小剩余修复：model fetch reconciliation 应保留当前已启用且由 model facts 支撑的 ID，但不要把它写进 provider inventory。当前实时 typecheck 也失败，因此即使不考虑 finding，这个 head 也尚不可合并。

本次由 Codex、三个独立 reviewer 与 OpenCode Go DeepSeek V4 Flash high 辅助；已核对当前 head、刷新 reconciliation、projection semantics、已有线程和实时 CI。

</details>

packages/storage/src/runtime-policy/coordinator.ts
Astro-Han: Verified fixed on current head `8c592fd52`: the complete replacement is serialized and checked against `MODEL_FACTS_DOCUMENT_MAX_BYTES` during preparation, before `clearAllConnectionLastTests`, and the regression preserves existing verification for an oversized candidate.

packages/core/src/model-facts.ts
```diff
@@ -0,0 +1,231 @@
...
+    'inputLimit',
+    'maxOutputTokens',
+    'knowledgeCutoff',
+    'structuredOutput',
+    'lastUpdated',
+    'capabilities',
+    'modalities',
+  ]);
+  for (const key of Object.keys(value))
+    if (!allowed.has(key)) throw new Error(`Unknown model fact field: ${key}`);
```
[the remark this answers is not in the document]
Astro-Han: Thanks — I do not think pricing belongs in this persisted schema. The linked #2330 decision explicitly keeps pricing overrides in the existing pricing editor and says not to duplicate them in `model-facts.json`. This file owns the non-pricing model facts that otherwise have no override channel, so rejecting `pricing` here preserves that authority boundary.

Astro-Han (commented):
The current head is semantically equivalent to the previously reviewed implementation plus rebase/formatting. I verified that the oversized-input validation, sparse-modality validation, and post-publication generation invalidation are now covered; I also resolved the pricing thread because #2330 explicitly keeps pricing in the existing pricing editor rather than duplicating it in `model-facts.json`.

The existing [provider-refresh P2](https://github.com/apache/maka/pull/3129#discussion_r3810799366) remains current and unresolved: refresh reconciliation still removes an enabled/default override-only model before the facts projection runs. No duplicate inline is needed. This also changes user-visible model selection by making override-only models selectable, but the PR has no visual evidence; please add an actual TUI or Desktop model-selector screenshot showing a facts-backed custom model. Exact-head CI has not run yet because the Apache Actions lanes are unavailable/queued.

AI-assisted review by OpenAI Codex. I verified the exact-head range-diff, current implementation and tests, issue #2330’s ownership decision, review-thread state, screenshot gate, provenance, and live checks.

当前 head 与上一轮相比主要是 rebase/format；oversized input、稀疏 modalities 和发布后 generation 失效均已有修复。pricing 线程不成立，因为 #2330 明确要求 pricing 继续由现有 pricing editor 负责，不应重复进入 `model-facts.json`。

已有的 provider refresh P2 仍然有效：刷新会在 facts projection 前删除已启用/默认的 override-only model，因此不重复发 inline。该改动会让自定义模型出现在用户选择器中，也需要补一张真实 TUI 或 Desktop 模型选择器截图。当前 exact-head CI 尚未实际运行。

本次由 OpenAI Codex 辅助，已核对 exact-head range-diff、实现与测试、#2330 权威边界、线程状态、截图门禁、来源披露和实时检查。

</details>

Nyvo-io:
Implemented and reviewed the provider-refresh correction in one follow-up commit: f321119a7.\n\n- Refresh reconciliation now preserves only facts-backed model IDs that are already enabled or are the connection's current default.\n- Raw provider inventories remain limited to live discovery results; custom facts are projected read-time.\n- The default target remains valid and the execution resolver still resolves the selected custom model.\n- Added Core and Storage regressions covering refresh, default retention, execution, inventory purity, and exclusion of unselected facts-backed models.\n- Rebased onto current main.\n\nVerification on the rebased head:\n- Core typecheck passed.\n- Storage typecheck passed after rebuilding Core.\n- Biome check passed for all changed files.\n- git diff --check passed.\n- Core reconciliation tests: 9 passed.\n- Storage model-facts tests: 6 passed.\n- Storage runtime-policy-stores tests: 44 passed.\n- Independent adversarial review found no actionable defect.\n\nPricing remains intentionally outside model-facts.json, per issue #2330's authority decision; the existing pricing editor remains the pricing source of truth.

packages/core/src/model-catalog.ts
```diff
@@ -290,11 +302,17 @@ function makeEntry(
     providerType: input.providerType,
     ...(input.connectionSlug ? { connectionSlug: input.connectionSlug } : {}),
     source,
-    capabilitySource: normalizedModel.capabilities
-      ? source
-      : metadata.capabilities
-        ? 'static_catalog'
-        : 'unknown',
+    capabilitySource: lookupModelFactOverride(
```
Astro-Han: [P1] This projection path has no production caller, so `capabilitySource: 'user_override'` can never fire. The ~94 lines threading `modelFactOverrides` through the catalog builders are reached only from `model-catalog.test.ts`; a search over the whole tree at this head finds no other caller. Production instead arrives here with a connection whose `models[]` was already merged by the storage-side projection in `coordinator.ts`, so `input.modelFactOverrides` is always `undefined`, `lookupModelFactOverride` returns undefined, and the ternary falls through to `'provider_api'`. Concretely: a user writes `{"openai:gpt-5":{"contextWindow":32000,"capabilities":{"vision":false}}}`, and Settings shows 32 000 attributed to the provider API. The user cannot distinguish their own pin from a provider report — which is exactly the signal they need months later when the provider has corrected the value and the pin is now the stale one. Either delete this path and derive provenance inside the storage projection (project a `factOverriddenFields` set alongside `models[]`), or wire the real callers. The regression test has to go through the same call shape production uses; the two existing tests pass `modelFactOverrides` directly, which is why this was invisible.

packages/storage/src/runtime-policy/coordinator.ts
```diff
@@ -1343,6 +1450,52 @@ export class RuntimePolicyCoordinator {
...
+    const facts = await this.readModelFacts(root);
+    return deepFreeze(
+      applyModelFactOverridesToCatalogSnapshot(
+        catalogSnapshot(await this.catalog.read(root)),
+        facts.document.overrides,
+      ),
+    );
+  }
+
+  private async readModelFacts(root: string) {
```
Astro-Han, line 1985: [P1] Make this pure — a read must not perform a durable catalog write. `readModelFacts` calls `clearAllConnectionLastTests` on a fingerprint change, which is a real write that bumps every connection's revision, and failures surface as `commit_outcome_unknown`. It is reached from `projectCatalogSnapshot`, `getModelFacts`, and `resolveExecutionConnection` — that last one being on the send path. Concretely: a user with six verified connections hand-edits `model-facts.json` to fix one Ollama model's window; the next `resolveExecutionConnection` for an unrelated Anthropic connection wipes verification on all six, which then render as not verified, and if the catalog write fails the user's send fails with a commit-unknown error rather than degrading to a stale read. Two problems compounded: the write happens on a read, and the invalidation is global when the change is per-model. Move the generation bump and the invalidation into an explicit write-path reconcile, and scope it to connections whose projected model set actually changed. Note also that `openInteractiveRuntimePolicyStoresForRead` builds this coordinator on a read-access lease and `runWithStorageRootLease` only checks identity rather than blocking writes — that variant has no production caller today, so the read-lease angle is inference, but it is the kind of assumption worth not leaving lying around.

packages/core/src/model-facts.ts
```diff
@@ -0,0 +1,243 @@
...
+  if ('apiProtocol' in value) {
+    if (
+      value.apiProtocol !== 'openai-chat' &&
+      value.apiProtocol !== 'openai-responses' &&
+      value.apiProtocol !== 'anthropic-messages'
+    )
+      throw new Error('Invalid apiProtocol');
+    result.apiProtocol = value.apiProtocol;
+  }
+  for (const key of ['contextWindow', 'inputLimit', 'maxOutputTokens'] as const) {
```
Astro-Han, line 152: [P2] An override that raises `contextWindow` is silently discarded by the `min` against `inputLimit`. This PR newly persists `inputLimit` on stored model rows and newly lets a user override `contextWindow`, but `resolveSelectedModelContextWindow` takes `narrowestPositiveLimit(model.contextWindow, model.inputLimit)`, and `applyModelFactOverride` sets only `contextWindow` and leaves the stored `inputLimit` alone. Reproduced by execution against verbatim copies: row `{contextWindow: 8192, inputLimit: 8192}` plus override `{contextWindow: 200000}` yields an applied row of `{contextWindow: 200000, inputLimit: 8192}` and a truncation budget of 8192 — while the catalog entry reports 200 000. So the number the user sees and the number that governs truncation disagree, which is the single worst outcome for a feature whose purpose is to correct that number. When an override sets `contextWindow`, clear or raise the row's `inputLimit` in `applyModelFactOverride`. Regression test: exactly the pair above. That a discovery adapter emits `inputLimit` today is inference — only the generated metadata sets it now — but the codec is asserted to pass it through, so the row shape is reachable.

packages/storage/src/runtime-policy/coordinator.ts
```diff
@@ -581,7 +642,10 @@ export class RuntimePolicyCoordinator {
       if (prepared.kind !== 'ready') return prepared;
       return deepFreeze({
         kind: 'ready' as const,
-        connection: structuredClone(connection),
+        connection: applyModelFactOverridesToConnection(
```
Astro-Han, line 782: [P2] Pick one authority for a user-declared context window. Merging the fact override into `models[]` here does not make it win: `resolveSelectedModelContextWindow` consults `relayModelProfile(connection, modelId)?.contextWindow` first and returns early, so on an `openai-compatible` connection the older setting silently beats the newer fact. Concretely: a user sets `relayModelProfiles['m'].contextWindow = 8192` in settings, later writes `{"openai-compatible:m":{"contextWindow":200000}}` in `model-facts.json`, and truncation stays at 8192 while the model picker shows 200 000 — no warning, no precedence documented anywhere. `RelayModelProfile` is the pre-existing per-model user-declaration seam and it already has a documented gate, so generalising it was the closest existing seam; running a second override table beside it is the thing the project's own rule asks us not to do. At minimum make `resolveSelectedModelContextWindow` the single resolver for both and state the order in one place — but the better answer is one table, not two.

packages/core/src/model-facts.ts
```diff
@@ -0,0 +1,243 @@
...
+  providerType: ProviderType | string,
+): string[] {
+  if (!overrides) return [];
+  const prefix = `${providerType.trim()}:`;
+  return Object.keys(overrides)
+    .filter((key) => key.startsWith(prefix))
+    .map((key) => key.slice(prefix.length));
+}
+
+export function decodeModelFactsDocument(value: unknown): ModelFactsDocument {
```
Astro-Han, line 97: [P2] Distinguish an unsupported future version from a malformed file, and refuse to write over one. `decodeModelFactsDocument` requires `schemaVersion === 1` exactly; `readWithDiagnostics` swallows the throw into `diagnostic: 'malformed'` with an empty document; and `replaceModelFacts` then overwrites the whole file with a v1 document. So a user who runs a newer build that writes `schemaVersion: 2` and then reverts to this one loses every override silently — the only signal is a `process.emitWarning` — has all their connection verification cleared on the next read, and has the v2 data destroyed by the first write. Version downgrade is a normal thing for users on a release channel to do, and this is a one-way door. Treat an unknown-but-parseable version as a refusal to write rather than as corruption, and add a regression test that a `schemaVersion: 2` document survives a `replace`.

packages/storage/src/model-facts-store.ts
```diff
@@ -0,0 +1,112 @@
...
+    let value: unknown;
+    try {
+      value = JSON.parse(new TextDecoder('utf-8', { fatal: true }).decode(bytes)) as unknown;
+      return { document: decodeModelFactsDocument(value), fingerprint };
+    } catch {
+      return { document: emptyDocument(), diagnostic: 'malformed', fingerprint };
+    }
+  }
+
+  async replace(root: string, overrides: ModelFactOverrides): Promise<ModelFactsDocument> {
```
Astro-Han: [P2] Adopt the revision-checked shape the sibling override store in this same package already uses. `replace(root, overrides)` swaps the whole table with no expected revision, so two windows that each read the table, each add one override, and each call `replace` silently lose the first one's entry — and `getModelFacts` returns no revision, so no caller can even detect the loss. `packages/storage/src/pricing-store.ts` is the existing user-override authority right next door and it does this correctly: `upsert(expectedRevision, ...)` / `delete(expectedRevision, ...)` with a `PricingRevisionConflictError`. Following it removes the lost-update entirely, and per-key writes also make the 256 KiB whole-document byte check unnecessary.

packages/core/src/model-facts.ts
```diff
@@ -0,0 +1,243 @@
...
+}
+
+const PROVIDER_ID_PATTERN = /^[^:\s]{1,128}$/;
+// Model ids may contain colons (for example, Ollama's `gpt-oss:120b`). The
+// provider is the only component that is constrained to the first separator.
+const MODEL_ID_PATTERN = /^[^\s]{1,256}$/;
+const PROVIDER_MODEL_KEY_PATTERN = /^([^:\s]{1,128}):([^\s]{1,256})$/;
+const MAX_FACT_NUMBER = 10_000_000_000;
+
+export function modelFactKey(providerType: ProviderType | string, modelId: string): string {
```
Astro-Han, line 58: [P2] Validate the provider segment against the registered providers, or report keys that matched nothing. `modelFactKey` types the provider as `ProviderType | string` and checks only that it has no colon, no whitespace and is at most 128 characters — so `"anthropic:claude-opus-5"` is happily stored for a connection whose `providerType` is `anthropic-api`, and the override then never applies anywhere. The document validates, `getModelFacts` reads it back intact, and the only symptom the user gets is that nothing changed — for a hand-edited JSON file, which is the primary way this feature is used, that is a bad failure mode. Either reject unknown provider segments on write, or have `getModelFacts` report unmatched keys so the UI can say which overrides are inert.

packages/core/src/model-facts.ts
```diff
@@ -0,0 +1,243 @@
...
+): ModelInfo {
+  if (!override) return { ...model };
+  return {
+    ...model,
+    ...override,
+    id: model.id,
+    ...(override.capabilities === undefined
+      ? {}
+      : { capabilities: { ...model.capabilities, ...override.capabilities } }),
+    ...(override.modalities === undefined ? {} : { modalities: override.modalities }),
```
Astro-Han: [P3] Merge modalities per direction, or document that they replace wholesale. `capabilities` merge field by field but `modalities` replace the whole object, and nothing in the type says so. Reproduced by execution: a row with `{input: ['text','image','pdf'], output: ['text']}` plus an override of `{input: ['text'], output: ['text']}` drops `image` and `pdf` — permanently, including any input modality the provider adds later. A user who only wanted to remove `pdf` has silently also removed `image`, and since the whole point of the pin is that it survives provider updates, they will not find out from a later correction. Per-direction merge matches how `capabilities` already behave and is what a reader would expect.

packages/core/src/model-facts.ts
```diff
@@ -0,0 +1,243 @@
...
+  const overrides: Record<string, ModelFactOverride> = {};
+  for (const key of keys) {
+    const match = PROVIDER_MODEL_KEY_PATTERN.exec(key);
+    if (!match || key.length > MODEL_FACT_KEY_MAX_LENGTH) throw new Error('Invalid model fact key');
+    overrides[key] = normalizeModelFactOverride(value.overrides[key]);
+  }
+  return { schemaVersion: MODEL_FACTS_SCHEMA_VERSION, overrides };
+}
+
+export function normalizeModelFactOverride(value: unknown): ModelFactOverride {
```
Astro-Han, line 118: [P3] Give the user a way to say "this fact is unknown", not just "this fact is that". `normalizeModelFactOverride` accepts only present-and-valid values, with no `null` tombstone, so when a provider reports a wrong `contextWindow` the user's only option is to substitute another number they would have to guess — they cannot express "ignore the provider here and fall back to bundled metadata", which for a correction feature is the more honest of the two things they might want. Reproduced by execution. A `null` sentinel in the allowlist would cover it.

packages/core/src/model-catalog.ts
```diff
@@ -34,6 +39,7 @@ export interface KnownModelCapabilities {
   reasoning?: true;
   functionCalling?: true;
   imageGeneration?: true;
+  webSearch?: true;
```
Astro-Han, line 66: [P3] `webSearch` here is a separate concern with no production reader. The only thing that reads `entry.capabilities.webSearch` is `model-catalog.test.ts`; `model-web-search.ts` reads `ModelInfo.capabilities.webSearch` off the model row, not the catalog entry. So this adds a field to a public shape for a consumer that does not exist, inside a PR about user-overridable facts. The repository asks that product changes stay separate unless they must ship together — either wire the reader or move it to its own change.

Astro-Han (commented):
Thanks — the storage half of this is genuinely careful, and I want to name that before the findings. Reviewed exact head `f321119a7f2ebf3576e1a9bc9decfa2ea8326249`.

The store gets the hard mechanics right: atomicity and temp-file recovery, `commit_outcome_unknown` on both the pre-clear and post-publication failure paths, oversize rejection that preserves verification, external-edit ticket supersession, and `__proto__` rejection — with the key grammar handling colon-bearing model ids like `ollama-cloud:gpt-oss:120b` correctly, because only the provider segment is colon-free. The numeric validator is tighter than the codec's and rejects zero, negatives, non-integers and infinities. Most importantly, no write path reads a projected snapshot and writes it back, so overrides never contaminate `connection-catalog.json` — that is what keeps a stale pin recoverable, and there is an explicit assertion for it. Pricing is not reachable from this PR at all, since the override allowlist excludes price fields and pricing has its own authority; the wrong-number risk here is truncation, not cost.

The problem is the precedence chain. Context window now has three authorities — `relayModelProfiles`, this new fact table, and bundled metadata — and `resolveSelectedModelContextWindow` consults the first one first and returns early, so on an `openai-compatible` connection a model fact silently loses to a setting the user may have configured months earlier, with no diagnostic. `RelayModelProfile` is the pre-existing per-model user-declaration seam with a documented gate; the project rule is to extend the closest existing seam rather than run a parallel one beside it, and that is what happened here. Separately, the PR threads a second, entirely dead projection path through the catalog builders, which is why the one provenance signal it adds can never fire.

Two P1s, six P2s and three P3s inline. Not approving while P1/P2 findings are open.

On tests: the storage suite is strong on mechanics and empty on precedence. Nothing asserts a fact override against a competing authority, nothing drives `capabilitySource` through the call shape production actually uses — which is precisely why the dead path went unnoticed — nothing covers a `contextWindow` override against a row carrying the `inputLimit` this same PR starts persisting, and nothing covers a non-1 `schemaVersion`, a concurrent `replace`, or an override for a model that no longer exists. Each of those inherits its severity from the finding it belongs to rather than standing on its own.

Review disclosure: this review was prepared with Claude Code, which read the diff at this head, enumerated the reader sites and the precedence chain by search rather than by assumption, and executed several of the pure functions against verbatim copies to check the merge semantics. Evidence grade is stated per finding — three were reproduced by execution, the rest are code reading or labelled inference. The human contributor reviewed this before posting.

Nyvo-io:
Final revision is ready for maintainer review.

This single follow-up commit adds the requested Desktop model-selector evidence and updates the runtime-policy regression expectation for the new facts provenance field.

Validation completed on head `7c4bbffa4`: Core 568/568 passed; Storage model-facts and runtime-policy-stores suites passed (11 + 44); Core/Storage/Runtime typechecks, Biome, `git diff --check`, and the local strict review all passed. The full Storage suite has one unrelated environment failure because this checkout lacks `node_modules/dugite/git/bin/git`; the remaining suite reported 832 passed, 14 skipped, and the existing bundled-Git smoke failure only.

No actionable findings remain from the local review. Ready for maintainer review.

Astro-Han:

- `packages/core/src/model-catalog.ts`

```
# resolve, then
git push --force-with-lease
```

---

packages/storage/src/model-facts-store.ts
```diff
@@ -0,0 +1,159 @@
...
+    const fingerprint = fingerprintBytes(bytes);
+    let value: unknown;
+    try {
+      value = JSON.parse(new TextDecoder('utf-8', { fatal: true }).decode(bytes)) as unknown;
+      return { document: decodeModelFactsDocument(value), fingerprint };
+    } catch (error) {
+      if (error instanceof UnsupportedModelFactsSchemaError) {
+        return { document: emptyDocument(), diagnostic: 'unsupported_schema', fingerprint };
+      }
+      return { document: emptyDocument(), diagnostic: 'malformed', fingerprint };
```
Astro-Han, line 65: [P3] This `diagnostic: 'malformed'` never reaches the person who caused it.

The only place in the repository that acts on it is `runtime-policy/coordinator.ts:1531-1537`, which calls `process.emitWarning('model-facts.json is <diagnostic>; ignoring its overrides')`, deduplicated by fingerprint. Both explicit guards — `model-facts-store.ts:111` and `coordinator.ts:247` — test only for `=== 'unsupported_schema'`. Nothing anywhere tests for `'malformed'`.

Trigger → path → outcome: a user hand-edits `model-facts.json` and mistypes one field → the **entire document** is rejected, not just that entry → every override disappears from the model picker → and the only explanation is a process warning on stderr. A desktop user will not look there. What they experience is "my overrides vanished and I don't know why."

`ModelFactsReadResult.diagnostic` is already part of the return value, so surfacing a notice in Settings should be inexpensive. Worth doing precisely because the write path is otherwise so careful — strict validation on write makes a silent whole-document rejection on read more surprising, not less.

Astro-Han (commented):
Reviewed on exact head `a70fbf3d3e5395b921ff921f3d8fcb3fa3af2340`. **No P0–P2.** One `[P3]` inline. Not approving only because `test` is red — see the end.

Two questions decided whether this feature is safe, and both come out well.

**Precedence between user overrides and built-in metadata is coherent, and the same rule is implemented in both layers for the same stated reason.** When `contextWindow` is overridden but `inputLimit` is not, `inputLimit` is carried to the new value, with the reasoning written down: a stale, narrower provider `inputLimit` must not silently win at resolution time. The execution layer agrees — `context-budget-policy.ts:441` makes a facts pin on `contextWindow` take precedence over the older relay-only declaration and then takes `narrowestPositiveLimit(contextWindow, inputLimit)`, justified as "catalog display and actual execution must use the same window." One rule, two layers, one rationale. That is the part I most want to credit.

**A bad override cannot brick a model.** The write path is strict: `normalizeModelFactOverride` whitelists fields and throws `Unknown model fact field` on anything else, caps strings at 2048, requires `isPositiveBoundedInteger` for numerics, restricts `apiProtocol` to three enum values, requires keys to match `provider:model` within 512 characters, and caps the override count at 512. Bad values are rejected as `RuntimePolicyStoreError('invalid_policy_input')` before they can be persisted. The read path is fail-open: `model-facts-store.ts:52-57` returns an empty document plus a diagnostic rather than throwing, so the model falls back to provider facts and stays usable.

**Candidates considered and dropped** — listed so they can be re-checked rather than re-derived:

- *"One typo permanently destroys 511 other overrides."* I initially graded this a P2 data-loss issue and **withdrew it**: the `fingerprint` is computed from the real bytes *before* decoding (`:48`), so optimistic concurrency checks the true file identity. `unsupported_schema` is explicitly protected from being overwritten while `malformed` is not, and that asymmetry is principled — unparseable content cannot be merged into, so replacing it is reasonable, whereas a file written by a newer version that this build simply does not understand must never be clobbered. Correct as designed.
- *`contextWindow` has no lower bound (`isPositiveBoundedInteger` accepts 1).* Not reported. This is an explicit user fact override; second-guessing it would defeat the premise of the feature.

**Gating:** `audit` is green, `test` is `completed/failure` on missing ASF license headers. That is a mechanical fix — running `npm run write:asf-headers` and pushing should clear it. Nothing on the code side blocks from my end; happy to re-check the terminal state afterwards.

Nyvo-io:
Rebased onto current main and updated the single commit. The Runtime Host catalog projection now keeps the persisted model-facts verification fingerprint out of the client-visible protocol header; empty model-fact overrides are also rejected before projection. Final CI is green: test and audit passed. Local lint, format, ASF-header, build, typecheck, Core, and focused Runtime Host regression checks also passed.

Nyvo-io:
Rebased onto current `main` and pushed the single verified commit `9dc0d5dc4`. The exact-head GitHub checks are all green (`test`, `audit`, and `windows_recovery`), and local build, lint, format, ASF-header, Core, Runtime, Storage, and Runtime Host verification passed. No new code changes were made after review. Ready for maintainer review.

Astro-Han (approved):
Thanks for your patience here. I re-reviewed exact head `e28b8217c3`; CI is green and I found no remaining P0/P1 issue. The earlier findings around refresh loss, read-side durable writes, context/input-limit precedence, relay precedence, schema downgrade, and provenance are fixed or superseded on this head.

Two non-blocking P2 observations remain:

- display-only fact changes currently participate in the connection-verification fingerprint and can unnecessarily clear a valid verification result;
- a provider inventory at the supported 2,048-entry boundary plus one enabled override-only model can produce catalog index 2,048, beyond the protocol’s accepted `0…2047` range.

I also found several concrete simplifications that should preferably be made in this PR:

- delete the entire `factBackedModelIds` plumbing: it is passed through several layers but never read by `reconcileConnectionAfterModelFetch`, whose existing behavior already preserves enabled IDs;
- delete the catalog builder’s raw `modelFactOverrides` input and its tests: no production caller uses it, while the real Runtime Host path already supplies projected `ModelInfo` with `factOverriddenFields`;
- remove the unused `ModelFactsDocumentOwner` wrappers and fingerprint helper;
- avoid production injection seams that exist only to make one directory-sync test convenient;
- unless a real Runtime Host/IPC writer is included here, remove the currently unconsumed `RuntimePolicyStores.modelFacts` get/replace facade and introduce it with its first production caller;
- move the two `.maka-shots` screenshots into the PR evidence and remove the unreferenced binaries from the repository.

These deletions remove parallel or unused paths without changing the user-facing model-facts behavior. With them treated as non-blocking cleanup, I’m approving the current solution.

OpenAI Codex and delegated `@reviewer` agents assisted with authority, production-path, test, and simplification analysis. Astro-Han reviewed the evidence and made the final approval decision.

<details>
<summary>中文对照</summary>

谢谢你耐心等待。我重新检查了当前 head `e28b8217c3`；CI 全绿，当前没有剩余 P0/P1。之前关于刷新丢失、读取路径写入、context/input limit 优先级、relay 优先级、schema 降级和 provenance 的意见，在当前 head 上已经修复或失效。

目前只剩两个非阻塞 P2：展示类事实也会参与连接验证 fingerprint，可能不必要地清除有效验证；以及 provider inventory 恰好达到支持的 2,048 项后再加入一个已启用的 override-only model，会生成协议不接受的 index 2,048。

另外有几组可以明确删除的代码：完全没有被 reconciler 消费的 `factBackedModelIds` 管线、只有测试使用的第二条 raw override projection、未使用的 store wrapper/fingerprint、只服务测试的生产注入 seam，以及当前没有 Host/IPC 消费者的 model-facts writer facade。两张没有在 PR 正文引用的截图也应转为 PR 证据并从仓库删除。

这些删除不会改变用户可见的 model-facts 行为，只会移除平行路径和未使用接口。把它们视为非阻塞清理后，我批准当前方案。

</details>

packages/core/src/model-facts.ts
```diff
@@ -0,0 +1,321 @@
...
+      lookupModelFactOverride(overrides, connection.providerType, model.id),
+    ),
+  );
+  const existing = new Set(models.map((model) => model.id));
+  const enabled = new Set(
+    connection.enabledModelIds ??
+      (connection.defaultModel === undefined ? [] : [connection.defaultModel]),
+  );
+  for (const modelId of enabled) {
+    if (models.length >= CONNECTION_CATALOG_MAX_MODELS_PER_CONNECTION) break;
```
Astro-Han: [P2] (category ① — supported catalog boundary)

Thanks for keeping the wire projection within the 2,048-model bound. Breaking as soon as the provider inventory already fills that bound means an enabled/default override-only model is never appended—the new test even fixes `custom-model` as absent. That model can still be selected through `enabledModelIds`, but `resolveExecutionConnection()` and the context-budget path consume this same projected list, so its user-provided `contextWindow` and other facts silently fall back to metadata/defaults. Could the bounded projection reserve space for enabled/default facts-backed models first, then fill the remaining slots with non-enabled provider rows, while leaving the raw provider inventory untouched? A regression should assert the result stays ≤2,048, the enabled custom model survives, and execution receives its override. Please feel free to push back if selection itself is rejected elsewhere whenever the model is beyond the provider projection.

Astro-Han (commented):
Thanks for simplifying the model-fact projection and removing the dead raw-catalog path. One supported catalog-boundary case still drops the override that execution needs. This is a suggestion from an outside review, so please do push back if enabled models are guaranteed to be present in the first 2,048 provider rows.

AI-assisted review disclosure: Codex ran independent model-authority and production/test analysis lanes; Astro-Han is the contributor of record for this review.

packages/runtime-host/src/protocol/runtime-policy.ts
```diff
@@ -693,6 +720,32 @@ function catalogPageItem(value: unknown): ConnectionCatalogPageItem {
   };
 }
 
+/** Decode catalog-only read-time provenance without admitting it to persisted models. */
+function decodeProjectedCatalogModel(value: unknown): ConnectionModel {
+  const item = requireShapedRecord(
+    value,
+    'projected connection model',
+    ['id'],
+    [...CONNECTION_MODEL_FIELDS.slice(1), 'factOverriddenFields'],
```
Astro-Han, line 729: Thanks for keeping this provenance read-only. Because this adds `factOverriddenFields` to the catalog wire shape, the compatibility epoch must fence older strict decoders before catalog admission. The same head instead moves `RUNTIME_HOST_COMPATIBILITY_EPOCH` backwards from the base’s 68 to 67 (current main is later), so an old epoch-67 Client can complete the handshake and then reject this field through `decodeConnectionModel`. The exact-head test is already failing on that backward epoch transition. I believe this is P1 on a category ② mixed-version/upgrade path: please rebase while preserving epoch history, assign this wire addition an epoch after current main, and cover old-Client/new-Host handshake + catalog at the gate. Please push back if another verified admission boundary excludes those peers.

Astro-Han (commented):
Thanks for the substantial simplification in this update. The exact-head test failure exposes one mixed-version protocol boundary that should be corrected before the branch is rebased forward. This is an AI-assisted review; I independently checked the wire decoder, epoch history, and current CI failure. Please do push back if the compatibility contract differs from what the strict decoder implies.

hqhq1025 (commented):
Reviewed exact head `4185b1efb07821822e8eef01736003cc34a4bb77`. **One P1 remains, so this is not merge-ready.**

The unresolved current-head thread at `packages/runtime-host/src/protocol/runtime-policy.ts:729` is valid. This head adds `factOverriddenFields` to the strict catalog model wire shape while declaring compatibility epoch 77. Current `main` is already epoch 78, and its epoch 77 represents a different wire change (`sessionTitle`). An epoch-77 Client from that history can therefore pass the handshake with this Host and then reject the projected model as `connection model contains unknown field 'factOverriddenFields'`. Rebase onto current `main`, preserve its epoch history, assign this change a new epoch after 78, and add an old-Client/new-Host handshake-plus-catalog regression. I did not duplicate the existing inline comment.

The previous 2,048-model P2 is fixed on this head: the projection reserves slots for enabled facts-backed models and evicts only unenabled provider rows (`packages/core/src/model-facts.ts:303-325`); both Core and Storage regressions verify a bounded 2,048-item catalog and execution visibility.

Scope checked: the complete 25-file diff, facts schema/read path, provider-to-user projection, model refresh, verification invalidation, execution context-budget precedence, strict Runtime Host paging codec, existing review threads, and current-main integration. Clean install, `npm run build:test`, full workspace typecheck, 46 focused Core/Runtime tests, 66 focused Storage tests, 78 focused Runtime Host tests, lint, format, ASF-header audit, and `git diff --check` passed. GitHub currently reports no hosted checks, and the branch is `CONFLICTING/DIRTY` against current `main` with conflicts in the protocol epoch history and document I/O.

Unable to verify until the rebase: the final merged epoch value, mixed-version rejection at that new boundary, and hosted platform checks.

Nyvo-io:
Rebased onto current `main` and resolved the protocol epoch and catalog-boundary requirements in one reviewed update.

All required checks are green: `test`, `audit`, and `windows_recovery`; GitHub reports `mergeable_state: clean`. Local lint, format, ASF-header, dependency builds, typechecks, focused Core/Storage/Runtime Host tests, and the mixed-version protocol regression also pass. No further code changes were made after review. Ready for maintainer review.

PR #3133, opened 2026-08-17
fix(runtime-host): page Session traces at the source

Astro-Han:
## Summary

- Page Session traces at the Storage source instead of materializing the complete Session inside the live Runtime Host. Host pages fill with complete AgentRuns up to the existing evidence, turn, and result budgets; one oversized run degrades to explicit partial coverage without blocking older history.
- Keep the Inspector's cost and cache summary independent of timeline pagination. Canonical attempts, legacy telemetry, pending repair, and unreadable provenance are queried by indexed `session_id` before decoding, so the summary remains Session-wide while the visible timeline stays bounded.
- Publish Session-scoped Usage invalidations from the Usage authority after durable writes. Runtime Host coalesces the `usage` domain, and Desktop subscribes synchronously through the current Host epoch instead of inferring cost changes from Session events.
- Load only the newest bounded timeline page initially. Each "Load earlier records" action reads exactly one continuation page and appends it; live refreshes rebuild only the page depth the user has requested so inserted or late evidence cannot leave a stale window.
- Use `(runId, turnId)` as the trace identity throughout projection, coverage, merge, and rendering. Distinguish oversized evidence from corrupt/unreadable evidence, and treat either known gap as partial rather than claiming the backend records no call details.
- Remove timeline search/filter state that would imply a complete in-memory dataset. Keep stable timestamps, visible initial/loading states, and explicit estimated/incomplete copy.
- Advance the Runtime Host compatibility epoch to 32 for the breaking trace continuation, Usage query, and Session-domain contract changes.

This changes user-visible Inspector behavior: long Sessions open normally, summary figures remain Session-wide estimates, each click loads one bounded earlier page, and incomplete accounting is never presented as a known zero or a complete timeline.

## Screenshots

The same 944-record, 1.18 MB Session is shown before and after the change.

| Before | After |
| --- | --- |
| Before: aggregate evidence limit error | After: Session-wide summary and paged timeline |

Additional Storybook verification covers the continuation control above the ascending timeline, one-page loading, stable timestamp labels, and explicit incomplete coverage. A clean isolated run produced no console warnings or errors.

## Verification

- Confirmed the predicted pre-fix failures for composite Run/Turn identity, oversized-vs-unreadable coverage, corrupt model-call evidence, Usage-authority refresh, one-page continuation reads, stale summaries, mutable page windows, and compatibility epoch; all pass after their owning-layer corrections.
- Focused Core, Runtime, Storage, Runtime Host, and Desktop suites: 139 tests passed.
- Biome format/lint over every changed TS/TSX file
- Storybook `TraceMoreHistory`: click-to-load appends the older page and a fresh isolated page reports no console warnings or errors.
- Repository-wide tests were not run locally; CI owns full coverage.

## Review

- Three independent AI reviewer passes examined the final rebased head: trace/pagination contracts, the end-to-end Usage invalidation path, and holistic architecture/test entropy.
- Final gate: no P0–P3 findings; all three reviewers returned GO.
- AI review is advisory and does not replace the required human contributor review before merge.

## AI use

Select exactly one:

Tool(s) and scope: Codex diagnosed the failure, shaped and authored the implementation and tests, adjudicated adversarial reviews, resolved the latest-main rebase, and performed focused and visual verification. Claude Fable provided earlier read-only architecture and code reviews. The human contributor must review the final diff, screenshots, and commit messages before merge.

## Checklist

- [ ] No


Astro-Han:
/agentic_review

Automated request by Codex on behalf of @Astro-Han to verify the newly installed Qodo OSS review integration.

packages/storage/src/agent-run-store.ts
```diff
@@ -416,6 +432,67 @@ class SqliteAgentRunStore implements DurableAgentRunStore {
     return { runs, truncated };
   }
 
+  async listSessionRunsPage(
+    sessionId: string,
+    input: AgentRunPageInput,
+  ): Promise<AgentRunPageResult> {
+    assertSafeId(sessionId, 'Invalid session id');
+    assertIdentitySearchLimit(input.limit);
+    if (input.before) assertSafeId(input.before.runId, 'Invalid AgentRun page cursor');
```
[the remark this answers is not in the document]
Astro-Han: Automated follow-up by Codex on behalf of @Astro-Han. Addressed in the current head: `listSessionRunsPage` rejects a non-finite `before.createdAt` before issuing SQL, and the focused storage regression test covers that cursor contract. Resolving as fixed.

packages/core/src/session-trace.ts
```diff
@@ -616,3 +630,96 @@ export function mergeTraceTotals(base: TraceTotals, next: TraceTotals): TraceTot
...
+  if (!first) throw new Error('At least one Session trace page is required');
+  return traces.slice(1).reduce((current, page) => {
+    if (page.schemaVersion !== current.schemaVersion || page.sessionId !== current.sessionId) {
+      throw new Error('Session trace pages do not describe the same Session');
+    }
+    const turns = new Map(current.turns.map((turn) => [traceTurnIdentityKey(turn), turn] as const));
+    for (const turn of page.turns) turns.set(traceTurnIdentityKey(turn), turn);
+    const ordered = [...turns.values()].sort(
+      (left, right) => left.startedAt - right.startedAt || left.runId.localeCompare(right.runId),
+    );
```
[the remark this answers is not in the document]
Astro-Han: Automated follow-up by Codex on behalf of @Astro-Han. Addressed in the current head: `mergeSessionTraces` now defines a total order by `(startedAt, runId, turnId)`, with a regression test covering tied timestamps and identities. Resolving as fixed.

packages/core/src/model-call-usage-projection.ts
```diff
@@ -108,12 +109,16 @@ function tokens(attempt: ModelCallAttempt): {
...
-  const cacheRead = attempt.cacheReadInputTokens ?? 0;
+  const cacheRead = clampCacheReadTokens(input, attempt.cacheReadInputTokens ?? 0);
   const cacheWrite = attempt.cacheWriteInputTokens ?? 0;
   const reasoning = attempt.reasoningTokens ?? 0;
   return { input, output, cacheMiss, cacheRead, cacheWrite, reasoning, total: input + output };
 }
 
+export function clampCacheReadTokens(inputTokens: number, cacheReadTokens: number): number {
+  return Math.min(cacheReadTokens, inputTokens);
+}
```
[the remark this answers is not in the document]
Astro-Han: Automated follow-up by Codex on behalf of @Astro-Han. Deferred intentionally for this PR. The clamp enforces valid accounting/display bounds; adding anomaly counters or coverage would introduce a separate diagnostics contract that is not required for source-level Session trace pagination. The source usage records remain authoritative for future diagnostics work. Resolving this thread as an acknowledged out-of-scope observability enhancement, not a correctness defect in this PR.

Astro-Han:

M4n5ter (commented):
<details open>
<summary>English</summary>

**[Important] Preserve supported cache-only provider evidence**

This is stronger than the observability concern originally raised in this thread. `strictProviderRequestUsage` deliberately preserves `cacheReadInputTokens` when the raw provider payload contains cache evidence but no trustworthy input total; the existing `provider-request-telemetry.test.ts` contract explicitly covers `{ cacheReadInputTokens: 10 }` without `inputTokens`.

In `model-call-usage-projection.ts`, the missing input total becomes `0`, so `clampCacheReadTokens` rewrites that valid cache evidence from `10` to `0`. The same projection feeds summaries, buckets, and canonical Usage logs, and the request also disappears from `cacheHitRequests`. This is therefore data loss on a supported provider path, not merely missing diagnostics for malformed `cacheRead > input` data.

Please preserve authoritative token fields in Usage aggregates and logs. If the displayed cache-hit ratio needs a trustworthy denominator or a `[0, 1]` bound, derive or clamp that presentation metric separately—or leave it unavailable when its basis is incomplete. Please add the existing cache-only provider shape as an end-to-end Usage projection regression.

**Non-blocking simplification**

Now that Session-wide cost and cache figures come from the Usage authority, `TraceTotals` has effectively lost its consumers: `SessionTrace.totals` has none, and only `TurnTrace.totals.costUsd` is read in production, where it can be derived from the existing model-call steps. Removing this aggregate contract would eliminate its projection, validation, wire, merge, and fixture maintenance without changing the Inspector timeline or per-turn cost.

The remaining findings are follow-up-level only: corrupt immutable RuntimeEvent rows can still pin trace pagination; malformed pre-v4 Usage rows may remain unattributed after migration; and reprojection markers have a narrow ABA window requiring an additional projection failure or crash. None of these should block this PR.

Apart from the cache-evidence loss above, the source pagination, composite identities, Desktop page-window state, Usage invalidation lifecycle, and integration with current `main` look sound. Please also update the PR body’s compatibility epoch from `23` to the final `28`.

</details>

<details>
<summary>简体中文</summary>

**[Important] 保留受支持的、只有 cache evidence 的 provider 数据**

这个问题比本 thread 最初提出的 observability 风险更严重。原始 provider payload 提供缓存证据、但没有可信 input total 时，`strictProviderRequestUsage` 会刻意保留 `cacheReadInputTokens`；现有 `provider-request-telemetry.test.ts` 契约也明确覆盖了没有 `inputTokens`、只有 `{ cacheReadInputTokens: 10 }` 的输入。

在 `model-call-usage-projection.ts` 中，缺失的 input total 会先变成 `0`，随后 `clampCacheReadTokens` 把有效的 cache evidence 从 `10` 改写为 `0`。同一投影还用于 summary、bucket 和 canonical Usage log，并导致该请求不再计入 `cacheHitRequests`。因此，这不是 malformed `cacheRead > input` 数据缺少诊断信号，而是受支持 provider 路径上的数据丢失。

请在 Usage aggregate 和 log 中保留权威 token 字段。如果展示层的 cache-hit ratio 需要可信分母或限制在 `[0, 1]`，应单独派生或 clamp 该展示指标；basis 不完整时也可以不展示比例。建议将已有的 cache-only provider shape 补成端到端 Usage projection 回归测试。

**非阻塞简化建议**

Session-wide cost 和 cache 数据改由 Usage authority 提供后，`TraceTotals` 已基本失去消费方：`SessionTrace.totals` 没有生产消费者，`TurnTrace.totals` 也只有 `costUsd` 被读取，而该值可以从现有 model-call steps 派生。删除这套 aggregate contract，可以一并去掉相关 projection、validator、wire、merge 和 fixture 维护，同时不改变 Inspector timeline 或每轮费用展示。

其余问题仅建议作为 follow-up：损坏的 immutable RuntimeEvent 仍可能卡住 trace pagination；迁移后无法归属的损坏 pre-v4 Usage row 会退出 Session provenance；reprojection marker 也存在一个还需叠加投影失败或进程退出才会造成影响的窄 ABA 窗口。这些都不应阻塞当前 PR。

除上述 cache evidence 丢失外，source pagination、复合身份、Desktop page-window 状态机、Usage invalidation 生命周期以及与最新 `main` 的集成都没有发现实质问题。PR body 中的 compatibility epoch 也请从 `23` 更新为最终的 `28`。

</details>

packages/storage/src/usage-stores.ts
```diff
@@ -399,18 +424,24 @@ function createWriterFacade(
...
+        admitSessionUsageMutation(record.sessionId, () => telemetry.insertLlmCall(record)),
       recordToolInvocation: (record) =>
         admit(() => run(() => telemetry.insertToolInvocation(record))),
     },
     modelCalls: {
-      modelCallAttempts: (range) => read(() => modelCalls.read(range)),
-      recordModelCallAttempt: (attempt) => admit(() => run(() => modelCalls.record(attempt))),
+      modelCallAttempts: (range, sessionId) => read(() => modelCalls.read(range, sessionId)),
+      recordModelCallAttempt: (attempt) =>
+        admitSessionUsageMutation(attempt.sessionId, () => modelCalls.record(attempt)),
```
hqhq1025: **P1: Do not publish an invalidation for every idempotent repair replay.** `readCanonicalUsage()` runs pending repair from the `usage.query` read path. If a marked run contains one valid attempt plus one undecodable authority event, repair replays the valid attempt but keeps the marker. This facade then publishes `usage:changed` after every identical upsert, and the Desktop subscriber schedules another summary read after 400 ms, creating a self-sustaining read -> upsert -> invalidation loop while the Inspector is open. On this exact head, two identical reads increased the notification count from 1 to 2 while `attempts=1`, `unreadableRecords=1`, `pendingRepairs=1`, and the marker remained. Publish only when the projected row actually changes (or use a non-publishing repair path), and add a regression proving that a second unchanged repair emits no new invalidation.

apps/desktop/src/renderer/session-inspector-overview-model.ts
```diff
@@ -125,18 +121,40 @@ export interface InspectorOverviewModel {
...
+  return summary.provenance.legacyRecords > 0 && summary.totalCostUsd > 0
+    ? summary.totalCostUsd
+    : undefined;
+}
+
+export function hasUnavailableSessionUsage(
+  summary: UsageSummaryV2 & { readonly provenance: UsageProvenance },
+): boolean {
+  return (
+    summary.totalRequests === 0 &&
```
hqhq1025: **P2: Surface known partial Session summaries even when some requests decoded.** This predicate reports a provenance gap only when `totalRequests === 0`. With one valid request plus `unreadableRecords > 0` or `pendingRepairs > 0`, the backend explicitly says the summary is incomplete, but the panel suppresses the current-session warning and renders the cost/cache figures through the normal path. Split the zero-data unavailable state from a partial/incomplete state, and cover a nonzero-request summary with a provenance gap.

hqhq1025 (changes requested):
Two new findings are attached inline. The P1 is merge-blocking.

Astro-Han:
@hqhq1025 @M4n5ter — re-review requested on the rebased head `7b69b81fc`.

The latest head addresses the three confirmed findings:

- preserves provider-reported cache-only evidence without inventing an input-token denominator; cache-hit rate stays unavailable for partial or missing usage;
- publishes Session Usage invalidations only when the underlying model-call projection or repair marker actually changes, preventing repeated repair refresh loops;
- reports incomplete Session Usage provenance even when some valid requests are present.

The `TraceTotals` simplification was deliberately kept out of this PR and is tracked separately in #3389.

Focused Core, Storage, Runtime Host, and Desktop checks pass, including 49 affected tests, Desktop typecheck, Biome, and `git diff --check`. Two independent Codex reviewer passes returned GO with no P0–P3 findings; these are advisory and do not replace human review. CI is currently running on the new head.

This update and comment were prepared with Codex assistance.

packages/storage/src/model-call-ledger.ts
```diff
@@ -118,75 +123,106 @@ class SqliteModelCallLedger implements ModelCallLedger {
...
-      this.#lease.database
+      const changed = this.#lease.database
         .prepare(`
           INSERT INTO usage_model_call_reprojection(session_id, run_id, marked_at)
           VALUES (?, ?, ?)
           ON CONFLICT(session_id, run_id) DO NOTHING
         `)
-        .run(sessionId, runId, Date.now());
+        .run(sessionId, runId, Date.now()).changes;
+      return changed > 0;
```
hqhq1025: [P2] Preserve a newer repair intent when this marker already exists

Returning `false` here conflates an idempotent replay with a newer attempt reusing an existing run marker. A repair can snapshot the old AgentRun events, a later authority append can hit the existing marker and make this mark a no-op, and the stale repair then clears that only marker. If the later Usage projection fails or the process exits before it, the new billed call remains only in AgentRun while `pendingReprojections()` is empty; no implemented full-stream sweep rediscovers it. I reproduced exactly `authority=[old,new]`, `stored=[old]`, `pending=[]`. Please version the marker and CAS-clear the generation read by repair, or use per-attempt repair intents.

hqhq1025 (changes requested):
One merge-blocking finding is attached inline.

Problem and mechanism: this PR correctly addresses unbounded Session inspection by paging AgentRuns at Storage, while keeping Session-wide Usage as an independent authority and refreshing it through Session-scoped invalidations. The problem definition is supported by the prior all-at-once materialization path, and source paging is the right first-principles boundary.

First principles and Occam: the pagination, composite trace identity, and independent Usage summary are materially simpler than loading an entire Session and inferring totals from a partial timeline. The previous invalidation loop and partial-summary presentation findings are fixed on this head. However, the repair intent is still represented by one unversioned row per run. That representation cannot distinguish an idempotent replay from a newer authority append while an older repair is in flight, so an old repair can delete the only durable recovery signal for a new billed call.

Optimality/refactor: preserve marker generations and compare-and-swap clear the exact generation observed by repair, or use per-attempt repair intents. The current insert/no-op plus unconditional clear is not a maintainable final state. Add a deterministic concurrency regression covering old repair snapshot -> new authority append/mark -> crash before projection -> old repair clear.

Deletion opportunities: SessionTrace.totals appears to have no Session-level production consumer after Usage became authoritative; removing that projection/wire/merge surface is a reasonable follow-up, not a blocker. I did not find a low-quality test that must be deleted for this fix; the missing test is the concurrent marker-generation case above.

Merge verdict: not ready to merge. Residual non-blocking gaps include corrupt persisted AgentRun/RuntimeEvent rows pinning pagination, the preload summary method not uniformly wrapping rejected promises in its declared Result envelope, and CI still running at review time.

Verification: reviewed head 7b69b81fc53d96deee1f685e1d0f6ca028a2070a; 164 focused Core, Storage, Runtime, Runtime Host, and Desktop tests passed; the full affected build completed after applying repository dependency patches (one initial sequential workspace resolution retry was required).

Astro-Han:
Rebased onto the latest `main` and replaced the independent Usage repair-marker lifecycle with a sequence-derived projection checkpoint.

The new invariant is: a run is pending exactly while its canonical AgentRun model-call high-water exceeds the Usage checkpoint. Projection rows and checkpoint advancement commit atomically, so the ABA race and the authority-append-before-marker crash window are both removed. A rebuildable run-level high-water index, maintained in the AgentRun append transaction and backfilled on migration, keeps lagging-run discovery from rescanning the full event history.

Current head: `6f52b7b59`

Validation:
- Storage workspace: 861 passed, 14 skipped, 0 failed
- Runtime Host / trace / protocol focused tests: 70 passed, 0 failed
- Two independent adversarial reviews: GO, no P0-P3 findings
- Full affected build and formatting checks passed

@hqhq1025, could you please re-review the updated approach and confirm whether the original race is resolved?

_Posted by Codex on behalf of the contributor._

Astro-Han:
Rebased again onto the latest `main` (`cabd68fb7`, including #3406) after it advanced during the previous CI run. The rebase was conflict-free and does not change the PR design; the new head is `1f509871d`.

The prior exact-head CI was fully green, including Desktop E2E, Storybook smoke, and installed CLI validation. Fresh CI is now running on the rebased head. @hqhq1025, when it is convenient, please re-review the sequence-derived catch-up architecture and dismiss the superseded changes-requested review if the current head addresses the findings.

_Posted by Codex on behalf of the contributor._

packages/storage/src/usage-stores.ts
Astro-Han: Addressed on the current head. The marker replay path no longer exists: catch-up reads only AgentRun model-call events beyond the durable per-run checkpoint, then advances that checkpoint in the same transaction as the projection writes. A subsequent unchanged catch-up finds no lagging run and emits no usage invalidation. The regression `does not republish idempotent model-usage mutations` covers two consecutive catch-ups and observes exactly one notification. The current exact-head CI is running after a conflict-free rebase; the previous exact-head CI was fully green.

apps/desktop/src/renderer/session-inspector-overview-model.ts
Astro-Han: Addressed in `f0d2bfc38`. `hasUnavailableSessionUsage()` now reports a provenance gap whenever `unreadableRecords > 0` or `pendingRepairs > 0`, independently of `totalRequests`. The regression asserts both the zero-request case and the same summary with `totalRequests: 1`, so valid totals can remain visible while the panel still marks the summary as incomplete.

packages/storage/src/model-call-ledger.ts
Astro-Han: Fixed at the authority owner rather than adding CAS semantics to the marker. The reprojection marker table and its mark/pending/clear APIs were removed. Pending work is now derived as `core_agent_runs.latest_model_call_sequence > usage_model_call_projection_checkpoints.applied_through_sequence`; AgentRun append advances the rebuildable source high-water in the same transaction as the canonical event, while projection writes and checkpoint advancement share another transaction. Therefore a later append always remains above the earlier checkpoint and is rediscovered without a second repair authority. Regressions cover a later authority append after an earlier catch-up, bounded catch-up, corrupt-event progress, and projection-failure rollback.

hqhq1025 (dismissed):
Re-reviewed the current head against the previous repair-intent race. No actionable findings remain.

Problem and mechanism: the PR correctly fixes unbounded Session inspection with source-level AgentRun pagination and keeps Session-wide accounting independent from timeline pages. The latest revision removes the independent reprojection marker entirely. AgentRun now publishes a durable per-run model-call sequence high-water in the same transaction as the authority event, while Usage advances an applied-through checkpoint in the same transaction as the projection writes. A committed event therefore remains discoverable until projection succeeds, without a second marker lifecycle that can suffer ABA deletion.

First principles and Occam: deriving projection lag from authority sequence minus applied checkpoint is the correct and simpler final model. It closes both the previously reproduced old/new marker race and the older crash window between authority append and marker creation. Corrupt authority evidence is counted while the checkpoint advances, and projection-storage failures roll back rather than skipping a valid billed call.

Optimality, deletion, and tests: replacing the marker table and its repair API removes accidental state rather than layering another generation protocol on top. I found no additional production code or low-value test that must be deleted, and no deeper refactor is required. The focused regressions cover real append high-water publication, marker-free recovery, later appends, unreadable evidence, bounded catch-up, idempotent invalidation, and transaction rollback on projection failure.

Merge verdict: the reviewed code is ready to merge once the currently running required CI check completes successfully. Residual non-blocking risks remain the previously documented corrupt AgentRun/RuntimeEvent pagination behavior and the conservative bounded catch-up latency for very large backlogs.

Verification: reviewed head 1f509871d146b61a3b7c9cb72f70a90310013d25; npm run build:test passed; 133 of 134 focused tests passed in the combined run and the lone unrelated implementation-child patch test passed immediately in isolation; git diff --check passed. Required CI had completed build, typecheck, lint, formatting, and related prechecks and was still running affected workspace tests at publication time.

Astro-Han:
Rebased onto the latest `main` (`c1e02d8ec`, including #3382 and #3092). The only conflict was the Runtime Host compatibility epoch: current `main` now owns epoch 33 for bounded sandbox failure frames, so this PR now uses epoch 34 for its trace cursor and Session usage wire changes while preserving the epoch 33 history. The adjacent protocol assertion was updated and autosquashed into the original protocol commit.

New head: `f20354b2a`. Local validation after the rebase:
- `npm run build:test`
- `npm run format:check`
- 96 focused Runtime Host / Usage / Storage tests
- `git diff --check`

All passed. @hqhq1025, the previous approval was necessarily dismissed by the rebased head; please re-approve this exact head when convenient. Once the fresh required CI is green, the contributor has authorized squash merge.

_Posted by Codex on behalf of the contributor._

hqhq1025 (commented):
Re-review of f20354b2a2ec4d1e5e6c44248b7c44fab16c44e7

No new code finding was introduced by the latest revision. The range-diff against the previously reviewed head shows the Session paging, Usage authority, repair checkpoint, invalidation, and Desktop state commits are patch-equivalent after the rebase. The earlier repair-intent race remains fixed by the transactional AgentRun high-water and applied-through checkpoint model.

Problem and mechanism: this PR correctly moves long-Session trace pagination to the Storage source while keeping Session-wide Usage independent from the visible timeline. Composite `(runId, turnId)` identity, bounded continuation pages, explicit incomplete coverage, and Session-scoped Usage invalidation remain the right ownership boundaries.

First principles, optimality, and deletion: the final architecture derives repair lag from durable authority state instead of maintaining a second marker authority, which is both simpler and safer. I found no additional production code or low-value test that must be removed, and no deeper refactor is required.

Current merge gate: not ready to merge only because current `main` advanced after this head and GitHub now reports a content conflict in `packages/runtime-host/src/protocol/index.ts` and `packages/runtime-host/src/__tests__/protocol.test.ts`. Both branches independently consumed compatibility epoch 34. Rebase onto current main, preserve the epoch-34 ScheduledTask entry, move this PR's Session-trace protocol change to epoch 35, and update the exact epoch regression accordingly. The new protocol epoch guard is expected to reject a same-number resolution.

After that rebase, this needs a short exact-head check of the two conflict resolutions and CI rather than another full architectural review.

Verification: required CI is green on f20354b2a2ec4d1e5e6c44248b7c44fab16c44e7; `npm run build:test` and `git diff --check` pass locally; the range-diff confirms all 19 previously reviewed follow-up commits are equivalent apart from the base-driven epoch renumbering.

Astro-Han:
Hi @hqhq1025, @M4n5ter, and @likun666661 — could one of you please review the current head `58de960e1`?

This branch has been rebased onto current `main` (`e44ddcec5`). The only conflict was the Runtime Host compatibility epoch: `main` now owns epoch 34 for the ScheduledTask wire change, so this PR advances the trace cursor and Session usage wire changes to epoch 35 while preserving the protocol history.

Current-head validation is complete:

- required `CI / test`: passed
- `npm run build:test`
- `npm run format:check`
- protocol epoch guard: 8/8
- focused Runtime Host / Usage / Storage tests: 98/98
- `git diff --check`

All existing review threads are resolved. An approval from a committer other than the author is the final merge gate.

_Posted by Codex on behalf of the contributor._

hqhq1025 (approved):
Re-review of 58de960e11ed7049836888dd42f19f15a586a123

Problem and mechanism: the PR fixes unbounded Session inspection by paging AgentRuns at the Storage source, while keeping Session-wide Usage independent from the visible timeline and refreshing it through Session-scoped invalidations. Composite `(runId, turnId)` identity, bounded continuation pages, explicit incomplete coverage, and the transactional authority high-water/applied checkpoint model remain the correct ownership boundaries.

Latest rebase: the range-diff confirms all 20 PR commits are patch-equivalent to the previously reviewed revision except for the required protocol conflict resolution. Current main owns compatibility epoch 34 for backend-free ScheduledTask templates; this revision preserves that history and advances Session trace pagination and Usage wire changes to epoch 35. The exact epoch assertion and merge-result guard agree with that ordering.

First principles, optimality, deletion, and tests: deriving projection lag from durable authority sequence minus the applied checkpoint remains simpler and safer than a second repair-marker authority. I found no additional production code or low-value test that should be deleted, and no deeper refactor is required.

Merge verdict: ready to merge. Required CI is green, GitHub reports the head mergeable, `git diff --check` passes, the protocol epoch guard reports 34 -> 35, and all eight epoch-guard tests pass. The previously documented corrupt persisted trace-row behavior and bounded catch-up latency remain non-blocking residual risks.

PR #3135, opened 2026-08-17
feat(desktop): add Work Board Phase 1 capture/list MVP

somewan820:
## Summary

Work Board Phase 1 (capture/list MVP) from #2560, built on the merged Phase 0 contract and store ([#3028](https://github.com/maka-agent/maka-agent/pull/3028)).

Adds a compact Work Board tab to the session workbar:

- global Inbox and current-project filtering;
- manual create, rename, move (Inbox <-> project), complete / reopen, archive / restore, and delete;
- empty, loading, and error states;
- local-first persistence through the existing operational-state database.

Boundary: the Desktop main process owns `WorkBoardStore`; the renderer is a read-only IPC projection that reloads on the `workBoard:changed` signal. No Runtime Host involvement, no model-visible tools, no turn-tail injection. `linkedSessions` and the linked-session projection remain deferred to Phase 3.

Refs #2560

## Verification

- `@maka/desktop` main and preload builds pass
- `@maka/desktop` typecheck passes (preload / main / renderer / storybook)
- Work Board IPC tests pass (2/2)
- Full desktop test suite runs in CI; several local suites require storage-root permissions unavailable in the sandbox

## Checklist

- [x] Tests cover the change and fail without it (IPC and store layers)

- [ ] No

## AI use

Select exactly one:

Tool(s) and scope: Codex (OpenAI) — implementation, tests, and documentation for Work Board Phase 1; the contributor reviewed the output and owns the final result. Affected commits carry `Generated-by: Codex` trailers.


apps/desktop/src/renderer/session-workbar-tabs.ts
```diff
@@ -54,6 +55,7 @@ export interface PersistedSessionWorkbarPanels {
 const PERSISTED_KINDS = new Set<SessionWorkbarTabKind>([
   'review',
   'tasks',
+  'work-board',
```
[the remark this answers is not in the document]
somewan820: Fixed in `57dde789c`: `isSessionWorkbarTabKind` now accepts `work-board`, so persisted Work Board tabs restore after restart.

apps/desktop/src/renderer/work-board-panel.tsx
```diff
@@ -0,0 +1,341 @@
...
+  const archivedItems = board.items.filter((item) => item.archived);
+
+  const runAction = async (action: () => Promise<unknown>): Promise<void> => {
+    setActionError(undefined);
+    try {
+      await action();
+    } catch (error) {
+      setActionError(error instanceof Error ? error.message : copy.actionFailed);
+    }
+  };
```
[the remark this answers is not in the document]
somewan820: Fixed in `57dde789c`: `runAction` now returns whether the mutation succeeded, and create/rename drafts are cleared or closed only on success.

apps/desktop/src/renderer/work-board-panel.tsx
```diff
@@ -0,0 +1,341 @@
...
+          onClick={() => setFilter('inbox')}
+        />
+        <Button
+          size="sm"
+          variant={filter === 'project' ? 'primary' : 'ghost'}
+          label={copy.project}
+          onClick={() => setFilter('project')}
+          isDisabled={props.projectId === null}
+          tooltip={props.projectId === null ? copy.noProject : undefined}
+        />
```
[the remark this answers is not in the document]
somewan820: Fixed in `57dde789c`: removed the incomplete `tablist` role, and the section aria-label now follows the active filter.

docs/work-board-phase1.md
```diff
@@ -0,0 +1,28 @@
+# Work Board Phase 1 — capture/list MVP
+
+Status: in development (`feat/2560-work-board-phase1`).
```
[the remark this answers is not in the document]
somewan820: Fixed in `57dde789c`: removed the branch-specific status line from the Phase 1 doc.

somewan820:
Addressed the review round in `57dde789c`:

- CI: regenerated the Astryx surface inventory so `work-board-panel.tsx` and `work-board.css` are tracked (fixes the failing `astryx_surface` check).
- Inline findings: `isSessionWorkbarTabKind` accepts `work-board`; create/rename drafts survive failed mutations; incomplete `tablist` role removed; branch-specific doc status removed.
- Nitpicks: mutations now rely on the `workBoard:changed` signal as the single reload path (no duplicate list), and panel copy moved into `DesktopConversationCopy` so locale completeness stays enforced.

Verification: full desktop typecheck passes, main build + Work Board IPC tests pass, Biome clean.

Copilot could not review this round because the requesting account hit its review quota; the change will be re-checked once quota resets.

somewan820:
@liugddx — Phase 1 (Work Board capture/list MVP) from the #2560 delivery plan is ready for review. It builds on the merged Phase 0 contract/store ([#3028](https://github.com/maka-agent/maka-agent/pull/3028)) and adds the workbar tab with Inbox/current-project filtering, create/rename/move/complete/reopen/archive/restore/delete, and main-process IPC ownership.

CI and bot feedback have been addressed: Astryx surface inventory regenerated (failing check fixed), persisted tab-kind restore fixed, create/rename drafts survive failed mutations, accessibility cleaned up, and panel copy moved into `DesktopConversationCopy`. Desktop typecheck, main build, Work Board IPC tests, and Biome all pass.

Could you take a look when you have a moment? Happy to adjust anything.

<details>

@liugddx —— #2560 delivery plan 里的 Phase 1（Work Board capture/list MVP）已就绪，等待 review。它基于已合并的 Phase 0 契约/store（[#3028](https://github.com/maka-agent/maka-agent/pull/3028)），新增 workbar tab，支持 Inbox/当前项目过滤、新增/改名/移动/完成/重开/归档/恢复/删除，以及 main 进程 IPC 所有权。

CI 和机器人反馈已处理：Astryx surface inventory 已重新生成（失败的检查已修复）、持久化 tab-kind 恢复已修复、失败时不再清空新增/改名草稿、可访问性已清理、面板文案已并入 `DesktopConversationCopy`。desktop typecheck、main build、Work Board IPC 测试和 Biome 均通过。

有空的话麻烦看一下，需要调整的地方请告诉我。

</details>

liugddx (commented):
## Review — problem framing & scope

Solid, disciplined engineering. My comments are almost entirely about **how the problem is defined** (in #2560), not the code in this PR, which is clean.

### What it solves / how (my read, please correct if off)

- **Solves**: the "capture deferred work without interrupting the active task" atom from #2560 — Phase 1 (capture/list MVP).
- **How**: a read-only Work Board tab in the workbar; `WorkBoardStore` owned by the main process, renderer is a projection that reloads on `workBoard:changed`; 6 fail-closed IPC handlers with a Result type; scope/creator/provenance/revision model. Correctly avoids Runtime Host, model tools, and turn-tail injection.

Execution quality is high: Result types, optimistic `revision` locking, single reload path (no second execution authority), IPC-layer tests. 👍

### First-principles / Occam concerns on the *definition*

1. **The problem is named after the solution.** The irreducible need is *"don't let me lose this idea; let me start it later."* But #2560 defines it as a Work Board with Inbox/project scope + lifecycle + provenance + linked-Session projection. Those are names of the *answer*. This locks all later phases to a board shape before we've asked whether a much smaller entity would do.

2. **Occam — cheaper entities exist for the same atom.** For an Agent product, "write the deferred item into a project `TODO.md` / issue" satisfies most acceptance criteria in #2560 (local-first, survives restart, auditable, later Agent-readable) with near-zero new machinery. The Non-goals say "not a Linear/Jira replacement," yet the structure being built (board, scope, lifecycle, status projection) is a smaller-shaped skeleton of exactly that. Worth an explicit note on why a store + state machine is required over a file.

3. **Riskiest assumption is validated last.** The load-bearing bet — *will users actually return to the board and start tasks from it?* — isn't exercised until Phase 3. Front-loading the store/state-machine/provenance and back-loading that validation is the reverse of lean. Consider a cheap end-to-end spike of the capture→revisit→start-task loop before investing in Phases 2–4.

### Credit where due

The boundary discipline is genuinely first-principles and correct: **not** polluting the Session Task Ledger (#2290), **not** injecting into every model turn, **not** creating a second execution-state authority. That separation of *user intent* vs *model execution state* is the strongest part of the design and this PR honors it.

### Ask before merge/continuation

- One paragraph in #2560 (or the Phase-1 doc) on **why a dedicated store beats a project file** for the atom — if it's provenance + Session linking, say so explicitly; that's the actual justification for the machinery.
- Consider resequencing so the capture→start-task loop gets a thin validation before Phase 2–4 build-out.

**Net: Approve on execution; request a scope/justification note on the problem definition before committing further phases.**

<details>

工程执行扎实,我的意见几乎都针对 #2560 的**问题定义**,不是本 PR 的代码。

**解决了什么 / 怎么解的**:交付 #2560 的 Phase 1(捕获/列表 MVP)。主进程独占 `WorkBoardStore`,渲染进程只读投影、收到 `workBoard:changed` 后 reload;6 个 fail-closed IPC handler + Result 类型;scope/creator/provenance/revision 模型;刻意不进 Runtime Host、不暴露模型工具、不注入每轮 turn。质量高(乐观锁、单一 reload 路径、IPC 测试)。

**第一性原理 / 奥卡姆的疑问(针对定义)**:
1. **用解法命名了问题**。原子需求只是"别让我忘了,以后能启动";却被定义成带 scope/lifecycle/provenance/Session 关联的看板。这些是答案的名字,会把后续所有 phase 锁死在"看板"形态。
2. **奥卡姆——同一原子需求有更省的实体**。对 Agent 产品,"写进项目 `TODO.md`/issue"几乎零新实体,却能满足本地优先、重启存活、可审计、Agent 可读等大部分验收标准。Non-goals 说不做 Linear/Jira,但所建结构正是其更小骨架。建议明确说明为何需要 store + 状态机而非一个文件。
3. **最该验证的假设放到最后**。"用户真会回来看看板并启动任务吗"直到 Phase 3 才触及。建议在 Phase 2-4 前,先廉价打通"捕获→回看→启动任务"闭环做验证。

**值得肯定**:边界划得非常清醒且符合第一性——不污染 Session Task Ledger(#2290)、不注入每轮上下文、不做第二套执行权威。这是设计最强的部分,本 PR 也严格遵守。

**合并/继续前建议**:在 #2560 或 Phase-1 文档补一段"为何用专用 store 而非项目文件"的理由(若是 provenance + Session 关联,请明说);并考虑重排顺序,先验证核心闭环再铺 Phase 2-4。

**结论:执行层面 Approve;在继续后续 phase 前,请补充问题定义的范围/理由说明。**

</details>

somewan820:
@liugddx — thanks for the review. Both asks are addressed in `6d261ee20`:

- **Why a dedicated store instead of a project file**: added to `docs/work-board-phase1.md`. A `TODO.md` / issue would cover the literal capture-and-list atom, but the #2560 acceptance criteria also require typed provenance + a bounded excerpt that survives side-chat fork deletion, stable per-item identity + revision CAS for concurrent Desktop writes, and later Session linking / result refs. Those are the load-bearing reasons for the store shape; if they were not in scope, a project file would indeed suffice.
- **Sequencing**: agreed. The doc now records the plan to validate a thin capture -> revisit -> start-as-task loop before expanding Phases 2 and 4.

Happy to adjust the wording if you would like the rationale stated differently.

<details>

@liugddx —— 感谢 review。两点已在 `6d261ee20` 处理：

- **为什么用专用 store 而不是项目文件**：已加入 `docs/work-board-phase1.md`。`TODO.md` / issue 能满足字面上的捕获与列表原子需求，但 #2560 的验收标准还要求强类型来源引用 + 在侧栏 fork 删除后仍存留的有界 excerpt、并发 Desktop 写入下稳定的逐项身份 + revision CAS，以及后续的 Session 关联 / result refs。这些才是 store 形态的承重理由；如果这些不在范围内，项目文件确实够用。
- **顺序安排**：同意。文档已记录计划：在铺开 Phase 2/4 之前，先用一条 thin 的 capture → 回看 → start-as-task 闭环做验证。

如果你希望这段 rationale 换个措辞，告诉我即可。

</details>

liugddx:
## Follow-up: concrete next steps (actionable)

My earlier comment was framing/critique. Here is what I'm actually asking for, as a checklist. **This PR is approvable as-is** — items below are gates on *continuing to Phase 2–4*, plus two tiny things to land with this PR.

### Land with this PR (small)

- [ ] **Add a "Why a store, not a file" note** (3–5 sentences) to `docs/work-board-phase1.md`. State the one thing that justifies the machinery over a project `TODO.md`: it's provenance + Session linking (Phase 3). If that's the reason, say it explicitly so the scope reads as intentional, not accidental.
- [ ] **Write down the assumption we're betting on**, in the same doc: *"Users will return to the board and start tasks from it."* One sentence. This becomes the thing Phase 3 must prove.

### Gate before Phase 2 (side-chat capture)

- [ ] **Do a thin Phase 3 spike FIRST, before Phase 2.** Wire one hard-coded item → "Start task" → new Session → link back. No polish. Goal: prove the capture→revisit→start loop has real pull. If nobody uses it, we stop here and the store stays a simple list.
- [ ] Put the spike behind a flag; it doesn't need to ship. It needs to answer "does the loop get used."

### Then resume the planned order

- [ ] Only after the spike shows pull: build Phase 2 (side-chat capture) and Phase 4 (evidence/refinement) as originally planned in #2560.

### What NOT to change (keep doing this)

- Keep the store in the main process as the single mutation authority.
- Keep the renderer read-only / reload-on-signal.
- Keep Work Board out of the Session Task Ledger, out of model turns, out of Runtime authority. This boundary is correct — don't soften it under any Phase.

**TL;DR for the maintainer:** merge this; add the two doc notes; then **build the Phase 3 "Start task" spike before Phase 2** to validate the loop; then continue #2560's plan unchanged.

<details>

上一条是框架性评论,这条是给你的**可执行清单**。本 PR 可以直接合并;下面是"继续做 Phase 2-4"的前置门槛,外加两个随本 PR 落地的小项。

**随本 PR 落地(小)**
- [ ] 在 `docs/work-board-phase1.md` 补 3-5 句"为何用 store 而非文件":唯一能撑起这套机制的理由是 provenance + Session 关联(Phase 3),请明说,让范围显得是有意为之。
- [ ] 同一文档写下我们在赌的假设:"用户会回到看板并从中启动任务。" 一句话,作为 Phase 3 必须验证的目标。

**Phase 2 之前的门槛**
- [ ] **先做一个极薄的 Phase 3 spike,插在 Phase 2 之前**:硬编码一个事项 → "开始任务" → 新 Session → 关联回来。不做打磨。目的:验证"捕获→回看→启动"闭环真有人用。若没人用,就停在这里,store 保持简单列表即可。
- [ ] spike 放在 flag 后,不必上线,只需回答"闭环有没有被用起来"。

**恢复既定顺序**
- [ ] 仅当 spike 证明有需求后:再按 #2560 原计划做 Phase 2(侧边栏捕获)和 Phase 4(证据/完善)。

**不要改(继续保持)**
- store 留在主进程,作为唯一写入权威;渲染进程只读、收信号 reload;Work Board 不进 Session Task Ledger、不进模型每轮上下文、不做 Runtime 权威。这条边界是对的,任何 phase 都别放松。

**一句话给维护者:** 合这个 PR;补两条文档;**在 Phase 2 之前先做 Phase 3 "开始任务" spike** 验证闭环;然后按 #2560 原计划继续。

</details>

somewan820:
@liugddx — checklist items are landed in `f47d56a68`:

- **Why a store, not a file**: `docs/work-board-phase1.md` now states in a few sentences that the one justification is provenance + Session linking (typed source refs / bounded excerpt surviving fork deletion, Phase 3 linking an item to the Session it starts), with stable identity + CAS for concurrent writers; if those were not in scope, a project file would suffice.
- **Assumption**: the doc records the bet — “users will return to the board and start tasks from it” — as the thing Phase 3 must prove.
- **Sequencing**: Phases 2 and 4 are gated behind a thin, flag-gated Phase 3 spike (hard-coded item -> “Start task” -> new Session -> link back, no polish).

The merge conflict with `main` is resolved by merging `origin/main` into this branch (`3eacc39a7`); the only conflict was the regenerated Astryx surface inventory. Desktop typecheck, main build, and Work Board IPC tests pass. The PR should now be mergeable.

<details>

@liugddx —— 清单项已在 `f47d56a68` 落地：

- **为什么用 store 而不是文件**：`docs/work-board-phase1.md` 现在用几句话明确：唯一撑起这套机制的理由是 provenance + Session 关联（side-chat 捕获保留强类型来源引用 / fork 删除后仍存的有界 excerpt，Phase 3 把看板事项关联到它启动的 Session），加上并发写入下的稳定身份 + CAS；如果这些不在范围内，项目文件确实够用。
- **假设**：文档记录了赌注——“用户会回到看板并从中启动任务”——作为 Phase 3 必须验证的目标。
- **顺序**：Phase 2 和 Phase 4 现在被一个薄的、flag 控制的 Phase 3 spike 门槛卡住（硬编码事项 -> “开始任务” -> 新 Session -> 关联回来，不做打磨）。

与 `main` 的合并冲突已通过把 `origin/main` 合入本分支解决（`3eacc39a7`）；唯一冲突是重新生成的 Astryx surface inventory。desktop typecheck、main build 和 Work Board IPC 测试均通过，PR 现在应该可以合并了。

</details>

liugddx (approved):
Thanks Quinn — the CAS + fork-surviving excerpt + Session linking is a fair reason a flat `TODO.md` can't cover, so the store shape reads as intentional now. Nice, disciplined boundary work too.

Approving. One thing to hold onto for later: before we build out Phase 2/4, let's land the thin capture → revisit → start-as-task loop first and confirm people actually come back to the board — as the doc now notes. No changes needed here.

<details>

谢谢 Quinn —— CAS + fork 删除后仍存留的 excerpt + Session 关联,确实是 `TODO.md` 覆盖不了的,现在这套 store 的范围读起来是有意为之的。边界也做得很克制,赞。

Approve。后续记一个点:在铺开 Phase 2/4 之前,先把 thin 的 捕获 → 回看 → 启动任务 闭环落地,确认用户真的会回到看板——正如文档现在所记。本 PR 无需再改。

</details>

somewan820:
@Astro-Han — could you take a quick look at this one when you have a moment? Status:

- Phase 1 (Work Board capture/list MVP) from #2560, built on the merged Phase 0 contract/store (#3028).
- liugddx has already **Approved** and confirmed the store rationale (provenance + Session linking) and the boundary discipline.
- All CI checks are green, including `astryx_surface`, typecheck, e2e, and CodeRabbit.
- The merge conflict with `main` was resolved by merging `origin/main` in.

No changes are expected from you unless something stands out; an approval would let this merge. Thanks!

<details>

@Astro-Han —— 方便的话请快速看一眼这个 PR：

- 这是 #2560 的 Phase 1（Work Board capture/list MVP），基于已合并的 Phase 0 契约/store（#3028）。
- liugddx 已经 **Approve**，并确认了 store 的理由（provenance + Session 关联）和边界划分。
- 所有 CI 检查全绿，包括 `astryx_surface`、typecheck、e2e 和 CodeRabbit。
- 与 `main` 的冲突已通过合并 `origin/main` 解决。

除非有需要指出的问题，不需要额外改动；approve 后即可合并。谢谢！

</details>

Astro-Han (commented):
The overall architecture is sound: `WorkBoardStore` remains the single mutation and persistence authority in Desktop main, the renderer is an IPC projection, and this does not create a second Runtime Host or Task Ledger authority. I also independently verified that the previous review threads are resolved on `f47d56a`, the existing approval covers this head, the PR is mergeable/clean, and the relevant CI is green.

I found no P0/P1 issues, but I think two P2 gaps should be closed before adding another approval:

1. **[P2] Preserve the store's pagination contract in the renderer projection.** `useWorkBoard()` discards `WorkBoardPage.nextCursor`, while the store intentionally has no total item cap and defaults to 50 results. Once an Inbox or project scope exceeds 50 active plus archived items, older items silently become unreachable; recently updated archived items can also crowd an older active item off the only page. Please retain the cursor and expose a bounded Load more path. Raising the limit to 100 would only move the cutoff.

2. **[P2] Keep the selected filter and effective mutation scope identical.** If the current project disappears while the Project filter is selected, `scopeForFilter()` silently falls back to Inbox, but the Project button and section label remain active. `create()` then writes the item to Inbox under a surface that still says Current project. Please derive one effective filter/scope and use it consistently for the label, query, and create operation, or atomically return the filter to Inbox when `projectId` becomes null.

One non-blocking follow-up:

- **[P3] Guard composing Enter in create and rename.** Both raw inputs treat every Enter as submission. Enter is also how CJK IMEs confirm a candidate, so this can create or rename an item with unfinished text. Reusing the established input seam, or applying the existing `isComposing` guard from `InlineRenameInput`, would close this cleanly.

The current Work Board tests exercise the main-process IPC/store boundary, but the Electron suite contains no Work Board renderer journey, so green CI does not cover these behaviors. A focused renderer/Electron regression for pagination/scope would provide the missing evidence without broadening the suite.

**Go/stop:** hold this head for the two small P2 renderer fixes; the P3 does not need to block. No PR split or architectural rewrite is needed. After those fixes, the Phase 1 shape looks ready to approve.

Codex assisted this review by tracing the current diff, existing feedback, owner boundaries, and CI evidence. The human reviewer is responsible for the final judgment and merge decision.

<details>

整体架构是正确的：`WorkBoardStore` 仍是 Desktop main 中唯一的变更与持久化权威，renderer 只是 IPC 投影，也没有引入第二套 Runtime Host 或 Task Ledger 权威。我还独立确认了当前 `f47d56a` 上前序 review threads 均已解决、已有批准覆盖该 head、PR 可干净合并且相关 CI 全绿。

没有 P0/P1，但建议在新增 Approve 前关闭两个 P2：

1. **[P2] renderer 应保留 store 的分页契约。** 当前 hook 丢弃 `nextCursor`，而 store 没有总量上限且默认只返回 50 条。某个 Inbox 或项目超过 50 条 active + archived item 后，旧事项会静默不可达；最近更新的归档项也可能把较旧的 active item 挤出唯一一页。请保留 cursor 并提供有界的“加载更多”，单纯把上限改成 100 只会移动截断点。
2. **[P2] UI 筛选与实际写入 scope 必须一致。** 当前项目消失时，Project filter 和区块标签仍保持选中，但查询已静默回退 Inbox，新增事项也会写入 Inbox。请让标签、查询和新增共用同一个 effective filter/scope，或在 `projectId` 变为 null 时原子回到 Inbox。

一个非阻塞 follow-up：

- **[P3] 新增和改名应忽略 IME composition 中的 Enter。** 中日韩输入法用 Enter 确认候选词，当前实现可能提前创建或保存未完成标题。复用现有输入 seam，或采用 `InlineRenameInput` 已有的 `isComposing` guard 即可。

当前测试只覆盖 main IPC/store，Electron suite 没有 Work Board renderer journey，因此全绿 CI 不能覆盖上述行为。补一条聚焦的 pagination/scope renderer/Electron 回归即可，无需扩大测试范围。

**结论：**先完成两个小的 P2 renderer 修复；P3 不阻塞。无需拆 PR 或改架构，修复后即可 Approve。

本次审查由 Codex 协助追踪当前 diff、前序反馈、职责边界和 CI 证据；最终判断与合并责任仍由人工 reviewer 承担。

</details>

somewan820:
@Astro-Han — both P2 gaps and the P3 are fixed in `8d761edea`:

1. **Pagination contract**: `useWorkBoard` now retains `WorkBoardPage.nextCursor` and the panel exposes a bounded “Load more” path, so items beyond the store's 50-item default page are reachable instead of silently disappearing.
2. **Scope consistency**: when the current project disappears, the filter atomically returns to Inbox, so the section label, list query, and create operation all use the same effective scope.
3. **IME (P3)**: create and rename ignore Enter while an IME composition is active.

Verification: full desktop typecheck, main build + Work Board IPC tests, and Biome all pass.

On the renderer/Electron regression suggestion: the desktop suite currently has no renderer test harness for this panel; I'd suggest adding a focused e2e journey in a follow-up rather than blocking this PR. Happy to add it after merge if you'd like.

<details>

@Astro-Han —— 两个 P2 和 P3 都已在 `8d761edea` 修复：

1. **分页契约**：`useWorkBoard` 现在保留 `WorkBoardPage.nextCursor`，面板提供有界的“加载更多”，store 默认 50 条之外的事项不再静默不可达。
2. **scope 一致性**：当前项目消失时 filter 原子回到 Inbox，区块标签、列表查询和新增操作都使用同一个 effective scope。
3. **IME（P3）**：输入法 composition 期间，新增和改名会忽略 Enter。

验证：desktop 全量 typecheck、main build + Work Board IPC 测试、Biome 均通过。

关于 renderer/Electron 回归测试：目前 desktop 测试体系没有这个面板的 renderer 测试 harness，建议作为 follow-up 加一条聚焦的 e2e journey，而不是阻塞本 PR。如果你需要，合并后我可以补。

</details>

apps/desktop/src/renderer/use-work-board.ts
```diff
@@ -0,0 +1,157 @@
...
+  const [snapshot, setSnapshot] = useState<WorkBoardSnapshot>(EMPTY_SNAPSHOT);
+
+  const load = useCallback(
+    (preserveItems: boolean) => {
+      const revision = ++revisionRef.current;
+      setSnapshot((current) => ({
+        items: preserveItems ? current.items : [],
+        nextCursor: preserveItems ? current.nextCursor : undefined,
+        loading: true,
+      }));
```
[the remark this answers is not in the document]
somewan820: Fixed in `e727d4fc6`: refresh / loadMore failures no longer replace the list with a fatal error when items already exist. A non-fatal continuation banner keeps the items visible, and retry re-runs the same failed cursor (or reloads page one for refresh failures). Desktop typecheck, main build + Work Board IPC tests, and Biome all pass.

apps/desktop/src/renderer/work-board-panel.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+  const startRename = (item: WorkBoardItem): void => {
+    setRenamingId(item.id);
+    setRenamingTitle(item.title);
+  };
+
+  const saveRename = async (item: WorkBoardItem): Promise<void> => {
+    const title = renamingTitle.trim();
+    if (!title) return;
+    if (await runAction(() => board.update(item.id, { title }))) {
```
Astro-Han: **P2 — Preserve the store's revision-CAS invariant in the renderer.** The panel has the exact `WorkBoardItem` revision it rendered, but rename (and the update/archive/remove handlers below) omit `expectedRevision`. If another Desktop writer changes revision 1 to 2 while this editor is open, this stale write succeeds as revision 3 and silently overwrites the concurrent change, despite `WorkBoardStore` already owning an atomic CAS seam. Pass `{ expectedRevision: item.revision }` for item mutations, keep the draft on `operation_conflict`, and reload through the existing change signal. Please add a two-writer/renderer conflict regression rather than adding a second client-side authority.

apps/desktop/src/renderer/use-work-board.ts
```diff
@@ -0,0 +1,198 @@
...
+      void window.maka.workBoard.list(query).then(
+        (result) => {
+          if (revision !== revisionRef.current) return;
+          if (result.ok) {
+            setSnapshot({
```
Astro-Han, line 127: **P2 — A mutation refresh discards every loaded page after page 1.** `load(true)` preserves the current items only while the request is pending; on success this assignment replaces the whole projection with the default first page (50 items) and resets `nextCursor`. After a user loads 50+10 items, any create/update/archive signal silently collapses the view back to 50. Preserve the loaded window by re-fetching to the prior extent or reconcile the mutation result locally while keeping main as the authority. Add a focused test: load two pages, emit `workBoard:changed`, and assert the already-loaded items/cursor remain coherent.

apps/desktop/src/renderer/work-board-panel.tsx
```diff
@@ -0,0 +1,314 @@
...
+  onArchive(): void;
+  onUnarchive(): void;
+  onRemove(): void;
+}) {
+  const { item, copy } = props;
+  return (
+    <li className="maka-work-board-row" data-archived={item.archived || undefined}>
+      <div className="maka-work-board-row-main">
+        {props.renaming ? (
+          <input
```
Astro-Han: **P2 — Reuse the established rename-input seam so rename is keyboard-usable.** Clicking Rename swaps the title for this raw input, but it has no `autoFocus`/focus effect, so keyboard input remains on the button until the user clicks again. The repository already has `InlineRenameInput`/Astryx `TextInput` behavior for focus/select, IME, Escape, and keyboard-event isolation. Reuse or extract that seam instead of maintaining a second raw-input implementation, and cover click Rename → immediate typing/save in a renderer test.

Astro-Han (commented):
Thanks for the Phase 1 implementation. The core architecture is sound: Desktop main owns the single persistent Work Board store, the renderer consumes an IPC projection, and the store already provides transactions, revisions, CAS, and keyset pagination. That is the right first-principles boundary; there is no need to introduce another Runtime or renderer authority.

The remaining issues are all at the renderer seam: the panel currently does not consume two guarantees the store already provides (revision CAS and stable pagination), and the rename interaction bypasses the repository's established input behavior. Fixing those locally is simpler and safer than changing the storage or IPC design.

**Conclusion: COMMENT — no P0/P1; three P2 items need handling before final approval.**

I did not retain a separate “rapid double-click creates duplicates” comment: it shares the same pending-mutation/input ownership boundary and should be covered by the focused renderer interaction tests added for these fixes.

*AI-assisted review disclosure: Codex coordinated two independent `@reviewer` passes and a read-only `ollama-cloud/deepseek-v4-flash:high` pass, then verified the surviving claims against exact head `1499ff3facd627c9da22f2c598ba62d1c17dfb03`, the store CAS contract, renderer hook, panel code, and current CI state. No local test suite was run. A human contributor remains responsible for deciding and verifying the final fix.*

<details>
<summary>中文摘要</summary>

Phase 1 的整体架构正确：Desktop main 是唯一持久化 authority，renderer 只消费 IPC 投影；store 已提供事务、revision CAS 和 keyset pagination，无需新增 Runtime 或 renderer 权威。

当前剩余问题都在 renderer seam：panel 没有使用 store 已有的 CAS；mutation 刷新会把已加载多页静默缩回第一页；rename 绕过既有输入组件且不会自动聚焦。最简单的方案是在现有 panel/hook 层修正并补聚焦的 renderer 测试，不改 storage/IPC 架构。

结论：COMMENT，无 P0/P1；3 个 P2 需处理后再最终批准。

</details>

Astro-Han:
This PR adds a substantial new Desktop surface. Could you please add screenshots of the Work Board UI before merge? At minimum, please show the populated board and one empty or editing state; one annotated composite is fine. Thanks!

somewan820:
Sorry for the delay — I have been tied up with personal matters over the past few days. I have reviewed the latest feedback and will follow up as soon as possible.\n\nI will prioritize the remaining items around renderer revision-CAS conflict handling, preserving loaded pagination across refreshes, using the shared TextInput/IME-safe rename and create controls, and releasing the WorkBoardStore during desktop shutdown. I will also update the AI-use disclosure and rerun the focused tests and checks after the fixes.\n\nThanks for the detailed review and your patience.

somewan820:
@coderabbitai review

somewan820:
Follow-up for the latest review round (commit `7d2b219f3`):

Implemented and pushed:

- Added an explicit `WorkBoardStore.close()` registration handle and close it during desktop shutdown; tests now close internally created stores.
- Passed `{ expectedRevision: item.revision }` through rename, complete, reopen, move, archive, restore, and remove mutations so stale renderer writes return `operation_conflict`.
- Preserved the number of already loaded Work Board items when mutation signals refresh the list, including continuation pagination.
- Replaced raw inputs with Astryx `TextInput` and retained Enter/Escape, IME composition, and `Process` handling for create and rename.

Verification already completed: desktop typecheck, main/preload/renderer builds, focused Work Board IPC tests (2/2), focused Work Board storage tests (18/18), and Biome checks.

AI-use disclosure: Codex (OpenAI) made a substantive contribution to the implementation, tests, and documentation for this review round. I reviewed the generated changes, verified the behavior and tests, and remain responsible for the final result. The commit includes the required `Generated-by: Codex` trailer.

docs/astryx-surface-file-inventory.md
```diff
@@ -158,6 +158,8 @@ Wiki bar: Design Conventions · API Use-the-System · Theming · Container Paddi
 | `apps/desktop/src/renderer/styles/sidebar.css` | shell-chrome-or-panel | n/a (css) | aligned — no off-rhythm control heights flagged | aligned |
 | `apps/desktop/src/renderer/styles/task-ledger.css` | styles | n/a (css) | aligned — no off-rhythm control heights flagged | aligned |
 | `apps/desktop/src/renderer/styles/theme-glass.css` | styles | n/a (css) | aligned — no off-rhythm control heights flagged | aligned |
+| `apps/desktop/src/renderer/styles/work-board.css` | styles | n/a (css) | aligned — no off-rhythm control heights flagged | aligned |
+| `apps/desktop/src/renderer/work-board-panel.tsx` | shell-chrome-or-panel | Banner, Button, EmptyState, Spinner | raw `<input` (API Use-the-System) | blocker |
```
hqhq1025: [P1] Regenerate the Astryx inventory after syncing current `main`. This row still records a raw `<input>` blocker, but this head now uses `TextInput` for both Work Board fields. On this exact SHA, `npm run astryx:surface-inventory` fails because the committed Markdown differs from the generator, and the selected `astryx_surface` CI job runs that same byte-for-byte check. Current `main` also conflicts in this generated file. Resolve the conflict, run `npm run astryx:surface-inventory:write`, and commit both generated artifacts.

apps/desktop/src/renderer/work-board-panel.tsx
```diff
@@ -0,0 +1,345 @@
...
+
+  const startRename = (item: WorkBoardItem): void => {
+    setRenamingId(item.id);
+    setRenamingTitle(item.title);
+  };
+
+  const saveRename = async (item: WorkBoardItem): Promise<void> => {
+    const title = renamingTitle.trim();
+    if (!title) return;
+    if (await runAction(() => board.update(item.id, { title }, { expectedRevision: item.revision }))) {
```
hqhq1025: [P2] Freeze the revision that the rename draft was opened against. If another Work Board panel updates revision 1 while this editor is open, the change signal reloads the row as revision 2 but leaves `renamingTitle` intact; this callback then receives the refreshed `item` and submits the stale draft with `expectedRevision: 2`, so CAS accepts it and silently overwrites the concurrent title. Store the base revision alongside the rename draft and use that value until save or cancel, preserving the draft when the store returns `operation_conflict`.

apps/desktop/src/renderer/work-board-panel.tsx
```diff
@@ -0,0 +1,345 @@
...
+    </li>
+  );
+}
+
+export function WorkBoardPanel(props: { projectId: string | null }) {
+  const copy = getDesktopConversationCopy(useUiLocale()).workBoardPanel;
+  const [filter, setFilter] = useState<'inbox' | 'project'>('inbox');
+  const query: WorkBoardListQuery = useMemo(
+    () => ({
+      scope: scopeForFilter(filter, props.projectId),
```
hqhq1025, line 150: [P2] Keep project-scoped Board items reachable after conflict-safe project relinking. The project catalog can absorb a conflicting project ID as an alias and rewrites its Session headers to the canonical ID, but existing Work Board rows retain the absorbed ID. This query uses only the canonical `projectId`, so those items match neither Inbox nor the current-project view and disappear from the UI. Canonicalize affected Work Board scopes in the relink transaction, or make the Board query resolve the canonical ID together with its aliases.

hqhq1025 (dismissed):
The current revision is not ready: one deterministic required-CI blocker and two material data-integrity/visibility defects remain. I rechecked the earlier lifecycle, input-control, and loaded-page fixes; those changes otherwise hold. The head also conflicts with current `main` and has no check runs, so the conflict-resolution revision will require a fresh review.

somewan820:
Conflict resolved against Apache Maka `main` (`57e08d83`). The branch now points at merge commit `34ed15f80`, with the Work Board Phase 1 changes preserved and the Astryx surface inventory regenerated to match the merged tree. GitHub now reports the PR as `MERGEABLE`.

hqhq1025 (commented):
Re-reviewed exact head `34ed15f8`. The Astryx inventory blocker is fixed, and the exact-head build and focused tests pass. Under policy v7’s materiality gate, the prior rename-revision concern has no supported current second-title-writer path in the single-instance Desktop, so I resolved that thread. The existing project-relink alias reachability P2 remains actionable; I am not duplicating its inline comment. GitHub currently reports no check runs on this head, so this is a COMMENT rather than an approval.

somewan820:
Addressed the remaining project-relink reachability P2 in `9eb42115a`:

- Work Board project queries now accept the canonical project ID plus absorbed aliases.
- The renderer derives the canonical project identity from the active project record and passes aliases through the workbar to the panel.
- Query normalization validates, deduplicates, and canonicalizes the identity set; pagination cursors include the complete project filter fingerprint.
- Storage regression coverage verifies that items persisted under an absorbed project ID remain visible after relink.

Verification:
- `@maka/core` build and Work Board contract: 15/15 passed.
- `@maka/storage` build and Work Board store: 17/17 relevant tests passed; the unrelated backup test is blocked in this sandbox by an existing `EPERM` while chmod-ing `~/Library/Caches/Maka/runtime-hosts`.
- Biome and `git diff --check` pass.

The new head is pushed and ready for fresh CI/review.

somewan820:
Follow-up on the remaining review blockers at `bf39a7bb7`:

- Rename drafts now capture the item revision when editing starts and keep that revision for save, so a change signal cannot turn a stale draft into an accepted overwrite. The draft remains open on CAS conflict.
- Project relink reachability is fixed in `9eb42115a`: canonical project IDs and absorbed aliases flow through the workbar and the storage query, with regression coverage.
- Loaded-page refresh is already preserved by `useWorkBoard`'s loaded-item window refetch; the continuation cursor remains available after `workBoard:changed`.
- Work Board IPC owns a closeable registration, and Desktop shutdown closes it; the boot path injects the store and the focused IPC tests cover registration lifecycle.
- Both Work Board fields use Astryx `TextInput`; `npm run astryx:surface-inventory` passes.

Verification on the exact branch:
- `@maka/core` build and Work Board contract tests pass.
- `@maka/storage` build and Work Board store tests pass for the focused suite.
- Biome, `git diff --check`, and the Astryx inventory check pass.

The PR is now marked ready for review so required CI can run. Please re-review the new head and confirm the resolved threads.

somewan820:
CI follow-up: after marking the PR ready and synchronizing `7afa7af2`, Apache CI is being created but immediately finishes as `action_required` with no jobs (`https://github.com/apache/maka/actions/runs/32450544416`). This is the fork workflow-approval gate, not a test failure. A maintainer needs to approve the workflow run; once approved, the required `test` check should execute on the current head.

The code and review-thread blockers have been addressed on `bf39a7bb7`/`9eb42115a`; the latest empty commit only retriggered the workflow after the draft-to-ready transition.

somewan820:
Four workflows are awaiting maintainer approval. Please approve and run them, then re-review the current head 7afa7af22. All review threads are resolved.

somewan820:
已按你这条 review 继续补了：Work Board 现在会把 `projectAliases` 一起带到查询里，`WorkBoardPanel -> useWorkBoard -> WorkBoardStore.list()` 这条链路也已经支持 absorbed alias 的可达性。`packages/storage/src/work-board-list-query.ts` 里 `projectIds` 会和当前 project scope 一起进查询，`packages/storage/src/__tests__/work-board-store.test.ts` 也有对应回归。麻烦你再看一下，如果这条已经覆盖到位，帮我把旧的 `Request changes` 结论撤掉/改成重新 review 一下。

packages/storage/src/work-board-list-query.ts
```diff
@@ -50,7 +50,7 @@ export function workBoardFilterFingerprint(value: WorkBoardListQuery): string {
     value.scope === undefined
       ? 'any'
       : value.scope.kind === 'project'
-        ? `project:${value.scope.projectId}`
+        ? `project:${[...(value.projectIds ?? [value.scope.projectId])].sort().join('|')}`
```
hqhq1025, line 73: [P2] Keep alias-filter cursors within their own decode contract

`workBoardFilterFingerprint()` now concatenates every canonical/absorbed project id, but the cursor decoder still rejects fingerprints longer than 512 characters and the public query rejects cursors longer than 1024. Project relinking has no alias-count bound, so a project with 14 absorbed UUID aliases can produce a valid first page whose own continuation cursor is rejected. I reproduced this with 15 identities and 101 rows: page one returned 100 rows and an 892-character cursor; page two failed with `invalid_input`. Use a fixed-length digest of the normalized identity set for the cursor fingerprint, or impose one consistent bound, and add a multi-page alias regression.

hqhq1025 (commented):
One actionable P2 finding remains on the current head.

Problem and mechanism: the PR correctly introduces a small user-owned Work Board with a main-process storage authority, CAS-protected mutations, bounded list pagination, and a Desktop workbar projection. The latest fixes preserve the revision captured when rename starts and make project-scoped rows reachable through the canonical project plus absorbed aliases. Both previous findings are fixed at their ownership boundaries.

First principles and Occam: passing the catalog's canonical identity set into one Storage query is simpler and more correct than rewriting historical Work Board rows during every project relink. However, the pagination cursor now embeds that unbounded identity set verbatim while the existing decoder accepts only a 512-character fingerprint. The first page can therefore mint a cursor that the second page itself refuses.

Optimality/refactor: keep the normalized alias set as the SQL authority, but bind the cursor to a fixed-length digest of that set (or introduce one consistent, evidenced bound across aliases, projectIds, fingerprint, and cursor). Do not raise only the decoder limit; the public cursor still has a separate 1024-character cap. Add a multi-page regression with enough UUID aliases to cross the old fingerprint limit.

Deletion and tests: I found no production code or low-value test that should be deleted. The current focused tests cover alias reachability but stop before a continuation cursor is required, which is why this deterministic failure remains invisible.

Merge verdict: not ready to merge until alias-scoped pagination can consume every cursor it emits. This review is a normal COMMENT under the maintainer's comment-only policy; it does not use REQUEST_CHANGES.

Verification: reviewed head a84c251ee4b3914e3ae767fa10a94fdd764379b6; npm run build:test passed; 35 focused Core, Storage, and Desktop Work Board tests passed; Astryx surface inventory passed; git diff --check passed. A real-store reproducer with 15 UUID project identities and 101 records returned 100 records plus an 892-character cursor on page one, then failed page two with invalid_input.

somewan820:
Update after continuing the work:

  - Rebased onto the latest `apache/maka/main`; the current head is `fd7eb300b`, and GitHub reports the PR as `MERGEABLE`.
  - Fixed the oversized pagination cursor issue for relinked project aliases. The query still uses the complete normalized `projectIds` set, while the cursor now stores a SHA-256
  base64url digest of that set, keeping the cursor within its validation limits.
  - Added a cross-page regression test covering 15 UUID aliases and 101 records, verifying that the cursor generated on page one can successfully fetch page two.
  - Core Work Board tests pass 15/15. Storage Work Board tests pass 18/19; the only failure is a local sandbox permission error when attempting to `chmod /Users/some/Library/
  Caches/Maka/runtime-hosts`, unrelated to this change.
  - `gh pr checks` still reports `no checks reported`. If the branch is still waiting at the fork workflow approval gate, please approve the workflow and rerun the required checks
  and review against the current head.

  Please review `fd7eb300b` again. The previously reported alias-scoped pagination cursor issue has been addressed.

apps/desktop/src/renderer/work-board-panel.tsx
```diff
@@ -0,0 +1,363 @@
...
+                    )
+                  }
+                  onArchive={() =>
+                    void runAction(() => board.archive(item.id, { expectedRevision: item.revision }))
+                  }
+                  onUnarchive={() =>
+                    void runAction(() => board.unarchive(item.id, { expectedRevision: item.revision }))
+                  }
+                  onRemove={() =>
+                    void runAction(() => board.remove(item.id, { expectedRevision: item.revision }))
```
Astro-Han: **[P3] Both row maps pass a batch of handlers the row will never call.**

`WorkBoardRow` picks its action set from `item.archived`: archived rows render `unarchive` + `delete`, active rows render `complete`/`reopen`, `rename`, `move`, `archive`. But the active map passes `onRemove` and `onUnarchive`, and the archived map passes no-op `onComplete` / `onReopen` / `onMove` / `onRename*`. In both directions those props are unreachable.

This is cosmetic today — the gating is correct and nothing misfires — but it makes the row's contract read as "any row can do anything", which is the opposite of what the component actually enforces. It also makes the no-op `() => undefined` handlers in the archived map look like placeholders for unfinished work rather than dead parameters.

Smallest fix: split the props into the two sets the component already branches on, so the type makes the branch explicit instead of the runtime doing it silently.

Astro-Han (commented):
Thanks — the capture/list MVP is well scoped, and the layering holds up: the store owns content and revisions, the IPC boundary owns the id, and the renderer is a projection rather than a second authority. We looked specifically for a duplicated source of truth and did not find one.

Reviewed at exact head `b617faf496a3d26001a1f4831f9065b4cfd6fdec`. **No P0–P2. One P3, inline.**

Two things we checked that are worth recording, because both are the kind of thing that looks wrong from the outside:

**The three lifecycle operations are not redundant.** We went in expecting to argue that `done` / `archive` / `delete` could collapse into fewer concepts, and they cannot: `done` is a state the item keeps while staying visible and reversible, `archive` moves it out of the active list while preserving the captured provenance excerpt, and `delete` is the only operation that actually erases that excerpt. Since the excerpt is user content captured from a conversation, being able to erase it is a distinct requirement that archiving cannot serve. `work-board-store.ts:221` also gates deletion behind `must_archive_first`, so the destructive path is two steps by construction rather than by UI convention.

**The `emitChanged` reload has no race.** `use-work-board.ts` bumps a revision on every load and discards the response of any load that is no longer the latest, and main emits `changed` only after the store write lands, so a reload always observes state at or after the mutation that triggered it. Rapid successive mutations settle on the last reload.

On CI: the `test` failure on this head is from a run created at 09:45 UTC today, well before `main` moved several times, so it is a stale merge base rather than something this PR introduced — but it is also not usable as a green signal. Separately, #3397 adds an ASF license header to the first line of ~2700 files and is close to landing; this branch will need a rebase once it does. We are holding APPROVE until there is a green run on a current head.

---

Astro-Han:

- `apps/desktop/src/renderer/app-shell.tsx`
- `apps/desktop/src/renderer/features/workbar/model/workbar-tabs.ts`
- `apps/desktop/src/renderer/features/workbar/ui/workbar-host.tsx`
- `…(more)`

```
# resolve, then
git push --force-with-lease
```

---

Astro-Han:
Heads-up on CI, not a review finding: **this head's CI run is red on the Astryx surface inventory gate, and it fails before any test runs.**

Its checks had never executed — the run was parked awaiting maintainer approval. I released it (`32684069657`), so the result is real rather than a new regression.

**Where it stops**: the `Astryx surface inventory` step of the `test` job (`npm run astryx:surface-inventory`).

**What it found** — reproduced locally at `6f6aa19fca525156d8bcd1ba0bee154f7fb17155`:

```
astryx surface inventory is stale; run: npm run astryx:surface-inventory:write
- .paths does not match generator output
- docs/astryx-surface-file-inventory.md does not match generator output
```

This gate is generated, not hand-maintained: it became mandatory for code changes in `33e6b7d95` (*ci: enforce Astryx inventory for code changes*, #3647), which landed after your last push — so this is a new requirement your branch has not seen, not something you broke.

**`npm run astryx:surface-inventory:write` regenerates both files**; commit them and the pipeline can proceed.

Separately, and unchanged from the earlier note on this PR: this branch still conflicts with current `main`, so it needs a rebase regardless.

<details>
<summary>中文</summary>

CI 提示，不是 review finding：**这个 head 红在 Astryx surface inventory 门禁上，在任何测试之前就失败了。**

它的检查此前从没执行过——run 卡在等放行，我已放行（`32684069657`），所以这个红是真实结果。

**停在哪**：`test` job 的 `Astryx surface inventory`（`npm run astryx:surface-inventory`）。

**报了什么**（我在 `6f6aa19f` 上本地复现）：清单已过期，`.paths` 和 `docs/astryx-surface-file-inventory.md` 都与生成器输出不一致。

这个门禁是**生成的、不是手写维护的**，而且它是在 `33e6b7d95`（*ci: enforce Astryx inventory for code changes*，#3647）里才对代码改动变成强制的——那个提交在你上次推送之后才合入，**所以这是你的分支没见过的新要求，不是你弄坏了什么。**

**`npm run astryx:surface-inventory:write` 会重新生成这两个文件**，提交后流水线就能继续往下走。

另外，与此前那条提醒一致：这个分支目前仍与当前 `main` 冲突，无论如何都需要 rebase。
</details>

Astro-Han:
Correcting one sentence in my previous comment.

I wrote that #3647 "landed after your last push". **That is wrong, and I should have checked it before posting.** #3647 (`33e6b7d95`) was committed at `2026-08-24T03:13:52+08:00`; this head (`6f6aa19f`) was committed at `2026-08-24T10:44:24+08:00` — **later**, not earlier.

The accurate version: **your branch does not contain `33e6b7d95`.** Its merge-base with `main` is `8adb74b4f`, which predates that commit, so the branch has never picked up the inventory generator's current output. That is a consequence of the branch being behind `main`, not of the gate arriving after you pushed.

Everything else in that comment stands unchanged: the run was parked and I released it, the failure is the generated inventory being stale, `npm run astryx:surface-inventory:write` regenerates both files, and the branch needs a rebase onto current `main` regardless — which is also what will bring the generator up to date.

<details>
<summary>中文</summary>

更正我上一条评论里的一句话。

我写了 #3647「在你上次推送之后才合入」，**这是错的，我发之前应该先查**。#3647（`33e6b7d95`）提交于 `2026-08-24T03:13:52+08:00`，而这个 head（`6f6aa19f`）提交于 `2026-08-24T10:44:24+08:00`，**在后不在前**。

准确的说法是：**你的分支不包含 `33e6b7d95`。** 它与 `main` 的 merge-base 是 `8adb74b4f`，早于那个提交，所以分支从未拿到清单生成器当前的输出。这是分支落后于 `main` 导致的，不是门禁在你推送之后才出现。

那条评论的其余部分不变：run 此前卡在等放行、是我放行的；失败原因是生成的清单已过期；`npm run astryx:surface-inventory:write` 会重新生成这两个文件；分支无论如何都需要 rebase 到当前 `main`——rebase 同时也会把生成器带到最新。
</details>

apps/desktop/src/renderer/work-board-panel.tsx
somewan820: Addressed in `1c8d833f8` and now outdated on the current diff. `WorkBoardRow` receives a discriminated `actions` union: active rows get only active actions, while archived rows get only `onUnarchive` and `onRemove`. The no-op and unreachable handler props were removed.

The current head `5d4481ba6` is green in both Linux CI and the Windows release check.

packages/storage/src/work-board-list-query.ts
somewan820: Addressed in `a05df8915` and verified again on the current head `5d4481ba6`.

`workBoardFilterFingerprint()` now hashes the complete normalized alias-filter identity with SHA-256/base64url, so the cursor keeps a fixed-size fingerprint without dropping any project aliases from the query contract. The storage regression uses 15 UUID identities and 101 rows: page one returns 100 rows with a cursor below the decoder bound, and page two returns the final row successfully.

Current-head CI is green:
- Linux CI: https://github.com/apache/maka/actions/runs/32715018884
- Windows release check: https://github.com/apache/maka/actions/runs/32715018879

somewan820:
Final verification on current head `5d4481ba6`:

- Added the focused renderer regression requested for paginated mutation refresh: load 50 + 10 items, emit `workBoard:changed`, then verify all 60 items remain loaded and the refresh requests the prior window depth.
- Rechecked the alias-cursor P2: the fingerprint is a fixed SHA-256/base64url digest of the complete normalized identity set, with the existing 15-alias / 101-row cross-page regression.
- All review threads are now answered and resolved; GitHub reports the PR as `MERGEABLE` against base `1e1c886a`.
- Linux CI passed: https://github.com/apache/maka/actions/runs/32715018884
- Windows release check passed: https://github.com/apache/maka/actions/runs/32715018879

The remaining merge-state blocker is `REVIEW_REQUIRED`; please re-review the current head.

apps/desktop/src/renderer/work-board-panel.tsx
```diff
@@ -0,0 +1,380 @@
...
+              event.preventDefault();
+              void create();
+            }
+          }}
+          placeholder={copy.createPlaceholder}
+        />
+        <Button
+          size="sm"
+          label={copy.create}
+          onClick={() => void create()}
```
jackwener, line 276: [P2] Prevent duplicate creates while the first request is in flight

`create()` has no synchronous in-flight guard, and the input/button remain enabled until `board.create()` resolves. I mounted the real `WorkBoardPanel` with a deferred IPC result and activated Create twice; this produced two `workBoard.create` calls while the first was still unresolved. Each create allocates a fresh UUID, so the CAS layer cannot deduplicate the pair and two tasks can be persisted from one activation burst.

Please guard the entry with synchronous pending state (not only a post-render visual flag), disable the input and button while pending, and add a renderer test asserting that a second Enter/click before resolution produces only one IPC create.

apps/desktop/src/renderer/features/workbar/tools/side-chat/quote-companion-panel.tsx
```diff
@@ -268,8 +268,6 @@ export function QuoteCompanionPanel(props: {
               pendingAttachments={pendingAttachments}
               onRemoveAttachment={removeAttachment}
               mentionSkills={props.mentionSkills}
```
jackwener, line 271: [P2] Forward the skill-catalog state together with `mentionSkills`

`QuoteCompanionPanel` still accepts `mentionSkillsUnavailable` and `mentionSkillsLoading`, and the AppShell-to-Workbar path still supplies them, but this seam now forwards only `mentionSkills`. During a catalog refresh, `useComposerMentions` intentionally clears the live list; `Composer` needs the settled/unavailable and loading facets to distinguish that transient state from a genuinely empty catalog, keep the menu geometry stable, and make activation a no-op while loading. Without them, Quote Companion can briefly render the empty-state message and change menu behavior on each refresh.

The Composer contract tests pass, so the lower-level behavior is intact; the state is lost here. Please restore both props and add a Quote Companion seam test that covers a temporarily cleared list while loading.

apps/desktop/src/renderer/features/workbar/tools/side-chat/quote-companion-panel.tsx
```diff
@@ -207,7 +208,6 @@ export function QuoteCompanionPanel(props: {
     >
       <ChatSurfaceLayout
         conversationKey={companion.companionSession?.id ?? props.sourceSession?.id}
-        scrollToBottomLabel={copy.scrollToBottom}
```
jackwener, line 210: [P3] Preserve the localized scroll-control label

Removing `scrollToBottomLabel={copy.scrollToBottom}` drops Maka's locale-specific label for the Quote Companion scroll button and falls back to the dependency's default copy. That regresses non-English and screen-reader text on this surface. Please restore this prop.

jackwener (commented):
Review of exact head `5d4481ba648963a9488b78fbc134acbdd9bc0ed7`: not ready to merge (2 P2, 1 P3).

The exact-head `test` and `package` checks are green. I also ran `build:test`, focused Core/Storage/Desktop tests (54/54), and the Composer mention-menu contract tests (10/10). A synthetic merge with current `main` built successfully and passed the same focused 54-test suite. The findings are inline below.

apps/desktop/src/renderer/work-board-panel.tsx
somewan820: Fixed in ddeb7410a. `create()` now takes a synchronous `createPendingRef` guard before its first await, disables the input and Add button while the request is in flight, and clears both the ref and state in `finally`. Added a deferred IPC regression test covering two clicks before the first request resolves. Desktop tests pass 1388/1388.

apps/desktop/src/renderer/features/workbar/tools/side-chat/quote-companion-panel.tsx
somewan820: Fixed in ddeb7410a. Quote Companion now forwards both `mentionSkillsUnavailable` and `mentionSkillsLoading` to Composer. Added a boundary contract test for the two catalog-state props; the desktop test suite passes 1388/1388.

apps/desktop/src/renderer/features/workbar/tools/side-chat/quote-companion-panel.tsx
somewan820: Fixed in ddeb7410a. Restored `scrollToBottomLabel={copy.scrollToBottom}` so Quote Companion keeps Maka's localized accessible label. The boundary test suite and full desktop tests pass.

Astro-Han (approved):
Approving on `8ebf4cf88ca3190bf8ae1a90c45c698f1dea9db4`, with `test` and `package` terminal green on that exact head.

I re-derived every finding I had left open rather than trusting the earlier round.

**The paginated-refresh P2 is properly fixed.** `use-work-board.ts` now re-fetches to the previously loaded extent through `listWindow`, which pages up to `loadedItemCountRef` with `WORK_BOARD_PAGE_SIZE_MAX` and drops duplicates by id, so a `workBoard:changed` signal after 50+10 items no longer collapses the view to the first page. The revision guard still discards responses from superseded loads, and a continuation failure keeps the existing items with a retry on the same cursor instead of replacing the list.

**The row-handler P3 is fixed better than I asked.** Splitting `WorkBoardRow`'s props into an `active | archived` discriminated union means the archived branch cannot be handed active-only callbacks at all — the compiler enforces what was previously a convention. That is a stronger fix than dropping the unused handlers.

**The double-submit guard on create is correct.** `createPendingRef` is checked and set synchronously before the first `await`, so a second Enter cannot slip through; the `createPending` state is only for rendering, and the `finally` restores both on the failure path.

**The Side Chat disposal fencing holds.** `performCompanionTurn` re-checks `isDisposed()` after each await, and a fork created inside the call is cleaned up when disposal wins the race before the send. The new tests construct the race with deferred promises rather than asserting a single ordering, so they lock the behaviour rather than the implementation.

One observation, not a finding: when disposal wins *after* a successful `send`, the created fork is not scheduled for cleanup. That looks deliberate — a run is already in flight, and `recoverOrphanedCompanionCopies` exists for exactly this reclamation — but if that is the intent, it is worth a comment, since the two neighbouring disposal branches do clean up and this one silently does not.

Merging this on @AstroHan's decision.

<details>
<summary>简体中文</summary>

已在 `8ebf4cf88ca3190bf8ae1a90c45c698f1dea9db4` 上 approve，该 exact head 的 `test` 与 `package` 均为终态绿。

我没有沿用上一轮的结论，而是把此前未闭合的每一条都重新从代码推导了一遍。

**分页刷新那条 P2 确实修好了。** `use-work-board.ts` 现在通过 `listWindow` 按之前已加载的规模重新取数：以 `WORK_BOARD_PAGE_SIZE_MAX` 翻页直到 `loadedItemCountRef`，并按 id 去重。因此加载了 50+10 条之后再来一次 `workBoard:changed`，视图不会再塌回第一页。代次守卫仍会丢弃被取代的加载结果；续页失败则保留已有条目并对同一 cursor 提供重试，而不是整体替换成错误态。

**行处理器那条 P3 修得比我要求的更好。** 把 `WorkBoardRow` 的 props 拆成 `active | archived` 判别联合后，archived 分支根本不可能拿到只属于 active 的回调——原先靠约定维持的东西现在由编译器保证。这比单纯删掉多余的 handler 更强。

**创建的防重复提交守卫是对的。** `createPendingRef` 在第一个 `await` 之前同步检查并置位，第二次回车无法穿过；`createPending` 状态只用于渲染；`finally` 在失败路径上也会把两者复位。

**Side Chat 的 disposal 围栏站得住。** `performCompanionTurn` 在每个 await 之后都重新检查 `isDisposed()`，且当 disposal 抢在 send 之前时，本次调用内创建的 fork 会被安排清理。新增的测试用 deferred promise 真正构造了竞态，而不是只断言某一种顺序——锁的是行为而不是实现。

一条观察，不是 finding：当 disposal 抢在**成功 send 之后**时，已创建的 fork 不会被安排清理。看起来是有意的——此时 run 已经发出，而 `recoverOrphanedCompanionCopies` 正是为这种回收准备的——但如果确实是有意的，建议补一句注释，因为相邻两个 disposal 分支都会清理，唯独这一处不清理。

本 PR 由 @AstroHan 决定合并，我按其决定执行。
</details>

Astro-Han:
LGTM. Merging at @AstroHan's request — `test` and `package` are green on `8ebf4cf88ca3190bf8ae1a90c45c698f1dea9db4`.

<details>
<summary>简体中文</summary>

LGTM，按 @AstroHan 的要求合并——`8ebf4cf88ca3190bf8ae1a90c45c698f1dea9db4` 上 `test` 与 `package` 均为绿。
</details>

PR #3136, opened 2026-08-17
test(runtime-host): reject a handshake with a mismatched compatibility epoch

### LINKED ISSUE #3130: Compatibility epoch mismatch has no handshake-level test
## Problem

`exchangeRuntimeHostHandshake` rejects a peer whose `compatibilityEpoch` differs (`packages/runtime-host/src/client/connection.ts:1441`), and other work relies on that rejection. #3126 narrows `UpdateCatalogConnectionResult` and depends on the handshake keeping epoch-21 Hosts away from the new Client — such a Host still answers a connection update with `invalid_default_target`, which the new decoder cannot parse.

That rejection has no direct test. `host-kernel.test.ts:317` covers an incompatible *protocol range*, not an epoch mismatch. And unlike `compositionId` — which `authenticated-websocket.test.ts:90` varies through a client argument — `RUNTIME_HOST_COMPATIBILITY_EPOCH` is a compile-time constant shared by both ends, so no parameter can put them at different epochs, and `exchangeRuntimeHostHandshake` is module-private.

## Desired outcome

A test asserting that a Host announcing a different compatibility epoch is rejected before any domain command is admitted.

Two possible seams: export `exchangeRuntimeHostHandshake` for direct testing, or give the transport layer a test entry point that can serve a forged handshake frame. The second is more useful — it would also cover the other handshake rejection branches.

1625567290:
## Summary

`exchangeRuntimeHostHandshake` already rejects a peer whose `compatibilityEpoch` differs, and later work depends on that rejection keeping a mismatched Host away from domain commands. That path had no handshake-level test: `host-kernel` only covered an incompatible protocol range, and both ends share the compile-time epoch constant.

This adds two seams:

- A live Host hello with `compatibilityEpoch + 1` is answered `incompatible`, then the connection is closed so a follow-up `host.status` cannot be admitted.
- A transport peer answers `accepted` with a different epoch while the registration file stays current. `connectRuntimeHost` returns `unavailable` / `handshake_failed`, and the peer never receives a domain request.

Fixes #3130

## Test plan

- [x] `@maka/runtime-host` typecheck
- [x] `handshake-compatibility` (1/1)
- [x] `host-kernel` (46/46), including the new epoch-mismatch case

### DIFF
--- a/packages/runtime-host/src/__tests__/handshake-compatibility.test.ts
+++ b/packages/runtime-host/src/__tests__/handshake-compatibility.test.ts
@@ -1,0 +1,160 @@
+import assert from 'node:assert/strict';
+import { randomUUID } from 'node:crypto';
+import { mkdtemp, rm } from 'node:fs/promises';
+import { createServer, type Server } from 'node:net';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { test } from 'node:test';
+import {
+  prepareStorageRootControlDirectory,
+  resolveStorageRoot,
+} from '@maka/storage/root-authority';
+import { connectRuntimeHost } from '../client/index.js';
+import { prepareRuntimeHostEndpoint } from '../control/endpoint.js';
+import { removeHostRegistration, writeHostRegistration } from '../control/registration.js';
+import {
+  decodeClientFrame,
+  encodeProtocolMessage,
+  RUNTIME_HOST_COMPATIBILITY_EPOCH,
+  RUNTIME_HOST_PROTOCOL_VERSION,
+  RUNTIME_HOST_REGISTRATION_SCHEMA_VERSION,
+  type HostFrame,
+  type RequestFrame,
+} from '../protocol/index.js';
+import { FramedTransport, RuntimeHostTransportError } from '../transport/framed-transport.js';
+
+const PROTOCOL = {
+  min: RUNTIME_HOST_PROTOCOL_VERSION,
+  max: RUNTIME_HOST_PROTOCOL_VERSION,
+} as const;
+
+test('rejects a Host that accepts with a different compatibility epoch before any domain command', async () => {
+  let admittedRequest: RequestFrame | undefined;
+  await withForgedHandshakePeer(
+    async (transport, hostEpoch, rootId) => {
+      const hello = decodeClientFrame(await transport.read(2_000));
+      assert.ok('kind' in hello && hello.kind === 'hello');
+      await writeProtocolFrame(transport, {
+        kind: 'accepted',
+        rootId,
+        hostEpoch,
+        connectionId: 'forged-epoch-connection',
+        selectedProtocol: RUNTIME_HOST_PROTOCOL_VERSION,
+        compatibilityEpoch: RUNTIME_HOST_COMPATIBILITY_EPOCH + 1,
+        compositionId: 'maka.interactive',
+        compositionRevision: '1',
+        state: 'ready',
+      });
+      try {
+        const next = decodeClientFrame(await transport.read(1_000));
+        if (!('kind' in next)) admittedRequest = next;
+      } catch (error) {
+        assert.ok(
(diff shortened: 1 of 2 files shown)


Astro-Han:
Thanks for the test — the forged-peer scaffolding is a nice end-to-end seam (no production test hooks), both tests verify the post-rejection state (connection closed / no domain requests admitted), and the `+1` epoch arithmetic stays valid across future bumps. CI is green.

**Conclusion: PASS with one P2.**

**P2-1 — the real upgrade flow that #3126 depends on has no test; the PR's tests cover only two seams.**
The client-side guard the PR pins (`connection.ts:1441`) can only be triggered by a forged/malicious peer — a genuinely old Host never emits a mismatched-epoch `accepted`. The actual upgrade scenario (#3126: new Client + still-running old Host with a stale registration file) is decided earlier, at `staleCompatibility` (`connection.ts:1290-1296`), which rewrites the hello protocol range to `{max+1, max+1}` and is the *first* defense that sends the peer down the `incompatible`/replacement path. Nothing in the repo tests that decision. A regression there (wrong comparison, dropped condition) would go green in CI while the upgrade world loses its only guard against admittance of an old Host that predates the server-side epoch check (`host-kernel.ts:437`). Consider adding a sibling test with the existing `withForgedHandshakePeer` scaffold: seed a registration with `compatibilityEpoch - 1`, assert the hello carries `{protocolMax+1, protocolMax+1}`, reply `incompatible`, and assert `connectRuntimeHost` returns `kind: 'incompatible'`. Or defer explicitly with the reasoning that the server-side check is the backstop.

Optional nits (P3): the `assert.equal(result.reason, 'handshake_failed')` pins the catch-all classification (any handshake failure produces the same reason; an `epoch_mismatch`-style dedicated reason would break it — the `kind === 'unavailable'` + `admittedRequest === undefined` assertions are the semantic ones); the ~90-line scaffold currently serves a single test (issue #3130 notes it would also cover the other rejection branches); `await transport.closed` has no bound (a regression that stops closing would hang to the CI timeout).

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagent on `ollama-cloud/deepseek-v4-flash`). The subagent traced the handshake/staleCompatibility paths and ran static mutation analysis (dropping each guarded check breaks exactly its test); the P2-1 gap is from that tracing, not an observed failure. Please weigh these findings with your own judgment.*

<details>

结论：PASS（1 个 P2）。伪造 peer 的端到端脚手架是好的接缝（无生产测试钩子），两个测试都验证拒绝后的状态（连接关闭 / 无 domain 请求被受理），epoch `+1` 写法在将来 bump 时自动保持有效。P2-1：PR 锁定的 client 端检查（connection.ts:1441）只能被伪造/恶意 peer 触发——真实旧 Host 不会发出 mismatched-epoch 的 accepted；#3126 实际依赖的升级场景（新 Client + 旧 Host + 旧注册文件）由更早的 staleCompatibility 决策（connection.ts:1290-1296，把 hello 协议范围改写为 {max+1,max+1}）决定，仓库内无任何测试覆盖该路径。建议用现有脚手架补 sibling 测试（注册文件写 epoch-1，断言 hello 带 {max+1,max+1}，回复 incompatible，断言 connectRuntimeHost 返回 incompatible），或显式延后（以服务端检查为兜底）。P3 可选：reason 断言锁定 catch-all 分类（语义弱）、约 90 行脚手架只服务一个测试、transport.closed 无界等待。
</details>

Astro-Han:
LGTM — merging. The P2-1 note (no test for the real upgrade-flow `staleCompatibility` decision) is a deferrable test gap per our review standard; the two pinned seams plus the server-side epoch check are sufficient for now. A follow-up sibling test would be welcome.

*AI-assisted review disclosure: this merge decision follows the review comment above (pi review subagent on `ollama-cloud/deepseek-v4-flash`); the human contributor reviewed the final diff and chose to merge.*

Issue #2974, opened 2026-08-13
[TRACKING] Apache Maka 0.2.0-incubating source release


Astro-Han:
<details open>
<summary>English</summary>

## Goal

Track the Apache Maka `0.2.0-incubating` source release from an exact commit through ASF approval and publication.

The single release authority is the source candidate `apache-maka-0.2.0-incubating-src.tar.gz`. Podling and IPMC voters review this exact archive. Everything in this issue is evaluated against its contents, provenance, and reproducibility.

## Scope

This issue covers only the ASF source release:

1. Select an exact immutable commit.
2. Produce the source candidate, SHA-512 checksum, and detached signature.
3. Verify the extracted candidate's provenance, licensing, headers, disclaimer, branding, build, and tests.
4. Publish the candidate to ASF `dist/dev` and run the podling and IPMC votes.
5. Promote the approved source release to ASF `dist/release` and announce it.

The following are separate, nonblocking work and are not part of this source-release exit condition:

- npm convenience artifacts (#3275).
- Desktop convenience artifacts (#3276).
- Standalone CLI binaries or GitHub Release binary assets.
- Apple and Windows code-signing infrastructure (#3414).
- Product installation, upgrade, and platform acceptance.
- Historical GitHub/npm distribution remediation (#3274), unless review identifies a concrete problem in the source candidate.

These artifacts may later be built from an approved source-release commit, but the ASF source release does not wait for them and does not approve their bytes.

## Source candidate contract

The candidate must:

- Identify one exact immutable commit and archive.
- Contain only source and repository material needed to build and test the release.
- Exclude installed dependencies and local workspace state, including `node_modules`, installed runtimes, caches, and external toolchains.
- Exclude compiled third-party executables and benchmark/runtime toolchains that are not needed to build or test Maka.
- Contain no bundled Category X material, including GPL, LGPL, or AGPL dependencies.
- Record the source, license, and required attribution for every third-party, adapted, generated, or non-text input included in the archive.
- Carry accurate root `LICENSE`, `NOTICE`, and `DISCLAIMER-WIP` files for its exact contents.
- Pass all release checks against the extracted archive presented for voting, not merely against the Git checkout.

Runtime dependencies and convenience artifacts have their own distribution obligations. They must not be copied into the source candidate simply to make development or installation more convenient.

## Pre-vote readiness checklist

Run this immediately before sending the podling vote. Every item is mechanical and independently checkable. Do not open a vote if any item fails.

[...]

PR #3140, opened 2026-08-17
docs: declare the bundled Geist webfonts in LICENSE

Joob1n:
Part of #2974 (Phase 3, *Declare Category B dependencies in LICENSE*).

## What ships that LICENSE did not declare

The desktop renderer bundle emits Geist Variable and Geist Mono Variable `.woff2` files, so binary artifacts redistribute OFL-1.1 material with no corresponding declaration. This adds one, in the same shape as the existing `trycua/cua` entry.

## Why a dependency scan does not find it

Both `@fontsource-variable/*` packages are declared as **devDependencies**, and neither appears in `app.asar`. Only the font files vite emits into `dist-renderer` ship. Verified by listing `dist-renderer/assets/*.woff2` — 11 files across the two families — against the packaged bundle.

Flagging it because the same reasoning narrows the rest of that checklist item:

| Audit entry | Verified | Needs a LICENSE declaration? |
|---|---|---|
| `@fontsource-variable/geist*` — OFL-1.1 | ships as `.woff2` in `dist-renderer` | **yes**, added here |
| `lightningcss` — MPL-2.0 | build-time only, not in any artifact | no |
| `caniuse-lite` | build-time only, and **CC-BY-4.0**, not MPL-2.0 | no |

So the item reduces to the fonts. Happy to be corrected if the intent was to declare build-time dependencies too — my reading of the release policy is that LICENSE covers what a release artifact actually redistributes.

## Checks

Lint and format pass. Source-only change, no code paths touched.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -245,8 +245,17 @@ License: SIL Open Font License, Version 1.1
 Maka's desktop renderer bundle embeds Geist Variable and Geist Mono Variable
 webfont files. The packages that carry them are build-time dependencies, but
 the `.woff2` files themselves are emitted into the renderer bundle and are
-therefore redistributed in binary artifacts. The following SIL Open Font
-License, Version 1.1 applies to that material:
+therefore redistributed in binary artifacts. The upstream license text for
+that material follows verbatim:
+
+This Font Software is licensed under the SIL Open Font License, Version 1.1.
+This license is copied below, and is also available with a FAQ at:
+http://scripts.sil.org/OFL
+
+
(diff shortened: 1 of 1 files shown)


likun666661 (approved):
Verified the pinned 5.3.0 package versions against `package-lock.json` and the published npm tarballs. The OFL text matches the upstream package licenses, the bundled `.woff2` files are correctly identified, and the packaging config also carries both upstream license files. CI is green. LGTM.

PR #3154, opened 2026-08-17
fix(scripts): say what drifted when the notices check fails

Joob1n:
## Problem

`npm run check:third-party-notices` compares a 15k-line generated file byte-for-byte and, on a mismatch, says only:

```
Error: Production dependency notices are stale. Run npm run generate:third-party-notices.
```

Two different causes produce that message, and the suggested command only fixes one of them:

- the dependency closure changed — regenerating is the fix;
- the same packages are listed but their notice text moved — regenerating will *also* silently absorb it, without anyone seeing what changed or why.

Whoever hits this is usually reading a CI log for an OS they are not sitting in front of, so there is nothing to act on but a guess.

## Change

The check now names the drift.

Changed closure — packages on each side, capped at ten with a count:

```
Error: Production dependency notices are stale. Run npm run generate:third-party-notices.
  present in the closure but missing from the committed file (1):
    mermaid@11.16.1
```

Same packages, different text — the first differing line, both versions quoted:

```
Error: Production dependency notices are stale. Run npm run generate:third-party-notices.
  the same packages are listed, so the difference is in the notice text itself
  first difference at line 430:
    committed:  "================================================================================"
    generated:  ""
```

A missing output file also gets its own message instead of sharing the stale one, since "run the generator" is the whole answer there.

No dependencies added; the comparison itself is unchanged.

## How I know it works

Both branches were exercised against the real file: deleting one package's section produced the first output above, and the second is the actual current output on `main`.

That second one is not hypothetical. It localized a real failure in one line — same packages, line 430 — which pointed at `LICENSE` content rather than a dependency change, and from there to #3140 having landed without the notices file being regenerated. The Windows `package` job that runs this check does not list `LICENSE` in its trigger paths, so that PR never ran it. I will send the regeneration separately, since it is a different concern from the diagnostics.

Verified with `npm ci` rather than an incremental install, so the closure matches what CI resolves.

## AI use

Tool and scope: Claude Code assisted with the implementation and diagnostics. The human contributor reviewed the work and accepts responsibility for the contribution.

## Checklist

- [x] Required CI checks pass

- [x] Yes — check-mode diagnostics only; notice generation and comparison semantics are unchanged
- [ ] No

### DIFF
--- a/scripts/generate-third-party-notices.mjs
+++ b/scripts/generate-third-party-notices.mjs
@@ -333,12 +333,66 @@ function validateAssetNotices() {
   }
 }
 
+/**
+ * Say what drifted, not just that something did.
+ *
+ * The comparison is byte-exact over a 15k-line generated file, so "stale" on
+ * its own leaves whoever hit it — often on a CI runner whose OS they are not
+ * holding — with nothing to act on. The two shapes worth separating are a
+ * changed dependency closure, which the suggested command fixes, and identical
+ * packages whose license bytes moved, which usually means something upstream
+ * or environmental rather than a missed regeneration.
+ */
+function describeNoticeDrift(committed, generated) {
+  const packagesIn = (text) =>
+    new Set(
+      text
+        .split('\n')
+        .filter((line) => line.startsWith('Package: '))
+        .map((line) => line.slice('Package: '.length)),
+    );
+  const committedPackages = packagesIn(committed);
+  const generatedPackages = packagesIn(generated);
+  const onlyGenerated = [...generatedPackages].filter((name) => !committedPackages.has(name));
+  const onlyCommitted = [...committedPackages].filter((name) => !generatedPackages.has(name));
+
+  const lines = [];
+  const list = (label, names) => {
+    if (names.length === 0) return;
+    const shown = names.slice(0, 10);
+    lines.push(`  ${label} (${names.length}):`);
+    for (const name of shown) lines.push(`    ${name}`);
+    if (names.length > shown.length) lines.push(`    …and ${names.length - shown.length} more`);
+  };
+  list('present in the closure but missing from the committed file', onlyGenerated);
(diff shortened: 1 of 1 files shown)


Joob1n:
Both bots landed on the same defect and both were right — fixed in `315b8f6f`.

I reproduced it before changing anything:

```
CRLF committed → {'a@1\r', 'b@2\r'}
LF   generated → {'a@1', 'b@2'}
only in generated: {'a@1', 'b@2'}
only in committed: {'a@1\r', 'b@2\r'}
```

Every package reported as both added and removed, which also means the text-drift branch never runs. So under CRLF the diagnostic would have misdescribed exactly the environment-specific failure it exists to explain — the worst possible case for this particular tool.

Both sides now split on either ending and trim the extracted name.

While there I added a third outcome. If every line matches once endings are normalized, the difference *is* the endings, and saying "regenerate" is wrong — the repository already forces LF through `.gitattributes`, so a CRLF checkout means that rule did not take effect:

```
  the same packages are listed, so the difference is in the notice text itself
  every line matches once line endings are normalized:
    committed uses CRLF: true
    generated uses CRLF: false
  check .gitattributes and the checkout, not the dependency closure
```

All three paths exercised against the real file: a deleted package section reports `mermaid@11.16.1`; a CRLF-converted copy of otherwise-current output produces the block above; and the actual drift on `main` still reports the same-packages case at line 430.

Note this PR's `package` job fails for a reason outside it — `main`'s notices file has been stale since #3140, which #3155 fixes. This PR is what localized that, and it goes green once #3155 lands.

jackwener:
Fast-path determination on head 1c14c53d: this is a low-impact, easy-to-revert diagnostics-only change. It does not alter notice generation, dependency resolution, or the byte-exact comparison; it only explains a failing check. CRLF handling is present, the AI-use disclosure is now explicit, and every required hosted check is terminal green on this head. I reviewed the final diff and found no blocking correctness issue.

PR #3148, opened 2026-08-17
build(desktop): stop shipping the renderer's dependency tree twice

Joob1n:
## Summary

`app.asar` shipped a second copy of the renderer's dependency sources. Vite emits everything the renderer loads into `dist-renderer`; electron-builder then walked the production dependency closure of `apps/desktop/package.json` and packaged those same packages again, as sources that are never loaded.

The first version of this PR excluded them with a `files` denylist. That was the wrong shape, and @likun666661's review showed why with measurements: a denylist has to name every transitive package too, so excluding `mermaid` does not stop electron-builder from independently collecting the hoisted packages `mermaid` pulls in, and `d3-*` never matched the bare `d3` meta-package. On that head the archive still carried ~69 renderer-only packages.

So the fix moved to the closure itself. The nine renderer-only direct dependencies now live in `devDependencies`, which takes them **and everything only reachable through them** out of the production closure. The denylist is gone rather than extended — there is nothing left for it to catch, and nothing left to keep in sync.

Moved: `@maka/ui`, `react`, `react-dom`, `@astryxdesign/core`, `@astryxdesign/theme-neutral`, `@dnd-kit/core`, `@dnd-kit/sortable`, `@xterm/xterm`, `@xterm/addon-fit`.

`@xterm/headless` and `@xterm/addon-unicode11` stay dependencies: `@maka/runtime` imports them for the PTY stack (`packages/runtime/src/pty-stack.ts`).

Refs #3146

## Verification

Packaged the macOS arm64 app from `main` and from this branch and compared the archives directly.

| | packages in asar | `node_modules` content | asar file |
| --- | ---: | ---: | ---: |
| `main` @ `efe381f4f` | 360 | 294.07 MiB | 309.4 MB |
| this branch | 240 | 123.55 MiB | 135.0 MB |
| delta | **−120** | **−170.52 MiB** | **−174.4 MB (−56.4%)** |

**Zero packages were added.** The largest entries that stop shipping:

```
79.1 MB  mermaid              11.3 MB  @mermaid-js/parser     3.9 MB  katex
18.8 MB  lucide-react          8.9 MB  cytoscape-fcose        2.6 MB  es-toolkit
15.7 MB  @astryxdesign/core    7.0 MB  react-dom              2.6 MB  @maka/ui
                               5.6 MB  @xterm/xterm           1.6 MB  dompurify
```

Every one of them is already inside `dist-renderer`; the copy that leaves is the unread source tree beside it.

[...]

PR #3155, opened 2026-08-17
fix(licenses): regenerate the notices after the bundled-font declaration

Joob1n:
`npm run check:third-party-notices` currently fails on `main`. This regenerates the file it checks.

## Cause

#3140 added the Geist OFL declaration to `LICENSE`. The notices generator folds `LICENSE` into its output, so `THIRD_PARTY_NOTICES.txt` went stale the moment that merged — my PR, my omission.

## Why CI did not catch it

The Windows `package` job is the only caller of `check:third-party-notices`. It runs on pull requests only, and its trigger paths do not include `LICENSE`:

```
07:30  #3140 merges — touches LICENSE alone, so this job never runs
   ↓
       main is now stale, and nothing on main re-runs the check
   ↓
08:38  #3148 touches electron-builder.config.mjs, which *is* a trigger path
       → the job wakes up and fails on debt from a different PR
```

So the failure surfaced on an unrelated PR, two paths and an hour removed from what caused it. Worth noting for anyone deciding whether that job should also run on `main`, or whether `LICENSE` belongs in its trigger paths — I have not changed either here, since both are policy calls.

## The change

Generator output only. 107 lines added, none removed, all of them the OFL text #3140 introduced.

Verified with `npm ci` rather than an incremental install, so the dependency closure matches what CI resolves; `check:third-party-notices` passes afterward.

Sent separately from #3154, which makes this class of failure self-describing — it is what localized this one, in one line, to LICENSE content rather than a dependency change.

Unblocks #3148.

### DIFF
--- a/apps/desktop/resources/licenses/npm/THIRD_PARTY_NOTICES.txt
+++ b/apps/desktop/resources/licenses/npm/THIRD_PARTY_NOTICES.txt
@@ -427,6 +427,112 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 
+
+Geist and Geist Mono (bundled webfonts)
+
+Source: https://github.com/vercel/geist-font
+Emitted by: @fontsource-variable/geist 5.3.0, @fontsource-variable/geist-mono 5.3.0
+Copyright 2024 The Geist Project Authors
+License: SIL Open Font License, Version 1.1
+
+Maka's desktop renderer bundle embeds Geist Variable and Geist Mono Variable
+webfont files. The packages that carry them are build-time dependencies, but
+the `.woff2` files themselves are emitted into the renderer bundle and are
+therefore redistributed in binary artifacts. The upstream license text for
+that material follows verbatim:
+
+This Font Software is licensed under the SIL Open Font License, Version 1.1.
+This license is copied below, and is also available with a FAQ at:
+http://scripts.sil.org/OFL
+
+
+-----------------------------------------------------------
+SIL OPEN FONT LICENSE Version 1.1 - 26 February 2007
+-----------------------------------------------------------
+
+PREAMBLE
+The goals of the Open Font License (OFL) are to stimulate worldwide
+development of collaborative font projects, to support the font creation
+efforts of academic and linguistic communities, and to provide a free and
+open framework in which fonts may be shared and improved in partnership
+with others.
+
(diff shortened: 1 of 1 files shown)


Astro-Han:
Thanks for the regeneration — I ran the actual generator (`npm run generate:third-party-notices` → `scripts/generate-third-party-notices.mjs`) against the PR head tree and byte-compared the output with the committed file: zero bytes differ. The change is exactly what the canonical flow produces (the Geist OFL block lands inside the `@ai-sdk/provider-utils@5.0.27` section via the `--- VERSION-PINNED LICENSE TEXT OVERRIDE ---`, since the generator folds repo LICENSE into that package's section — same mechanism as the existing trycua/cua entry). I also cross-checked the other notices artifacts (cargo generator, `THIRD_PARTY_LICENSES.txt` for vendored SVGs, git NOTICE) — none needed updating.

**Conclusion: PASS — no blocking findings (no P0/P1/P2).**

P3 (optional): the generated npm notices path (`apps/desktop/resources/licenses/npm/THIRD_PARTY_NOTICES.txt`) is missing from the trigger list in `.github/workflows/release-windows-check.yml:22-36` (only the cargo notices and git NOTICE are listed), so this PR's own change never ran the owning `package` job — the body transparently defers the LICENSE-trigger gap, but the npm path itself is also outside the triggers. A one-line trigger addition would keep the next LICENSE-adjacent drift visible; otherwise fine as a stated deferral. Also, the body says "107 lines added"; the actual diff is 106 additions.

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagent on `ollama-cloud/deepseek-v4-flash`). The subagent executed the real generator at the PR head tree and verified byte-identical output; the P3 CI-trigger point is from static workflow inspection. Please weigh these findings with your own judgment.*

<details>

结论：PASS，无阻塞项。subagent 实际运行生成脚本（npm run generate:third-party-notices）并与 PR 提交的文件逐字节比对——零差异，改动就是规范流程的输出（Geist OFL 块经 VERSION-PINNED LICENSE TEXT OVERRIDE 落入 @ai-sdk/provider-utils 段，与既有 trycua/cua 条目同机制）。其他 notices 制品（cargo 生成器、SVG 的 THIRD_PARTY_LICENSES.txt、git NOTICE）均无需更新，无遗漏。P3（可选）：npm notices 路径不在 release-windows-check.yml 的 trigger 列表（只列了 cargo notices 和 git NOTICE），本 PR 自身改动从未跑过 package job——body 已显式延后 LICENSE trigger 缺口，但 npm 路径本身也在 trigger 外，建议补一行或显式声明延后；另外 body 说 "107 lines" 实际是 106。
</details>

Astro-Han:
LGTM. Merging!

PR #3170, opened 2026-08-17
fix(ui): restore usable scrolling for one-line code blocks

MicroGery:
## Summary

- move the copy action into a dedicated CodeBlock toolbar so it no longer covers long single-line content
- keep the Astryx native overflow viewport and platform scrollbar instead of adding a second scrolling implementation
- prevent Turn-level pointer capture from stealing native scrollbar gestures or edge-selection autoscroll
- vertically center one-line code while preserving multiline CodeBlock behavior

## Verification

- `npm run rebuild`
- `npm --workspace @maka/ui test` — 182 passed
- `npm run lint -- --diagnostic-level=error`
- `npm --workspace @maka/desktop run e2e -- e2e/code-scroll.spec.ts` — 1 passed
- manually compared the latest `origin/main` behavior with this branch in the desktop GUI

## Root cause

Astryx places the copy button absolutely when a plaintext CodeBlock has no header, so the control can cover the end of a long one-line value. Independently, Maka captures pointer gestures on the enclosing Turn for quote selection; that retargets pointer movement away from the nested code viewport and prevents Chromium from continuing native scrollbar or edge-selection scrolling. The fix gives CodeBlock a structural toolbar and lets its native viewport retain ownership of scrolling and selection gestures.

## AI use

Select exactly one:

Tool(s) and scope: OpenAI Codex assisted with root-cause analysis, implementation, regression tests, rebase, and local verification. The contributor reviewed the resulting changes.

## Checklist

- [ ] No


Astro-Han:
Thanks for the fix — the broken scrolling is real and the diagnosis is precise (the absolute-positioned copy button covers the top-right of untitled code blocks, the turn-level `setPointerCapture` in the quote hook swallows scrollbar dragging and cross-edge selection autoscroll, and macOS's hidden overlay scrollbar makes the row unreachable), and the cure is scoped correctly: `:has()`/`title=""` toolbar structure, pointer-capture exemption, and OverlayScrollbars in custom-elements mode (keeping the native viewport so selection/keyboard scrolling aren't lost) are three distinct fixes for three distinct causes, with the e2e directly exercising the four previously-broken interactions (scrollbar drag, track click, keyboard, selection autoscroll) — verified against the Astryx source that the theme mapping is right (`os-theme-light` on dark app), that vertical wheel still chains to the chat scroller, and that quote-from-code survives (selectionchange path untouched). CI is green.

**Conclusion: PASS — no P0/P1/P2.**

**P3-1 (merge gate): the PR has no before/after screenshots.** This visually changes every code block (plain-text blocks gain a 32px bordered toolbar, single-line blocks grow to ~72px, language-block content shifts down as the Astryx `-spacing-2` pull-up is neutralized, long lines gain a persistent scrollbar). Per the repo gate, UI changes need before/after screenshots — please add light/dark pairs.

P3 (optional): `title=""` also enables the structural header for plain-text blocks, which with the always-passed `isCollapsible` makes plain-text blocks ≥10 lines collapsible for the first time — a new interaction not mentioned in the PR (acceptable for consistency, but please confirm it's intended); the scrollbar slot overlays a 10px strip at the bottom (`z-index:1`) and a pointerdown there triggers clickScroll instead of text selection/copy — worth a CJK long-command screenshot to confirm; OverlayScrollbars (~30KB) buys "persistent themed scrollbar + touch/pen drag" over the native thin-scrollbar path — defensible as a product choice but the tradeoff isn't stated in the PR; the e2e doesn't verify the key nested-scrolling promise (hovering a code block doesn't break the outer chat vertical scroll) or toolbar copy-button clickability — per convention, test gaps are P3; the custom ArrowLeft/Right keydown handler in markdown-body.tsx is redundant with the native keyboard scroll on the tabIndex=0 viewport and unconditionally preventDefaults (also swallowing selection-collapse+scroll when a selection exists) — either delete it or only intercept when there's remaining scroll in that direction.

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagent on `opencode-go/deepseek-v4-flash`), which traced the spacer/ChatLayout geometry, the pointer-capture path, and verified OS 2.16 / Astryx 0.4.0 source. P3-1 is from the PR having no screenshots. Please weigh these findings with your own judgment.*

<details>

结论：PASS（无 P0/P1/P2）。问题三源精确定位：绝对定位复制按钮盖住无标题块右上角 + turn 级 setPointerCapture 吞掉滚动条拖拽/跨边缘选区 autoscroll + macOS 隐藏 overlay scrollbar 使长行不可达；修复各就各位（工具栏化+指针豁免+OverlayScrollbars custom-elements 模式保留原生 viewport 保选区/键盘滚动），e2e 直接验证四个曾被破坏的交互（滚动条拖拽/轨道点击/键盘/选区 autoscroll），主题映射/垂直滚不被吞/quote-from-code 存活均已核对。CI 全绿。P3-1（合并 gate）：无 before/after 截图——本 PR 视觉上改变每个代码块（纯文本块新增 32px 带边框工具栏、单行块增高至 ~72px、语言块内容因中和 Astryx -spacing-2 上拉而下移、长行新增持久滚动条），按仓库规则需要 light/dark 前后截图。P3（可选）：title="" 使纯文本块首次拥有 header，配合恒传 isCollapsible 使 ≥10 行纯文本块可变折叠——新交互 PR 未提及，需确认是有意；滚动条 slot 覆盖底部 10px 条带（z-index:1），该处 pointerdown 触发 clickScroll 而非文本选区/复制——建议用 CJK 长命令截图核实；OverlayScrollbars（~30KB）换持久主题滚动条+触控拖拽是可选产品决策但 PR 未陈述取舍；e2e 未验证嵌套滚动核心承诺（悬停代码块不破坏外层 chat 垂直滚动）与工具栏复制按钮可点；markdown-body.tsx 自定义 ArrowLeft/Right handler 与 tabIndex=0 viewport 的原生键盘滚动冗余且无条件 preventDefault（含提示词选区折叠+滚动），建议删除或仅在仍有滚动余量时拦截。
</details>

Astro-Han (commented):
The scrolling solution is well targeted: it preserves Astryx's native focusable viewport, isolates OverlayScrollbars to one-line blocks, and covers pointer, keyboard, selection, thumb, and track paths without building a parallel code-block component. Two boundary details remain.

From first principles, a control needs a name and a shipped dependency needs a stable reviewed license identity. For headerless plaintext, either do not expose collapse or provide a localized accessible name; for OverlayScrollbars, make the version/notice contract explicit rather than combining a moving caret range with an exact-version override.

Review performed with two Codex reviewer agents and DeepSeek V4 Flash as advisory tools; I verified the findings against the latest head and live green CI.

滚动方案很克制：保留 Astryx 原生可聚焦 viewport，只对单行 block 隔离使用 OverlayScrollbars，并覆盖 pointer、keyboard、selection、thumb 和 track 路径，没有新建并行 code-block 组件。仍有两个边界问题。

从第一性原理看，control 必须有名称，发布依赖必须有稳定、已审查的 license identity。对无 header 的 plaintext，要么不暴露 collapse，要么提供本地化 accessible name；对 OverlayScrollbars，应显式固定 version/notice 契约，而不是把移动的 caret range 与精确版本 override 组合。

本次审查使用了两位 Codex reviewer agents 与 DeepSeek V4 Flash 作为辅助工具；我已依据最新 head 和实时绿色 CI 复核问题。

</details>

packages/ui/src/markdown-body.tsx
```diff
@@ -186,17 +196,109 @@ function MarkdownCode(props: {
...
+      className={`maka-markdown-code maka-markdown-code-${props.density}`}
+      data-maka-code-layout={isSingleLine ? 'single-line' : 'multi-line'}>
       <CodeBlock
         code={props.code}
         language={props.language}
+        // Astryx otherwise overlays the copy button on headerless plaintext.
+        // An empty title deliberately enables its structural header without
+        // inventing a visible or unlocalised label; language fences still show
+        // their normal language label in the same toolbar.
+        title=""
```
Astro-Han: **P2 — Long plaintext blocks get an unnamed collapse button.** With `title=""` and unconditional `isCollapsible`, Astryx renders a focusable `role=button` header once a plaintext block reaches its collapse threshold, but the empty title and decorative chevron provide no accessible name. Screen readers announce only “button.” Disable collapse for headerless/plaintext blocks or provide a localized accessible label; add a 10+ line plaintext accessibility assertion.

packages/ui/package.json
```diff
@@ -28,6 +28,7 @@
     "clsx": "^2.1.1",
     "lucide-react": "^1.28.0",
     "mermaid": "^11.16.0",
+    "overlayscrollbars": "^2.16.0",
```
Astro-Han: **P2 — This range can move past the exact reviewed license override.** The notice generator's special-case text is keyed to `overlayscrollbars@2.16.0`, while `^2.16.0` permits later versions whose tarballs/notice obligations may differ. A routine lock refresh can therefore fail notice generation or change shipped attribution without an explicit dependency review. Pin `2.16.0`, or make the override/version upgrade contract explicit and tested.

apps/desktop/e2e/code-scroll.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+    return {
+      rect: { x: rect.x, y: rect.y, width: rect.width, height: rect.height },
+      clientWidth: node.clientWidth,
+      scrollWidth: node.scrollWidth,
+      topInset: textRect.top - rect.top,
+      bottomInset: rect.bottom - textRect.bottom,
+    };
+  });
+  expect(metrics.scrollWidth).toBeGreaterThan(metrics.clientWidth);
+  expect(Math.abs(metrics.topInset - metrics.bottomInset)).toBeLessThanOrEqual(1);
```
Astro-Han: **P1 — Replace this glyph-bounds assertion with a layout metric.** `Range.getBoundingClientRect()` measures font ink ascent/descent, which is not expected to be symmetric inside the viewport; required CI receives a 10px difference and fails. Delete this assertion or measure the actual line wrapper/layout box after visual verification. Do not alter production padding merely to satisfy glyph metrics.

Astro-Han (commented):
The move back to native scrolling removes the earlier dependency/license drift, and polling is better than the fixed geometry timeout. The latest required E2E still fails because the new vertical-centering assertion measures glyph ink bounds rather than the layout box it intends to constrain.

Following Occam's razor, delete this low-quality metric or replace it with the actual line wrapper/viewport layout box after visual verification; changing production spacing to satisfy font ascent/descent asymmetry would optimize for the test rather than the user-visible requirement. The existing accessible-name finding on plaintext collapse is already public, so I am not duplicating it here.

Reviewed with Codex using two independent review passes and DeepSeek V4 Flash as an external adversarial pass; I verified the metric and live E2E failure against this exact head.

回退到原生滚动已消除之前的依赖/许可证漂移，polling 也优于固定等待。但最新 required E2E 仍失败，因为新的垂直居中断言测量的是 glyph ink bounds，而不是它真正想约束的 layout box。

按奥卡姆剃刀，应删除这个低质量指标，或经视觉核验后改测真实 line wrapper/viewport box；不要为了字体 ascent/descent 不对称去修改生产布局。plaintext collapse 的 accessible name 问题已有公开评论，这里不重复。

本次由 Codex 两轮独立审查，并使用 DeepSeek V4 Flash 做外部对抗审查；我核对了当前 head 和 E2E 失败。

</details>

apps/desktop/e2e/code-scroll.spec.ts
Astro-Han: [P1] This new required E2E currently fails on the exact head (`Received: 10` in run 32152606634). `Range.getBoundingClientRect()` measures font glyph ink, whose ascent/descent is inherently asymmetric and platform-dependent; the production CSS centers the line/layout box, not glyph ink. Please delete this assertion or measure the actual line wrapper/layout box instead, then rerun the required E2E. Do not distort production spacing to satisfy a font-dependent metric.

packages/ui/src/markdown-body.tsx
```diff
@@ -175,6 +176,7 @@ function MarkdownCode(props: {
   density: 'default' | 'compact';
   renderMermaid: boolean;
 }) {
+  const copy = getSharedUiCopy(useUiLocale()).markdown;
```
Astro-Han: [P2] This makes every non-Mermaid code block—and even Mermaid before the early return—require Maka's `LocaleProvider`. `MarkdownBody` is publicly exported, while `useUiLocale()` throws without that provider, so previously valid standalone consumers now crash merely by rendering a code fence. Please keep the localization need behind an existing provider boundary or provide the component's established fallback contract, and add a standalone public-component regression test.

Astro-Han (commented):
The production scrolling seam is mostly right: Astryx keeps ownership of the native viewport, and the pointer-capture exception is narrowly scoped. The current head is not ready because its new required E2E asserts font glyph-ink symmetry rather than the layout box the CSS centers, and that exact assertion is failing on Linux CI. I also found a public-component compatibility regression from making `useUiLocale()` unconditional for every code block.

AI-assisted review: Codex coordinated two independent reviewer passes and an OpenCode Go DeepSeek V4 Flash high-effort adversarial pass. I verified exact head `15e89b85f99ddd27f3d3477cf2e2098ad2d4d1e4`, Astryx/CSS ownership, the public export and locale hook, and the failing required E2E. No local tests were run.

<details>
<summary>中文审查</summary>

生产滚动 seam 基本正确：原生 viewport 仍由 Astryx 管理，pointer-capture 豁免范围也足够窄。当前不能合并，因为新增 required E2E 测量的是字体 glyph ink 对称性，不是 CSS 居中的 layout box；该断言正在 Linux CI 失败。另外，所有代码块无条件调用 `useUiLocale()`，给公共组件新增了未声明的 Provider 硬依赖。

本次为 AI 辅助审查，已核验精确 head、Astryx/CSS 责任边界、公共导出、locale hook 和失败的 required E2E；未运行本地测试。

</details>

packages/ui/src/__tests__/markdown-body.test.ts
```diff
@@ -26,6 +27,69 @@ it('keeps raw HTML inert instead of expanding the Markdown trust surface', () =>
...
+      text: ['```', `ssh-ed25519 ${'A'.repeat(200)}`, '```'].join('\n'),
+    }),
+  }));
+
+  const toolbarIndex = markup.indexOf('astryx-codeblock-header');
+  const copyButtonIndex = markup.indexOf('astryx-codeblock-copy-button');
+  const scrollViewportIndex = markup.indexOf('role="group"');
+
+  assert.match(markup, /data-maka-code-layout="single-line"/);
+  assert.doesNotMatch(markup, /maka-markdown-code-scrollbar-slot/);
```
Astro-Han: [P3] This assertion is leftover negative coupling to the removed OverlayScrollbars iteration: `maka-markdown-code-scrollbar-slot` is not produced anywhere on this head, so the assertion cannot protect the native-scroll behavior. Please delete this line and the identical multiline assertion at line 58; the positive toolbar/viewport and E2E assertions already cover the shipped contract.

Astro-Han (approved):
This is now a focused fix at the existing Astryx seam: the copy control remains in the native CodeBlock toolbar, the one-line body keeps the native horizontal viewport and keyboard/selection behavior, and only pointer gestures originating inside that viewport bypass Turn-level pointer capture. The locale/provider and layout-metric regressions from earlier heads are resolved.

I found no reproducible correctness issue on this head. All current required checks are green. The inline P3 is only an Occam cleanup and does not block approval.

AI-assisted review disclosure: Codex performed the final review using two independent reviewer passes and OpenCode Go DeepSeek V4 Flash (high effort) as an adversarial advisory pass. I verified the Astryx DOM seam, selection/scroll boundary, accessibility name, exact head `dce234bd0`, and live CI.

当前实现已经收敛到 Astryx 的既有扩展点：copy control 留在原生 CodeBlock toolbar；单行代码继续使用原生横向 viewport、键盘和文本选择；只有从该 viewport 发起的 pointer gesture 才跳过 Turn 层 pointer capture。此前 head 的 locale/provider 与布局度量问题都已解决。

当前 head 未发现可复现的正确性问题，所有必需检查均已通过。行内 P3 仅是奥卡姆剃刀式清理，不阻塞 approve。

AI 辅助审查说明：Codex 使用两轮独立 reviewer 审查，并以 OpenCode Go DeepSeek V4 Flash（high effort）进行对抗性辅助审查；我已核对 Astryx DOM seam、selection/scroll 边界、无障碍名称、精确 head `dce234bd0` 与实时 CI。

</details>

jackwener:
Refreshed this branch onto current `main@0ef1c55b5` and pushed head `46a36deeb`.

The product change is unchanged. The only review follow-up is the approved P3 cleanup: I removed the two negative assertions coupled to the abandoned `maka-markdown-code-scrollbar-slot` implementation name. Positive toolbar/viewport coverage and the behavioral E2E remain.

Fresh verification on this exact head:

- `npm --workspace @maka/ui test` — 188 passed
- `npm --workspace @maka/desktop run e2e -- e2e/code-scroll.spec.ts` — 1 passed
- all required hosted checks green

@Astro-Han, please re-review the refreshed exact head so the independent human-review gate can close.

Astro-Han (approved):
The refreshed head `46a36de` preserves the previously reviewed production behavior and applies only the agreed Occam cleanup: the two assertions coupled to the abandoned scrollbar implementation name are removed. I rechecked the exact delta from the approved head, the current PR diff, prior findings, and live required checks. The accessibility, standalone-provider, layout-metric, native-scroll, pointer-selection, and deterministic-E2E boundaries remain resolved.

AI-assisted review disclosure: Codex verified the refreshed exact head and current GitHub state; the human maintainer reviewed the final result and authorized approval and merge.

<details>
<summary>中文评论</summary>

刷新后的 `46a36de` 保留了此前已审查的生产行为，只执行了已同意的奥卡姆剃刀式清理：删除两条与废弃 scrollbar 实现名耦合的断言。我重新核对了相对已批准 head 的精确变化、当前 PR diff、既有 findings 和实时 required checks。无障碍名称、standalone provider、布局度量、原生滚动、pointer selection 与确定性 E2E 边界仍然全部解决。

AI 辅助审查说明：Codex 核验了刷新后的精确 head 和当前 GitHub 状态；人工维护者审阅最终结果并明确授权 approve 与 merge。

</details>

PR #3171, opened 2026-08-17
fix(desktop): contain narrow workbar content

ARE404:
## Summary

Keep the launcher and side-chat composer inside the 320px minimum right workbar. An explicit shrinkable launcher frame owns the width constraint instead of relying on the headed Astryx `List` wrapper, which otherwise resolves to its 360px max-content width and clips shortcut hints.

Electron coverage directly measures the shortcut hint containers and composer send button at the narrow boundary.

## Verification

- `npx biome check apps/desktop/src/renderer/session-workbar.tsx apps/desktop/src/renderer/styles/chat-detail.css apps/desktop/e2e/session-workbar.spec.ts` — passed
- `npx playwright test --config e2e/playwright.config.ts e2e/session-workbar.spec.ts` — 3 passed
- `npm --workspace @maka/desktop run typecheck` — passed before review
- `npm --workspace @maka/desktop run build:renderer` — passed before review

## Review focus

The wrapper avoids coupling product CSS to Astryx's internal headed-list DOM. The regression test targets actual `kbd` hint containers and uses tolerant geometry polling, incorporating both resolved review findings.

## AI use

Select exactly one:

Tool(s) and scope: OpenAI Codex diagnosed the workbar layout, implemented and tested the containment fix, incorporated review feedback, and drafted this description. The human contributor reviewed the work, chose to submit it, and remains responsible for its accuracy, provenance, and licensing.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/e2e/session-workbar.spec.ts
+++ b/apps/desktop/e2e/session-workbar.spec.ts
@@ -16,13 +16,16 @@ async function openGitChanges(page: Page) {
 
 async function setRightWorkbarWidth(page: Page, width: number) {
   const workbar = page.locator('.maka-session-workbar[data-placement="right"]');
+  await expect(workbar).toBeVisible();
   await workbar.evaluate((element, nextWidth) => {
     (element as HTMLElement).style.setProperty(
       '--maka-session-workbar-width',
       `${nextWidth}px`,
     );
   }, width);
-  await expect.poll(async () => (await workbar.boundingBox())?.width).toBe(width);
+  await expect
+    .poll(async () => (await workbar.boundingBox())?.width)
+    .toBeCloseTo(width, 0);
   return workbar;
 }
 
@@ -41,7 +44,7 @@ test('narrow right workbar keeps launcher shortcuts and side-chat send button in
 
   const workbarBox = await workbar.boundingBox();
   const shortcutBoxes = await launcher
-    .getByRole('img')
+    .locator('kbd')
     .evaluateAll((elements) =>
       elements.map((element) => {
         const box = element.getBoundingClientRect();
--- a/apps/desktop/e2e/session-workbar.spec.ts
+++ b/apps/desktop/e2e/session-workbar.spec.ts
@@ -14,6 +14,76 @@ async function openGitChanges(page: Page) {
   return page.getByRole('region', { name: 'Git 变更' });
 }
 
+async function setRightWorkbarWidth(page: Page, width: number) {
+  const workbar = page.locator('.maka-session-workbar[data-placement="right"]');
+  await workbar.evaluate((element, nextWidth) => {
+    (element as HTMLElement).style.setProperty(
+      '--maka-session-workbar-width',
+      `${nextWidth}px`,
+    );
+  }, width);
+  await expect.poll(async () => (await workbar.boundingBox())?.width).toBe(width);
+  return workbar;
+}
+
+test('narrow right workbar keeps launcher shortcuts and side-chat send button inside', async ({
+  window: page,
+}) => {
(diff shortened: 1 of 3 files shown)


Astro-Han:
Thanks for the follow-up — the core fix is correct: an explicit shrinkable width owner (`.maka-workbar-launcher-frame` with `width: 100%; min-width: 0` as the grid item) collapses the whole inner chain (Astryx bare flex wrapper → ul → items) to the container width, so the launcher shortcut hint no longer overflows the 320px workbar. The 360px cap keeps wide-window behavior unchanged, the scoping is clean (no leakage to other placements), the e2e asserts the kbd element actually lands inside the 320px workbar (same `SESSION_WORKBAR_MIN_WIDTH = 320` constant as the resize clamp), and CI is green.

**Conclusion: PASS — one P2 carried over from #3160 (unchanged), two P3s.**

**P2 — the #3160 narrow-window drag-region issue is untouched and unre-deferred.** The rules at `chat-detail.css:114-132` (`z-index: calc(var(--z-titlebar)+1); height: var(--maka-plate-titlebar-clearance); padding-inline: ... var(--maka-titlebar-overlay-right-width); -webkit-app-region: drag`, keyed on `[data-placement="right"] > .maka-session-workbar-toolbar`) still apply when `@media (max-width: 990px)` moves the workbar to the window bottom (`data-placement` stays "right" — placement is static state). Result: a 32px window-drag strip across the bottom-placed workbar (drag on empty space moves the whole window), and the collapse button sits ~160–200px from the right edge on Windows. Since this PR is explicitly the "narrow workbar follow-up" to #3160, it's the natural place to close that gap — either scope those rules to `@media (min-width: 991px)` (or a layout attribute) and add a narrow-window e2e, or at least note the explicit deferral in the PR body.

P3 (optional): the e2e runs on the fixed 1240×820 fixture with the panel squeezed to 320px via a CSS variable — the `@media (max-width: 990px)` layout still has zero coverage (that's the layout the P2 lives in); `.maka-workbar-launcher-list { min-width: 0 }` is a no-op (the class lands on the inner `ul`, where min-width doesn't participate in automatic minimum sizing — the real shrink chain is frame → wrapper → ul `width: 100%`), harmless but removable.

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagent on `opencode-go/deepseek-v4-flash`). The subagent verified the shrink chain, placement selectors, media query, and e2e fixture statically; the narrow-window behavior is a prediction from CSS inspection, not an observed failure. Please weigh these findings with your own judgment.*

<details>

结论：PASS（1 个从 #3160 延续的 P2 + 2 个 P3）。核心修复正确：显式可收缩宽度所有者（.maka-workbar-launcher-frame，width:100%; min-width:0）让整条内部链塌缩到容器宽度，320px 面板下 launcher 快捷键提示不再溢出；360px 封顶保持宽窗口行为不变；e2e 直接断言 kbd 落在 320px workbar 内（同 SESSION_WORKBAR_MIN_WIDTH=320 常量），CI 全绿。P2：#3160 的窄窗口（≤990px）drag region 问题原样保留、未收口——底部 workbar 顶部 32px 带仍是窗口级拖拽面、Windows 上收起按钮距右缘 ~160-200px 死区；本 PR 正是"窄窗口后续修复"，建议顺手收进 @media (min-width: 991px) 并补窄窗口 e2e，或至少在正文写明延后。P3：e2e 固定 1240×820 fixture，@media (max-width: 990px) 布局零覆盖；.maka-workbar-launcher-list min-width:0 是无操作声明（className 落在内层 ul 上，min-width 不参与自动最小尺寸）。
</details>

Astro-Han (approved):

PR #3174, opened 2026-08-17
feat(windows-sandbox): place sandboxed children on a private desktop

liugddx:
## What & why

Closes the **private-desktop** half of the umbrella #2142 Phase 4 ③ sandbox hardening — specifically initial private-desktop **placement** plus a hardened per-launch desktop object. #3161 delivered the production-identity readiness probe and deferred the private desktop to a follow-up (RFC v1 §6.5); this is that follow-up.

Non-interactive AppContainer workers previously inherited the creator's interactive `Winsta0\Default` desktop as their *initial* desktop. AppContainer fences files, network and token, but does **not** choose a non-interactive desktop — so code that never re-attaches still ran amid the user's interactive windows. RFC v1 §6.3 promised a private desktop; it was designed but not enforced.

## What this does

- Each AppContainer launch — both the production `create_appcontainer_child` path and the `probe_appcontainer_child` readiness path — creates a **per-launch alternate desktop** on the existing window station (`CreateDesktopExW`, bounded 512 KiB heap so ten-way concurrency stays ~5 MiB of the documented 48 MiB system budget; name = `maka-sandbox-desktop.<pid>.<128-bit CSPRNG nonce>`, RNG failure fails the launch closed).
- Its security descriptor: a leading **deny ACE** strips `DESKTOP_SWITCHDESKTOP`/`HOOKCONTROL`/journal (0x138) from the launching-user SID (which the AppContainer child carries as an effective SID); allow ACEs grant the launching user and Local System control for cleanup and the child's AppContainer SID only the minimal `DESKTOP_*` rights; an explicit **Low no-write-up mandatory label** makes the granted rights pass MIC at the child's Low IL.
- `STARTUPINFOEXW.lpDesktop` points the child at it; the handle lives until the child settles, then `CloseDesktop`.
- **Fail closed**: if the desktop, its DACL, or its name nonce can't be built, the launch fails — `auto`/`require` never fall back to the host desktop.
- The `--boundary-probe` self-attests its initial desktop name against the launcher-owned `maka-sandbox-desktop.` prefix; `--readiness-probe` emits a machine-readable attestation (exact SID, specific Job, settlement, desktop placement) that `readiness-probe-smoke.ps1` asserts field by field.
- `windows_launcher_desktop_tests.rs` covers the SDDL builder (deny-before-allow, minimal app-SID mask, Low label), the placement predicate, and a real-OS test holding **ten live desktops with ten distinct names** inside the heap budget.

## Scope of the guarantee (narrowed per review)

This PR ships initial-desktop **placement** plus a DACL/label-protected per-launch desktop object — **not** escape-proof confinement:

- ✅ **Enforced:** the worker *starts* on a launcher-created private desktop, never the interactive `Default`; the desktop object is protected (deny ACE, minimal app mask, Low label, protected DACL) and per-launch unique (CSPRNG nonce; `CreateDesktopExW` would silently reopen an existing name, so uniqueness is part of the object contract).
- ⚠️ **Not claimed:** `lpDesktop` selects only the *initial* desktop. Nothing structural stops in-process code from `OpenDesktopW("Default")` + `SetThreadDesktop` re-attaching — keeping a worker *unable* to enumerate/message/hook the user's windows after an escape attempt requires the deferred no-Win32k / dedicated window-station / token-boundary gates (§6.5).
- ⏸️ **Deferred (§6.5): clipboard isolation.** The clipboard belongs to the *window station*, which both desktops still share.
- ⏸️ **Deferred (§6.5): full window-station separation and no-Win32k mitigation.**
- ⏸️ **Deferred (§6.5): end-to-end Low-IL GUI proof.** The Low label makes the granted create-window/write rights pass MIC, but no probe creates a window in-child; the worker does no GUI work, so this is labeled-usable, not proven.

Digest is unchanged — the desktop is a launch-time detail and does not enter the manifest, so no `profile_digest_mismatch`.

## Stacked on #3161

This PR is **stacked on #3161** (it modifies the `probe_appcontainer_child` code that #3161 introduces). Until #3161 merges, the diff here includes its commits; they drop out automatically once #3161 lands. **Please merge #3161 first.**

## Verification (Windows 11 x64, local)

- `cargo fmt --check` clean; `cargo build --locked`; `cargo test --locked` → 49 pass (incl. desktop SDDL/placement tests and the ten-distinct-desktops live test).
- Runtime behavior (child initially on `maka-sandbox-desktop.*`, file/network/Job boundary intact, readiness attestation fields) is proven by `appcontainer-smoke.ps1` and `readiness-probe-smoke.ps1` on the real `windows-sandbox-w0` CI lane.
- `@maka/core` + `@maka/runtime` rebuild clean; `default-sandbox-manager` node test → 11 pass.

## RFC

`windows-sandbox-rfc-v1.md` and `.zh-CN.md` §6.3/§6.4/§6.5 updated: the normative guarantee is stated as initial-desktop placement (no unconditional "cannot" claims); the deny ACE, Low label, heap bound, and readiness attestation are recorded as enforced; no-Win32k, dedicated window station, clipboard isolation, and in-child window-creation proof stay on the deferred-gates list.

## AI use

Select exactly one:

Tool(s) and scope: Claude Fable 5 (Anthropic), driven by the human contributor of record, authored the private-desktop placement (`create_confined_desktop`, the two launch-path wirings, the boundary-probe self-attestation), the smoke assertions, the Rust unit tests, and the RFC updates, all under human review. The commit carries a `Generated-by: Claude Fable 5` trailer; the final squash commit must retain it.


hqhq1025 (changes requested):
Request changes on the current head `c999eefd4`.

The alternate-desktop implementation is a useful partial hardening step, but the security guarantee currently documented by this PR is stronger than the mechanism:

`CreateDesktopW` creates the new desktop on the launcher's current window station, and the RFC explicitly defers a dedicated window station. On Windows, windows/messages/hooks are desktop-scoped, but the clipboard belongs to the window station. Therefore this change can isolate the child from `Default` windows and desktop hooks, but it does not by itself establish the claimed clipboard isolation while both desktops remain in the same window station.

The current `desktopIsolated` attestation only proves that the desktop name is non-empty and not `Default`; the smoke additionally proves the generated prefix. Neither test attempts `OpenClipboard`/`GetClipboardData`, proves a different window station, or proves that the child cannot reopen `Winsta0\\Default`.

Please either:

- narrow the RFC/code comments and PR claim to the desktop-scoped guarantees actually enforced here, with clipboard isolation remaining deferred; or
- create a dedicated window station plus desktop under one RAII lifecycle, pass the full `winsta\\desktop` name, and add negative tests for clipboard access, opening/switching to `Default`, child escape, and residual object cleanup.

This PR is stacked on #3161, whose readiness ownership and failure-cleanup blockers also need to be resolved before this can merge.

experiments/windows-sandbox/launcher/src/windows_launcher.rs
```diff
@@ -137,6 +144,207 @@ pub fn launch_appcontainer(request: &LaunchRequest) -> Result<u8, String> {
...
+unsafe fn probe_appcontainer_child(
+    cmd_path: &str,
+    job: HANDLE,
+    app_container_sid: *mut c_void,
+) -> Result<u8, String> {
+    let executable = wide(cmd_path);
+    // Build the command line directly: cmd.exe's `/c` switch must stay
+    // unquoted, so the per-token quoting used for arbitrary workloads would
+    // make cmd treat "/c" as a program to run and exit non-zero.
+    let mut command = wide(&format!("\"{cmd_path}\" /c exit 0"));
```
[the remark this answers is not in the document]
liugddx: Analyzed against this PR's scope (private-desktop). This finding lands on the readiness-probe seam introduced by the stacked #3161, not on the private-desktop change — but assessing it anyway:

The probe **deliberately** launches `cmd.exe /c exit 0` with **System32 as its working directory**. The `STATUS_DLL_INIT_FAILED` (0xC0000142) case documented in `builtin-tools.ts` applies to arbitrary shell workloads whose *inherited* cwd is default-denied inside a capability-less AppContainer. System32 is readable by `ALL APPLICATION PACKAGES`, so the AppContainer child's DLL init succeeds and `/c exit 0` returns cleanly — which is exactly why the readiness-probe smoke passes on this backend.

The probe is also **representative**: it builds the real AppContainer identity/token + kill-on-close Job that the production worker uses, then verifies job membership (`IsProcessInJob`) and the exact requested AppContainer SID (`TokenAppContainerSid` + `EqualSid`). It does not produce false-unavailable on a host that can run the worker.

**Assessment: false positive. No change.**

按本 PR(private-desktop)范围评估。此条落在 stacked 的 #3161 readiness 探针代码上,非本 PR 改动;仍逐条判断:

探针**刻意**以 **System32 作为 cwd** 启动 `cmd.exe /c exit 0`。`builtin-tools.ts` 记录的 `0xC0000142` DLL-init 失败,针对的是"任意 shell workload + 继承的默认拒绝 cwd"。System32 对 `ALL APPLICATION PACKAGES` 可读,故 AppContainer 子进程 DLL init 成功、`/c exit 0` 干净返回——这正是本机 readiness-probe smoke 能过的原因。探针本身有代表性(构造真实 AppContainer identity/token + kill-on-close Job,校验 Job 归属与精确 SID),不会对可运行 worker 的主机误判为不可用。

**结论:误报,不改。**
</details>

packages/runtime/src/sandbox/default-sandbox-manager.ts
```diff
@@ -33,7 +85,7 @@ function builtinWindowsBackend(
   return new WindowsBrokerSandboxBackend({
     clientPath,
     writeManifest: createWindowsBrokerManifestWriter(),
-    isAvailable: () => existsSync(clientPath),
+    isAvailable: createWindowsReadinessProbe(clientPath),
```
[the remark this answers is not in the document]
liugddx: Analyzed against this PR's scope (private-desktop); this concerns the #3161 readiness seam, assessed anyway.

`isBuiltinFilesystemWorkerSandboxAvailable(win32)` is an **existence-only advisory** gate, not the enforcement seam. The enforcing readiness probe is wired into `WindowsBrokerSandboxBackend.isAvailable` (the `spawnSync('--readiness-probe')`), which is the gate `SandboxManager` consults when it actually **selects a backend to launch**. At `execution-composition.ts:307` the `FilesystemWorkerClient` is constructed *with that sandboxManager*, so on a host where the binary exists but the readiness probe fails, composition may advertise a managed worker, but any real launch **fail-closes at the backend** — it never silently downgrades to host execution.

Consolidating the two into one signal would move the enforcing `spawnSync` probe onto the Runtime Host startup hot path, which #3161 deliberately kept lazy/memoized inside the backend. This is a defense-in-depth split (advisory existence vs. enforcing probe), not a bypass.

**Assessment: by design, fail-closed at launch. No change.**

按本 PR(private-desktop)范围评估;此条针对 #3161 readiness 代码,仍逐条判断。

`isBuiltinFilesystemWorkerSandboxAvailable(win32)` 只是"存在性 advisory"门,不是执行门。真正的 enforcing 探针接在 `WindowsBrokerSandboxBackend.isAvailable`(`spawnSync('--readiness-probe')`)上,这才是 `SandboxManager` 实际**选择 backend 启动**时查的门。`execution-composition.ts:307` 处 `FilesystemWorkerClient` 是**带着该 sandboxManager** 构造的;因此二进制存在但探针失败的主机上,composition 可能 advertise managed worker,但任何真实 launch 都会**在 backend 层 fail-closed**,绝不静默降级到宿主执行。

把两者并成一个信号=把 enforcing `spawnSync` 探针搬进 Runtime Host 启动热路径,而 #3161 刻意把它 lazy/memoized 放在 backend 内。这是"存在性 advisory vs. enforcing 探针"的纵深分层,不是绕过。

**结论:按设计,启动即 fail-closed,不改。**
</details>

packages/runtime/src/sandbox/default-sandbox-manager.ts
```diff
@@ -24,6 +25,57 @@ function builtinWindowsClientPath(
...
+function probeWindowsReadiness(clientPath: string): boolean {
+  const cached = windowsReadinessCache.get(clientPath);
+  if (cached !== undefined) return cached;
+  let available = false;
+  if (existsSync(clientPath)) {
+    try {
+      const result = spawnSync(clientPath, ['--readiness-probe'], {
+        timeout: WINDOWS_READINESS_PROBE_TIMEOUT_MS,
+        windowsHide: true,
+        stdio: 'ignore',
```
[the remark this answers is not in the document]
liugddx: Analyzed against this PR's scope (private-desktop); this concerns the #3161 readiness seam, assessed anyway.

The launcher **self-bounds** the throwaway child at `READINESS_PROBE_TIMEOUT_MS = 10s`, strictly **less than** the TS `spawnSync` external timeout of 15s. So under a normal timeout the Rust side reaps first, runs `Drop` (the sole `DeleteAppContainerProfile` call) and closes the Job handle as a kernel backstop — well before the 15s hard kill. A leak requires a hang so severe the launcher's own 10s bound never fires.

For that residual case the profile name already carries a wall-clock-nanos **nonce** (`request_id = readiness-probe.<pid>.<nonce>`), added precisely so a leaked profile can never collide with a later probe (a PID-only name would fail `create (already exists)` on PID reuse). Fully closing the residual window (a panic-safe external cleanup registry) adds persistent state + cleanup-test burden that conflicts with the nonce approach, for negligible benefit.

**Assessment: low-severity reliability edge, largely mitigated (10s < 15s ordering + nonce), deferred by design. No change.**

按本 PR(private-desktop)范围评估;此条针对 #3161 readiness 代码,仍逐条判断。

launcher **自限**抛弃式子进程 `READINESS_PROBE_TIMEOUT_MS = 10s`,严格**小于** TS `spawnSync` 外部 15s。正常超时下 Rust 侧先回收、跑 `Drop`(唯一的 `DeleteAppContainerProfile`)、关 Job 句柄作内核兜底——远早于 15s 硬 kill。只有"连自身 10s 都不触发的灾难性 hang"才会泄漏。

而该残余场景 profile 名已带纳秒 **nonce**(`request_id = readiness-probe.<pid>.<nonce>`),正是为让泄漏 profile 不与后续探针撞名(纯 PID 名会在 PID 复用时 `create (already exists)` 失败)。彻底堵这个残余窗口(panic-safe 外部清理注册表)会引入持久状态 + 清理测试负担,与 nonce 方案冲突,收益极低。

**结论:低危可靠性边角,已大部分缓解(10s<15s 顺序 + nonce),按设计延后,不改。**
</details>

experiments/windows-sandbox/launcher/src/windows_launcher.rs
```diff
@@ -250,6 +458,111 @@ pub(crate) fn appcontainer_profile_name(request_id: &str) -> Vec<u16> {
...
+    // app mask (enough to keep a live handle and later close it) always
+    // succeeds. No `DESKTOP_SWITCHDESKTOP`: the launcher never foregrounds it.
+    let handle = unsafe {
+        CreateDesktopW(
+            name.as_ptr(),
+            null(),
+            null(),
+            0,
+            CONFINED_DESKTOP_APP_MASK,
+            &security,
```
M4n5ter, line 815: <details open>
<summary>English</summary>

**[P1] Set and verify a Low mandatory integrity label on the desktop**

This security descriptor contains only a DACL. The desktop therefore receives the creator's integrity level—normally Medium—while AppContainer processes run at Low IL. Mandatory Integrity Control is evaluated before the DACL, so granting the package SID `DESKTOP_CREATEWINDOW` and `DESKTOP_WRITEOBJECTS` does not by itself let the child write to a Medium-integrity desktop.

The current Windows smoke does not exercise this: `--boundary-probe` only reads the assigned desktop's name, while the readiness child runs with `CREATE_NO_WINDOW`. A production worker or tool that creates a hidden window/menu can therefore fail initialization even though both probes pass.

Please either:

- assign and verify a Low mandatory label on the created desktop, then add an AppContainer probe that successfully creates and destroys a hidden window; or
- explicitly make Win32k/UI unavailable and narrow the rights and compatibility claims accordingly.

Microsoft documents both that AppContainers run at Low IL and that Low-IL processes cannot write to higher-integrity objects even when the DACL grants access: [Launch an AppContainer](https://learn.microsoft.com/en-us/windows/win32/secauthz/implementing-an-appcontainer), [Mandatory Integrity Control](https://learn.microsoft.com/en-us/windows/win32/secauthz/mandatory-integrity-control).

</details>

<details>
<summary>简体中文</summary>

**[P1] 为 private desktop 设置并验证 Low mandatory integrity label**

这里构造的 security descriptor 只有 DACL。因此 desktop 会获得创建者的 integrity level——通常是 Medium——而 AppContainer 进程运行在 Low IL。Mandatory Integrity Control 会先于 DACL 检查，所以即使 DACL 给 package SID 授予了 `DESKTOP_CREATEWINDOW` 和 `DESKTOP_WRITEOBJECTS`，Low-IL child 也不一定能写入 Medium-integrity desktop。

当前 Windows smoke 没有覆盖这一点：`--boundary-probe` 只读取当前 desktop 名称，readiness child 又使用了 `CREATE_NO_WINDOW`。因此，生产 worker 或工具一旦创建 hidden window/menu，仍可能初始化失败，而现有两个 probe 都会通过。

请二选一：

- 给新 desktop 设置并验证 Low mandatory label，并增加一个 AppContainer probe，实际创建和销毁 hidden window；或
- 明确禁用 Win32k/UI，并相应收窄权限与兼容性声明。

微软文档明确说明 AppContainer 运行在 Low IL，且即使 DACL 授权，Low-IL principal 仍不能写入更高 integrity 的对象：[Launch an AppContainer](https://learn.microsoft.com/en-us/windows/win32/secauthz/implementing-an-appcontainer)、[Mandatory Integrity Control](https://learn.microsoft.com/en-us/windows/win32/secauthz/mandatory-integrity-control)。

</details>

M4n5ter (changes requested):
<details open>
<summary>English</summary>

Follow-up review on current head `c999eefd4`.

The previously reported window-station/clipboard blocker remains and is not duplicated here. One additional blocker was found: the new desktop receives a DACL but no explicit Low mandatory integrity label. Since AppContainer processes run at Low IL, the current probes prove only desktop assignment and name visibility—not that the advertised create-window/write rights are usable by production children.

Please add and verify the appropriate integrity label, with a child-side window-creation test, or explicitly narrow the backend to a no-Win32k contract.

The commit also discloses material Claude authorship but contains only `Co-Authored-By`; project policy requires a `Generated-by: <tool>` trailer on every materially AI-authored commit.

All current GitHub checks pass, including 41 Rust tests and the Windows smoke, but those tests do not exercise clipboard access, reopening `Winsta0\Default`, desktop write/create access, or concurrent desktop-heap pressure.

</details>

<details>
<summary>简体中文</summary>

对当前 head `c999eefd4` 的补充审查。

先前指出的 window-station/clipboard blocker 仍然成立，此处不重复评论。另发现一个 blocker：新 desktop 配置了 DACL，却没有显式设置 Low mandatory integrity label。AppContainer 进程运行在 Low IL，因此当前 probe 只证明了 child 被分配到指定名称的 desktop，不能证明生产 child 实际拥有所声明的 create-window/write 权限。

请设置并验证正确的 integrity label，同时增加 child-side window-creation 测试；或者明确把 backend 收窄为禁止 Win32k 的合同。

此外，该提交披露了 Claude 的实质性参与，但只有 `Co-Authored-By`。项目政策要求每个包含实质性 AI 内容的提交带有 `Generated-by: <tool>` trailer。

当前 GitHub checks 全部通过，包括 41 个 Rust 测试和 Windows smoke；但这些测试没有覆盖 clipboard、重新打开 `Winsta0\Default`、desktop write/create 权限以及并发 desktop-heap 压力。

</details>

liugddx:
Thanks both — pushed `0df8547d7`, taking the narrowing path you offered rather than building a full window station in this PR.

**@hqhq1025** — I took option (a): narrow the claim to the desktop-scoped guarantees actually enforced, with clipboard deferred.

- The RFC (§6.3, both languages), the `create_confined_desktop` doc comment, and the PR body now claim only that the worker is **confined off the interactive `Default` desktop** — it cannot enumerate/post messages to the user's interactive windows or install desktop hooks against them.
- Clipboard isolation is explicitly **not** claimed here: the RFC now states the clipboard is window-station-scoped and stays shared, and §6.5 lists "full window-station separation **and clipboard isolation**" as the later gate. The `desktopIsolated` attestation is scoped to exactly what it proves (name is non-empty and not `Default`) — a confinement check, not a clipboard claim.

**@M4n5ter** — the DACL-vs-Low-IL / integrity-label finding: I took your "explicitly narrow the backend to a no-Win32k contract" option.

- The shipped contract is now stated as **confinement only** — the worker performs no Win32k/GUI work. The create-window/write rights in the desktop DACL are granted but **explicitly not relied upon**; §6.5 (both languages) adds a deferred gate "Usable Low-IL desktop rights / no-Win32k contract" noting the desktop carries no explicit Low mandatory-integrity label, so a verified label plus a child-side window-creation test is deferred. So the probes now prove only what the contract claims (desktop assignment + off-`Default`), matching the narrowed guarantee.
- AI attribution: the commit now carries `Generated-by: Claude Fable 5`, and the PR template's **AI use** section is filled in with tool and scope.

Both the readiness-ownership and failure-cleanup blockers you referenced on #3161 are addressed in `4506446f2` there (single availability authority warmed at composition; settlement-path reuse; deterministic self-reconciling cleanup). This PR stays stacked on #3161 — please merge #3161 first.

<details>

感谢两位 —— 已推 `0df8547d7`,采用你们给出的收窄方案,而不是在本 PR 里建完整 window station。

**@hqhq1025** —— 采用 option (a):把声明收窄到实际强制的 desktop 作用域,clipboard 延后。

- RFC(§6.3,双语)、`create_confined_desktop` doc 注释、PR 正文现在只声明 worker 被**挡在交互 `Default` 桌面之外** —— 不能枚举/向用户交互窗口发消息,或对其装 desktop hook。
- 此处明确**不**声明 clipboard 隔离:RFC 现说明 clipboard 是 window-station 作用域、仍为共用,§6.5 把"完整 window-station 分离**与 clipboard 隔离**"列为后续门禁。`desktopIsolated` 断言严格对应它所证明的(名字非空且非 `Default`)—— 是 confinement 检查,不是 clipboard 声明。

**@M4n5ter** —— DACL vs Low-IL / integrity-label 那条:采用你的"明确把 backend 收窄为 no-Win32k 契约"选项。

- 落地契约现在表述为**仅 confinement** —— worker 不做任何 Win32k/GUI 工作。desktop DACL 里的 create-window/write 权限被授予但**明确不依赖**;§6.5(双语)新增延后门禁"可用的 Low-IL 桌面权限 / no-Win32k 契约",说明该桌面没有显式 Low mandatory-integrity label,故经验证的 label 加 child 侧 window-creation 测试暂缓。这样 probe 只证明契约所声明的内容(桌面分配 + 非 `Default`),与收窄后的保证一致。
- AI attribution:该 commit 现带 `Generated-by: Claude Fable 5`,PR 模板 **AI use** 段已按工具与范围填好。

你在 #3161 提到的 readiness-ownership 与 failure-cleanup blocker,已在那边 `4506446f2` 处理(availability 收敛为单一 authority 并在 composition warm;复用 settlement path;确定性自我调和清理)。本 PR 仍 stack 在 #3161 上 —— 请先合 #3161。

</details>

hqhq1025 (changes requested):
Follow-up review on current head `0df8547d7`.

The private-desktop-specific concerns from my previous review are resolved at the contract level:

- the RFC and code comments now explicitly state that the alternate desktop does not isolate the window-station clipboard;
- dedicated window-station/clipboard isolation is correctly deferred;
- Low-IL create-window/write usability is no longer claimed, and the shipped behavior is narrowed to a no-Win32k confinement contract;
- the AI attribution trailers are present.

I would accept this PR's private-desktop delta on that narrowed scope. I cannot approve the stacked revision yet because it includes #3161 head `4506446f2`, whose fixed readiness-profile implementation still has blocking namespace, cross-process ownership, unsettled-cleanup, and permanent-negative-cache issues. See my follow-up review on #3161 for the exact failure paths.

Once #3161 lands with those lifecycle issues resolved and this stack is rebased onto that accepted head, I will re-check the final desktop delta and CI. The current #3174 Windows W0 check is green; some package/E2E checks are still running at review time.

liugddx:
Rebased onto the updated #3161 head (`c8b6527d2`) — the private-desktop delta itself is unchanged; only the readiness-probe base commit moved to pick up the four lifecycle fixes reviewed on #3161. Re-validated the launcher on the rebased head: `cargo fmt --check`, `cargo build --locked`, and `cargo test --locked` (45 passing, including the private-desktop desktop tests). No content change to the desktop scope reviewed here.

已 rebase 到更新后的 #3161 head(`c8b6527d2`):private-desktop 改动本身不变,仅 readiness-probe 基座 commit 前移以纳入 #3161 上评审的四条 lifecycle 修复。在 rebase 后的 head 复验 launcher:cargo fmt/build/test(45 通过,含 private-desktop 桌面测试)。此处评审的 desktop 作用域无内容变化。
</details>

hqhq1025 (changes requested):
Follow-up review on current head `d861e1ee7`.

The earlier clipboard and Low-IL overclaims are fixed, but the private-desktop delta still does not establish confinement from the interactive desktop.

`STARTUPINFO.lpDesktop` selects the child's initial desktop; it does not prevent code inside that child from calling `OpenDesktopW("Default")` and moving a thread with no windows/hooks via `SetThreadDesktop`. This PR creates and protects the private desktop, but it neither structurally denies access to `Winsta0\Default` nor enables a no-Win32k mitigation. The current smoke proves only the initial desktop name, so `desktopIsolated` overstates what was verified.

Please make one of the intended boundaries real: preferably enable and validate the no-Win32k mitigation for the Node worker, or use a token/window-station design that prevents reopening `Default`. Add child-side negative tests for `OpenDesktop(Default)`, `SetThreadDesktop`, window enumeration, and hooks. Until then this is initial desktop placement, not malicious-code desktop confinement.

This stack also includes #3161, which still has the readiness retry and unsettled-profile lifecycle blockers noted in my latest review there. All current CI is green, including the Windows W0 lane, but it does not cover the escape path above. Not ready to merge.

experiments/windows-sandbox/launcher/src/windows_launcher.rs
```diff
@@ -250,6 +601,132 @@ pub(crate) fn appcontainer_profile_name(request_id: &str) -> Vec<u16> {
...
+    let security = SECURITY_ATTRIBUTES {
+        nLength: size_of::<SECURITY_ATTRIBUTES>() as u32,
+        lpSecurityDescriptor: descriptor as *mut c_void,
+        bInheritHandle: 0,
+    };
+    // The launching user holds GA in the DACL above, so requesting the minimal
+    // app mask (enough to keep a live handle and later close it) always
+    // succeeds. No `DESKTOP_SWITCHDESKTOP`: the launcher never foregrounds it.
+    let handle = unsafe {
+        CreateDesktopW(
```
M4n5ter: <details open>

**[P1] Bound the per-launch WinSta0 desktop-heap cost**

Every live sandbox invocation now creates a separate desktop with `CreateDesktopW`, which uses the default interactive-desktop heap. Microsoft documents a 48 MiB system desktop-heap limit and a default 3,072 KiB allocation for each desktop on `WinSta0`. The repository already treats ten concurrent launches as supported, so sustained concurrency can consume roughly 30 MiB before accounting for `Default`, `Winlogon`, `ScreenSaver`, or desktops created by other software. A few more parallel or long-running tool calls can therefore make `CreateDesktopW` fail and reject otherwise valid sandbox operations until earlier commands exit.

The current ten-way smoke does not establish headroom because its children exit immediately; it does not hold all ten desktops live simultaneously.

Please make this resource budget explicit—for example, use [`CreateDesktopExW`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-createdesktopexw) with a measured small heap suitable for the no-GUI workload, reuse a bounded desktop boundary, or enforce a concurrency limit below a measured capacity—and add a Windows test that keeps the supported maximum alive. If the existing confinement blocker is resolved by enforcing Win32k-disable and removing the private desktop, that would eliminate this failure mode as well.

Microsoft documents the limit and default allocation in [`CreateDesktopW`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-createdesktopw).

</details>

<details>

**[P1] 限制每次启动占用的 WinSta0 desktop heap**

现在每个存活的 sandbox invocation 都通过 `CreateDesktopW` 创建一个独立 desktop，因此会使用默认的交互式 desktop heap。微软文档说明系统 desktop heap 上限为 48 MiB，而 `WinSta0` 上每个 desktop 默认分配 3,072 KiB。仓库已经把十路并发启动作为受支持场景；如果这些 child 持续运行，仅它们就会消耗约 30 MiB，还没有计入 `Default`、`Winlogon`、`ScreenSaver` 以及其他软件创建的 desktop。再增加少量并行或长时间运行的 tool call，就可能让 `CreateDesktopW` 失败，使原本合法的 sandbox 操作在已有命令退出前持续被拒绝。

当前十路并发 smoke 的 child 会立即退出，因此没有证明十个 desktop 会同时存活，也没有验证剩余容量。

请显式处理这项资源预算，例如使用 [`CreateDesktopExW`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-createdesktopexw) 为 no-GUI workload 设置经过实测的较小 heap、复用数量有界的 desktop boundary，或把强制并发上限设在实测容量以内；同时增加一个让受支持最大并发数持续存活的 Windows 测试。如果现有 confinement blocker 最终通过强制 Win32k-disable 并删除 private desktop 解决，也会同时消除此故障模式。

相关限制和默认分配见微软的 [`CreateDesktopW` 文档](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-createdesktopw)。

</details>

M4n5ter (dismissed):
<details open>

Request changes on current head `d861e1ee7`.

I found one additional blocker not covered by the existing threads: allocating a default-size `WinSta0` desktop for every live launch puts ordinary sustained parallel workloads close to the system desktop-heap limit.

I am not repeating hqhq1025’s valid current-head finding that `lpDesktop` only selects the initial desktop and that no Win32k/UI restriction prevents malicious code from reopening `Default`.

No other new material correctness or design findings remain after reviewing the private-desktop delta, its composition with stacked #3161, existing discussions, and the simplification surface. The simplest coherent design may be to enforce the stated no-Win32k boundary and remove the per-launch desktop state, subject to Node-worker compatibility evidence.

Verification on `d861e1ee7`:

- All current GitHub checks are successful, including the Windows W0 lane.
- `git diff --check` and `cargo fmt --all -- --check` pass.
- Windows-target production and test code compile with `cargo check`.
- Desktop-heap exhaustion was established from the documented Windows resource contract and reachable sustained-concurrency path; the current CI does not reproduce it because its ten child processes exit quickly.

</details>

<details>

在当前 head `d861e1ee7` 上请求修改。

我发现了一个现有讨论尚未覆盖的 blocker：每个存活的 launch 都在 `WinSta0` 上分配默认大小的 desktop，会让普通的持续并发 workload 很快逼近系统 desktop-heap 上限。

hqhq1025 已经准确指出：`lpDesktop` 只选择初始 desktop，当前也没有 Win32k/UI restriction 阻止恶意代码重新打开 `Default`。这里不重复该 finding。

检查 private-desktop 增量、它与 stacked #3161 的组合、全部现有讨论和简化空间后，没有其他新的实质性 correctness/design finding。一个可能更简单且完整的方向是强制落实文档声明的 no-Win32k boundary，并删除按 launch 创建 desktop 的状态；前提是补齐 Node worker 的兼容性证据。

在 `d861e1ee7` 上的验证：

- 当前所有 GitHub checks 均成功，包括 Windows W0 lane。
- `git diff --check` 与 `cargo fmt --all -- --check` 通过。
- Windows target 的生产代码和测试代码均通过 `cargo check`。
- Desktop-heap 问题基于微软公开的资源契约和可达的持续并发路径确认；当前 CI 的十个 child 很快退出，因此没有复现持有全部 desktop 时的耗尽场景。

</details>

liugddx:
@hqhq1025 Your latest point is correct and I've converged on de-claiming rather than shipping the escape-proof boundary in this PR. New head: `c000c6423` (rebased onto the fixed #3161 head `8b03f9f61`).

**`lpDesktop` is initial placement, not confinement — confirmed.** `STARTUPINFOW.lpDesktop` selects only the child's *initial* desktop; in-process code can still `OpenDesktopW("Default")` + `SetThreadDesktop` to re-attach, because no no-Win32k mitigation, dedicated window station, or token boundary is enforced. The old `desktopIsolated` attestation proved only the initial desktop name, so it overstated. Changes:
- Renamed the attestation `desktopIsolated` → **`desktopPrivatePlacement`** (JSON key, `desktop_is_isolated` → `desktop_is_private_placement`, comments, smoke assertion, unit test), so the name matches what it proves: initial-desktop placement + a DACL-protected private desktop.
- RFC §6.3 and §6.5 (EN+zh) now state plainly this is **placement + DACL protection, not escape-proof confinement**, and explicitly note the `OpenDesktopW("Default")`+`SetThreadDesktop` re-attach path.
- Moved the intended boundaries — no-Win32k process mitigation, dedicated window station (which also isolates the clipboard), token boundary, and the child-side negative tests (`OpenDesktop(Default)`, `SetThreadDesktop`, window enumeration, hooks) — into explicit §6.5 deferred gates.
- Rewrote the commit subject `confine …` → `place … on a private desktop` and its body to match.

Why de-claim rather than enable no-Win32k now: making one boundary real (no-Win32k for the Node worker, or a token/window-station design) plus child-side escape tests is a genuine hardening slice with its own compatibility surface across Node/PowerShell/cmd/Git. Bundling it here would blow this PR's scope; the honest shipped guarantee is placement + DACL, and the worker does no GUI work. I'd rather ship that truthfully now and land the escape-proof boundary as its own gated slice than overclaim.

Since your prior review already accepted the private-desktop delta on the narrowed scope, and the stack is now rebased onto the fixed #3161 head, this should be re-checkable once #3161 lands. Not self-merging.

你的最新意见正确,本 PR 选择**降级声明**而非在此塞入防逃逸边界。新 head:`c000c6423`(已 rebase 到修好的 #3161 head `8b03f9f61`)。

**`lpDesktop` 是初始放置、不是 confinement —— 属实。** `lpDesktop` 只选初始桌面;进程内代码仍可 `OpenDesktopW("Default")` + `SetThreadDesktop` 重新挂回,因为没有 no-Win32k mitigation、独立 window station 或 token 边界。旧的 `desktopIsolated` 只证了初始桌面名,故 overclaim。改动:把断言 `desktopIsolated` → **`desktopPrivatePlacement`**(JSON key、`desktop_is_isolated`→`desktop_is_private_placement`、注释、冒烟断言、单测),让命名与所证一致:初始桌面 placement + DACL 保护的私有桌面;RFC §6.3/§6.5(EN+zh)明确为 **placement + DACL 保护、非防逃逸 confinement**,并写明 `OpenDesktopW("Default")`+`SetThreadDesktop` 重挂路径;把预期边界(no-Win32k mitigation、独立 window station 兼隔离 clipboard、token 边界,及 child 侧 `OpenDesktop(Default)`/`SetThreadDesktop`/窗口枚举/hook 负测)移入 §6.5 明确延后门禁;commit 主题 `confine …` → `place … on a private desktop` 并同步正文。

为何不在此启用 no-Win32k:把某条边界做实(Node worker 的 no-Win32k,或 token/window-station 设计)加 child 侧逃逸负测,是一块有自身跨 Node/PowerShell/cmd/Git 兼容面的硬化切片,塞进本 PR 会撑爆作用域;当前如实交付的保证是 placement + DACL,且 worker 不做任何 GUI 工作。我宁愿现在如实交付它、把防逃逸边界作为独立门禁切片落地,也不 overclaim。

你上一轮已在收窄作用域上接受 private-desktop delta,现栈已 rebase 到修好的 #3161 head,待 #3161 落地即可复检。不自合。
</details>

Astro-Han:
Thanks for the rework — the desktop confinement is genuinely the right mechanism for the gap (AppContainer fences files/network/token but not desktop scope: window-message shatter and desktop hooks stay open on the interactive desktop), and the implementation is lean: ~120 lines of Rust (`create_confined_desktop` + `lpDesktop` on both launch paths + child-side self-attestation) extending the existing `create_appcontainer_child`/`probe_appcontainer_child` seam, no parallel machinery. I verified the details: `CreateDesktopW` with per-object protected DACL (`D:P`) matches the #3182 lesson, handles close via RAII on all paths, both paths build the desktop before `CreateProcessW` and set `lpDesktop` before launch, the probe child also lands on the private desktop with mutex lease covering the whole cycle, failure is fail-closed (no fallback to Default), the private desktop sits on the launcher's winsta with clipboard sharing declared, and the windows-2025 CI lane genuinely runs `cargo test` + smoke scripts (asserting `desktopPrivatePlacement:true`, prefix match, rejecting `"desktop":"Default"`) green on real Windows. Both broker paths route through `launch_appcontainer`, so production coverage is complete.

**Conclusion: PASS — no P0/P1.**

**P2-1 (handle or explicitly defer): the DACL claim vs the child's *effective* permissions don't match.** The `desktop_sddl` grants `GA` to the user SID (`D:P(A;;GA;;;{owner_sid})(A;;GA;;;SY)(A;;0xc7;;;{app_sid})`), and the commit/RFC declare the child "never" gets `DESKTOP_SWITCHDESKTOP`/`HOOKCONTROL`/journal. But an AppContainer token keeps the user SID as an *effective* SID (only Everyone→ etc. become deny-only), so the first ACE gives the child full desktop rights — including SWITCHDESKTOP, HOOKCONTROL, journal recording/replay, WRITE_DAC/WRITE_OWNER. The "minimal non-interactive, never SWITCHDESKTOP/HOOKCONTROL/journal" guarantee only holds at the app-SID ACE level, not for the child's effective access, and the test `desktop_dacl_grants_owner_and_system_full_control` bakes the GA-for-user into intent. Impact is limited (the child having GA on its *own* desktop crosses no desktop boundary — switching requires permission on the *target/input* desktop, whose DACL is system-owned), so this is a claim-vs-effective-access inconsistency rather than an exploitable hole — but future additions of user privileges would silently flow into the child. Fix (either): append a deny ACE for the user SID covering 0x138 (SWITCHDESKTOP|HOOKCONTROL|JOURNALRECORD|JOURNALPLAYBACK) after the grant ACE (launcher doesn't need those; CreateProcessW needs only the 0xc7 subset), or correct the commit/RFC/comment wording and defer explicitly.

P3 (optional): `desktop_is_private_placement` only excludes `Default`/empty — any other desktop name passes the child-side self-attestation; match the `maka-sandbox-desktop.` prefix instead (the smoke script's prefix assertion is the real backstop); the readiness gate proves SID+Job but not desktop *placement* (the probe child is `cmd /c exit 0` — a future refactor dropping `lpDesktop` would stay green on the gate and only the CI smoke would catch it); the legacy restricted-token paths (`--atomic`/`create_child_atomic` via `CreateProcessWithTokenW`/`CreateProcessAsUserW`) still run on the interactive desktop with no flag noting the gap — production broker doesn't use them, but a comment or reusing `create_confined_desktop` there would harden the invitation-to-misuse.

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagent on `opencode-go/deepseek-v4-flash`), which traced the DACL effective-permission semantics, the launch paths, and the CI lane from the PR head. P2-1 is an access-control analysis (AppContainer token retains user SID as effective), not an observed exploit. Please weigh these findings with your own judgment.*

<details>

结论：PASS（无 P0/P1）。桌面隔离正是这个缺口的正确机制（AppContainer 只围文件/网络/token，不围 desktop scope：window-message shatter 与 desktop hooks 在交互桌面仍开放），实现精简：~120 行 Rust（create_confined_desktop + 两条 launch 路径接线 lpDesktop + 子进程侧自证）扩展现有 create_appcontainer_child/probe_appcontainer_child seam，无并行机制。已核实：CreateDesktopW+逐对象受保护 DACL（D:P）符合 #3182 教训；handle 全路径 RAII 关闭；两条路径都在 CreateProcessW 前建桌面并设 lpDesktop；探针 child 同落私有桌面且 mutex lease 覆盖全程；失败 fail-closed 无 fallback；私有桌面在 launcher 的 winsta 上、剪贴板共享已声明；windows-2025 CI lane 真跑 cargo test + smoke（断言 desktopPrivatePlacement:true、前缀匹配、拒绝 "desktop":"Default"）在真实 Windows 全绿；broker 两条生产路径都经 launch_appcontainer 全覆盖。P2-1：DACL 声明与子进程有效权限不符——desktop_sddl 给 user SID 授 GA（D:P(A;;GA;;;owner)(A;;GA;;;SY)(A;;0xc7;;;app)），commit/RFC 声明"绝不给 SWITCHDESKTOP/HOOKCONTROL/journal"；但 AppContainer token 保留 user SID 为有效 SID（仅 Everyone 等为 deny-only），首条 ACE 即给 child 全部桌面权限（含 SWITCHDESKTOP/HOOKCONTROL/journal 录制回放/WRITE_DAC/WRITE_OWNER）；"最小权限绝不给"只在 app SID ACE 层面成立，测试还把 GA-for-user 固化为意图。影响有限（child 对自身桌面 GA 不跨任何桌面边界——切换需目标/输入桌面的权限，其 DACL 系统所有），是声明-有效权限不一致而非可利用漏洞，但未来给 user 增权会静默流入 child。修复二选一：grant ACE 后追加 user SID deny ACE 覆盖 0x138（SWITCHDESKTOP|HOOKCONTROL|JOURNALRECORD|JOURNALPLAYBACK，launcher 不需这些权限、CreateProcessW 只需 0xc7 子集），或修正 commit/RFC/注释措辞并显式延后。P3（可选）：desktop_is_private_placement 只排除 Default/空名，任何其他桌面名都通过子进程自证——建议直接匹配 maka-sandbox-desktop. 前缀（冒烟脚本前缀断言才是真兜底）；readiness gate 只证 SID+Job 不证桌面放置（探针 child 是 cmd /c exit 0，未来重构丢 lpDesktop 门禁仍绿仅 CI 冒烟能抓）；遗留 restricted-token 路径（--atomic/CreateProcessWithTokenW/CreateProcessAsUserW 无 lpDesktop）仍在交互桌面裸奔——生产 broker 不用它们，但建议加注释或复用 create_confined_desktop。
</details>

liugddx:
Thanks @Astro-Han — took the deny-ACE option and swept the whole PR; head is now `3ec518309` (rebased onto #3161's `/d` fix), point by point:

**P2-1 (owner `GA` is effective for the AppContainer child) — fixed with the deny ACE.** You're right that the launching-user SID stays *effective* in the AppContainer token, so the owner full-control allow named the child as a grantee of the interactive-control rights the commit/RFC said it would "never" hold — a claim-vs-DACL mismatch even where the lowbox dual access check (user grant ∩ package grant = `0xc7`) would have intersected them away in practice. Rather than de-claim, I made the DACL say what the docs say: `desktop_sddl` now emits `D:P(D;;0x138;;;owner)(A;;GA;;;owner)(A;;GA;;;SY)(A;;0xc7;;;app)` — a leading deny ACE (canonical order) stripping `SWITCHDESKTOP | HOOKCONTROL | JOURNALRECORD | JOURNALPLAYBACK` from the user SID, with a named `DENIED_DESKTOP_INTERACTIVE_MASK` const documenting why. The launcher is unaffected (it only ever requests the `0xc7` app mask, disjoint from the deny), and future user-SID privilege additions can no longer silently flow into the child. New test `desktop_dacl_denies_interactive_control_to_the_effective_user_sid` asserts the deny ACE exists **and precedes every allow ACE**; commit message and RFC §6.3/§6.5 (EN + zh) now describe the deny. Real-machine `appcontainer-smoke.ps1` green under the new DACL (child still starts, attests the private desktop).

**P3 attestation prefix — fixed.** `desktop_is_private_placement` now requires the launcher-owned `maka-sandbox-desktop.` prefix (case-insensitive, non-empty suffix) instead of merely "not Default/empty" — `Winlogon`, screensaver desktops, or another product's alternate desktop now fail the child-side check too; the test covers those cases and the smoke prefix assertion remains the CI backstop.

**P3 legacy restricted-token paths — documented.** `create_child` / `create_child_atomic` now carry explicit doc comments: W0 diagnostic paths, NOT production (the packaged broker routes only through `launch_appcontainer`), children run on the interactive desktop with no placement, and wiring them through `create_confined_desktop` is deliberately out of scope while they remain negative-evidence prototypes.

**P3 readiness gate doesn't prove placement — acknowledged, deferred.** True: the probe child (`cmd /d /c exit 0`) reports nothing, so a refactor dropping `lpDesktop` would keep the gate green and only the CI smoke would catch it. Teaching the probe child to self-attest its desktop belongs with the §6.5 "full-policy readiness coverage" gate; not expanding the probe protocol in this PR.

**Self-sweep result (beyond the quoted lines):** your P2-1 plus the pattern across previous rounds — claims exceeding enforcement — prompted a full sweep of this PR instead of another spot-fix. The sweep caught one more instance we had missed: the `create_confined_desktop` doc comment still called this "a confinement boundary under a no-Win32k contract … the shipped guarantee is confinement." That comment (and two "confine the child" call-site comments) are now placement-scoped like the RFC and commit message.

<details>
<summary>中文</summary>

**P2-1（owner `GA` 对 AppContainer 子进程有效）— 以 deny ACE 修复。** 你说得对：发起用户 SID 在 AppContainer token 中仍*有效*，owner 全控 allow 因而把子进程列为了 commit/RFC 声称"绝不给"的交互控制权限的被授予者——即使 lowbox 双重访问检查（user 授权 ∩ package 授权 = `0xc7`）实际会交集掉它们，声明与 DACL 本身仍不一致。这次不降级措辞，而是让 DACL 与文档一致：`desktop_sddl` 现在输出 `D:P(D;;0x138;;;owner)(A;;GA;;;owner)(A;;GA;;;SY)(A;;0xc7;;;app)`——前置 deny ACE（规范顺序）从用户 SID 剥离 `SWITCHDESKTOP|HOOKCONTROL|JOURNALRECORD|JOURNALPLAYBACK`，并有命名常量 `DENIED_DESKTOP_INTERACTIVE_MASK` 说明缘由。launcher 不受影响（只申请与 deny 不相交的 `0xc7`），未来给用户 SID 增权也不会再静默流入子进程。新测试断言 deny ACE 存在**且在所有 allow 之前**；commit/RFC §6.3/§6.5（中英）同步描述 deny。真机 `appcontainer-smoke.ps1` 在新 DACL 下全绿。

**P3 自证前缀 — 已修。** `desktop_is_private_placement` 改为要求 launcher 自有的 `maka-sandbox-desktop.` 前缀（大小写不敏感、后缀非空），不再是"非 Default/非空"——`Winlogon`、屏保桌面或第三方 alternate desktop 现在同样通不过子进程侧自证；测试覆盖这些用例，冒烟前缀断言仍是 CI 兜底。

**P3 遗留 restricted-token 路径 — 已注释。** `create_child`/`create_child_atomic` 现有显式文档注释：W0 诊断路径、非生产（打包 broker 只走 `launch_appcontainer`）、子进程在交互桌面上、无 placement；接入 `create_confined_desktop` 在其仍为负面证据原型期间刻意不做。

**P3 readiness gate 不证 placement — 认同并延后。** 探针子进程（`cmd /d /c exit 0`）不汇报任何东西，丢掉 `lpDesktop` 的重构确实只有 CI 冒烟能抓。让探针子进程自证桌面归入 §6.5 "full-policy readiness coverage" 门禁，本 PR 不扩探针协议。

**自查结果（超出被引用行）：** 你的 P2-1 加上前几轮反复出现的同一模式——声明超出强制——促使这次对整个 PR 做全量清扫而非再打一个点补丁。清扫抓到一处我们此前漏掉的实例：`create_confined_desktop` 的文档注释仍写着 "confinement boundary under a no-Win32k contract … the shipped guarantee is confinement"。该注释（及两处 "confine the child" 调用点注释）现已与 RFC/commit 一致收窄为 placement。
</details>

Astro-Han (commented):
The private-desktop direction strengthens the production AppContainer boundary, and the protected DACL, exact SID/Job checks, separate readiness namespace, and fail-closed Runtime cache are well aligned. Existing public reviews already identify the unaligned token-information UB, WinSta0 desktop-heap exhaustion, and possible existing-desktop reuse; I am not duplicating those inline threads, but they remain blockers.

From first principles, release evidence must attest the boundary properties, not only the launcher's success bit. The additional inline finding asks the independent readiness path to emit and verify its exact desktop/SID/Job attestation, matching what production confinement claims.

Review performed with three Codex reviewer agents and DeepSeek V4 Flash as advisory tools; I verified the novel finding against the latest head and live green CI.

private desktop 方向强化了生产 AppContainer 边界；protected DACL、exact SID/Job 校验、独立 readiness namespace 和 fail-closed Runtime cache 均符合职责边界。现有公开 review 已指出 token information 未对齐读取 UB、WinSta0 desktop heap 耗尽和旧 desktop 复用风险，我不重复这些行内线程，但它们仍是 blocker。

从第一性原理看，release evidence 必须证明边界属性，而不只是 launcher 返回成功。以下新增行内问题建议让独立 readiness 路径输出并验证其 exact desktop/SID/Job attestation，与生产 confinement 的声明一致。

本次审查使用了三位 Codex reviewer agents 与 DeepSeek V4 Flash 作为辅助工具；我已依据最新 head 和实时绿色 CI 复核新增问题。

</details>

experiments/windows-sandbox/readiness-probe-smoke.ps1
```diff
@@ -0,0 +1,43 @@
...
+if (-not (Test-Path -LiteralPath $launcher)) {
+  throw "Missing launcher binary: $launcher"
+}
+
+# The production-identity readiness probe (RFC §6.4) stands up the real
+# AppContainer identity, token, and kill-on-close Job and launches a throwaway
+# confined child. A clean exit 0 proves the host can actually create and enforce
+# the boundary — this is what the runtime's `isAvailable` fails closed on, so it
+# must never regress into file-presence-only availability.
+$firstOutput = & $launcher --readiness-probe 2>&1
```
Astro-Han, line 36: **P2 — This smoke test does not attest the boundary it claims to prove.** It accepts any `--readiness-probe` implementation that exits 0, so removing/bypassing the throwaway child's private-desktop, exact AppContainer SID, or specific Job verification would leave release evidence green. The production `appcontainer-smoke.ps1` exercises a separate path. Emit a bounded machine-readable attestation from this readiness child (desktop isolation + SID + Job) and assert those fields here.

experiments/windows-sandbox/launcher/src/windows_launcher.rs
liugddx: Fixed in `70eea182a` (first option, minus the in-child window probe): `desktop_sddl` now appends `S:(ML;;NW;;;LW)`, pinning the desktop at Low integrity with No-Write-Up so the DACL's create-window/write grants pass MIC for the Low-IL AppContainer child — labeling at Low needs no privilege since it is below the creator's own level, and higher-IL access is unaffected. `desktop_sddl_pins_a_low_integrity_no_write_up_label` asserts the label. The in-child hidden-window create/destroy probe stays explicitly deferred (RFC §6.5): the worker performs no GUI work and the probe child is `cmd /d /c exit 0`, so the shipped claim remains placement with labeled-usable rights, not proven GUI capability — the RFC (EN+zh) says exactly that now.

中文:已修——SDDL 追加 `S:(ML;;NW;;;LW)` 把桌面钉在 Low integrity(No-Write-Up),使 DACL 授予的 create-window/write 对 Low-IL 子进程通过 MIC;单测断言标签存在。child 内建窗探针按 §6.5 显式延后:worker 不做 GUI,声明保持"placement + 标签可用",不宣称端到端 GUI 能力。

experiments/windows-sandbox/launcher/src/windows_launcher.rs
liugddx: Fixed in `70eea182a` with the `CreateDesktopExW` option: each private desktop is now created with an explicit 512 KiB heap (`CONFINED_DESKTOP_HEAP_KB`), a sixth of the default 3,072 KiB, so the supported ten-way concurrency costs ~5 MiB of the documented 48 MiB system budget instead of ~30 MiB. Your "hold the maximum live" point is covered by a real-OS test, `ten_confined_desktops_can_be_held_live`, which creates and holds ten confined desktops simultaneously (desktops own the heap allocation; children need not run) and runs in the windows-2025 CI lane. The budget and rationale are documented on the constant and in RFC §6.3/§6.5 (EN+zh).

中文:已修——改用 `CreateDesktopExW` 显式 512 KiB heap(默认 3,072 KiB 的六分之一),十路并发约 5 MiB;新增真机测试 `ten_confined_desktops_can_be_held_live` 同时持有十个私有桌面证明预算,跑在 windows-2025 CI lane。

experiments/windows-sandbox/launcher/src/windows_launcher.rs
```diff
@@ -891,6 +1291,68 @@ unsafe fn child_process_is_in_job(process: HANDLE) -> Result<bool, String> {
...
+            buffer.as_mut_ptr() as *mut c_void,
+            needed,
+            &mut needed,
+        )
+    } == 0
+    {
+        return Err(last_error("GetTokenInformation(TokenAppContainerSid)"));
+    }
+    let info = buffer.as_ptr() as *const TOKEN_APPCONTAINER_INFORMATION;
+    let sid = unsafe { (*info).TokenAppContainer };
```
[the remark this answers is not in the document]
liugddx: Fixed in `70eea182a`: the buffer is now sized in `usize` words (`vec![0usize; needed / size_of::<usize>() + 1]`), matching the `current_user_sid_string` pattern this file already uses, so reading `TOKEN_APPCONTAINER_INFORMATION` (leading pointer field) through it is properly aligned instead of UB-by-luck.

experiments/windows-sandbox/readiness-probe-smoke.ps1
liugddx: Fixed in `70eea182a`: `--readiness-probe` now emits a bounded machine-readable attestation of the facts it verified parent-side — `{"appContainerSidVerified":true,"jobVerified":true,"settled":true,"appContainerSid":"S-1-15-2-…","desktop":"maka-sandbox-desktop.…","desktopPrivatePlacement":…}` — and this smoke asserts every field plus the desktop prefix and rejects `Default`, so dropping the exact-SID match, the specific-Job membership check, the settlement drain, or the `lpDesktop` placement turns release evidence red instead of leaving a hollow exit-0 gate green. One scoping note kept honest in the output comment: `desktop` is the parent-side `lpDesktop` placement fact (the throwaway cmd child cannot self-report), consistent with the placement — not escape-proof — claim in RFC §6.3.

中文:已修——`--readiness-probe` 输出父侧已验证事实的机器可读 attestation,冒烟逐字段断言(SID 前缀、Job、settlement、桌面前缀、拒绝 Default);删除任一验证都会让发布证据变红。`desktop` 字段为父侧 lpDesktop placement 事实(cmd 子进程无法自证),与 §6.3 placement 声明一致。

liugddx:
Head is now `70eea182a` (rebased onto #3161's `a59c2554e`, plus one commit), closing all four open inline threads — details and per-thread replies inline: Low no-write-up mandatory label pinned on the desktop (M4n5ter P1; in-child window probe stays deferred per RFC §6.5), desktop heap bounded via `CreateDesktopExW` at 512 KiB with a real-OS test holding ten confined desktops live simultaneously (M4n5ter P1), the `TOKEN_APPCONTAINER_INFORMATION` unaligned-read UB fixed with a word-aligned buffer (CodeRabbit), and `--readiness-probe` now emits a machine-readable attestation (exact SID, specific Job, settlement, desktop placement) that `readiness-probe-smoke.ps1` asserts field by field (Astro-Han P2). RFC EN+zh updated to record label + heap bound as enforced and to narrow the deferred gate to no-Win32k + in-child window-creation coverage. Local: 49 cargo tests + both real-machine smokes green.

head 更新为 `70eea182a`(rebase 到 #3161 `a59c2554e` 并加一个 commit),关闭全部四条行内线程:桌面钉 Low no-write-up 标签(child 内建窗探针按 §6.5 延后);`CreateDesktopExW` 限 512 KiB heap + 真机十桌面并存测试;`TOKEN_APPCONTAINER_INFORMATION` 未对齐读取 UB 改字对齐缓冲;`--readiness-probe` 输出机器可读 attestation(精确 SID/特定 Job/settlement/桌面放置)且冒烟逐字段断言。RFC 中英同步。本地 49 cargo 测试 + 两冒烟全绿。
</details>

experiments/windows-sandbox/launcher/src/acl_ledger.rs
```diff
@@ -234,6 +267,12 @@ impl LedgerLock {
         if let Some(error) = create_error {
             return Err(error);
         }
+        if already_exists {
+            if let Err(error) = validate_existing_lock_owner(handle, user_sid) {
```
Astro-Han, line 270: **P1 — The owner check rejects legitimate elevated mutexes.** The required Windows run fails because the existing object is owned by `S-1-5-32-544` (Administrators), while validation accepts only `TokenUser` or SYSTEM. That makes same-user/elevated follow-up acquisition fail as squatting before it can wait on the lock. Create the mutex with an explicit expected owner and validate that identity (or use the token's actual owner consistently); do not simply trust any Administrators owner. Add repeated elevated acquisition coverage and rerun W0.

experiments/windows-sandbox/launcher/src/windows_launcher.rs
```diff
@@ -250,6 +618,178 @@ pub(crate) fn appcontainer_profile_name(request_id: &str) -> Vec<u16> {
...
+/// back to the interactive desktop.
+pub(crate) unsafe fn create_confined_desktop(
+    app_container_sid: *mut c_void,
+) -> Result<ConfinedDesktop, String> {
+    let owner_sid = current_user_sid_string()?;
+    let package_sid = unsafe { sid_string(app_container_sid) }?;
+    // A per-launch name so concurrent sandboxes never share a desktop, and a
+    // name leaked by a killed launch never collides with a live one. Mirrors
+    // the readiness-probe profile nonce (§6.4).
+    let nonce = SystemTime::now()
```
Astro-Han: **P2 — Fail closed on desktop-name collisions.** PID plus wall-clock nanoseconds is not a guaranteed unique identity, and clock failure collapses the nonce to `0`. If `CreateDesktopExW` opens an existing object, the supplied security descriptor is ignored, defeating the per-launch DACL boundary. Use a GUID/cryptographic nonce and explicitly reject or retry `ERROR_ALREADY_EXISTS`; never fall back to a shared deterministic name.

Astro-Han (commented):
The latest readiness attestation, Low-IL label, aligned token storage, and bounded desktop heap address the earlier findings. The new mutex-owner hardening currently rejects a legitimate elevated owner in required CI, and the desktop name still lacks a fail-closed collision contract.

The simplest authority is explicit and consistent identity: create the mutex with the exact owner SID the validator accepts, and use a GUID-quality desktop nonce while rejecting/retrying `ERROR_ALREADY_EXISTS`. Do not broaden trust to every Administrators-group owner merely to make CI pass.

Reviewed with Codex using two independent review passes and DeepSeek V4 Flash as an external adversarial pass; I verified the Windows object-identity paths and live required-check failure against this exact head.

最新 readiness attestation、Low-IL label、token 对齐存储和 desktop heap 上限已修复旧问题。新的 mutex owner 加固在 required CI 中拒绝了合法 elevated owner；desktop 名称也仍缺少 fail-closed 的冲突契约。

最小权威方案是显式且一致的 identity：创建 mutex 时设置 validator 接受的精确 owner SID；desktop 使用 GUID 级 nonce，并在 `ERROR_ALREADY_EXISTS` 时拒绝或重试。不要为了 CI 直接信任所有 Administrators group owner。

本次由 Codex 两轮独立审查，并使用 DeepSeek V4 Flash 做外部对抗审查；我核对了当前 head 和 required check。

</details>

liugddx:
Rebased onto #3161's `0df509bc3` (elevated-host fix for the lock-owner check that turned `windows_sandbox_w0_protocol` red); head is now `2797e0eb4`, desktop delta unchanged. 中文:已 rebase 到 #3161 的提权宿主修复上,head `2797e0eb4`,桌面增量不变。

hqhq1025 (changes requested):
Follow-up review on current head `2797e0eb4`.

Findings

1. P2 — The RFC still states an escape-proof result before immediately disclaiming it.

`docs/architecture/windows-sandbox-rfc-v1.md:148-170` normatively says workers “cannot enumerate or post window messages to the user's interactive windows, or install desktop hooks against them,” but the same paragraph correctly explains that `lpDesktop` is only initial placement and in-process code can reopen `Winsta0\Default` with `OpenDesktopW` + `SetThreadDesktop`. The PR description also still calls this “desktop-scoped confinement” and says the worker is kept off `Default`, despite the latest commit deliberately narrowing the implementation to placement. This PR exists partly to align the RFC with shipped behavior, so the remaining top-level guarantee must be narrowed to “starts on a launcher-created alternate desktop” and must not use an unconditional “cannot” claim. Apply the same correction to the Chinese RFC and PR description/title.

2. P2 — Desktop creation has no fail-closed uniqueness contract.

`experiments/windows-sandbox/launcher/src/windows_launcher.rs:779-826` derives the object name from PID plus wall-clock nanoseconds and collapses clock failure to `0`. Microsoft documents that `CreateDesktopExW` succeeds by returning the existing desktop when the specified name already exists; the supplied security descriptor and heap parameters therefore do not establish a new per-launch object in that case. PID reuse, clock rollback/failure, or same-process calls at the same clock tick can reopen an older desktop while the code and attestation accept the launcher prefix as proof of this launch's object. Use a GUID/cryptographic nonce and avoid the `unwrap_or(0)` fallback. The Windows test should retain ten handles and assert ten distinct names/objects, not only that ten calls returned handles.

Problem and architecture assessment

The underlying problem is correctly narrowed now: AppContainer alone does not choose a non-interactive initial desktop, and this PR adds initial placement plus a protected per-launch DACL. That is a useful defense-in-depth layer, but it is not confinement from `Default`; no-Win32k or a dedicated window-station/token boundary remains the correct deeper architecture for that stronger requirement. The current implementation is otherwise direct and uses the existing launch paths rather than introducing a parallel broker.

Deletion opportunities

The hand-written `json_string` helper and its test can be replaced with the existing `serde_json::to_string`. The explicit `"desktop":"Default"` smoke rejection is redundant after the stronger launcher-prefix assertion. These are non-blocking cleanup items.

Verification

All live GitHub checks are green. On this exact head I ran `cargo fmt --check`, Windows MSVC cross-target `cargo check --locked`, the full repository build, focused runtime readiness/Windows sandbox tests (22 passed), Biome, diff checks, and merge-conflict checks. Real Windows CI exercises creation, the Low label, the 512 KiB heap, and the smoke paths, but it does not exercise name collision or escape back to `Default`.

Merge verdict: not ready until the two claim/object-identity issues above are fixed. After that, the remaining no-Win32k/window-station work can stay explicitly deferred.

liugddx:
Thanks @hqhq1025 — both points fixed at head `cecc80542`, and both are accepted as valid, not argued down:

**1. Normative claim vs disclaimed reality — fixed at the sentence that mattered.** You're right that the earlier narrowing only fixed the *annotation* layer while the top-level §6.3 guarantee still said workers "cannot enumerate or post window messages … or install desktop hooks." The normative sentence now reads: workers **start on a launcher-created private (alternate) desktop, never the interactive `Default`**, with enumerate/message/hook resistance explicitly tied to the deferred no-Win32k/window-station gates — no unconditional "cannot" remains (EN + zh). The PR title is now "place sandboxed children on a private desktop" (was "confine…"), and the description's "Scope of the guarantee" section was rewritten the same way, including an explicit **Not claimed** bullet for the `OpenDesktopW("Default")` + `SetThreadDesktop` escape.

**2. Desktop-name uniqueness — now fail-closed.** Accepted in full: `CreateDesktopExW` opens an existing desktop on name collision (silently ignoring the supplied DACL and heap size), and PID + wall-clock nanoseconds with `unwrap_or(0)` was both predictable and clock-failure-degenerate. The nonce is now **128 bits from the OS CSPRNG** (`BCryptGenRandom`, `BCRYPT_USE_SYSTEM_PREFERRED_RNG`); RNG failure fails the launch closed instead of collapsing to a constant name. `ten_confined_desktops_can_be_held_live` now retains all ten handles **and asserts ten distinct names that each pass the placement predicate**, not merely ten returned handles.

**Deletion opportunities — both taken.** `json_string` now delegates to `serde_json::to_string` (hand-rolled escaping and its test deleted), and the redundant `"desktop":"Default"` rejections were removed from both smokes — the launcher-prefix assertions already exclude `Default`.

Local verification on `cecc80542`: `cargo fmt --check`, `cargo build --locked`, 49 tests pass (including the ten-distinct-desktops live test), and both real-machine smokes green — the attestation output now shows the 128-bit hex nonce (`maka-sandbox-desktop.<pid>.<32 hex>`).

<details>
<summary>中文</summary>

**1. 规范句与免责说明矛盾 — 已在真正的规范句上修复。** 此前的收窄只改了注解层,§6.3 顶层保证仍写"cannot enumerate/post/hook"。现在规范句为:worker **以 launcher 创建的 private (alternate) desktop 作为初始桌面,绝不落在交互 `Default` 上**;枚举/消息/钩子抗性显式挂到暂缓的 no-Win32k/window-station 门禁,全文无无条件 "cannot"(中英同步)。PR 标题改为 "place…"(原 "confine…"),描述的保证范围段同步重写并加入显式 **Not claimed** 条目(`OpenDesktopW("Default")` + `SetThreadDesktop` 逃逸)。

**2. 桌面名唯一性 — 已 fail-closed。** 完全接受:`CreateDesktopExW` 同名时打开旧桌面并静默忽略传入 DACL/heap,而 PID+时钟纳秒+`unwrap_or(0)` 既可预测又在时钟失败时退化。nonce 改为 **OS CSPRNG 128 位**(`BCryptGenRandom`),RNG 失败即启动 fail closed。十桌面测试持有全部句柄并**断言十个互异且通过 placement 谓词的名字**。

**删除机会 — 两条都采纳。** `json_string` 委托 `serde_json::to_string`(手写转义及其测试已删);两个冒烟中冗余的 `"desktop":"Default"` 拒绝已删(前缀断言已覆盖)。

本机 `cecc80542`:fmt/build 干净,49 测试过(含十互异桌面真机测试),两冒烟绿,attestation 已显示 128 位十六进制 nonce。
</details>

hqhq1025 (approved):
Follow-up review on current head `cecc80542`.

No actionable finding remains in the latest delta. The two blockers from my previous review are resolved.

Problem, root cause, and mechanism

The PR now defines the problem correctly: AppContainer does not choose a non-interactive initial desktop, so production children previously started on `Winsta0\Default`. The patch adds initial placement on a per-launch alternate desktop with a protected DACL, Low mandatory label, bounded heap, fail-closed creation, and placement attestation. It explicitly does not claim escape-proof confinement; no-Win32k and dedicated window-station/token boundaries remain deferred.

The latest commit fixes the object-identity gap directly. Desktop names now use a 128-bit nonce from `BCryptGenRandom`; RNG failure aborts the launch instead of falling back to a predictable value, and the real-Windows test holds ten desktops live while asserting ten distinct launcher-owned names. The RFC, Chinese RFC, PR title, and description now consistently describe initial placement rather than an unconditional inability to reopen `Default`.

First-principles and optimality assessment

For the stated placement guarantee, this is the simplest coherent architecture: reuse the existing production/readiness launch paths, create one scoped desktop object per launch, keep its handle alive through settlement, and fail closed on every setup error. Adding a no-Win32k state machine or dedicated window station here would solve a stronger problem than this PR claims and should remain a separate compatibility-tested hardening slice.

Deletion and test quality

The hand-written JSON escaper and duplicate `Default` smoke assertions were correctly deleted. The remaining SDDL, placement, heap, distinct-name, production smoke, and readiness-attestation tests protect observable boundary properties rather than mirroring implementation details. No further production code or low-quality test needs deletion before merge.

Verification

On this exact head, `git diff --check`, merge-conflict inspection, `cargo fmt --check`, and Windows MSVC cross-target `cargo check --locked` pass. Live Windows W0, Windows recovery, installed Windows CLI, package, typecheck, and CodeRabbit checks are green. At review time, `test_workspaces`, `test_runtime_host`, `storybook`, and `e2e` are still queued or running; they must finish successfully before merge.

Residual risks

The delivered boundary is initial placement, not prevention of `OpenDesktopW("Default")`/`SetThreadDesktop`; clipboard remains window-station scoped; Low-IL GUI rights are labeled but not exercised by an in-child window test; and the readiness lifecycle retains the documented #3161 deferred gaps. These are explicitly documented and do not contradict the current guarantee.

Merge verdict: the reviewed revision is ready once the remaining required checks complete successfully.

Astro-Han (approved):
The current head resolves the earlier private-desktop blockers while keeping the security claim narrow and accurate. It proves initial placement on a launcher-created desktop, uses unpredictable names, protects the DACL and Low-IL label, verifies the exact AppContainer SID/Job identity, and keeps the desktop handle owned through child settlement. The RFC correctly leaves window-station clipboard isolation and stronger escape resistance out of this W0 claim.

I found no reproducible P0–P3 issue in the code on this head. The implementation extends the existing Windows launcher/readiness path and does not create a second sandbox authority.

CI note: the required run is not currently green. The E2E failure is the unrelated `slash-command-menu.spec.ts` compact-session assertion (`Fake backend received: after compact` was not observed), and Storybook remains stuck in progress. This approval is for the reviewed code; the PR is not merge-ready until required CI is resolved.

AI-assisted review disclosure: Codex performed the final review using two independent reviewer passes and OpenCode Go DeepSeek V4 Flash (high effort) as an adversarial advisory pass. I verified the desktop security/lifecycle paths against exact head `cecc80542` and inspected the live E2E log. No local tests were run.

当前 head 已解决此前 private desktop 的阻塞项，同时把安全声明保持在准确、有限的范围：它证明 child 初始位于 launcher 创建的 desktop，使用不可预测名称，保护 DACL 与 Low-IL label，校验精确 AppContainer SID/Job 身份，并由 launcher 持有 desktop handle 直到 child settle。RFC 也正确地没有把 window-station clipboard isolation 和更强 escape resistance 纳入本 W0 声明。

当前 head 的代码未发现可复现 P0–P3。实现扩展既有 Windows launcher/readiness 路径，没有新增第二个 sandbox authority。

CI 说明：必需 run 目前不是绿色。E2E 失败来自无关的 `slash-command-menu.spec.ts` compact-session 断言（未观察到 `Fake backend received: after compact`），Storybook 仍卡在 in progress。本次 approve 针对已审代码；必需 CI 解决前不能合并。

AI 辅助审查说明：Codex 使用两轮独立 reviewer 审查，并以 OpenCode Go DeepSeek V4 Flash（high effort）进行对抗性辅助审查；我已针对精确 head `cecc80542` 核对 desktop security/lifecycle 路径并检查实时 E2E 日志。本轮未运行本地测试。

</details>

PR #3175, opened 2026-08-17
fix(desktop): indent project session rows

ARE404:
## Summary

Restore the standard `spacing-6` (24px) SideNav nesting step for sessions rendered under a project. The CSS contract test pins the hierarchy, and the Electron test verifies the visible indentation after waiting for both measured controls to become visible.

## Verification

- `npx biome check apps/desktop/src/renderer/styles/sidebar.css packages/ui/src/session-history-list.tsx apps/desktop/src/main/__tests__/session-project-hierarchy-contract.test.ts apps/desktop/e2e/sidebar-project-row.spec.ts` — passed
- `node --test apps/desktop/src/main/__tests__/session-project-hierarchy-contract.test.ts` — 1 passed
- `npx playwright test --config e2e/playwright.config.ts e2e/sidebar-project-row.spec.ts` — 3 passed
- Desktop build and focused compiled contract tests passed before review

## Review focus

The implementation restores the existing design-system spacing token rather than introducing a new value. The geometry assertion waits for the project and first-session controls to be visible before reading `boundingBox()`, addressing the review reliability concern.

## AI use

Select exactly one:

Tool(s) and scope: OpenAI Codex audited the sidebar hierarchy, implemented and tested the CSS/E2E fix, incorporated review feedback, and drafted this description. The human contributor reviewed the work, chose to submit it, and remains responsible for its accuracy, provenance, and licensing.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/e2e/sidebar-project-row.spec.ts
+++ b/apps/desktop/e2e/sidebar-project-row.spec.ts
@@ -34,6 +34,8 @@ test('project navigation and actions remain adjacent keyboard controls', async (
   const controlledGroup = page.locator(`[id="${controlledGroupId}"]`);
   const firstSessionControl = controlledGroup.locator('[data-session-id] button').first();
 
+  await expect(navigation).toBeVisible();
+  await expect(firstSessionControl).toBeVisible();
   const [projectNavigationBox, firstSessionBox] = await Promise.all([
     navigation.boundingBox(),
     firstSessionControl.boundingBox(),
--- a/apps/desktop/e2e/sidebar-project-row.spec.ts
+++ b/apps/desktop/e2e/sidebar-project-row.spec.ts
@@ -34,6 +34,14 @@ test('project navigation and actions remain adjacent keyboard controls', async (
   const controlledGroup = page.locator(`[id="${controlledGroupId}"]`);
   const firstSessionControl = controlledGroup.locator('[data-session-id] button').first();
 
+  const [projectNavigationBox, firstSessionBox] = await Promise.all([
+    navigation.boundingBox(),
+    firstSessionControl.boundingBox(),
+  ]);
+  expect(projectNavigationBox).not.toBeNull();
+  expect(firstSessionBox).not.toBeNull();
+  expect(firstSessionBox!.x - projectNavigationBox!.x).toBeGreaterThanOrEqual(20);
+
   await expect(projectRow.locator('button button')).toHaveCount(0);
   await expect(navigation).toHaveAttribute('aria-expanded', 'true');
 
--- a/apps/desktop/src/main/__tests__/session-project-hierarchy-contract.test.ts
+++ b/apps/desktop/src/main/__tests__/session-project-hierarchy-contract.test.ts
@@ -1,0 +1,27 @@
+import assert from 'node:assert/strict';
+import { existsSync, readFileSync } from 'node:fs';
+import { describe, it } from 'node:test';
+
+const sidebarCssUrl = [
+  new URL('../../renderer/styles/sidebar.css', import.meta.url),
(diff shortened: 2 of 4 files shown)


Astro-Han:
Thanks — the fix is correct and minimal: the old `padding-inline-start: 0 !important` at `sidebar.css:237-240` genuinely deviated from the design system (Astryx's default childrenInner is already `--spacing-6`/24px), and the change is exactly one CSS value plus comment correction and tests. The `!important` necessity check is accurate (Astryx Stylex classes carry (3,1,0) specificity via `:not(#\#)` — Maka's (0,2,2) loses without it), and the `> div > [role="group"] > div` combinator chain lands only on childrenInner without hitting deeper nesting. Both the new unit test and the e2e geometry assertion run in CI (both green). Time-sort rows keeping the left edge while project-sort rows inset 24px is a stated, documented choice — fine as a product decision.

**Conclusion: PASS — all P3, nothing blocking.**

P3-1 — the contract test pins the *mechanism*, not the visual result: if someone later deletes this now-redundant override (pixel-identical), the regex match fails and `assert.ok(projectChildrenRule)` goes red — the test would block a zero-pixel simplification. Conversely, if Astryx changes its own nesting step in the future, the `!important` override keeps Maka silently stuck at 24px with all tests green, and the contract drifts from the design system. A computed-style assertion would pin the visual outcome instead.

P3-2 — the unit test's candidate-path fallback has a dead branch: in the desktop build `dist/renderer/styles/sidebar.css` doesn't exist (renderer goes through vite bundle), so the dist run always falls to the second candidate reading the source; the first candidate implies an intent that doesn't exist. Dropping the dead candidate (or asserting against the vite artifact explicitly) would make the intent clear.

P3-3 — the e2e threshold `toBeGreaterThanOrEqual(20)` at `e2e/sidebar-project-row.spec.ts:44` doesn't reference `--spacing-6` (24px) with a tolerance note; it only guards against "zeroed", and a spacing-5 change would pass it. A one-line comment tying it to the token would help. Also, the PR body says "Electron sidebar test" but it's a Playwright spec — minor description nit.

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagent on `opencode-go/deepseek-v4-flash`). The subagent verified specificity math, the descendant combinator chain, and CI execution paths statically; P3 items are static observations, not observed failures. Please weigh these findings with your own judgment.*

<details>

结论：PASS，全 P3。问题真实（main 上 padding-inline-start: 0 !important 偏离设计系统，Astryx 默认 childrenInner 本就是 --spacing-6/24px），改动最小（唯一行为变更是一个 CSS 值 0→var(--spacing-6)，其余是注释校正与测试）。!important 必要性判断准确（Astryx Stylex 类经 :not(#\#)×3 特异性达 (3,1,0)，Maka (0,2,2) 不加 !important 会输）；子组合器链只落到 childrenInner 不误伤更深嵌套。单测 + e2e 都在 CI 真实执行且全绿。P3-1：契约测试钉的是实现机制而非视觉结果——将来删掉这条像素零变化的多余 override 会被测试拦下；反向若 Astryx 改了自己的嵌套步进，!important 会让 Maka 静默停在 24px 且测试全绿，契约与设计系统悄悄脱节（建议改用 computed-style 断言）。P3-2：单测路径回退有死分支（desktop 构建里 dist/renderer/styles/sidebar.css 不存在，dist 运行态必落第二个 candidate 读源码）。P3-3：e2e 阈值 20 是魔法数，未指向 --spacing-6(24px) 容差；PR 描述称 Electron sidebar test 实为 Playwright spec。附注：时间排序贴左缘 vs 项目排序 24px inset 是明示取舍，属产品判断。
</details>

Astro-Han (approved):

Issue #3177, opened 2026-08-18
test(desktop): streaming-remount E2E intermittently misses the active Turn


M4n5ter:
## What happened

The Desktop E2E test `keeps a completed reply after an interrupted turn and conversation remount` intermittently times out while waiting for the Stop button after submitting the second Turn.

Observed failure:

```text
e2e/streaming-remount.spec.ts:80
Locator: getByRole("button", { name: "停止" })
Expected: visible
Timeout: 20000ms
```

The preceding assertion confirms that the user message is visible, but that does not necessarily fence the later state in which the replacement Turn is active and the Stop control is rendered. The failing assertion was introduced with the remount coverage in #3057.

Expected: once the large-response prompt is submitted, the test should reliably observe the active Turn and continue with steering.

Actual: on an otherwise unrelated PR, the user message became visible but the Stop button did not appear within 20 seconds.

## How to reproduce

1. Run the CI E2E suite
2. Observe `apps/desktop/e2e/streaming-remount.spec.ts:80`
3. The failure is intermittent: the focused test passed locally 3/3 times

Known occurrence: https://github.com/maka-agent/maka-agent/actions/runs/32088379562/job/95565550054

The failing PR does not modify the composer send path, Turn state projection, Stop control, or this test. All other CI jobs passed, including 962/962 Runtime Host tests.

## Environment

- Maka commit: `0c47963ff999c2f531682def1c58c7f8fc0d622b`
- Base commit: `5d9ce0d20`
- OS: GitHub-hosted Linux runner
- Surface: Desktop E2E
- Test origin: #3057

## Logs, screenshots, or additional context

This appears to be a missing readiness fence or an intermittent live-state projection failure around the second Turn after interruption/remount. The issue should be diagnosed from the durable Turn state and renderer subscription state at timeout rather than fixed by increasing the timeout.

Sun-GLiang:
I was able to reproduce this in a CI-like Docker environment.

### Environment

- Commit: `23a2a561a5cb7e145a806b932f32ed994d5417a7`
- Linux `amd64` container
- Node.js 24
- Xvfb
- The container was running under architecture emulation, so the observed failure rate should not be treated as representative of normal CI.

### Results

- The first focused run passed.
- Repeating the focused test 10 times produced 8 passes and 2 failures.
- Both failures timed out at the same Stop-button assertion in `apps/desktop/e2e/streaming-remount.spec.ts` (line 135 at this commit).

[...]

PR #3178, opened 2026-08-18
test(desktop): fence active turn state in streaming-remount e2e

rookepoole:
### What changed
- Re-ordered composer readiness check (`getByRole('button', { name: '发送' }).toBeEnabled()`) to precede `composer.fill(...)` in `apps/desktop/e2e/streaming-remount.spec.ts` so React state transitions after an interrupted turn settle before filling text.
- Added an `expect.poll` fence verifying `runningTurnIds.length > 0` on the session before expecting the Stop button to be rendered, ensuring active Turn state is fenced in durable session state.

Addresses #3177

---
Generated from a bounded immutable repository snapshot by PARS-Agent using Gemini 3.6 Flash. Tests listed above are recommendations unless GitHub checks report otherwise.

### DIFF
--- a/apps/desktop/e2e/streaming-remount.spec.ts
+++ b/apps/desktop/e2e/streaming-remount.spec.ts
@@ -124,14 +124,21 @@ test('keeps a completed reply after an interrupted turn and conversation remount
     ), originalSessionId!),
     { timeout: 20_000 },
   ).toBe(0);
-  await composer.fill(FAKE_WAIT_FOR_STEERING_LARGE_RESPONSE_PROMPT);
   await expect(page.getByRole('button', { name: '发送' })).toBeEnabled({
     timeout: 20_000,
   });
+  await composer.fill(FAKE_WAIT_FOR_STEERING_LARGE_RESPONSE_PROMPT);
   await composer.press('Enter');
   await expect(page.locator('.maka-user-message', {
     hasText: FAKE_WAIT_FOR_STEERING_LARGE_RESPONSE_PROMPT,
   })).toBeVisible();
+  await expect.poll(
+    () => page.evaluate(async (sessionId) => (
+      (await window.maka.sessions.list()).find((session) => session.id === sessionId)
+        ?.runningTurnIds?.length ?? 0
+    ), originalSessionId!),
+    { timeout: 20_000 },
+  ).toBeGreaterThan(0);
   await expect(page.getByRole('button', { name: '停止' })).toBeVisible({
     timeout: 20_000,
   });


Astro-Han:
Thanks for the attempt — but this change makes the test *deterministically fail*, the opposite of the flake fix you're going for. Please hold on merge until verified.

**Conclusion: FAIL — P1 + P2, both need resolution before merge.**

**P1 — the reordering inverts the assertion: the send button is disabled on an empty draft, so `expect(发送).toBeEnabled()` now times out every run.** The spec now asserts the button *before* `composer.fill(...)` (diff at spec ~128-131). At that point the composer is guaranteed empty: `sendCurrent` clears the input unconditionally after a successful submit (`packages/ui/src/composer.tsx:1078-1080` `textPort.setValue('')`), and the first hold-open turn has already been submitted in this test — nothing refills the draft (draft persistence only restores `new-task:`-prefixed keys). The send button's enabled state depends on a non-empty draft (`composer.tsx:1220-1224` `sendDisabled = ... || !text.trim() || ...`), and after the interrupted turn `streaming=false` with empty text, the slot renders a disabled 发送 (`composer.tsx:1796-1816`). So this turns an occasional flake into a guaranteed 20s timeout on every CI run. The original order (fill → assert enabled) was correct — the fence must evaluate *after* the text exists. Quick check: run the single spec (`playwright test streaming-remount.spec.ts -g 'keeps a completed reply'`) on this branch — it should time out deterministically.

**P2 — the new durable-poll fence likely won't fix #3177's failing case and is unverified as a fence.** The poll is honest (it reads the kernel's live `runningTurnIds` via `listSessions`, re-queried per call, so it can't hide a bad renderer — the later 停止 assertion still catches renderer failures). But in the #3177 failure shape — user message visible, 停止 not appearing — the user message only renders *after* the host accepted the turn, at which point durable `runningTurnIds` is already >0, so the poll passes immediately and the 停止 assertion still times out. It sharpens the failure boundary but doesn't address the renderer subscription/projection lag the issue itself suspects. The change as merged has no demonstrated effect on the flake: the PR body states tests are "recommendations unless GitHub checks report otherwise," and `gh pr checks` shows no test run (only CodeRabbit). Requirement: revert/fix the reorder (P1), run the spec locally multiple times on this branch to show the flake is actually gone (or that the failure mode is now correctly reported), and state what the fence is expected to prove (host-side start vs renderer-side projection are two different witnesses — this fence proves the former only).

P3 (optional): the new poll is a structural twin of the existing `runningTurnIds === 0` poll in the same file (same evaluate shape, same 20s timeout) — a small `runningTurnIdsOf(id)` helper would dedupe; no behavioral impact.

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagent on `opencode-go/deepseek-v4-flash`). The subagent traced the composer's send/disabled contract and the diff statically; the P1 timeout is a prediction from that contract (the spec was not executed under read-only constraints), so please confirm with a single-spec run. Please weigh these findings with your own judgment.*

<details>

结论：FAIL（P1 + P2，合并前需解决）。P1：重排让断言反转——发送按钮在空草稿时恒为 disabled，现在 expect(发送).toBeEnabled() 在 composer.fill() 之前断言，必然每次运行 20s 超时（从偶发 flake 变成确定性红）。原因：sendCurrent 提交成功后无条件清空输入（composer.tsx:1078-1080），测试里第一轮 hold-open turn 已提交、draft 无回填机制；发送按钮 enabled 依赖草稿非空（composer.tsx:1220-1224），中断轮结束后 streaming=false 且 text 为空，渲染为 disabled 的发送按钮。原顺序（fill 后再查 enabled）才是对的——fence 必须在文本存在之后求值。验证方式：单跑该 spec 应确定性超时。P2：新增的 durable 轮询大概率治不了 #3177 的失败类别——用户消息可见意味着 host 已接受 turn、runningTurnIds 早已 >0，轮询立即通过，随后 停止 断言照样 20s 超时；轮询只证明 host 侧状态，不能证明 renderer 订阅/投影通道健康（#3177 疑点所在）。且 PR 无实际测试运行记录（gh pr checks 只有 CodeRabbit），不能声称 fixes #3177。要求：回退/修正重排（P1）、本地连续多轮运行该 spec 证明 flake 真的消失、说明 fence 证明的是 host 侧启动而非 renderer 侧投影。P3：新轮询与文件内既有轮询是结构孪生，可抽 runningTurnIdsOf(id) 助手去重。
</details>

M4n5ter:
Thanks for taking the time to investigate #3177 and propose a synchronization fence.

We ultimately traced the intermittent failure to a production renderer race rather than a missing readiness check in the test: session settlement could combine an older catalog result with a newer live-Turn projection and incorrectly clear the Stop witness.

#3189 has now been merged with the root-cause fix. It binds settlement to the accepted authoritative list read, uses a per-projection compare-and-swap, projects observer-owned running Turn identities, and adds deterministic regression coverage.

This supersedes the test-only approach here, so we can close #3178 without asking you to make further changes. The current ordering would also check whether Send is enabled before filling the empty composer, as noted above, but there is no need to spend time revising it now unless an independent test improvement remains.

Thanks again for helping investigate the failure.

PR #3181, opened 2026-08-18
fix(cli,runtime-host): recover from turn consumer lag without eviction loops

### LINKED ISSUE #3180: fix(cli): TUI turn stream dies permanently when its local event consumer falls behind (no resubscribe, unlike Desktop)
## What happens

When the TUI's per-turn event consumer falls more than `MAX_PENDING_EVENTS_PER_TURN` (1024) events behind — e.g. a slow terminal under dense Agent Graph / tool activity — the `RuntimeHostSessionChannel` permanently kills that turn's event stream:

```ts
// packages/cli/src/runtime-host-session-channel.ts (SessionEventQueue.push)
if (this.#items.length >= MAX_PENDING_EVENTS_PER_TURN) {
  this.fail(new Error('Runtime Host Session event consumer is too slow'));
  return;
}
```

The Host turn keeps running, but `runMakaPiTuiTurn` (`drainGoalTurn`) sees the iterator reject, marks the turn as errored, and live projection stops. The channel itself stays healthy and keeps consuming host frames into the dead queue, so nothing ever recovers — the user only heals when the turn reaches its terminal transcript refresh. This is the same “stream appears frozen” symptom class that #2630 fixed on Desktop.

Note: **host-side slow-consumer eviction is already handled** — `#canRecover` covers `slow_consumer` and the channel resubscribes (verified with a driver test pushing `subscription.closed { reason: 'slow_consumer' }` on a reconnecting connection: the channel auto-reopened the subscription and the live turn stream continued). The gap is only the **client-local** slow-consumer path above, which never reaches the recovery machinery.

## Desktop reference

`apps/desktop/src/main/runtime-host-session-subscription-owner.ts` (#2630) treats local slowness as recoverable: its pump awaits frame acceptance, so a slow consumer overflows the bounded subscription queue, the failure surfaces as `RuntimeHostSubscriptionError('slow_consumer')`, and the owner re-establishes the subscription, restoring canonical state without the renderer resubscribing. The TUI has the equivalent recovery machinery (`#scheduleRecovery` / `#recover`) but the per-turn queue failure never invokes it.

## Proposal

Make the per-turn overflow recoverable instead of fatal:

- On overflow, shed offset-bearing deltas (`text_delta` / `thinking_delta`, healed by the next canonical resync / `text_complete`) and evict the oldest sheddable delta to make room for non-delta events, so terminal events still land — never fail the stream for slowness.
- Notify the channel once per lag episode (latched until the backlog drains); the channel retires the healthy-but-lagged subscription and runs the existing `#scheduleRecovery` path, resubscribing and re-syncing canonical state (transcript replacement heals the rendered view), matching Desktop behavior.

## Verification plan

- Driver tests: flood an unconsumed turn queue past 1024 deltas → assert the channel auto-resubscribes (replacement subscription opened), the turn stream never rejects, terminal events still arrive, and live deltas continue after recovery.
- Existing CLI suite stays green.

I’d like to work on this fix.

me2seeks:
## Summary

A TUI turn event consumer that fell more than `MAX_PENDING_EVENTS_PER_TURN` (1024) events behind had its stream failed permanently (`Runtime Host Session event consumer is too slow`) while the Host turn kept running; the channel kept draining frames into the dead queue and nothing recovered until the terminal transcript refresh — the same "stream appears frozen" class #2630 fixed on Desktop.

The queue now sheds offset-bearing deltas (`text_delta` / `thinking_delta`, healed by the next canonical resync or text completion) and evicts the oldest sheddable delta to make room for non-delta events so terminal records always land, and notifies the channel once per lag episode. The channel retires the healthy-but-lagged subscription through the existing `#scheduleRecovery` path — the same resubscribe a Host slow-consumer eviction triggers — instead of killing the stream.

Fixes #3180

## Verification

- `npm test` in `packages/cli`: 264 tests pass, including 2 new regression tests that flood an unconsumed turn stream past the bound — the channel resubscribes, the stream never rejects, live deltas continue after recovery, and terminal events still land while deltas are shed
- Biome check and `git diff --check` clean
- Pre-PR simplify-audit of the diff slice: no candidates, no decision gates

## AI use

Tool(s) and scope: Maka (this project's agent) investigated the Desktop/TUI divergence, authored the fix and its tests, and ran the simplify-audit review.

## Checklist

## Dogfood root-cause (a1ec77d2d)

Running this branch against a long resumed session reproduced a worse failure than #3180: the Host evicted the TUI's subscription as `slow_consumer` seconds after every (re)subscribe, because an active turn's thinking/text delta flood outpaces the coordinator's one-awaited-send-at-a-time flush and overflows the 32-frame subscriber queue. The channel's resubscribe path then either died permanently (a clean iterator end mid-catch-up was a non-recoverable error) or froze silently. Direct probe against the live host: fresh subscription evicted at +9.6s with 109/110 frames being deltas.

Two fixes:

- `runtime-host`: the coordinator coalesces a queued assistant delta into its queued tail when it continues the same stream contiguously (absolute `startOffset` semantics make a merged frame byte-identical in content; absorbed frames never spend a sequence). Eviction remains the backstop for genuinely undrainable backlogs.
- `cli`: a live stream that ends without `subscription.closed` is now `connection_closed`-recoverable, routing through the same resync instead of failing the channel.

Verification: `packages/runtime-host` 964/964 (incl. 2 new coalescing tests; eviction tests now flood with non-coalescable alternating streams); `packages/cli` 269/269 (incl. clean-end resubscribe regression, mutation-verified — reverting the channel fix hangs it, reverting coalescing fails the two new tests); biome clean.

Known remaining boundary: a truly wedged consumer (render loop stalled) still settles into the latched shed state by design — the run continues server-side and the transcript heals on the next attach; un-wedging the driver from render backpressure is follow-up work.

### DIFF
--- a/packages/cli/src/__tests__/runtime-host-session-driver.test.ts
+++ b/packages/cli/src/__tests__/runtime-host-session-driver.test.ts
@@ -1781,9 +1781,7 @@ describe('turn consumer lag recovery (#3180)', () => {
     await resynced.promise;
 
     await floodTurnStream(replacement, 1_024, 5, 'subscription-2');
-    replacement.push(
-      projectionFrame(1_025, completedTurn('turn-1', 'run-1'), 2, 'subscription-2'),
-    );
+    replacement.push(projectionFrame(1_025, completedTurn('turn-1', 'run-1'), 2, 'subscription-2'));
     await delay(0);
     assert.ok(
       await drainUntilDone(switched.activeTurn.events),
@@ -1804,9 +1802,7 @@ describe('turn consumer lag recovery (#3180)', () => {
     await floodToolStream(replacement, 1_024, 'subscription-2');
     // The terminal outcome must land even though no delta can be evicted;
     // process the frame before draining so the backlog is still full.
-    replacement.push(
-      projectionFrame(1_025, completedTurn('turn-1', 'run-1'), 2, 'subscription-2'),
-    );
+    replacement.push(projectionFrame(1_025, completedTurn('turn-1', 'run-1'), 2, 'subscription-2'));
     await delay(0);
     assert.ok(
       await drainUntilDone(switched.activeTurn.events),
@@ -1849,9 +1845,7 @@ describe('turn consumer lag recovery (#3180)', () => {
     // must land even though no delta can be evicted; otherwise the live tool
     // card stays running until the durable transcript heals it.
     replacement.push(toolResultFrame(1_025, 'subscription-2'));
-    replacement.push(
-      projectionFrame(1_026, completedTurn('turn-1', 'run-1'), 2, 'subscription-2'),
-    );
+    replacement.push(projectionFrame(1_026, completedTurn('turn-1', 'run-1'), 2, 'subscription-2'));
     await delay(0);
 
     let sawToolResult = false;
@@ -1881,9 +1875,7 @@ describe('turn consumer lag recovery (#3180)', () => {
     // (which would leave the live card stuck at "running" until the durable
     // transcript heals it).
     replacement.push(toolResultFrame(1_025, 'subscription-2'));
-    replacement.push(
-      projectionFrame(1_026, completedTurn('turn-1', 'run-1'), 2, 'subscription-2'),
-    );
+    replacement.push(projectionFrame(1_026, completedTurn('turn-1', 'run-1'), 2, 'subscription-2'));
     await delay(0);
 
     let sawToolResult = false;
--- a/packages/cli/src/__tests__/runtime-host-session-driver.test.ts
+++ b/packages/cli/src/__tests__/runtime-host-session-driver.test.ts
@@ -1680,13 +1680,24 @@ describe('turn consumer lag recovery (#3180)', () => {
     await delay(0);
   }
 
+  async function floodToolStream(
+    subscription: InstanceType<typeof FakeSubscription>,
+    count: number,
+  ): Promise<void> {
+    for (let index = 0; index < count; index += 1) {
+      subscription.push(toolStartFrame(index + 1, index));
+      if (index % 64 === 63) await delay(0);
+    }
+    await delay(0);
+  }
+
   async function waitForSubscriptions(connection: FakeConnection, count: number): Promise<void> {
     const deadline = Date.now() + WAIT_BUDGET_MS;
     while (connection.openedSubscriptions !== count && Date.now() < deadline) await delay(5);
     assert.equal(connection.openedSubscriptions, count);
   }
 
-  test('resubscribes instead of failing when a turn event consumer falls behind', async () => {
+  function lagRecoveryFixture() {
     const initial = new FakeSubscription(
       continuitySnapshot(),
       Promise.resolve([assistantMessage('turn-1', 'Hello')]),
@@ -1697,6 +1708,7 @@ describe('turn consumer lag recovery (#3180)', () => {
       'subscription-2',
     );
     const connection = new FakeConnection([initial, replacement], true);
+    const resynced = deferred<void>();
     const driver = createRuntimeHostMakaSessionDriver({
       connection: connection.value,
       cwd: '/tmp',
@@ -1704,69 +1716,98 @@ describe('turn consumer lag recovery (#3180)', () => {
       model: 'gpt-5',
       now: () => 50,
     });
+    driver.subscribeTranscriptReplacements!((_sessionId, _turnId, _messages, reason) => {
+      if (reason === 'reconnect') resynced.resolve();
+    });
+    return { initial, replacement, connection, driver, resynced };
+  }
+
+  async function drainUntilDone(events: AsyncIterable<unknown>): Promise<boolean> {
+    const iterator = events[Symbol.asyncIterator]();
+    let completed = false;
+    for (let index = 0; index < 1_200; index += 1) {
+      const result = await iterator.next();
+      if (result.done) return completed;
+      if ((result.value as { type?: string }).type === 'complete') completed = true;
+    }
+    return false;
+  }
+
+  test('resubscribes instead of failing when a turn event consumer falls behind', async () => {
+    const { initial, replacement, connection, driver, resynced } = lagRecoveryFixture();
     const switched = await driver.switchSession('session-1');
     assert.ok(switched.activeTurn);
 
     // Flood the unconsumed turn stream past its 1024-event bound.
     await floodTurnStream(initial, 1_100, 5);
 
-    // The channel retires the lagged subscription and resubscribes instead of
-    // failing the stream.
+    // The channel retires the lagged subscription, resubscribes, and compacts
+    // the sheddable backlog the canonical resync supersedes.
     await waitForSubscriptions(connection, 2);
+    await resynced.promise;
 
-    // Drain part of the backlog, then confirm live events keep flowing.
-    const iterator = switched.activeTurn.events[Symbol.asyncIterator]();
-    for (let index = 0; index < 100; index += 1) {
-      const result = await iterator.next();
-      assert.equal(result.done, false);
-    }
+    // The stream never rejected, and live events land right away.
     replacement.push(deltaFrame(1, 'turn-1', 5, ' world', 'subscription-2'));
-    let seen = '';
-    for (let index = 0; index < 1_100 && seen !== ' world'; index += 1) {
-      const result = await iterator.next();
-      assert.equal(result.done, false);
-      seen = (result.value as { text?: string }).text ?? '';
-    }
-    assert.equal(seen, ' world');
+    assert.equal((await nextEvent(switched.activeTurn.events)).text, ' world');
   });
 
   test('lands terminal events while shedding deltas from a lagging consumer', async () => {
-    const initial = new FakeSubscription(
-      continuitySnapshot(),
-      Promise.resolve([assistantMessage('turn-1', 'Hello')]),
-    );
-    const replacement = new FakeSubscription(
-      continuitySnapshot({ projectionRevision: 2 }),
-      Promise.resolve([assistantMessage('turn-1', 'Hello')]),
-      'subscription-2',
-    );
-    const connection = new FakeConnection([initial, replacement], true);
-    const driver = createRuntimeHostMakaSessionDriver({
-      connection: connection.value,
-      cwd: '/tmp',
-      llmConnectionSlug: 'openai-main',
-      model: 'gpt-5',
-      now: () => 50,
-    });
+    const { initial, replacement, connection, driver, resynced } = lagRecoveryFixture();
     const switched = await driver.switchSession('session-1');
     assert.ok(switched.activeTurn);
 
     await floodTurnStream(initial, 1_100, 5);
     await waitForSubscriptions(connection, 2);
+    await resynced.promise;
 
     replacement.push(projectionFrame(1, completedTurn('turn-1', 'run-1'), 2, 'subscription-2'));
-    const iterator = switched.activeTurn.events[Symbol.asyncIterator]();
-    let completed = false;
-    let done = false;
-    for (let index = 0; index < 1_200 && !done; index += 1) {
-      const result = await iterator.next();
-      if (result.done) {
-        done = true;
-        break;
-      }
-      if ((result.value as { type?: string }).type === 'complete') completed = true;
-    }
-    assert.ok(completed, 'terminal complete event survived the lagged backlog');
-    assert.ok(done, 'turn stream finished cleanly');
+    await delay(0);
+    assert.ok(
+      await drainUntilDone(switched.activeTurn.events),
+      'terminal complete event survived the lagged delta backlog',
+    );
+  });
+
+  test('admits a terminal outcome when the lagged backlog holds no deltas', async () => {
+    const { initial, replacement, connection, driver, resynced } = lagRecoveryFixture();
+    const switched = await driver.switchSession('session-1');
+    assert.ok(switched.activeTurn);
+
+    // Fill the bound with non-delta events: nothing sheddable to evict.
+    await floodToolStream(initial, 1_100);
+    await waitForSubscriptions(connection, 2);
+    await resynced.promise;
+
+    // The terminal outcome must land even though no delta can be evicted;
+    // process the frame before draining so the backlog is still full.
+    replacement.push(projectionFrame(1, completedTurn('turn-1', 'run-1'), 2, 'subscription-2'));
+    await delay(0);
+    assert.ok(
+      await drainUntilDone(switched.activeTurn.events),
+      'terminal complete event was admitted over a non-delta backlog',
+    );
   });
 });
+
+function toolStartFrame(
+  sequence: number,
+  index: number,
+  subscriptionId = 'subscription-1',
+): SubscriptionFrame {
+  return {
+    kind: 'subscription.session_event',
+    hostEpoch: 'host-1',
+    subscriptionId,
+    sequence,
+    sessionId: 'session-1',
+    runId: 'run-1',
+    event: {
+      type: 'tool_start',
+      id: `tool-${index}`,
+      turnId: 'turn-1',
+      ts: 10,
+      toolUseId: `tool-${index}`,
+      toolName: 'Bash',
+    },
+  };
+}
--- a/packages/cli/src/__tests__/runtime-host-session-driver.test.ts
+++ b/packages/cli/src/__tests__/runtime-host-session-driver.test.ts
@@ -1840,6 +1840,43 @@ describe('turn consumer lag recovery (#3180)', () => {
     assert.fail('stream ended without the terminal complete event');
   });
 
+  test('resubscribes when the live stream ends without a terminal close', async () => {
+    const initial = new FakeSubscription(
+      continuitySnapshot(),
+      Promise.resolve([assistantMessage('turn-1', 'Hello')]),
+    );
+    const replacement = new FakeSubscription(
+      continuitySnapshot({ projectionRevision: 2 }),
+      Promise.resolve([assistantMessage('turn-1', 'Hello world')]),
+      'subscription-2',
+    );
+    const connection = new FakeConnection([initial, replacement], true);
+    const driver = createRuntimeHostMakaSessionDriver({
+      connection: connection.value,
+      cwd: '/tmp',
+      llmConnectionSlug: 'openai-main',
+      model: 'gpt-5',
+      now: () => 50,
+    });
+    const switched = await driver.switchSession('session-1');
+    assert.ok(switched.activeTurn);
+    const transcript = deferred<StoredMessage[]>();
+    driver.subscribeTranscriptReplacements!((_sessionId, _turnId, messages, reason) => {
+      assert.equal(reason, 'reconnect');
+      transcript.resolve(messages);
+    });
+
+    // A clean iterator end with no subscription.closed frame — e.g. the Host
+    // evicted the subscription as a slow consumer while the channel was still
+    // buffering the catch-up transcript — used to fail the channel
+    // permanently. It must resubscribe and continue the live stream instead.
+    await initial.close();
+    assert.deepEqual(await transcript.promise, [assistantMessage('turn-1', 'Hello world')]);
+    assert.equal(connection.openedSubscriptions, 2);
+    replacement.push(deltaFrame(1, 'turn-1', 11, '!', 'subscription-2'));
+    assert.equal((await nextEvent(switched.activeTurn.events)).text, '!');
+  });
+
   test('re-arms lag detection exactly at the hysteresis watermark', async () => {
     const initial = new FakeSubscription(
       continuitySnapshot(),
--- a/packages/cli/src/__tests__/runtime-host-session-driver.test.ts
+++ b/packages/cli/src/__tests__/runtime-host-session-driver.test.ts
@@ -1683,9 +1683,13 @@ describe('turn consumer lag recovery (#3180)', () => {
   async function floodToolStream(
     subscription: InstanceType<typeof FakeSubscription>,
     count: number,
+    subscriptionId = 'subscription-1',
+    startSequence = 1,
   ): Promise<void> {
     for (let index = 0; index < count; index += 1) {
-      subscription.push(toolStartFrame(index + 1, index));
+      subscription.push(
+        toolStartFrame(startSequence + index, startSequence + index, subscriptionId),
+      );
       if (index % 64 === 63) await delay(0);
     }
     await delay(0);
@@ -1834,7 +1838,7 @@ describe('turn consumer lag recovery (#3180)', () => {
     }
     second.push(deltaFrame(1, 'turn-1', 5, ' world', 'subscription-2'));
     for (let index = 0; index < 100; index += 1) {
-      second.push(toolStartFrame(100 + index, 2_000 + index, 'subscription-2'));
+      second.push(toolStartFrame(2 + index, 2_000 + index, 'subscription-2'));
     }
     await delay(0);
     let fresh = '';
@@ -1846,8 +1850,8 @@ describe('turn consumer lag recovery (#3180)', () => {
     assert.equal(fresh, ' world');
 
     // A second lag episode is a new episode, not a dead latch: it triggers a
-    // fresh canonical resync.
-    await floodToolStream(second, 1_100);
+    // fresh canonical resync. The stream stays contiguous on `second`.
+    await floodToolStream(second, 1_100, 'subscription-2', 102);
     await waitForSubscriptions(connection, 3);
     await waitFor(() => resyncs === 2);
 
--- a/packages/cli/src/__tests__/runtime-host-session-driver.test.ts
+++ b/packages/cli/src/__tests__/runtime-host-session-driver.test.ts
@@ -1969,6 +1969,80 @@ describe('turn consumer lag recovery (#3180)', () => {
     assert.equal((await nextEvent(switched.activeTurn.events)).text, '!');
   });
 
+  for (const [name, replacementRoot] of [
+    ['the same terminal turn', completedTurn('turn-1', 'run-1')],
+    ['a successor turn', runningTurn('turn-2', 'run-2')],
+  ] as const) {
+    test(`preserves an unconsumed terminal event across a replacement with ${name}`, async () => {
+      const initial = new FakeSubscription(
+        continuitySnapshot(),
+        Promise.resolve([assistantMessage('turn-1', 'Hello')]),
+      );
+      const replacement = new FakeSubscription(
+        continuitySnapshot({ projectionRevision: 3, rootTurn: replacementRoot }),
+        Promise.resolve([
+          assistantMessage('turn-1', 'Hello'),
+          turnStateMessage('turn-1', 'completed'),
+          ...(replacementRoot.turnId === 'turn-2' ? [userMessage('turn-2', 'Continue')] : []),
+        ]),
+        'subscription-2',
+      );
+      const connection = new FakeConnection([initial, replacement], true);
+      const driver = createRuntimeHostMakaSessionDriver({
+        connection: connection.value,
+        cwd: '/tmp',
+        llmConnectionSlug: 'openai-main',
+        model: 'gpt-5',
+        now: () => 50,
+      });
+      const switched = await driver.switchSession('session-1');
+      assert.ok(switched.activeTurn);
+
+      initial.push(projectionFrame(1, completedTurn('turn-1', 'run-1'), 2));
+      await delay(0);
+      initial.fail(new RuntimeHostSubscriptionError('connection_closed', 'connection lost'));
+      await waitForSubscriptions(connection, 2);
+
+      assert.equal((await nextEvent(switched.activeTurn.events)).type, 'complete');
+      assert.equal((await switched.activeTurn.events[Symbol.asyncIterator]().next()).done, true);
+    });
+  }
+
+  test('exhausts recovery after repeated one-frame clean-EOF replacements', async () => {
+    const initial = new FakeSubscription(
+      continuitySnapshot(),
+      Promise.resolve([assistantMessage('turn-1', 'Hello')]),
+    );
+    const ended = Array.from({ length: 8 }, (_, index) => {
+      const subscription = new FakeSubscription(
+        continuitySnapshot({ projectionRevision: index + 2 }),
+        Promise.resolve([assistantMessage('turn-1', 'Hello')]),
+        `subscription-${index + 2}`,
+      );
+      subscription.push(deltaFrame(1, 'turn-1', 5, String(index), `subscription-${index + 2}`));
+      return subscription;
+    });
+    for (const subscription of ended) await subscription.close();
+    const connection = new FakeConnection([initial, ...ended], true);
+    const driver = createRuntimeHostMakaSessionDriver({
+      connection: connection.value,
+      cwd: '/tmp',
+      llmConnectionSlug: 'openai-main',
+      model: 'gpt-5',
+      now: () => 50,
+    });
+    const switched = await driver.switchSession('session-1');
+    assert.ok(switched.activeTurn);
+
+    await initial.close();
+    await assert.rejects(async () => {
+      for await (const _event of switched.activeTurn!.events) {
+        // Drain each replacement's single live frame until recovery fails.
+      }
+    }, /recovery exhausted its retry budget/u);
+    assert.equal(connection.openedSubscriptions, 9);
+  });
+
   test('re-arms lag detection exactly at the hysteresis watermark', async () => {
     const initial = new FakeSubscription(
       continuitySnapshot(),
--- a/packages/cli/src/__tests__/runtime-host-session-driver.test.ts
+++ b/packages/cli/src/__tests__/runtime-host-session-driver.test.ts
@@ -1807,6 +1807,26 @@ describe('turn consumer lag recovery (#3180)', () => {
     );
   });
 
+  test('drops the entire pre-resync tool backlog at the canonical cut', async () => {
+    const { initial, replacement, connection, driver, resynced } = lagRecoveryFixture();
+    const switched = await driver.switchSession('session-1');
+    assert.ok(switched.activeTurn);
+
+    await floodToolStream(initial, 1_100);
+    await waitForSubscriptions(connection, 2);
+    await resynced.promise;
+
+    replacement.push(toolStartFrame(1, 9_000, 'subscription-2'));
+    replacement.push(projectionFrame(2, completedTurn('turn-1', 'run-1'), 2, 'subscription-2'));
+    await delay(0);
+
+    const iterator = switched.activeTurn.events[Symbol.asyncIterator]();
+    const first = await iterator.next();
+    assert.equal(first.done, false);
+    assert.equal(first.value.type, 'tool_start');
+    if (first.value.type === 'tool_start') assert.equal(first.value.toolUseId, 'tool-9000');
+  });
+
   test('admits a tool result when the lagged backlog holds no deltas', async () => {
     const { initial, replacement, connection, driver, resynced } = lagRecoveryFixture();
     const switched = await driver.switchSession('session-1');
@@ -1904,6 +1924,51 @@ describe('turn consumer lag recovery (#3180)', () => {
     assert.equal((await nextEvent(switched.activeTurn.events)).text, '!');
   });
 
+  test('backs off several immediate clean-EOF replacements before recovering', async () => {
+    const initial = new FakeSubscription(
+      continuitySnapshot(),
+      Promise.resolve([assistantMessage('turn-1', 'Hello')]),
+    );
+    const ended = [2, 3, 4].map(
+      (index) =>
+        new FakeSubscription(
+          continuitySnapshot({ projectionRevision: index }),
+          Promise.resolve([assistantMessage('turn-1', 'Hello')]),
+          `subscription-${index}`,
+        ),
+    );
+    for (const subscription of ended) await subscription.close();
+    const stable = new FakeSubscription(
+      continuitySnapshot({ projectionRevision: 5 }),
+      Promise.resolve([assistantMessage('turn-1', 'Hello world')]),
+      'subscription-5',
+    );
+    const connection = new FakeConnection([initial, ...ended, stable], true);
+    const driver = createRuntimeHostMakaSessionDriver({
+      connection: connection.value,
+      cwd: '/tmp',
+      llmConnectionSlug: 'openai-main',
+      model: 'gpt-5',
+      now: () => 50,
+    });
+    const switched = await driver.switchSession('session-1');
+    assert.ok(switched.activeTurn);
+    const resynced = deferred<void>();
+    driver.subscribeTranscriptReplacements!((_sessionId, _turnId, _messages, reason) => {
+      if (reason === 'reconnect') resynced.resolve();
+    });
+
+    await initial.close();
+    await waitForSubscriptions(connection, 2);
+    await delay(5);
+    assert.equal(connection.openedSubscriptions, 2, 'the first repeated EOF is backoff-gated');
+
+    await resynced.promise;
+    assert.equal(connection.openedSubscriptions, 5);
+    stable.push(deltaFrame(1, 'turn-1', 11, '!', 'subscription-5'));
+    assert.equal((await nextEvent(switched.activeTurn.events)).text, '!');
+  });
+
   test('re-arms lag detection exactly at the hysteresis watermark', async () => {
     const initial = new FakeSubscription(
       continuitySnapshot(),
@@ -1934,11 +1999,15 @@ describe('turn consumer lag recovery (#3180)', () => {
     const switched = await driver.switchSession('session-1');
     assert.ok(switched.activeTurn);
 
-    // Latch the lag flag with a full non-delta backlog: all 1_024 queued
-    // events stay because nothing is sheddable.
+    // Latch the lag flag with a non-delta backlog. The canonical cut clears
+    // every pre-cut event, then a still-wedged consumer fills again without
+    // triggering a resubscribe loop.
     await floodToolStream(initial, 1_100);
     await waitForSubscriptions(connection, 2);
     await waitFor(() => resyncs === 1);
+    await floodToolStream(second, 1_100, 'subscription-2', 1);
+    await delay(20);
+    assert.equal(connection.openedSubscriptions, 2, 'the post-cut lag latch stayed armed');
 
     // Draining to one event above the watermark (513 pending) must NOT
     // re-arm: a fresh overflow on the still-latched queue is the same lag
@@ -1948,7 +2017,7 @@ describe('turn consumer lag recovery (#3180)', () => {
     for (let index = 0; index < 511; index += 1) {
       assert.equal((await iterator.next()).done, false);
     }
-    await floodToolStream(second, 600, 'subscription-2', 2_000);
+    await floodToolStream(second, 600, 'subscription-2', 1_101);
     await delay(20);
     assert.equal(connection.openedSubscriptions, 2, 'lag latch held above the watermark');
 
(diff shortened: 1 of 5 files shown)


me2seeks:
## Known limitation (and the two qodo findings)

Documenting one boundary I was already aware of, which overlaps with the qodo-code-review findings above:

**Terminal drop on a fully non-delta backlog** (qodo finding 2). When the queue is full of tool/control events with no sheddable delta to evict, an incoming terminal event (`complete` / `error` / `abort`) is dropped, and `finish()` then lets the consumer reach end-of-stream without the terminal outcome — the turn surfaces as `errored` ("Session turn ended without a completion event") instead. I judged this acceptable for v1: it requires a wedged consumer *and* an all-control 1024-deep backlog, and the outcome is still strictly better than before this PR (previously the stream failed permanently at the same depth; now only this pathological corner mis-reports). That said, guaranteed admission for terminal outcomes (let them exceed the bound, or evict any oldest event regardless of sheddability) is a small, local change — happy to fold it into this PR if reviewers prefer.

**Stale sheddable backlog after lag recovery** (qodo finding 1). Correct as stated, with one nuance: `seedActive(false)` does not replay assistant text, so the canonical resync heals the transcript via `onTranscriptReplaced`; the *live stream* hole (shed deltas; the TUI applies text deltas append-only and ignores `startOffset`) persists until `text_complete` replaces the text. Compacting the stale sheddable backlog in the queues at the canonical-replacement seam would shorten that window — also a candidate follow-up.

Both are refinements on top of this PR's guarantee (the stream no longer dies); neither reintroduces the permanent failure this PR removes. I'm glad to address either here or in a follow-up — maintainer's call.

packages/cli/src/runtime-host-session-channel.ts
```diff
@@ -560,12 +586,32 @@ class SessionEventQueue implements AsyncIterable<SessionEvent>, AsyncIterator<Se
...
-      this.fail(new Error('Runtime Host Session event consumer is too slow'));
-      return;
+      // A consumer that falls behind must not kill the stream. Shed
+      // offset-bearing deltas (the next canonical resync or completion heals
+      // them) and make room for every other event so terminal records always
+      // land; the channel resubscribes to re-sync state, like the Desktop
+      // subscription owner does (#2630).
+      if (isSheddableDelta(event)) {
+        this.#noteLag();
+        return;
```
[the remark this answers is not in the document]
me2seeks: Fixed in 4f58bc1cf: when the canonical replacement lands, every lagging turn queue now drops the sheddable backlog the resync supersedes (`shedLaggedDeltas`), so the fresh post-recovery stream is no longer shed behind stale deltas. Pinned by `resubscribes instead of failing when a turn event consumer falls behind`, which fails without the compaction (the post-resync marker delta is shed behind the full backlog).

packages/cli/src/runtime-host-session-channel.ts
```diff
@@ -560,12 +586,32 @@ class SessionEventQueue implements AsyncIterable<SessionEvent>, AsyncIterator<Se
...
+      }
+      const shedIndex = this.#items.findIndex(isSheddableDelta);
+      if (shedIndex === -1) {
+        this.#noteLag();
+        return;
```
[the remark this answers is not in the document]
me2seeks: Fixed in 4f58bc1cf: terminal outcomes (`complete` / `error` / `abort`) now get guaranteed admission — when the full backlog holds no sheddable delta to evict, the oldest event is evicted instead, so the consumer never reaches end-of-stream without a terminal outcome. Pinned by `admits a terminal outcome when the lagged backlog holds no deltas`, which fails without this branch.

packages/cli/src/runtime-host-session-channel.ts
```diff
@@ -560,12 +590,48 @@ class SessionEventQueue implements AsyncIterable<SessionEvent>, AsyncIterator<Se
...
+    if (!this.#lagging) return;
+    for (let index = this.#items.length - 1; index >= 0; index -= 1) {
+      if (isSheddableDelta(this.#items[index]!)) this.#items.splice(index, 1);
+    }
+  }
+
+  #noteLag(): void {
+    if (this.#lagging) return;
+    this.#lagging = true;
+    this.#onLag();
```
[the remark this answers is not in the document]
me2seeks: Addressed in a65cfdeca with the second half of the suggested remedy — permit a later canonical recovery when fresh output is shed. The latch now re-arms with hysteresis: once the consumer drains the backlog to half the bound, it has demonstrably made progress, so a later lag episode is treated as a new episode and schedules another canonical resync; a wedged consumer never drains, so it cannot loop resubscribes.

I deliberately did not take the other half (compacting the non-terminal backlog): unread tool/interaction events are not reconstructible from the transcript tail and dropping them would lose live UI state, whereas shed deltas are always healed by the canonical resync or `text_complete`.

Pinned by `recovers again when the consumer lags again after making progress`: fresh output lands after the first recovery, then a second lag episode (with 100 events still queued, so the backlog never empties) opens a third subscription. The test fails without the hysteresis re-arm.

me2seeks:
**CI note:** the `test_workspaces` failure was `packages/eval`'s `a request arriving while the proxy drains is refused, not counted` — a timing-sensitive proxy-drain test in a workspace this PR does not touch (`packages/cli` only). It passes 3/3 locally on this branch. The latest push re-triggers the lane; if it flakes again on the same test I'd call it a main-branch flake to track separately.

packages/cli/src/__tests__/runtime-host-session-driver.test.ts
```diff
@@ -1662,3 +1662,222 @@ async function waitFor(predicate: () => boolean): Promise<void> {
...
+    for (let index = 0; index < 425; index += 1) {
+      const result = await iterator.next();
+      assert.equal(result.done, false);
+      fresh = (result.value as { text?: string }).text ?? '';
+    }
+    assert.equal(fresh, ' world');
+
+    // A second lag episode is a new episode, not a dead latch: it triggers a
+    // fresh canonical resync.
+    await floodToolStream(second, 1_100);
```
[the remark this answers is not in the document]
me2seeks: Confirmed and fixed in 5ec410b77: the second subscription's stream now goes 1 (the fresh delta), 2..101 (the 100 trailing tool events), 102..1201 (the second flood), all with `subscriptionId: 'subscription-2'` — contiguous and correctly identified, so the test no longer relies on the fake's missing sequence validation.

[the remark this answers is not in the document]
me2seeks: (commit SHA updated to e9aed361d after a format-only amend; content unchanged)

Astro-Han:
Thanks for the fix — the diagnosis is right (#3180/#2630's frozen-stream failure is a dead queue: `push` fails the session at ≥1024 buffered events and the turn stream dies permanently), and the approach is the correct minimal one: reuse the host eviction's existing `#scheduleRecovery → #recover → #acceptCanonicalReplacement` resubscription path, which only works because the CLI connection is a reconnecting connection (`runtime-host-cli-context.ts:129`). I ran the test file locally: 27/27 pass, and tests 1/2/4 genuinely fail on main (old code fails the stream at event 1025), so the core claim holds: a lagged consumer is recovered instead of killing the stream. The race analysis is sound — pump teardown vs subscription swap is guarded, the `#ready`/`#recoveryTask`/per-queue `#lagging` latches prevent notification storms and re-subscribe loops, `shedLaggedDeltas` runs before the `seedActive` replay, and a consumer parked in `next()` can't hold a full backlog.

**Conclusion: PASS — one P2 (partially documented; see below), otherwise P3s.**

**P2 — non-delta, non-terminal events are silently dropped when there's no sheddable delta to evict, and the canonical resync does not replay them.** Your "Known limitation" covers the terminal case (`complete`/`error`/`abort` dropped on a fully non-delta backlog — fine, explicitly deferred for v1). What it doesn't cover: the *tool stream* events — `tool_output_delta`, `tool_progress`, `tool_result`/`tool_result_preview`, and `text_complete`/`thinking_complete` (`session-projector.ts:483,496-507,302`) — are also not sheddable, so during a tool-output flood a lagged consumer silently loses chunks of the live tool panel (tool stuck at "running", live text missing), and `seedActive`'s replay only covers text/thinking deltas, interactions, steering, and queue updates — never the tool stream. The durable transcript later heals the *messages* (`reconcileToolsWithStoredMessages`/`replaceTranscriptWithStoredMessages` in `pi-transcript.ts:308,340`), so the impact is limited to the live panels — which is exactly the class of symptom this PR is supposed to fix, shrunk to tool-card granularity, and no test pins the boundary. Either extend the shed/replay story to the tool stream, or document "tool-stream events may be lost during lag; terminal state is healed from the durable transcript" and add a test asserting that boundary. Suggest a regression test flooding `tool_output_delta` to pin the actual behavior.

P3: the `#retiringSubscriptions` guard swallows genuine concurrent errors on the retiring subscription's pump (recovery-path errors still surface via `#fail`, so not a silent deafness — a comment/test would pin it); the hysteresis boundary (`LAG_REARM_PENDING_EVENTS`) is only indirectly covered by test 4; the escalation from per-queue failure to session-wide recovery when an abandoned old turn's queue refills is benign but worth a comment.

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagent on `opencode-go/deepseek-v4-flash`). The subagent ran the test file locally (27/27) and traced the pump/recovery/seedActive paths; the P2 is a static trace of the non-delta drop path (no triggered failure observed), consistent with your own known-limitation note. Please weigh these findings with your own judgment.*

<details>

结论：PASS（1 P2，部分已被作者文档化）。诊断正确（#3180/#2630 的流冻结是死队列：push 在积压≥1024 时 fail 会话，turn 流永久死亡），方案最小且正确（复用 host eviction 的 #scheduleRecovery→#recover→#acceptCanonicalReplacement 重订阅路径，CLI 连接本就是 reconnecting 连接）。本地实跑测试文件 27/27 通过，测试 1/2/4 在 main 上必然失败（旧代码第 1025 个事件死流）——核心主张成立。竞态分析健全（泵拆除 vs 订阅交换有守卫、#ready/#recoveryTask/#lagging 闩锁防通知风暴和死循环、shedLaggedDeltas 先于 seedActive 重放、next() 等待态的消费者不可能持满积压）。P2：非 delta 非终局事件在无 delta 可逐出时被静默丢弃且 canonical resync 不重放——你的 Known limitation 只覆盖 terminal 事件，没覆盖工具流事件（tool_output_delta/tool_progress/tool_result/text_complete），工具输出洪峰时实时工具面板缺失、工具卡停留在"运行中"；durable 转录能治愈消息列表但影响限于实时面板（正是本 PR 想修的症状在工具卡粒度的缩小版），无测试界定该边界。建议把 shed/replay 扩展到工具流，或显式文档化"lag 期间工具流事件可能丢失、终局由 durable 转录治愈"并补 flood tool_output_delta 的回归测试。P3：#retiringSubscriptions 守卫吞掉退役订阅泵上的真实并发错误（恢复路径错误仍经 #fail 上浮，非静默失聪，建议注释/测试钉住）、hysteresis 边界仅测试 4 间接覆盖、单队列失败升级为会话级恢复属良性超集行为。
</details>

me2seeks:
先转回 draft：用本 PR 的构建 dogfood 时观察到一例疑似相关的卡住案例，需要定位清楚再交 review。

现象（resume 的会话 a24c5b58，新构建的 CLI + runtime-host）：
- turn 进行中 TUI 报了一次 error，随后 error 消失、界面刷新了一下（疑似走了重订阅/快照恢复路径）；
- 之后 TUI 永远停在同一帧（最后一个 tool result + Thinking…），不再渲染任何新事件；
- 但 runtime host 侧该 run 一直在正常推进：core_agent_run_events 与 runtime_events 持续增长（卡住后又产生了 200+ 事件），无 pending interaction；
- TUI 进程空转（18 分钟 8s CPU，ep_poll），与 runtime host 的 socket 双向 Recv-Q/Send-Q 均为 0 —— 即服务端不再向该 TUI 投递任何帧，而客户端也以为一切正常、没有再触发恢复。

初步怀疑方向：error → 恢复 → 刷新之后，新订阅的推流链路（server 端 per-subscriber flush 或 client 端 recovery 后的事件接收）存在静默中断的可能，且中断后没有任何一侧重新发现异常。正在继续定位根因，确认与本 PR 改动的关系后再转 ready。

me2seeks:
根因已定位并修复（a1ec77d2d），draft 状态等 CI 绿后转回 ready。

**Dogfood 复现链**（探针直连 live host 实测）：活跃 turn 的 thinking/text delta 洪峰（每秒几十条小帧）超过 coordinator 一帧一 awaited send 的排空速度，32 帧的订阅者队列秒级溢出 → Host 以 slow_consumer 驱逐 → 本分支的重订阅路径在洪峰下每次重订阅后几秒内再次被驱逐，形成驱逐-恢复循环；且循环中客户端有两条死路：① catch-up 期间订阅迭代器干净结束（closed 帧被缓冲未处理）→ 不可恢复的永久失败；② 恢复挂起 → 无错误静默冻结（用户实机卡的最终状态）。

**修复**：
- runtime-host：订阅者队列内对同一 assistant 流的连续 delta 做内容保真合并（按绝对 startOffset 语义，合并帧与原始帧序列逐字节等价，被吸收帧不占 sequence）——实测 64 帧洪峰合并为 1 帧，不再触发驱逐；真正排不动的积压仍走驱逐（交替流测试覆盖）。
- cli：live 流无 subscription.closed 直接结束 → 归为 connection_closed 可恢复错误，走既有 resync 恢复而非 #fail。

**验证**：runtime-host 964/964、cli 269/269；两处修复均做变异验证（回退 channel 修复 → 新回归测试挂起；回退合并 → 两个新测试挂）。

**遗留边界**（文档化在 PR body）：真正卡死的消费端（渲染循环停滞）最终仍停在 latched shed 态——run 在服务端继续，重连后 transcript 治愈；让 driver 不被渲染背压卡住是后续工作。

Astro-Han:
Thanks for the deep dogfooding — the delta-storm root cause and the fix direction are right (deltas shouldn't be able to trigger cancel semantics: the `queue.fail('too slow')` path is gone, `tool_output_delta` is now covered by `isSheddableDelta` with 4 new tests pinning "tool_result and complete still land after a tool_output_delta storm", terminal events get a guaranteed admission path, and resync sheds the lagged backlog). The channel-side latch/hysteresis/recovery design checks out with no re-entry storm (tests pin the water-level boundaries). Running your head locally: I built the protocol decoder against the repo and reproduced a frame-level failure the CI can't see — details below.

**Conclusion: FAIL — P1: the server-side merge can produce frames beyond the receiver's protocol limits, causing connection-level failure + recovery loop (the exact eviction loop this PR promises to kill, just moved from frame-count eviction to frame-size rejection).**

**P1 — merge budget vs receiver ceiling mismatch.** `session-continuity-coordinator.ts:1361-1382` merges `mergeableAssistantDeltaText` checking only the 256 KiB queue budget — it never validates the merged text/frame size — while the receiver hard-caps at `SESSION_SUBSCRIPTION_FRAME_MAX_BYTES` (64 KiB−1) per frame and `SESSION_LIVE_DELTA_MAX_BYTES` (16 KiB) per delta text (`session-continuity.ts:298, :678-681`). I reproduced it: three consecutive legally-shaped 16 KiB deltas (16334 B text, 16566 B encoded — each within limits) merge into a 49002 B text frame, and `decodeSubscriptionFrame` throws `invalid_frame` immediately. The merge budget (256 KiB) is far above the 64 KiB receiver ceiling, and `#pump` doesn't re-check size before sending, so the oversized frame is inevitably emitted and inevitably rejected.

**Predicted failure path:** under the exact dogfood storm (delta flood + lagging consumer with send backpressure), the merge tail accumulates past 16 KiB within seconds (tens of frames/sec × 3s, or 3 large chunks) → receiver rejects the frame → `decodeHostFrame` → `#fail` → the **entire connection goes terminal** (all subscriptions `connection_closed`, all pending requests rejected, transport aborted — `connection.ts:753, :790, :1021-1045`) → recovery resubscribes + reloads the full transcript → the storm continues → the merge tail exceeds again → reject again… a **connection-level recovery loop**, and each cycle kills every other subscription and in-flight request on that connection (wider blast radius than the pre-fix channel-local failure). Your 9.6s/109-frame dogfood sample (108 deltas) is well within trigger range. The server already has the precedent of capping every emitted frame (`:456` PTY path) and the delta-splitting logic (`:1510-1521`) — the merge path just lacks it.

**Fix (small):** bound the merge by `min(256 KiB queue budget, SESSION_SUBSCRIPTION_FRAME_MAX_BYTES − envelope, SESSION_LIVE_DELTA_MAX_BYTES − existing text)` and fall back to plain enqueue when exceeded (the existing `:1383-1387` eviction path backstops). Add a test feeding large deltas (~16 KiB each) and asserting `byteLength ≤ SESSION_SUBSCRIPTION_FRAME_MAX_BYTES` and merged text ≤ `SESSION_LIVE_DELTA_MAX_BYTES`, or run the merge output through `decodeSubscriptionFrame`. (The two existing merge tests only merge 64 tiny `chunk-N` frames ~500 B total, so green CI doesn't contradict this.)

P2 residual (your documented deferral — acceptable, but please name `tool_progress`/`tool_result` in the PR body's known-limitation list since code comments cover them but the body doesn't): with an all-non-delta 1024 backlog, `tool_progress`/`tool_result`/`text_complete`/`thinking_complete` are still silently discarded in the `push` else-branch, and canonical resync's `seedActive(false)` doesn't replay assistant text or tool streams (healed by durable transcript reload + terminal events). A one-line body clarification + one boundary test would close it.

P3 (optional): `#retiringSubscriptions`' guard swallows concurrent real errors on a retiring subscription's pump (argued safe via recovery-path re-exposure, but untested); server `slow_consumer` eviction on a non-mergeable storm (tool events/alternating streams) still loops recover→evict — pre-existing, but the PR's claim that "genuinely stuck consumers stabilise in latched shed state" only holds for mergeable delta storms, worth tightening the wording; shed text/thinking deltas aren't backfilled until `text_complete` (TUI is append-only, no `startOffset` consumer — documented).

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagent on `opencode-go/deepseek-v4-flash`), which built the protocol decoder and empirically reproduced the oversized-frame rejection on the PR head, and traced the merge/eviction/pump paths. P1 is a reproduced failure, not a prediction. Please weigh these findings with your own judgment.*

<details>
<summary>中文摘要</summary>

复评（head a1ec77d2）结论：**FAIL（P1）**——服务端合并帧可超出接收端协议上限，造成连接级失败+恢复循环（正是本 PR 承诺消灭的 eviction loop，只是从帧数驱逐变成帧大小拒绝）。P1：session-continuity-coordinator.ts:1361-1382 合并 mergeableAssistantDeltaText 只校验 256 KiB 队列预算，从不校验合并后的 text/帧大小；接收端硬上限 SESSION_SUBSCRIPTION_FRAME_MAX_BYTES（64KiB−1）/帧 + SESSION_LIVE_DELTA_MAX_BYTES（16 KiB）/delta text（session-continuity.ts:298,678-681）。已实测复现：3 个各自合法（16334 B text/16566 B 编码均在限内）的连续 16 KiB delta 合并成 49002 B text 帧后 decodeSubscriptionFrame 直接抛 invalid_frame。合并预算（256 KiB）远超接收端 64 KiB 上限，#pump 发送前不复查大小→超限帧必发必拒。预测失败路径：dogfood 洪峰（+消费落后+send 背压）数秒内合并 tail 超 16 KiB→接收端拒帧→decodeHostFrame→#fail→整个连接 terminal（全部订阅 connection_closed、全部 pending 请求 reject、transport abort，connection.ts:753,790,1021-1045）→恢复重订阅+全量 transcript 重载→洪峰继续→再超限→恢复循环以连接级形式复现，且每轮杀死该连接所有其他订阅与在途请求（爆炸半径大于修复前的 channel 局部失败）。你的 9.6s/109 帧 dogfood 样本完全在触发范围内。服务端已有先例约束每帧 ≤ 64 KiB（:456 PTY 路径）与 delta 拆分（:1510-1521），合并路径唯独缺。修复很小：合并预算改 min(256 KiB 队列预算, SESSION_SUBSCRIPTION_FRAME_MAX_BYTES−信封, SESSION_LIVE_DELTA_MAX_BYTES−已有 text)，超限回退普通入队（现有 :1383-1387 eviction 兜底）+ 补"大 delta 合并产物 ≤ 上限 / 过 decodeSubscriptionFrame"测试（现有两个合并测试只合并 64 个 ~500 B 小帧，CI 绿不构成反驳）。P2 残余（你已文档化延后，可接受，但建议 PR body 的 known limitation 点名 tool_progress/tool_result——代码注释已覆盖但 body 未点名）：全非-delta 1024 backlog 时 tool_progress/tool_result/text_complete/thinking_complete 仍在 push else 分支被静默丢弃，canonical resync seedActive(false) 不重放 assistant text/工具流（靠 durable transcript + terminal 事件治愈）；补一句 body 说明+一条边界测试即闭环。P3（可选）：#retiringSubscriptions 守卫吞并发真实错误（有安全论证无测试）；非可合并洪峰（工具事件/交替流）下服务端 slow_consumer 驱逐-恢复循环仍在（pre-existing），"真正卡死的消费端会稳定在 latched shed 态"仅对可合并 delta 洪峰成立，措辞应收敛；shed 的 text/thinking delta 在 text_complete 前不补（TUI 纯 append 无 startOffset 消费，已文档化）。
</details>

Astro-Han:
Important pushback welcome on this one — new finding on your head, CI can't see it:

**`session-continuity-coordinator.ts:1361-1382`** — the merge path checks only the 256 KiB queue budget and never the receiver's hard caps: `SESSION_SUBSCRIPTION_FRAME_MAX_BYTES` (64 KiB−1) per frame and `SESSION_LIVE_DELTA_MAX_BYTES` (16 KiB) per delta text. I reproduced it against the repo's own decoder: three legal 16 KiB deltas merge into a 49 KiB text frame → `decodeSubscriptionFrame` throws `invalid_frame` → the whole connection goes terminal + resubscribe → the storm re-fills the merge tail → loop. That's the eviction loop this PR promises to kill, just moved to frame-size rejection, with a wider blast radius (kills all subscriptions/requests on the connection).

If this reads wrong to you (e.g. merging only happens below a threshold I misread), please correct me — I'd rather be wrong than block on a false positive. Otherwise the fix is small: cap the merge by `min(256 KiB, FRAME_MAX − envelope, DELTA_MAX − existing text)` and fall back to plain enqueue, plus one boundary test.

---

*AI-assisted review: pi subagent reproduced the oversized-frame rejection on head a1ec77d2. This is a review finding, not a demand — if the pump path or merge conditions invalidate it, tell me and I'll recheck.*
<details>
<summary>中文摘要</summary>

欢迎反驳的新发现（CI 覆盖不到）：session-continuity-coordinator.ts:1361-1382 合并路径只校验 256 KiB 队列预算、不校验接收端硬上限（帧 ≤64KiB−1 / delta text ≤16 KiB）；已用仓库自带解码器实测复现：3 个合法 16 KiB delta 合并成 49 KiB 帧 → decodeSubscriptionFrame 抛 invalid_frame → 整连接 terminal+重订阅 → 洪峰再灌 → 循环——正是本 PR 承诺消灭的 eviction loop，只是变成帧大小拒绝且爆炸半径更大（连带杀死同连接所有订阅/请求）。若这个读法有误（例如合并只在某阈值以下发生），请纠正，我宁愿错了也不愿误报阻塞；否则修复很小：合并上限改 min(256 KiB, 帧上限−信封, delta 上限−已有 text)，超限回退普通入队 + 一条边界测试。
</details>

packages/runtime-host/src/server/session-continuity-coordinator.ts
```diff
@@ -1352,6 +1352,33 @@ export class SessionContinuityCoordinator implements SessionContinuityService {
...
+      const mergedText = mergeableAssistantDeltaText(tail.frame, frame);
+      if (mergedText !== undefined && tail.frame.kind === 'subscription.session_delta') {
+        const merged: SubscriptionFrame = {
+          ...tail.frame,
+          delta: { ...tail.frame.delta, text: mergedText },
+        };
+        const mergedEncodedBytes = encodeProtocolMessage(merged).byteLength;
+        if (
+          subscriber.queuedBytes - tail.encodedBytes + mergedEncodedBytes + terminalBytes <=
+          MAX_SUBSCRIBER_QUEUED_BYTES
```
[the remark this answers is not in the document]
me2seeks: Fixed in c0cab7984. Coalescing now reuses the protocol 16 KiB delta-text and 64 KiB subscription-frame limits in addition to the 256 KiB queue budget. The regression feeds individually valid deltas whose merge would exceed the cap and decodes every emitted frame with the receiver decoder; removing the guard fails the test.

packages/cli/src/runtime-host-session-channel.ts
```diff
@@ -560,12 +612,58 @@ class SessionEventQueue implements AsyncIterable<SessionEvent>, AsyncIterator<Se
...
+      } else if (isTerminalOutcome(event)) {
+        // Terminal outcomes always land, even when the backlog holds no
+        // delta to evict: without one the consumer reaches end-of-stream
+        // without a result.
+        this.#items.shift();
+      } else {
+        // A non-delta, non-terminal event with nothing sheddable to evict
+        // (e.g. tool_result behind an all-control backlog) is dropped. The
+        // durable transcript heals the final state; the live panel may show
+        // a stale tool card until then. Documented boundary for v1.
```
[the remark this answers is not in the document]
me2seeks: Fixed in 97bde9b67. tool_result now shares the priority-outcome admission path with complete/abort/error, so it displaces an older non-delta event when the 1,024-event queue is saturated. The flood regression fails when tool_result is removed from that predicate. CLI and Runtime Host full suites pass after rebasing to b9a514fe0.

packages/cli/src/runtime-host-session-channel.ts
```diff
@@ -361,6 +379,10 @@ export class RuntimeHostSessionChannel {
       this.#now,
       this.#subscription.activeAssistantStreams,
     );
+    // Deltas a lagging consumer has not seen are superseded by this canonical
+    // resync; keeping them would shed the fresh post-recovery stream behind
+    // them.
+    for (const queue of this.#turns.values()) queue.shedLaggedDeltas();
```
Astro-Han: **P2 — Drop the entire pre-resync backlog, not only deltas.** A queue can reach the lag threshold with non-sheddable `tool_start`, `tool_progress`, permission, or other control events. After the replacement transcript/snapshot is installed, this method leaves those old events queued, so the turn consumer replays stale starts/progress/interactions after `onTranscriptReplaced()` has already established newer canonical state. Treat resync as a sequence cut: clear all unconsumed pre-cut events, then seed the active/terminal/interaction state from the replacement snapshot and accept only frames from the replacement subscription. Add a lag test whose backlog contains non-delta tool/control events and assert they are not replayed after recovery.

packages/cli/src/runtime-host-session-channel.ts
```diff
@@ -263,9 +265,25 @@ export class RuntimeHostSessionChannel {
           this.#accept(frame);
         }
       }
-      if (!this.#closing) throw new Error('Runtime Host Session subscription ended unexpectedly');
+      // A stream that ends without a subscription.closed frame is a broken
+      // live channel, not a terminal state: the Host may have torn the
+      // subscription down mid-recovery (e.g. slow_consumer eviction while the
+      // transcript reload was still buffering). Route it through the same
+      // resync recovery as an explicit close instead of killing the channel.
+      if (!this.#closing)
```
Astro-Han, line 293: **P2 — Bound or back off repeated clean-EOF recovery.** This newly makes every clean EOF recoverable, while `#recover()` loops immediately on recoverable open/catch-up failures. A reconnecting transport that repeatedly opens a subscription which ends cleanly (for example a proxy or incompatible Host accepting then closing) can therefore churn subscriptions in a tight async loop indefinitely. Please put exponential backoff and a bounded/connection-state-aware retry policy at the reconnecting subscription boundary, with a test for several immediate clean-EOF replacements that proves there is no hot loop.

Astro-Han (commented):
Thanks for addressing both sides of the lag problem. Host-side contiguous delta coalescing reduces avoidable pressure, and converting CLI lag from a fatal channel error into canonical resynchronization is the right recovery model. The frame-size/sequence checks and terminal-outcome preservation are thoughtful. Independent reviewer passes and a separate read-only DeepSeek V4 Flash high-effort pass reviewed this head.

The remaining invariant is that a canonical resync is a cut: no event from before that cut may be replayed afterward unless it is intentionally regenerated from the replacement snapshot. Today only text/thinking/tool-output deltas are removed; queued tool/control events survive, so a slow consumer can observe stale progress, starts, or interaction events after `onTranscriptReplaced()` and the new snapshot have already established later state. Clean EOF is also newly routed into a recovery loop with no retry budget or backoff.

From first principles, replace the lagged queue wholesale at the canonical cut, seed the necessary active/terminal/interaction state from the replacement snapshot, and then accept only post-open frames. Put retry/backoff ownership at the reconnecting subscription boundary. This is simpler than classifying an expanding list of old event kinds as individually safe to retain.

I did not run a local test suite; this review is based on the latest head, focused static tracing, current green CI, independent `@reviewer` passes, and a read-only `ollama-cloud/deepseek-v4-flash` high-effort review. AI-assisted review; I verified the queue and recovery paths.

感谢同时处理 lag 的 Host 和 CLI 两侧。Host 合并连续 delta 可以减少不必要压力；CLI 从致命错误改为 canonical resync，也是正确的恢复模型。frame size、sequence 和 terminal outcome 的约束都考虑得很细。

剩余不变量是：canonical resync 必须形成一个 cut，cut 之前的事件不能在之后重新播放，除非由 replacement snapshot 明确重建。目前只删除 text/thinking/tool-output delta，排队中的 tool/control 事件仍保留，因此慢消费者可能在 `onTranscriptReplaced()` 和新 snapshot 已建立较新状态后，再收到旧 progress/start/interaction。clean EOF 也被新接入一个没有退避或次数上限的恢复循环。

更符合第一性原理和奥卡姆剃刀的方案，是在 canonical cut 直接替换整个 lagged queue，从 replacement snapshot 重建必要的 active/terminal/interaction 状态，之后只接收新 subscription 的 post-open frame；重试退避由 reconnecting subscription boundary 统一拥有。这样无需持续分类哪些旧事件可以保留。

本次未运行本地测试；结论来自最新 head 静态追踪、当前绿色 CI、独立 reviewer 和一次只读的 DeepSeek V4 Flash high-effort 审查。AI 辅助审查；我已人工复核 queue 与 recovery 路径。

</details>

packages/cli/src/runtime-host-session-channel.ts
me2seeks: Fixed in e3c4e1913. Canonical resync now clears the entire unconsumed pre-cut backlog, seeds replacement state from the authoritative snapshot, and admits only post-open frames. The mixed tool/control backlog regression proves no stale event replays after replacement.

packages/cli/src/runtime-host-session-channel.ts
me2seeks: Fixed in e3c4e1913. Clean-EOF recovery now uses a bounded exponential backoff at the reconnecting subscription boundary instead of a hot loop. The immediate-EOF sequence regression verifies the retry delays and terminal budget while preserving protocol frame caps.

packages/cli/src/runtime-host-session-channel.ts
```diff
@@ -251,6 +260,7 @@ export class RuntimeHostSessionChannel {
     try {
       for await (const frame of subscription) {
         if (this.#closing || this.#subscription !== subscription) return;
+        if (frame.kind !== 'subscription.closed') this.#markRecoveryStable(subscription);
```
Astro-Han: [P1] Resetting the attempt counter on the first non-closed frame makes the eight-attempt bound bypassable. A replacement that emits one catch-up/live frame and immediately reaches EOF resets to zero each time, causing unbounded open/frame/EOF recovery churn. Let only the existing `RECOVERY_STABLE_AFTER_MS` window establish stability; do not reset on one frame. Add a repeated one-frame-then-EOF test that eventually exhausts the budget.

packages/cli/src/runtime-host-session-channel.ts
```diff
@@ -560,12 +668,51 @@ class SessionEventQueue implements AsyncIterable<SessionEvent>, AsyncIterator<Se
...
+        return;
+      }
+      this.#noteLag();
     }
     this.#items.push(event);
   }
 
+  /** Drop the entire unseen pre-cut backlog after canonical replacement. */
+  cutBacklog(): void {
+    this.#items.length = 0;
```
Astro-Han, line 723: [P1] This can discard the only unconsumed `complete`/`error`/`abort` while leaving `#finishAfterItems = true`. If the replacement snapshot represents the same terminal turn, `sameTerminalTurn` suppresses reseeding; the next `next()` therefore returns clean EOF with no terminal event, and the TUI reports a successful turn as “ended without a completion event.” Preserve a queued guaranteed terminal across the cut, or skip cutting an already-finished queue, and test both same-terminal and successor replacement interleavings.

Astro-Han (commented):
The canonical replacement approach is the right owner-level response to consumer lag, and Host-side frame bounding remains coherent. Two P1 lifecycle failures remain in the CLI channel: the retry budget is reset by any single frame, permitting endless one-frame/EOF reconnect churn, and the canonical backlog cut can discard the only queued terminal outcome while retaining the queue's finished state. Both violate the PR's central promise that recovery converges without turning a successful terminal turn into an error.

AI-assisted review: Codex coordinated two independent reviewer passes and an OpenCode Go DeepSeek V4 Flash high-effort adversarial pass. I verified exact head `e3c4e1913a5fd0eb4e3bb41177ed029563d73c50`, queue state transitions, replacement seeding, retry/backoff behavior, and current CI. No local tests were run.

<details>
<summary>中文审查</summary>

canonical replacement 是处理 consumer lag 的正确责任层，Host 侧 frame bound 也保持一致。CLI channel 仍有两个 P1：任何单帧都会重置 retry budget，导致“一帧后 EOF”无限重连；canonical cut 会删除唯一的 terminal outcome，却保留 queue 已结束状态，使成功 turn 变成无 completion 的错误。两项都违反本 PR“恢复最终收敛”的核心目标。

本次为 AI 辅助审查，已核验精确 head、queue 状态转换、replacement seeding、retry/backoff 和当前 CI；未运行本地测试。

</details>

packages/cli/src/runtime-host-session-channel.ts
me2seeks: Fixed in 13c0e1b79. A canonical backlog cut now retains an already-queued unconsumed complete/error/abort whenever the queue is marked finish-after-items. Parameterized regressions cover both a same-terminal replacement and a successor-turn replacement, and both deliver the terminal exactly once before EOF. CLI passes 286/286.

packages/cli/src/runtime-host-session-channel.ts
me2seeks: Fixed in 13c0e1b79. Recovery attempts are no longer reset by the first non-closed frame; only the existing one-second stable subscription window resets the budget. A regression feeds eight replacements that each emit one frame and immediately reach clean EOF, then proves recovery exhausts after nine total subscriptions. CLI passes 286/286.

me2seeks (commented):
Publishing verified review remediation.

packages/cli/src/__tests__/runtime-host-session-driver.test.ts
```diff
@@ -1663,3 +1663,587 @@ async function waitFor(predicate: () => boolean): Promise<void> {
...
+
+    // The terminal outcome must land even though no delta can be evicted;
+    // process the frame before draining so the backlog is still full.
+    replacement.push(projectionFrame(1, completedTurn('turn-1', 'run-1'), 2, 'subscription-2'));
+    await delay(0);
+    assert.ok(
+      await drainUntilDone(switched.activeTurn.events),
+      'terminal complete event was admitted over a non-delta backlog',
+    );
+  });
```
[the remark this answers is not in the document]
me2seeks: Fixed in 3ac0ea023. After the canonical replacement callback, the four admission tests now refill the replacement queue to exactly 1,024 events: text deltas for shedding, non-delta tool starts for guaranteed terminal/tool-result admission, and tool-output deltas for the realistic noisy-tool path. The admitted outcome uses the next subscription sequence, so each assertion now exercises the full-queue branch instead of the pre-cut backlog. Focused CLI driver tests pass 40/40.

packages/cli/src/runtime-host-session-channel.ts
```diff
@@ -431,6 +465,40 @@ export class RuntimeHostSessionChannel {
...
+      );
+      await new Promise<void>((resolve) => setTimeout(resolve, delayMs));
+    }
+    this.#recoveryAttemptsWithoutLiveFrame += 1;
+  }
+
+  #scheduleRecoveryStable(subscription: RuntimeHostSessionSubscription): void {
+    this.#clearRecoveryStableTimer();
+    const timer = setTimeout(() => {
+      if (!this.#closing && this.#subscription === subscription && this.#ready) {
```
Astro-Han, line 504: [P1] The retry budget can still reset without any live frame. `#scheduleRecoveryStable()` starts immediately after transcript hydration, so a replacement that stays silent for one second and then reaches clean EOF resets `#recoveryAttemptsWithoutLiveFrame` to zero on every cycle. That permits unbounded open/load/silent/EOF churn even though the counter is explicitly scoped to attempts “without live frame.” Start the stability window only after the first post-hydration live frame (or track that condition separately), and add a replacement sequence that remains silent beyond the stability window before each EOF.

packages/cli/src/runtime-host-session-channel.ts
```diff
@@ -263,9 +272,25 @@ export class RuntimeHostSessionChannel {
...
+      // transcript reload was still buffering). Route it through the same
+      // resync recovery as an explicit close instead of killing the channel.
+      if (!this.#closing)
+        throw new RuntimeHostSubscriptionError(
+          'connection_closed',
+          'Runtime Host Session subscription ended unexpectedly',
+        );
     } catch (error) {
       if (this.#closing || this.#subscription !== subscription) return;
+      // A subscription retired because a turn consumer fell behind is closed
```
Astro-Han, line 300: [P2] Initial hydration can turn a recoverable slow-consumer close into an `open()` failure. While `#ready` is false, the pump buffers `subscription.closed`; if transcript loading resolves before the iterator reaches EOF, the pending-frame drain calls `#accept()` and throws outside the recovery branch, so the outer `open()` catch closes the channel instead of resubscribing. Route recoverable errors from the initial pending-frame drain through `#recover()` (or record the pending closure before hydration completes), and cover the interleaving where `subscription.closed: slow_consumer` arrives before transcript resolution while EOF is delayed.

packages/cli/src/runtime-host-session-channel.ts
```diff
@@ -560,12 +661,53 @@ class SessionEventQueue implements AsyncIterable<SessionEvent>, AsyncIterator<Se
...
+      if (shedIndex !== -1) {
+        this.#items.splice(shedIndex, 1);
+      } else if (isGuaranteedOutcome(event)) {
+        // Turn-terminal outcomes and tool results always land, even when the
+        // backlog holds no delta to evict: without a turn outcome the
+        // consumer reaches end-of-stream without a result, and without a
+        // tool_result the live tool card stays running until the durable
+        // transcript heals it. The oldest queued event is sacrificed.
+        this.#items.shift();
+      } else {
```
Astro-Han, line 707: [P2] A full non-delta backlog can discard the authoritative assistant completion. `text_complete` and `thinking_complete` are neither sheddable deltas nor guaranteed outcomes, so they hit this drop branch when 1,024 control events are queued. The projector has already marked that accumulator complete, and the later terminal projection skips settled accumulators; for `maka run`, `TurnOutcomeAccumulator` then receives `complete` without the `text_complete` that supplies `finalOutput`. Treat assistant completion events as guaranteed outcomes and add a full non-delta backlog → `text_complete` → terminal regression that asserts the final output.

Astro-Han (commented):
The layered direction is sound: Runtime Host reduces avoidable pressure by coalescing compatible deltas, while the CLI uses the existing canonical resubscription path as the recovery backstop. The current head also closes the previously reported single-frame retry reset and terminal-cut issues.

Three concrete edge cases remain in the admission and recovery boundaries below. They are local fixes rather than reasons to redesign the approach. Current required checks are green and the head is cleanly mergeable, but I recommend addressing these before approval.

Reviewed with Codex using three independent reviewer agents and OpenCode Go DeepSeek V4 Flash (high); I verified the findings against the exact current head and live CI.

整体分层方向是对的：Runtime Host 通过合并兼容 delta 降低可避免的压力，CLI 则复用现有 canonical resubscription 作为恢复后盾。当前 head 也已经修复此前提出的单帧重置重试预算和 terminal cut 问题。

目前还剩下下面三个具体的 admission / recovery 边界问题。它们都可以局部修复，不需要重新设计整体方案。当前必需检查全绿、分支可干净合并，但建议修复后再批准。

本次由 Codex 配合三个独立 reviewer agent，以及 OpenCode Go DeepSeek V4 Flash（high）审查；我按当前精确 head 和实时 CI 复核了这些结论。

</details>

packages/cli/src/runtime-host-session-channel.ts
me2seeks: Fixed in 930fdeaf3. Recovery stability is now armed only by the first non-closed frame received after replacement hydration; transcript hydration and subscription.closed never reset the budget. The new regression keeps the first replacement silent beyond the one-second window, then proves the remaining replacements exhaust the eight-attempt budget. Focused driver tests pass 43/43 and the full CLI suite passes 289/289.

packages/cli/src/runtime-host-session-channel.ts
me2seeks: Fixed in 930fdeaf3. text_complete and thinking_complete now share the guaranteed-outcome admission path with turn terminals and tool_result, so a saturated all-control backlog cannot discard an authoritative assistant accumulator. The regression refills the replacement queue to exactly 1,024 non-delta events, sends text_complete followed by terminal projection, and asserts Hello final answer survives as final output. Focused driver tests pass 43/43 and the full CLI suite passes 289/289.

packages/cli/src/runtime-host-session-channel.ts
me2seeks: Fixed in 930fdeaf3. Initial hydration now catches recoverable errors while draining buffered frames and enters the same canonical recovery path instead of letting open close the channel. The regression buffers subscription.closed with reason slow_consumer before transcript resolution while EOF remains delayed, then verifies replacement hydration and live delivery. Focused driver tests pass 43/43 and the full CLI suite passes 289/289.

Astro-Han (approved):
The latest head closes all three previously reported boundaries: retry stability now requires a post-hydration live frame, initial buffered slow-consumer closure enters canonical recovery, and assistant completion events receive guaranteed admission under a full non-delta backlog. The new regressions exercise each exact interleaving, including final output preservation.

I found no remaining concrete P0–P3 in the changed queue/recovery paths. One conservative edge remains: a live frame buffered during replacement hydration does not count toward stability, which can exhaust the retry budget earlier than necessary; that fails closed and is suitable as a follow-up rather than a blocker.

The current `test_runtime_host` run failed in the unrelated owned-candidate prompt-exit test while affected CLI coverage passed. This approval is for the code at the exact head; the failed required check still needs a successful rerun before merge.

Reviewed with Codex using two independent reviewer agents and OpenCode Go DeepSeek V4 Flash (high); I verified the exact delta, all prior findings, recovery/admission ordering, focused tests, and live CI.

最新 head 已修复之前三个边界：retry stability 只有在 hydration 后收到 live frame 才成立；initial buffered slow-consumer close 会进入 canonical recovery；full non-delta backlog 下 assistant completion 也获得 guaranteed admission。新增回归测试覆盖了每个精确交错，包括 final output 保留。

当前 queue/recovery 增量中未发现剩余具体 P0–P3。仍有一个保守边界：replacement hydration 期间已缓冲的 live frame 不计入 stability，可能较早耗尽 retry budget；该行为 fail closed，适合作为 follow-up，不阻塞本次。

当前 `test_runtime_host` 在无关的 owned-candidate prompt-exit 测试中失败，而受影响 CLI coverage 已通过。本次批准针对精确 head 的代码；合并前仍需把失败的 required check 重跑成功。

本次由 Codex 配合两个独立 reviewer agent，以及 OpenCode Go DeepSeek V4 Flash（high）审查；我核验了精确增量、全部既有 finding、recovery/admission 顺序、聚焦测试和实时 CI。

</details>

me2seeks:
The exact-head CI failure is runner-side: test_workspaces spent about 4h43m in Install Linux runtime dependencies and was cancelled before npm install, build, or any workspace test started. All other lanes on this head passed, and the existing exact-head approval remains valid. I attempted a failed-lane rerun, but GitHub requires repository admin rights for this run. Maintainer action requested: rerun the failed jobs for https://github.com/apache/maka/actions/runs/32228596412. I have not changed the approved head and have not merged.

PR #3189, opened 2026-08-18
fix(desktop): preserve live turns across session refreshes

### LINKED ISSUE #3177: test(desktop): streaming-remount E2E intermittently misses the active Turn

M4n5ter:
## Summary

- bind transient settlement to the accepted authoritative session-list read and the exact live projections observed before that read began
- clear only an unchanged per-session projection with an identity compare-and-swap, so a concurrent confirmation, replacement, or content advance cannot be retired by an older catalog snapshot
- merge observer-owned nonterminal root Turn identities into Desktop session-list summaries without persisting them or changing Runtime Host wire data

Fixes #3177

## Root cause

Settlement previously combined two independently versioned inputs: a React `sessions` state snapshot and whatever live-Turn projection happened to be current when an effect ran. Changing the effect phase could narrow one scheduling window, but it could not establish causality. A list result could already be queued, a `sessions:changed` notification could then confirm or replace the Turn, and React could commit the older list afterward; reconciliation would treat that older catalog state as evidence about the newer projection and clear the Stop witness.

Runtime Host-backed Desktop lists also omitted the observer's current running Turn identity, so later catalog refreshes could not preserve a live Turn when the durable header lagged or was stale.

## Design

`createSessionListRefresher` now captures renderer context immediately before each authority read and passes it only with the accepted generation. Settlement runs at that acceptance boundary, before the list enters React state. The state controller conditionally clears a projection only when it is still the same object captured for that read. Superseded responses discard both their catalog result and their context, while a trailing read captures a fresh context.

This removes the React effect and adds no persisted state, protocol field, or renderer epoch. `SessionSummary.runningTurnIds` remains an ephemeral list projection sourced from the existing Runtime Host observer.

## Verification

- compiled Desktop test suite — 904/904 passed
- focused causal reconciliation, request-context, and Runtime Host observer tests — 45/45 passed
- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm --workspace @maka/desktop run build`
- independent correctness and design reviews found no actionable findings
- scoped simplification audit found no deletion candidate, deletion probe, or architecture decision gate

## AI use

Select exactly one:

Tool(s) and scope: OpenAI Codex investigated the race, implemented the fix, and added regression coverage under M4n5ter's direction and review. M4n5ter accepts responsibility for the result.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/src/main/__tests__/runtime-host-session-catalog-running-turns.test.ts
+++ b/apps/desktop/src/main/__tests__/runtime-host-session-catalog-running-turns.test.ts
@@ -1,0 +1,65 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+import type { SessionCatalogProjection } from '@maka/runtime-host/protocol';
+import type { IpcHandler } from '../ipc-reconnect-policy.js';
+import { registerRuntimeHostSessionCatalogIpc } from '../runtime-host-session-catalog-ipc-main.js';
+
+test('projects observed running Turn identities into renderer Session lists', async () => {
+  const handlers = new Map<string, IpcHandler>();
+  registerRuntimeHostSessionCatalogIpc(
+    {
+      client: {
+        listSessions: async () => [session('running'), session('idle')],
+      } as never,
+      runningTurnIds: (sessionId) =>
+        sessionId === 'running' ? ['turn-live'] : [],
+      resolveCreateProject: async () => ({ kind: 'host_path', path: '/workspace' }),
+      emitSessionsChanged() {},
+      releaseSessionResources() {},
+      sessionCopyCleanup: {
+        recover: async () => ({ failed: [] }),
+      } as never,
+    },
+    {
+      handle: (channel, listener) => handlers.set(channel, listener),
+      handleReconnectableRead: (channel, listener) => handlers.set(channel, listener),
+    },
+  );
+
+  const list = handlers.get('sessions:list');
+  assert.ok(list);
+  const projected = await list({} as never) as Array<{ id: string; runningTurnIds?: string[] }>;
+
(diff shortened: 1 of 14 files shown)


M4n5ter:
fast path: The current changes are minor and reversible

PR #3191, opened 2026-08-18
fix(eval): keep leftover processes through framework timeout

### LINKED ISSUE #3150: eval: define a consistent framework-timeout policy for background processes across arms
Follow-up from #2971 review. Arms differ in how subject-owned background processes survive agent shutdown vs framework timeout (see the DSH PTY-tree teardown found in the 89-task run). Define one policy, apply it to every arm, and test it. (Raised by @M4n5ter in the approving review.)

1625567290:
## Summary

The verifier still scores a trial that hits the framework timeout (`subject_failed` plus the verifier reward). The relay used the same cancel path as host abort, so it signalled the recorded process group and could delete the environment — leftover services survived a clean subject exit and disappeared on timeout.

- Framework timeout stops only the subject leader so `environment.exec` can return. It does not signal the process group.
- If the leader never acknowledges TERM/KILL, the relay fails closed instead of publishing a scoreable timeout frame over an environment the subject may still be mutating.
- Host abort still settles or destroys, because that trial is abandoned.
- The Maka hosted-execution client uses `abortPolicy: preserve_environment` on this path, so a timeout abort does not cancel Host-owned work the verifier is about to score.

Fixes #3150

## Verification

- `python3 harbor/test_relay_lifecycle.py` — 14 pass, 3 skipped
- `python3 harbor/test_relay_contract.py` — 20/20

## Checklist

- [x] Yes — a framework timeout no longer tears down leftover subject processes or deletes the environment when the subject stops; an unconfirmed timeout is no longer reported as a scoreable `framework_timeout`

### DIFF
--- a/packages/eval/harbor/relay_agent.py
+++ b/packages/eval/harbor/relay_agent.py
@@ -29,6 +29,7 @@ import re
 import shlex
 import tempfile
 from pathlib import Path
+from types import SimpleNamespace
 from typing import Any
 
 from eval_framework import selected
@@ -91,6 +92,7 @@ class RelayAgent(BaseAgent):
         relay_port: int,
         relay_token: str,
         teardown_timeout_ms: int,
+        framework_timeout_ms: int | None = None,
         **kwargs: Any,
     ):
         super().__init__(*args, **kwargs)
@@ -100,6 +102,23 @@ class RelayAgent(BaseAgent):
         if not isinstance(teardown_timeout_ms, int) or teardown_timeout_ms <= 0:
             raise RuntimeError("Maka Eval teardown timeout is invalid")
         self._teardown_timeout = teardown_timeout_ms / 1000
+        if framework_timeout_ms is None:
+            framework_timeout_ms = teardown_timeout_ms
+        if not isinstance(framework_timeout_ms, int) or framework_timeout_ms <= 0:
+            raise RuntimeError("Maka Eval framework timeout is invalid")
+        self.set_framework_timeout_ms(framework_timeout_ms)
+
+    def set_framework_timeout_ms(self, timeout_ms: int | None) -> None:
+        if timeout_ms is None:
+            self._framework_timeout = self._teardown_timeout
+            return
+        if (
+            isinstance(timeout_ms, bool)
+            or not isinstance(timeout_ms, int)
+            or timeout_ms <= 0
+        ):
+            raise RuntimeError("Maka Eval framework timeout is invalid")
+        self._framework_timeout = min(timeout_ms / 1000, self._teardown_timeout)
 
     @staticmethod
     def name() -> str:
@@ -116,6 +135,7 @@ class RelayAgent(BaseAgent):
         execution: asyncio.Task[Any] | None = None
         decision: asyncio.Task[dict[str, Any]] | None = None
         request: dict[str, Any] | None = None
+        cwd = ""
         execution_reported = False
         scope_path = f"/logs/agent/.maka-eval-{self._token}.pid"
         environment_path = f"/tmp/maka-eval-{self._token}.env"
@@ -153,6 +173,9 @@ class RelayAgent(BaseAgent):
                 decision.result()
                 raise RelayTransportClosed("Maka Eval relay received control before execution")
             result = execution.result()
+            result = await _load_captured_subject_outputs(
+                environment, cwd, result, request, self._token
+            )
             await _persist_subject_outputs(environment, result)
             stdout, diagnostic = _project_result(result, request)
             # A subject that exited on its own leaves the shared environment as
@@ -160,8 +183,9 @@ class RelayAgent(BaseAgent):
             # waiting on those processes here — `environment.exec` has already
             # returned — so tearing them down would not unblock anything; it
             # would only edit the thing about to be measured, and edit it for
-            # some subjects and not others. Cancellation still quiesces, because
-            # there the subject has not stopped and the trial is being abandoned.
+            # some subjects and not others. Host abort still quiesces, because
+            # that trial is abandoned. Framework timeout does not: the verifier
+            # still scores what the subject left.
             if not await _send(
                 writer,
                 {
@@ -181,60 +205,25 @@ class RelayAgent(BaseAgent):
                 "verify",
             )
         except asyncio.CancelledError:
+            # A cancelled task would otherwise fail every await in this
+            # handler. Uncancel so cleanup can run; a second cancel or any
+            # cleanup error destroys the environment before we re-raise.
+            current = asyncio.current_task()
+            if current is not None and hasattr(current, "uncancel"):
+                current.uncancel()
+            framework_timeout_handled = False
             if request is not None and execution is not None:
-                execution_terminal = execution.done() and not execution.cancelled()
-                terminal_projection = None
-                if execution_terminal:
-                    terminal_result = execution.result()
-                    terminal_projection = _project_result(terminal_result, request)
-                # A subject that already exited has nothing left to settle, and
-                # tearing its scope down here would remove what the verifier is
-                # about to score -- the same environment edit this relay stopped
-                # making on the ordinary path. Only a subject still running is
-                # brought to a stop.
-                if terminal_projection is not None:
-                    result = terminal_result
-                else:
-                    result = await _settle_or_destroy(
-                        environment, cwd, scope_path, execution, self._teardown_timeout
-                    )
-                if result is not None:
-                    await _persist_subject_outputs(environment, result)
-                if (
-                    result is not None
-                    and not execution_reported
-                    and (execution_terminal or not _host_teardown_requested)
-                ):
-                    stdout, diagnostic = terminal_projection or _project_result(result, request)
-                    with contextlib.suppress(Exception):
-                        await _send(
-                            writer,
-                            {
-                                "token": self._token,
-                                "kind": "executed",
-                                "termination": "exited" if execution_terminal else "framework_timeout",
-                                "exitCode": result.return_code if execution_terminal else 124,
-                                "stdout": stdout,
-                                "diagnostic": diagnostic,
-                            },
-                        )
-                elif not execution_reported and not _host_teardown_requested:
-                    with contextlib.suppress(Exception):
-                        await _send(
-                            writer,
-                            {
-                                "token": self._token,
-                                "kind": "executed",
-                                "termination": "framework_timeout",
-                                "exitCode": 124,
-                                "stdout": "",
-                                "diagnostic": (
-                                    _carrier_diagnostic("result-frame-missing", b"")
-                                    if request.get("captureStdout", True)
-                                    else {"category": "none"}
-                                ),
-                            },
-                        )
+                framework_timeout_handled = await self._cleanup_cancelled_execution(
+                    environment,
+                    cwd,
+                    scope_path,
+                    execution,
+                    request,
+                    writer,
+                    execution_reported,
+                )
+            if framework_timeout_handled and not _host_teardown_requested:
+                return
             raise
         except RelayTransportClosed:
             if request is not None and execution is not None:
@@ -243,6 +232,9 @@ class RelayAgent(BaseAgent):
                 )
                 if result is not None:
                     with contextlib.suppress(Exception):
+                        result = await _load_captured_subject_outputs(
+                            environment, cwd, result, request, self._token
+                        )
                         await _persist_subject_outputs(environment, result)
         except BaseException:
             if request is not None and execution is not None:
@@ -251,6 +243,9 @@ class RelayAgent(BaseAgent):
                 )
                 if result is not None:
                     with contextlib.suppress(Exception):
+                        result = await _load_captured_subject_outputs(
+                            environment, cwd, result, request, self._token
+                        )
                         await _persist_subject_outputs(environment, result)
             raise
         finally:
@@ -260,9 +255,16 @@ class RelayAgent(BaseAgent):
                     await decision
             if request is not None:
                 with contextlib.suppress(Exception):
+                    stdout_path, stderr_path, stdout_fifo, stderr_fifo, collectors_path = (
+                        _capture_paths(self._token)
+                    )
                     await asyncio.wait_for(
                         environment.exec(
-                            f"rm -f -- {shlex.quote(scope_path)} {shlex.quote(environment_path)}",
+                            "rm -f -- "
+                            f"{shlex.quote(scope_path)} {shlex.quote(environment_path)} "
+                            f"{shlex.quote(stdout_path)} {shlex.quote(stderr_path)} "
+                            f"{shlex.quote(stdout_fifo)} {shlex.quote(stderr_fifo)} "
+                            f"{shlex.quote(collectors_path)}",
                             cwd=cwd,
                             timeout_sec=1,
                         ),
@@ -271,6 +273,101 @@ class RelayAgent(BaseAgent):
             with contextlib.suppress(BrokenPipeError, ConnectionError, RuntimeError, TimeoutError):
                 writer.close()
                 await asyncio.wait_for(writer.wait_closed(), timeout=1)
+
+    async def _cleanup_cancelled_execution(
+        self,
+        environment: Any,
+        cwd: str,
+        scope_path: str,
+        execution: asyncio.Task[Any],
+        request: dict[str, Any],
+        writer: Any,
+        execution_reported: bool,
+    ) -> bool:
+        cleanup_timeout = (
+            self._teardown_timeout
+            if _host_teardown_requested
+            else self._framework_timeout
+        )
+        deadline = asyncio.get_running_loop().time() + cleanup_timeout
+        try:
+            return await self._finalize_cancelled_execution(
+                environment,
+                cwd,
+                scope_path,
+                execution,
+                request,
+                writer,
+                execution_reported,
+                cleanup_timeout,
+            )
+        except BaseException:
+            remaining = deadline - asyncio.get_running_loop().time()
+            with contextlib.suppress(Exception):
+                if remaining > 0:
+                    await asyncio.wait_for(
+                        _settle_or_destroy(
+                            environment, cwd, scope_path, execution, remaining
+                        ),
+                        timeout=remaining,
+                    )
+            raise
+
+    async def _finalize_cancelled_execution(
+        self,
+        environment: Any,
+        cwd: str,
+        scope_path: str,
+        execution: asyncio.Task[Any],
+        request: dict[str, Any],
+        writer: Any,
+        execution_reported: bool,
+        cleanup_timeout: float,
+    ) -> bool:
+        execution_terminal = execution.done() and not execution.cancelled()
+        # A subject that already exited has nothing left to settle, and
+        # tearing its scope down here would remove what the verifier is
+        # about to score -- the same environment edit this relay stopped
+        # making on the ordinary path. A still-running subject is
+        # stopped so the execution call can return. Host abort then
+        # quiesces the leftover group; framework timeout does not,
+        # because the verifier still runs.
+        if execution_terminal:
+            result = execution.result()
+        elif _host_teardown_requested:
+            result = await _settle_or_destroy(
+                environment, cwd, scope_path, execution, cleanup_timeout
+            )
+        else:
+            result = await _stop_subject_for_timeout(
+                environment, cwd, scope_path, execution, cleanup_timeout
+            )
+        if result is not None:
+            result = await _load_captured_subject_outputs(
+                environment, cwd, result, request, self._token
+            )
+            await _persist_subject_outputs(environment, result)
+        if (
+            result is not None
+            and not execution_reported
+            and (execution_terminal or not _host_teardown_requested)
+        ):
+            stdout, diagnostic = _project_result(result, request)
+            try:
+                execution_reported = await _send(
+                    writer,
+                    {
+                        "token": self._token,
+                        "kind": "executed",
+                        "termination": "exited" if execution_terminal else "framework_timeout",
+                        "exitCode": result.return_code if execution_terminal else 124,
+                        "stdout": stdout,
+                        "diagnostic": diagnostic,
+                    },
+                )
+            except Exception:
+                execution_reported = False
+        return execution_reported and not _host_teardown_requested
 
 
 async def _prepare_command(
@@ -321,16 +418,111 @@ async def _prepare_command(
         if secret_path is not None:
             secret_path.unlink(missing_ok=True)
     subject = shlex.join([request["command"], *request["args"]])
-    output_redirect = "" if capture_stdout else " >/dev/null"
+    stdout_path, stderr_path, stdout_fifo, stderr_fifo, collectors_path = _capture_paths(token)
     scope_error = shlex.quote(f"{SCOPE_ERROR_PREFIX} {result_token}\\n")
-    inner = (
+    prefix = (
         "umask 077; "
-        f"{{ echo $$ > {shlex.quote(scope_path)}; }} 2>/dev/null || "
-        f"{{ printf {scope_error}; exit 111; }}; "
+        f"setsid_path=$(command -v setsid) || {{ printf {scope_error}; exit 111; }}; "
+        f"shell_path=$(command -v sh) || {{ printf {scope_error}; exit 111; }}; "
         f". {shlex.quote(container_path)}; command -p rm -f {shlex.quote(container_path)}; "
-        f"exec {subject}{output_redirect}"
+        f"{{ printf '%s %s\\n' \"$$\" \"$$\" > {shlex.quote(scope_path)}; }} "
+        f"2>/dev/null || {{ printf {scope_error}; exit 111; }}; "
+    )
+    if capture_stdout:
+        # Keep the Docker exec streams detached from the subject tree while a
+        # bounded collector drains each inherited FIFO for as long as a task
+        # service keeps it open. `dd bs=1` writes incrementally, then `cat`
+        # continues draining through the same open descriptor after the cap.
+        # The collector exits naturally when every inherited writer closes; it
+        # must outlive the relay while a preserved service still writes, since
+        # killing its only reader would terminate that service with SIGPIPE.
+        def collector(fifo: str, output: str, descriptor: int) -> str:
+            return (
+                f"(exec {descriptor}< {shlex.quote(fifo)}; "
+                f"command -p dd bs=1 count={RESULT_CARRIER_LIMIT_BYTES + 1} "
+                f"<&{descriptor} > {shlex.quote(output)} 2>/dev/null; "
+                f"command -p cat <&{descriptor} >/dev/null) "
+                "</dev/null >/dev/null 2>&1"
+            )
+
+        stdout_collector = collector(stdout_fifo, stdout_path, 3)
+        stderr_collector = collector(stderr_fifo, stderr_path, 4)
+        inner = (
+            prefix
+            + "rm -f -- "
+            + " ".join(
+                shlex.quote(path)
+                for path in (
+                    stdout_path,
+                    stderr_path,
+                    stdout_fifo,
+                    stderr_fifo,
+                    collectors_path,
+                )
+            )
+            + "; "
+            + f"command -p mkfifo {shlex.quote(stdout_fifo)} {shlex.quote(stderr_fifo)} || "
+            + f"{{ printf {scope_error}; exit 111; }}; "
+            + f": > {shlex.quote(stdout_path)}; : > {shlex.quote(stderr_path)}; "
+            + f"\"$setsid_path\" \"$shell_path\" -c {shlex.quote(stdout_collector)} "
+            + "</dev/null >/dev/null 2>&1 & stdout_collector=$!; "
+            + f"\"$setsid_path\" \"$shell_path\" -c {shlex.quote(stderr_collector)} "
+            + "</dev/null >/dev/null 2>&1 & stderr_collector=$!; "
+            + f"printf '%s %s\\n' \"$stdout_collector\" \"$stderr_collector\" > "
+            + f"{shlex.quote(collectors_path)}; "
+            + f"{subject} > {shlex.quote(stdout_fifo)} 2> {shlex.quote(stderr_fifo)} & leader=$!; "
+        )
+    else:
+        inner = prefix + f"{subject} >/dev/null 2>&1 & leader=$!; "
+    inner += (
+        f"{{ printf '%s %s\\n' \"$$\" \"$leader\" > {shlex.quote(scope_path)}; }} "
+        f"2>/dev/null || {{ kill \"$leader\" 2>/dev/null || true; printf {scope_error}; exit 111; }}; "
+        "wait \"$leader\"; status=$?; exit \"$status\""
     )
     return f"setsid --wait sh -c {shlex.quote(inner)}"
+
+
+def _capture_paths(token: str) -> tuple[str, str, str, str, str]:
+    prefix = f"/tmp/maka-eval-{token}"
+    return (
+        f"{prefix}.stdout",
+        f"{prefix}.stderr",
+        f"{prefix}.stdout.fifo",
+        f"{prefix}.stderr.fifo",
+        f"{prefix}.collectors",
+    )
+
+
+async def _load_captured_subject_outputs(
+    environment: Any,
+    cwd: str,
+    result: Any,
+    request: dict[str, Any],
+    token: str,
+) -> Any:
+    if not request.get("captureStdout", True):
+        return result
+    stdout_path, stderr_path, _, _, _ = _capture_paths(token)
+    stdout_marker = f"MAKA-EVAL-CAPTURE-STDERR-V1 {token}"
+    captured = await environment.exec(
+        f"cat {shlex.quote(stdout_path)} 2>/dev/null; "
+        f"printf '\\n%s\\n' {shlex.quote(stdout_marker)}; "
+        f"cat {shlex.quote(stderr_path)} 2>/dev/null",
+        cwd=cwd,
+        timeout_sec=1,
+    )
+    raw = str(getattr(captured, "stdout", "") or "")
+    separator = f"\n{stdout_marker}\n"
+    if separator in raw:
+        stdout, _, stderr = raw.partition(separator)
+    else:
+        stdout = str(getattr(result, "stdout", "") or "")
+        stderr = str(getattr(result, "stderr", "") or "")
+    return SimpleNamespace(
+        return_code=result.return_code,
+        stdout=stdout,
+        stderr=stderr,
+    )
 
 
 async def _require_constrained_subject(environment: Any) -> None:
@@ -568,7 +760,7 @@ async def _settle(environment: Any, cwd: str, scope_path: str, execution: Any) -
     else:
         result = None
         for signal, timeout in (("TERM", 20), ("KILL", 10)):
-            await _signal(environment, cwd, scope_path, signal)
+            await _signal_group(environment, cwd, scope_path, signal)
             try:
                 result = await asyncio.wait_for(asyncio.shield(execution), timeout=timeout)
                 break
@@ -582,6 +774,59 @@ async def _settle(environment: Any, cwd: str, scope_path: str, execution: Any) -
             raise RuntimeError("Maka Eval subject did not settle")
     await _quiesce_scope(environment, cwd, scope_path)
     return result
+
+
+async def _stop_subject_for_timeout(
+    environment: Any,
+    cwd: str,
+    scope_path: str,
+    execution: Any,
+    timeout: float,
+) -> Any:
+    # The verifier still scores this trial. Stop only the subject so
+    # `environment.exec` can return; do not hunt descendants or delete the
+    # environment. Those leftovers are what the task asked the subject to leave.
+    if execution.cancelled():
+        raise RuntimeError("Maka Eval subject execution was cancelled")
+    if execution.done():
+        return execution.result()
+    loop = asyncio.get_running_loop()
+    deadline = loop.time() + timeout
+    destroy_reserve = min(20.0, timeout * 0.2)
+    stop_deadline = deadline - destroy_reserve
+    remaining = stop_deadline - loop.time()
+    if remaining > 0:
+        # Escalate inside one environment call. Docker-backed environments pay
+        # a material control-plane round trip for every exec; separate TERM and
+        # KILL calls can consume a short framework budget even when KILL works.
+        # This command still targets only the session leader, so task-owned
+        # background processes remain available to the verifier.
+        await _stop_leader(
+            environment,
+            cwd,
+            scope_path,
+            grace_sec=min(0.1, max(0.001, remaining * 0.1)),
+            timeout_sec=remaining,
+        )
+        remaining = stop_deadline - loop.time()
+        if remaining > 0:
+            try:
+                return await asyncio.wait_for(
+                    asyncio.shield(execution), timeout=remaining
+                )
+            except asyncio.CancelledError:
+                if execution.cancelled():
+                    raise RuntimeError("Maka Eval subject execution was cancelled") from None
+                raise
+            except (TimeoutError, asyncio.TimeoutError):
+                pass
+    if execution.done() and not execution.cancelled():
+        return execution.result()
+    # A verifier cannot measure a stable environment while the subject may
+    # still be mutating it. Fail closed instead of publishing a scoreable
+    # framework_timeout frame without positive leader-exit evidence.
+    await _destroy_environment(environment, execution, deadline, loop)
+    raise RuntimeError("Maka Eval could not confirm subject exit after framework timeout")
 
 
 async def _settle_or_destroy(
@@ -600,24 +845,35 @@ async def _settle_or_destroy(
             timeout=max(0.001, deadline - loop.time() - stop_reserve),
         )
     except Exception:
-        try:
-            remaining = max(0.001, deadline - loop.time())
-            await asyncio.wait_for(environment.stop(delete=True), timeout=remaining)
-        except Exception:
-            pass
-        finally:
-            if not execution.done():
-                execution.cancel()
-            with contextlib.suppress(asyncio.CancelledError, Exception):
-                remaining = max(0.001, deadline - loop.time())
-                await asyncio.wait_for(execution, timeout=remaining)
+        await _destroy_environment(environment, execution, deadline, loop)
         return None
 
 
-async def _signal(environment: Any, cwd: str, scope_path: str, signal: str) -> None:
+async def _destroy_environment(
+    environment: Any,
+    execution: Any,
+    deadline: float,
+    loop: asyncio.AbstractEventLoop,
+) -> None:
+    try:
+        remaining = max(0.001, deadline - loop.time())
+        await asyncio.wait_for(environment.stop(delete=True), timeout=remaining)
+    except Exception:
+        pass
+    finally:
+        if not execution.done():
+            execution.cancel()
+        with contextlib.suppress(asyncio.CancelledError, Exception):
+            remaining = max(0.001, deadline - loop.time())
+            await asyncio.wait_for(execution, timeout=remaining)
+
+
+async def _signal_group(environment: Any, cwd: str, scope_path: str, signal: str) -> None:
     command = (
-        f"pgid=$(cat {shlex.quote(scope_path)} 2>/dev/null) || exit 0; "
-        "case $pgid in ''|0|*[!0-9]*) exit 0;; esac; "
+        f"read -r pgid leader extra < {shlex.quote(scope_path)} 2>/dev/null || exit 0; "
+        "case \"$pgid\" in ''|0|*[!0-9]*) exit 0;; esac; "
+        "case \"$leader\" in '' ) leader=$pgid;; 0|*[!0-9]*) exit 0;; esac; "
+        "test -z \"$extra\" || exit 0; "
         f"kill -{signal} -- \"-$pgid\""
     )
     with contextlib.suppress(Exception):
@@ -628,11 +884,39 @@ async def _signal(environment: Any, cwd: str, scope_path: str, signal: str) -> N
         )
 
 
+async def _stop_leader(
+    environment: Any,
+    cwd: str,
+    scope_path: str,
+    grace_sec: float,
+    timeout_sec: float = 5,
+) -> bool:
+    command = (
+        f"read -r pgid leader extra < {shlex.quote(scope_path)} 2>/dev/null || exit 1; "
+        "case \"$pgid\" in ''|0|*[!0-9]*) exit 1;; esac; "
+        "case \"$leader\" in '' ) leader=$pgid;; 0|*[!0-9]*) exit 1;; esac; "
+        "test -z \"$extra\" || exit 1; "
+        "kill -TERM -- \"$leader\" 2>/dev/null || exit 1; "
+        f"sleep {grace_sec:.6f}; "
+        "kill -0 -- \"$leader\" 2>/dev/null || exit 0; "
+        "kill -KILL -- \"$leader\" 2>/dev/null"
+    )
+    try:
+        result = await environment.exec(
+            command,
+            cwd=cwd,
+            timeout_sec=max(0.001, timeout_sec),
+        )
+        return result.return_code == 0
+    except Exception:
+        return False
+
+
 async def _quiesce_scope(environment: Any, cwd: str, scope_path: str) -> None:
     if not await _scope_active(environment, cwd, scope_path):
         return
     for signal, timeout in (("TERM", 10), ("KILL", 5)):
-        await _signal(environment, cwd, scope_path, signal)
+        await _signal_group(environment, cwd, scope_path, signal)
         deadline = asyncio.get_running_loop().time() + timeout
         while asyncio.get_running_loop().time() < deadline:
             if not await _scope_active(environment, cwd, scope_path):
@@ -643,8 +927,10 @@ async def _quiesce_scope(environment: Any, cwd: str, scope_path: str) -> None:
 
 async def _scope_active(environment: Any, cwd: str, scope_path: str) -> bool:
     result = await environment.exec(
-        f"pgid=$(cat {shlex.quote(scope_path)} 2>/dev/null) || exit 4; "
-        "case $pgid in ''|0|*[!0-9]*) exit 4;; esac; "
+        f"read -r pgid leader extra < {shlex.quote(scope_path)} 2>/dev/null || exit 4; "
+        "case \"$pgid\" in ''|0|*[!0-9]*) exit 4;; esac; "
+        "case \"$leader\" in '' ) leader=$pgid;; 0|*[!0-9]*) exit 4;; esac; "
+        "test -z \"$extra\" || exit 4; "
         "kill -0 -- \"-$pgid\" 2>/dev/null; status=$?; "
         "if [ $status -eq 0 ]; then exit 0; fi; exit 3",
         cwd=cwd,
--- a/packages/eval/harbor/run_trial.py
+++ b/packages/eval/harbor/run_trial.py
@@ -24,11 +24,12 @@ import hashlib
 import importlib
 import importlib.metadata
 import json
+import math
 import os
 import signal
 import sys
 from pathlib import Path
-from typing import AsyncIterator
+from typing import Any, AsyncIterator
 
 FRAMEWORK_VERSION_MISMATCH_EXIT_CODE = 78
 
@@ -145,6 +146,7 @@ async def run_trial(framework: str, expected_version: str, config_file: Path) ->
                     publish_ready_task(namespace, target)
         else:
             trial = await trial_type.create(config)
+        bind_framework_timeout_budget(framework, trial)
         await trial.run()
     finally:
         config_file.unlink(missing_ok=True)
@@ -160,6 +162,87 @@ def apply_subject_egress_policy(task: object) -> None:
     agent = task.config.agent
     agent.network_mode = "allowlist"
     agent.allowed_hosts = [allowed_host]
+
+
+def resolved_framework_timeout_ms(config: object, default_timeout: float | None) -> int | None:
+    base_timeout = config.agent.override_timeout_sec or default_timeout
+    if base_timeout is None:
+        return None
+    maximum = config.agent.max_timeout_sec or float("inf")
+    multiplier = (
+        config.agent_timeout_multiplier
+        if config.agent_timeout_multiplier is not None
+        else config.timeout_multiplier
+    )
+    timeout_ms = math.floor(min(base_timeout, maximum) * multiplier * 1000)
+    if timeout_ms <= 0:
+        raise RuntimeError("framework agent timeout is shorter than one millisecond")
+    return timeout_ms
+
+
+def bind_framework_timeout_budget(framework: str, trial: object) -> None:
+    """Set the RelayAgent cleanup budget from each framework's effective phase timeout."""
+
+    if framework == "harbor":
+        original = getattr(trial, "_run_agent_phase", None)
+        agent = getattr(trial, "agent", None)
+        if not callable(original) or not hasattr(agent, "set_framework_timeout_ms"):
+            raise RuntimeError("Harbor timeout integration is unavailable")
+
+        async def run_agent_phase(*args: Any, **kwargs: Any) -> Any:
+            timeout_sec = kwargs.get("timeout_sec")
+            agent.set_framework_timeout_ms(
+                None if timeout_sec is None else math.floor(timeout_sec * 1000)
+            )
+            return await original(*args, **kwargs)
+
+        trial._run_agent_phase = run_agent_phase
+        return
+
+    if framework == "pier":
+        agent = getattr(trial, "_agent", None)
+        execution = getattr(trial, "_execution", None)
+        run_agent = getattr(trial, "_execute_agent", None)
+        run_step_agent = getattr(trial, "_execute_step_agent", None)
+        resolve_step_timeout = getattr(trial, "_resolve_step_timeout", None)
+        if (
+            not hasattr(agent, "set_framework_timeout_ms")
+            or execution is None
+            or not callable(run_agent)
+            or not callable(run_step_agent)
+            or not callable(resolve_step_timeout)
+        ):
+            raise RuntimeError("Pier timeout integration is unavailable")
+
+        async def execute_agent() -> Any:
+            timeout_sec = execution.agent_timeout_sec
+            agent.set_framework_timeout_ms(
+                None if timeout_sec is None else math.floor(timeout_sec * 1000)
+            )
+            return await run_agent()
+
+        async def execute_step_agent(step: object, result: object) -> Any:
+            default_timeout = (
+                step.agent.timeout_sec
+                if step.agent.timeout_sec is not None
+                else trial._task.config.agent.timeout_sec
+            )
+            timeout_sec = resolve_step_timeout(
+                override=trial.config.agent.override_timeout_sec,
+                default=default_timeout,
+                max_val=trial.config.agent.max_timeout_sec,
+                specific_multiplier=trial.config.agent_timeout_multiplier,
+            )
+            agent.set_framework_timeout_ms(
+                None if timeout_sec is None else math.floor(timeout_sec * 1000)
+            )
+            return await run_step_agent(step, result)
+
+        trial._execute_agent = execute_agent
+        trial._execute_step_agent = execute_step_agent
+        return
+
+    raise RuntimeError("framework must be harbor or pier")
 
 
 async def create_harbor_trial(trial_type: type, config: object) -> object:
--- a/packages/eval/harbor/test_relay_lifecycle.py
+++ b/packages/eval/harbor/test_relay_lifecycle.py
@@ -24,6 +24,7 @@ import json
 import os
 import signal
 import shutil
+import socket
 import subprocess
 import sys
 import tempfile
@@ -58,9 +59,15 @@ class LocalEnvironment:
             shell=True,
             executable="/bin/bash",
             check=False,
+            capture_output=True,
+            text=True,
             timeout=timeout_sec,
         )
-        return SimpleNamespace(return_code=completed.returncode)
+        return SimpleNamespace(
+            return_code=completed.returncode,
+            stdout=completed.stdout,
+            stderr=completed.stderr,
+        )
 
 
 class SimultaneousEnvironment:
@@ -94,9 +101,20 @@ class SimultaneousEnvironment:
             )
             self.finished.set()
             return SimpleNamespace(return_code=0, stdout=frame, stderr=None)
-        if command.startswith("test -r") or command.startswith("pgid="):
+        if command.startswith("test -r") or command.startswith("read -r pgid"):
             return SimpleNamespace(return_code=3, stdout="", stderr="")
         return SimpleNamespace(return_code=0, stdout="", stderr="")
+
+
+class DummyWriter:
+    def is_closing(self):
+        return False
+
+    def write(self, _value):
+        return None
+
+    async def drain(self):
+        return None
 
 
 class ClosedWriter:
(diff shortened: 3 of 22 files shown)


packages/eval/harbor/relay_agent.py
```diff
@@ -472,6 +479,78 @@ async def _settle(environment: Any, cwd: str, scope_path: str, execution: Any) -
...
+            try:
+                return await asyncio.wait_for(
+                    asyncio.shield(execution), timeout=min(0.1, remaining)
+                )
+            except asyncio.CancelledError:
+                if execution.cancelled():
+                    raise RuntimeError("Maka Eval subject execution was cancelled") from None
+                raise
+            except (TimeoutError, asyncio.TimeoutError):
+                break
```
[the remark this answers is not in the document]
1625567290: Addressed in 3aca1fad0.

The vanished-leader grace now waits for the remaining stop deadline, not 100 ms. Harbor exec completion can take longer than that, and a finished subject must still be admitted rather than destroyed.

packages/eval/harbor/test_relay_lifecycle.py
```diff
@@ -177,9 +200,17 @@ async def exec(self, command, cwd=None, timeout_sec=None):
 
 
 def _is_teardown(command: str) -> bool:
-    # `kill -0` is how the relay asks whether the scope is still there; only
-    # TERM and KILL end it.
-    return "kill -TERM" in command or "kill -KILL" in command
+    # Host abort signals the recorded process group. Framework timeout stops
+    # only the subject leader (`"$pgid"`), which must not count as teardown.
+    return ('-"$pgid"' in command) and ("kill -TERM" in command or "kill -KILL" in command)
```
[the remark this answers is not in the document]
1625567290: Addressed in 3aca1fad0.

The predicates now match the group-signal form the relay emits (`"-$pgid"`). A focused test pins group TERM as teardown and leader TERM as leader-stop. The unconfirmed-timeout fixture comment now describes the vanished-leader path; a second fixture delivers TERM/KILL and still fail-closes if the subject does not exit.

packages/eval/README.md
```diff
@@ -142,11 +142,13 @@ What the verifier scores is the environment the task was left in, so a subject t
...
-scored. The same rule binds the agent frameworks: the DeepSeek Harness owns a persistent PTY tree and
-kills every descendant on shutdown, so the Eval-patched DSH subprocess skips that kill under
-`DSH_PRESERVE_BACKGROUND_PROCESSES`, which Eval always sets.
+subjects Eval classifies as failed and not for the others. Framework timeout is the same
+measurement: the verifier still runs, the status is `subject_failed`, and leftover processes stay.
+The relay stops only the subject so the execution call can return; it does not signal the process
+group or delete the environment. Host abort is the exception, because that trial is abandoned
+rather than scored. The same rule binds the agent frameworks: the DeepSeek Harness owns a persistent
+PTY tree and kills every descendant on shutdown, so the Eval-patched DSH subprocess skips that kill
+under `DSH_PRESERVE_BACKGROUND_PROCESSES`, which Eval always sets.
```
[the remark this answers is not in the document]
1625567290: Addressed in 3aca1fad0.

The README now states the fail-closed case: if the leader never acknowledges TERM or KILL inside the reserved budget, the relay destroys the environment and raises instead of publishing a scoreable framework_timeout.

packages/runtime-host/src/client/hosted-execution.ts
```diff
@@ -120,21 +131,57 @@ export async function runHostedExecutionWithDependencies(
...
+  const detach = () => {
+    if (abortPolicy !== 'preserve_environment' || detached) return;
+    detached = true;
+    host.releaseToEnvironment();
+    closeForDetach = connection.close().catch(() => undefined);
+  };
   const cancel = () => {
+    if (abortPolicy === 'preserve_environment') {
+      detach();
+      return;
```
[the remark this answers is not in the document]
1625567290: Addressed in ecffafffd.

preserve_environment now detaches only after hosted.execution.start has been dispatched. An abort before that admission returns a cancelled projection and settles the unused Host. It does not report that execution continues.

packages/eval/harbor/relay_agent.py
```diff
@@ -516,11 +595,28 @@ async def _signal(environment: Any, cwd: str, scope_path: str, signal: str) -> N
...
+        f"pgid=$(cat {shlex.quote(scope_path)} 2>/dev/null) || exit 1; "
+        "case $pgid in ''|0|*[!0-9]*) exit 1;; esac; "
+        f"kill -{signal} -- \"$pgid\""
+    )
+    try:
+        result = await environment.exec(
+            command,
+            cwd=cwd,
+            timeout_sec=5,
+        )
```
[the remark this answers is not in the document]
1625567290: Addressed in ecffafffd.

Each leader-stop exec is now capped by the remaining stop deadline, and that remaining value is recomputed after the signal before waiting on environment.exec. A 50 ms teardown budget can no longer spend five seconds on the kill command.

packages/eval/harbor/relay_agent.py
```diff
@@ -472,6 +479,78 @@ async def _settle(environment: Any, cwd: str, scope_path: str, execution: Any) -
...
+        return result
+    if execution.done() and not execution.cancelled():
+        return execution.result()
+    # A verifier cannot measure a stable environment while the subject may
+    # still be mutating it. Fail closed instead of publishing a scoreable
+    # framework_timeout frame without positive leader-exit evidence.
+    try:
+        remaining = max(0.001, deadline - loop.time())
+        await asyncio.wait_for(environment.stop(delete=True), timeout=remaining)
+    except Exception:
```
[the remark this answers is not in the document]
1625567290: This destroy is the fail-closed path, not the confirmed-timeout path.

When the leader has exited, the relay leaves descendants and the environment for the verifier. When TERM/KILL never produce that exit, there is no stable leftover environment to score: the subject may still be mutating the cell. Publishing framework_timeout in that state would let the verifier race a live subject. Deleting the environment and raising keeps that trial an infrastructure failure instead of a scored timeout.

The README now names this as the second exception next to host abort.

Astro-Han:
Thanks for the fix — the inconsistency is real (on timeout the old `run()` cancelled the recorded process group and could delete the environment, so leftover services vanished on timeout but survived clean exit), and the solution is sound: `_host_teardown_requested` distinguishes host abort (still settle/destroy) from framework timeout (stop only the leader), the budget math in `_stop_subject_for_timeout` is correct (TERM ≤20s / KILL ≤10s slices bounded by `remaining`, destroy reserve `min(20, 0.2*timeout)` — 30s budget math checks out), `_signal_leader`'s `kill -- "$pgid"` is correct bash semantics (pgid file = setsid'd sh PID = group leader, exec'd into the subject), and the fail-closed paths (leader ignoring TERM/KILL, vanished leader with unconfirmed exec) destroy the environment and stop publishing scoreable frames — all under test (`test_unconfirmed_…`, `test_ignored_…`). The architecture boundary holds: only the Runtime Host *client* (`abortPolicy` option, default `'cancel'` unchanged) and the eval relay changed; `@maka/eval` doesn't construct Runtime or touch host implementation. I ran the relay tests on the PR head (20/20 + harbor lifecycle 20, 3 skipped) and verified red→green by swapping main's `relay_agent.py` back in — 4 new timeout tests fail on old code, so the coverage is genuine. The live-process test (`test_framework_timeout_preserves_a_live_background_service`) really keeps a connection alive through leader TERM.

**Conclusion: PASS with two P2s (Windows lane infra-flake + end-to-end validation gap) — no P0/P1.**

**P2-1 — the other side of this contract lives outside the repo: end-to-end behavior is unverified against the real framework.** The fix keeps leftovers alive only if the framework's own timeout mechanism doesn't independently cancel `environment.exec`'s recorded process group or destroy the environment. The framework is a pip dependency installed by `run_trial.py:98-100` and isn't in the diff, so this can't be confirmed or refuted from the repo. The live test uses `LocalEnvironment` (local process exec), not real docker exec cancel semantics. Please either run a real Harbor cell (start background service → trigger framework timeout → verify the service reachable in the validate phase) or explicitly state this premise and the verification plan in the PR description.

**P2-2 — the failed Windows lane is an environment flake, not this PR.** The failure ("Validate installed CLI Windows x64 / Node 24" → "checking the interactive TUI setup path" → `INTERNAL_STARTUP_FAILURE`) matches byte-for-byte the same step on an unrelated dependabot PR (32105259238), the PR's own earlier commits passed the same lane, and the final commit only touches eval relay Python (not executed by the Windows smoke) and the hosted-execution client (not in the TUI startup path). Please re-run that lane (or record it as a known infra issue) rather than attributing it to this PR.

P3 (optional): the `elif not execution_reported and not _host_teardown_requested:` branch in `relay_agent.py:201-209` is now dead code (`_stop_subject_for_timeout` can't return None — it either returns or raises), so the result-frame-missing fallback frame can no longer be produced on the timeout path; `await closeForDetach` at `hosted-execution.ts:184` has no timeout (a hung close means the relay budget exhausts → destroy + fail-closed, safe but unbounded); `_signal_leader` treats any exec exception (including transient) as "leader vanished" — if the subject then doesn't finish within `remaining`, the environment is destroyed and the trial unscoreable (documented fail-closed, low probability); README has two paragraphs squeezed onto one line, and `_stop_subject_for_timeout` is annotated `-> Any | None` but never returns None.

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagent on `opencode-go/deepseek-v4-flash`). The subagent ran the relay tests on both the PR head and the old implementation (red→green proof), traced the budget math, and pulled and compared the Windows failure log against the unrelated dependabot run; P2-1 is an unverifiable-from-repo gap the PR itself implies, not an observed failure. Please weigh these findings with your own judgment.*

<details>

结论：PASS（2 个 P2：Windows lane 环境 flake + 端到端验证缺口；无 P0/P1）。问题真实（timeout 时旧 run() 无条件 cancel recorded process group 且可删环境→leftover 在 timeout 时消失、clean exit 时存活），方案正确：_host_teardown_requested 区分 host abort（仍 settle/destroy）与 framework timeout（只停 leader）；_stop_subject_for_timeout 预算 math 正确（TERM≤20s/KILL≤10s 各 slice 被 remaining 约束、destroy 保留 min(20,0.2*timeout)，30s 预算推算成立）；_signal_leader 的 kill -- "$pgid" 在 bash 语义下正确（pgid 文件=setsid'd sh PID=组 leader，exec 后即 subject）；fail-closed 路径（leader 无视 TERM/KILL、vanished leader 且 exec 未确认）销毁环境并停止发布可计分 frame，均有测试。架构边界保持：只改 Runtime Host client（abortPolicy 新选项，默认 'cancel' 不变）与 eval relay；@maka/eval 未构造 Runtime、未碰 host 实现。实跑 relay 测试（20/20 + harbor lifecycle 20（3 skipped））并把 main 的 relay_agent.py 换回后 4 个新 timeout 测试全失败——red→green 证明测试真实；实况测试用真实进程验证 background service 在 leader TERM 后仍可 TCP 连接。P2-1：修复的另一侧在仓库外——framework 自身 timeout 机制若独立 cancel recorded process group 或销毁环境，则 end-to-end 不成立；框架是 pip 依赖（run_trial.py:98-100 安装）不在 diff 中，仓库内无法证实/证伪；实况测试用 LocalEnvironment 不覆盖真实 docker exec 的 cancel 语义。建议用真实 Harbor cell 验证（启动 background service→触发 framework timeout→验证阶段服务可达）或至少在描述中声明前提与验证方式。P2-2：Windows lane 失败是环境 flake 非本 PR 引入——同一错误签名在无关 dependabot PR（32105259238）同步骤逐字复现、本 PR 更早 commits 同 lane 通过、最后 commit 只改 eval relay Python（Windows smoke 不执行）与 hosted-execution client（不在 TUI 启动路径）。建议重跑该 lane 或按已知 infra 问题延后。P3（可选）：relay_agent.py:201-209 的 elif 分支已成死代码（_stop_subject_for_timeout 不可能返回 None，结果帧缺失兜底 frame 在 timeout 路径已无法产生）；hosted-execution.ts:184 closeForDetach 无超时（挂起时 relay 预算耗尽→destroy+fail-closed，安全但无界）；_signal_leader 对待瞬态 exec 异常一律当"leader 已消失"，若 subject 未在 remaining 内完成则环境被毁、trial 不可计分（文档化 fail-closed，概率低）；README 两段挤一行、_stop_subject_for_timeout 标注 -> Any|None 但从不返回 None。
</details>

Astro-Han (commented):
Separating framework timeout from host abort is the correct ownership model: the verifier may need subject leftovers, while an abandoned trial still requires destructive cleanup. The current implementation loses that distinction at two race boundaries—before Hosted Execution admission is known, and when timeout cleanup itself is cancelled or fails.

The first-principles solution is an explicit server-owned admission token plus one shielded, bounded cleanup state machine. Preserve/detach is allowed only after admission is acknowledged; a second cancellation or execution error transitions that same cleanup state machine to environment destruction before cancellation is re-raised. This is simpler than inferring ownership from a client-side boolean and nested exception handlers.

Review performed with two Codex reviewer agents and DeepSeek V4 Flash as advisory tools; I verified both failure paths against the latest head and current `main`.

区分 framework timeout 与 host abort 是正确的 ownership 模型：verifier 可能需要 subject leftovers，而被放弃的 trial 仍必须执行破坏性清理。当前实现会在两个 race 边界丢失这一区分：Hosted Execution admission 尚未确认时，以及 timeout cleanup 自身再次被取消或失败时。

更符合第一性原理的方案是显式的 server-owned admission token，加上一套 shielded、bounded cleanup state machine。只有 admission ack 后才允许 preserve/detach；第二次 cancellation 或 execution error 必须让同一 cleanup state machine 转入 environment destruction，完成后再重新抛出 cancellation。这比依赖客户端 boolean 和嵌套 exception handler 推断 ownership 更简单可靠。

本次审查使用了两位 Codex reviewer agents 与 DeepSeek V4 Flash 作为辅助工具；我已依据最新 head 和当前 `main` 复核两个失败路径。

</details>

packages/runtime-host/src/client/hosted-execution.ts
```diff
@@ -120,21 +131,69 @@ export async function runHostedExecutionWithDependencies(
...
+      return {
+        projection: indeterminate(execution.executionId, 'Hosted execution was cancelled'),
+        detached: false,
+      };
+    }
+    cancel();
+  }
   try {
-    return await connection.request('hosted.execution.start', execution);
+    dispatched = true;
```
Astro-Han: **P1 — `dispatched` is set before the server has admitted the execution.** An abort can arrive while `connection.request()` is still queued, being written, or not yet accepted by Runtime Host; `detach()` then unreferences the Owned Host and reports preservation even though execution may never start. Use a server admission acknowledgement/token: before ack, abort must cancel/settle the Host; only an admitted execution may be released to the environment. Add a delayed-frame/pre-admission abort test.

packages/eval/harbor/relay_agent.py
```diff
@@ -171,14 +172,20 @@ async def run(self, instruction: str, environment: Any, context: Any) -> None:
...
+                # because the verifier still runs.
                 if terminal_projection is not None:
                     result = terminal_result
-                else:
+                elif _host_teardown_requested:
                     result = await _settle_or_destroy(
                         environment, cwd, scope_path, execution, self._teardown_timeout
                     )
+                else:
+                    result = await _stop_subject_for_timeout(
```
Astro-Han: **P1 — A second cancellation or execution error inside this timeout cleanup bypasses environment destruction.** This call runs inside the outer `except asyncio.CancelledError`; if `_stop_subject_for_timeout()` re-raises a second `CancelledError` (lines 522/534) or propagates an `environment.exec()`/execution exception, sibling `except BaseException` handlers do not catch exceptions raised from this handler. `finally` removes files and closes the relay but never calls `environment.stop(delete=True)`, leaking the container/descendants on host abort. Run cleanup in a bounded shielded task that switches to `_settle_or_destroy()` on re-cancellation/error, then re-raise.

packages/eval/harbor/relay_agent.py
```diff
@@ -472,6 +479,75 @@ async def _settle(environment: Any, cwd: str, scope_path: str, execution: Any) -
...
+    destroy_reserve = min(20.0, timeout * 0.2)
+    stop_deadline = deadline - destroy_reserve
+    result = None
+    for signal, slice_timeout in (("TERM", 20.0), ("KILL", 10.0)):
+        remaining = stop_deadline - loop.time()
+        if remaining <= 0:
+            break
+        signalled = await _signal_leader(
+            environment, cwd, scope_path, signal, timeout_sec=min(5.0, remaining)
+        )
```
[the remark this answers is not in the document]
1625567290: Addressed in cdcf81860.

`_signal_leader` now uses the remaining stop deadline directly. There is no five-second cap that can turn a slow Harbor exec into a vanished-leader miss.

packages/runtime-host/src/client/hosted-execution.ts
1625567290: Addressed in cdcf81860.

Preserve/detach now happens only when the interrupted start reports dispatch `dispatched`. A queued, not-yet-admitted start settles as cancelled and does not release the Host. The pre-admission abort test covers a delayed start frame that closes as `not_dispatched`.

packages/eval/harbor/relay_agent.py
1625567290: Addressed in cdcf81860.

The cancel handler uncancels the task so cleanup can run. A second CancelledError during that cleanup switches to `_settle_or_destroy` before the cancellation is re-raised. The environment is not left alive when timeout cleanup itself is cancelled.

packages/eval/harbor/test_relay_lifecycle.py
```diff
@@ -142,6 +144,57 @@ async def stop(self, delete=False):
...
+        if _is_leader_stop(command):
+            return SimpleNamespace(return_code=0, stdout="", stderr="")
+        return await super().exec(command, cwd=cwd, timeout_sec=timeout_sec)
+
+
+class SlowLeaderStopEnvironment(IgnoringLeaderEnvironment):
+    async def exec(self, command, cwd=None, timeout_sec=None):
+        if _is_leader_stop(command):
+            await asyncio.sleep(0.05)
+        return await super().exec(command, cwd=cwd, timeout_sec=timeout_sec)
```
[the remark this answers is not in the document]
1625567290: Addressed in eed0e211b. SlowLeaderStopEnvironment now signals leader_stop_started before it sleeps; the test waits on that event before the second cancel.

packages/runtime-host/src/client/hosted-execution.ts
```diff
@@ -120,24 +131,73 @@ export async function runHostedExecutionWithDependencies(
...
+              execution.executionId,
+              'Hosted execution continues for environment verification',
+            ),
+        detached: true,
+      };
+    }
+    return { projection, detached: false };
+  } catch (error) {
+    if (signal?.aborted && abortPolicy === 'preserve_environment') {
+      if (isAdmittedInterrupt(error)) {
```
Astro-Han: **P1 — A dispatched frame is not a server admission acknowledgement.** The connection marks a request in-flight around frame write; if the connection closes before Runtime Host processes it, the interruption is still classified as `dispatched`. This branch then detaches and reports that execution continues even though no Host execution may exist to settle. Preserve the environment only after a server-owned admission acknowledgement/token, and test frame-written-but-not-admitted interruption.

packages/eval/harbor/relay_agent.py
```diff
@@ -253,6 +225,76 @@ async def run(self, instruction: str, environment: Any, context: Any) -> None:
...
+        scope_path: str,
+        execution: asyncio.Task[Any],
+        request: dict[str, Any],
+        writer: Any,
+        execution_reported: bool,
+    ) -> None:
+        execution_terminal = execution.done() and not execution.cancelled()
+        terminal_projection = None
+        if execution_terminal:
+            terminal_result = execution.result()
```
Astro-Han: **P1 — Cleanup exceptions can still bypass environment destruction.** This call can raise when the execution task completed exceptionally, and `_persist_subject_outputs` below can also raise. Because both occur inside the `except asyncio.CancelledError` handler, the later sibling `except BaseException` does not catch them; `finally` closes files/transports but never calls settle-or-destroy, leaving the environment and descendants behind. Funnel every cleanup exception through one shielded, deadline-bounded settle-or-destroy path, then rethrow. Add execution-exception and persistence-failure regressions during host abort.

Astro-Han (commented):
The latest commits improve the timeout boundary: leader shutdown now uses the remaining deadline, unconfirmed shutdown fails closed, and the second-cancel test now enters the intended cleanup window. Two authority gaps remain.

The simplest correct model is a server-owned admission token followed by one shielded, deadline-bounded cleanup state machine. Transport dispatch is not admission, and every exception raised while finalizing cancellation must converge on settle-or-destroy before being rethrown. This removes the current sibling-exception special cases.

Reviewed with Codex using two independent review passes and DeepSeek V4 Flash as an external adversarial pass; I verified the cited state transitions against this exact head and current `main`.

最新提交改进了 timeout 边界：leader stop 使用剩余 deadline，无法确认退出时 fail-closed，二次 cancel 测试也进入了正确窗口。但仍有两个权威缺口。

最小正确模型是：由 server 返回 admission token；随后所有取消收尾都进入一个 shielded、带 deadline 的状态机。transport dispatch 不能代表 admission，收尾中的任何异常都必须先 settle-or-destroy 再重抛。

本次由 Codex 两轮独立审查，并使用 DeepSeek V4 Flash 做外部对抗审查；我核对了当前 head 与最新 `main`。

</details>

packages/runtime-host/src/client/hosted-execution.ts
1625567290: Addressed in 9281bc99b.

`dispatch === 'dispatched'` is only the client write. preserve_environment now calls `hosted.execution.admit` first and detaches only after the Host returns an `admissionToken`. A frame-written interrupt on admit, with no token, settles the Host and reports cancellation.

Compatibility epoch is 26 so mixed-version peers fail at handshake instead of sending an unknown admit command.

packages/eval/harbor/relay_agent.py
1625567290: Addressed in 9281bc99b.

`_cleanup_cancelled_execution` wraps finalize. Any exception — including `execution.result()` after an exceptional task and `_persist_subject_outputs` — runs `_settle_or_destroy` before the error is rethrown. Tests cover execution-exception and persist-failure during host abort.

packages/eval/harbor/relay_agent.py
```diff
@@ -255,6 +219,106 @@ async def run(self, instruction: str, environment: Any, context: Any) -> None:
...
+                    {
+                        "token": self._token,
+                        "kind": "executed",
+                        "termination": "exited" if execution_terminal else "framework_timeout",
+                        "exitCode": result.return_code if execution_terminal else 124,
+                        "stdout": stdout,
+                        "diagnostic": diagnostic,
+                    },
+                )
+        elif not execution_reported and not _host_teardown_requested:
```
Astro-Han: [P3] This fallback is unreachable under the current contract. On the framework-timeout path, `_stop_subject_for_timeout()` either returns a result or raises; the only branch that can leave `result` as `None` is Host teardown, where `_host_teardown_requested` is true and this condition is false. Please delete this `elif` block and narrow `_stop_subject_for_timeout` from `Any | None` to `Any`; the ordinary executed-frame branch already owns every successful timeout result.

Astro-Han (dismissed):
The current head closes the two earlier authority gaps. Framework timeout and Host abort now enter distinct relay paths; cancellation cleanup converges on bounded settle-or-destroy; and preserve/detach is permitted only after a server-owned admission acknowledgement. If that acknowledgement is lost, the client deliberately fails closed and settles the owned Host rather than claiming preservation it cannot prove.

The solution keeps Runtime Host as the execution authority and uses one cleanup state machine instead of inferring ownership from transport dispatch. I found no blocking correctness issue. The inline P3 removes an unreachable fallback left behind by the stronger fail-closed contract.

CI note: all completed checks are green; `test_workspaces` is still in progress, so merge readiness still depends on that final check.

AI-assisted review disclosure: Codex performed the final review using two independent reviewer passes and OpenCode Go DeepSeek V4 Flash (high effort) as an adversarial advisory pass. I verified admission/abort state transitions, deadline handling, exact head `9281bc99b`, and live CI; no local tests were run.

当前 head 已关闭此前两个 authority 缺口：framework timeout 与 Host abort 进入不同 relay 路径；取消清理统一收敛到有界的 settle-or-destroy；只有收到 server-owned admission ack 后才允许 preserve/detach。如果 ack 丢失，client 会刻意 fail closed 并 settle owned Host，而不会声称一个无法证明的 preservation。

方案保持 Runtime Host 为 execution authority，并使用单一 cleanup state machine，不再从 transport dispatch 推断 ownership。未发现阻塞性正确性问题。行内 P3 只删除强化 fail-closed 契约后留下的不可达 fallback。

CI 说明：所有已完成 checks 均为绿色；`test_workspaces` 仍在运行，因此是否可合并还取决于这个最后检查。

AI 辅助审查说明：Codex 使用两轮独立 reviewer 审查，并以 OpenCode Go DeepSeek V4 Flash（high effort）进行对抗性辅助审查；我已核对 admission/abort 状态转换、deadline、精确 head `9281bc99b` 与实时 CI。本轮未运行本地测试。

</details>

packages/eval/harbor/relay_agent.py
1625567290: Addressed in the current head (`210eb5cbd`). The unreachable `elif not execution_reported...` fallback has been removed, and `_stop_subject_for_timeout()` now returns `Any` rather than `Any | None`. The framework-timeout path therefore either produces a concrete execution result or fails closed, while Host teardown remains the only nullable settle path. Verified with the relay lifecycle suite (23 passed, 3 skipped) and the Eval lifecycle boundary suite (19 passed).

1625567290:
@Astro-Han Rebased onto current `main` (`6ada6b546`) and force-pushed `774fb641c`.

The protocol conflict is resolved by retaining the epochs already on `main` and assigning hosted execution admission epoch 40. ASF headers are present, and the old README changes are no longer part of this PR.

Focused local validation passed: Runtime Host hosted-execution/protocol 58/58, Eval lifecycle 19/19, relay lifecycle 23 tests with 3 conditional skips, relay contract 30/30, and CLI operator command 4/4. Lint, format, ASF header, and diff checks also pass. Fresh CI is pending and may need workflow approval.

Astro-Han:
Hi — this PR was behind `main` and could not be merged as-is. Since `maintainer_can_modify` was enabled, I rebased it onto current `main` and force-pushed the result; the rebase applied with **zero conflicts**, so your changes are unchanged in content (author attribution is preserved — only the commit SHAs moved).

Please `git fetch && git reset --hard <your-remote>/<branch>` before your next push so you do not re-introduce the old base.

CI should now run against the rebased head. I will pick this up for review once the checks are terminal.

---
*AI-assisted maintenance action. It is not a review and does not count as the required human review under `CONTRIBUTING.md` §Review.*

1625567290:
Thanks for rebasing it. I confirm the PR is now at `ef88e934e5597aafc6990fd7869a444d24341ee2`; I will not push the previous local head back onto this branch.

The fresh exact-head CI run exists as `32624459195`, but GitHub currently marks it `action_required`. Please approve it when convenient so validation can run on the maintainer-rebased head.

1625567290:
@Astro-Han The maintainer-rebased head `ef88e934e5597aafc6990fd7869a444d24341ee2` now has terminal green CI: run `32624459195` passed in 6m17s. No branch changes were made after your rebase. Ready for review.

1625567290:
@Astro-Han `main` advanced again after the maintainer rebase, so I rebased the exact remote head onto current `main` (`a2f2a1afa`). The new head is `197d6eb02`, and GitHub reports it mergeable.

The protocol conflict was resolved at epoch 42: current `main` owns epoch 40 for queue entry mutation and epoch 41 for typed compaction outcomes; this PR's hosted-execution admission operation now owns epoch 42.

Exact-head local validation:
- relevant Runtime Host tests: 59/59
- Eval lifecycle tests: 19/19
- CLI operator-command tests: 4/4
- relay lifecycle tests: 23 run, 3 conditional skips
- ASF header check
- `git diff --check`

Ready for the fresh exact-head CI and review.

The fresh workflow is waiting for maintainer approval: [CI 32638150680](https://github.com/apache/maka/actions/runs/32638150680). Please approve this exact-head run.

Astro-Han (commented):
Incremental re-review at exact head `197d6eb0201454097144e2cb341642e085cbecb0`. Exact-head run `32638150680` (attempt 2, job `97193042407`) is terminal `SUCCESS`, with Build, Typecheck, affected workspace tests, and Runtime Host tests all genuinely executed rather than skipped.

The previously-raised items are closed: timeout versus Host-abort routing, cancel/error destroy, and vanished-leader fail-closed all hold at this head. The protocol epoch move to 42 came in with the rebase and isn't counted as a finding here.

Two findings remain, one of them blocking.

### `[P1]` `admit` has no server-owned at-most-once identity, so a fast-settling execution runs twice — `packages/runtime-host/src/server/hosted-execution-coordinator.ts:104-150`

The admission token is minted and returned, but never required or consumed. `hosted.execution.start` decodes the *same* `HostedExecutionStartInput` as `admit` (`protocol/hosted-execution.ts:104-115`), which carries no token, and the client stores the token locally (`client/hosted-execution.ts:177-181`) without sending it back. Grepping the coordinator, `admissionToken` only ever appears where it is created or handed out — there is no comparison anywhere.

The identity record is also short-lived. `#admit` and `#start` both call `#ensureExecution`, which starts the run immediately, and the record removes itself on settle:

```js
.finally(() => {
  this.#executions.delete(input.executionId);
});
```

So if the admitted run settles before the client's `start` arrives, `#ensureExecution` finds nothing in the map and **starts a second run under the same `executionId`**. Calling `start` with no prior `admit` runs too, for the same reason.

Reproduced against this head's emitted dist with a minimal probe: `admit` → `setImmediate` → `start` yields `runs=2` with the start projection reporting `inputTokens=2`, while a direct `start` yields `ok` with `runs=1`.

In the real runner the second attempt will most likely go indeterminate after colliding on the same `session.create`, which means the first run's canonical completed/failed projection is lost — and it can wrongly settle or reclaim an environment that was supposed to be handed to a verifier. That is the opposite of the at-most-once property this change exists to establish.

The existing coordinator test at `:46-69` keeps the first task deliberately pending via a deferred, which is exactly why the fast-settle window isn't covered.

**Fix direction:** have the server retain and consume the admitted record — a terminal-result cache keyed by `executionId`, with token validation as one part of it — and cover no-admission, wrong-token, cross-connection/epoch reuse, and fast settle.

### `[P2]` The real-Harbor E2E gap is still open

`test_relay_lifecycle.py:43-64, 1033-1095` runs against a local `LocalEnvironment`/subprocess, and `lifecycle-boundaries.test.ts:585-650` uses a fake Host client and executor. Nothing exercises a real `DockerEnvironment`/Harbor trial to show that after a framework timeout the leader ends while the background service survives and stays visible to the verifier, and that a Host abort destroys it instead. Same finding as the previous round.

### Validation

Ran: Eval build plus `test:dist` (Node 75/75, relay lifecycle 23/23), Runtime Host focused 59/59, and the dist-level probe described under the P1. Not run: any real Harbor trial.

1625567290:
Addressed both blocking findings at exact head `887140c4a`.

Admission identity and at-most-once execution:
- `hosted.execution.start` now requires `{ execution, admissionToken }`; the server validates the token against the exact execution input, Host epoch, and connection identity.
- The admission record remains as the terminal replay cache. A correct retry returns the same settled result without invoking the execution again.
- No admission, a wrong token, changed execution input, another connection, and another Host epoch all fail closed.
- The fast-settle `admit -> setImmediate -> start` regression proves one invocation and one replayed result.
- Compatibility epoch 43 prevents older peers from entering this contract.

Real Harbor lifecycle coverage:
- Eval changes now select a CI lane that installs pinned `harbor==0.20.0` into an isolated environment and runs the opt-in Docker lifecycle tests.
- The framework-timeout case uses a real Harbor `DockerEnvironment` and `Trial`: the leader exits, a background heartbeat remains available to the verifier, the verifier scores the preserved cell, and Harbor removes the environment after verification.
- The Host-abort case proves the exact Compose project and main container existed, then verifies teardown, cancellation, and absence of an executed result frame.

Post-rebase local validation on current `main`: build, typecheck, lint, format, ASF headers, CI planner 30/30, Runtime Host 1120/1120, Eval Node 75/75, relay contract 30/30, and relay lifecycle 23 with 3 platform skips. Harbor 0.20 parsed the generated task fixture in a disposable environment. Docker is not installed on this machine, so the two real-container tests were skipped locally; they are now mandatory in the selected Linux CI lane rather than being claimed as local evidence.

The fresh exact-head checks are ready for maintainer approval.

1625567290:
Follow-up on the exact-head CI failure: run [32643831667](https://github.com/apache/maka/actions/runs/32643831667) failed before creating any job because the Harbor lifecycle `run:` scalar was invalid YAML. The quoted executable path ended before the remaining command text.

Fixed at `0ef94520c14f073d3271e9c29e282a31543144df` by using a YAML block scalar for the command.

Local validation:
- workflow YAML parse: clean
- CI planner contract: 30/30
- repository format check
- ASF header check
- `git diff --check`

The lifecycle implementation and its earlier focused test results are unchanged; this commit only makes the hosted workflow executable. Fresh exact-head [CI](https://github.com/apache/maka/actions/runs/32644979336) and [Dependency audit](https://github.com/apache/maka/actions/runs/32644979330) runs are waiting for maintainer approval.

packages/eval/harbor/relay_agent.py
```diff
@@ -272,6 +236,106 @@ async def run(self, instruction: str, environment: Any, context: Any) -> None:
...
+        # because the verifier still runs.
+        if terminal_projection is not None:
+            result = terminal_result
+        elif _host_teardown_requested:
+            result = await _settle_or_destroy(
+                environment, cwd, scope_path, execution, self._teardown_timeout
+            )
+        else:
+            result = await _stop_subject_for_timeout(
+                environment, cwd, scope_path, execution, self._teardown_timeout
```
Astro-Han: P1：framework timeout 路径复用了 host 的 5000ms destructive teardown budget；当前 Harbor agent timeout 为 1s，外层 deadline 会先到期，导致 cleanup 未完成、`executed` frame 不发送，verifier 无法运行。请为 framework cancellation 使用不超过 framework deadline 的 cleanup budget；host abort 可继续保留独立的较长 teardown budget。

Astro-Han (commented):
独立复核当前 exact head `0ef94520`。发现 1×P1，已在 `relay_agent.py:299` 行内说明：framework timeout 的 cleanup 复用了 5000ms host teardown budget，而 Harbor agent timeout 为 1s；真实 Harbor lifecycle test 在当前 CI 以 `AgentTimeoutError` 失败，`executed` frame 未发送，verifier 未运行。请将该路径限制在 framework deadline 内，同时保留 host abort 的独立破坏性 teardown budget。

旧 admission token / at-most-once 路径已复核闭合，runtime-host focused tests 80/80 与 relay lifecycle 23/23 通过；当前 `test` failure、`audit` success，且 PR 仍 conflict，因此本次只给 COMMENT。

packages/eval/harbor/relay_agent.py
1625567290: Fixed at exact head `2ff514282` after rebasing onto current `main` (`3bb645e99`).

Reproduction before the fix:
- The pinned Harbor lifecycle lane failed with `AgentTimeoutError` at a 1s agent timeout.
- A deterministic relay probe with a 50ms framework timeout and 5000ms teardown timeout showed the leader-stop command receiving about 4.0s and the outer operation ending in `TimeoutError`, even though a timeout frame was eventually emitted.

The relay now receives a separate `framework_timeout_ms`, derived from Harbor’s resolved agent timeout after task loading (including overrides, maximums, multipliers, and the shortest multi-step timeout). Framework cancellation uses only that budget; Host abort retains the existing independent destructive teardown budget. The relay suppresses the framework cancellation only after the executed frame is confirmed delivered. Cleanup failure, transport failure, or re-cancellation still converges on bounded settle-or-destroy and rethrows.

Post-rebase validation:
- Eval Node tests: 75/75
- relay contract: 30/30
- relay lifecycle: 24 run, 3 platform skips
- Runtime Host: 1129/1129
- CI planner: 31/31
- workflow YAML parse, lint, format, ASF headers, and `git diff --check`

The two pinned real-Harbor Docker tests remain unavailable locally because Docker is not installed on this machine; they are still mandatory in the selected Linux CI lane. Fresh exact-head CI is pending.

1625567290:
The fresh exact-head workflows are waiting for maintainer approval:
- [CI 32682514406](https://github.com/apache/maka/actions/runs/32682514406)
- [Dependency audit 32682514470](https://github.com/apache/maka/actions/runs/32682514470)

The PR is now mergeable against `main`; the rebase conflict was resolved by assigning the hosted-execution admission contract compatibility epoch 45, after the epochs already owned by current `main`.

packages/eval/harbor/relay_agent.py
1625567290: Fixed in `8e9b931d5`. I reproduced the deadline failure before changing the signal loop: with a 200 ms framework cleanup budget, the previous implementation spent the full leader-stop window on `TERM`, never attempted `KILL`, destroyed the environment, and raised `Maka Eval could not confirm subject exit after framework timeout`. The CI Harbor case was the same boundary at a 1 s agent timeout.

The timeout path now keeps the framework-specific deadline and divides the remaining stop window across every remaining signal, so a TERM-delayed leader cannot consume the KILL opportunity. The same reproduction now records `TERM` then `KILL`, returns the subject result, and leaves the environment intact. Host abort still uses the independent destructive teardown budget.

Added a short-deadline regression covering the delayed-TERM/required-KILL case. Local verification on Node 24.19.0: relay lifecycle 25/25 with 3 platform skips, relay contract 30/30, Eval dist suite, Runtime Host 1130/1130, full workspace suite, typecheck, lint, format, ASF headers, and 73/73 CI/Windows contract tests all pass. The real Docker/Harbor test is not runnable on this machine because Docker is unavailable, so the refreshed exact-head CI remains the final integration proof.

1625567290:
Follow-up for the failing real-Harbor CI is now pushed at `8e9b931d5`.

Before the fix, a short framework deadline let the TERM attempt consume the entire non-destructive stop window. The 200 ms local reproduction issued only TERM, destroyed the environment, and raised the same unconfirmed-exit error seen by the 1 s Harbor test. The updated loop reserves time for every remaining signal; the same reproduction now issues TERM and KILL, observes subject completion, and preserves the verifier environment.

Validation on the rebased head with Node 24.19.0:
- full workspace test suite passed
- Runtime Host 1130/1130 passed
- Eval build and complete dist suite passed
- relay lifecycle 25/25 with 3 platform skips; relay contract 30/30
- typecheck, lint, format check, and ASF header check passed
- CI/Windows contract suite 73/73 passed

The Docker-backed Harbor lifecycle test remains CI-only on this machine because Docker is unavailable. The branch is rebased on current `main`; its 21-file diff contains no documentation or Java test sources.

1625567290:
Follow-up on exact-head CI run 32687216792: the real Harbor lifecycle test still failed at 8e9b931d5 in test_framework_timeout_preserves_background_for_verifier. The previous change ensured that both TERM and KILL were attempted, but it still paid two Docker environment.exec round trips. Under the 1s framework budget, those control-plane trips plus execution-result observation exhausted the 0.8s non-destructive stop window even after KILL had been issued.

Fixed at 39e1807eca008593886ca3424d91b51c97f5b8ab, rebased onto current main d8bd23d46. TERM, a bounded grace interval, and KILL now run in one container command against the session leader only. The remaining budget is used to confirm the subject execution result; the background process group is still left intact for verification.

Before-fix local timing reproduction with 250ms signal and result-observation latency:
- two control round trips, TERM then KILL
- 0.801s elapsed
- unconfirmed-exit RuntimeError and environment destruction

After-fix reproduction with the same delays:
- one control round trip containing TERM and KILL
- 0.503s elapsed
- subject exit confirmed and environment preserved

A real local process probe also confirmed that the leader exits with SIGKILL while its task-owned background process remains alive and its heartbeat continues advancing; no environment deletion was requested.

Final exact-head validation with Node 24.19.0:
- full npm test: all workspaces passed
- relay lifecycle: 25/25 passed with 3 platform skips
- full workspace typecheck, lint, format check, and ASF header audit passed
- CI/Windows contract suite: 70/70 passed
- git diff check passed
- 21-file PR diff contains no documentation or Java test sources

Docker is unavailable on this Mac, so the exact Harbor DockerEnvironment case remains the required hosted confirmation. The fresh workflows are waiting for maintainer approval:
- CI: https://github.com/apache/maka/actions/runs/32690207738
- Dependency audit: https://github.com/apache/maka/actions/runs/32690207860

packages/runtime-host/src/server/hosted-execution-coordinator.ts
```diff
@@ -62,36 +72,101 @@ export class HostHostedExecutionCoordinator {
...
-    this.#executions.set(input.executionId, { input: structuredClone(input), abort, task });
-    return { ok: true, result: structuredClone(await task) };
+    const execution = {
+      input: structuredClone(input),
+      authority: { hostEpoch: context.hostEpoch, connectionId: context.connectionId },
+      abort,
+      admissionToken,
+      task,
+    };
+    this.#executions.set(input.executionId, execution);
```
Astro-Han, line 176: [P2] Completed execution records are never released, so a long-lived Host grows a receipt per distinct execution id.

`#executions` is only ever written — `set` at `:168`, `get` at `:88`/`:129`/`:176` — with no `delete` anywhere and no drain on the settle path. A preserved record for a completed or failed execution stays strongly referenced for the Host's lifetime. `releaseToEnvironment()` does not help: in `client/launcher.ts:96-98` it is exactly `child.unref()`.

The public Runtime Host protocol lets one connection admit different execution ids, sequentially or concurrently, so this is reachable without misuse. An exact-head probe ran 1,000 sequential then 1,000 concurrent admits successfully; before draining, 2,000 runs and 0 aborts, and `beginDrain()` then produced 2,000 aborts — i.e. all 2,000 completed records were still held.

Calibration, so this is not overstated: the ordinary Harbor path mints a fresh UUID, a fresh root and an owned one-shot Host per invocation, so it will not accumulate thousands on a single Host in practice. That is why this is a P2 and not a P1.

The smallest fix is to encode the real one-shot contract in the coordinator: after the first execution identity, reject a different id and keep exactly one replay receipt until the Host exits. Note that deleting the same-id record after settle is *not* a safe shortcut — that would revive the fast-settlement at-most-once race. Supporting genuine multi-execution instead would need an explicit, bounded receipt lifecycle.

<details>
<summary>中文</summary>

`#executions` 只写不删：`:168` set，`:88`/`:129`/`:176` get，全文件没有 `delete`，settle 分支也不清理。已完成或失败的 execution 记录会被强引用到 Host 退出为止；`releaseToEnvironment()` 帮不上忙——它在 `client/launcher.ts:96-98` 就是一句 `child.unref()`。

公开的 Runtime Host 协议允许同一 connection 顺序或并发 admit 不同的 execution id，所以这条路径不需要误用就能走到。exact head 上的 probe：1,000 次顺序 + 1,000 次并发 admit 全部成功，drain 前 2,000 runs / 0 aborts，`beginDrain()` 之后产生 2,000 个 abort，说明这 2,000 条已完成记录仍然全部被持有。

校准一下以免夸大：正常的 Harbor 路径每次调用都会新建 UUID、新建 root、并拥有一个 one-shot Host，实际不会在单个 Host 上堆到几千条——所以定 P2 而不是 P1。

最小修法是把真正的 one-shot 契约写进 coordinator：确定首个 execution identity 之后拒绝不同 id，只保留一条 replay receipt 直到 Host 退出。注意 settle 之后删除同 id 记录**不是**安全的捷径，那会让 fast-settlement 的 at-most-once 竞态复活。若确实要支持多 execution，则需要一个显式且有界的 receipt 生命周期。
</details>

Astro-Han (commented):
Review at `39e1807e`. One **[P2]**, filed inline on `hosted-execution-coordinator.ts`. No P0/P1.

What holds: within a single Host epoch the admission token is bound correctly to the exact input, `hostEpoch` and `connectionId`. The same id will run again on a *fresh* Host, but the first-party Eval path neither reconnects nor resubmits and its adapter mints a fresh id and root each time, so that does not become a second finding.

Gate status: `audit` is green on this head; `test` was still in progress at review time, so this is not an approval.

<details>
<summary>中文</summary>

在 `39e1807e` 上审，一条 **[P2]**，行内提在 `hosted-execution-coordinator.ts`，无 P0/P1。

成立的部分：单个 Host epoch 内，admission token 与精确输入、`hostEpoch`、`connectionId` 的绑定是正确的。同一 id 在**全新的** Host 上会再次运行，但第一方 Eval 路径既不 reconnect 也不 resubmit，其 adapter 每次都新建 id 和 root，所以这一点不构成第二条 finding。

门禁状态：本 head 上 `audit` 已绿，`test` 在发布时仍在运行中，因此这不是一条批准。
</details>

packages/eval/harbor/relay_agent.py
```diff
@@ -584,6 +656,59 @@ async def _settle(environment: Any, cwd: str, scope_path: str, execution: Any) -
...
+                raise
+            except (TimeoutError, asyncio.TimeoutError):
+                pass
+    if execution.done() and not execution.cancelled():
+        return execution.result()
+    # A verifier cannot measure a stable environment while the subject may
+    # still be mutating it. Fail closed instead of publishing a scoreable
+    # framework_timeout frame without positive leader-exit evidence.
+    await _destroy_environment(environment, execution, deadline, loop)
+    raise RuntimeError("Maka Eval could not confirm subject exit after framework timeout")
```
Astro-Han, line 829: [P1] The PR's headline path fails on its own new test at this exact head.

`test` run [32690207738](https://github.com/apache/maka/actions/runs/32690207738) is completed/**failure** on `39e1807e`. The failing step is *Run real Harbor lifecycle tests*, and the failing case is `test_framework_timeout_preserves_background_for_verifier` (`test_harbor_trial_lifecycle.py:84`):

```
RuntimeError: Maka Eval could not confirm subject exit after framework timeout
  relay_agent.py:203  run -> _cleanup_cancelled_execution
  relay_agent.py:268  _cleanup_cancelled_execution -> _finalize_cancelled_execution
  relay_agent.py:320  _finalize_cancelled_execution -> _stop_subject_for_timeout
  relay_agent.py:709  raise RuntimeError(...)
```

This is not inherited: `test` is green on current `main` (`b877962`), and both the failing step and the failing test are added by this PR.

The consequence is the feature's own goal. Under a 1s framework timeout, cleanup cannot confirm the leader exited, so the relay never emits a scoreable `framework_timeout`, the verifier never runs, and the "preserve the background process for the verifier to score" path does not hold.

A re-run will not settle this — the fix has to make the real Docker/Harbor leader-stop plus `environment.exec` settle reliably inside the framework deadline while still leaving the background process alive. The real Harbor lifecycle lane needs to reach terminal green.

<details>
<summary>中文</summary>

本 head 上 `test` run `32690207738` 已 completed/**failure**，失败步骤是 *Run real Harbor lifecycle tests*，失败用例是 `test_framework_timeout_preserves_background_for_verifier`（`test_harbor_trial_lifecycle.py:84`），异常沿 `relay_agent.py:203 → :268 → :320 → :709` 抛出 `Maka Eval could not confirm subject exit after framework timeout`。

这不是继承来的红：当前 `main`（`b877962`）上 `test` 是绿的，而失败的步骤和用例都是本 PR 新增的。

后果正是本功能自身的目标：在 1 秒的 framework timeout 下，cleanup 无法确认 leader 已退出，relay 就不会发出可评分的 `framework_timeout`，verifier 不会运行，"保留后台进程供 verifier 评分"这条路径因此不成立。

重跑解决不了这个问题——需要让真实 Docker/Harbor 的 leader-stop 加 `environment.exec` 在 framework deadline 内可靠 settle，同时保留后台进程；real Harbor lifecycle 这条 lane 必须达到终态绿。
</details>

Astro-Han (commented):
Updating my earlier review on the same head `39e1807e`: **now NO-GO with 1×P1 + 1×P2.**

The P1 is new information rather than a revision of judgement — the `test` run on this head reached a terminal state after that review was posted, and it failed on this PR's own real-Harbor lifecycle test. Details inline.

The P2 from the earlier review (unbounded execution receipts on a long-lived Host) is unchanged and still open.

`audit` remains green on this head.

<details>
<summary>中文</summary>

更新同一 head `39e1807e` 上的结论：**现为 NO-GO，1×P1 + 1×P2。**

这条 P1 是新出现的事实而非改判——上一条 review 发布之后，本 head 的 `test` 才跑到终态，并且失败在本 PR 自己的 real Harbor lifecycle 测试上，细节见行内。上一条 review 里的 P2（长寿命 Host 上 execution receipt 无界增长）维持不变，仍未闭。本 head 的 `audit` 仍为绿。
</details>

packages/eval/harbor/relay_agent.py
1625567290: Addressed in exact head d151aebab.

The failure was not in the TERM/KILL selection itself. Harbor 0.20.0 runs DockerEnvironment.exec through docker compose exec, streams combined command output, and waits for that stream to reach EOF. For captureStdout=false, the relay redirected only stdout. The preserved background process inherited stderr, so the exec stream stayed open after the subject leader exited; the cleanup deadline expired and the relay reported that it could not confirm exit.

The non-capturing command now detaches both streams with >/dev/null 2>&1. This does not affect output-capturing commands.

Before the change, a local pipe reproduction using (sleep 2) & exit 0 with only stdout redirected did not settle within 250 ms because the background child retained stderr. With both streams detached, the same probe returned exit code 0 in 9 ms. A relay regression now locks the generated command to both redirects.

Exact-head validation after rebasing onto current main:
- full workspace suite passed, including Desktop 1364/1364
- relay lifecycle 26 passed with 3 platform skips
- relay contract 30/30
- Runtime Host 1132/1132
- typecheck, lint, format, ASF headers, and 70/70 CI/Windows contract tests

Docker is not installed on this machine, so the refreshed real Harbor lifecycle check remains the final integration proof.

packages/runtime-host/src/server/hosted-execution-coordinator.ts
1625567290: Addressed in exact head d151aebab.

The coordinator now encodes the existing one-shot Host contract. The first admit or cancel binds the execution identity. Replays for that same identity retain the single cached receipt, preserving the fast-settlement at-most-once guarantee. Every different execution identity is rejected with operation_conflict, so one Host cannot accumulate an unbounded set of completed receipts.

A direct coordinator probe before the change accepted and ran all 200 distinct ids, retained all 200 records, and produced 200 aborts on drain. After the change it accepted one id, rejected the other 199, ran once, and produced one abort on drain. Tests also cover concurrent different-id admission and cancellation as the first identity-binding operation.

Exact-head validation after the rebase: coordinator tests 9/9, Runtime Host 1132/1132, the full workspace suite, typecheck, lint, format, ASF headers, and 70/70 CI/Windows contract tests all pass.

1625567290:
Updated PR #3191 to exact head d151aebab after rebasing onto current main b87796278.

This update addresses both newly reported blockers:
- The real Harbor timeout failure came from the preserved background process retaining stderr while Harbor waited for the combined docker compose exec output stream to close. Non-capturing execution now detaches both stdout and stderr.
- The Runtime Host coordinator now enforces its one-shot execution identity and retains at most one replay receipt; different execution ids conflict.

Both review threads have detailed before/after reproductions and are resolved.

Exact-head local verification passed: the full workspace suite, relay lifecycle 26 with 3 platform skips, relay contract 30/30, Runtime Host 1132/1132, coordinator 9/9, Desktop 1364/1364, typecheck, lint, format, ASF headers, and 70/70 CI/Windows contract tests.

The fresh CI and dependency-audit workflows were created for this head and currently require maintainer approval:
- https://github.com/apache/maka/actions/runs/32691809036
- https://github.com/apache/maka/actions/runs/32691809045

Docker is unavailable locally, so the approved real Harbor lifecycle job is still the final integration gate.

1625567290:
@Astro-Han Rebased this PR onto current `main` (`1e1c886ae`) at exact head `461687f51`; the merge conflicts are resolved.

The conflicts were substantive in two places:
- Current `main` already owns Runtime Host compatibility epochs 45 and 46, so hosted-execution admission is now epoch 47 instead of colliding with the queue-entry epoch.
- CI keeps both current `main`'s live Eval egress-proxy lifecycle and this PR's pinned Harbor 0.20.0 lifecycle job.

Exact-head local validation passed:
- full `npm test`: every workspace passed;
- Runtime Host: 1143/1143;
- relay lifecycle: 26 tests, with the 3 platform/opt-in cases skipped as designed;
- CI planner: 32/32;
- typecheck, lint (2680 files), format (1618 files), ASF headers (2815 covered files), and `git diff --check`.

Docker is not available on this machine, so I did not claim a local run of the real Harbor container lifecycle. The pre-rebase head's required checks were green; the rebased head now needs the fresh CI run to re-establish that integration evidence against current `main`.

1625567290:
Rebased onto current main 51eeaa60b; the exact head is e2d3f2cd6. The hosted checks are terminal and successful:

- CI: https://github.com/apache/maka/actions/runs/32929060480
- Dependency audit: https://github.com/apache/maka/actions/runs/32929060502

Exact-head local validation also passed: Runtime Host tests; Eval 75/75 Node tests and all selected Python lifecycle, relay, policy, artifact, and egress suites; CI planner 37/37; full workspace typecheck; ASF headers; and changed-file lint/format checks.

The earlier CI attempt failed only in the unchanged main WorkHub E2E. The replacement head has an identical file tree, and its full hosted CI completed successfully. There are no unresolved review threads. Ready for human review.

1625567290:
Exact head `28e2fe28d2` is based on `main@bfba253613`. All hosted workflows are terminal and successful:

- CI: https://github.com/apache/maka/actions/runs/32955273146
- Dependency audit: https://github.com/apache/maka/actions/runs/32955273218
- Windows recovery: https://github.com/apache/maka/actions/runs/32955273125

The hosted CI passed affected workspace tests, the live Eval egress proxy, the real Harbor lifecycle suite, Runtime Host tests, Desktop E2E, alignment audit, and Storybook smoke. The exact PR diff contains no documentation/Markdown or Java test sources, GitHub reports it mergeable, and there are no unresolved review threads. Ready for human review.

1625567290:
Rebased onto current `main@a6866d059` and force-updated the PR to exact head `cb0efa8c8`.

The conflict in `packages/runtime-host/src/protocol/index.ts` was resolved semantically: current main owns compatibility epochs 51 and 52, so the hosted-execution admission boundary now uses epoch 53. This preserves main's durable steering echo contract and this PR's server-issued admission token / at-most-once execution contract without reusing an incompatible epoch.

Exact-head local validation with a clean dependency install passed: full test build, Runtime Host dist suite, Eval Node and Python lifecycle/policy/artifact suites, CLI dist suite, and CI planner 44/44. `range-diff` shows the other 11 commits unchanged; only the two epoch-touching commits differ as expected. The 21-file diff contains no documentation/Markdown or Java test sources. GitHub reports the PR mergeable with no unresolved review threads.

1625567290:
Follow-up on exact head `cb0efa8c8`: GitHub has not created any check suite after the recorded synchronize event. I verified the PR is open and mergeable and the repository workflows are active. A close/reopen recovery on #3819 also created no run, while direct workflow dispatch from this fork contributor is rejected with HTTP 403 (admin rights required). Maintainer action is needed to approve or trigger the GitHub-hosted workflows for this exact head; the exact-head local validation reported above remains green.

packages/eval/harbor/run_trial.py
```diff
@@ -162,10 +163,43 @@ def apply_subject_egress_policy(task: object) -> None:
...
+    if timeout_ms <= 0:
+        raise RuntimeError("Harbor agent timeout is shorter than one millisecond")
+    config.agent.kwargs["framework_timeout_ms"] = timeout_ms
+
+
 async def create_harbor_trial(trial_type: type, config: object) -> object:
     trial_type._resolve_agent_skills(config)
     task, task_download_result = await trial_type._load_task(config)
     apply_subject_egress_policy(task)
+    apply_framework_timeout_budget(config, task)
```
Astro-Han: [P1 · category ①] This budget is applied only inside create_harbor_trial; run_trial sends the supported Pier arm directly through trial_type.create(config). createPierExecutor and the CLI both reach that branch, so Pier RelayAgent receives no framework_timeout_ms and falls back to the 110s destructive teardown budget even when its framework agent/step timeout is much shorter. Cancellation can therefore block or destroy the environment instead of preserving leftovers under the same public contract. Please inject the effective timeout at a shared construction seam for Pier too, and add Pier wiring plus timeout-cleanup coverage; the current policy test and selected CI lifecycle test exercise Harbor only.

Astro-Han (commented):
I rereviewed the exact head after the Harbor timeout fixes. The confirmed Harbor path is materially stronger, but the same contract is not yet applied to the other supported arm. Exact head is cb0efa8c855caf02e34cd6db0eb0c6ca44becf54 and current merge state is CONFLICTING/DIRTY. Review analysis was assisted by Codex and independent @reviewer agents. Astro-Han verified the public Pier composition, pinned framework behavior, test coverage, and severity before publication and owns this review.

packages/eval/harbor/run_trial.py
```diff
@@ -162,10 +163,43 @@ def apply_subject_egress_policy(task: object) -> None:
...
+    timeouts = []
+    for default_timeout in defaults:
+        base_timeout = config.agent.override_timeout_sec or default_timeout
+        if base_timeout is None:
+            continue
+        maximum = config.agent.max_timeout_sec or float("inf")
+        timeouts.append(min(base_timeout, maximum) * multiplier)
+    if not timeouts:
+        return
+    timeout_ms = math.floor(min(timeouts) * 1000)
```
Astro-Han: [P1 · category ①] Taking min(timeouts) once and storing it on the shared RelayAgent applies the shortest step cleanup budget to every step. Harbor MultiStepTrial executes each step with its own effective timeout; with A=1s and B=60s, B still receives framework_timeout_ms=1000 and can be terminated/reported after one second. The current policy test merely asserts the minimum and therefore locks in the bug. Please carry the current step timeout into each execution rather than collapsing the whole trial to one minimum, and add a two-step lifecycle regression with different budgets.

packages/eval/harbor/relay_agent.py
```diff
@@ -321,7 +393,10 @@ async def _prepare_command(
         if secret_path is not None:
             secret_path.unlink(missing_ok=True)
     subject = shlex.join([request["command"], *request["args"]])
-    output_redirect = "" if capture_stdout else " >/dev/null"
+    # An exit-code subject has no structured stdout to preserve. Detach both
+    # streams so a task-owned background process cannot keep Harbor's
+    # `docker compose exec` output pipe open after the subject leader exits.
+    output_redirect = "" if capture_stdout else " >/dev/null 2>&1"
```
Astro-Han: [P1 · category ①] The leftover-preservation fix only detaches pipes when captureStdout is false, but the structured protocol path defaults captureStdout to true. A normal subject that starts a background service and exits leaves that descendant holding stdout/stderr; Harbor DockerEnvironment.exec waits for stream EOF, so leader-only TERM/KILL never makes execution settle. The cleanup deadline then destroys the environment, losing the service the verifier was meant to score (and a logging child can keep accumulating captured output). The real-Harbor test forces captureStdout=false, so it misses the default path. Please make structured capture compatible with descendant preservation and cover capture=true with a child holding/writing the pipe.

Astro-Han (commented):
Two additional exact-head gaps remain in the framework-timeout contract. They are independent of the already-reported Pier wiring gap and both affect supported Harbor execution shapes. Review analysis was assisted by Codex and independent @reviewer agents. Astro-Han verified the pinned Harbor behavior, production command construction, per-step timeout semantics, and severity before publication and owns this review.

packages/eval/harbor/run_trial.py
1625567290: Fixed in exact head 595fcc17b. Timeout binding now runs after both Harbor and Pier trial construction. Pier single-agent execution receives its configured agent timeout, and each Pier multi-step execution receives the exact resolved timeout for that step. Focused policy tests cover both paths.

packages/eval/harbor/run_trial.py
1625567290: Fixed in exact head 595fcc17b. The static shortest-timeout assignment was removed. Harbor updates RelayAgent immediately before each agent phase with that phase's effective timeout, so a 1-second first step no longer pins a later 60-second step to 1 second. A mixed-budget regression verifies the per-phase values.

packages/eval/harbor/relay_agent.py
1625567290: Fixed in exact head 595fcc17b. Structured stdout and stderr capture now uses bounded file collectors outside the Docker exec output stream, while the subject and its descendants write to FIFOs that remain drained until final cleanup. A background descendant can therefore keep writing without holding Harbor exec open or growing an unbounded in-memory capture. The real Harbor lifecycle case now uses captureStdout=true and verifies both continued heartbeat and background log output.

1625567290: Follow-up in exact head 5b65297d6 after rebasing onto current main 5058b29f1. Hosted Linux CI reached the standard Eval suite and exposed two stale GNU-setsid test assumptions: raw subprocess stdout is intentionally empty after bounded capture moved to collector files, and the scope record now publishes pgid plus leader. The tests now read through the production capture loader, parse both scope fields, and terminate their collector groups during cleanup. A local GNU-setsid-compatible run passes all 26 relay lifecycle cases; the full Eval suite, Runtime Host focused 127/127, CLI 644/644, and CI planner 49/49 are green.

1625567290: Final follow-up in exact head `b77806537`. The first hosted run after the bounded-capture change exposed one additional real-Harbor boundary: relay cleanup killed the FIFO collectors before verification, so a preserved service that inherited stdout/stderr lost its only readers and could terminate with SIGPIPE. The collectors now outlive the relay and drain until all inherited writers close; relay cleanup removes only the capture pathnames. The regression keeps the preserved service attached to stdout/stderr so this contract is exercised.

Exact-head CI is now terminal and successful, including the standard Linux workspace suite and the real Harbor `captureStdout=true` background-service lifecycle case: https://github.com/apache/maka/actions/runs/33322610344

1625567290:
Exact-head follow-up for `b77806537`:

- CI `test` succeeded, including standard Linux workspace tests, live Eval egress proxy, the real Harbor lifecycle suite, Runtime Host tests, Desktop E2E, alignment audit, Storybook smoke, and installed CLI release-candidate validation: https://github.com/apache/maka/actions/runs/33322610344
- Dependency audit and Windows recovery succeeded: https://github.com/apache/maka/actions/runs/33322610315 and https://github.com/apache/maka/actions/runs/33322610317

All reported timeout-contract findings are resolved, with zero unresolved review threads. The 22-file PR diff contains no documentation/Markdown files or Java test sources. `main` has since advanced to `5d519d6f9`; GitHub reports this exact head mergeable, and a local merge-tree check finds no textual conflict. Ready for maintainer review.

hqhq1025 (commented):
Exact-head review of `b778065372cef2bbabe161d709ac2ff1f5e032fb` found one blocking correctness issue in the structured-output capture path. The per-phase timeout wiring, preserved-descendant behavior, and one-shot Hosted Execution changes otherwise close the earlier findings, and the current hosted `test`, `audit`, and `windows_recovery` checks are green, including the real Harbor lifecycle step. Details are inline.

Local validation: clean install; `npm run build:test`; full typecheck; Eval dist suite (75 Node tests plus all Python suites, with platform skips only); CI planner/operator tests 55/55; changed-file Biome; `git diff --check`; and a clean merge-tree against current `main`. A focused Runtime Host run was 126/127 locally because the sandbox-context integration failed on this Linux host, while the hosted Runtime Host step is green.

packages/eval/harbor/relay_agent.py
```diff
@@ -321,18 +418,113 @@ async def _prepare_command(
...
+    cwd: str,
+    result: Any,
+    request: dict[str, Any],
+    token: str,
+) -> Any:
+    if not request.get("captureStdout", True):
+        return result
+    stdout_path, stderr_path, _, _, _ = _capture_paths(token)
+    stdout_marker = f"MAKA-EVAL-CAPTURE-STDERR-V1 {token}"
+    captured = await environment.exec(
```
hqhq1025, line 507: [P1] Synchronize this read with the detached collectors. `environment.exec` can return as soon as the subject leader exits, while the FIFO collector is still draining bytes into `stdout_path`; a preserved descendant intentionally keeps the FIFO open, so EOF cannot be used as the boundary. This immediate `cat` can therefore read only a prefix before the terminal result frame is durable. On this exact head, a full RelayAgent socket-path probe with 40 KB of valid pre-frame output and a descendant holding stdout open returned `termination: exited`, `exitCode: 0`, and `result-frame-missing` in 10/10 runs (the direct capture helper reproduced the partial read in 30/30 runs). `packages/eval/src/maka-subject.ts:93-106` then converts the otherwise successful evaluation into `infra_failed`. Add an explicit collector acknowledgement/terminal-frame handoff (or a dedicated result channel) and wait for that durable boundary before loading the files, while still allowing the collector to drain later descendant output without waiting for FIFO EOF.

PR #3195, opened 2026-08-18
fix(desktop): contain first-run onboarding viewport

ARE404:
## Summary

- Keep the first-run onboarding surface inside ChatLayout's actual available height instead of recomputing it from `100dvh`.
- Remove the transcript-only empty spacer and composer dock padding only while onboarding owns the empty state, so the page itself cannot scroll.
- Preserve the existing onboarding card and provider-list scrolling, and add a real Electron regression test for the viewport contract.

## Root cause

ChatMessageList reserves a flexible spacer before ordinary empty states. The onboarding surface also declared a viewport-derived minimum height, so it overflowed its half-height flex slot and expanded ChatLayout's self-scroll range. In the failing test, a 952px chat viewport produced 1159px of scrollable content even though the card itself fit on screen.

## Verification

- `npx biome check apps/desktop/src/renderer/styles/onboarding.css apps/desktop/e2e/fixtures.ts apps/desktop/e2e/onboarding-viewport.spec.ts`
- `npx playwright test --config e2e/playwright.config.ts e2e/onboarding-viewport.spec.ts`
- Visually inspected a 3300×1984 real Electron capture: the card and skip action stay within the chat viewport and the outer scrollbar is gone.

## AI use

Select exactly one:

Tool(s) and scope: Codex diagnosed the flex/viewport interaction, implemented the scoped CSS fix, and added the Electron E2E regression test.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/e2e/fixtures.ts
+++ b/apps/desktop/e2e/fixtures.ts
@@ -339,6 +339,7 @@ async function withE2eWindow(
 
 export const test = base.extend<{
   window: Page;
+  onboardingWindow: Page;
   gitReviewWindow: { page: Page; projectRoot: string };
   invocableSkillsWindow: Page;
   linkColorWindow: Page;
@@ -349,6 +350,14 @@ export const test = base.extend<{
   // Seeded: a pre-staged connection clears onboarding so the composer is ready.
   window: async ({}, use) => {
     await withE2eWindow({ seed: true, readinessSelector: COMPOSER_INPUT, locale: 'zh' }, use);
+  },
+  onboardingWindow: async ({}, use) => {
+    await withE2eWindow({
+      seed: false,
+      readinessSelector: '[data-maka-contract="onboarding-card"]',
+      locale: 'zh',
+      showWindow: true,
+    }, use);
   },
   gitReviewWindow: async ({}, use) => {
     await withE2eWindow(
--- a/apps/desktop/e2e/onboarding-viewport.spec.ts
+++ b/apps/desktop/e2e/onboarding-viewport.spec.ts
@@ -1,0 +1,98 @@
+import { expect, test } from './fixtures';
+
+test('first-run onboarding stays within the chat viewport', async ({ onboardingWindow }) => {
+  const geometry = await onboardingWindow.evaluate(() => {
+    const pageRoot = document.scrollingElement;
+    const scrollContainer = document.querySelector<HTMLElement>('[data-chat-scroll-container="true"]');
+    const surface = document.querySelector<HTMLElement>('[data-maka-contract="onboarding-surface"]');
+    const card = document.querySelector<HTMLElement>('[data-maka-contract="onboarding-card"]');
+    if (!pageRoot || !scrollContainer || !surface || !card) {
+      throw new Error('Onboarding geometry is unavailable');
+    }
+
+    const scrollRect = scrollContainer.getBoundingClientRect();
+    const surfaceRect = surface.getBoundingClientRect();
+    const cardRect = card.getBoundingClientRect();
+    return {
+      pageClientHeight: pageRoot.clientHeight,
+      pageScrollHeight: pageRoot.scrollHeight,
+      clientHeight: scrollContainer.clientHeight,
+      scrollHeight: scrollContainer.scrollHeight,
+      surfaceTop: surfaceRect.top,
+      surfaceBottom: surfaceRect.bottom,
+      viewportTop: scrollRect.top,
+      viewportBottom: scrollRect.bottom,
+      cardTop: cardRect.top,
+      cardBottom: cardRect.bottom,
+    };
+  });
+
+  expect(geometry.pageScrollHeight, JSON.stringify(geometry, null, 2)).toBe(
+    geometry.pageClientHeight,
+  );
+  expect(geometry.scrollHeight, JSON.stringify(geometry, null, 2)).toBe(geometry.clientHeight);
+  expect(geometry.surfaceTop).toBeGreaterThanOrEqual(geometry.viewportTop);
+  expect(geometry.surfaceBottom).toBeLessThanOrEqual(geometry.viewportBottom);
+  expect(geometry.cardTop).toBeGreaterThanOrEqual(geometry.viewportTop);
+  expect(geometry.cardBottom).toBeLessThanOrEqual(geometry.viewportBottom);
+});
+
+test('first-run onboarding remains reachable at the minimum window size', async ({
+  onboardingWindow,
+}) => {
+  await onboardingWindow.setViewportSize({ width: 480, height: 320 });
+
+  const surface = onboardingWindow.locator('[data-maka-contract="onboarding-surface"]');
+  const initialScrollTop = await surface.evaluate((element) => element.scrollTop);
+  await surface.hover();
+  await onboardingWindow.mouse.wheel(0, 10_000);
+  await expect.poll(() => surface.evaluate((element) => element.scrollTop)).toBeGreaterThan(
+    initialScrollTop,
+  );
+
+  const geometry = await onboardingWindow.evaluate(() => {
+    const pageRoot = document.scrollingElement;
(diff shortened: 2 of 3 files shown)


hqhq1025 (approved):
Re-reviewed head `dd33cc1` against current `main`. No actionable findings.

The root cause is real: the transcript-only flex spacer and the onboarding surface's viewport-derived minimum height combined to expand ChatLayout beyond its available block size. The final patch keeps outer scroll ownership with ChatLayout, preserves user-reachable internal scrolling at the supported minimum height, and scopes the behavior to the mounted onboarding surface.

Independent verification on a clean detached worktree:
- Biome check passed for all three changed files.
- `npm --workspace @maka/desktop run build:with-deps` passed.
- `npm --workspace @maka/desktop run typecheck` passed.
- Targeted real Electron Playwright suite passed: 2/2, including wheel-driven scrolling at 320px height.
- All required GitHub checks are green; the PR is mergeable and clean.

Optimality: this is the smallest maintainable correction for the current ownership boundary. It adds no runtime state, public API, configuration, dependency, or parallel layout path. I do not recommend deleting any production code or either regression test, and a deeper component refactor is not warranted unless more full-viewport chat states appear.

Residual risk: the scoped CSS necessarily depends on Astryx ChatMessageList/ChatLayout DOM structure. The real Electron geometry and user-scroll assertions provide appropriate regression coverage for that contract.

Ready to merge.

Astro-Han:
Thanks for the fix — the phantom scrollbar is real (the empty-state flex spacer in `ChatMessageList` pushes onboarding content down, and the surface's `min-height` self-sizing combined with the spacer means content that fits still scrolls at 952px → 1159px), and the scoping via `:has(.maka-onboarding-surface)` is clean (no pollution of normal conversations; the spacer selector `> div > [aria-hidden]:has(+ :last-child)` precisely matches and can't hit the sentinel's sibling). At default sizes (1240×820) the e2e assertion `scrollHeight == clientHeight` holds and there's no regression. CI is green.

**Conclusion: PASS with one P2 (reached at the repo's supported minimum window) plus P3s.**

**P2 — at the supported minimum height (480×320) the onboarding content is hard-clipped and unreachable — an accessibility regression relative to before.** The three-layer `overflow` chain (root `overflow-y: hidden` → first child `overflow: hidden` → surface `height:100%; min-height:0; overflow:hidden`) leaves no scroll path anywhere, but the card is ~590-660px tall (`onboarding-hero.tsx` four-row provider list + buttons, wordmark + card + skip) while a 320px window leaves only ~240-260px of chat viewport — the lower half (Browse providers / connect / Skip) is cut off with no way to reach it. Before this PR the root scrollbar could reach the clipped content (surface had no overflow constraint, messageArea `flex: 1 0 auto` doesn't shrink), so this PR introduces the regression. `@media (max-height: 720px)` only changes `align-items: start`, providing no internal scrolling fallback. The repo explicitly maintains 320px as a tested contract (`window-state.ts:31-37` `SAFE_MIN_HEIGHT`, `main-window.ts:311-318` minHeight clamp) — Qodo flagged the same point. Suggested fix: keep the outer containment, change the surface to `overflow-y: auto` (scroll internally when content doesn't fit), and add a 480×320 e2e (the fixture's size env is only read when a scenario is set — `main-window.ts:600-616`); the existing `cardBottom <= viewportBottom` assertion is already the right test, it just lacks size coverage.

P3 (optional): the commit's `Generated-by: Codex` trailer is escaped as a literal `\n` in the commit message (not a real trailer line) — the repo's Ai Use Disclosure check warns on this; worth rewriting the trailer properly at squash; the e2e only covers the default 1240×820 (which is why the P2 slipped through); the structural selector depends on Astryx DOM ordering (the spacer being last-child's previous sibling) — commented and e2e-guarded, fine, but fragile to upstream empty-state restructuring.

---

*AI-assisted review disclosure: this review was produced with AI assistance (pi review subagent on `opencode-go/deepseek-v4-flash`). The subagent traced the spacer/ChatLayout geometry and the three-layer overflow chain from the PR head source; the P2 is a geometry calculation (590-660px card vs 240-260px viewport), not an observed render. Please weigh these findings with your own judgment.*

<details>

结论：PASS（1 个 P2 + P3）。问题真实（ChatMessageList 空态前恒有 aria-hidden flex spacer 把内容推到底部，onboarding surface 的 min-height 自撑与其叠加——内容放得下却有页面滚动条，952px 视口→1159px 滚动范围）；:has(.maka-onboarding-surface) 作用域干净不污染普通对话，spacer 选择器精确匹配不会误伤 sentinel。默认尺寸（1240×820）下 e2e 断言 scrollHeight==clientHeight 成立无回归。CI 全绿。P2：仓库支持的最小高度（480×320）下 onboarding 内容被三层 overflow hidden 链硬裁剪且不可达——卡片固有 ~590-660px（onboarding-hero 四行 provider 列表+双按钮、wordmark+Card+skip），320px 窗口聊天视口仅 ~240-260px，下半部分（Browse providers/连接/Skip）被裁掉且全链无任何滚动路径；PR 之前 root 滚动能到达被裁内容（surface 无 overflow 约束、messageArea flex:1 0 auto 不收缩），所以这是本 PR 引入的可达性回归。@media (max-height:720px) 只改 align-items:start 无内部滚动兜底。仓库显式把 320px 维护为受测契约（window-state.ts:31-37 SAFE_MIN_HEIGHT）。建议：保留外层 contain、surface 改 overflow-y:auto（放不下时内部滚动），并补 480×320 e2e（fixture 尺寸 env 只在设 scenario 时读取，main-window.ts:600-616）；现有 cardBottom<=viewportBottom 断言正好是现成测试，缺的只是尺寸覆盖。P3（可选）：commit 的 Generated-by: Codex trailer 被写成字面 \n 转义（非真实 trailer 行，Ai Use Disclosure 检查已 warning，squash 时应重写）；e2e 只覆盖默认 1240×820（正因如此 P2 得以通过 CI）；结构性选择器依赖 Astryx DOM 顺序（spacer 需是 last-child 前一兄弟），已注释+e2e 兜底但脆弱。
</details>

Astro-Han (approved):
LGTM

Astro-Han (approved):
Thanks for carrying the short-viewport fix through to the final head. I reviewed exact head `ba16360dd791e8952eab5d43be7ecb57a3b14723`.

The problem definition is correct: the onboarding empty-state spacer and viewport-sized surface created an outer scroll authority that did not belong in the chat layout. The current solution stays at the nearest existing seam: onboarding-scoped CSS removes the spacer/dock contribution, keeps the outer layout contained, and gives the onboarding surface its own `overflow-y: auto` fallback. That is simpler than threading a second layout mode through shared UI components.

The final 480×320 regression now proves that the content remains reachable through internal scrolling, while the normal-size test protects the original phantom-scrollbar fix. Current-head checks are green, and the PR includes visual-inspection evidence.

**Conclusion: APPROVE — no P0–P2 findings.**

Unverified, non-blocking gap: the minimum-height case uses Playwright viewport sizing rather than a separate native `BrowserWindow` resize/zoom matrix; the current Electron coverage and geometry assertions are sufficient for this localized change.

*AI-assisted review disclosure: Codex coordinated an independent `@reviewer` pass and a read-only `ollama-cloud/deepseek-v4-flash:high` pass. The review traced the ChatLayout/ChatMessageList DOM and overflow chain, checked the final short-viewport fix and regression tests, and refreshed current-head CI/review state. No local Electron run was performed in this final pass. A human contributor remains responsible for the final merge decision.*

<details>
<summary>中文摘要</summary>

已复评最终 head。问题根因是 onboarding 空态 spacer 与视口高度共同制造了不应存在的外层滚动；当前方案在 onboarding 现有 CSS seam 内解决，没有新增并行布局状态。最终版本使用 surface 内部滚动，480×320 回归测试证明最小高度下内容仍可达，正常尺寸测试继续保护 phantom scrollbar 修复；当前 checks 全绿且 PR 已提供视觉检查证据。没有 P0–P2，结论：APPROVE。

非阻断缺口：最小高度用的是 Playwright viewport，而不是额外的原生窗口 resize/zoom 矩阵。

</details>

PR #3145, opened 2026-08-17
feat(runtime-host): add remote project registration

M4n5ter:
<details open>
<summary>English</summary>

## Summary

Remote Desktop targets could select only Projects already registered on the Runtime Host. This PR adds a constrained Host-side directory picker so users can browse folders published by that Host and register one through the existing Project Catalog authority.

The service user's Home is published by default. Operators can replace it with up to eight named `--project-root <label>=<absolute-path>` entries. Remote browsing uses opaque root IDs and validated path segments; Local keeps the native OS picker.

Refs #2522

## Verification

- `npm run lint` — passed
- `npm run format:check` — passed
- `npm run typecheck` — passed across all workspaces
- `npm run astryx:surface-inventory` — passed with 188 aligned surfaces and 0 blockers
- affected Runtime Host, CLI, and Desktop tests — passed
- `npm test` — 951/952 passed; the existing Candidate-arbitration timeout passed when rerun in isolation
- exercised the Desktop flow against a real remote Runtime Host over SSH: switch between two published roots, toggle hidden folders, navigate, register a Project, and select it for a new task

## AI use

- [ ] No generative tool made a substantive contribution
- [x] Generative tooling made a substantive contribution

Tool(s) and scope: OpenAI Codex contributed implementation, tests, analysis, and remote Desktop verification under maintainer direction

## Checklist

- [x] Tests cover the change and fail without it
- [x] Lint, format, typecheck and the affected suites pass locally

Does this PR entail a change in behavior?

- [x] Yes — described under Summary above
- [ ] No

</details>

<details>
<summary>简体中文</summary>

## 摘要

此前，Desktop 的远程目标只能选择已经在 Runtime Host 注册的 Project。本 PR 增加受限的 Host-side 目录选择器，让用户浏览该 Host 发布的文件夹，并通过现有 Project Catalog authority 注册所选目录。

默认发布运行服务用户的 Home。Operator 可以用最多八个命名的 `--project-root <label>=<absolute-path>` 替换默认范围。远程浏览只使用不透明 root ID 和经过校验的路径分段；Local 继续使用系统原生选择器。

关联 #2522

## 验证

- `npm run lint` — 通过
- `npm run format:check` — 通过
- `npm run typecheck` — 所有 workspace 通过
- `npm run astryx:surface-inventory` — 188 个 surface 全部 aligned，0 blocker
- 受影响的 Runtime Host、CLI 与 Desktop 测试 — 通过
[...]

PR #3199, opened 2026-08-18
feat(goal): let the user arm a Goal from the composer

Astro-Han:
## Summary

A Goal could only be armed by the model, from inside a Turn, with the GoalSet tool. There was no Host operation for it, no bridge method, and no control anywhere in the product — the user could stop a Goal but never start one.

This adds `goal.arm` as a real Host operation, next to `goal.query` and `goal.control`, and carries it out to the ＋ menu:

- **`HostGoalCoordinator#arm`** creates the Goal through the same `GoalManager` the tool uses, under the same Session admission gate, with the same durable record. It schedules nothing: a Goal armed outside a Turn takes hold on the next one, which `beginObservedTurn` binds to the live control lease. A Session that already has an unfinished Goal gets `operation_conflict`.
- **One authority for the bounds.** The two budget ceilings and the token floor move to `@maka/core/goal`, so the protocol codec and the GoalSet tool schema validate against one number instead of two copies of it.
- **Remote owners are granted `goal.arm`**, at the same tier as `goal.control`. Withholding it would withhold nothing: a remote owner sends Turns, and the model arms its own Goal inside one. The only thing a refusal removes is the explicit path the user can see and stop.
- **The IPC handler takes the Session from the scoped channel**, never from the renderer's frame, and normalizes the budgets without clamping them — an out-of-range value is refused once, by the Host, rather than quietly becoming a Goal the user did not ask for.
- **The ＋ menu gains one "设定 Goal…" row** in its action group, at the same 28px rhythm as the rest, and app-shell opens a dialog that collects the condition and the two budgets. The row explains itself when a Goal is already running (arming would conflict) or a Turn is in flight (the Goal would only take hold on the next one).

> **Stacked on #3198.** That PR restructured the same menu, and this row belongs in the action group it establishes. Both branches live on the same fork, so GitHub cannot base this one on that branch directly — it is opened against `main` and its diff therefore contains #3198's commit as well. **Review only the second commit** (`feat(goal): let the user arm a Goal from the composer`), and merge #3198 first.

### What it looks like

Shot live in Storybook, same viewport, each surface cropped to its own bounds and composited at 1:1 pixel scale. (The mode rows below the divider are #3198's doing, not this PR's.)

image

Left: 设定 Goal… joins 添加文件或目录 and 选择技能 in the action group — the same 28px row, no new component kind in the panel. Middle: arming refuses a second Goal, so the row says why instead of opening a dialog that would fail on submit; it is the same mechanism the Skills row already uses for an empty catalog, and the only row that grows past 28px is the one carrying an explanation. Right: the two budgets that stop a Goal are visible while it is being armed, the condition counter is the shared `GOAL_CONDITION_TEXT_LIMIT`, and 开始 stays disabled until there is a condition.

## Verification

- `packages/runtime-host`: `goal-protocol.test.ts` (+3 tests — exact frame, budget bounds, remote-owner grant) and `goal-coordinator.test.ts` (+1 — arm, conflict, not-found, archived) pass.
- `apps/desktop`: `runtime-host-client-operations.test.ts` (+1 — one request, no retry on conflict) and `runtime-host-session-domains-ipc-main.test.ts` (+1 — scoped Session wins over the frame's, budget normalization) pass.
- `packages/runtime` `goal-tools.test.ts` / `goal-state.test.ts` and `packages/core` `goal-authority.test.ts` pass unchanged, confirming the constant move is behaviour-preserving.
- Typecheck: `@maka/core`, `@maka/runtime`, `@maka/runtime-host`, `@maka/ui`, and all four `apps/desktop` tsconfig projects. `npm run format:check` and `biome lint` clean.
- Storybook: `GoalDialogOpen` and `GoalAlreadySet` added. Measured live — the seven ＋ rows stay 28px, and the disabled Goal row grows to 48px only because it carries its explanation, the same way the Skills row already does with an empty catalog.
- Not run: repository-wide suites (left to CI); no E2E added.

## AI use

Tool(s) and scope: Claude Code (Opus) wrote the diff, the tests, and the commit message. The decision to add a real Host operation rather than route the user through the model is the human contributor's. The commit carries a `Generated-by: Claude Code` trailer.

## Checklist

### Review focus

Runtime Host is the sole execution authority, and this adds an operation to it. Two things deserve the attention:

1. **`#arm` schedules nothing.** A Goal armed while a Turn is in flight does not drive that Turn — it registered before the Goal existed, and `activateGoal` declines exactly that case (`goal_not_observed`). The UI disables the entry during a Turn so this is never a silent no-op, but the Host accepts it either way; the Goal simply starts from the next Turn.
2. **The remote-owner grant.** `REMOTE_OWNER_OPERATION_GRANTS` is fail-closed by design, and this PR opens one entry in it. The reasoning is in the test that asserts it; if you disagree, the grant is the one line to revert.

### DIFF
--- a/apps/desktop/src/main/runtime-host-client.ts
+++ b/apps/desktop/src/main/runtime-host-client.ts
@@ -1189,12 +1189,7 @@ export class DesktopRuntimeHostClient {
    * that already has an unfinished Goal fails with `operation_conflict`, and
    * that is an answer for the user, not a race to re-run.
    */
-  armGoal(input: {
-    sessionId: string;
-    condition: string;
-    maxIterations: number | null;
-    tokenBudget: number | null;
-  }): Promise<OperationOutput<"goal.arm">> {
+  armGoal(input: OperationInput<"goal.arm">): Promise<OperationOutput<"goal.arm">> {
     return this.request("goal.arm", input);
   }
 
--- a/apps/desktop/src/main/runtime-host-client.ts
+++ b/apps/desktop/src/main/runtime-host-client.ts
@@ -1183,6 +1183,21 @@ export class DesktopRuntimeHostClient {
     return this.request("goal.query", { sessionId });
   }
 
+  /**
+   * Arm a Goal the user asked for. No optimistic retry loop like `clearGoal`:
+   * arming names no revision, so there is no stale one to refresh — a Session
+   * that already has an unfinished Goal fails with `operation_conflict`, and
+   * that is an answer for the user, not a race to re-run.
+   */
+  armGoal(input: {
+    sessionId: string;
+    condition: string;
+    maxIterations: number | null;
+    tokenBudget: number | null;
+  }): Promise<OperationOutput<"goal.arm">> {
+    return this.request("goal.arm", input);
+  }
+
   controlGoal(
     goal: Pick<GoalProjection, "sessionId" | "goalId" | "revision">,
     action: GoalControlAction,
--- a/apps/desktop/src/preload/preload.ts
+++ b/apps/desktop/src/preload/preload.ts
@@ -170,6 +170,7 @@ import {
   requireDesktopTargetScope,
   type DesktopTargetScope,
 } from '../shared/runtime-host-identity.js';
+import type { GoalArmRequest } from '../shared/goal-arm.js';
 import {
   projectDesktopAttachmentRefs,
   projectDesktopDailyReviewSummary,
@@ -1910,13 +1911,8 @@ const makaBridge = {
     get(sessionId: string): Promise<GoalState | null> {
       return invokeProjectedSessionRuntimeHost('goal:get', sessionId);
     },
-    arm(input: {
-      sessionId: string;
-      condition: string;
-      maxIterations?: number | null;
-      tokenBudget?: number | null;
-    }): Promise<GoalState> {
-      return invokeProjectedSessionRuntimeHost('goal:arm', input.sessionId, input);
+    arm(sessionId: string, goal: GoalArmRequest): Promise<GoalState> {
+      return invokeProjectedSessionRuntimeHost('goal:arm', sessionId, goal);
     },
     clear(sessionId: string): Promise<void> {
       return invokeSessionRuntimeHost('goal:clear', sessionId);
--- a/apps/desktop/src/preload/preload.ts
+++ b/apps/desktop/src/preload/preload.ts
@@ -1910,6 +1910,14 @@ const makaBridge = {
     get(sessionId: string): Promise<GoalState | null> {
       return invokeProjectedSessionRuntimeHost('goal:get', sessionId);
     },
+    arm(input: {
+      sessionId: string;
+      condition: string;
+      maxIterations?: number | null;
+      tokenBudget?: number | null;
+    }): Promise<GoalState> {
+      return invokeProjectedSessionRuntimeHost('goal:arm', input.sessionId, input);
+    },
     clear(sessionId: string): Promise<void> {
       return invokeSessionRuntimeHost('goal:clear', sessionId);
     },
--- a/apps/desktop/src/renderer/app-shell.tsx
+++ b/apps/desktop/src/renderer/app-shell.tsx
@@ -227,6 +227,7 @@ function rebaseWorkspaceFileReferences(
 
 import { useSettingsModal } from './use-settings-modal';
 import { RemoteProjectDirectoryDialog } from './remote-project-directory-dialog';
+import { GoalDialog } from './goal-dialog.js';
 import { useSystemUiLocale } from './use-system-ui-locale';
 import {
   isSessionWorkspaceUnavailableError,
@@ -776,6 +777,12 @@ function AppShellContent({
     },
     [],
   );
+  /**
+   * The Session the Goal dialog is arming, or `undefined` when it is closed.
+   * Keyed by Session rather than a boolean so switching Sessions while the
+   * dialog is open can never arm the wrong one.
+   */
+  const [goalDialogSessionId, setGoalDialogSessionId] = useState<string>();
   // Set of session ids whose backend / connection is no longer usable —
   // drives the sidebar "已过期" pill (PR108g, paired with the PR108e chat
   // header banner). Derivation is pure (see `stale-sessions.ts`) so the
@@ -3271,6 +3278,17 @@ function AppShellContent({
                   onOrchestrationModeChange={(mode) => {
                     void setOrchestrationMode(mode);
                   }}
+                  onSetGoal={
+                    activeId && activeBoundarySurface.localInteractionAvailable
+                      ? () => setGoalDialogSessionId(activeId)
+                      : undefined
+                  }
+                  goalActive={activeGoal !== null}
+                  goalDisabledReason={
+                    activeStreamingLive || (activeId && turnActive)
+                      ? shellCopy.goalTurnActive
+                      : undefined
+                  }
                     />
                   </>
                 }
@@ -3586,6 +3604,10 @@ function AppShellContent({
         }}
       />
 
+      <GoalDialog
+        {...(goalDialogSessionId ? { sessionId: goalDialogSessionId } : {})}
+        onClose={() => setGoalDialogSessionId(undefined)}
+      />
       <RuntimeHostSshTerminalDialog />
 
       <RemoteProjectDirectoryDialog
--- a/apps/desktop/src/renderer/goal-dialog.tsx
+++ b/apps/desktop/src/renderer/goal-dialog.tsx
@@ -19,7 +19,11 @@ import { NumberInput } from '@astryxdesign/core/NumberInput';
 import { Text } from '@astryxdesign/core/Text';
 import { TextArea } from '@astryxdesign/core/TextArea';
 import { useUiLocale } from '@maka/ui';
-import { GOAL_CONDITION_TEXT_LIMIT, GOAL_MAX_ITERATIONS_LIMIT } from '@maka/core/goal';
+import {
+  GOAL_CONDITION_TEXT_LIMIT,
+  GOAL_MAX_ITERATIONS_LIMIT,
+  GOAL_TOKEN_BUDGET_MINIMUM,
+} from '@maka/core/goal';
 import { getShellCopy, localizedShellErrorMessage } from './locales/shell-copy.js';
 
 export function GoalDialog(props: {
@@ -54,8 +58,7 @@ export function GoalDialog(props: {
     setArming(true);
     setError(undefined);
     try {
-      await window.maka.goal.arm({
-        sessionId,
+      await window.maka.goal.arm(sessionId, {
         condition: condition.trim(),
         maxIterations,
         tokenBudget,
@@ -115,7 +118,7 @@ export function GoalDialog(props: {
                   description={copy.tokenBudgetDescription}
                   value={tokenBudget}
                   onChange={setTokenBudget}
-                  min={1_000}
+                  min={GOAL_TOKEN_BUDGET_MINIMUM}
                   step={1_000}
                   isOptional
                   isDisabled={arming}
--- a/apps/desktop/src/renderer/goal-dialog.tsx
+++ b/apps/desktop/src/renderer/goal-dialog.tsx
@@ -1,0 +1,149 @@
+/**
+ * The form behind the ＋ menu's "Set a goal…" entry.
+ *
+ * A Goal is the one control here that keeps spending tokens after the user
+ * stops typing, so it is a dialog rather than a menu row: the condition needs
+ * a sentence, and the two budgets that stop it should be visible while it is
+ * being armed, not discovered afterwards.
+ *
+ * Bounds are the Host's. This sends what the user chose and reports what the
+ * Host answers; it does not clamp a rejected number into an accepted one,
+ * because the Goal that then runs would not be the one they asked for.
+ */
+import { useEffect, useState } from 'react';
+import { Button } from '@astryxdesign/core/Button';
+import { Dialog, DialogHeader } from '@astryxdesign/core/Dialog';
+import { Layout, LayoutContent, LayoutFooter } from '@astryxdesign/core/Layout';
+import { HStack, VStack } from '@astryxdesign/core/Stack';
+import { NumberInput } from '@astryxdesign/core/NumberInput';
+import { Text } from '@astryxdesign/core/Text';
+import { TextArea } from '@astryxdesign/core/TextArea';
+import { useUiLocale } from '@maka/ui';
+import { GOAL_CONDITION_TEXT_LIMIT, GOAL_MAX_ITERATIONS_LIMIT } from '@maka/core/goal';
+import { getShellCopy, localizedShellErrorMessage } from './locales/shell-copy.js';
+
+export function GoalDialog(props: {
+  /** The Session to arm. `undefined` closes the dialog. */
+  sessionId?: string;
+  onClose(): void;
+}) {
+  const locale = useUiLocale();
+  const copy = getShellCopy(locale).goalDialog;
+  const [condition, setCondition] = useState('');
+  const [maxIterations, setMaxIterations] = useState<number | null>(null);
+  const [tokenBudget, setTokenBudget] = useState<number | null>(null);
+  const [arming, setArming] = useState(false);
+  const [error, setError] = useState<string>();
+
+  // Each opening starts from empty: the previous Session's condition is not a
+  // useful default for this one, and a stale error would outlive its cause.
+  useEffect(() => {
+    if (props.sessionId === undefined) return;
+    setCondition('');
+    setMaxIterations(null);
+    setTokenBudget(null);
+    setError(undefined);
+    setArming(false);
+  }, [props.sessionId]);
+
+  const sessionId = props.sessionId;
+  const canSubmit = condition.trim().length > 0 && !arming;
+
+  async function arm(): Promise<void> {
+    if (!sessionId || !canSubmit) return;
+    setArming(true);
+    setError(undefined);
+    try {
+      await window.maka.goal.arm({
+        sessionId,
+        condition: condition.trim(),
+        maxIterations,
+        tokenBudget,
+      });
+      props.onClose();
+    } catch (cause) {
+      setError(localizedShellErrorMessage(cause, copy.failedFallback, locale));
+    } finally {
+      setArming(false);
+    }
+  }
+
+  return (
+    <Dialog
+      isOpen={sessionId !== undefined}
+      onOpenChange={(open) => {
+        if (!open && !arming) props.onClose();
+      }}
+      purpose="form"
+      width={520}
+      className="goalDialog"
+    >
+      <Layout
+        header={<DialogHeader title={copy.title} onOpenChange={(open) => {
+          if (!open && !arming) props.onClose();
+        }} />}
+        content={
+          <LayoutContent padding={4}>
+            <VStack gap={4}>
+              <Text type="body" color="secondary">{copy.description}</Text>
+              <TextArea
+                label={copy.conditionLabel}
+                description={copy.conditionDescription}
+                placeholder={copy.conditionPlaceholder}
+                value={condition}
+                onChange={setCondition}
+                rows={3}
+                maxLength={GOAL_CONDITION_TEXT_LIMIT.codeUnits}
+                isRequired
+                isDisabled={arming}
+                hasAutoFocus
+                {...(error ? { status: { type: 'error' as const, message: error } } : {})}
+              />
+              <HStack gap={3}>
+                <NumberInput
+                  label={copy.maxIterationsLabel}
+                  description={copy.maxIterationsDescription}
+                  value={maxIterations}
+                  onChange={setMaxIterations}
+                  min={1}
+                  max={GOAL_MAX_ITERATIONS_LIMIT}
+                  isOptional
+                  isDisabled={arming}
+                />
+                <NumberInput
+                  label={copy.tokenBudgetLabel}
+                  description={copy.tokenBudgetDescription}
+                  value={tokenBudget}
+                  onChange={setTokenBudget}
+                  min={1_000}
+                  step={1_000}
+                  isOptional
+                  isDisabled={arming}
+                />
+              </HStack>
+            </VStack>
+          </LayoutContent>
+        }
+        footer={
+          <LayoutFooter hasDivider>
+            <HStack gap={2} hAlign="end">
+              <Button
+                variant="ghost"
+                label={copy.cancel}
+                isDisabled={arming}
+                onClick={props.onClose}
+              />
+              <Button
+                variant="primary"
+                label={copy.submit}
+                isDisabled={!canSubmit}
+                isLoading={arming}
+                onClick={() => void arm()}
+              />
+            </HStack>
+          </LayoutFooter>
+        }
+      />
+    </Dialog>
+  );
+}
--- a/apps/desktop/src/renderer/goal-dialog.tsx
+++ b/apps/desktop/src/renderer/goal-dialog.tsx
@@ -8,16 +8,20 @@
  *
  * Bounds are the Host's. This sends what the user chose and reports what the
  * Host answers; it does not clamp a rejected number into an accepted one,
- * because the Goal that then runs would not be the one they asked for.
+ * because the Goal that then runs would not be the one they asked for. For the
+ * same reason the budget fields hold the user's own text rather than a parsed
+ * number: a field that quietly keeps the last value it could parse can arm a
+ * budget the form is no longer showing, and a field that quietly keeps nothing
+ * can arm no budget at all where the user asked for a tight one.
  */
 import { useEffect, useState } from 'react';
 import { Button } from '@astryxdesign/core/Button';
 import { Dialog, DialogHeader } from '@astryxdesign/core/Dialog';
 import { Layout, LayoutContent, LayoutFooter } from '@astryxdesign/core/Layout';
 import { HStack, VStack } from '@astryxdesign/core/Stack';
-import { NumberInput } from '@astryxdesign/core/NumberInput';
 import { Text } from '@astryxdesign/core/Text';
 import { TextArea } from '@astryxdesign/core/TextArea';
+import { TextInput } from '@astryxdesign/core/TextInput';
 import { useUiLocale } from '@maka/ui';
 import {
   GOAL_CONDITION_TEXT_LIMIT,
@@ -25,6 +29,29 @@ import {
   GOAL_TOKEN_BUDGET_MINIMUM,
 } from '@maka/core/goal';
 import { getShellCopy, localizedShellErrorMessage } from './locales/shell-copy.js';
+
+type BudgetReading =
+  | { readonly kind: 'empty' }
+  | { readonly kind: 'value'; readonly value: number }
+  | { readonly kind: 'invalid' };
+
+/**
+ * Read one budget field, refusing anything the Host would not take.
+ *
+ * `min`/`max` repeat the operation's own bounds so the refusal can be shown
+ * beside the field instead of arriving as a rejected frame; the numbers
+ * themselves come from the shared Goal limits, so there is still one authority
+ * for what they are.
+ */
+function readGoalBudget(text: string, min: number, max?: number): BudgetReading {
+  const trimmed = text.trim();
+  if (!trimmed) return { kind: 'empty' };
+  if (!/^\d+$/.test(trimmed)) return { kind: 'invalid' };
+  const value = Number(trimmed);
+  if (!Number.isSafeInteger(value) || value < min) return { kind: 'invalid' };
+  if (max !== undefined && value > max) return { kind: 'invalid' };
+  return { kind: 'value', value };
+}
 
 export function GoalDialog(props: {
   /** The Session to arm. `undefined` closes the dialog. */
@@ -34,8 +61,8 @@ export function GoalDialog(props: {
   const locale = useUiLocale();
   const copy = getShellCopy(locale).goalDialog;
   const [condition, setCondition] = useState('');
-  const [maxIterations, setMaxIterations] = useState<number | null>(null);
-  const [tokenBudget, setTokenBudget] = useState<number | null>(null);
+  const [maxIterationsText, setMaxIterationsText] = useState('');
+  const [tokenBudgetText, setTokenBudgetText] = useState('');
   const [arming, setArming] = useState(false);
   const [error, setError] = useState<string>();
 
@@ -44,14 +71,20 @@ export function GoalDialog(props: {
   useEffect(() => {
     if (props.sessionId === undefined) return;
     setCondition('');
-    setMaxIterations(null);
-    setTokenBudget(null);
+    setMaxIterationsText('');
+    setTokenBudgetText('');
     setError(undefined);
     setArming(false);
   }, [props.sessionId]);
 
   const sessionId = props.sessionId;
-  const canSubmit = condition.trim().length > 0 && !arming;
+  const maxIterations = readGoalBudget(maxIterationsText, 1, GOAL_MAX_ITERATIONS_LIMIT);
+  const tokenBudget = readGoalBudget(tokenBudgetText, GOAL_TOKEN_BUDGET_MINIMUM);
+  const canSubmit =
+    condition.trim().length > 0 &&
+    maxIterations.kind !== 'invalid' &&
+    tokenBudget.kind !== 'invalid' &&
+    !arming;
 
   async function arm(): Promise<void> {
     if (!sessionId || !canSubmit) return;
@@ -60,8 +93,8 @@ export function GoalDialog(props: {
     try {
       await window.maka.goal.arm(sessionId, {
         condition: condition.trim(),
-        maxIterations,
-        tokenBudget,
+        maxIterations: maxIterations.kind === 'value' ? maxIterations.value : null,
+        tokenBudget: tokenBudget.kind === 'value' ? tokenBudget.value : null,
       });
       props.onClose();
     } catch (cause) {
@@ -103,25 +136,27 @@ export function GoalDialog(props: {
                 {...(error ? { status: { type: 'error' as const, message: error } } : {})}
               />
               <HStack gap={3}>
-                <NumberInput
+                <TextInput
                   label={copy.maxIterationsLabel}
                   description={copy.maxIterationsDescription}
-                  value={maxIterations}
-                  onChange={setMaxIterations}
-                  min={1}
-                  max={GOAL_MAX_ITERATIONS_LIMIT}
+                  value={maxIterationsText}
+                  onChange={setMaxIterationsText}
                   isOptional
                   isDisabled={arming}
+                  {...(maxIterations.kind === 'invalid'
+                    ? { status: { type: 'error' as const, message: copy.maxIterationsInvalid(GOAL_MAX_ITERATIONS_LIMIT) } }
+                    : {})}
                 />
-                <NumberInput
+                <TextInput
                   label={copy.tokenBudgetLabel}
                   description={copy.tokenBudgetDescription}
-                  value={tokenBudget}
-                  onChange={setTokenBudget}
-                  min={GOAL_TOKEN_BUDGET_MINIMUM}
-                  step={1_000}
+                  value={tokenBudgetText}
+                  onChange={setTokenBudgetText}
                   isOptional
                   isDisabled={arming}
+                  {...(tokenBudget.kind === 'invalid'
+                    ? { status: { type: 'error' as const, message: copy.tokenBudgetInvalid(GOAL_TOKEN_BUDGET_MINIMUM) } }
+                    : {})}
                 />
               </HStack>
             </VStack>
--- a/docs/astryx-surface-file-inventory.md
+++ b/docs/astryx-surface-file-inventory.md
@@ -5,7 +5,7 @@ Each row is one on-disk product surface file. Regenerated inventory must stay in
 
 Wiki bar: Design Conventions · API Use-the-System · Theming · Container Padding.
 
-**Totals:** 188 files — blocker 0, polish 0, aligned 188.
+**Totals:** 189 files — blocker 0, polish 0, aligned 189.
 
 ## Exclusions (explicit)
 
@@ -43,6 +43,7 @@ Wiki bar: Design Conventions · API Use-the-System · Theming · Container Paddi
 | `apps/desktop/src/renderer/command-palette.tsx` | dialog-overlay | EmptyState | aligned — uses Astryx (EmptyState) | aligned |
 | `apps/desktop/src/renderer/custom-pet-companion.tsx` | other | none | aligned — no raw controls; no Astryx JSX usage | aligned |
 | `apps/desktop/src/renderer/error-boundary.tsx` | other | Button, Card | aligned — uses Astryx (Button, Card) | aligned |
+| `apps/desktop/src/renderer/goal-dialog.tsx` | dialog-overlay | Button, Dialog, DialogHeader, HStack, Layout, LayoutContent, Text, VStack | aligned — uses Astryx (Button, Dialog, DialogHeader, HStack, Layout, LayoutContent, Text, VStack) | aligned |
 | `apps/desktop/src/renderer/keyboard-help.tsx` | dialog-overlay | Dialog, DialogHeader, Heading, Layout, LayoutContent | aligned — uses Astryx (Dialog, DialogHeader, Heading, Layout, LayoutContent) | aligned |
 | `apps/desktop/src/renderer/live-turn-reconciler.tsx` | other | none | aligned — no raw controls; no Astryx JSX usage | aligned |
 | `apps/desktop/src/renderer/maka-tokens.css` | styles | n/a (css) | aligned — no off-rhythm control heights flagged | aligned |
--- a/docs/astryx-surface-file-inventory.md
+++ b/docs/astryx-surface-file-inventory.md
@@ -43,7 +43,7 @@ Wiki bar: Design Conventions · API Use-the-System · Theming · Container Paddi
 | `apps/desktop/src/renderer/command-palette.tsx` | dialog-overlay | EmptyState | aligned — uses Astryx (EmptyState) | aligned |
 | `apps/desktop/src/renderer/custom-pet-companion.tsx` | other | none | aligned — no raw controls; no Astryx JSX usage | aligned |
 | `apps/desktop/src/renderer/error-boundary.tsx` | other | Button, Card | aligned — uses Astryx (Button, Card) | aligned |
-| `apps/desktop/src/renderer/goal-dialog.tsx` | dialog-overlay | Button, Dialog, DialogHeader, HStack, Layout, LayoutContent, Text, VStack | aligned — uses Astryx (Button, Dialog, DialogHeader, HStack, Layout, LayoutContent, Text, VStack) | aligned |
+| `apps/desktop/src/renderer/goal-dialog.tsx` | dialog-overlay | Button, Dialog, DialogHeader, HStack, Layout, LayoutContent, Text, TextInput, VStack | aligned — uses Astryx (Button, Dialog, DialogHeader, HStack, Layout, LayoutContent, Text, TextInput) | aligned |
 | `apps/desktop/src/renderer/keyboard-help.tsx` | dialog-overlay | Dialog, DialogHeader, Heading, Layout, LayoutContent | aligned — uses Astryx (Dialog, DialogHeader, Heading, Layout, LayoutContent) | aligned |
 | `apps/desktop/src/renderer/live-turn-reconciler.tsx` | other | none | aligned — no raw controls; no Astryx JSX usage | aligned |
 | `apps/desktop/src/renderer/maka-tokens.css` | styles | n/a (css) | aligned — no off-rhythm control heights flagged | aligned |
--- a/packages/runtime-host/src/protocol/goal.ts
+++ b/packages/runtime-host/src/protocol/goal.ts
@@ -1,6 +1,8 @@
 import {
   GOAL_CONDITION_TEXT_LIMIT,
+  GOAL_MAX_ITERATIONS_LIMIT,
   GOAL_REASON_TEXT_LIMIT,
+  GOAL_TOKEN_BUDGET_MINIMUM,
   isGoalStatus,
   type GoalStatus,
 } from '@maka/core/goal';
@@ -24,6 +26,7 @@ const QUERY_ERRORS = [
   'internal_failure',
 ] as const;
 const CONTROL_ERRORS = [...QUERY_ERRORS, 'session_archived', 'operation_conflict'] as const;
+const ARM_ERRORS = CONTROL_ERRORS;
 
 export type GoalControlAction = 'pause' | 'resume' | 'clear';
 
@@ -54,6 +57,31 @@ export interface GoalQueryResult {
   readonly goal: GoalProjection | null;
 }
 
+/**
+ * Arm a Goal for a Session from outside a Turn.
+ *
+ * The model arms its own Goal with the GoalSet tool, inside the Turn that owns
+ * the control lease. A user arming one has no Turn to speak from, so this is
+ * the Host's own entry point for it — same authority, same durable record, no
+ * second implementation of what a Goal is.
+ *
+ * `maxIterations` and `tokenBudget` are nullable rather than optional: the
+ * frame carries an exact key set, and "the caller did not choose" has to be a
+ * value on the wire rather than an absent field. Null means the Goal takes the
+ * Runtime default (and, for the budget, no budget at all).
+ */
+export interface GoalArmInput {
+  readonly sessionId: string;
+  readonly condition: string;
+  readonly maxIterations: number | null;
+  readonly tokenBudget: number | null;
+}
+
+export interface GoalArmResult {
+  readonly sessionId: string;
+  readonly goal: GoalProjection;
+}
+
 export interface GoalControlInput {
   readonly sessionId: string;
   readonly goalId: string;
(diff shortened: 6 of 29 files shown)


apps/desktop/src/main/runtime-host-client.ts
```diff
@@ -1185,6 +1185,21 @@ export class DesktopRuntimeHostClient {
...
+   * arming names no revision, so there is no stale one to refresh — a Session
+   * that already has an unfinished Goal fails with `operation_conflict`, and
+   * that is an answer for the user, not a race to re-run.
+   */
+  armGoal(input: {
+    sessionId: string;
+    condition: string;
+    maxIterations: number | null;
+    tokenBudget: number | null;
+  }): Promise<OperationOutput<"goal.arm">> {
```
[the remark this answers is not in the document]
Astro-Han: Agreed and fixed in 974a40734.

`armGoal(input: OperationInput<"goal.arm">)`. The local type was a copy of the operation input that could drift from it silently — the compiler would have kept accepting it after the protocol changed shape. Every sibling on this client already spells its input this way.

apps/desktop/src/renderer/goal-dialog.tsx
```diff
@@ -0,0 +1,149 @@
...
+                <NumberInput
+                  label={copy.tokenBudgetLabel}
+                  description={copy.tokenBudgetDescription}
+                  value={tokenBudget}
+                  onChange={setTokenBudget}
+                  min={1_000}
+                  step={1_000}
+                  isOptional
+                  isDisabled={arming}
+                />
```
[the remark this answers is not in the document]
Astro-Han: Agreed and fixed in 974a40734.

`min={GOAL_TOKEN_BUDGET_MINIMUM}`, and no `max` — you are right that no shared ceiling exists, and there should not be one: `tokenBudget` is a stopping condition, so omitting it means no token stop at all. A ceiling here would make the dialog stricter than leaving the field empty.

This was my own inconsistency: the constant is introduced by this PR precisely so the tool and the protocol validate one number, and then the dialog restated it as a literal.

apps/desktop/src/preload/preload.ts
```diff
@@ -1875,6 +1872,14 @@ const makaBridge = {
...
       return invokeProjectedSessionRuntimeHost('goal:get', sessionId);
     },
+    arm(input: {
+      sessionId: string;
+      condition: string;
+      maxIterations?: number | null;
+      tokenBudget?: number | null;
+    }): Promise<GoalState> {
+      return invokeProjectedSessionRuntimeHost('goal:arm', input.sessionId, input);
+    },
```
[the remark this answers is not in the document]
Astro-Han: Confirmed, and this was mine to catch — fixed in ed73ad723.

You are right about the effect: my previous commit made the validator refuse `sessionId`, and preload was passing the whole `input` including it, so every arm from the dialog would have failed. My IPC test invoked the channel directly, so it exercised the validator and never the caller.

The fix is not a strip, because a strip leaves the two sides free to disagree again. Preload declared the payload shape and the validator declared the key list, and nothing tied them together. Both now read `shared/goal-arm.ts`, preload takes the Session as its own parameter like every sibling on this bridge, and putting `sessionId` back in the dialog's literal is a compile error (TS2353) rather than a runtime rejection. Verified both ways.

apps/desktop/src/renderer/app-shell.tsx
```diff
@@ -3208,58 +3201,37 @@ function AppShellContent({
...
+                    activeId && activeBoundarySurface.localInteractionAvailable
+                      ? () => setGoalDialogSessionId(activeId)
+                      : undefined
+                  }
+                  goalActive={activeGoal !== null}
+                  goalDisabledReason={
+                    activeStreamingLive || (activeId && turnActive)
+                      ? shellCopy.goalTurnActive
+                      : undefined
                   }
```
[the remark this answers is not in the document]
Astro-Han: Confirmed and fixed in 3e6bbc3fd. `setGoal(null)` now runs with the effect, before the first fetch, so the state resets with the Session instead of with the response.

apps/desktop/src/main/__tests__/session-mode-ipc-main.test.ts
```diff
@@ -0,0 +1,119 @@
...
+    client: {
+      async updateSessionConfiguration(sessionId: string, patch: DesktopSessionConfigurationPatch) {
+        patches.push(patch);
+        return projection(sessionId);
+      },
+    },
+    resolveCreateProject: async () => ({}),
+    emitSessionsChanged() {},
+    releaseSessionResources() {},
+    sessionCopyCleanup: { recover: async () => ({ cleaned: [] }) },
```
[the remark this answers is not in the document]
Astro-Han: Confirmed and fixed on the base branch in c05d4c78f (this file belongs to #3198). The stub returns `{ cleaned: [], failed: [] }`.

Your reading of why it passes today is right, which is the reason to fix it: the rejection is swallowed, so the stub would surface as an unrelated failure in whatever test touched `sessions:list` next.

packages/ui/src/__tests__/composer-plus-menu.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+const base = {
+  onSend: () => undefined,
+  onStop: () => undefined,
+  sessionMode: 'default' as const,
+  onSessionModeChange: () => undefined,
+};
+
+test('the mode choice alone opens the menu on a row, not on a rule', () => {
+  assert.equal(plusMenu(base).includes('astryx-dropdown-menu-divider'), false);
+});
```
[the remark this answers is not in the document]
Astro-Han: Confirmed and fixed on the base branch in c05d4c78f (this file belongs to #3198). `plusMenu` now asserts the marker before slicing. Verified: renaming the class in the built composer turns both tests red, where before it left them green.

On the divider class — agreed that it couples the test to an Astryx version, and I am leaving it. The behaviour under test is "is there a rule at the top of this menu", and the rendered class is the only signal the SSR markup carries for it. A rename would fail loudly and be a one-line update, which I prefer to asserting something weaker.

packages/runtime-host/src/server/goal-coordinator.ts
```diff
@@ -306,6 +320,49 @@ export class HostGoalCoordinator {
...
+      try {
+        header = await this.#stores.sessionStore.readHeaderSnapshot(input.sessionId);
+      } catch (error) {
+        if (isSessionNotFoundError(error)) return notFound('Session does not exist');
+        throw error;
+      }
+      if (header.isArchived) {
+        return sessionArchived('Archived Session cannot be given a Goal');
+      }
+      const created = this.manager.create(input.sessionId, input.condition, {
```
M4n5ter, line 367: <details open>

**Blocking — Persist the “waiting for first observed Turn” phase**

<code>#arm</code> persists an <code>active</code> Goal with <code>currentExecution: null</code>, which is indistinguishable from a Goal waiting for its next continuation. If the Host stops after arming but before the user sends another Turn, <code>recover()</code> passes this record to <code>recoverActiveGoal()</code>, which immediately creates a continuation intent and schedules it. The Goal therefore starts autonomous work and token spending after restart even though the public contract explicitly says that arming alone starts nothing.

Please persist an explicit pre-activation phase or marker and only recover Goals that have already been activated by an observed Turn. A regression test should cover arm → close/reopen → recover, verify that no Turn is admitted before the user’s next Turn, and separately verify that an already activated Goal still resumes correctly.

</details>

<details>

**阻塞问题——持久化“等待首个 observed Turn”的阶段**

<code>#arm</code> 持久化的是 <code>active</code> 且 <code>currentExecution: null</code> 的 Goal，这与等待下一次 continuation 的 Goal 无法区分。如果 Host 在设置 Goal 后、用户发送下一条 Turn 前停止，<code>recover()</code> 会把该记录交给 <code>recoverActiveGoal()</code>，后者立即创建 continuation intent 并调度执行。因此，Goal 会在重启后自行工作并消耗 token，违反公开契约中“仅设置 Goal 不会启动执行”的约定。

请持久化明确的预激活阶段或标记，只恢复已经由 observed Turn 激活的 Goal。回归测试应覆盖 arm → close/reopen → recover，证明用户发送下一条 Turn 前不会 admission，同时验证已经激活的 Goal 仍能正常恢复。

</details>

packages/runtime-host/src/server/goal-coordinator.ts
```diff
@@ -306,6 +320,49 @@ export class HostGoalCoordinator {
...
+        ...(input.maxIterations === null ? {} : { maxIterations: input.maxIterations }),
+        ...(input.tokenBudget === null ? {} : { tokenBudget: input.tokenBudget }),
+        tokensAtStart: this.#tokenCache.get(input.sessionId) ?? 0,
```
M4n5ter: <details open>

**Blocking — Count the first Goal-bound Turn against a manually selected token budget**

<code>GoalManager.create()</code> always sets <code>tokensBaselinePending: true</code>, which is appropriate when <code>GoalSet</code> is called from inside the Turn that creates the Goal. It is not correct for an externally armed Goal. When the first Goal-bound Turn settles, <code>settleTurn()</code> replaces <code>tokensAtStart</code> with the post-Turn token count and skips the budget check. A Goal armed with a 1,000-token budget can therefore spend more than 1,000 tokens during its first Turn, remain active, and admit at least one additional continuation.

Please distinguish whether the baseline has already been established. For external arming, establish the accurate baseline before the first Goal-bound Turn starts, while preserving the pending-baseline behavior for <code>GoalSet</code> inside a Turn. Tests should cover both arming while idle and arming while an unrelated Turn is already running.

</details>

<details>

**阻塞问题——手动设置的 token budget 必须包含首个 Goal Turn**

<code>GoalManager.create()</code> 总是写入 <code>tokensBaselinePending: true</code>。这适用于在当前 Turn 内调用 <code>GoalSet</code> 的情况，却不适用于外部设置的 Goal。首个受 Goal 约束的 Turn 结束时，<code>settleTurn()</code> 会用 Turn 完成后的 token 数重置 <code>tokensAtStart</code>，并跳过预算检查。因此，一个设置了 1,000 token 预算的 Goal 可以在首个 Turn 中消耗超过 1,000 token，仍保持 active，并至少再 admission 一次 continuation。

请明确区分 baseline 是否已经建立。外部 arm 应在首个 Goal Turn 开始前建立准确 baseline，同时保留 Turn 内 <code>GoalSet</code> 所需的 pending-baseline 语义。测试应同时覆盖空闲时 arm，以及已有无关 Turn 运行时 arm。

</details>

apps/desktop/src/renderer/goal-dialog.tsx
```diff
@@ -0,0 +1,152 @@
...
+    if (props.sessionId === undefined) return;
+    setCondition('');
+    setMaxIterations(null);
+    setTokenBudget(null);
+    setError(undefined);
+    setArming(false);
+  }, [props.sessionId]);
+
+  const sessionId = props.sessionId;
+  const canSubmit = condition.trim().length > 0 && !arming;
```
M4n5ter: <details open>

**Blocking — Do not arm a different budget from the value still shown in the form**

Astryx <code>NumberInput</code> only calls <code>onChange</code> when its pending text passes validation. Invalid text remains internal until blur, while <code>canSubmit</code> ignores the budget fields. For example, entering <code>500</code> for the token budget leaves the controlled <code>tokenBudget</code> as <code>null</code>, but Start remains enabled. Clicking Start then arms the Goal with <code>null</code>, which the Host interprets as no token ceiling, even though the form was still showing the user’s attempted limit. An invalid maximum-turn value can similarly submit its last valid prefix.

Please make the raw budget validity part of the form state and disable submission while either field contains invalid pending input. The fields should also be integer-only, and an interaction test should prove that an invalid displayed budget cannot call <code>window.maka.goal.arm</code>.

</details>

<details>

**阻塞问题——不要设置与表单当前显示值不同的预算**

Astryx <code>NumberInput</code> 只会在待输入文本通过校验时调用 <code>onChange</code>。无效文本会保留在组件内部直到失焦，而 <code>canSubmit</code> 完全没有检查预算字段。例如，输入 token budget <code>500</code> 时，受控状态 <code>tokenBudget</code> 仍然是 <code>null</code>，但“开始”按钮保持可用。点击后会用 <code>null</code> 设置 Goal，Host 将其解释为没有 token 上限，尽管表单当时仍显示用户尝试输入的限制。无效的最大 Turn 数也可能提交最后一个有效前缀。

请把预算输入的原始有效性纳入表单状态，只要任一字段仍有无效待输入内容就禁止提交。同时应将字段限制为整数，并增加交互测试，证明显示无效预算时不会调用 <code>window.maka.goal.arm</code>。

</details>

packages/runtime-host/src/protocol/goal.ts
```diff
@@ -79,6 +107,18 @@ export const GOAL_OPERATION_SPECS = {
       }
     },
   }),
+  'goal.arm': defineOperation({
```
M4n5ter, line 110: <details open>

**Important — Bump the compatibility epoch for the new wire operation**

This adds a new wire operation while <code>RUNTIME_HOST_COMPATIBILITY_EPOCH</code> remains 24. A new Desktop can therefore complete the handshake with an older epoch-24 standalone or remote Host, expose Set Goal, and then send an operation that the old decoder rejects as an unknown key. That protocol error tears down the connection instead of rejecting the incompatible pair during the handshake and entering the Host-upgrade path.

Please advance the compatibility epoch after rebasing and update the handshake/registration fixtures. This is the same compatibility boundary used by recent additions and removals of Runtime Host wire contracts.

</details>

<details>

**重要问题——新增 wire operation 后需要提升 compatibility epoch**

这里增加了新的 wire operation，但 <code>RUNTIME_HOST_COMPATIBILITY_EPOCH</code> 仍为 24。因此，新 Desktop 可以与旧的 epoch-24 standalone 或 remote Host 成功握手、展示“设置 Goal”，随后发送旧 Host decoder 无法识别的 operation。该 protocol error 会导致连接被拆除，而不是在握手阶段拒绝不兼容组合并进入 Host 升级路径。

请在 rebase 后提升 compatibility epoch，并更新 handshake/registration fixtures。仓库近期新增或删除 Runtime Host wire contract 时也使用了相同的兼容性边界。

</details>

apps/desktop/src/main/runtime-host-client.ts
```diff
@@ -1182,6 +1182,16 @@ export class DesktopRuntimeHostClient {
...
   }
 
+  /**
+   * Arm a Goal the user asked for. No optimistic retry loop like `clearGoal`:
+   * arming names no revision, so there is no stale one to refresh — a Session
+   * that already has an unfinished Goal fails with `operation_conflict`, and
+   * that is an answer for the user, not a race to re-run.
+   */
+  armGoal(input: OperationInput<"goal.arm">): Promise<OperationOutput<"goal.arm">> {
+    return this.request("goal.arm", input);
```
M4n5ter, line 1193: <details open>

**Important — Reconcile a dispatched interruption before reporting arm failure**

<code>goal.arm</code> is a control operation, so a connection loss after dispatch deliberately produces an unknown outcome instead of being retried. The Host may already have created and durably flushed the Goal before the response is lost, but this pass-through makes the dialog report “The goal could not be set. Try again” and leaves it retryable. The next user Turn can then run under a Goal the UI just said failed, while retrying the dialog only produces a conflict.

Please give arming an idempotency or reconciliation path rather than blindly retrying it. For example, after a dispatched interruption, reconnect and query the canonical Goal while accounting for the known pre-state and request; a durable operation identity would provide the stronger contract. Add a test covering commit-success followed by response loss.

</details>

<details>

**重要问题——报告 arm 失败前需要协调已经 dispatch 的中断**

<code>goal.arm</code> 是 control operation，因此 dispatch 后发生连接中断时不会自动重试，而是产生结果未知的异常。Host 可能已经创建并持久化 Goal，只是响应在返回前丢失；但当前透传会让弹窗显示“Goal 设置失败，请重试”，并允许再次提交。用户的下一条 Turn 可能在 UI 刚刚宣称失败的 Goal 下运行，而再次提交只会得到 conflict。

请为 arm 增加幂等或状态协调路径，不要直接重发 mutation。例如，在 dispatched interruption 后重新连接并查询 canonical Goal，同时结合已知的请求前状态和请求内容判断结果；持久化 operation identity 会提供更强的契约。还应增加“提交成功但响应丢失”的测试。

</details>

M4n5ter (changes requested):
<details open>

Requesting changes. I found five non-duplicate correctness and compatibility issues:

1. an armed-but-not-started Goal begins autonomously after Host restart;
2. the first Goal-bound Turn is excluded from a manually selected token budget;
3. invalid pending budget text can silently arm a Goal with different or unlimited bounds;
4. the new wire operation does not advance the compatibility epoch;
5. a response-lost arm can commit durably while the UI reports failure.

The Host admission, scoped Session binding, conflict/archive handling, and durable flush paths otherwise look coherent. The simplify audit also found a second Desktop-side runtime schema for <code>goal.arm</code>; consolidating it into the canonical Host operation decoder is a worthwhile non-blocking follow-up, but I am not opening a separate inline thread because it substantially overlaps the existing single-source and exact-IPC-shape discussions.

One additional upgrade reminder: credentials issued to remote owners before this change do not contain the new <code>goal.arm</code> grant. The current Remote Host user population is effectively negligible and reissuing those credentials is inexpensive, so the practical impact can be ignored for this PR; it should not block the change.

All checks on the current head are green, but the affected restart, first-Turn budget, mixed-version, invalid pending-input, and dispatched-interruption paths are not covered by those tests.

</details>

<details>

请求修改。本次发现 5 个没有与现有评论重复的 correctness 和 compatibility 问题：

1. 尚未开始的手动 Goal 会在 Host 重启后自行执行；
2. 首个受 Goal 约束的 Turn 不计入手动选择的 token budget；
3. 无效的待输入预算可能被静默替换为其他限制或无限制；
4. 新增 wire operation 后没有提升 compatibility epoch；
5. arm 响应丢失时，Goal 可能已经持久化，但 UI 会错误报告失败。

除此之外，Host admission、scoped Session 绑定、conflict/archive 处理和 durable flush 路径整体一致。<code>simplify-audit</code> 还发现 Desktop 为 <code>goal.arm</code> 建立了第二套 runtime schema；将其收敛到 Host operation 的 canonical decoder 是有价值的非阻塞 follow-up，但它与现有 single-source 和 exact IPC shape 讨论高度相邻，因此不再单独创建 inline thread。

另有一项升级提醒：本次变更之前签发给 remote owner 的 credential 不包含新增的 <code>goal.arm</code> grant。目前 Remote Host 用户量几乎可以忽略，重新签发 credential 的成本也很低，因此该项的实际影响可以忽略，不作为本 PR 的阻塞理由。

当前 head 的所有 checks 均为绿色，但 restart、首个 Turn 预算、混合版本、无效待输入以及 dispatched interruption 这些路径没有被现有测试覆盖。

</details>

packages/runtime-host/src/server/goal-coordinator.ts
Astro-Han: <details open>

Fixed in 341b2a423. Confirmed the path: `#arm` persists `active` with `currentExecution: null`, `recover()` hands that to `recoverActiveGoal`, and it builds a continuation intent from `status` alone.

I did not add the pre-activation marker, because a new field on `GoalState` has to enter `GOAL_STATE_SHAPE`, and `hasExactShape` rejects any key outside `allowed`. A record written by the new build would then fail to decode in `prepareRecovery` on an older one, turning a rollback into a Host that will not start.

The distinction is already persisted. Only a Turn that settled while bound to the Goal raises `iterations`, so `iterations === 0` is exactly "no Turn has carried this Goal". `recover()` now reads that as `hasCarriedTurn` and leaves an uncarried Goal for `beginObservedTurn` to bind, as it would have without the restart. The current-execution branch is unchanged: an execution in flight is its own proof of carrying.

Regression test covers arm → close → reopen → `recover()`, asserting no admission and no evaluation. Both hooks record rather than throw, because a throw inside admission is swallowed by the drain and only surfaces as a paused Goal, which names the wrong failure. On the previous code it fails with `the restart admitted a Turn for a Goal no Turn had carried`. The already-activated case stays covered by the existing current-execution recovery tests.

One behaviour change to flag: this also covers a Goal that `GoalSet` created inside a Turn the Host never lived to settle. That Turn is gone after the restart, so it now waits for the user's next Turn instead of self-starting — the same answer, in the safe direction.

</details>

<details>

已在 341b2a423 修复。链路确认：`#arm` 持久化 `active` + `currentExecution: null`，`recover()` 交给 `recoverActiveGoal`，后者仅凭 `status` 就建立 continuation intent。

未采用预激活标记。`GoalState` 新增字段必须进入 `GOAL_STATE_SHAPE`，而 `hasExactShape` 拒绝 `allowed` 之外的任何键——新版本写出的记录会让旧版本在 `prepareRecovery` 里 decode 失败，回滚将变成 Host 起不来。

该区分其实已经持久化了：只有绑定 Goal 结算的 Turn 会抬高 `iterations`，因此 `iterations === 0` 精确等于「无 Turn 承载过这个 Goal」。`recover()` 现在以 `hasCarriedTurn` 读它，未承载的 Goal 留给 `beginObservedTurn` 在下一个 Turn 绑定，与没有重启时一致。current-execution 分支不变——飞行中的 execution 本身就是被承载的证据。

回归测试覆盖 arm → 关闭 → 重开 → `recover()`，断言不 admission、不 evaluate。两个钩子都记录而非抛错：抛错会被 drain 吞掉、只表现为 Goal 被 pause，那会指认错误的失败原因。在旧代码上它报 `the restart admitted a Turn for a Goal no Turn had carried`。已激活 Goal 的恢复仍由既有 current-execution 测试覆盖。

需要披露一处行为变化：这同样覆盖了 `GoalSet` 在 Turn 内创建、而 Host 未活到该 Turn 结算的情形。重启后那个 Turn 已经不在，因此改为等待用户的下一个 Turn 而非自启——同一个答案，且在安全方向上。

</details>

packages/runtime-host/src/server/goal-coordinator.ts
Astro-Han: <details open>

Taking the finding and rejecting the remedy — the code turned out to be further from what it looks like.

`opts.tokensAtStart` was never read by any decision, on either path. `create` writes it into `tokensAtStart` and `tokensNow` together, so the spend it implies is zero whatever it holds, and the first settled Turn overwrites both under `tokensBaselinePending` before any budget is compared. `GoalSet`'s `deps.getTokenCount` read and arm's `#tokenCache` read were two equivalent inert plugs. Removed in 526d289b8, along with the now-unused dep.

Arming cannot supply an accurate baseline either. `#tokenCache` is only written by `getRecentContext` during continuation evaluation, so a Session that has never run a Goal has no entry at all. And an unrelated Turn may be in flight at arm time — `#arm` documents that it settles outside the Goal — so a snapshot read there would charge its spend to a Goal that never caused it, and could reach `budget_limited` with a `lastReason` naming money the Goal never spent. An accurate baseline needs a token count read at the moment a Turn binds the Goal; `beginObservedTurn` is synchronous and in another package, so that read point does not exist today.

So the semantics are now stated rather than changed: the budget starts at the first Turn carrying the Goal and covers everything the Goal drives after it. That Turn is one the user is present for — `GoalSet` runs inside it, and an armed Goal takes hold on the one the user sends next — while the budget exists to bound what runs unattended. `GoalSet` has behaved this way since it shipped; arm now matches it instead of carrying a parameter that suggested otherwise. A new test pins it: the first settle establishes the baseline without being charged, and the budget is enforced from there.

The asymmetry left over — that first Turn counts toward `maxIterations` but not toward the budget — predates this PR and applies to `GoalSet` equally. Happy to open an issue if you want it closed.

</details>

<details>

接受问题，拒绝该修法——实际情况比看上去更远。

`opts.tokensAtStart` 在两条路径上都从未被任何判断读取。`create` 把它同时写入 `tokensAtStart` 和 `tokensNow`，因此它取何值所隐含的消耗都是 0；首个结算的 Turn 又会在任何预算比较之前，依据 `tokensBaselinePending` 把两者整体覆盖。`GoalSet` 的 `deps.getTokenCount` 读数与 arm 的 `#tokenCache` 读数是两个等价的死插头。已在 526d289b8 连同不再使用的依赖一并删除。

arm 也给不出准确 baseline。`#tokenCache` 只在 continuation 评估的 `getRecentContext` 中写入，所以从未跑过 Goal 的 Session 根本没有条目。而且 arm 时可能有无关 Turn 正在飞行——`#arm` 自己就写明它在 Goal 之外结算——此时读快照会把它的消耗记到从未导致它的 Goal 上，甚至可能带着谎报的 `lastReason` 触发 `budget_limited`。准确的 baseline 需要在某个 Turn 绑定 Goal 的瞬间读取 token 计数；`beginObservedTurn` 是同步的且位于另一个包，该读点今天并不存在。

因此语义被明确而非改变：预算从首个承载 Goal 的 Turn 起算，覆盖此后由 Goal 驱动的全部内容。那个 Turn 是用户在场发起的——`GoalSet` 就在它内部运行，而 arm 的 Goal 在用户发出的下一个 Turn 上生效——预算的目的则是约束无人值守的消耗。`GoalSet` 自上线以来就是这个语义，arm 现在与之一致，而不是继续携带一个暗示相反语义的参数。新测试固定了这一点：首个结算建立 baseline 且不被计入，预算自此开始生效。

剩下的不对称——首个 Turn 计入 `maxIterations` 却不计入预算——先于本 PR，且对 `GoalSet` 同样成立。如果你希望一并解决，我可以另开 issue。

</details>

apps/desktop/src/renderer/goal-dialog.tsx
Astro-Han: <details open>

Fixed in 66f5d3931, and the exposure is wider than the example. `NumberInput` renders `type="text"`, so unparseable text reaches it too: both out-of-range and non-numeric input leave the controlled value untouched and revert silently on blur. Typing a 500-token budget armed `null` — not a smaller ceiling but no ceiling at all, which inverts the one gesture meant to tighten a run.

Making the raw validity part of the form state is not reachable with `NumberInput`: it never exposes `pendingInput`, and by blur the text is already gone. So the fields now hold the user's own text and the dialog reads it. Display and payload cannot diverge, whole numbers are required, and an entry the Host would refuse blocks Start with the reason beside the field. `readGoalBudget` repeats the operation's bounds only to say that locally instead of as a rejected frame; the numbers come from the shared Goal limits, and the copy takes them as arguments so no sentence carries its own copy of a limit.

Interaction test as asked: `apps/desktop/e2e/goal-dialog-budget.spec.ts`. It fails on the previous dialog with Start still enabled under a 500-token budget, and it asserts what was armed by reading the Goal back from the Host rather than by watching the bridge call — the bridge is a `contextBridge` object and cannot be intercepted anyway.

Trade-off worth naming: this gives up `NumberInput`'s stepper and thousands formatting on these two fields.

</details>

<details>

已在 66f5d3931 修复，且暴露面比示例更广。`NumberInput` 渲染的是 `type="text"`，因此无法解析的文本同样能输入：超界与非数字输入都会让受控值保持不变，并在失焦时静默回退。输入 500 的 token 预算会 arm 出 `null`——那不是更小的上限，而是完全没有上限，恰好把唯一用于收紧运行的操作反了过来。

把原始有效性纳入表单状态对 `NumberInput` 不可行：它从不暴露 `pendingInput`，而到失焦时文本已经消失。因此字段现在持有用户自己的文本，由 dialog 读取。显示与发送不可能分叉，必须是整数，Host 会拒绝的输入则直接禁用「开始」并把原因显示在字段旁。`readGoalBudget` 重复 operation 的边界，只是为了在本地给出拒绝理由而不是收到一个被拒的 frame；数值仍来自共享的 Goal 限制，文案以参数接收它们，所以没有任何一句话自带一份限制副本。

按你的要求补了交互测试：`apps/desktop/e2e/goal-dialog-budget.spec.ts`。它在旧对话框上会因 500 预算下「开始」仍可点而失败，并且是通过**从 Host 读回 Goal** 来断言实际 arm 了什么，而不是监视 bridge 调用——bridge 是 `contextBridge` 对象，本来也无法拦截。

需要点明的取舍：这两个字段因此失去了 `NumberInput` 的步进按钮和千位格式化。

</details>

packages/runtime-host/src/protocol/goal.ts
Astro-Han: <details open>

Fixed in bb24a6234. Rebased first, and `main` had already reached 25, so this advances it to 26.

Handshake and registration fixtures are unchanged on purpose: they name deliberately stale epochs (9, 16, 20, 23) to exercise rejection, so pinning them to the current one would remove what they test. That matches the precedent — both #3145, which added an operation, and #3165, which only removed catalog filters, were single-line changes to `RUNTIME_HOST_COMPATIBILITY_EPOCH`.

</details>

<details>

已在 bb24a6234 修复。先做了 rebase，`main` 此时已到 25，因此本次提升为 26。

握手与注册 fixture 有意保持不变：它们使用的是故意过期的 epoch（9、16、20、23）以触发拒绝路径，若改为当前值反而会抹掉它们所测试的东西。这与先例一致——新增 operation 的 #3145 和仅删除 catalog filter 的 #3165，都是对 `RUNTIME_HOST_COMPATIBILITY_EPOCH` 的单行改动。

</details>

apps/desktop/src/main/runtime-host-client.ts
Astro-Han: <details open>

Agreed on the symptom, deferring the remedy — tracked as #3234.

`retryable = mode === 'query'` at `client/connection.ts:331` is the contract for every control operation in the Host, not something particular to arming. `goal.control`, session continuity, hosted execution and the turn and message mutations all report a dispatched interruption the same way, and there is no operation identity on the wire and no reconciliation helper anywhere to build on — `isReconnectableReadFailure` in `ipc-reconnect-policy.ts` covers queries only. Giving `goal.arm` its own idempotency would add a public contract inside this PR rather than extend an existing seam.

The exposure is also narrower than the dialog's message suggests: `useSessionGoal` subscribes to `goal-change` and renders a persistent Goal indicator with a one-click clear, so a Goal that did commit appears there with a way to stop it. It is still a wrong message, which is why it is worth fixing — just not here.

I recorded your reconcile-after-reconnect shape in the issue as the first level, since `goal.query` is already a retryable query and the dialog could report "connection lost; the current Goal is X" without any protocol change, with durable operation identity as the stronger second level.

</details>

<details>

症状认同，方案推迟——已记录为 #3234。

`client/connection.ts:331` 的 `retryable = mode === 'query'` 是 Host 中**所有** control 操作的契约，并非 arm 特有。`goal.control`、session continuity、hosted execution 以及 turn 和 message 的各类 mutation 都以同样方式报告 dispatched interruption，而 wire 上不存在 operation identity，也没有任何可供扩展的协调设施——`ipc-reconnect-policy.ts` 里的 `isReconnectableReadFailure` 只覆盖 query。单独为 `goal.arm` 造一套幂等，是在本 PR 内新增公共契约，而不是扩展既有接缝。

暴露面也比弹窗文案所暗示的更窄：`useSessionGoal` 订阅 `goal-change` 并渲染常驻的 Goal 指示器（带一键清除），因此真正提交成功的 Goal 会在那里显示并可被停止。但这仍然是一条错误的提示，所以值得修——只是不在这里修。

我把你提出的「重连后对账」形状作为第一层记进了 issue：`goal.query` 本身就是可重试的 query，弹窗可以在不改协议的前提下报告「连接中断，当前 Goal 为 X」；持久化 operation identity 作为更强的第二层。

</details>

Astro-Han:
Thanks for the careful pass — the restart and budget findings were both real, and chasing them turned up more than the diff showed.

All five are answered in their own threads: 1, 3 and 4 are fixed; for 2 I took the finding but not the remedy, since the baseline parameter turned out to be inert on both paths, so it is removed and the semantics stated; 5 is deferred to #3234, where it belongs as a contract shared by every control operation. Two of your suggested implementations I did not take, with reasons in the threads.

Would appreciate another look when you have time.

*AI disclosure: this round's code, tests and thread replies were produced with Claude Code. I reviewed the final diff and own the result; each fix was verified by first making its test fail for its own reason.*

---

感谢细致 review——重启和预算这两条都成立，而且顺着查下去发现的比 diff 上能看到的更多。

五条都已在各自 thread 中回复：1、3、4 已修；第 2 条我接受问题但未采用该修法——那个 baseline 参数在两条路径上都从未生效，因此删除并明确语义；第 5 条推迟到 #3234，它本就属于所有 control 操作共享的契约。另有两处你建议的实现我没有采纳，理由写在对应 thread 里。

方便时麻烦再看一轮。

*AI 披露：本轮代码、测试与 thread 回复由 Claude Code 产出。我审阅了最终 diff 并对结果负责；每处修复都先让其测试因它自己的原因失败来验证。*

docs/astryx-surface-file-inventory.md
```diff
@@ -43,6 +43,7 @@ Wiki bar: Design Conventions · API Use-the-System · Theming · Container Paddi
 | `apps/desktop/src/renderer/command-palette.tsx` | dialog-overlay | EmptyState | aligned — uses Astryx (EmptyState) | aligned |
 | `apps/desktop/src/renderer/custom-pet-companion.tsx` | other | none | aligned — no raw controls; no Astryx JSX usage | aligned |
 | `apps/desktop/src/renderer/error-boundary.tsx` | other | Button, Card | aligned — uses Astryx (Button, Card) | aligned |
+| `apps/desktop/src/renderer/goal-dialog.tsx` | dialog-overlay | Button, Dialog, DialogHeader, HStack, Layout, LayoutContent, Text, TextInput, VStack | aligned — uses Astryx (Button, Dialog, DialogHeader, HStack, Layout, LayoutContent, Text, TextInput) | aligned |
```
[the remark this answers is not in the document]
Astro-Han: Not manually edited — this row is generator output, and both artifacts were regenerated with `npm run astryx:surface-inventory:write` before the push. `npm run astryx:surface-inventory` passes on this commit.

`LayoutFooter` and `TextArea` are absent because the generator does not track them: `ASTRYX_COMPONENTS` in `scripts/generate-astryx-surface-inventory.mjs` lists `Layout`, `LayoutContent`, `LayoutHeader` and `LayoutPanel` but not `LayoutFooter`, and `TextInput` but not `TextArea`. Neither name appears anywhere in `docs/astryx-surface-file-inventory.md`, for any of the 189 files — so this is a property of the tracked set, not of this row.

Widening that set is a separate change to the inventory tooling, and it would rewrite rows across the repository rather than this one.

packages/runtime-host/src/server/goal-coordinator.ts
M4n5ter: <details open>

Verified on `66f5d39`. `recover()` no longer enqueues a continuation for a durable Goal with neither a current execution nor a settled carried Turn. The regression exercises arm → close → prepare/recover against the durable store and proves that recovery neither evaluates the Goal nor admits a Turn. Reusing the persisted iteration count also avoids adding a rollback-sensitive storage field.

Strictly, `iterations === 0` means that no Goal-bound Turn has settled; an abrupt crash after a Turn binds but before it settles can therefore leave the Goal waiting for the user’s next Turn. That is a conservative, narrow recovery outcome rather than unattended execution, so it does not block this change.

Resolved.

</details>

<details>

已在 `66f5d39` 上验证。对于既没有 current execution、也没有已结算承载 Turn 的持久化 Goal，`recover()` 不再创建 continuation。回归测试通过 durable store 覆盖 arm → close → prepare/recover，并证明恢复过程既不会 evaluate Goal，也不会 admit Turn。复用已持久化的 iteration count 也避免了新增一个影响回滚兼容性的存储字段。

严格来说，`iterations === 0` 表示尚无 Goal-bound Turn 完成结算；如果 Turn 已绑定但在结算前发生突然崩溃，Goal 可能会等待用户的下一轮。这是保守且窗口很窄的恢复结果，不会造成无人值守执行，因此不阻塞本次变更。

已解决。

</details>

apps/desktop/src/main/runtime-host-client.ts
M4n5ter: <details open>

The ambiguous-outcome symptom remains real, but I accept the deferral to #3234. It is part of the shared Runtime Host control-operation contract, the triggering window requires a disconnect after dispatch but before the response, and subscription recovery exposes a committed Goal again with a clear action available. The incorrect failure message remains worth fixing, but it is now a tracked follow-up rather than an approval blocker for this PR.

Resolved as tracked by #3234.

</details>

<details>

结果不确定的问题仍然真实存在，但我接受推迟到 #3234。它属于 Runtime Host 所有 control operation 共用的契约；触发需要连接恰好在 dispatch 后、响应返回前中断；订阅恢复后，已经提交的 Goal 会重新显示，并提供 clear 操作。错误的失败提示仍值得修复，但现在作为已跟踪的 follow-up，不再阻塞本 PR 的 approval。

按 #3234 跟踪处理，已解决本 thread。

</details>

packages/runtime-host/src/server/goal-coordinator.ts
M4n5ter: <details open>

Verified. The creation-time `tokensAtStart` value was inert on both callers: creation wrote it to both counters, and the first settlement replaced both before any budget comparison. Removing the parameter and its duplicate reads is the correct simplification.

I accept the established semantics made explicit here: the first Goal-carrying Turn establishes the baseline, and the token budget bounds the Goal-driven work that follows it. The remaining difference—maximum iterations counts that first Turn while the token budget does not—predates this arm path and can be reconsidered separately if the product wants one unified definition; it is not a blocker for this PR.

Resolved.

</details>

<details>

已验证。创建时传入的 `tokensAtStart` 对两个调用方都没有实际作用：创建逻辑会把它同时写入两个计数器，而首次结算又会在任何预算比较之前同时覆盖两者。删除该参数及两处重复读取是正确的简化。

我接受这里明确下来的既有语义：首个承载 Goal 的 Turn 用于建立 baseline，token budget 约束其后由 Goal 驱动的工作。剩余的不对称——maximum iterations 会计算首个 Turn，而 token budget 不会——早于本次 arm 路径；如果产品希望统一定义，可以单独重新评估，但不阻塞本 PR。

已解决。

</details>

packages/runtime-host/src/protocol/goal.ts
M4n5ter: <details open>

Verified. The rebased branch advances the compatibility epoch from main’s 25 to 26, so mixed old-Host/new-client pairs are rejected during the handshake before `goal.arm` can reach an older decoder. The deliberately stale handshake fixtures do not need to track the current epoch.

Resolved.

</details>

<details>

已验证。Rebase 后的分支把 main 的 compatibility epoch 从 25 提升到 26，因此旧 Host 与新 client 的混合组合会在握手阶段被拒绝，不会让 `goal.arm` 进入旧 decoder。那些有意使用过期 epoch 的 handshake fixture 不需要跟随当前值更新。

已解决。

</details>

apps/desktop/src/renderer/goal-dialog.tsx
M4n5ter: <details open>

Verified on the real path. The fields now retain the user’s raw text, one parse determines both validity and the submitted payload, and non-integer or out-of-range input disables Start. The E2E regression proves that invalid displayed input leaves the Host without a Goal, then reads the exact valid budgets back from the Host after submission.

Resolved.

</details>

<details>

已按真实路径验证。字段现在保留用户输入的原始文本，同一次解析同时决定有效性和提交 payload；非整数或超界输入都会禁用「开始」。E2E 回归证明显示无效输入时 Host 中仍然没有 Goal，随后提交有效值后又会从 Host 读回完全一致的预算。

已解决。

</details>

M4n5ter (dismissed):
<details open>

Approved on head `66f5d39` for the Goal-specific delta stacked after `c759107` (#3198).

I reverified all five previous findings. The restart behavior, invalid budget input, and compatibility epoch are fixed on their real paths. Removing the inert creation-time token baseline correctly exposes the existing Goal budget semantics. The dispatched-interruption ambiguity remains real but is narrow, recoverable through the canonical Goal projection, and tracked across control operations in #3234; it is a non-blocking follow-up here.

No new P0–P2 issue was found. The automated composite-Session-key concern is a false positive: preload resolves the Desktop key to its Host scope and Host-local Session id before invoking `goal:arm`, then projects the returned identifiers back to Desktop keys. All checks on this head are green.

Scope caveat: this PR is explicitly stacked on #3198. This approval assumes #3198 lands first with its outstanding mode-pair authority fix, followed by the documented rebase; it does not supersede that review.

</details>

<details>

批准 head `66f5d39` 中叠加在 `c759107`（#3198）之后的 Goal 专属增量。

我重新验证了上一轮的五个问题。重启行为、无效预算输入和 compatibility epoch 均已在真实路径上修复。删除创建时无效的 token baseline，正确显式化了既有 Goal budget 语义。Dispatch 后中断造成的结果不确定仍然真实存在，但触发窗口较窄，可以通过 canonical Goal projection 恢复确认，并已在 #3234 中按所有 control operation 的共用问题跟踪；这里降为非阻塞 follow-up。

未发现新的 P0–P2。自动审查提出的 composite Session key 问题是误报：preload 会先把 Desktop key 解析为对应 Host scope 和 Host-local Session id，再调用 `goal:arm`，最后才把返回标识重新投影为 Desktop key。当前 head 的所有 checks 均为绿色。

范围说明：本 PR 明确叠加在 #3198 之上。本 approval 假设 #3198 会先带着仍待修复的 mode-pair authority 问题合并，随后按说明 rebase；它不会覆盖 #3198 的 review 结论。

</details>

Astro-Han:
Two force-pushes since the last review, both mechanical — and one real delta worth naming.

**The rebases.** #3198 was squash-merged, so this branch was replayed onto the new `main` and its base retargeted from the stacked branch back to `main`. That rewrote the SHAs, which is what dismissed the approval. The second push is the same tree with a fresh SHA: `ci.yml` triggers on `pull_request: branches: [main]`, and the first push landed while the base was still the stacked branch, so CI never ran on it.

**What actually changed since `66f5d393`.** The eight commits are the same eight, but they are not byte-identical to what was approved. Rebasing across #3027 resolved three conflicts — `goal:pause` / `goal:resume` now sit beside `goal:arm` in the IPC block, `runGoalControl` and the Goal dialog's session state are two independent declarations in `app-shell.tsx`, and the two new client tests share a file tail. `main` also moved the compatibility epoch to 27, so `goal.arm` now retires 27 and lands on 28.

CI is green on `6a074c06c`: `test` and `windows_sandbox_w0_protocol` both pass, and the plan step selected this PR's 28 files against the merged `main`.

@M4n5ter thanks again for the earlier pass — the two deltas above are the only things worth re-reading. @jackwener @likun666661 @hqhq1025 if any of you have time, a second pair of eyes would be very welcome: the parts that matter most are the new `goal.arm` Host operation with its epoch bump, and the Goal dialog wiring in `app-shell.tsx`.

packages/runtime-host/src/server/goal-coordinator.ts
```diff
@@ -179,12 +193,24 @@ export class HostGoalCoordinator {
...
+   * restart. An execution that was in flight is its own proof of carrying, so
+   * that branch keeps recovering unconditionally.
+   */
   async recover(): Promise<void> {
     if (!this.#prepared) throw new Error('Goal recovery was not prepared');
     for (const snapshot of this.#authorityBySession.values()) {
       if (snapshot.record.currentExecution) {
         await this.#recoverCurrentExecution(snapshot.record.currentExecution);
-      } else {
+      } else if (hasCarriedTurn(snapshot.record.goal)) {
```
hqhq1025: [P1] Preserve explicit Resume intent across restart

`iterations > 0` proves only that a Goal-bound Turn settled; it does not prove whether autonomous continuation was explicitly requested. I reproduced: arm a Goal, pause it before any Turn settles, resume it while admission is busy, then restart before `currentExecution` is recorded. The durable record is `active` with `iterations === 0`, so this branch skips recovery. After restart the UI still shows an active Goal, but `admitTurn` is never called and the explicit Resume is silently lost until another user Turn arrives.

Please make the recovery predicate preserve the resumed/autonomous phase instead of equating it with a settled iteration, and add this pause -> resume -> restart sequence to the durable recovery test.

packages/runtime-host/src/server/goal-coordinator.ts
```diff
@@ -306,6 +332,48 @@ export class HostGoalCoordinator {
...
+   * scheduled here: a Goal armed with no Turn running takes effect on the next
+   * Turn, which `beginObservedTurn` binds to the live control lease, and the
+   * loop starts when that Turn settles. Arming does not itself start spending.
+   *
+   * A Turn already in flight keeps the standing it registered with — it was
+   * bound before this Goal existed, so it settles outside the Goal, and the one
+   * after it is the first the Goal drives.
+   */
+  #arm(input: GoalArmInput): Promise<OperationOutcome<'goal.arm'>> {
+    return this.#sessionAdmission.run(input.sessionId, async () => {
```
hqhq1025, line 350: [P1] Recheck draining after the Session admission wait

The kernel begins composition drain before waiting for already-admitted operations, and `beginDrain()` clears the Goal manager immediately. An admitted `goal.arm` can still be queued behind another operation in `SessionAdmissionGate`; when that holder releases after drain began, this callback continues, recreates the cleared Goal, flushes it, and returns success. I reproduced the exact sequence and read the new Goal back from durable storage after drain.

Recheck `#draining` inside the admitted callback before any read or mutation and return `host_draining`. The regression should hold the Session gate, queue an arm, begin drain, release the gate, and verify that no in-memory or durable Goal is created.

hqhq1025 (changes requested):
Requesting changes. I confirmed two lifecycle blockers on the current head.

Problem and mechanism

This PR correctly identifies the missing product capability: a user could control an existing Goal but could not arm one outside a model Turn. The Host-owned `goal.arm` operation, scoped Desktop IPC, exact protocol frame, shared limits, and dialog are the right overall mechanism.

First-principles assessment

The single Host authority and the rule that arming alone schedules nothing are sound. The current implementation is not yet a complete final architecture because it infers autonomous recovery intent from `iterations > 0`, although an explicit Resume can create an active autonomous intent before any Turn settles. It also allows an already-admitted arm to cross the composition drain boundary after the Goal manager has been disposed.

Optimality and simplification

No production code is clearly deletable, and the added protocol, coordinator, IPC, and E2E coverage is generally meaningful. No low-quality test needs deletion; instead, the two reproduced sequences need regression coverage. A repository-wide refactor is not required, but the durable recovery predicate must represent explicit Resume intent, and `goal.arm` must recheck admission availability after waiting in the Session gate.

Merge verdict

Not ready to merge at `6a074c06c4082ea66926db9c9c16bb96d6516247`. In addition to the two findings, current `main` now conflicts in `apps/desktop/stories/app-shell.stories.tsx`, so this head needs another rebase and fresh CI.

Verification

- Runtime Host Goal tests: 12/12 passed.
- Runtime/Core Goal tests: 22/22 passed.
- Desktop client and IPC tests: 33/33 passed.
- Goal dialog Electron E2E: 3/3 repeated runs passed.
- Desktop production build, Storybook typecheck, Biome, Astryx inventory, and `git diff --check`: passed.
- Independent lifecycle reproducer confirmed both findings: explicit Resume recovered with zero admissions, and a queued arm released after drain returned success and persisted the Goal.

Residual risk: the already tracked ambiguous control-operation response problem remains in #3234, and previously issued remote credentials still require replacement to gain the new grant. Those are follow-ups; the two inline findings are the blockers for this revision.

packages/runtime-host/src/server/goal-coordinator.ts
Astro-Han: Thanks — confirmed. I first tried the other way, making resume not drive a never-carried Goal so `iterations` would stop being an inference. The existing `one Host Goal is shared across clients…` test disproves it: a model-set Goal paused before its birth Turn settles must drive on resume, and `pause` drops the registration, so resume is its only way back. The distinction has to be recorded.

So there is a durable bit, but named for the birth rather than the resume: `armedAt`, stamped once by `goal.arm` and never cleared. `isDrivingGoal = armedAt === undefined || iterations > 0` — one write at one site, versus a resumed phase that would need maintaining at resume, pause, settle, clear and every terminal status. The rule now lives in `restoreDrive`, shared by recovery and resume, so the coordinator no longer keeps its own copy and `hasCarriedTurn` is gone.

Your sequence is covered, and fail-first both ways: commenting out `armed: true` turns both armed tests red. It also fixes a case neither of us named — a model-set Goal that crashed at `iterations: 0` was skipped too. (`864f337e4`)

packages/runtime-host/src/server/goal-coordinator.ts
Astro-Han: Thanks — confirmed exactly as described, and fixed as you asked. `#arm` rechecks `#draining` inside the admitted callback and returns `host_draining`, already a declared error, matching `message-coordinator`'s recheck-after-the-wait. Kept per-operation rather than in `SessionAdmissionGate`, which every domain shares.

Arm is the only hole: every other Goal operation reads the cleared manager and answers `not_found`. `GoalManager.dispose()` was just `goals.clear()`, so it now really disposes and `create()` throws afterwards — the recheck means nobody reaches it, the throw is so the next forgetful entry point fails loudly.

The regression holds the gate, queues an arm, drains, releases: asserts `host_draining` and nothing in memory or storage. (`2902eccda`)

Astro-Han:
Rebased onto current `main`; the `app-shell.stories.tsx` conflict was #3292 removing the inline completion. Verified on `2902eccda`: goal-coordinator 8/8, Goal suites across core/storage/runtime/runtime-host 69/69, desktop 33/33, `@maka/ui` 183/183, four tsconfig projects, Biome, Astryx inventory. Both findings checked fail-first.

Astro-Han:

Astro-Han:
Rebased onto current `main` (`fa0dc52b6`).

Two adaptations: #3332 took epoch 28, so `goal.arm` moved to 29 — the collision flagged above; #3249 removed `backend` from `CreateSessionInput`, so the four Session fixtures here drop that field.

Verified on the rebased tree: `typecheck`, `build`, Biome, Astryx inventory, plus runtime-host Goal 23/23, runtime Goal 82/82, desktop 33/33, `@maka/ui` composer 6/6.

@hqhq1025 both findings are answered in the threads above — re-review appreciated. @M4n5ter @jackwener @likun666661 a second pair of eyes is very welcome too, whenever you have time.

*(Rebase and local verification run with Claude Code (Opus 5).)*

hqhq1025 (changes requested):
[P1] The existing Resume-intent finding remains unresolved on this head.

At `packages/runtime/src/goal-continuation.ts:421`, `resumeFromControl()` returns an active Goal but `restoreDrive()` refuses to schedule it when `armedAt` exists and `iterations === 0`. This is a normal product path: arm a Goal, pause it before the first carried Turn, then press Resume. The UI promises that resuming continues autonomous iteration immediately and renders the Goal as running, but no continuation runs, and a restart preserves the inert state. The new regression at `packages/runtime-host/src/__tests__/goal-coordinator.test.ts:743` asserts this mismatch.

Please persist explicit current drive/resume intent, or an equivalently simple phase, and use it for both Resume and recovery. The regression should cover pause -> Resume while admission is busy -> restart before `currentExecution` is recorded.

I did not open a duplicate inline thread because this is the existing unresolved `hqhq1025` finding. I reviewed the complete revision; the required check and 73 focused local tests pass, but this P1 keeps this exact head from merge readiness.

hqhq1025 (approved):
Reviewed the current revision, including the fix for the prior resume/restart finding. The durable `armedAt` transition now distinguishes an armed Goal that is still waiting from one whose first settlement or explicit resume made it autonomous, and the recovery regression covers busy admission across restart. The repository build and 96 focused Goal tests passed, and the required CI check is green.

Issue #2522, opened 2026-08-08
[Tracking] Runtime Host remote access and production hardening


M4n5ter:
<details open>
<summary>English</summary>

## Goal

Complete the product integration and production hardening of Runtime Host as the single long-lived authority for one State Root and its Runtime work.

Local IPC and authenticated WebSocket Clients use the same Host state, operation routing, execution authority, recovery, and shutdown lifecycle. Desktop may connect to multiple independent Hosts as a Client-side composition; CLI and TUI continue to use one selected Host per process.

Stable architecture belongs in the [Runtime Host architecture guide](https://github.com/maka-agent/maka-agent/blob/main/docs/architecture/runtime-host-architecture.md). This issue tracks remaining product and operational work rather than duplicating protocol or TypeScript designs.

## Stable constraints

- One State Root has at most one writer Host.
- Local IPC and WebSocket are transport adapters for the same Kernel, Composition, Domain Modules, and canonical Stores.
- Root ID identifies the durable Host State Root. A profile identifies one immutable way to connect to it.
- Host-owned identifiers are scoped to one State Root. A Client that aggregates Hosts must retain Host identity on every resource reference.
- A remote target connects only; it never joins local discovery, election, candidate spawning, or fallback.
- Runtime Host is the only authority that resolves a workspace target to a Host path. Remote Clients select Host Projects and never reinterpret Client-local paths as Host paths.
- Reconnect restores canonical observation. It never blindly replays commands whose outcome may be unknown.
- TLS, SSH, and explicitly acknowledged plaintext are distinct connection methods. No method silently downgrades to another.
- Multi-Host Desktop remains a Client-side composition. It does not create a federated Host, cross-Host durable database, or second execution authority.

## Delivered foundation

- [x] Local standalone service, common transport/listener boundaries, and explicit ephemeral/service lifecycle ([#2583](https://github.com/maka-agent/maka-agent/pull/2583))
- [x] Authenticated WebSocket/WSS access, exact grants, root pinning, Origin policy, and native remote connector ([#2591](https://github.com/maka-agent/maka-agent/pull/2591))
- [x] Host-owned Project Catalog and canonical workspace authority through the closed project/host-path target model ([#2603](https://github.com/maka-agent/maka-agent/pull/2603), [#2642](https://github.com/maka-agent/maka-agent/pull/2642))
[...]

PR #3205, opened 2026-08-18
feat(cli): manage Linux Runtime Host services

M4n5ter:
## Summary

<details open>
<summary>English</summary>

Add a managed Runtime Host service lifecycle to the CLI so a Linux Host can remain available after its SSH session ends.

The lifecycle contract, persisted configuration, normalized status, and clean-uninstall behavior are platform-neutral. Linux supplies a systemd user backend; a future macOS LaunchAgent backend can reuse the same CLI and lifecycle semantics. Installation pins the exact Node.js and Maka CLI paths, binds the WebSocket listener to loopback, and preserves the existing port and project roots on reinstall. Uninstall removes the managed unit and configuration while retaining the State Root and user data.

Refs #2522

</details>

<details>
<summary>中文</summary>

为 CLI 增加 managed Runtime Host service lifecycle，使 Linux Host 能在 SSH 会话结束后继续运行。

Lifecycle contract、持久配置、归一化状态和 clean uninstall 均保持平台无关；Linux 仅提供 systemd user backend，未来 macOS LaunchAgent backend 可以复用相同的 CLI 与 lifecycle 语义。安装会固定精确的 Node.js 与 Maka CLI 路径、只监听 loopback，并在重复安装时保留已有端口和 Project roots。卸载会移除托管 unit 与配置，同时保留 State Root 和用户数据。

Refs #2522

</details>

## Verification

<details open>
<summary>English</summary>

- `npm --workspace maka-agent test` — 269 tests passed
- `npm run build` — passed
- `npm run typecheck` — all workspaces passed after a clean dependency install and build
- `npm run lint` and `npm run format:check` — passed
- CLI release pack and offline smoke validation — passed
- Packaged CLI installed, restarted, inspected, and cleanly uninstalled a real systemd user service; the service survived the setup SSH session and uninstall retained the State Root

</details>

<details>
<summary>中文</summary>

- `npm --workspace maka-agent test` — 269 项测试通过
- `npm run build` — 通过
- `npm run typecheck` — 干净安装依赖并构建后，所有 workspace 通过
- `npm run lint` 与 `npm run format:check` — 通过
- CLI release pack 与离线 smoke validation — 通过
- 使用发布形态 CLI 在真实 systemd user 环境完成安装、重启、状态查询和干净卸载；service 不依赖 setup SSH 会话存活，卸载后 State Root 保留

</details>

## AI use

Select exactly one:

Tool(s) and scope: Codex assisted with implementation, tests, documentation, and validation under maintainer direction. The commit includes the required `Generated-by: Codex` trailer.

## Checklist

- [ ] No

### DIFF
--- a/packages/cli/src/runtime-host-service-manager.ts
+++ b/packages/cli/src/runtime-host-service-manager.ts
@@ -1,14 +1,21 @@
 import { randomUUID } from 'node:crypto';
 import { createServer } from 'node:net';
-import { mkdir, open, readFile, realpath, rename, rm } from 'node:fs/promises';
-import { dirname, isAbsolute, join, resolve } from 'node:path';
+import { mkdir, open, readFile, realpath, rename, rm, stat } from 'node:fs/promises';
+import { homedir } from 'node:os';
+import { dirname, isAbsolute, join, relative, resolve, sep } from 'node:path';
 import {
   PROJECT_DIRECTORY_MAX_ROOTS,
   PROJECT_DIRECTORY_ROOT_LABEL_MAX_BYTES,
+  RUNTIME_HOST_PROTOCOL_VERSION,
 } from '@maka/runtime-host/protocol';
+import { connectExistingRuntimeHost } from '@maka/runtime-host/client';
+import { withFileUpdateLock } from '@maka/storage/file-update-lock';
 
 const SERVICE_CONFIG_FILE = 'runtime-host-service.json';
 const DEFAULT_WEBSOCKET_PATH = '/runtime-host';
+const SERVICE_OPERATION_LOCK_TIMEOUT_MS = 60_000;
+const SERVICE_READY_TIMEOUT_MS = 45_000;
+const SERVICE_READY_POLL_MS = 50;
 
 export interface RuntimeHostManagedServiceConfig {
   readonly schemaVersion: 1;
@@ -25,7 +32,6 @@ export interface RuntimeHostManagedServiceConfig {
   readonly launch: {
     readonly nodePath: string;
     readonly cliPath: string;
-    readonly packageVersion: string;
   };
 }
 
@@ -47,8 +53,9 @@ export interface RuntimeHostServiceBackendStatus {
 }
 
 export interface RuntimeHostServiceBackend {
+  readonly operationLockPath: string;
   preflightInstall(): Promise<void>;
-  install(config: RuntimeHostManagedServiceConfig): Promise<void>;
+  install(config: RuntimeHostManagedServiceConfig): Promise<RuntimeHostServiceDeployment>;
   status(): Promise<RuntimeHostServiceBackendStatus>;
   start(): Promise<void>;
   stop(): Promise<void>;
@@ -56,8 +63,11 @@ export interface RuntimeHostServiceBackend {
   uninstall(): Promise<void>;
 }
 
+export interface RuntimeHostServiceDeployment {
+  rollback(): Promise<void>;
+}
+
 export interface RuntimeHostManagedServiceStatus extends RuntimeHostServiceBackendStatus {
-  readonly configured: boolean;
   readonly config: RuntimeHostManagedServiceConfig | null;
 }
 
@@ -86,11 +96,16 @@ export interface RuntimeHostManagedServiceInput {
   readonly websocketPath?: string;
   readonly nodePath: string;
   readonly cliPath: string;
-  readonly packageVersion: string;
 }
 
 interface RuntimeHostServiceManagerDeps {
   readonly allocateLoopbackPort: () => Promise<number>;
+  readonly waitForReady: (
+    config: RuntimeHostManagedServiceConfig,
+    backend: RuntimeHostServiceBackend,
+  ) => Promise<void>;
+  readonly environment: NodeJS.ProcessEnv;
+  readonly homeDir: string;
 }
 
 export class RuntimeHostServiceManagerError extends Error {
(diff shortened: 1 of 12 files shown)


Astro-Han (commented):
The platform-neutral lifecycle and Linux backend are separated cleanly, and the loopback binding, exact executable paths, linger preflight, restrictive umask, and retained State Root are good boundaries. Existing reviews already cover the failure-atomic reinstall, systemd error classification, and readiness gaps, so I am not duplicating those threads.

The first-principles simplification is still one transactional launch authority: validate every path/argument first, stage config and unit together, activate, prove Runtime Host readiness, then commit—or restore the previous pair. That removes mixed states instead of adding more status heuristics. The two inline findings below are additional input-corruption/validation gaps on the latest head.

Review performed with three Codex reviewer agents and DeepSeek V4 Flash as advisory tools; I verified the novel findings against the latest head and live green CI.

平台无关 lifecycle 与 Linux backend 的分层清晰；loopback binding、精确 executable path、linger preflight、限制性 umask 和保留 State Root 都是正确边界。已有 review 已覆盖 reinstall 非原子、systemd 错误分类和 readiness，我不重复这些线程。

更符合第一性原理的简化仍是单一、事务性的 launch authority：先验证所有 path/argument，再一起 stage config 与 unit，激活并证明 Runtime Host ready，最后 commit；任何失败恢复上一对配置。这样可直接消除 mixed state，而不是增加更多 status heuristic。以下两个行内问题是最新 head 上额外的输入破坏/验证缺口。

本次审查使用了三位 Codex reviewer agents 与 DeepSeek V4 Flash 作为辅助工具；我已依据最新 head 和实时绿色 CI 复核新增问题。

</details>

packages/cli/src/runtime-host-systemd-service.ts
```diff
@@ -0,0 +1,364 @@
...
+
+function quoteSystemdArgument(value: string): string {
+  if (/[\u0000-\u001f\u007f]/u.test(value)) {
+    throw new TypeError('systemd arguments cannot contain control characters');
+  }
+  return `"${value
+    .replaceAll('\\', '\\\\')
+    .replaceAll('"', '\\"')
+    .replaceAll('%', '%%')
+    .replaceAll('$', '$$$$')}"`;
```
Astro-Han, line 501: **P2 — `$` is over-escaped in `ExecStart`.** systemd needs `$$` in the unit text to pass one literal dollar, but replacing one `$` with four produces two dollars in the launched argument. A valid root such as `/srv/$project` or label `Cash$` is silently changed. Replace with `.replaceAll('$', '$$')` and add an exact render/parse round-trip assertion for a dollar-containing argument.

packages/cli/src/runtime-host-service-manager.ts
```diff
@@ -0,0 +1,396 @@
...
+  if (!input.cliPath) {
+    throw new RuntimeHostServiceManagerError(
+      'invalid_launch',
+      'The current Maka CLI entry point could not be resolved',
+    );
+  }
+  const requestedRoot = resolve(input.rootPath ?? previous?.rootPath ?? input.defaultRootPath);
+  const projectDirectoryRoots = await Promise.all(
+    (input.projectDirectoryRoots ?? previous?.projectDirectoryRoots ?? []).map(
+      async ({ label, path }) => ({ label, path: await realpath(path) }),
```
Astro-Han: **P2 — Canonicalizing project roots is not enough to make them valid Host roots.** `realpath()` accepts regular files and duplicate/symlink-equivalent directories, while `HostProjectDirectoryAuthority` rejects both. Install can therefore persist config and report a successful restart before the Host exits into a systemd restart loop. Validate that every canonical root is a directory and that canonical paths are unique before writing config/unit; cover a file path and duplicate canonical roots.

packages/cli/src/runtime-host-systemd-service.ts
M4n5ter: Thanks for flagging this. I verified that the current code is intentionally correct: JavaScript replacement strings interpret `$$` as one literal dollar, so `.replaceAll("$", "$$$$")` writes exactly `$$` into the generated unit. systemd then converts that `$$` to the intended single literal `$` in the launched argument. Changing the replacement string to `$$` would write only one dollar into the unit and would not preserve the argument. The existing render assertion confirms the generated `Cash$$` form, so no code change is needed here.

_Codex assisted with the technical verification and wording of this maintainer-directed reply._

packages/cli/src/runtime-host-systemd-service.ts
```diff
@@ -0,0 +1,497 @@
...
+  removeRuntimeHostServiceFile,
+  RuntimeHostServiceManagerError,
+  type RuntimeHostManagedServiceConfig,
+  type RuntimeHostServiceBackend,
+  type RuntimeHostServiceBackendStatus,
+  type RuntimeHostServiceDeployment,
+  writeRuntimeHostServiceFile,
+} from './runtime-host-service-manager.js';
+
+const SYSTEMD_UNIT_NAME = 'maka-runtime-host.service';
```
Astro-Han: **P2 — Scope the unit name to the configuration/profile authority.** Every invocation uses the same `maka-runtime-host.service`, while CLI profiles (including the dev CLI) can use different `clientDataRoot` values and therefore different managed-service JSON. Installing/status/stopping/uninstalling from one profile can overwrite or control another profile's unit/process. Derive a stable unit identity from the profile/data root, or reject service management outside the supported default profile; add a dev/release mixed-profile regression.

Astro-Han (commented):
The latest head resolves the earlier failure-atomic install, readiness/PID ownership, project-root canonicalization, and systemd error-reporting findings. One remaining authority collision prevents approval: the persisted config is profile/data-root scoped, but the systemd unit identity is global to the Unix user.

The first-principles fix is to give the managed service the same identity as its configuration authority. Either derive a stable unit name from the Maka profile/data-root identity, or explicitly reject service management from non-release/non-default profiles. One profile must never control another profile's process.

Reviewed with Codex using two independent review passes and DeepSeek V4 Flash as an external adversarial pass; I verified the identity paths against this exact head and current `main`.

最新 head 已修复之前的原子安装、readiness/PID 归属、project root 规范化和 systemd 错误投影问题。仍有一个权威冲突：配置按 profile/data root 隔离，但 systemd unit 对同一 Unix 用户是全局的。

最小修复是让 service identity 与配置 authority 一致：unit 名包含稳定的 profile/data-root identity，或明确禁止非发布/非默认 profile 使用 service management。一个 profile 不能控制另一个 profile 的进程。

本次由 Codex 两轮独立审查，并使用 DeepSeek V4 Flash 做外部对抗审查；我核对了当前 head 与最新 `main`。

</details>

Astro-Han (approved):
The current head resolves the earlier service-identity collision and the previous lifecycle blockers. The managed unit is now scoped to the Client Data Root identity; install/reinstall is failure-atomic; readiness proves Runtime Host ownership; project roots are canonicalized and validated; and systemd failures remain failures rather than being projected as `not_installed`. The implementation now has one coherent configuration/service authority without adding a parallel lifecycle path.

I found no reproducible P0–P3 issue on this head. All current required checks are green.

AI-assisted review disclosure: Codex performed the final review using two independent reviewer passes and OpenCode Go DeepSeek V4 Flash (high effort) as an adversarial advisory pass. I verified the identity, rollback, readiness, and validation paths against exact head `fc93a2167` and live CI.

当前 head 已解决此前的 service identity 冲突和 lifecycle 阻塞项。managed unit 现在与 Client Data Root 身份一致；install/reinstall 具备失败原子性；readiness 会验证 Runtime Host 归属；project root 会做规范化和目录校验；systemd 失败也不会再被投影成 `not_installed`。整体保持了单一、连贯的配置与服务权威，没有引入并行 lifecycle。

当前 head 未发现可复现的 P0–P3 问题，所有必需检查均已通过。

AI 辅助审查说明：Codex 使用两轮独立 reviewer 审查，并以 OpenCode Go DeepSeek V4 Flash（high effort）进行对抗性辅助审查；我已针对精确 head `fc93a2167` 与实时 CI 复核 identity、rollback、readiness 和 validation 路径。

</details>

PR #3210, opened 2026-08-18
feat(cli): add one-shot user commands

### LINKED ISSUE #3209: feat(cli): add one-shot user commands
## Problem

The Maka TUI has no way for a user to run a local command without opening another terminal or asking the model to call Bash. An interactive `/shell` would make the TUI another terminal owner, introducing PTY input, resize, recovery, and terminal-restoration lifecycle that is unnecessary for one command.

A user command and a model Bash call are different capabilities. The command and its output must not be represented as a model tool result or become readable through a Runtime resource reference.

## Desired outcome

Provide a Codex-style `!<command>` interaction in the idle TUI that:

- runs one command in the current Session working directory through Runtime Host;
- keeps the TUI rendering instead of handing the terminal to an interactive shell;
- shows the resource result only as a local user-command card and does not create an agent turn;
- retains a terminal resource update that reaches the client before its card exists;
- stores the run with `visibility: "user"` and makes Runtime reject model reads of that resource, while allowing the client projection to display it.

Users can open another terminal or ask the model to use Bash. The former breaks flow; the latter intentionally creates a model-visible agent tool result. An interactive `/shell` is broader than this problem and duplicates terminal ownership in the TUI.

me2seeks:
## Summary

Adds idle TUI `!<command>` execution through the selected Runtime Host and current Session workspace. The command appears as a local user-command card and does not create an agent turn. Submitting bare `!` shows `Usage: !<command>`; `!!pwd` executes `!pwd`.

**Runtime boundary.** Runtime Host remains the execution and resource-lifecycle authority. One-shot user commands persist `visibility: "user"`. Runtime model-facing Read, Write, and Stop return the same non-disclosing not-found result used for inaccessible resources, while the owning Client can inspect and stop the exact command resource.

**Lifecycle boundary.** Host admission rechecks that the Session is active and resolves its authoritative working directory before launch. A failed post-launch inspection triggers best-effort cleanup. The CLI driver owns observation ordering and stops active user commands before Session switch, new Session, close, or driver disposal; TUI reconnect preserves local command cards without turning them into transcript messages.

**Protocol compatibility.** `runtime.resource.start` accepts an optional bounded one-shot `command`, and durable Shell Run records carry `visibility`. `RUNTIME_HOST_COMPATIBILITY_EPOCH` is 57, so epoch-56 peers are rejected during admission rather than failing on the widened wire shape.

**Privilege.** `runtime.resource.start` already belongs to `REMOTE_OWNER_OPERATION_GRANTS` for the interactive login shell. This change adds no grant and no CLI-owned executor.

Fixes #3209

<details>
<summary>中文说明</summary>

TUI 空闲时输入 `!<command>`，命令会由当前选中的 Runtime Host 在当前 Session 工作目录执行，并显示为本地 User command 卡片，不创建 agent turn。提交裸 `!` 会显示 `Usage: !<command>`；`!!pwd` 会执行 `!pwd`。

Runtime Host 继续拥有执行、工作目录准入和资源生命周期。一次性用户命令持久化为 `visibility: "user"`：模型侧 Read、Write、Stop 均得到非披露的 not-found；拥有该资源的 Client 仍可检查和停止它。

Host 在真正启动前会在同一准入边界内再次确认 Session 仍为 active，并读取权威工作目录；启动后检查失败会尽力停止资源。CLI driver 在切换 Session、新建 Session、关闭或释放前停止活动命令。重连会保留本地命令卡片，但不会把它写进对话 transcript。

协议增加可选且有长度上限的 `command`，持久化 Shell Run 增加 `visibility`，兼容 epoch 升为 57；epoch 56 的旧对端会在准入时被拒绝。

</details>

## Verification

- Node 24.18.1: full repository `npm test` passed with all workspace suites at 0 failures.
- Affected builds passed: Core, Storage, MCP, Eval, Runtime, Runtime Host, and CLI.
- Focused suites passed: Runtime Shell Run manager (60 passed, 4 Windows-only skipped), Runtime Host resource protocol/coordinator (24 passed), CLI driver/transcript/TUI (all passed; TUI 150/150).
- Scoped Biome checks and `git diff --check` passed.
- Regression coverage includes model/client visibility separation, Read/Write/Stop non-disclosure, Session archive during queued admission, launch cleanup, stop-before-switch lifecycle, rejected-stop recovery, fast terminal-update replay, reconnect card preservation, bare `!`, and `!!command`.

## AI use

Select exactly one:

Tool(s) and scope: Codex prepared the implementation, tests, Issue, and PR description. The human contributor reviews the diff and remains responsible for submission, correctness, provenance, and licensing.

## Checklist

- [x] Lint, format, typecheck, focused suites, and the full repository suite pass locally

- [ ] No

## Visual evidence

Not included. The prior screenshots represented an earlier implementation with an inline localized hint that is intentionally outside this reconstructed scope.


packages/runtime-host/src/server/runtime-resource-coordinator.ts
```diff
@@ -329,26 +329,27 @@ export class HostRuntimeResourceCoordinator
...
       const launched = await this.runBackgroundBash({
         sessionId: input.sessionId,
         sourceTurnId: input.launchId,
         sourceToolCallId: input.launchId,
+        visibility: 'user',
         cwd: header.cwd,
         command,
         env,
-        pty: true,
+        pty: input.command === undefined,
```
Astro-Han: [P1] Please provide a cancellation/teardown path for pipe-backed user commands. `runBackgroundBash` receives neither a timeout nor a client lifecycle owner here, while TUI shutdown only calls `driver.stop()`, which returns when there is no agent turn. A command such as `!sleep 3600` therefore survives TUI exit, retains Runtime Host residency, and can consume live-shell capacity with no UI left to stop it. The existing `runtime.resource.stop` client path is the natural owner: retain active user-command refs in the CLI, stop them on close, and expose the same stop path from Ctrl+C while a card is running. Add coverage for both start-in-flight and already-running close races.

packages/cli/src/pi-tui-runner.ts
```diff
@@ -1340,6 +1366,19 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
     requestRender();
   };
 
+  const runUserCommand = async (command: string): Promise<void> => {
+    if (!input.driver.runUserCommand) {
+      throw new Error('User commands are unavailable on this session driver.');
+    }
+    const started = await input.driver.runUserCommand(command);
+    appendUserCommandToTranscript(state, { command, ...started });
```
Astro-Han, line 1523: [P2] This card is only appended to transient transcript state, so a Runtime Host reconnect removes it. The reconnect handler replaces the transcript from durable model messages, then resource hydration can only update an already-existing card; it never recreates this user-owned entry. A running command can consequently finish successfully while its output becomes permanently invisible. Please preserve these local cards across same-session replacement or reconstruct them from the authoritative user-visible resource snapshots, with a reconnect → terminal-update regression test.

packages/cli/src/pi-transcript.ts
```diff
@@ -1546,6 +1571,10 @@ function findToolEntry(
     );
 }
 
+function isShellRunToolCard(tool: MakaPiToolEntry): boolean {
+  return tool.toolName === 'Bash' || tool.userOwned === true;
```
Astro-Han, line 1999: [P2] Treating user-owned cards as generic shell-run cards also routes their running → terminal transition through `pushShellRunSettledNotice`, producing an extra `Background task completed/failed` notice beside the `User command` card. That is duplicate output with the wrong ownership semantics. Please exclude `userOwned` cards from the background-task settlement notice and add a focused running → completed/failed assertion.

Astro-Han (commented):
The core direction is sound: Runtime Host remains the sole execution and resource-lifecycle authority, and the CLI adds only a local projection. I found three concrete lifecycle/projection gaps that should be addressed before merge: commands have no cancellation or close-time cleanup, reconnect discards their only UI projection, and terminal updates reuse the model-background-task notice. The smallest design is to keep Runtime Host authoritative while having the CLI retain active user-command refs, stop them through the existing client-owned resource API, and reconstruct their cards from user-visible resource snapshots after reconnect.

AI-assisted review: Codex coordinated two independent reviewer passes and an OpenCode Go DeepSeek V4 Flash high-effort adversarial pass. I verified these findings against exact head `313edbbcf18f961714c116ebb21669f313e299cb`, the relevant call paths, and current CI. No local tests were run in this review.

<details>
<summary>中文审查</summary>

整体方向正确：执行权和资源生命周期仍由 Runtime Host 统一负责，CLI 只增加本地投影。当前有三个需要修复的具体缺口：用户命令没有取消或退出清理路径；重连会删除唯一的 UI 卡片；终态更新会错误复用模型后台任务通知。最小方案是不新增执行权，而是让 CLI 保存活跃 user-command ref，通过已有 client-owned resource stop API 清理，并在重连后从用户可见资源快照重建卡片。

本次为 AI 辅助审查：Codex 协调两轮独立 reviewer 审查及 OpenCode Go DeepSeek V4 Flash high 对抗性复核，并核验了精确 head、相关调用链和当前 CI。本轮未运行本地测试。

</details>

Astro-Han:
This PR adds a new user-visible TUI flow. Could you please add a screenshot showing a ! command in the idle TUI and the resulting User command card, including its completed output state? One annotated screenshot is fine. Thanks!

me2seeks:
Added the requested screenshot to the PR body. It shows an idle `!echo maka-user-ok` command and the completed, expanded User command card with output. No source or commits on this branch were changed.

packages/cli/src/pi-transcript.ts
```diff
@@ -281,7 +284,7 @@ export function applyShellRunViewUpdateToTranscript(
   }
   if (
     !tool ||
-    tool.toolName !== 'Bash' ||
+    !isShellRunToolCard(tool) ||
```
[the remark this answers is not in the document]
me2seeks: Fixed in exact head 2c65cc52f. User-owned cards are now excluded from the generic background-task settlement notice, and the focused running-to-terminal transcript test asserts that no notice entry is appended.

packages/cli/src/pi-transcript.ts
```diff
@@ -304,11 +307,33 @@ export function applyShellRunUpdateToTranscript(
...
 
+/** Adds a local-only card for a `!<command>` resource without creating a model turn. */
+export function appendUserCommandToTranscript(
+  state: MakaPiTranscriptState,
+  input: { commandId: string; command: string; result: ShellRunSnapshotResult },
+): void {
+  state.entries.push({
+    kind: 'tool',
+    toolUseId: input.commandId,
+    toolName: 'User command',
```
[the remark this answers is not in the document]
me2seeks: Fixed in exact head 2c65cc52f. Same-session reconnect replacement now preserves the local user-owned cards, while ordinary session replacement still drops them; a reconnect-to-terminal-update TUI regression verifies the output remains visible.

packages/runtime-host/src/server/runtime-resource-coordinator.ts
me2seeks: Fixed in exact head 2c65cc52f. The CLI tracks active user-command refs and stops them through runtime.resource.stop on Ctrl-C and driver close. Tests cover both an already-running command and the start-in-flight versus close race. The Node 24 CLI, Runtime, Runtime Host, and Core suites all pass locally.

packages/cli/src/pi-tui-runner.ts
me2seeks: Fixed in exact head 2c65cc52f. Same-session transcript replacement preserves transient user-owned cards so later authoritative resource updates still have a projection target; session switches continue to clear them. The reconnect-to-terminal-update regression passes in the full CLI suite.

packages/cli/src/pi-transcript.ts
me2seeks: Fixed in exact head 2c65cc52f. Settlement notices now remain limited to model/background-task cards; user-owned command completion stays only in its User command card. The focused transcript regression asserts the absence of a duplicate notice.

packages/cli/src/runtime-host-session-driver.ts
```diff
@@ -272,6 +283,92 @@ class RuntimeHostMakaSessionDriverImpl implements RuntimeHostMakaSessionDriver {
...
+    } catch (error) {
+      if (capture) this.#shellRunListeners.delete(capture);
+      throw error;
+    } finally {
+      releaseStartBarrier?.();
+      this.#userCommandStartBarriers.delete(startBarrier);
+    }
+  }
+
+  stopUserCommands(): Promise<void> {
```
Astro-Han, line 448: [P2] Stop or re-project live user commands when leaving their Session

`stopUserCommands()` is reached on Ctrl+C only while the current transcript still contains a running user-owned card, and on full driver shutdown. Normal `/session`, `/new`, and rewind transitions replace the transcript without preserving those cards and do not stop the tracked resources. A reproducible `!sleep 3600` → `/new` therefore leaves the old Session command running invisibly; `hasRunningUserCommand()` becomes false, so Ctrl+C can no longer stop it before CLI exit. Please either await the existing start-barrier-aware stop path before changing Session identity, or reconstruct user-owned cards from that Session’s authoritative resource snapshots with an explicit stop affordance. Add a running-command → switch/new regression.

Astro-Han (commented):
The current-head rereview confirms that the prior teardown, same-session reconnect, and duplicate-settlement findings are fixed. The actual terminal capture also matches the current UX: completed user commands are expanded by default and the bare `!` hint is visible. All six commits carry `Generated-by: Codex`, and the PR body discloses the substantive scope.

One nonduplicate session-transition P2 remains below. Exact-head CI has not run yet because the Apache Actions lanes are still queued/unavailable, so this head is not merge-ready independently of the finding.

AI-assisted review by OpenAI Codex. I verified the range-diff from the previously reviewed head, lifecycle and projection paths, existing threads, current screenshot, provenance, and live check state.

当前 head 已修复上一轮的退出清理、同 Session 重连和重复终态通知问题；真实终端截图也与当前 UX 一致。6 个提交均保留 `Generated-by: Codex`，PR body 的 AI 披露完整。

下面仍有一个新的 Session 切换 P2。当前 exact-head CI 还没有实际运行，因此即使不考虑 finding，也尚不可合并。

本次由 OpenAI Codex 辅助，已核对上一轮 head 的 range-diff、生命周期与投影路径、已有线程、截图、来源披露和实时检查状态。

</details>

packages/cli/src/runtime-host-session-driver.ts
me2seeks: Fixed in `32253928a`.

`switchSession()` now awaits the existing start-barrier-aware `stopUserCommands()` path before changing Session identity (covers `/session` and rewind, which routes through it), and `startNewSession()` synchronously bumps the stop generation/pending count before clearing identity so an in-flight start self-stops on resolution while already-active commands are stopped via the serialized stop tail. `#stopUserCommand` addresses resources by their captured owner `sessionId`, so the stop still lands on the old Session after the identity change.

Added two regressions in `runtime-host-session-driver.test.ts`: running-command → `switchSession` (asserts `runtime.resource.stop` for the old Session before the switch completes) and running-command → `startNewSession`. `@maka/cli`: 328/328 pass; Biome clean.

packages/runtime-host/src/protocol/runtime-resource.ts
```diff
@@ -223,13 +227,27 @@ export const RUNTIME_RESOURCE_OPERATION_SPECS = {
 } as const;
 
 export function decodeRuntimeResourceStartInput(value: unknown): RuntimeResourceStartInput {
-  const input = requireExactRecord(value, 'Runtime Resource start input', [
-    'sessionId',
-    'launchId',
-  ]);
+  const input = requireShapedRecord(
```
Astro-Han, line 249: [P1] Bump `RUNTIME_HOST_COMPATIBILITY_EPOCH` — this widens a protocol input without one. `decodeRuntimeResourceStartInput` moves from `requireExactRecord(…, ['sessionId','launchId'])` to `requireShapedRecord(…, ['sessionId','launchId'], ['command'])`, and `protocol/index.ts:75` is still `27`, the same value as `main`. A post-PR CLI and a pre-PR remote Runtime Host therefore pass admission — the epochs match — and then the first `!ls` sends `command`, which the old host's `requireExactRecord` -> `assertExactKeys` rejects as an unknown field, surfacing as an opaque `invalid_request` rather than the version refusal the epoch exists to produce. Epoch 27's own comment states the rule for exactly this case. Note 28 is currently contested by #3236, #3199 and #3133, all taking 28 from a base at 27, so take 28 and re-check it at merge time. Confirmed by reading the source at this head. Regression test needed: a decode test that an epoch-27 peer's `requireExactRecord` path rejects `command`, plus the constant itself.

packages/cli/src/runtime-host-session-driver.ts
```diff
@@ -471,6 +568,12 @@ class RuntimeHostMakaSessionDriverImpl implements RuntimeHostMakaSessionDriver {
       summary = runtimeHostSessionSummary(session);
       await assertSessionResumeAvailable(summary, this.#executionLocation);
     }
+    // Leaving the current Session must not orphan its live user commands:
+    // the switch replaces the transcript, so their cards and the Ctrl+C stop
+    // affordance would disappear while the commands keep running. Await the
+    // start-barrier-aware stop path before changing Session identity so an
+    // in-flight start cannot land after the switch (#3210).
+    await this.stopUserCommands();
```
Astro-Han: [P1] Move this above the cwd relocation, or make its failure non-fatal — as placed it can abandon a Session in a half-switched state. `#commitCwdRelocation` at `:560` is durable and already committed by the time control reaches here. `stopUserCommands()` returns a promise that rejects whenever any `#stopUserCommand` rejects, and `#stopUserCommand` (`:1177-1193`) awaits `runtime.resource.stop` with no catch, so a `host_draining` or a transport failure propagates straight out of `switchSession`. The Session identity is then never changed — `#openSessionChannel` below never runs — but the user's session has already been moved to the new cwd. The command also stays in `#activeUserCommands` with no retry and nothing surfaced. Confirmed by reading the source at this head. Regression test needed: `switchSession` with `relocateCwd` set and a `runtime.resource.stop` that rejects; assert either the switch completes or the relocation did not commit.

packages/cli/src/pi-tui-runner.ts
```diff
@@ -3036,6 +3081,16 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
...
       return { consume: true };
     }
+    if (
+      !turnRunning &&
+      matchesKey(data, Key.ctrl('c')) &&
+      hasRunningUserCommand(state) &&
+      input.driver.stopUserCommands
+    ) {
+      lastIdleCtrlCAt = 0;
+      void runControl(() => input.driver.stopUserCommands!());
```
Astro-Han: [P2] Do not reset `lastIdleCtrlCAt` on a path that can stay armed forever. This branch sits above the idle double-Ctrl+C exit at `:3139-3150` and clears the chord timestamp unconditionally, and `hasRunningUserCommand(state)` only goes false when a terminal resource update arrives. `#stopUserCommand` publishes that update *after* `runtime.resource.stop` resolves, so if the stop rejects, nothing is published, the card reads running for the rest of the session, and Ctrl+C is captured permanently — the user never even sees "Press Ctrl+C again to exit." A milder version happens on every stop: while the RPC is in flight `runControl` returns immediately on `if (busy) return` (`:639`), so each Ctrl+C in that window is consumed with zero feedback, which against a remote host is a full round trip. Ctrl+D still exits, so this is a lost affordance rather than a hang. The new `Ctrl-C stops a running user command without exiting the TUI` test cannot catch it: its fake `stopUserCommands()` only increments a counter and never publishes a terminal update, and the test presses Ctrl+C once. Confirmed by reading the source at this head. Regression test needed: with a rejecting `stopUserCommands`, assert a second Ctrl+C shows the exit prompt and a third exits.

packages/cli/src/runtime-host-session-driver.ts
```diff
@@ -618,12 +728,21 @@ class RuntimeHostMakaSessionDriverImpl implements RuntimeHostMakaSessionDriver {
 
   async stop(): Promise<void> {
     const turn = this.#channel?.snapshot.rootTurn;
-    if (!turn || isTerminalTurn(turn)) return;
-    await this.#request('turn.stop', {
-      sessionId: turn.sessionId,
-      turnId: turn.turnId,
-      runId: turn.runId,
-    });
+    const stops: Promise<unknown>[] = [this.stopUserCommands()];
```
Astro-Han: [P2] Do not let a user-command stop failure be reported as a failed turn interrupt. `stop()` is the turn-interrupt authority — `pi-tui-runner.ts:805` reaches it from double-Escape and from Ctrl+C while `turnRunning` — and this line puts `stopUserCommands()` unconditionally into the same `Promise.allSettled` whose first rejection is rethrown at `:745`. Two consequences. First, `turn.stop` can succeed while the user-command stop fails, and the caller sees the interrupt itself as failed: `requestTurnInterrupt`'s catch resets `interruptRequested = false`, re-enables submit, and reports the error, so the turn is silently re-armed for interruption. Second, interrupting the model turn now also kills any independently running `!` command, with nothing in the UI saying so — a `!` command is by design not part of the turn. Confirmed by reading the source at this head. Regression test needed: running turn plus running user command, interrupt with a rejecting `runtime.resource.stop`, and assert the turn interrupt is not reported as failed.

packages/cli/src/pi-transcript.ts
```diff
@@ -314,17 +323,48 @@ export function applyShellRunUpdateToTranscript(
...
+  const userCommands = options.preserveUserCommands
+    ? state.entries.filter(
+        (entry): entry is MakaPiToolEntry => entry.kind === 'tool' && entry.userOwned === true,
+      )
+    : [];
   const view = materializeSession(messages);
-  state.entries = foldStoredShellRunChildren(view.items.flatMap(chatItemToTranscriptEntries));
+  state.entries = [
+    ...foldStoredShellRunChildren(view.items.flatMap(chatItemToTranscriptEntries)),
+    ...userCommands,
```
Astro-Han: [P3] Preserved cards are re-appended at the tail, so a reconnect reorders the transcript. `state.entries` is rebuilt as stored-messages-then-`userCommands`, so a `!` command run before several model turns jumps to the bottom after a same-session reconnect and the transcript stops reading chronologically. Cosmetic, and strictly better than the erasure this replaced. The covering test (`preserves local user-command cards only for same-session reconnect replacement`) passes an empty message list, so ordering is never exercised — that is what would need to change to pin a fix. Confirmed by reading the source at this head.

packages/cli/src/skill-highlight-editor.ts
```diff
@@ -38,13 +44,27 @@ export class MakaSkillHighlightEditor extends Editor {
...
+    const lines = super
       .render(width)
       .map((line) =>
         line.replace(pattern, (whole, name: string) =>
           this.isInvocable(name) ? ansi.accent(whole) : whole,
         ),
       );
+    if (this.getText() !== '!' || !this.userCommandHint) return lines;
+
+    const cursor = '\x1b[7m \x1b[0m';
```
Astro-Han: [P3] This string-matches a private rendering detail of `@earendil-works/pi-tui`. `'\x1b[7m \x1b[0m'` is what `Editor.render` emits only when the cursor sits at end-of-line (`node_modules/@earendil-works/pi-tui/dist/components/editor.js:442`); with the cursor over a grapheme it emits `\x1b[7m<grapheme>\x1b[0m` instead (`:436`). For a bare `!` the end-of-line form is always the one produced, so the hint works today — but a dependency bump that changes the glyph or adds an attribute makes `findIndex` return -1 and the hint silently disappears. The covering test asserts the rendered string, so it would fail loudly; the undeclared coupling is the point, and a comment naming the pi-tui version this was read from would be enough. Confirmed by reading the source and the installed dependency at this head.

packages/runtime-host/src/server/runtime-resource-coordinator.ts
```diff
@@ -344,36 +344,37 @@ export class HostRuntimeResourceCoordinator
...
-      } else {
+      } else if (command === undefined) {
         const executable = (shell.exe ?? shell.displayName).replace(/'/g, "''");
         command = `& '${executable}' -NoLogo`;
       }
       const launched = await this.runBackgroundBash({
         sessionId: input.sessionId,
         sourceTurnId: input.launchId,
         sourceToolCallId: input.launchId,
+        visibility: 'user',
```
Astro-Han: Contract decision rather than a defect, so no P level. This sets `visibility: 'user'` on **every** `runtime.resource.start`, not only the command-carrying ones this PR adds — including the Desktop-owned interactive login shell, which reaches this same call with `input.command === undefined` and `pty: true` two lines down. Before this PR that PTY had no `visibility`, so the model could read it through `readRuntimeResource`, stop it through `stopBackgroundTask`, and see it in the context summary; all three now return the non-disclosing result. The coordinator test was amended to assert `visibility: 'user'` for `desktop-launch-1`, so this is deliberate — but it is a user-visible change to what the model can see in a Desktop session, it is not in the PR body's behaviour section, and it is a second intent that one revert would undo along with the feature. Worth either scoping the flag to `input.command !== undefined` or stating the Desktop change explicitly. Confirmed by reading the source at this head.

Astro-Han (commented):
COMMENT. One P1 that I think blocks the merge, plus four smaller findings and a contract question.

Scope note first: I reviewed this at `787ad6aae`, after the rebase and the `fix(cli): stop user commands before leaving their Session` commit. The start-barrier and stop-generation machinery in `stopUserCommands` is a real improvement over the previous head, and the two threads I opened earlier — cancellation/teardown at `runtime-resource-coordinator.ts:352` and stop-on-session-change at `runtime-host-session-driver.ts:355` — are both genuinely addressed. Two of the findings below are residuals of those fixes rather than repeats of them.

The P1 is the compatibility epoch. This PR widens `runtime.resource.start`'s input to accept `command` and leaves `RUNTIME_HOST_COMPATIBILITY_EPOCH` at 27, which is what `main` has, so a new CLI and an old remote Runtime Host still shake hands and then fail on the first `!` command. Details inline. The same PR also adds `visibility` to the durable `ShellRunRecord` (`packages/core/src/shell-run.ts:117`, key set at `:305`), which an older binary reading the same workspace operational-state DB rejects as a malformed field — and `listSessionShellRuns` maps over the whole session, so one such row takes out that session's entire shell-run projection. Same root cause, same fix.

Two things I could not confirm and am not filing:

- The close path. `beginClose` (`pi-tui-runner.ts:695-704`) fires `void input.driver.stop().catch(() => {})` and returns, so `/exit` with a running `!sleep 3600` does not await the `runtime.resource.stop` before the caller tears the connection down, and `beginClose` skips `driver.stop()` entirely when `interruptRequested` is already set. Against a managed host that leaves the command running and holding a `maxLiveShellRuns` slot. I am reasonably confident this is real but I could not construct the failure end to end at this head, so I am recording it as an observation on the existing teardown thread rather than as a numbered finding.
- Whether `runtime.resource.start` being in `REMOTE_OWNER_OPERATION_GRANTS` is a privilege change. I do not think it is — a remote owner could already start the login-shell PTY, acquire a controller and type anything into it — but the new surface is non-interactive and scriptable, which is a different exposure profile. An explicit "no new privilege" line in the PR body would settle it.

Test gaps, none of which I am giving a P level:

- `runtime-resource-coordinator.test.ts`'s `starts a one-shot user command in pipes without exposing it to the model` asserts only the `ShellRunBashInput` handed to a fake `runBackgroundBash`; the harness's `readRuntimeResource` has no visibility gate, so the test cannot prove non-exposure. The real coverage is in `shell-run-manager.test.ts`; the name overclaims.
- `stops a running user command before switching Sessions (#3210)` asserts that a `runtime.resource.stop` exists in the request log, not that it preceded the identity change — which is the actual claim.
- `UserCommandDriver` in `pi-tui-runner.test.ts` returns `status: 'completed'` from `runUserCommand`, so the running-to-terminal card transition is never exercised in the runner suite.
- The PR body's "removes `/shell`" is stale: neither `/shell` nor `openInteractiveShell` exists on `main` or in this diff.

`biome format` is clean on every changed TS file at this head.

AI disclosure: this review was assisted by Claude (Opus), which performed the initial code search and cross-checking. Every finding published here I re-derived myself against the source at `787ad6aae`, and I dropped the ones I could not confirm at that head — the two observations above are marked as such rather than filed.

packages/runtime-host/src/protocol/runtime-resource.ts
me2seeks: Fixed in `d7552f796`. `RUNTIME_HOST_COMPATIBILITY_EPOCH` is now 28 with a comment covering both sides of the widening (the optional `command` on `runtime.resource.start` and the durable `visibility` on Shell Run records). Regression test added in `runtime-resource-protocol.test.ts`: it pins the constant at 28, asserts the epoch-27 `requireExactRecord` decode path rejects `command`, and asserts a command-less start still decodes. Noted on 28 being contested — will re-check at merge time.

packages/cli/src/runtime-host-session-driver.ts
me2seeks: Fixed in `d7552f796` by moving the stop above the relocation. `stopUserCommands()` now runs right after the execution-boundary check, before `#commitCwdRelocation`: a rejecting `runtime.resource.stop` aborts the switch with nothing durable committed and the driver still owns the original Session. Regression test: `a rejecting user-command stop aborts the switch before any durable relocation commits (#3210)` asserts the switch rejects, no `session.workspace.relocate` request was ever sent, and `getSessionId()` is unchanged.

packages/cli/src/pi-tui-runner.ts
me2seeks: Fixed in `d7552f796`. The Ctrl+C branch now runs the stop through a guarded action: on rejection it sets a `userCommandStopRejected` latch (and reports the error as a notice), which disarms the capture so later presses fall through to the idle double-Ctrl+C exit chord. Regression test: `a rejected user-command stop hands Ctrl-C back to the exit chord (#3210)` — with a rejecting `stopUserCommands`, the first press is captured, the second shows "Press Ctrl+C again to exit.", and the third exits.

packages/cli/src/runtime-host-session-driver.ts
me2seeks: Fixed in `d7552f796`. `stop()` now wraps `stopUserCommands()` in `.catch(() => undefined)` before the `Promise.allSettled`, so only a `turn.stop` rejection can fail the interrupt; the user-command stop stays best-effort on this path because `stop()` is also the close authority that tears commands down on TUI exit. On the second consequence: the stopped command's card does flip to its terminal state via the published resource update, so the kill is visible — and the callers that own user-command lifecycle (Ctrl+C, Session switch) await `stopUserCommands()` directly and surface its errors. Regression test: `a rejecting user-command stop does not fail the turn interrupt (#3210)` — running turn plus running user command, rejecting `runtime.resource.stop`, and `stop()` resolves.

packages/cli/src/pi-transcript.ts
me2seeks: Fixed in `d7552f796`. Preserved cards are now re-inserted at their chronological position: each card's shell-run `startedAt` is compared against the materialized ChatItem timestamps and the card is placed before the first later item, with a final fold that cannot touch user-owned cards (folding only merges a stored shell-run child into a Bash parent sharing its ref). The new test `reconnect re-inserts preserved user-command cards at their chronological position (#3210)` passes a non-empty message list and asserts the card stays ahead of the later turn.

packages/cli/src/skill-highlight-editor.ts
me2seeks: Fixed in `d7552f796`. Added a comment at the match site naming the coupling: the glyph is pi-tui's private end-of-line cursor rendering, read from @earendil-works/pi-tui 0.83.0 (`components/editor.js:442`), with a note that the covering test fails loudly if a bump changes it.

packages/runtime-host/src/server/runtime-resource-coordinator.ts
me2seeks: Scoped in `d7552f796`. `visibility: 'user'` is now set only when `input.command !== undefined`; the Desktop interactive login shell reaches the same call without a `command` and keeps its prior model-visible behavior, so this PR no longer changes what the model can see in a Desktop session. The coordinator test is amended back to assert `visibility: undefined` for `desktop-launch-1`, while the one-shot command test keeps asserting `visibility: 'user'`. The PR body now also states the no-new-privilege point on `REMOTE_OWNER_OPERATION_GRANTS` explicitly.

me2seeks:
Rebased onto current main; new head `a87fc5229`.

- The epoch moved 28 → 29 because main landed the relay Fast service-tier change as 28 while this was in flight. The comment and the gate test were renumbered (the test now asserts `>= 29` relative to the constant instead of an exact value, so the next advance cannot silently pass this gate).
- The tip commit's trailer now follows CONTRIBUTING's `Generated-by: <tool>` format; the epoch reference in its message was updated to match.
- Local suites on the rebased branch: runtime-host 1031/1031 (the previously failing epoch assertion fixed by the renumber), cli 353/353, core 558/558, desktop main green.

Everything else from the 2026-08-20 11:31 review is unchanged from `d7552f796`: the widened start input is gated by the epoch, `visibility` is scoped to command-carrying starts, stop failures are isolated from turn interrupt and abort a Session switch before anything is committed, and reconnect re-inserts preserved cards chronologically.

<details>
<summary>中文说明</summary>

已 rebase 到最新 main；epoch 因 main 占用 28 而推进到 29，gate 测试改为相对断言，tip 提交补上合规的 Generated-by trailer。本地套件全绿。11:31 review 的全部修复保持不变，等待重新 review。

</details>

me2seeks:
Rebased onto current `main` (`2e5fe9c55`, 8 commits replayed). Resolutions:

1. **Compatibility epoch** → **40**: main landed 33–39 while this was in flight; this PR's wire delta (one-shot `command` on `runtime.resource.start` + Shell Run `visibility`) is now the 40 entry, both comment lists preserved.
2. Mid-turn submit routing (`pi-tui-runner.ts`): main's `/transcript` branch and this PR's user-command refusal are kept as sequential checks.
3. ASF header on `skill-highlight-editor.ts` (#3397) merged with this PR's widened pi-tui import.

Verified on new head `8d9da7d46`: full root build, CLI 399/399, runtime-host 1092/1092 (handshake-compatibility gate asserts relative to the constant), biome clean.

Astro-Han:
Hi — this PR was behind `main` and could not be merged as-is. Since `maintainer_can_modify` was enabled, I rebased it onto current `main` and force-pushed the result; the rebase applied with **zero conflicts**, so your changes are unchanged in content (author attribution is preserved — only the commit SHAs moved).

Please `git fetch && git reset --hard <your-remote>/<branch>` before your next push so you do not re-introduce the old base.

CI should now run against the rebased head. I will pick this up for review once the checks are terminal.

---
*AI-assisted maintenance action. It is not a review and does not count as the required human review under `CONTRIBUTING.md` §Review.*

packages/runtime-host/src/server/runtime-resource-coordinator.ts
```diff
@@ -363,36 +363,40 @@ export class HostRuntimeResourceCoordinator
       const header = await this.#sessionHeaders.readHeader(input.sessionId);
       const shell = await this.#resolveShell();
       const env = { ...process.env };
-      let command: string;
-      if (shell.kind === 'git-bash') {
+      let command = input.command;
```
ARE404: [P2] Keep the Session header read inside the same admission section as launch. The new one-shot path captures `header.cwd` at line 363, then awaits shell resolution before `runBackgroundBash()` acquires Session admission. A concurrent `session.workspace.relocate` can therefore commit `/new` during that gap, after which this command is admitted with stale `/old`. Reproduction: pause `#resolveShell`, relocate the Session, then resume; `!pwd` runs in the old cwd, violating the current-Session-cwd contract. Please acquire admission before reading the header (or re-read it inside the admitted launch), and cover this interleaving.

packages/runtime-host/src/server/runtime-resource-coordinator.ts
```diff
@@ -363,36 +363,40 @@ export class HostRuntimeResourceCoordinator
...
         command = `& '${executable}' -NoLogo`;
       }
       const launched = await this.runBackgroundBash({
         sessionId: input.sessionId,
         sourceTurnId: input.launchId,
         sourceToolCallId: input.launchId,
+        // Only the one-shot `!<command>` resources this Client owns are hidden
+        // from the model; the Desktop interactive login shell (no `command`)
+        // keeps its prior model-visible visibility (#3210).
+        ...(input.command === undefined ? {} : { visibility: 'user' as const }),
```
ARE404: [P2] Make start plus initial snapshot atomic from the client's perspective, or clean up on post-launch failure. This one-shot command is already live after `runBackgroundBash()` returns at line 388, but the handler then performs a separate `inspectResource()` at line 407. If that inspection/encoding throws, the operation reports failure while the process keeps running; the CLI rejects `runUserCommand`, removes its pre-card capture, and never creates a card. Retrying executes the command twice. A focused harness where launch succeeds and inspect throws reproduces the invisible live resource. Return the launch snapshot directly if sufficient, or stop the launched ref before returning failure.

packages/cli/src/runtime-host-session-driver.ts
```diff
@@ -597,6 +702,13 @@ class RuntimeHostMakaSessionDriverImpl implements RuntimeHostMakaSessionDriver {
   }
 
   startNewSession(): void {
+    // `/new` replaces the transcript without preserving user-command cards,
+    // so a still-running command would lose both its projection and its
+    // Ctrl+C stop affordance. Stop tracked commands before the identity
+    // change; the generation/pending bump is synchronous, so an in-flight
+    // start self-stops when it resolves even though this method stays sync
+    // (#3210).
+    void this.stopUserCommands().catch(() => undefined);
```
ARE404: [P2] Do not clear Session identity and the only command projection when this stop rejects. `/new` immediately replaces the transcript, while this fire-and-forget call suppresses `runtime.resource.stop` failures. With `!sleep 3600` active and a rejected stop (transport loss/host draining), `startNewSession()` still sets `#sessionId = null`; the old command keeps running without its card or Ctrl+C affordance. This is the residual failure path of the earlier stop-on-Session-change fix. Make `/new` awaitable and abort the transition on stop failure, or preserve/reconstruct the old resource projection, with a rejecting-stop regression.

packages/cli/src/skill-highlight-editor.ts
```diff
@@ -57,13 +63,34 @@ export class MakaSkillHighlightEditor extends Editor {
...
     const pattern = new RegExp(SKILL_INVOCATION_TOKEN_SOURCE, 'g');
-    return super
+    const lines = super
       .render(width)
       .map((line) =>
         line.replace(pattern, (whole, name: string) =>
           this.isInvocable(name) ? ansi.accent(whole) : whole,
         ),
       );
+    if (this.getText() !== '!' || !this.userCommandHint) return lines;
```
ARE404: [P3] Please add the documented before/after UI evidence for this visible inline-hint change. `CONTRIBUTING.md` requires before/after screenshots or a recording for UI changes; the PR currently has only an after-state capture, and it is labeled from older head `ecfa72feb` rather than the current exact head.

ARE404 (commented):
Exact-head review at `f0149f9d5ea6d191f0708ec7b98c1da35de95f20`: **NO-GO**. Spec axis: three P2 lifecycle/atomicity findings inline. Standards axis: one P3 evidence gap inline; no hard code-standard violation. The permission review found no exposed new grant or model confirmation bypass: remote owners already had the broader interactive-shell path, while model Read/Write/Stop and context-summary paths are gated for `visibility: user`. The only path-filtered check expected here is `test`; it is terminal **failure** because `packages/cli/src/pi-tui-runner.ts` has a formatter-rejected blank line at EOF. No audit or Windows sandbox check is expected for these paths.

me2seeks:
Rebased onto `main` (`c79e9eb4`) with the three [P2]s now implemented on top (`720151a7`):

1. **Stop-on-`/new` atomicity** — `startNewSession` is awaited end-to-end. The driver runs the barrier-aware user-command stop before any identity mutation and rejects on failure; the runner catches it, posts the reason as an error notice, and leaves transcript/cards/state untouched. Regression tests cover both layers (driver: rejection keeps `getSessionId()`; runner: notice + no wipe). The control gate lets /new through while a user command runs — stopping them is its first step — fenced by the existing generation guards.
2. **Start+snapshot atomicity** — when the post-launch `inspectResource` fails, the launched process is stopped (best-effort) and the operation reports failure, so a retry cannot double-execute. Harness test asserts `stopCount === 1` and `ok === false`.
3. **cwd admission gap** — header read, shell resolution, launch, and initial snapshot now share one Session admission section; the relocate race can no longer admit a stale cwd.

Verified locally: CLI workspace 439/439, runtime-host 1127 green including the new regression tests; epoch guard 44 → 45 passes against the new base. The [P3] before/after captures will be refreshed from this head once CI is green.

Astro-Han:
Heads-up on CI, not a review finding: **this head's CI run is red on the formatter, and it fails before any test runs.**

Its checks had never executed — the run was parked awaiting maintainer approval. I released it (`32687618896`), so the result is real and not a new regression from your rebase.

**Where it stops**: the `Check formatting` step of the `test` job (`npm run format:check`).

**What it found** — reproduced locally at `720151a700ae9940011fa489b68fdf4ad3303b2d`, 3 errors across 1605 files, in three files:

- `packages/cli/src/pi-transcript.ts`
- `packages/cli/src/pi-tui-runner.ts`
- `packages/runtime-host/src/server/runtime-resource-coordinator.ts`

One of them is a line-length wrap Biome wants split:

```
- env.SHELL ||= userInfo().shell || (process.platform === 'darwin' ? '/bin/zsh' : '/bin/sh');
+ env.SHELL ||=
+   userInfo().shell || (process.platform === 'darwin' ? '/bin/zsh' : '/bin/sh');
```

plus one stray blank line before a `} catch (error) {`.

**`npm run format` fixes all three**; commit the result and the rest of the pipeline gets a chance to run. Note that `Lint` and `Check formatting` sit ahead of `Build`, `Typecheck` and the test steps, so **there is currently no test signal for this head** — the three [P2] fixes you say you implemented on top of the rebase have not actually been exercised by CI yet.

<details>
<summary>中文</summary>

CI 提示，不是 review finding：**这个 head 红在格式化检查上，而且它在任何测试之前就失败了。**

它的检查此前从没执行过——run 卡在等放行，我已放行（`32687618896`），所以这个红是真实结果，不是你这次 rebase 引入的新回归。

**停在哪**：`test` job 的 `Check formatting`（`npm run format:check`）。

**报了什么**（我在 `720151a7` 上本地复现）：1605 个文件里 3 个错，分布在 `packages/cli/src/pi-transcript.ts`、`packages/cli/src/pi-tui-runner.ts`、`packages/runtime-host/src/server/runtime-resource-coordinator.ts`。其中一处是 Biome 要求换行的长行（`env.SHELL ||= …`），另有一处 `} catch (error) {` 前多余的空行。

**`npm run format` 可以全部修好。** 注意 `Lint` 和 `Check formatting` 排在 `Build`、`Typecheck` 和测试步骤之前，所以**这个 head 目前没有任何测试信号**——你说在 rebase 之上实现的那三条 [P2] 修复，CI 还完全没有验证过。
</details>

Astro-Han (commented):
Update on `f6139efc15`:

**[P1] `RUNTIME_HOST_COMPATIBILITY_EPOCH` must strictly increase — still 49**

PR claims epoch bumped to handle new `command` + `visibility` fields, but head `f6139e` sets `RUNTIME_HOST_COMPATIBILITY_EPOCH = 49` which is already used by main for WorkHub (also 49). Rebased PR re-used main's occupied epoch instead of allocating a fresh one (needs 50). A main-built Host (epoch 49 WorkHub) will accept a PR-built Client but reject the new `command` key / `visibility` record — incompatible change not fenced.

Fix: bump epoch to a new value strictly greater than main's 49 and keep comment load-bearing to force conflict.

Also `mergeStateStatus=CONFLICTING/DIRTY` — needs rebase.

Checks on `f6139efc15ce735e8878e2313d704db7060b8153` are `test: SUCCESS` only — not full green.

<details><summary>简体中文</summary>epoch 未真正递增，已占用，与 main 冲突。</details>

likun666661 (approved):

PR #3223, opened 2026-08-18
fix(eval): close metering admission at child exit

me2seeks:
## Summary

The external-subject metering proxy stayed open for new admissions after the child process had exited while `runChild()` was still draining stdout and stderr. A leftover descendant or established keep-alive connection could issue another provider request in that window, so the supposedly post-exit request was counted instead of refused.

This closes proxy admission synchronously from the child `exit`/`error` boundary. `report()` and `close()` reuse the same idempotent transition, then drain only requests admitted before that cut.

The failure is reproducible on current `main`; PR #3208's otherwise unrelated workspace lane failed this exact regression with `2 !== 1`.

## Verification

- focused provider-admission regression: 50/50 consecutive passes
- `npm --workspace @maka/eval run build`
- `npm --workspace @maka/eval run test:dist` — 75/75 Node tests; all Python Harbor suites pass (4 environment-dependent cases skipped as expected)
- `npx biome check packages/eval/src/harbor-external-subject.ts`
- `git diff --check`
- `simplify-audit`: no P0-P3 candidates; `accepting` remains the single admission authority

AI disclosure: Codex implemented and verified this change under me2seeks's direction and review.

### DIFF
--- a/packages/eval/src/__tests__/provider-admission-integration.test.ts
+++ b/packages/eval/src/__tests__/provider-admission-integration.test.ts
@@ -324,11 +324,21 @@ test('a request arriving while the proxy drains is refused, not counted', async 
     `import { readFileSync, writeFileSync } from 'node:fs';
 const [, , baseUrl, outcomePath, childExitedPath] = process.argv;
 const held = fetch(\`\${baseUrl}/responses\`, { method: 'POST', body: '{}' }).catch(() => undefined);
-for (;;) {
+const deadline = Date.now() + 5_000;
+let sawChildExit = false;
+while (Date.now() < deadline) {
   try {
-    if (JSON.parse(readFileSync(childExitedPath, 'utf8')).phase === 'child_exited') break;
+    if (JSON.parse(readFileSync(childExitedPath, 'utf8')).phase === 'child_exited') {
+      sawChildExit = true;
+      break;
+    }
   } catch {}
   await new Promise((resolve) => setTimeout(resolve, 10));
+}
+if (!sawChildExit) {
+  await held;
+  writeFileSync(outcomePath, JSON.stringify({ status: -2 }));
+  process.exit(2);
 }
 let status = 0;
 try {
--- a/packages/eval/src/__tests__/provider-admission-integration.test.ts
+++ b/packages/eval/src/__tests__/provider-admission-integration.test.ts
@@ -314,14 +314,22 @@ test('a request arriving while the proxy drains is refused, not counted', async 
   assert.ok(address && typeof address !== 'string');
   const child = join(root, 'child.mjs');
   const straggler = join(root, 'straggler.mjs');
-  // Outlives its parent, holds one request open across the subject's exit, then
-  // starts another -- the shape of a background service the agent left running.
+  const childExited = join(root, 'logs/agent/opencode.wrapper-state.json');
+  // Outlives its parent and holds the child's stdout pipe open, keeping the
+  // wrapper inside runChild until the exit callback publishes its barrier.
+  // The second request therefore cannot race ahead of that callback, while a
+  // report-time-only admission cut deadlocks instead of satisfying the test.
   await writeFile(
     straggler,
-    `import { writeFileSync } from 'node:fs';
-const [, , baseUrl, outcomePath] = process.argv;
+    `import { readFileSync, writeFileSync } from 'node:fs';
+const [, , baseUrl, outcomePath, childExitedPath] = process.argv;
 const held = fetch(\`\${baseUrl}/responses\`, { method: 'POST', body: '{}' }).catch(() => undefined);
-await new Promise((resolve) => setTimeout(resolve, 300));
+for (;;) {
+  try {
+    if (JSON.parse(readFileSync(childExitedPath, 'utf8')).phase === 'child_exited') break;
+  } catch {}
+  await new Promise((resolve) => setTimeout(resolve, 10));
+}
 let status = 0;
 try {
   status = (await fetch(\`\${baseUrl}/responses\`, { method: 'POST', body: '{}' })).status;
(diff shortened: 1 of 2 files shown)


Astro-Han (commented):
The production fix is at the right authority and is smaller than alternative socket or reporting changes: admission closes synchronously in the child `error`/`exit` callbacks, while already-admitted requests still drain into the settled snapshot. I found no production-code regression. The remaining P2 is in the verification layer: the existing regression test schedules the purported post-exit request with a blind 300 ms timer, so loaded CI can still send it before the child exit callback and reproduce `2 !== 1` even with this fix. Because this PR is specifically a flake fix, the regression should pin the exit boundary deterministically.

AI-assisted review: Codex coordinated two independent reviewer passes and an OpenCode Go DeepSeek V4 Flash high-effort adversarial pass. I verified the exact head `4990f314f4324848e6a0eacc307651f9cc3abff0`, child/proxy event ordering, the existing provider-admission integration test, and current CI. No local tests were run in this review.

<details>
<summary>中文审查</summary>

生产代码修复落在正确责任层，也比关闭 socket 或推迟到 report 的方案更简单：child 的 `error`/`exit` 回调同步停止 admission，已经接收的请求继续 drain 到最终快照。未发现生产代码回归。剩余 P2 在验证层：现有测试用固定 300ms 定时器制造“退出后”请求，CI 调度变慢时它仍可能在 child exit callback 前发送，从而让修复后的代码继续出现 `2 !== 1`。这个 PR 本身就是 flake fix，因此回归测试应确定性锁定 exit boundary。

本次为 AI 辅助审查：Codex 协调两轮独立 reviewer 审查及 OpenCode Go DeepSeek V4 Flash high 对抗性复核，并核验了精确 head、child/proxy 事件顺序、现有 provider-admission integration test 和当前 CI。本轮未运行本地测试。

</details>

packages/eval/src/harbor-external-subject.ts
```diff
@@ -468,8 +469,14 @@ async function runChild(
...
   const stderr = captureStreamDiagnostic(running.stderr);
   const exitCode = await new Promise<number>((resolveExit, reject) => {
-    running.once('error', reject);
-    running.once('exit', (code) => resolveExit(code ?? 1));
+    running.once('error', (error) => {
+      onExit();
+      reject(error);
+    });
+    running.once('exit', (code) => {
+      onExit();
```
Astro-Han: [P2] Please add a deterministic regression pin for this exact boundary. The current integration fixture sends its second request after a fixed 300 ms delay; if request 1 or child scheduling takes longer on CI, request 2 is legitimately pre-exit and is still counted, producing the same `2 !== 1` failure on this head. Drive the second request from a barrier emitted after this callback has called `onExit` (or add a focused test around an injected child-exit event), then verify base fails and head passes. That distinguishes this fix from a weaker report-time admission cut and removes the residual timer race.

me2seeks: Fixed in 444a0bbe2. The straggler now holds the child stdout pipe open and waits for the existing child_exited wrapper-state barrier. The exit callback closes proxy admission before publishing that barrier, so the second request is deterministically post-callback. Reverting to a report-time-only cut makes the same test time out at 8s; the fixed test passes in about 1.1s. Eval passes 75/75 plus all Python suites (4 expected skips).

packages/eval/src/__tests__/provider-admission-integration.test.ts
```diff
@@ -314,14 +314,22 @@ test('a request arriving while the proxy drains is refused, not counted', async
...
+    `import { readFileSync, writeFileSync } from 'node:fs';
+const [, , baseUrl, outcomePath, childExitedPath] = process.argv;
 const held = fetch(\`\${baseUrl}/responses\`, { method: 'POST', body: '{}' }).catch(() => undefined);
-await new Promise((resolve) => setTimeout(resolve, 300));
+for (;;) {
+  try {
+    if (JSON.parse(readFileSync(childExitedPath, 'utf8')).phase === 'child_exited') break;
+  } catch {}
+  await new Promise((resolve) => setTimeout(resolve, 10));
+}
```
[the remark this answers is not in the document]
me2seeks: Fixed in 8cca51e45. The detached straggler now waits at most five seconds for the child_exited barrier. If the barrier never appears, it waits for the held request, writes an explicit failing outcome, and exits nonzero, closing the inherited stdout instead of blocking wrapper drain indefinitely. The provider-admission integration file passes 5/5.

Astro-Han (approved):
Approved. Child exit/error is the earliest truthful metering-admission boundary: stopAccepting closes the single proxy authority synchronously before the child_exited barrier is published, while already admitted requests remain in the active set and drain into the final report. The settled guard deduplicates error/exit, and report/close reuse the same idempotent transition.

The latest test no longer guesses with a fixed delay: it waits for the child-exit barrier, proves an in-flight request is retained, and proves a post-exit request is rejected. Current head is clean against main and all required checks are green.

AI-assisted review by Codex with two independent reviewers and OpenCode Go DeepSeek V4 Flash (high); I verified lifecycle ordering, late/in-flight requests, error cleanup, test discrimination, and live CI.

批准。child exit/error 是最早且真实的 metering admission 边界：stopAccepting 在发布 child_exited barrier 前同步关闭唯一 proxy authority；已受理请求仍保留在 active 集合中并进入最终 report。settled guard 去重 error/exit，report/close 复用同一个幂等 transition。

最新测试不再依赖固定延迟，而是等待 child-exit barrier，同时证明在途请求保留、exit 后请求拒绝。当前 head 与 main 干净合并，必需检查全绿。

本次由 Codex、两个独立 reviewer 和 OpenCode Go DeepSeek V4 Flash high 辅助；已核对生命周期顺序、late/in-flight requests、错误清理、测试判别力和实时 CI。

</details>

PR #2684, opened 2026-08-11
feat(runtime): support explicit Git Bash selection on Windows

me2seeks:
## Summary

- add a Host-owned `auto` / `git_bash` shell preference while keeping the existing PowerShell-first Windows detection unchanged by default
- validate an explicit `bash.exe` on the Runtime Host before persisting it, then use the resolved `ShellPlan` for Bash tools, child agents, background and PTY execution, the integrated terminal, and model-facing shell guidance
- add the Desktop setting and preserve older Runtime Policy documents through the schema migration; the legacy `System32\bash.exe` WSL shim uses stdin for non-PTY commands

The executable path is intentionally resolved on the Runtime Host, not the Desktop Client. This keeps remote Desktop connections honest and makes an invalid or later-missing explicit path fail closed instead of silently returning to PowerShell.

Fixes #2195

### Desktop

Git Bash setting in Desktop

## Verification

- `npm run build:test`
- `npm run test:dist:serial` — all workspace tests passed
- `npm run typecheck`
- `npx biome check .`
- `git diff --check`
- rendered and inspected `Product/Settings/Pages/General Git Bash` in Storybook at 1440×1000

## Checklist

- [x] Tests cover the change and fail without it
- [x] Lint, format, typecheck and the affected suites pass locally

Does this PR entail a change in behavior?

- [x] Yes — described under Summary above
- [ ] No

<details>
<summary>中文对照</summary>

## 概要

- 增加由 Runtime Host 持有的 `auto` / `git_bash` shell 偏好；默认仍保持现有的 Windows PowerShell 优先检测
- 显式选择时，由 Runtime Host 在持久化前验证 `bash.exe`，并将解析出的同一个 `ShellPlan` 用于 Bash 工具、子 Agent、后台与 PTY 执行、集成终端以及面向模型的 shell 方言说明
- 增加 Desktop 设置，并通过 schema 迁移兼容旧 Runtime Policy 文档；旧版 `System32\bash.exe` WSL shim 的非 PTY 命令通过 stdin 输入

可执行文件路径刻意在 Runtime Host 而不是 Desktop Client 上解析。这样远程 Desktop 连接不会误用客户端机器的路径；显式路径无效或之后消失时也会失败关闭，而不会静默回退到 PowerShell。

</details>

## AI use

- [ ] No generative tool was used for implementation.
- [x] Generative tooling was used and the result was reviewed and verified by the author.

[...]

PR #3224, opened 2026-08-18
test(runtime): stabilize proxied body deadline coverage

me2seeks:
## Summary

The default production `proxiedFetch` body-deadline test uses one timeout for both response setup and the intentionally stalled body read. At 50 ms, loaded runners can expire before response headers arrive, so the test fails at `fetchForConnectionEffect()` instead of exercising the body-read deadline it is meant to cover.

On current `main`, the unchanged test passed only 13/20 isolated runs; the #2684 worktree passed 16/20, confirming this is a baseline timing flake rather than a shell-selection regression.

This raises only the default production `proxiedFetch` test budget to 250 ms. The explicitly constructed direct-transport body-deadline test remains at 50 ms, preserving its tighter coverage.

The test does not install an active proxy: with the module default of `proxy = null`, this production wrapper reaches the response through direct Undici.

## Verification

- focused default `proxiedFetch` regression: 50/50 consecutive passes
- `npm --workspace @maka/runtime test` — 2877 total / 2871 pass / 0 fail / 6 skip
- `npx biome check packages/runtime/src/network/__tests__/scoped-fetch-transport.test.ts`
- `git diff --check`
- `simplify-audit`: no P0-P3 candidates; the two timeout values intentionally cover different setup boundaries

AI disclosure: Codex implemented and verified this test stabilization under me2seeks's direction and review.

### DIFF
--- a/packages/runtime/src/network/__tests__/scoped-fetch-transport.test.ts
+++ b/packages/runtime/src/network/__tests__/scoped-fetch-transport.test.ts
@@ -180,7 +180,9 @@ describe('connection effect network transport', () => {
       const response = await fetchForConnectionEffect(
         undefined,
         `http://127.0.0.1:${port}/models`,
-        { timeoutMs: 50 },
+        // Leave enough time for the production proxy to establish the response;
+        // the assertion is that its deadline remains armed while the body stalls.
+        { timeoutMs: 250 },
       );
       await assert.rejects(
         response.readJson(),
--- a/packages/runtime/src/network/__tests__/scoped-fetch-transport.test.ts
+++ b/packages/runtime/src/network/__tests__/scoped-fetch-transport.test.ts
@@ -180,8 +180,8 @@ describe('connection effect network transport', () => {
       const response = await fetchForConnectionEffect(
         undefined,
         `http://127.0.0.1:${port}/models`,
-        // Leave enough time for the production proxy to establish the response;
-        // the assertion is that its deadline remains armed while the body stalls.
+        // Leave enough time for the default production proxiedFetch path to
+        // receive the response; its deadline must remain armed while the body stalls.
         { timeoutMs: 250 },
       );
       await assert.rejects(


packages/runtime/src/network/__tests__/scoped-fetch-transport.test.ts
```diff
@@ -180,7 +180,9 @@ describe('connection effect network transport', () => {
       const response = await fetchForConnectionEffect(
         undefined,
         `http://127.0.0.1:${port}/models`,
-        { timeoutMs: 50 },
+        // Leave enough time for the production proxy to establish the response;
```
Astro-Han: [P3] Please call this the default production `proxiedFetch` path rather than saying a production proxy is being established. This test never calls `setActiveProxy`; with the module default of `null`, `proxiedFetch` builds no proxy dispatcher and uses direct Undici. The test still provides useful coverage of the real default wrapper and its body deadline, so this is a wording/scope clarification rather than a blocking code issue.

Astro-Han (approved):
Approved at exact head `a6233cd18dce3ed91ec2eab00a6d015c2cf82f1c`. Raising only the default production-fetch-path budget preserves the tighter injected-transport assertion while still checking that the outer deadline remains active during a stalled body read. No P0–P2 findings. I left one non-blocking P3 clarification because no active proxy is installed in this test.

AI-assisted review: Codex coordinated two independent reviewer passes and an OpenCode Go DeepSeek V4 Flash high-effort adversarial pass. I verified the exact diff and traced `fetchForConnectionEffect(undefined, ...)` through `proxiedFetch` and active-proxy resolution. No local tests were run in this review; current CI was still completing when reviewed.

<details>
<summary>中文审查</summary>

已审查精确 head `a6233cd18dce3ed91ec2eab00a6d015c2cf82f1c`。仅放宽默认生产 fetch 路径的预算，同时保留注入 transport 的 50ms 紧约束，仍能验证外层 deadline 在 body 卡住时持续生效。未发现 P0–P2；留了一条不阻塞的 P3 表述建议，因为测试没有安装 active proxy。

本次为 AI 辅助审查：Codex 协调两轮独立 reviewer 审查及 OpenCode Go DeepSeek V4 Flash high 对抗性复核，并追踪了 `fetchForConnectionEffect(undefined, ...)`、`proxiedFetch` 和 active-proxy resolution。本轮未运行本地测试；审查时当前 CI 尚未全部结束。

</details>

packages/runtime/src/network/__tests__/scoped-fetch-transport.test.ts
me2seeks: Fixed in f1a503f29. The inline comment and PR description now call this the default production proxiedFetch path and describe receiving the response through direct Undici, without implying that an active proxy dispatcher is established. Runtime passes 2877 total (2871 pass, 6 skip).

Astro-Han (approved):
Approved. The only change since the previously approved head is a comment clarifying that the test exercises the default production proxiedFetch path. The 250 ms budget preserves the discriminating behavior: headers must arrive before the deadline, the body remains deliberately incomplete, and readJson must still fail with the specific timeout classification. No production code or assertion strength changed.

Current head is clean against main and all required checks are green. AI-assisted review by Codex with one independent reviewer and OpenCode Go DeepSeek V4 Flash (high); I verified the exact delta, production wrapper path, deadline ownership, and live CI.

批准。相对上次已批准 head，唯一新增改动只是注释澄清该测试覆盖默认生产 proxiedFetch 路径。250 ms 仍保持测试判别力：headers 必须在 deadline 前到达，body 故意不完成，readJson 仍必须以明确的 timeout 分类失败。生产代码和断言强度都没有变化。

当前 head 与 main 干净合并，必需检查全绿。本次由 Codex、一个独立 reviewer 和 OpenCode Go DeepSeek V4 Flash high 辅助；已核对精确 delta、生产 wrapper、deadline ownership 和实时 CI。

</details>

Issue #3213, opened 2026-08-18
chore(runtime): Code Mode is on @ai-sdk/code-mode now — delete superseded custom scaffolding, consolidate byte counting, fold the package into runtime


Astro-Han:
Status update from the owner: Code Mode is active in-flight work, and the migration to the QuickJS approach has **already landed** — `packages/code-mode/src/quickjs.ts` is a 157-line adapter over `@ai-sdk/code-mode`'s `experimental_runCodeMode` (deps: `@ai-sdk/code-mode@1.0.23`, `ai@7.0.66`). What remains in the package is custom scaffolding that predates or duplicates what the SDK now owns.

## Deletable now

1. **`CodeModeLimits` + `DEFAULT_CODE_MODE_LIMITS`** (`packages/code-mode/src/index.ts:149-172`): a 1:1 rename mapping onto the SDK's `executionPolicy` (`maxSandboxTimeMs`→`timeoutMs`, etc.). Use the SDK's policy shape directly. The only consumer outside the package is `packages/runtime/src/ai-sdk-backend.ts:3082,3114` reading `maxToolOutputBytes` as a constant — keep one constant.
2. **The hand-rolled execution queue** (`index.ts:214-264`, ~50 lines): global one-cell-at-a-time serialization with a queue depth of one. The SDK already governs concurrency (`CODE_MODE_CONCURRENCY_LIMIT`, `maxInFlightBridgeRequests`). Unless there is a measured constraint that QuickJS instances must be serially reused, this is duplicate governance. Confirm the SDK's concurrency semantics before removing.
3. **`serializedByteLength`** (`index.ts:10-147`, ~140 lines): cannot be deleted outright — `packages/runtime/src/tool-runtime.ts:1372` uses it in production — but it exposes a wider duplication: the repo has **three hand-written JSON byte counters** (`@maka/code-mode`, `packages/core/src/additional-permissions.ts:195`, `packages/core/src/sandbox-boundary.ts:755`), plus the SDK's own byte-limit enforcement. One "serialized byte length" fact, four authorities. Consolidate to a single implementation in `@maka/core`.

## Cascade

After the above, the package is ~200 lines of genuine integration glue (the `callTool` bridge, fatal-tool-error handling, drain-to-quiescence semantics — keep those). At that size the standalone workspace loses its reason to exist: fold the remainder into `@maka/runtime` and delete the package manifest, exports, tsconfig, root/desktop/CLI build wiring, stale-dist check, and release-dependency entries.

[...]

PR #3225, opened 2026-08-18
refactor: retire SDK-superseded code-mode scaffolding and rehome cell admission

Astro-Han:
## Summary

`packages/code-mode` predates `@ai-sdk/code-mode@1.0.23` taking over most of what it did. This retires what the SDK genuinely superseded, and moves the one thing it cannot to the side that owns execution.

1. **`CodeModeLimits` → the SDK's `CodeModeExecutionPolicy` shape** — it was a 1:1 rename mapping. Product defaults unchanged, still shipped as a constant, and overrides now fail closed.
2. **Three hand-written JSON byte counters → one** at `@maka/core/serialized-byte-length`.
3. **Cell admission moves to `AiSdkBackend`**, on the runtime's existing `AdmissionLimiter`, leaving the adapter stateless.
4. **The package folds into `@maka/runtime`.** At 224 lines of glue with one consumer, a standalone workspace cost more than it explained.

Part of #3213; the `toolMode = 'code_mode'` producer remains out of scope.

## Why admission moved rather than disappeared

This PR first deleted the hand-rolled execution queue as duplicate governance. Review showed the premise was wrong, and the finding reproduced: with the worker cap pinned to one and each cell cancelled while its tool ignores cancellation, host operations outstanding went **1 → 4**.

The queue's admission covered a cell's complete lifecycle, drain included. The SDK worker cap cannot: on cancellation `runCodeMode` releases its worker and rejects at once, by design, while host operations may still be running with durable side effects. Only Maka waits for those, so only Maka can bound them.

Two things were wrong with where it lived:

- **It was never named.** `codeCellActive` read as "one cell at a time"; what it provided was full-lifecycle admission, because the flag spanned the whole promise and the drain sits inside it. That gap is what made it look superseded — and the concept *had* been recognised once, in a `fix(code-mode): bound execution admission` subcommit of #2549 whose wording never reached the code. The word `admission` appeared zero times in the package.
- **Its scope was an accident of module loading.** A module-level flag is shared by every session in the process, so one session's cell made another session's cell queue and a third fail outright. Host operations run through a session's own ToolRuntime; the session is the granularity whose side effects need bounding.

It now lives on `AiSdkBackend`, which owns cell execution and reaches those host operations through `scope.toolRuntime`. Acquire and release sit in one method around one call, so the permit covers exactly what `executeCodeCell` promises to finish — a contract the adapter now states explicitly. Composed end to end, the same repeated-cancellation probe holds at **1**.

It is not a new primitive. `ChildAgentRunLimiter` was already an abort-aware FIFO permit pool with a capacity; it is renamed `AdmissionLimiter` and now serves two boundaries at two lifetimes — child runs one instance per turn, cells one per backend. The only difference, turning a cell away instead of queueing it, is a product decision that reads better at the call site than inside a permit pool: the caller checks `waitingCount` before `acquire`, with no await between, so the pair is atomic.

## The fold

`index.ts` and `quickjs.ts` merge into `packages/runtime/src/code-mode.ts` — the split only existed to give the package an entry point distinct from its implementation. Tests move alongside.

Twelve wiring sites drop their entry: the workspace manifest, tsconfig and project reference, both root build chains, the desktop and CLI build chains, the stale-dist pairs, the release script's internal-package and build-order lists, the CLI validation path filters, and the Windows sandbox build step. `@ai-sdk/code-mode` moves to the runtime's dependencies, so the release script's `npm ls` closure still reaches it; both third-party notice inventories already listed it under its own name and are unchanged — `check:cli-third-party-notices` and `check:third-party-notices` both pass.

One comment needed rewriting rather than deleting. The dependabot ignore for typescript majors was justified by an in-house Code Mode transpiler pinned to 5.9 for the typescript JS API. That transpiler is gone, `@ai-sdk/code-mode` owns transpilation, and no workspace pins typescript — so the ignore has no subject. The comment now says so; dropping the ignore is a dependency-policy call left to a maintainer.

## Behavior change

- **Admission is scoped to a backend, not the process.** The bound is unchanged (one active cell, one queued, the third turned away), but cells in different sessions no longer block each other. The state lives on a backend generation rather than a session, and an earlier revision of this description claimed a session could exceed the bound across a backend rebuild. No reachable path was found: a draining cell holds its turn open through `activeToolSettlements`, which holds the run open, and `clearBackendQuarantineForActivation` throws rather than activating a successor generation while a predecessor holds an active run.
- **Policy overrides fail closed.** The old boundary used `??`, so `null`/`undefined` kept the product default while zero, negatives and strings passed through to the SDK. Overrides are now admitted only for known fields and only as positive integers. This also fixes a regression this PR introduced and review caught: an earlier merge excluded only `undefined`, so `{ maxSourceBytes: null }` reached the SDK and its `??` restored the 256 KiB default in place of Maka's 64 KiB.

## Review focus

The byte counters were not semantically identical, so that consolidation is not a pure rename: the retained one is bounded and reports `Infinity` where the core wrappers threw. Both core call sites pass validated plain payloads for which the two agree byte for byte, pinned by a test.

One deliberate departure: a top-level `undefined` counts as four bytes rather than `Infinity`, because an absent result is not an oversized one — `Infinity` would make `tool-runtime` reject a tool that returned nothing as too large. Four bytes conservatively bounds what either publication path emits (a cell substitutes `null`; a tool result becomes empty text). An earlier revision of this contract named the wrong mechanism for the tool-result path and has been corrected.

## Verification

`@maka/core` 558 pass · focused runtime suites 307 pass (the 30 moved Code Mode tests included) · release policy suite 28 pass · both third-party notice checks pass · typecheck clean across `@maka/core`, `@maka/runtime`, `@maka/runtime-host`, `maka-agent` and `@maka/eval` · `format:check` and lint clean. All three review findings reproduced before fixing and re-measured after, including the composed admission probe.

The backend wiring is now covered too, which it was not: stubbing `CodeCellAdmission` to a no-op left 218 backend tests green, so re-deleting the acquire/release would not have turned CI red. Two mutations now fail the new test — a no-op admission outright, and releasing the permit before the host drain (the worker cap's semantics) by timeout. Each commit built and tested independently; repository-wide suite left to CI.

## AI use

Tool(s) and scope: Claude Code (Opus) — read the source and SDK, made the edits, ran the checks above, drafted this description. Commits carry `Generated-by: Claude Code`.

## Checklist

- [x] Yes — described above
- [ ] No


packages/core/src/serialized-byte-length.ts
```diff
@@ -0,0 +1,147 @@
...
+function countSerializedValue(
+  value: unknown,
+  budget: SerializedByteBudget,
+  position: 'top' | 'array' | 'object',
+): boolean {
+  if (value === undefined || typeof value === 'function' || typeof value === 'symbol') {
+    if (position === 'object') return false;
+    if (position === 'top' && value !== undefined) return false;
+    addSerializedBytes(budget, 4);
+    return true;
```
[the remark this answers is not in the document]
Astro-Han: Confirmed the observation, but not taking the suggested change — the behavior is load-bearing.

`packages/runtime/src/tool-runtime.ts:1371` bounds every nested tool result with this counter and throws `ToolResultLimitError` on overflow. Making a top-level `undefined` report infinity would reject a tool that simply returned nothing as though its result were too large, which is both a behavior regression and a misleading error. `null` is what callers publish in place of an absent value — a Code Mode cell through `value ?? null` in `quickjs.ts`, a tool result through `coerceResultContent` — so four bytes is what actually reaches the wire.

This is also pre-existing behavior carried over unchanged from `@maka/code-mode`; this PR is a consolidation that deliberately preserves each call site's current semantics.

The useful part of the finding was that the departure was undocumented and unpinned, so e615eec6a addresses that instead of the behavior:

- documents the departure in the counter's contract, with the reason;
- pins `serializedByteLength(undefined) === 4` in `packages/core/src/__tests__/serialized-byte-length.test.ts`;
- adds `admits a nested tool that returned nothing under the result limit` in `packages/runtime/src/__tests__/tool-runtime-durable-boundary.test.ts`, so the downstream consequence is covered rather than implied.

Verified by temporarily applying your suggested `if (position === 'top') return false;`: both tests fail, the runtime one on exactly the regression described above. Everything JSON genuinely cannot represent — functions, symbols, bigints, cycles, `toJSON` hooks, non-plain prototypes — still reports infinity, and that is now pinned too.

packages/runtime/src/code-mode.ts
```diff
@@ -75,20 +68,7 @@ export async function executeCodeCellImpl(
...
-          maxResultBytes: limits.maxOutputBytes,
-          maxConsoleOutputBytes: 1,
-          maxSourceBytes: limits.maxSourceBytes,
-          maxToolInputBytes: limits.maxToolInputBytes,
-          maxToolOutputBytes: limits.maxToolOutputBytes,
-          maxBridgeRequests: limits.maxToolCalls,
-          maxInFlightBridgeRequests: limits.maxToolConcurrency,
-        },
-      },
+      options: { executionPolicy },
```
hqhq1025, line 149: [P1] Keep admission ownership until host operations finish draining.

Delegating outer-cell admission to `run` does not preserve the guarantee provided by the removed queue. On cancellation or timeout, `runCodeMode` releases its process-global worker slot before this adapter finishes `drainHostToolOperations`. If `callTool` ignores cancellation or performs slow cleanup, another cell is admitted while the previous host operation is still running.

I reproduced this with `setMaxWorkers(1)`: repeatedly starting a tool, aborting its cell, and starting the next cell accumulated eight simultaneously retained host operations even though the worker cap was one. This can create unbounded host-side work, overlapping mutations, and turns that never finish draining.

Please keep a Runtime-owned admission permit across the complete cell lifecycle, including host-operation drainage, and add a multi-wave cancellation regression test. The SDK worker cap can remain the lower-level QuickJS bound.

packages/code-mode/src/quickjs.ts
```diff
@@ -10,25 +11,17 @@ import type {
...
-      input.limits?.maxToolInputBytes ?? DEFAULT_CODE_MODE_LIMITS.maxToolInputBytes,
-    maxToolOutputBytes:
-      input.limits?.maxToolOutputBytes ?? DEFAULT_CODE_MODE_LIMITS.maxToolOutputBytes,
-    maxOutputBytes: input.limits?.maxOutputBytes ?? DEFAULT_CODE_MODE_LIMITS.maxOutputBytes,
-  };
+  // An explicit `undefined` override keeps the product default rather than
+  // falling through to the SDK's looser one.
+  const executionPolicy: CodeModeExecutionPolicy = { ...DEFAULT_CODE_MODE_EXECUTION_POLICY };
+  for (const [key, value] of Object.entries(input.executionPolicy ?? {})) {
+    if (value !== undefined) (executionPolicy as Record<string, unknown>)[key] = value;
```
hqhq1025: [P2] Reject invalid overrides instead of falling through to looser SDK defaults.

This loop copies `null` because it excludes only `undefined`. The SDK resolves policy fields with `??`, so a runtime JavaScript or JSON caller can pass `{ maxSourceBytes: null }` and silently restore the SDK default of 256 KiB instead of Maka's 64 KiB default. The same widening affects tool output, bridge requests, in-flight requests, and console output.

I reproduced this with a 70 KiB source: the default and an explicit `undefined` override were rejected at 65,536 bytes, while the `null` override executed successfully.

Please whitelist the policy fields and accept only valid positive integers, treating both `null` and `undefined` as retaining Maka's product default. Add a regression test for runtime-invalid values.

hqhq1025 (changes requested):
## Findings

1. [P1] Removing the local queue drops full-cell lifecycle admission. The SDK worker cap is released before aborted host operations finish draining, so repeated cancellation can accumulate unbounded host-side work beyond `maxWorkers`.
2. [P2] The new generic policy merge accepts `null`, which causes the SDK's `??` resolution to silently replace Maka's tighter limits with looser SDK defaults.

## Problem and mechanism

The PR removes a lossless limits-renaming layer, delegates cell concurrency to `run`, and consolidates JSON byte counting in Core. Removing the rename layer and duplicate counters is sound. The concurrency premise is incomplete: the old admission state bounded the complete cell lifecycle, while the SDK cap bounds only active QuickJS workers.

## First principles and optimality

The solution is not yet optimal. Resource admission must cover every operation whose lifetime and side effects belong to the admitted cell. A lower-level worker permit cannot replace that ownership boundary when host work may outlive the worker after cancellation. Policy overrides must also fail closed rather than widening limits through invalid runtime values.

The simplest final structure is:

- Runtime owns bounded admission across `runCodeMode` plus host-operation drainage.
- `run` retains responsibility for its process-global QuickJS worker cap.
- The adapter validates a closed set of positive-integer policy overrides before calling the SDK.
- Core remains the single owner of package-neutral serialized byte counting.

## Deletion and tests

The legacy field-by-field rename mapping can remain deleted. Do not restore the previous queue implementation verbatim; replace it with a small full-lifecycle admission primitive at the Runtime ownership boundary.

The deleted queue-specific tests were correctly removed, but they need replacement coverage for overlapping cells, repeated abort/timeout waves, and overflow. Add a `null`/invalid-policy regression test. The new `undefined` byte-count test should describe the actual boundary policy rather than claiming that ToolRuntime publishes `null`; the SDK uses an empty sentinel and ToolRuntime durably publishes empty text content.

## Merge verdict

Not ready to merge at `e615eec6a5d5552be3d5ee371f8f8029d756f6d7`.

## Verification

I reviewed the complete diff, current dependency source, live CI, and the PR merged onto current `main`. The merged revision passed:

- Code Mode: 29 tests
- Core: 564 tests
- Focused Runtime: 37 tests
- Biome on all changed source files
- 10,000 randomized JSON-compatible byte-count comparisons

GitHub CI is green; E2E remains skipped. The cancellation and `null`-override failures were independently reproduced against the reviewed head.

packages/runtime/src/code-mode.ts
Astro-Han: Right on both counts. Reproduced before fixing: `setMaxWorkers(1)`, each cell cancelled while its tool ignores cancellation — retained host operations went **1 → 4**.

Reverted (`c902aab23`), then moved the bound to the side that owns execution (`c9f40e014`), onto the runtime's existing `AdmissionLimiter` rather than a second hand-written class (`41cdb6302`). The permit is released in a `finally` around `executeCodeCell`, which settles only after `drainHostToolOperations` — failure path included. The repeated-cancellation probe now holds at **1**.

On the multi-wave test: I don't have one worth shipping. The primitive-level version went red on the same mutation as the existing single-wave test — one flag, one slot, no accumulator, so waves add nothing. The backend-level version passed **with the admission removed entirely**, so I dropped it rather than ship a test that proves nothing.

What did land is the wiring coverage that was genuinely missing: stubbing admission to a no-op left 218 backend tests green. `1359c0b98` adds `bounds cells outstanding on one backend, across the host drain`, red under two mutations — a no-op admission, and releasing the permit before the drain, which is the worker-cap semantics you caught.

If you see a wave-loop shape I don't, I'll write it.

packages/code-mode/src/quickjs.ts
Astro-Han: Confirmed and fixed in `e8e4398d3`. This was a regression I introduced here: the original nine `?? DEFAULT.x` lines were safe because `??` covers `null` too; I rewrote them as a loop and excluded only `undefined`.

Overrides are now whitelisted to the known policy fields and admitted only as positive integers, so `null`, `undefined`, `0`, negatives and strings all keep Maka's default. Regression test in `packages/runtime/src/__tests__/code-mode.test.ts` (moved from `packages/code-mode` in `23c5354f0`).

One consequence visible in the diff: the policy is now the SDK's shape, so `maxConsoleOutputBytes` became overridable where `CodeModeLimits` had no field for it. It can only be raised from 1, and no production caller passes overrides at all. I kept it uniform rather than carve out a special case — say if you'd rather exclude it.

Astro-Han:
Rebased onto current `main`; head is `ac0ec4da3`. Both findings addressed — mechanism and evidence are in the inline replies.

- **P1** — admission moved to the side that owns execution, onto the runtime's existing `AdmissionLimiter` rather than a second hand-written class. The permit is released in a `finally` around `executeCodeCell`, which settles only after `drainHostToolOperations`, failure path included. Your `setMaxWorkers(1)` probe now holds at 1 (it went 1 → 4 before). No multi-wave test — reasoning inline; happy to write one if you have a shape in mind.
- **P2** — overrides are whitelisted to the known policy fields and admitted only as positive integers, so `null`, `undefined`, `0`, negatives and strings all keep Maka's default.

Also folded `packages/code-mode` into `@maka/runtime` here instead of as a follow-up — one consumer, no independent contract left.

`test`, `audit` and `windows_sandbox_w0_protocol` are green. `package` is red on the known Windows lane: 8 pass / 53 fail across 20 different branches over that workflow's last 100 runs. Both of my failures land in `verify-windows-autoupdate.mjs` — owned by #3265 per #3241's own scope note — and the packaged-app install and upgrade steps passed in both runs.

@hqhq1025 ready for another look.

packages/runtime/src/code-mode.ts
```diff
@@ -1,34 +1,109 @@
...
+  // default instead of the tighter one Maka ships.
+  const executionPolicy: CodeModeExecutionPolicy = { ...DEFAULT_CODE_MODE_EXECUTION_POLICY };
+  const overrides = input.executionPolicy;
+  if (overrides) {
+    for (const key of Object.keys(DEFAULT_CODE_MODE_EXECUTION_POLICY) as Array<
+      keyof CodeModeExecutionPolicy
+    >) {
+      const value = overrides[key];
+      if (typeof value === 'number' && Number.isInteger(value) && value > 0) {
+        executionPolicy[key] = value;
```
hqhq1025: [P2] Preserve the previous product override boundary

Iterating every SDK policy key makes `maxConsoleOutputBytes` newly caller-overridable, although the removed adapter always pinned it to `1` and the old `CodeModeLimits` contract did not expose it. A built probe using `{ maxConsoleOutputBytes: 100 }` allowed `console.log('hello')` from cell code to reach host stdout. The same predicate accepts integers above the SDK maximum of `2,147,483,647`; for example `maxResultBytes: 2,147,483,648` produces an `execution_error` instead of retaining Maka's default as the comment promises. Please map only the former product-supported override fields and enforce the SDK range, or remove production overrides and keep a narrower test-only injection point.

hqhq1025 (changes requested):
Re-review of exact head `ac0ec4da3578c7dc15b0a358313fef701b80f135`.

The two previously blocking issues are fixed. Admission now belongs to the backend that owns the complete cell lifecycle, and the permit remains held until sandbox execution and every started host operation have settled. The policy merge also rejects `null` and other non-positive/non-integer values instead of falling back to the SDK's looser defaults. Repeated cancellation/admission runs passed 10/10 locally, and the focused Core/Runtime suites passed 80/80.

One merge-blocking policy-boundary regression remains: replacing the former product-specific `CodeModeLimits` mapping with the SDK policy shape exposes an override that Maka previously kept fixed. The inline finding includes a direct reproducer and the related SDK-range validation gap.

Problem definition and mechanism: removing the private `@maka/code-mode` package and SDK-superseded scaffolding is sound, but Maka still owns host-operation admission and its stricter execution policy. The package fold, generalized limiter, and shared serialized-byte counter are the right first-principles boundaries. Reusing one limiter and deleting the package/build glue follow Occam's razor; no deeper architecture change is required.

Deletion/test assessment: the obsolete package and duplicate byte-counting implementation are correctly deleted. I found no additional production code that should be removed. The backend admission test and direct cancellation/drain test cover different seams and are both useful; no low-quality test should be deleted.

Merge verdict: not ready to merge until the override surface is restored to the previous product boundary and the SDK's numeric range is enforced with regression coverage. Separately, the Windows package job reached successful build, update, installer handoff, and relaunch, then failed while cleaning up an already-exited PID. That appears unrelated to this PR's code and is also the subject of PR #3327, but the required check should still be rerun before merge.

packages/runtime/src/code-mode.ts
Astro-Han: Confirmed both, and reproduced both.

Console — the field is load-bearing in a way its name hides. It is the only thing keeping cell code out of host stdout, which the CLI writes its TUI and command output to:

```
default (1)  → host stdout: []
{ …: 100 }   → host stdout: [MARKER_100_…]
```

Range — a different failure than `execution_error`. `isValidRunLimit` is `integer && > 0 && <= 2_147_483_647`, and `assertSourceSize` returns early when it fails, so an out-of-range value **disables** the check rather than erroring. That hole predates this PR — the old per-field `?? DEFAULT` had it too. Only the `maxConsoleOutputBytes` exposure is new.

I took your second option. `executionPolicy` has no production caller: the backend passes none, and `code-mode.ts` is not on the package's export surface. It exists so tests can reach a limit they cannot practically hit at its default, above all the 30s deadline. So it now takes a **complete** policy — `Readonly<Required<CodeModeExecutionPolicy>>`, with the body reduced to `input.executionPolicy ?? DEFAULT_CODE_MODE_EXECUTION_POLICY`. The merge is gone.

That deletes the `null` regression test you asked for last round, so let me argue it rather than slip it past you. `null`, `undefined`, `0`, negatives, floats, strings, out-of-range, and which fields may be tuned at all are all questions about *merging*. With no merge they stop being representable instead of being guarded. Mapping a narrower key list and adding a range clause keeps a validation apparatus policing a parameter nothing in production can reach.

In its place, the coverage that was actually missing: nothing pinned the console property, which is why it took a probe to find. `keeps sandbox console output out of the host process stdout` runs a probe cell in a child process — the sandbox writes from a worker thread Node pipes into the parent's stdout, so patching `process.stdout.write` in-process cannot see it. It goes red when the default is raised from 1.

`0ed2085f6`. Focused Code Mode suites 50 pass, biome clean. Agreed on rerunning `package` before merge; #3327 looks like the one that covers both of my signatures.

hqhq1025 (approved):
Approved at exact head `0ed2085f6b317daf926d378ed6a3fb6beecad23d`.

No actionable findings remain. The previous policy-boundary finding is fixed by removing the per-field merge rather than adding another validation layer. Production still omits `executionPolicy`, `code-mode.ts` is not on the Runtime package export surface, and the internal test seam now requires a complete SDK policy. That makes the frozen Maka policy the only policy on the product path while still allowing focused limit tests.

The new child-process regression test covers the missing product invariant directly: sandbox console output must not reach host stdout, which is also used by the CLI. I independently verified both sides of the probe: the default policy produced empty stdout, while a complete test policy with `maxConsoleOutputBytes: 100` emitted the marker. The focused Core/Runtime suites passed 79/79, the policy/admission subset passed 10/10 repeated runs, Core/Storage/Runtime builds passed, Biome passed, notice inventories were current, and the CI planner tests passed 17/17.

Problem definition and mechanism: the PR correctly removes SDK-superseded workspace scaffolding while retaining the two responsibilities Maka still owns: full-cell admission through host-operation drainage and a stricter product execution policy. The backend-scoped admission limiter, stateless Code Mode adapter, shared Core byte counter, and folded Runtime dependency are the appropriate ownership boundaries.

First principles and Occam's razor: yes. The follow-up deletes the unnecessary merge and its validation tests instead of maintaining a configurable surface with no production caller. I found no further production code or low-quality tests that should be deleted, and no deeper refactor is required.

Merge verdict: the code at this head is ready. GitHub CI, dependency audit, and the Windows package check are still pending, so do not merge until the required checks complete successfully. The previous Windows cleanup race remains an external verification risk; if it repeats, the harness fix in #3327 should land and this branch should be refreshed before rerunning.

Issue #3211, opened 2026-08-18
test(release): FakeBackend is a first-class BackendKind and desktop E2E material ships in release artifacts


Astro-Han:
Found by a repo-wide simplification audit (static analysis; zero-consumer claims re-verified with `git grep -aI`).

`FakeBackend` is a test backend, but it is wired as first-class production surface, and desktop E2E material ships in release artifacts:

- `BackendKind` defines `'fake'` as a formal session backend: `packages/core/src/session.ts:262`.
- The normal Runtime Host composition unconditionally registers it: `packages/runtime-host/src/server/execution-composition.ts:289`.
- Unknown/legacy providers fall back to `'fake'`: `packages/core/src/llm-connections.ts:375`.
- The production candidate entry statically imports the desktop E2E composition: `packages/runtime-host/src/execution-candidate-main.ts` (top-of-file imports).
- Release packaging copies whole `dist/` trees, so FakeBackend, E2E fixture schema, and `--desktop-e2e` branches ship in the CLI package and the Electron app: `scripts/release-cli-package.mjs:367-369`, `apps/desktop/electron-builder.config.mjs`.

Product layers then carry `fake` special-cases (search exclusion in `apps/desktop/src/main/search/thread-search.ts`, model selector, stale-session handling). Note the fake-rejection in `chat-readiness.ts` (`assertSessionCanSend`) currently has no production caller — only `errorMessage` is imported by `main-window.ts` — so it is residue rather than an active guard.

**Suggested direction:** keep E2E going through the real composition via the existing `primaryBackendFactory` seam, but move FakeBackend + fixture schema + candidate bootstrap to a test-only compile entry, drop `'fake'` from `BackendKind` and the unknown-provider fallback, and stop shipping fixture material in release artifacts.

PR #3226, opened 2026-08-18
fix(release): stop shipping FakeBackend and desktop E2E material in production artifacts

Astro-Han:
## Summary

`FakeBackend` is a test backend, but it was wired as a first-class production surface and left the build inside the release artifacts. The normal Runtime Host composition registered it unconditionally (`packages/runtime-host/src/server/execution-composition.ts:289`), and the production candidate entry statically imported the Desktop E2E composition, so both the CLI tarball and the packaged Electron app carried a test backend plus the `--desktop-e2e` bootstrap.

This PR is the packaging half of #3211: get the test backend and the Desktop E2E bootstrap out of the production build graph and out of both artifacts, without weakening the rule that E2E must run through the real Runtime Host composition.

**What changed**

- `FakeBackend` moves to `@maka/runtime/test-only/fake-backend`; `DesktopE2eBackend` and its composition move to `packages/runtime-host/src/test-only/desktop-e2e-execution.ts`, with a new `test-only/execution-candidate-e2e-main.ts` entry beside it.
- The production candidate entry loses its E2E import, and `--desktop-e2e` disappears from the candidate CLI, the launcher, `connect-or-spawn` and the Desktop candidate manager. The entry module is now the switch: `runtime-host-boot.ts` resolves `@maka/runtime-host/test-only/execution-candidate-e2e-main` when `isE2e`, and the production entry otherwise.
- The production composition no longer registers a `fake` backend. Every deterministic-backend consumer — Desktop E2E, the in-process composition tests, and the forked `execution-host` test fixture — now goes through the existing `primaryBackendFactory` seam, which is exactly the seam #3211 asked to preserve.
- Release packaging drops `test-only/` on both artifact paths, and `packages/runtime/README.md` records the convention.

**Design decisions this PR commits to (issue items 1 and 4)**

*Test-only compile entry.* A `test-only/` source directory plus a `./test-only/*` package export, extending the convention `@maka/runtime` already used for `./test-only/observation-text-reader`. This was chosen over build-time exclusion or export conditions because it makes one fact — "no production module imports this" — checkable statically, from one directory name, on both artifact paths.

*Release-exclusion verification, without building a release.* Three static checks, no packing required:

1. `isMakaDevelopmentArtifact` in `scripts/release-cli-file-policy.mjs` now rejects any `test-only/` segment, with unit tests in `scripts/release-cli-file-policy.test.mjs` (run by `npm run check:release`). `test-only` is deliberately *not* added to `DEVELOPMENT_DIRECTORIES`, so a third-party package shipping a directory by that name is unaffected — there is a test for that too.
2. The same predicate already gates `validatePackedFiles` in `scripts/release-cli-package.mjs`, so if the copy filter ever let such a file through, packing throws rather than shipping it. `copyRuntimeDist` drops the directory at copy time.
3. `apps/desktop/electron-builder.config.mjs` gains `'!**/test-only/**'` beside the existing `'!**/__tests__/**'`.

The strongest guarantee is not a file filter, though: a new test in `packages/runtime-host/src/__tests__/execution-candidate-main.test.ts` walks the built module graph from the production candidate entry across all bundled `@maka/*` packages and fails if it can reach any `test-only/` module. Even a stray byte in an artifact would now be unreachable code.

**Design decided here, implemented in the follow-up (issue items 2 and 3)**

`'fake'` stays in `BackendKind` in this PR. Removing it is a persistence and product-surface change, not a packaging one, and it is large enough to review separately. The decisions, so the follow-up does not re-litigate them:

*Persistence compatibility (item 2).* Verified: a stored `backend: 'fake'` is **not** tolerated by decode today. `packages/storage/src/session-store.ts:1146` (`isBackendKind`) is a hard validator — narrowing the union there turns every legacy fake session into `Invalid session header … malformed fields`, i.e. an unreadable session, not a degraded one. `packages/core/src/agent-run.ts:712` and `packages/core/src/scheduled-task.ts:492` have the same shape. So the follow-up must split the two jobs `'fake'` currently does: selecting a live backend factory, and tagging a retired legacy session. `BackendKind` (the live selector) narrows to `'ai-sdk'`; the persisted header keeps a wider `'ai-sdk' | 'fake'` type that decode still accepts and that resolves to no factory. No data migration — the product already treats these rows as retired (`session-send-projection.ts:118` → `fake_backend`, `stale-sessions.ts:27`, search exclusion in `thread-search.ts:171`), and rewriting them to `'ai-sdk'` would make dead sessions look sendable.

*Unknown-provider fallback (item 3).* `packages/core/src/llm-connections.ts:375` currently returns `'fake'` for an unknown or legacy `providerType`. That is a behavior change to make explicitly: `backendKindOf` will throw on an unrecognized provider rather than silently resolving a connection onto a test backend. Callers that need a non-throwing answer use the readiness projection, which already reports "not a real connection" without needing a backend kind. This is called out here so the follow-up PR carries it under its own heading.

**Behavior change**

A session or Automation persisted with `backend: 'fake'` no longer runs the test backend. The composition registers an explicit refusal at that kind instead, throwing the canonical `NO_REAL_CONNECTION:fake_backend` error, which `parseNoRealConnectionError` already turns into the copy both surfaces show for these rows: the task came from the retired local simulation, add a real model and start a new one.

An earlier revision of this PR left no factory at all there. That was wrong: the Desktop composer does gate these rows (`sendBlocked` reads a destructive `sessionHealthNotice`, raised by `projectSessionSendOutcome` for `fake_backend`), but that is a renderer gate, not an authority. `maka run` resumes an existing session through a readiness check that only inspects the connection catalog and never reads the session header's backend, and `scheduled-task.ts:492` still accepts `execution.backend: 'fake'` in a persisted template — both reach activation directly, and would have surfaced a bare `No backend factory registered for kind="fake"`.

Rewriting the durable header to `'ai-sdk'` on the read path was the alternative and is worse: it destroys the fact that drives the correct copy, and leaves a session that looks runnable while its `llmConnectionSlug` still points at nothing, so the failure resurfaces later and less specifically.

**Not in scope**

- `apps/desktop/src/main/e2e-fixture*` (the dev screenshot fixture and `@maka/core/e2e-fixture` schema) still ships in the asar. It is guarded — `resolveE2eFixture` throws when `app.isPackaged` — and removing it from the build graph means making `main.ts` load it dynamically, which is a desktop-startup refactor, not a packaging filter.
- `apps/desktop/src/main/chat-readiness.ts`'s `assertSessionCanSend` fake-rejection has no production caller (only `errorMessage` is imported, by `main-window.ts`). It is residue, not an active guard; it is not treated as one here and is left for the follow-up's cleanup.

Part of #3211

## Verification

Built `@maka/code-mode` → `@maka/core` → `@maka/storage` → `@maka/mcp` → `@maka/runtime` → `@maka/runtime-host` → `@maka/ui` → `@maka/desktop` from clean.

- `@maka/runtime-host` full workspace suite: **979 passed, 0 failed** (`node --test --test-concurrency=4 "dist/**/*.test.js"`).
- `@maka/runtime` affected files: **237 passed, 0 failed** (`fake-backend`, `session-manager`, `runtime-continuation-crash`, `stream-graph-coordinator`).
- `npm run check:release`: **30 passed, 0 failed**, including the new `test-only` file-policy tests.
- `npm run lint`, `npm run format`, and `@maka/desktop`'s four-config `typecheck` all clean.
- Desktop Playwright E2E, the paths that exercise the moved material: **19 passed** across `send-message`, `slash-command-menu`, `streaming-remount`, `composer-inline-completion`, `parent-session-deletion`. `slash-command-menu.spec.ts:34` ("compacts the active session") drives `DesktopE2eBackend.compactHistory`, so it proves the new E2E entry still reaches the real composition through `primaryBackendFactory`.

Module-graph evidence for the packaging claim, from the built `dist`:

```
production entry : {"size":622,"reached":[]}
e2e entry        : {"size":1,"reached":["test-only/execution-candidate-e2e-main.js"]}
```

622 modules reachable from `execution-candidate-main.js`, none of them test-only; the E2E entry is itself test-only, as expected. This is the assertion the new test makes.

The retired-backend refusal has a red-to-green regression test (`a legacy fake-backend session is refused with the product reason, not a registry error`): removing the registration makes it fail on `No backend factory registered`.

On "fails without it": the `release-cli-file-policy` tests do fail on `main` — the predicate did not reject `test-only/` paths. The module-graph test is a forward guard rather than a red-to-green test: on `main` the same modules exist under different names, so the assertion it encodes ("the production entry cannot reach test-only material") only becomes expressible once the directory convention exists.

Not run locally: a full CLI `npm pack` and a full `electron-builder` pack. Both are CI-side, and the point of the file-policy unit tests plus the module-graph test is to catch a regression without them.

## AI use

Tool(s) and scope: Claude Code (Opus) — investigated the issue against current `main`, drafted the design above, wrote the implementation and the new tests, and ran the verification listed. The human contributor of record reviews the final diff and owns accuracy and provenance.

## Checklist

- [ ] No


YayoiNanoka:
## PR Assessment

### What problem does this PR solve?

This PR removes `FakeBackend` and the Desktop E2E bootstrap from production module graphs and release artifacts, while preserving the existing Runtime Host composition as the E2E execution path.

It also replaces the old production `--desktop-e2e` switch with a dedicated test-only candidate entrypoint and gives persisted legacy `backend: 'fake'` sessions an explicit product-level refusal instead of an unregistered-backend error.

### How does this PR solve the problem?

- Moves `FakeBackend` and Desktop E2E composition code under `test-only/` exports.
- Removes the unconditional fake backend registration and the production `--desktop-e2e` CLI flow.
- Selects the E2E candidate entrypoint from `runtime-host-boot.ts`, while the production candidate remains free of E2E imports.
- Keeps deterministic tests and Desktop E2E on the existing `primaryBackendFactory` seam.
- Excludes `test-only/` files from CLI and Electron release packaging.
- Adds a production module-graph test that follows static imports, exports, side-effect imports, and literal dynamic imports.
- Registers an explicit `fake_backend` refusal so legacy sessions receive the existing user-facing product reason.

### Is the problem correctly defined?

Correct.

The PR addresses both sides of the release-isolation problem: preventing test-only code from entering production artifacts and ensuring that removing the production fake backend does not degrade legacy persisted sessions into an internal registry error.

### Principle-based assessment

- First principles: The production artifact must not depend on test-only modules, and durable `backend: 'fake'` values must retain their retired-session semantics rather than being silently rewritten into a runnable backend. The implementation preserves both invariants.
- Occam's razor: The PR removes the old production flag and import path, reuses the existing composition factory seam, and adds only the release filters and graph checks needed to enforce the boundary.
- Minimal sufficient solution: The dedicated test-only entrypoint is sufficient to keep E2E behavior while removing the production E2E branch. The explicit legacy refusal avoids reintroducing `FakeBackend` or corrupting durable session metadata.

## Review Findings

### P1 Blocker

None.

### P2 Should Fix

None.

### Suggestion

None.

## Verification

- Base: `781fa8d18d7f1d5e983de8283b3e0a61cd05751d`
- Head: `2f4bb27ad8a3eee1e8afad13c79c75ab2fdf8ada`
- Executed:
  - Built `@maka/code-mode`, `@maka/core`, `@maka/storage`, `@maka/mcp`, `@maka/runtime`, and `@maka/runtime-host`.
  - Ran the affected Runtime Host candidate, composition, CLI, and Desktop E2E unit tests.
  - Runtime Host composition tests: `13 passed, 0 failed`.
  - Candidate and Desktop E2E tests: `7 passed, 0 failed`.
  - Release policy suite: `30 passed, 0 failed`.
  - Checked the complete PR diff and verified no whitespace errors.
  - Checked GitHub Actions for the PR head: CI, Runtime Host tests, Desktop E2E, CLI package validation, Windows release checks, Windows recovery/baseline, and dependency audit all passed.
- Key results: The production module graph reached no `test-only/` module; the legacy fake-backend regression test returned the canonical `fake_backend` product reason; release packaging validation passed.
- Temporary tests: Not created; no temporary files were committed or pushed.
- Original workspace: Confirmed unchanged; the review ran in an isolated clone.
- Not verified: A local macOS Electron package was not rebuilt separately, but the Windows packaging workflow and release validation workflow passed for this exact head.

## Conclusion

Can merge.

The PR satisfies the stated packaging boundary, preserves the real Runtime Host composition for E2E, handles legacy fake sessions with an intentional product-level refusal, and has passing local and CI verification.
<details>
<summary>点击展开中文</summary>

## PR 判断

### 这个 PR 解决了什么问题？

这个 PR 将 `FakeBackend` 和 Desktop E2E 启动代码从生产模块图和发布产物中移除，同时保留现有 Runtime Host composition 作为 E2E 的执行路径。

它还用独立的 test-only candidate entrypoint 替代了生产环境中的 `--desktop-e2e` 开关，并为持久化的旧 `backend: 'fake'` session 提供明确的产品级拒绝，避免出现未注册 backend 的内部错误。

### 这个 PR 如何解决这个问题？

- 将 `FakeBackend` 和 Desktop E2E composition 移到 `test-only/` 导出路径。
- 删除无条件的 fake backend 注册和生产环境的 `--desktop-e2e` CLI 流程。
- 在 `runtime-host-boot.ts` 中选择 E2E candidate entrypoint，使生产 candidate 不再导入 E2E 代码。
- 测试和 Desktop E2E 继续使用现有的 `primaryBackendFactory` seam 注入确定性 backend。
- CLI 和 Electron 发布打包都排除 `test-only/` 文件。
- 新增生产模块图测试，覆盖静态 import、export、side-effect import 和字面量动态 import。
- 注册明确的 `fake_backend` 拒绝逻辑，使旧 session 得到现有的用户可理解错误原因。

### 这个问题定义得对吗？

正确。

这个 PR 同时处理了发布隔离问题的两面：防止 test-only 代码进入生产产物，并确保删除生产 fake backend 后，旧的持久化 session 不会退化成内部 registry 错误。

### 原则性判断

- 第一性原理：生产产物不能依赖 test-only 模块；持久化的 `backend: 'fake'` 必须保留其“已废弃 session”的语义，不能被静默改写成可运行 backend。实现满足这两个不变量。
- 奥卡姆剃刀：PR 删除了旧的生产开关和导入路径，复用了现有 composition factory seam，只增加了维持边界所需的发布过滤和模块图检查。
- 最小充分解：独立的 test-only entrypoint 足以保留 E2E 行为，同时移除生产 E2E 分支。显式拒绝旧 session 避免重新引入 `FakeBackend`，也避免破坏持久化 session 元数据。

## Review Findings

### P1 阻塞

无。

### P2 应该改

无。

### 建议

无。

## 验证

- Base：`781fa8d18d7f1d5e983de8283b3e0a61cd05751d`
- Head：`2f4bb27ad8a3eee1e8afad13c79c75ab2fdf8ada`
- 已执行：
  - 构建 `@maka/code-mode`、`@maka/core`、`@maka/storage`、`@maka/mcp`、`@maka/runtime` 和 `@maka/runtime-host`。
  - 运行 Runtime Host candidate、composition、CLI 和 Desktop E2E 相关测试。
  - Runtime Host composition 测试：`13 passed, 0 failed`。
  - Candidate 和 Desktop E2E 测试：`7 passed, 0 failed`。
  - 发布策略测试：`30 passed, 0 failed`。
  - 检查完整 PR diff，并确认没有 whitespace error。
  - 检查 PR head 对应的 GitHub Actions：CI、Runtime Host 测试、Desktop E2E、CLI package validation、Windows 发布检查、Windows recovery/baseline 和 dependency audit 全部通过。
- 关键结果：生产模块图没有到达任何 `test-only/` 模块；旧 fake-backend session 的回归测试返回规范的 `fake_backend` 产品原因；发布打包验证通过。
- 临时测试：未创建；没有临时文件被提交或推送。
- 原始工作区：已确认未修改；审查在隔离 clone 中进行。
- 未验证：没有单独在本地重新构建 macOS Electron 包，但对应 commit 的 Windows packaging 和 release validation workflow 已通过。

## 结论

可以合入。

这个 PR 满足发布边界要求，保留了真实 Runtime Host composition 作为 E2E 路径，对旧 fake session 提供了明确且有意设计的产品级拒绝，并且本地和 CI 验证均通过。

</details>

PR #3227, opened 2026-08-19
fix(storage): retire completed legacy migration metadata

liugddx:
## Summary

- retire the released `cutover_journal`, `runtime_import_sources`, and `session_metadata_import_sources` tables during the existing atomic operational-state migration
- validate the released column shapes (name/type/notNull/pk) and row-level migration evidence before removing obsolete metadata; column-shape validation does not inspect CHECK/FK constraints, and row validation is the authoritative content gate
- require every cutover row to be completed and internally valid
- preserve interrupted, malformed, or unfamiliar migration state unchanged and fail closed

## Root cause

Released workspaces can still contain migration bookkeeping created before SQLite became the sole operational authority. The old readers and writers were removed, but their tables were not retired. Current schema convergence therefore upgrades the real application tables, then rejects the otherwise valid database because strict target-schema validation sees `cutover_journal` (followed by the two import-source tables) as unexpected objects.

This presents as `OPERATIONAL_STATE_MIGRATION_BLOCKED` on startup even though SQLite integrity is healthy.

## Verification

- `npm run clean --workspace @maka/storage`
- `npm run build --workspace @maka/storage`
- operational-state suite under the packaged Electron Node 24 runtime: **29 pass / 0 fail**
- Biome 2.5.6 check for both changed source files
- `git diff --check`
- replayed the migration against a copy of the affected released workspace database:
  - migration completed
  - `PRAGMA quick_check` returned `ok`
  - schema versions converged to runtime 12, session metadata 27, workflow 9, operational 2
  - all three retired metadata tables were absent afterward

The full storage suite was also attempted locally, but unrelated tests that load `fs-native-extensions` cannot start because this checkout lacks its Darwin prebuilt addon. The focused suite and real-database-copy replay do not load that optional native path; CI remains authoritative for the complete package suite.

The real workspace database was never modified; all migration replay used temporary copies.

## AI use

Select exactly one:

Tool(s) and scope: Codex and Claude Code contributed to the implementation, regression tests, historical writer-contract analysis, and review-response changes. The human contributor reviewed the final diff, owns the submission, and owns the merge decision.

## Checklist

- [x] Affected package build, formatting, and focused tests pass
- [x] `git diff --check` passes

Generated-by: Codex

### DIFF
--- a/packages/storage/src/operational-state-store.ts
+++ b/packages/storage/src/operational-state-store.ts
@@ -58,6 +58,26 @@ const OPERATIONAL_SCHEMA_VERSIONS: ReadonlyMap<string, number> = new Map([
 const REMOVED_OPERATIONAL_SCHEMA_VERSIONS: ReadonlyMap<string, number> = new Map([
   ['automation', 2],
 ]);
+const LEGACY_CUTOVER_JOURNAL_COLUMNS = [
+  { name: 'store_name', type: 'TEXT', notNull: 0, primaryKey: 1 },
+  { name: 'source_path', type: 'TEXT', notNull: 1, primaryKey: 0 },
+  { name: 'source_fingerprint', type: 'TEXT', notNull: 1, primaryKey: 0 },
+  { name: 'state', type: 'TEXT', notNull: 1, primaryKey: 0 },
+  { name: 'started_at', type: 'INTEGER', notNull: 1, primaryKey: 0 },
+  { name: 'completed_at', type: 'INTEGER', notNull: 0, primaryKey: 0 },
+  { name: 'validation_json', type: 'TEXT', notNull: 0, primaryKey: 0 },
+] as const;
+const LEGACY_RUNTIME_IMPORT_SOURCE_COLUMNS = [
+  { name: 'source_path', type: 'TEXT', notNull: 0, primaryKey: 1 },
+  { name: 'fingerprint', type: 'TEXT', notNull: 1, primaryKey: 0 },
+  { name: 'imported_at', type: 'INTEGER', notNull: 1, primaryKey: 0 },
+] as const;
+const LEGACY_SESSION_IMPORT_SOURCE_COLUMNS = [
+  { name: 'source_path', type: 'TEXT', notNull: 0, primaryKey: 1 },
+  { name: 'fingerprint', type: 'TEXT', notNull: 1, primaryKey: 0 },
+  { name: 'session_id', type: 'TEXT', notNull: 1, primaryKey: 0 },
+  { name: 'imported_at', type: 'INTEGER', notNull: 1, primaryKey: 0 },
+] as const;
 
 const require = createRequire(import.meta.url);
 const owners = new Map<string, OperationalStateDatabaseOwner>();
@@ -365,6 +385,7 @@ export function migrateOperationalStateDatabaseInternal(db: DatabaseSync, now: (
       DROP TABLE IF EXISTS automation_authority_state;
       DELETE FROM operational_schema_migrations WHERE scope = 'automation';
     `);
+    retireCompletedLegacyMigrationMetadata(db);
     assertCurrentOperationalTargetSchema(db);
     for (const [scope, version] of OPERATIONAL_SCHEMA_VERSIONS) {
       registerSchema(db, scope, version, appliedAt);
@@ -374,6 +395,158 @@ export function migrateOperationalStateDatabaseInternal(db: DatabaseSync, now: (
     rollback(db);
     throw error;
   }
+}
+
+/**
+ * Retire import and cutover evidence written before SQLite became the sole
+ * operational authority. Only exact released tables with internally valid,
+ * completed rows may be removed; interrupted or unfamiliar state stays
+ * fail-closed and the surrounding migration transaction rolls back unchanged.
+ */
+function retireCompletedLegacyMigrationMetadata(db: DatabaseSync): void {
+  if (hasTable(db, 'cutover_journal')) {
+    assertReleasedLegacyTableShape(db, 'cutover_journal', LEGACY_CUTOVER_JOURNAL_COLUMNS);
+    const rows = db
+      .prepare(`
+        SELECT
+          store_name,
+          source_path,
+          source_fingerprint,
+          state,
+          started_at,
+          completed_at,
+          validation_json
+        FROM cutover_journal
+        ORDER BY store_name
+      `)
+      .all() as Array<Record<string, unknown>>;
+    for (const row of rows) assertCompletedLegacyCutoverJournalRow(row);
+  }
+  if (hasTable(db, 'runtime_import_sources')) {
+    assertReleasedLegacyTableShape(
+      db,
+      'runtime_import_sources',
+      LEGACY_RUNTIME_IMPORT_SOURCE_COLUMNS,
+    );
+    const rows = db
+      .prepare(`
+        SELECT source_path, fingerprint, imported_at
+        FROM runtime_import_sources
+        ORDER BY source_path
+      `)
+      .all() as Array<Record<string, unknown>>;
+    for (const row of rows) assertLegacyImportSourceRow(row);
+  }
+  if (hasTable(db, 'session_metadata_import_sources')) {
+    assertReleasedLegacyTableShape(
+      db,
+      'session_metadata_import_sources',
+      LEGACY_SESSION_IMPORT_SOURCE_COLUMNS,
+    );
+    const rows = db
+      .prepare(`
+        SELECT source_path, fingerprint, session_id, imported_at
+        FROM session_metadata_import_sources
+        ORDER BY source_path
+      `)
+      .all() as Array<Record<string, unknown>>;
+    const sessionExists = db.prepare('SELECT 1 FROM session_metadata WHERE session_id = ?');
+    for (const row of rows) {
+      assertLegacyImportSourceRow(row);
+      if (
+        typeof row.session_id !== 'string' ||
+        row.session_id.length === 0 ||
+        !sessionExists.get(row.session_id)
+      ) {
+        throw new Error('Legacy session import source is incomplete or invalid');
+      }
+    }
+  }
+
+  db.exec(`
+    DROP TABLE IF EXISTS session_metadata_import_sources;
+    DROP TABLE IF EXISTS runtime_import_sources;
+    DROP TABLE IF EXISTS cutover_journal;
+  `);
+}
+
+function assertReleasedLegacyTableShape(
+  db: DatabaseSync,
+  table: string,
+  expectedColumns: ReadonlyArray<{
+    readonly name: string;
+    readonly type: string;
+    readonly notNull: number;
+    readonly primaryKey: number;
+  }>,
+): void {
+  const columns = db
+    .prepare(`
+      SELECT name, type, "notnull" AS not_null, pk
+      FROM pragma_table_info(?)
+      ORDER BY cid
+    `)
+    .all(table) as Array<{ name?: unknown; type?: unknown; not_null?: unknown; pk?: unknown }>;
+  const releasedShape = columns.every((column, index) => {
+    const expected = expectedColumns[index];
+    return (
+      expected !== undefined &&
+      column.name === expected.name &&
+      column.type === expected.type &&
+      column.not_null === expected.notNull &&
+      column.pk === expected.primaryKey
+    );
+  });
+  if (columns.length !== expectedColumns.length || !releasedShape) {
+    throw new Error(`Legacy operational metadata table ${table} has an unfamiliar schema`);
+  }
+}
+
+function assertCompletedLegacyCutoverJournalRow(row: Record<string, unknown>): void {
+  if (
+    typeof row.store_name !== 'string' ||
+    row.store_name.length === 0 ||
+    typeof row.source_path !== 'string' ||
+    row.source_path.length === 0 ||
+    typeof row.source_fingerprint !== 'string' ||
+    row.source_fingerprint.length === 0 ||
+    row.state !== 'completed' ||
+    !isNonnegativeInteger(row.started_at) ||
+    !isNonnegativeInteger(row.completed_at) ||
+    typeof row.validation_json !== 'string'
+  ) {
+    throw new Error('Legacy operational cutover journal is incomplete or invalid');
+  }
+  let validation: unknown;
+  try {
+    validation = JSON.parse(row.validation_json);
+  } catch {
+    throw new Error('Legacy operational cutover journal has invalid validation evidence');
+  }
+  if (
+    typeof validation !== 'object' ||
+    validation === null ||
+    Array.isArray(validation) ||
+    !Object.values(validation).every(isNonnegativeInteger)
+  ) {
+    throw new Error('Legacy operational cutover journal has invalid validation evidence');
+  }
+}
+
+function assertLegacyImportSourceRow(row: Record<string, unknown>): void {
+  if (
+    typeof row.source_path !== 'string' ||
+    row.source_path.length === 0 ||
+    typeof row.fingerprint !== 'string' ||
+    row.fingerprint.length === 0 ||
+    !isNonnegativeInteger(row.imported_at)
+  ) {
+    throw new Error('Legacy operational import source is incomplete or invalid');
+  }
+}
+
+function isNonnegativeInteger(value: unknown): value is number {
+  return typeof value === 'number' && Number.isSafeInteger(value) && value >= 0;
 }
 
 function registerSchema(db: DatabaseSync, scope: string, version: number, appliedAt: number): void {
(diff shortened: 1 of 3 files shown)


packages/storage/src/operational-state-store.ts
```diff
@@ -376,6 +397,162 @@ export function migrateOperationalStateDatabaseInternal(db: DatabaseSync, now: (
...
+    readonly name: string;
+    readonly type: string;
+    readonly notNull: number;
+    readonly primaryKey: number;
+  }>,
+): void {
+  const columns = db
+    .prepare(`
+      SELECT name, type, "notnull" AS not_null, pk
+      FROM pragma_table_info(?)
```
Astro-Han: [P2] Validate the complete released schema before destructive cleanup

This column projection cannot distinguish the released table from one with an altered `CHECK`/foreign-key contract or an added trigger/index; I reproduced an altered `CHECK` plus trigger returning the exact same column metadata. The subsequent DROP therefore removes schema this build has not actually recognized, which conflicts with the fail-closed invariant. Could we reuse or consolidate the full `sqlite_schema` signature seam in `operational-target-schema.ts`, and add one altered-constraint or extra-trigger regression case? The historical tables have a stable released definition, so this can remain a single explicit legacy contract rather than a parallel migration system.

<details>
<summary>中文</summary>

这里只读取列信息，无法区分已发布表与修改过 CHECK/外键或附加 trigger/index 的陌生表；实测修改约束并增加 trigger 后，返回的列信息完全相同。随后 DROP 会删除当前版本并未真正识别的结构，违背 fail-closed。建议复用或合并 `operational-target-schema.ts` 中完整的 `sqlite_schema` signature，并增加 altered constraint 或 extra trigger 的回归测试。

</details>

packages/storage/src/operational-state-store.ts
```diff
@@ -376,6 +397,162 @@ export function migrateOperationalStateDatabaseInternal(db: DatabaseSync, now: (
...
+    );
+  });
+  if (columns.length !== expectedColumns.length || !releasedShape) {
+    throw new Error(`Legacy operational metadata table ${table} has an unfamiliar schema`);
+  }
+}
+
+function assertCompletedLegacyCutoverJournalRow(row: Record<string, unknown>): void {
+  if (
+    typeof row.store_name !== 'string' ||
```
Astro-Han: [P2] Reject journal contracts this build does not recognize

The current predicate accepts any non-empty `store_name` and any non-empty object of non-negative integers. A completed row such as `future_store` with `{"anything": 0}` therefore passes and the journal is deleted, even though no released writer in this build defines that contract. That turns unfamiliar migration evidence into accepted completion evidence. Please recognize only the store/evidence contracts actually emitted by the released writers and abort the transaction for any unknown row; this keeps the promised fail-closed behavior without adding another recovery path.

<details>
<summary>中文</summary>

当前判断接受任意非空 `store_name` 和任意非空的非负整数对象。例如 `future_store` 加 `{"anything": 0}` 也会通过并导致 journal 被删除，但当前版本没有任何已发布 writer 定义这种 contract。建议只接受历史发布 writer 实际产生过的 store/evidence contract；遇到未知记录时回滚并保留原库。

</details>

Astro-Han (commented):
AI-assisted review by Codex. I inspected the current head `6f27a`, the historical released writers and table definitions, the atomic migration path, the focused tests, current CI, and the existing Qodo/CodeRabbit feedback.

The problem definition and the chosen seam are sound: obsolete migration metadata should be retired inside the existing atomic schema-convergence transaction, and any state this build cannot recognize should leave the database unchanged. The transaction and rollback ordering preserve that boundary correctly.

The remaining gap is the authority for “released and completed.” Today that decision is split between column-only DDL recognition and permissive row checks. The two inline P2 findings are facets of that same invariant. The smallest coherent correction is to consolidate recognition at the existing schema-signature seam and accept only journal contracts emitted by released writers, while keeping the DROP statements in this transaction.

I would not add a `completed_at >= started_at` requirement: the released writer used separate `Date.now()` calls and guaranteed only non-negative timestamps, so wall-clock rollback can produce a legitimate reverse ordering.

<details>
<summary>中文审查意见</summary>

问题定义和处理位置是正确的：应在现有原子 schema convergence 事务内清理已完成的旧迁移元数据，无法识别的状态必须保持数据库原样。

当前剩余问题是“已发布且已完成”的权威定义被拆成了两套不完整判断：只看列形状的 DDL 判断，以及过于宽松的行内容判断。两条行内 P2 属于同一个不变量缺口。最小而完整的方案是复用现有 schema signature 能力，并且只接受历史发布 writer 实际产生过的 journal contract，DROP 仍保留在当前事务中。

不建议增加 `completed_at >= started_at`：历史 writer 分两次调用 `Date.now()`，只保证非负数，并不保证墙上时钟单调。

</details>

packages/storage/src/operational-state-store.ts
```diff
@@ -376,6 +399,123 @@ export function migrateOperationalStateDatabaseInternal(db: DatabaseSync, now: (
...
+    validation = JSON.parse(row.validation_json);
+  } catch {
+    throw new Error('Legacy operational cutover journal has invalid validation evidence');
+  }
+  if (
+    typeof validation !== 'object' ||
+    validation === null ||
+    Array.isArray(validation) ||
+    Object.keys(validation).length === 0 ||
+    !Object.values(validation).every(isNonnegativeInteger)
```
Astro-Han: [P2] Validate the evidence keys for each released store

The new store-name allowlist closes the future-store case, but the evidence contract is still open: a completed session_metadata row with validation_json = {"anything":0} passes this predicate and the transaction drops cutover_journal, although no released writer emitted that key. The same applies to missing or extra keys for every known store, so this still retires migration evidence the build has not recognized. Please make the historical authority a map from each released store_name to its exact validation-key set, require an exact key-set match here, and add a regression with a known store plus an unknown key that asserts the journal survives rollback.

新的 store_name 白名单关闭了未知 store 路径，但 evidence contract 仍然开放：已完成的 session_metadata 行若使用 validation_json = {"anything":0}，仍会通过这里并导致 cutover_journal 被删除，而历史发布 writer 从未产生该 key。已知 store 的缺失或额外 key 同样会被接受，因此仍可能退休当前版本不认识的迁移证据。建议把历史权威收敛为 store_name 到精确 validation key 集合的映射，在这里要求 key 集完全匹配，并补一个“已知 store + 未知 key”的回归测试，验证回滚后 journal 保留。

</details>

Astro-Han (commented):
The latest head correctly closes the full-schema half of the prior review: it reuses the existing sqlite_schema signature seam, matches the released DDL variants, rejects altered constraints and attached objects, and keeps retirement inside the existing atomic migration. The released store-name set is also complete, and the current required checks are green.

One part of the previous invariant remains open: store identity is recognized, but validation evidence is only type-checked, not matched to the released writer for that store. Three independent reviewer passes reproduced the same known-store/unknown-key path; the external DeepSeek pass independently verified the historical writer contracts and the focused 35-test result. The inline P2 is the smallest remaining correction. A completed_at >= started_at constraint should not be added because the released writer used independent wall-clock reads.

I am leaving a review comment rather than approving this head. AI-assisted review by Codex with three independent reviewer agents and OpenCode Go DeepSeek V4 Flash (high); I verified the current head, historical DDL and writers, transaction boundary, focused test evidence, live CI, and deduplicated bot feedback.

最新 head 已正确关闭完整 schema 识别问题：复用现有 sqlite_schema signature，覆盖发布过的 DDL，拒绝修改约束和额外对象，并保持在原子迁移事务中。store_name 白名单也完整，CI 全绿。

但上一轮不变量仍有一半未闭合：当前只识别 store 名，validation_json 仍只做类型检查，没有匹配该 store 的历史 writer evidence contract。已知 store 加未知 key 会被接受并删除 journal。三轮独立 reviewer 均复现了该路径，DeepSeek 也核对了历史 writer contract。请按行内 P2 补齐后再 Approve。不应增加 completed_at >= started_at，因为历史 writer 使用独立 wall-clock 读取。

本次使用 Codex、三个独立 reviewer 和 OpenCode Go DeepSeek V4 Flash high；核对了当前 head、历史 DDL/writer、事务边界、focused tests、实时 CI，并去重了 bot feedback。

</details>

liugddx:
## Review follow-up

Pushed `62ca27fae` to address the outstanding review findings, and re-ran the generic adversarial pre-review checks against the resulting change.

### Addressed findings

**Astro-Han [P2]: validation evidence was not bound to the released store contract.**

The previous `RELEASED_CUTOVER_STORE_NAMES` allowlist only checked the store name. A completed row for a known store could therefore carry arbitrary, missing, or renamed validation keys and still be deleted. The implementation now maps every released `store_name` to the exact key set emitted by its final released `importAndValidate` writer (`1caea265c^`). Missing, extra, or renamed keys fail closed and preserve the journal.

The `session_metadata` fixture was updated to include the complete released table-count contract. The contract is explicitly pinned to the final writer generation; an older or unrecognized evidence shape remains preserved and startup-blocked rather than being destructively removed.

**CodeRabbit: completion timestamp ordering.**

Completed rows now also require `completed_at >= started_at`. A row with valid integer timestamps but reversed ordering is treated as invalid evidence and remains unchanged.

### Verification

- Biome check: clean
- `npm run build --workspace @maka/storage`: passed
- `git diff --check`: clean
- Focused operational-state suite: **37 pass / 0 fail / 1 skipped**
- Delete thought experiment: disabling either new guard makes its regression test fail; all new guards are load-bearing.

The full storage suite remains subject to CI because unrelated tests require the unavailable Darwin `fs-native-extensions` prebuilt addon in this environment.

### Review verdict

The previously identified findings are addressed. The change remains fail-closed for unfamiliar, incomplete, malformed, or internally inconsistent migration evidence. No new P0/P1 issue was found in the follow-up pass.

<details>
<summary>中文摘要</summary>

已推送 `62ca27fae` 修复审核问题：

- 不再只校验 `store_name`，现在按最终发布版本 writer 绑定每个 store 的完整 validation key 集合；缺少、增加或改名的 key 都会 fail-closed 并保留 journal。
- 增加 `completed_at >= started_at` 校验，倒序时间戳不会触发删除。
- Biome、storage build、diff check 均通过；focused suite 为 **37 pass / 0 fail / 1 skipped**。
- 删除思想实验验证通过：移除任一新 guard，对应回归测试都会失败。

复审结论：之前发现的问题已处理，未发现新的 P0/P1 问题。
</details>

packages/storage/src/operational-state-store.ts
```diff
@@ -376,6 +436,141 @@ export function migrateOperationalStateDatabaseInternal(db: DatabaseSync, now: (
...
+    typeof row.source_path !== 'string' ||
+    row.source_path.length === 0 ||
+    typeof row.source_fingerprint !== 'string' ||
+    row.source_fingerprint.length === 0 ||
+    row.state !== 'completed' ||
+    !isNonnegativeInteger(row.started_at) ||
+    !isNonnegativeInteger(row.completed_at) ||
+    // A completed row whose finish precedes its start is internally
+    // inconsistent evidence, not a row this build ever wrote.
+    row.completed_at < row.started_at ||
```
Astro-Han: **[P2] Accept completed rows across wall-clock rollback**

The released writers call `Date.now()` (or the injected `now`) independently when recording `started_at` and `completed_at`; their schema only guarantees non-negative timestamps. An NTP/manual clock rollback between those writes can therefore produce a valid completed journal with `completed_at < started_at`. Rejecting it here preserves the legacy tables but also throws `OperationalStateMigrationBlockedError` on every startup, so the workspace can never converge.

Remove the ordering predicate (keep both non-negative integer checks) and replace the current reversed-time rejection test with a case proving that such released evidence is retired.

Astro-Han (commented):
The exact per-store validation-key map now matches the released writers, and the drop remains inside the existing atomic migration transaction. One compatibility blocker remains: the new timestamp ordering rule rejects state that the released writer could legitimately produce.

Several CI jobs are still in progress, so this head is not merge-ready independently of the finding.

AI-assisted review disclosure: Codex reviewed exact head `62ca27f`, the released writer provenance, current CI, and thread state, with two independent reviewer passes.

<details>
<summary>中文说明</summary>

当前按 store 精确匹配 validation keys 的方向正确，且删除旧表仍处于原有原子迁移事务内。但新增的时间先后约束超出了已发布 writer 的契约：系统时钟回拨可产生合法的 `completed_at < started_at`，当前实现会让 workspace 每次启动都被迁移阻塞。另有数个 CI job 仍在运行。

</details>

liugddx:
## Scope follow-up

I reviewed the latest feedback and narrowed the patch back to the PR's actual contract.

The exact per-store `validation_json` key-set check remains: it directly closes the outstanding P2 because a known store with unknown, missing, or renamed evidence must not be destructively retired.

I removed the added `completed_at >= started_at` requirement in `92f94d988`. The released writer used independent wall-clock reads and only guaranteed non-negative timestamps; a clock rollback can legitimately produce `completed_at < started_at`. Rejecting that state would make an otherwise valid released workspace remain startup-blocked, so this constraint was outside the PR's compatibility contract.

The corresponding regression test was removed as well. The focused suite is now **36 pass / 0 fail / 1 skipped**, with Biome, storage build, and `git diff --check` clean.

The remaining CodeRabbit suggestion to prefer plain `CREATE TABLE` is a style nit with no correctness impact and is intentionally left out of this PR.

This should be the final scope-limited follow-up: the original evidence-contract blocker is addressed, while the unrelated clock-ordering behavior is not imposed.

Astro-Han (commented):
Reviewed exact head `92f94d98858c791a257ba2b2f6bf6414b108f0f5`.

The storage change is technically coherent: retirement requires the full released legacy signature, preserves fail-closed and rollback paths, and keeps the current schema/store authority intact. Focused Storage tests and current CI pass, there are no unresolved threads, and this does not affect UI/UX.

One provenance gate remains before approval. The PR body does not provide a complete AI-use declaration naming the tools and their scope. Of five substantive commits, only `4af7a288` has a `Generated-by:` trailer; `6f27a197`, `381d4318`, and `62ca27fa` identify Claude as a co-author but have no matching generation trailer, and `92f94d988` also has none.

Please add the PR-template AI-use declaration and tool/scope, then add the appropriate `Generated-by: <tool>` trailer to every materially AI-authored commit (or explicitly clarify any human-only commit) and preserve the trailers in the final squash. No code P0-P2 findings remain after that provenance correction.

AI-assisted review disclosure: Codex reviewed the exact-head storage lifecycle, schema signatures, rollback paths, focused tests, live CI, threads, and provenance metadata. No external model was used. Astro-Han authorized this review campaign.

<details>
<summary>中文说明</summary>

代码层面已经闭环：legacy signature、fail-closed、rollback 和当前 schema/store 权威都保持正确；Storage 测试及 CI 全绿，无未解决线程，也不涉及 UI/UX。

批准前还缺 AI provenance：PR body 需要按模板明确工具和范围；5 个实质 commit 只有 1 个有 `Generated-by:`，其中 3 个还标注了 Claude co-author，但没有对应 generation trailer。请为实质 AI-authored commits 补 trailer，或明确哪些 commit 完全由人工完成。补齐后没有剩余代码 P0-P2。

</details>

liugddx:
## Provenance gate addressed

The latest review identified a repository-compliance blocker, not a new code defect: the PR body lacked the required AI-use declaration and the materially AI-authored commits did not all carry the required trailer.

This is now addressed without expanding the code scope:

- The PR description now selects **Generative tooling made a substantive contribution** and names Codex and Claude Code, with their scope.
- All five substantive commits in this PR now carry `Generated-by: Codex` trailers, preserved in their commit messages.
- The code remains unchanged from the scope-limited head: exact per-store validation-key matching stays; the incompatible timestamp-ordering rule remains removed.
- No new code P0-P2 finding remains in the latest review. The CREATE TABLE formatting suggestion is a non-blocking nit and is intentionally outside this PR's scope.

The rewritten commit chain was force-pushed with `--force-with-lease`; the PR body and final commit history are now aligned with the repository's AI attribution requirements.

Astro-Han (approved):
Approved exact head `6edf57b83d4693719737da3b256a63f17f2b7094`.

The prior provenance blocker is fully resolved: the PR body now discloses the Codex and Claude Code scopes, all five material commits carry `Generated-by: Codex`, and the retained Claude co-author trailers match the disclosure. The storage-only diff remains sound, all review threads are resolved, and I found no P0–P3 issue. There is no UI/UX change, so no screenshot is required.

This approval is conditional on the remaining queued required checks finishing green before merge.

AI-assisted review disclosure: OpenAI Codex performed the exact-head code, provenance, review-thread, and CI analysis; I verified the commit trailers, disclosure scope, and live GitHub state before approving.

此前的 AI provenance 缺口已完整修复：body 说明清楚，5 个实质性提交都有 `Generated-by: Codex`，Claude co-author 信息也一致。代码没有 P0–P3，线程已全部解决，无 UI 变化。剩余 required checks 全绿后才算 merge-ready。

</details>

PR #3251, opened 2026-08-19
fix(desktop): reconcile goal.arm after connection loss

### LINKED ISSUE #3234: Reconcile a dispatched control operation instead of reporting it as failed
## Problem

Every Runtime Host control operation reports a connection loss after dispatch as an unknown outcome that must not be retried:

```ts
// packages/runtime-host/src/client/connection.ts
this.retryable = mode === 'query';
```

That is the correct contract — the Host may already have committed and durably flushed the mutation before the response was lost. What is missing is anything a caller can do about it. There is no operation identity on the wire and no reconciliation helper, so every control caller passes the interruption straight through to the user as a failure, and a retry can only produce a conflict.

Desktop has one facility here, `isReconnectableReadFailure` in `apps/desktop/src/main/ipc-reconnect-policy.ts`, and it applies to queries only.

## Where it shows

Raised in review of #3199 for `goal.arm`, which is the sharpest case because a committed Goal spends tokens: the dialog says the Goal could not be set, while the Session may in fact be armed, and the user's next ordinary Turn then runs under it. The Goal indicator (`useSessionGoal`) does show the real state with a one-click clear, so the outcome is visible and reversible, but the message the user was given is wrong.

The same shape applies to the other control operations — `goal.control`, session continuity, hosted execution, message and turn mutations.

## Suggested direction

Two levels, either of which is worth having on its own:

1. **Reconcile after a dispatched interruption.** `goal.query` is a retryable query, so a caller can reconnect, read the canonical state, and compare it against the known pre-state and the request it sent. That needs no protocol change and would let the dialog say "connection lost; the current Goal is X" instead of "the goal could not be set".
2. **Durable operation identity.** A per-operation id the Host records with the commit would give the stronger contract — an unambiguous answer to "did my mutation land", rather than an inference from state. This is a public contract change and needs its own architecture decision.

Reviewer's framing on #3199: <https://github.com/maka-agent/maka-agent/pull/3199#discussion_r3809515712>

## Not in scope for #3199

Building either of these for `goal.arm` alone would add a public contract inside a feature PR rather than extend an existing seam, so #3199 keeps the shared contract and this is tracked separately.

Sun-GLiang:
## Summary

- Add a reusable Desktop IPC reconciled-control seam that dispatches a mutation once, then waits for a replacement candidate and runs only the reconciliation query.
- Reconcile dispatched `goal.arm` connection loss against canonical `goal.query` state, including prior Goal identity, normalized condition, effective iteration limit, and token budget.
- Keep the Goal dialog open and locked for reconciled or unavailable outcomes, with Chinese and English copy and stale async-result protection across Session changes.
- Do not change the Runtime Host compatibility epoch, wire codec, or persistence schema.

Fixes #3234

## Verification

- `npm test -w @maka/runtime-host` — 984 passed
- `npm test -w @maka/desktop` — 944 passed
- `npm run typecheck -w @maka/desktop` — passed
- `npm run lint` — passed
- `npm run e2e -w @maka/desktop -- goal-dialog-budget.spec.ts` — 1 passed

## Dependency

- [ ] #3199 is merged

This Draft PR is based on the current head of #3199. Its diff against `main` therefore includes #3199 until that PR lands. After #3199 merges, this branch will be rebased onto the latest `upstream/main` and the PR will be marked ready for review.

## AI use

Select exactly one:

Tool(s) and scope: Codex implemented the Desktop IPC reconciliation seam, `goal.arm` integration, renderer behavior, tests, and verification under user direction.

## Checklist

- [ ] No


Sun-GLiang:
Superseded by #3363, rebuilt from the merged #3199 baseline as a fresh single-commit branch.

PR #2628, opened 2026-08-10
fix(desktop): plan titlebar chrome from the front surface

jackwener:
## Summary
- Titlebar **paint** stays frame-level (drag / OS controls); **interactive chrome** follows the front surface only.
- Add pure `planTitlebarChrome` (+ unit tests) so Settings no longer special-cases one button while workbar tools still leak.
- Refines #2617: same product outcome, one ownership rule instead of ad-hoc `!settingsOpen` + scattered module sets.

## Test plan
- [x] `app-shell-titlebar-chrome` unit tests
- [ ] Existing e2e: `settings owns the window chrome while a session remains active`
- [ ] Manual: open session → open Settings → no search / identity / workbar buttons in titlebar; window still drags

PR #3260, opened 2026-08-19
fix(desktop): hide workbar chrome behind settings

chihumyum:
## Summary

Hide the complete session Workbar while Settings owns the window, including its tabs, add button, and right-panel toggle. Closing Settings restores the previously expanded Workbar and its active tabs.

`ChatWorkbar` already forwards the shell's obscured state as `hidden`, but `SessionWorkbar` only applied it to panel content. Its titlebar-level toolbar remained visible whenever the right Workbar was expanded. Folding `hidden` into the placement visibility condition removes the whole Workbar frame without unmounting its state.

This fixes the residual Workbar-owned toolbar leak after #2617. It is related to #2628, but that PR plans AppShell titlebar actions and does not hide the tab strip rendered inside `SessionWorkbar`.

## Verification

- `npm --workspace @maka/desktop run build:with-deps`
- `npx playwright test e2e/settings.spec.ts --config e2e/playwright.config.ts` — 2 passed
- `npm run lint` — 2,446 files checked
- `npm run format:check` — 1,539 files checked
- `git diff --check`

The regression test opens a real Workbar tab, opens Settings, asserts the entire right Workbar is hidden, then closes Settings and asserts the prior tab is restored.

## Visual evidence

| Before — Workbar chrome leaks above Settings | After — Settings owns the full window |
| --- | --- |
| Before: the Workbar tab, add button, and right-panel toggle remain visible above Settings | After: Settings is shown without any Workbar chrome |

## AI use

Select exactly one:

Tool(s) and scope: Codex (OpenAI) — root-cause analysis, implementation, regression test, and validation. The contributor reviewed the result and owns the final submission. The affected commit carries a `Generated-by: Codex` trailer.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/src/renderer/session-workbar.tsx
+++ b/apps/desktop/src/renderer/session-workbar.tsx
@@ -703,7 +703,8 @@ export function SessionWorkbar(props: {
         const activeTab = panel.tabs.find((tab) => tab.id === panel.activeTabId);
         const showingLauncher = panel.launcherOpen || !activeTab;
         const visible =
-          placement === 'right' ? !props.rightCollapsed : props.bottomOpen;
+          !props.hidden &&
+          (placement === 'right' ? !props.rightCollapsed : props.bottomOpen);
         return (
           <Card
             key={placement}
--- a/apps/desktop/e2e/settings.spec.ts
+++ b/apps/desktop/e2e/settings.spec.ts
@@ -19,3 +19,31 @@ test('opening settings commits an active titlebar rename', async ({ window: page
 
   await expect(identity).toContainText('renamed before settings');
 });
+
+test('settings hides expanded workbar chrome and restores it on close', async ({
+  window: page,
+}) => {
+  const composer = page.locator(COMPOSER_INPUT);
+  await composer.fill('create a session with an expanded workbar');
+  await composer.press('Enter');
+
+  await page.getByRole('button', { name: '展开任务工作栏' }).click();
+  const workbar = page.locator('.maka-session-workbar[data-placement="right"]');
+  const workbarToolbar = workbar.getByRole('toolbar', { name: '任务工作栏标签' });
+  await expect(workbarToolbar).toBeVisible();
+  await expect(workbarToolbar.getByRole('button', { name: '打开工作栏标签' })).toBeVisible();
+  await expect(workbarToolbar.getByRole('button', { name: '收起任务工作栏' })).toBeVisible();
+  await page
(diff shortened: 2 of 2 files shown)


apps/desktop/src/renderer/session-workbar.tsx
```diff
@@ -703,7 +703,8 @@ export function SessionWorkbar(props: {
         const activeTab = panel.tabs.find((tab) => tab.id === panel.activeTabId);
         const showingLauncher = panel.launcherOpen || !activeTab;
         const visible =
-          placement === 'right' ? !props.rightCollapsed : props.bottomOpen;
+          !props.hidden &&
```
Astro-Han, line 706: [P3] Cover the bottom placement too, since this expression now governs both. The fix is correct for `bottom` as well — `!props.hidden` gates the placement branch rather than sitting inside it — but the new spec only exercises `right`, so the bottom Workbar's freedom from the same leak is currently a property of how the condition happens to be parenthesised rather than something a test would defend. A future refactor that splits the condition per placement, which is exactly the shape the code had before this PR, would regress `bottom` silently. Asserting both placements disappear and return costs one more locator in the spec you already added.

Astro-Han (commented):
Thanks — this is the right one-line fix in the right place. Reviewed exact head `35558ed7b26a622bbd435f1588eb43d5ea26726b`.

The diagnosis holds up: `ChatWorkbar` already forwards the shell's obscured state, and `SessionWorkbar` was applying it only to panel content while the titlebar-level toolbar kept rendering. Folding `hidden` into the placement visibility condition removes the whole frame through the existing `data-collapsed` path, which is `display: none` in `chat-detail.css` — so the Workbar leaves the layout and the accessibility tree together and nothing stays focusable behind Settings. Panel state is untouched, which is what makes the restore-on-close assertion in the new spec meaningful. The Chinese accessible names in the spec match the fixture, which seeds `uiLocale: 'zh'`.

One coverage point inline. No findings above P3, and no CI checks are currently reported for this head.

Review disclosure: this review was prepared with Claude Code, which read the diff at this head, checked the collapsed-state CSS and the e2e locale fixture, and looked for other consumers of `props.hidden` in the component. The human contributor reviewed it before posting.

Astro-Han (approved):
Approving at `35558ed7b26a622bbd435f1588eb43d5ea26726b`. My review of this head found nothing above P3, so nothing here should hold the merge — the inline note stands as a suggestion, to take or leave.

PR #3261, opened 2026-08-19
ci: reduce pull request runner usage

Astro-Han:
## Summary

Reduce automatic pull request runner starts to one stable core gate.

- Run planning, planner contract tests, lint/format, build/typecheck, affected workspace tests, Runtime Host tests, Electron E2E/alignment, and Storybook as selected steps in one Ubuntu `test` job.
- For changes that can affect the published CLI, build and validate an installed Linux/Node 24 tarball in that same job. The planner is the single authority for selecting this check.
- Keep the full CLI matrix, including macOS, Windows, Node 22, and real Harbor/Pier validation, available through manual dispatch and the release `workflow_call`; it no longer starts automatically for pull requests.
- Keep Windows baseline on its existing nightly schedule and manual dispatch. Keep Windows recovery on manual dispatch. Neither starts automatically for pull requests.
- Run the core job for pull requests and `main` pushes. The `main` run validates the exact merged delta because this repository does not currently require pull requests to stay up to date with the base branch.

For this PR's own CI-sensitive diff, the automatic path changes from nine jobs to `CI / test` only. The tradeoff is less platform evidence on every pull request and less lane-level parallelism; the retained installed-package smoke covers the common CLI packaging failure mode without another runner.

This does not add a new daily or weekly tier. The existing Windows nightly remains in place.

Fixes #3259

## Verification

- `node --test --test-concurrency=1 scripts/ci-test-plan.test.mjs scripts/release-cli-workflow-policy.test.mjs` (22 tests)
- `npm run windows:inventory` (3 tests; generated inventory current)
- `actionlint .github/workflows/ci.yml .github/workflows/windows-baseline.yml .github/workflows/windows-recovery.yml .github/workflows/cli-package-validation.yml`
- `npm run format:check`
- `git diff --check`
- `npm --userconfig=/dev/null run release:cli:pack && npm --userconfig=/dev/null run release:cli:smoke` (clean isolated build, offline installed-package validation passed)

Repository-wide tests and the Electron E2E suite were not run locally. The changed workflow will provide the selected CI evidence.

## Discussion and rollout

The `dev@maka.apache.org` discussion remains open, and this PR is ready for concrete review. Do not merge while there are unresolved objections. If there are no objections after the stated 72-hour window, it can proceed under lazy consensus after human review of the final diff.

The stable required check remains `CI / test`. Live verification found no branch protection or repository ruleset that requires the removed platform job names.

## AI use

Select exactly one:

- [ ] No generative tool made a substantive contribution
[...]

PR #3262, opened 2026-08-19
chore: add .asf.yaml for ASF infrastructure self-service

Astro-Han:
## Summary

ASF committers have no GitHub admin UI, so `.asf.yaml` is the only self-service path to repository settings. This adds one.

| Setting | Now | After |
|---|---|---|
| Branch protection on `main` | none at all | 1 approving review, dismiss stale reviews, require the `test` check |
| Topics | `[]` | 13 topics |
| Description | `Maka — local-first AI desktop assistant` | replaced with the incubating description |
| Delete branch on merge | off | on |

`features`, `enabled_merge_buttons` and `dependabot_alerts` already match what is declared. They are written down as a baseline for the next change to this file, not as drift detection: asfyaml returns early when the file is unchanged, so it never reconciles settings altered elsewhere.

No `notifications` block: `commits@maka.apache.org` already receives PR, issue and push events by Infra default, 97 messages in the last month. No `dependabot_updates` either, since that key maps to `disable_automated_security_fixes()` and Dependabot security updates are live here (#3256, #3257).

Release tags stay unprotected. `protected_tags` is a no-op in asfyaml (`protected_tags.py` only prints a notice; GitHub retired tag protections in favour of rulesets), so covering tags means a `rulesets` entry with `type: tag`. That is worth doing and is not in this PR.

## Required check

`#3261` landed, so `ci.yml` now runs a single `test` job on every pull request, covering lint, format, build, tsc, knip, the selected workspace suites, e2e and Storybook. Requiring that one context is complete coverage, and no other workflow defines a job by that name.

## Review focus

**This is a governance change, and the cost is concrete.** Of the last 30 merged PRs, 14 had zero approving reviews, and 15 were merged by their own author. Approvals came almost entirely from one person. After this merges, every one of those paths needs a second person, which makes the fast path in [CONTRIBUTING.md:65](../blob/main/CONTRIBUTING.md#L65) unusable as written.

Dependabot is the sharpest edge: up to 20 open PRs a week, none of which can self-approve, and `allow_auto_merge` stays off. Turning it on is a separate governance decision and is not in this PR.

Discussion on dev@: https://lists.apache.org/thread/p0g84t1pqcop6bk7859k0j63s952ls98

## Remaining work

- [ ] Rewrite the fast-path clause in `CONTRIBUTING.md` (lines 65, 67)
- [ ] Mirror it in `CONTRIBUTING.zh-CN.md` (lines 63, 65)
- [ ] Update `CONTRIBUTING.md:43` / `CONTRIBUTING.zh-CN.md:41`, which still say project decisions move to the ASF list "once available" — dev@maka.apache.org now exists

## Verification

`yaml.safe_load` is not enough here: asfyaml parses twice, and the round trip changes the result. Run from the repository root:

```sh
uv run --with strictyaml python - <<'PY'
import strictyaml as sy
gh = sy.dirty_load(open('.asf.yaml').read(), allow_flow_style=True)['github']
one, two = gh['description'].data, sy.dirty_load(gh.as_yaml(), allow_flow_style=True)['description'].data
print(len(one), len(two), "OK" if one == two else "DESCRIPTION POLLUTED")
PY
```

Prints `184 184 OK`. Before the last revision it printed `184 238 DESCRIPTION POLLUTED`: `as_yaml()` re-indents a folded scalar, which pulled a following comment into it and would have appended that comment to the repository description.

`contexts: [test]` verified against a real run:

```sh
gh api repos/apache/maka/commits/101a3ce4475efc2e197101c5f4309b333f1a0baf/check-runs \
  --jq '.check_runs[] | "\(.name) \(.conclusion)"'
```

The job carries no `if:` condition, so it runs on every pull request and always reports. The check-run name is `test`, not `CI / test`.

Not run: lint, format, typecheck and the test suites. This PR touches only `.asf.yaml`, which none of them cover.

## AI use

Select exactly one:

Tool(s) and scope: Claude Code drafted the file and this description, ran the validation above, and reviewed the result adversarially. That review caught the folded-scalar bug, the `dependabot_updates` regression, and an earlier claim that required status checks could not be configured yet. The human contributor reviewed the final diff and owns the decision to submit it.

## Checklist

- [ ] No

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,0 +1,64 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Reference: https://github.com/apache/infrastructure-asfyaml
+
+github:
+  description: >-
+    Apache Maka (Incubating) is a local-first AI agent workspace.
+    Model messages, tool calls, tool results, permission decisions,
+    and termination events are recorded as an append-only log.
+  labels:
+    - ai
+    - ai-agent
+    - agent-runtime
+    - local-first
+    - llm
+    - tool-use
+    - event-sourcing
+    - desktop
+    - electron
+    - typescript
+    - cli
+    - apache
+    - maka
+  features:
+    issues: true
+    wiki: false
+    projects: true
+    discussions: false
+  enabled_merge_buttons:
+    squash: true
+    merge: false
+    rebase: false
+  pull_requests:
+    del_branch_on_merge: true
+  dependabot_alerts: true
+  protected_branches:
+    main:
+      required_pull_request_reviews:
+        dismiss_stale_reviews: true
+        required_approving_review_count: 1
+      required_status_checks:
+        strict: false
+        # test is the single job in .github/workflows/ci.yml, covering lint,
+        # format, build, tsc, knip, the selected workspace suites, e2e and
+        # Storybook. Renaming it there, or adding a paths filter that stops
+        # ci.yml from running, freezes every pull request: the check never
+        # reports and no committer can override it.
+        contexts:
+          - test
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -39,8 +39,8 @@ github:
   features:
     issues: true
     wiki: false
-    projects: true
-    discussions: false
+    projects: false
+    discussions: true
   enabled_merge_buttons:
     squash: true
     merge: false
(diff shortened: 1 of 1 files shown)


Astro-Han:
@wolfboys Thanks for the pointer, this is the .asf.yaml you suggested.

It covers repository metadata, merge policy, and branch protection on main.

Discussion on dev@: https://lists.apache.org/thread/p0g84t1pqcop6bk7859k0j63s952ls98

The governance item is branch protection: requiring one approving review and a
passing `test` check on main, which has no protection of any kind today. A
mentor review before this leaves draft would be welcome.

(Updated. An earlier version of this comment said the file also configured
notification routing, and that merging it would start mailing-list delivery.
Both were wrong: commits@maka.apache.org already receives PR, issue, and push
events by Infra default, so the notifications block was dropped. The correction
is on the dev@ thread above.)

Astro-Han:
@M4n5ter This changes how merging works for everyone with write access, and you are the person it affects most. Of the last 40 merged PRs, 15 were yours, and you and I are the only two who have actually pressed merge.

After this lands, every PR needs an approving review from someone else. Yours will need mine, mine will need yours. `allow_auto_merge` stays off, so Dependabot PRs need a human each time too, and `dismiss_stale_reviews` means a force-push resets the approval.

I think that trade is right for a podling, but you are the one who will live with it daily. If you think the cost is wrong, or that the count should be different, I would rather hear it now than after it is enforced.

Discussion on dev@: https://lists.apache.org/thread/p0g84t1pqcop6bk7859k0j63s952ls98

M4n5ter:
Thanks for calling this out. I agree that requiring one independent approval is the right tradeoff for the podling, and I’m prepared to share the review load. I would keep the count at one.

Before merging, please update both CONTRIBUTING files so the fast-path and dev@ guidance match the policy being enforced.

tisonkun (approved):

PR #3263, opened 2026-08-19
fix(desktop): preserve side chat composer elevation

chihumyum:
## Summary

The right-side Side Chat composer clipped the elevation painted by Astryx's
inner `ChatComposer` body, leaving the white composer visually unbounded against
the white workbar surface.

This change removes the side-only wrapper clipping and redundant wrapper radius
while preserving the compact Side Chat radius token. It also adds an Electron
E2E assertion for the overflow contract that keeps the inner elevation visible.

### Before

before

### After

after

## Verification

- `npm run format:check`
- `npm run lint`
- `npm run build`
- `npm run typecheck`
- `npm --workspace @maka/desktop run e2e -- session-workbar.spec.ts` (3 passed)
- `git diff --check`

## AI use

Select exactly one:

Tool(s) and scope: Codex diagnosed and implemented the scoped CSS fix, added the
regression assertion, ran validation, and prepared this pull request.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/e2e/session-workbar.spec.ts
+++ b/apps/desktop/e2e/session-workbar.spec.ts
@@ -68,6 +68,8 @@ test('narrow right workbar keeps launcher shortcuts and side-chat send button in
   await setRightWorkbarWidth(page, 320);
 
   const composerCard = companion.locator('.maka-composer-astryx');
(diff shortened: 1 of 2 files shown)


Astro-Han (approved):
LGTM. Nice and Clean fix.

PR #3266, opened 2026-08-20
feat(runtime): materialize verified native PDF inputs

### LINKED ISSUE #3164: feat(runtime): materialize native PDF inputs for verified PDF-capable providers
## Problem

Maka synchronizes PDF-input metadata and exposes `resolveModelPdfSupport`, but the runtime never consumes that capability. `AiSdkBackend.appendImageParts` filters attachments to `kind === image`; a PDF remains only a text placeholder even when the selected provider/model has a verified native PDF contract.

The storage side is already present: PDF attachments have a canonical kind and MIME type, are persisted as Session artifacts, and can be read through the existing bounded `AttachmentByteReader`. The missing boundary is provider request materialization.

## Provider research

AI SDK encoding support and provider support are separate facts. The current locked SDK produces these request shapes from one generic `file` part:

- OpenAI-compatible Chat: `type: file` with `filename` and Base64 `file_data`;
- Anthropic Messages: `type: document` with a Base64 `application/pdf` source;
- OpenAI Responses: `type: input_file` with `filename` and Base64 `file_data`.

That proves the SDK can lower a generic file part. It does not prove every OpenAI-compatible provider accepts the resulting wire.

### Verified first-party contracts

- **Anthropic Messages:** all active Claude models support PDF processing. The API accepts URL, Base64, or Files API references. The complete request is limited to 32 MB; PDFs must be standard and unencrypted; the page limit is 600, or 100 when the request context window is below 1M tokens.
- **OpenAI:** native Chat Completions and Responses accept PDF file inputs. Responses uses `input_file`; PDF processing includes extracted text and page images. Each file and the combined files in one request must remain below 50 MB.

### Not yet verified

- **Generic OpenAI-compatible providers:** the adapter can emit the OpenAI Chat `file_data` shape, but each provider must opt in only after its own contract is verified. Unknown compatible providers must remain unsupported.
- **Alibaba Token Plan / `qwen3.8-max`:** Alibaba Chat documentation lists text, image, audio, and video content but no `file`/`file_data` PDF shape. Alibaba Responses documents `input_file`, but currently states that this content type is supported only by `qwen3.5-ocr` and requires `file_url`. Therefore `qwen3.8-max` is not an acceptance target for this issue. Its generated `pdf` modality is insufficient authority for sending local PDF bytes.
- Alibaba PDF support, if added later, needs a separate provider contract. If the service continues to require `file_url`, it also needs a safe upload/temporary-URL design rather than silently exposing local files.

## Scope status

This is an independent Runtime attachment-capability issue. It is **not** a follow-up requirement for Qwen3.8 support, #3156, #3157, #3162, or #3163, and it does not block those deliverables. Alibaba/Qwen remains in this issue only as a negative boundary showing why generic OpenAI-compatible PDF support must not be inferred without provider evidence.

## Desired outcome

A model on a **verified PDF-capable provider/wire** receives an attached PDF as a provider-native file part on the current Turn and during durable replay. Unsupported, unknown, or unverified provider/model combinations continue to receive the bounded placeholder and never receive raw PDF bytes.

### First slice: Anthropic and native OpenAI

1. Replace the unused boolean-only assumption with an explicit provider/wire PDF-input contract. Do not infer generic-compatible support from a model name or generated modality alone.
2. Carry the resolved contract into `AiSdkBackend` from the production Runtime Host composition. The former Headless production path no longer exists on current `main` and is not part of this slice.
3. Materialize `AttachmentRef.kind === pdf` through the existing attachment byte reader as one generic AI SDK `file` part with exact `application/pdf`, filename, and bytes. Let the verified AI SDK provider adapter perform the wire-specific lowering.
4. Cover native Anthropic Messages and native OpenAI Chat/Responses. Keep all generic OpenAI-compatible providers disabled unless separately verified.
5. Keep current Turn, RuntimeEvent replay, StoredMessage fallback, steering, compaction replacement, and restart behavior consistent.

### Budget and failure policy

- Add a PDF sub-budget and a combined raw-binary attachment budget. Separate independent image/PDF limits are not sufficient because Anthropic applies its 32 MB cap to the complete request and Base64 adds roughly one-third overhead.
- Preserve the existing 12 MB image sub-limit. Choose the PDF and combined limits from measured encoded request size, leaving room for text, schemas, JSON, and other attachments.
- Deduplicate budget decisions by durable occurrence across provider steps, as the image path does today.
- Missing, unreadable, oversized, unsupported, or locally invalid PDFs must degrade to a safe bounded explanation; PDF bytes must never enter transcript text, logs, RuntimeEvents, or diagnostics.
- Preflight parsing for encrypted/page-limit PDFs is tracked in #3284; bounded recovery after provider rejection is tracked in #3285. Neither is a merge gate for this first slice. Ordinary provider errors remain governed by the existing bounded `ModelFailure` handling.

## Acceptance criteria

- `resolveModelPdfSupport` is no longer dead metadata; the runtime has an explicit verified provider/wire gate.
- Native Anthropic receives a `document` block with Base64 `application/pdf` data through the SDK.
- Native OpenAI Chat receives `type: file`; native OpenAI Responses receives `type: input_file`.
- A generic OpenAI-compatible connection does not receive PDF bytes by default.
- Current Turn, RuntimeEvent replay, StoredMessage fallback, steering, compaction, and restart all present the same durable PDF occurrence without duplicating the user message.
- Image and PDF attachments respect both their subtype caps and the combined binary-request budget.
- Unsupported models and local attachment failures keep the safe placeholder. Provider request failures use the existing bounded `ModelFailure` path; this slice does not add PDF-specific automatic retry.
- Focused live verification covers at least one native Anthropic model and one native OpenAI model when protected credentials are available.

## Non-goals

- Claiming PDF support for `qwen3.8-max` or another generic-compatible model without provider evidence.
- Uploading local PDFs to a third-party file service or creating public temporary URLs.
- OCR fallback for models without native PDF support.
- Automatically extracting PDF text into the prompt.
- Treating arbitrary Office or binary files as provider-safe PDF/document inputs.
- Preflight parsing for encrypted PDFs/page counts (#3284) or automatic PDF-specific retry after provider rejection (#3285).

## References

- OpenAI file inputs: https://developers.openai.com/api/docs/guides/file-inputs
- Anthropic PDF support: https://platform.claude.com/docs/en/build-with-claude/pdf-support
- Alibaba Chat API: https://help.aliyun.com/zh/model-studio/qwen-api-via-openai-chat-completions
- Alibaba Responses API: https://help.aliyun.com/zh/model-studio/qwen-api-via-openai-responses

Disclosure: this issue was drafted and revised with OpenAI Codex assistance and reviewed and approved by the human contributor.

tianrking:
## Summary

- Connect the existing PDF modality metadata to an explicit native-provider contract: Anthropic Messages and native OpenAI Chat/Responses are enabled, while generic-compatible, subscription, unknown-model, and inventory-denied routes remain fail-closed.
- Materialize durable PDF artifacts through the existing bounded byte reader across the current Turn, RuntimeEvent replay, StoredMessage fallback, steering, and compaction replacement without copying bytes into transcript text or diagnostics.
- Preserve the 12 MiB image cap, add a 16 MiB PDF cap and an 18 MiB combined raw-binary cap, and cache keep/omit decisions by durable occurrence across provider steps.
- Lock the AI SDK lowering shapes for Anthropic `document`, OpenAI Chat `file`, and OpenAI Responses `input_file`, including a Host-to-provider streaming integration test.

Fixes #3164

## Verification

- `npm run build:test` — all workspaces built successfully.
- `npm run typecheck` — all workspace typechecks passed.
- `biome lint .` — 2,451 files checked, passed.
- `biome format .` — 1,544 files checked, passed.
- focused Runtime backend and provider-wire tests — 213 passed.
- focused Core attachment/model tests — 11 passed.
- Host composition PDF wire test — 1 passed against a real local streaming HTTP boundary.
- `git diff --check` and staged secret-pattern scan — passed.

The full Runtime command was also attempted on Windows. The affected PDF/attachment tests passed, but the overall suite did not finish cleanly because unrelated platform fixtures failed or hung in SQLite temp-file cleanup (`EBUSY`), process-tree, shell, and PTY paths. The 11 SQLite failures reproduced in isolation with the same locked-file cleanup error and do not touch changed code.

Protected OpenAI/Anthropic live calls were not run because no provider credentials were available. Exact request bodies are covered against the locked SDK adapters, and the Host composition test exercises artifact reading, backend construction, streaming transport, and final OpenAI Chat wire serialization without an external credential.

## Review focus

The provider/wire contract is intentionally narrower than SDK encoding support: a compatible relay does not receive raw PDF bytes merely because its adapter can serialize a file part. Budget accounting uses bytes actually read, does not charge failed reads, preserves the existing image limit, and applies one combined cap before Base64 expansion.

## AI use

Select exactly one:

Tool(s) and scope: OpenAI Codex assisted with issue and duplicate-PR analysis, implementation, tests, local verification, and PR drafting. @tianrking requested the contribution and remains the human contributor of record.

## Checklist

- [ ] No

### DIFF
--- a/packages/core/src/attachments.ts
+++ b/packages/core/src/attachments.ts
@@ -21,9 +21,14 @@ import type { AttachmentRef, StorageRef } from './events.js';
 import { isCanonicalArtifactEntityId } from './artifacts.js';
 
 /** Lives in core so @maka/runtime and @maka/storage share one type without a package cycle. */
-export type AttachmentByteReader = (
-  ref: StorageRef,
-) => Promise<{ ok: true; bytes: Uint8Array } | { ok: false; reason: string }>;
+export type AttachmentByteReader = (ref: StorageRef) => Promise<
+  | {
+      ok: true;
+      bytes: Uint8Array;
+      mimeType?: string; // MIME sniffed from the loaded bytes when available.
+    }
+  | { ok: false; reason: string }
+>;
 
 export const ATTACHMENT_RESOURCE_PREFIX = 'maka://runtime/attachments';
 
@@ -77,6 +82,20 @@ export const READ_IMAGE_TOO_LARGE_MESSAGE = `Image exceeds the ${MAX_READ_IMAGE_
 export const MAX_PROVIDER_IMAGE_REQUEST_BYTES = 12 * 1024 * 1024;
 export const PROVIDER_IMAGE_BUDGET_EXCEEDED_MESSAGE = `Image was read, but the per-request image budget (${MAX_PROVIDER_IMAGE_REQUEST_BYTES / 1024 / 1024}MB across all images this turn) was exceeded; earlier images were sent and this one was omitted. Read fewer or smaller images.`;
 
+/**
+ * Native PDF requests are Base64 encoded. Sixteen raw MiB expands to roughly
+ * 21.4 MiB, leaving headroom under Anthropic's 32 MiB whole-request limit for
+ * text, tool schemas, JSON framing, and other content.
+ */
+export const MAX_PROVIDER_PDF_REQUEST_BYTES = 16 * 1024 * 1024;
+
+/**
+ * Shared raw-byte ceiling across image and PDF inputs. Eighteen raw MiB
+ * expands to 24 MiB in Base64, preserving 8 MiB of whole-request headroom on
+ * the strictest verified native PDF route.
+ */
+export const MAX_PROVIDER_BINARY_REQUEST_BYTES = 18 * 1024 * 1024;
+
 const MIME_BY_EXTENSION: Readonly<Record<string, string>> = {
   png: 'image/png',
   jpg: 'image/jpeg',
@@ -107,10 +126,9 @@ export function guessMimeFromName(fileName: string): string {
 
 /**
  * Route a MIME type to an {@link AttachmentRef} kind. The runtime
- * consumption split is image vs. everything-else (images become provider
- * image parts; other kinds are read on demand by the model via Read), so this
- * only needs to single out the kinds that change
- * consumption or display. Unknown / unmapped MIME falls back to `other`.
+ * consumption split singles out images and PDFs (authorized routes can send
+ * them as provider file parts); text-like kinds are read on demand by the
+ * model via Read. Unknown / unmapped MIME falls back to `other`.
  *
  * `fileName` is consulted for kinds whose MIME is unreliable across OSes
  * (Office documents arrive as `application/octet-stream` or a long
--- a/packages/runtime/src/ai-sdk-backend.ts
+++ b/packages/runtime/src/ai-sdk-backend.ts
@@ -95,10 +95,12 @@ import {
   YIELD_AGENT_GRAPH_TOOL_NAME,
   type YieldAgentGraphToolResult,
 } from './stream-graph-supervisor-tools.js';
-import type { AttachmentByteReader } from '@maka/core/attachments';
 import {
+  MAX_PROVIDER_BINARY_REQUEST_BYTES,
   MAX_PROVIDER_IMAGE_REQUEST_BYTES,
+  MAX_PROVIDER_PDF_REQUEST_BYTES,
   PROVIDER_IMAGE_BUDGET_EXCEEDED_MESSAGE,
+  type AttachmentByteReader,
 } from '@maka/core/attachments';
 import { stripUndefinedDeep } from '@maka/core/tool-args-identity';
 import { pricingModelKey } from '@maka/core/usage-stats/pricing';
@@ -178,7 +180,7 @@ import { compactionDecisionDiagnosticPatch } from './compaction-boundary.js';
 import type {
   AutomaticMemoryCompactionDecision,
   AutomaticMemoryCompactionDispatch,
-  ProviderImageBudget,
+  ProviderAttachmentBudget,
 } from './ai-sdk-compaction.js';
 import {
   contextDiagnosticsCompactionOf,
@@ -207,7 +209,6 @@ import {
   formatTextWithInlineRefs,
   steeringMessagesMissingFromBase,
   steeringModelMessage,
-  steeringProviderOptions,
   stripSteeringMessages,
   type RuntimeEventModelReplayItem,
   type RuntimeEventModelReplayPlan,
@@ -835,9 +836,10 @@ export interface AiSdkBackendInput extends AiSdkCompactionCapabilities {
    */
   recordToolArtifacts?: ToolArtifactRecorder;
   /**
-   * Optional attachment byte reader. When set, image attachments on the current
-   * user turn may be rendered as provider image parts instead of placeholder text.
-   * Caller wires this to the session ArtifactStore; runtime never imports storage.
+   * Optional attachment byte reader. When set, authorized image and PDF
+   * attachments may be rendered as provider file parts instead of placeholder
+   * text. Caller wires this to the session ArtifactStore; runtime never imports
+   * storage.
    */
   readAttachmentBytes?: AttachmentByteReader;
   /**
@@ -845,7 +847,11 @@ export interface AiSdkBackendInput extends AiSdkCompactionCapabilities {
    * image parts; false/unknown stay as text refs with a fallback note.
    */
   supportsVision?: boolean;
+  /** True only when Runtime Host verified the provider identity and resolved wire for native PDFs. */
+  supportsNativePdfInput?: boolean;
   maxProviderImageRequestBytes?: number;
+  maxProviderPdfRequestBytes?: number;
+  maxProviderBinaryRequestBytes?: number;
   /** Host-owned bounded long-term-memory extraction. Source tools are Runtime-reserved. */
   memoryExtraction?: MemoryExtractionSourceCapabilities;
 }
@@ -989,11 +995,15 @@ class TurnScope {
   watchdog: StreamWatchdog | null = null;
   runTrace: RunTrace | null = null;
   /**
-   * Image allowance for this turn, accumulated across its provider steps. Owned
-   * by the scope so an overlapping turn cannot spend it, and non-null for the
-   * scope's whole life so no path has to decide what "no budget" means.
+   * Binary attachment allowance for this turn, accumulated across its provider
+   * steps. Owned by the scope so an overlapping turn cannot spend it, and
+   * non-null for the scope's whole life so no path has to decide what "no
+   * budget" means.
    */
-  readonly imageBudget: ProviderImageBudget = { used: 0, decisions: new Map() };
+  readonly attachmentBudget: ProviderAttachmentBudget = {
+    used: { image: 0, pdf: 0, total: 0 },
+    decisions: new Map(),
+  };
   /**
    * User messages steered into this turn, drained from the caller's queue at
    * step boundaries. Each entry is the canonical envelope-wrapped user
@@ -1118,8 +1128,8 @@ export class AiSdkBackend implements AgentBackend {
       modelAdapter: this.modelAdapter,
       createProviderRequestTracker: (trackerInput) =>
         this.createProviderRequestTracker(trackerInput),
-      materializeRuntimeReplayPlan: (plan, imageBudget, checkpoint) =>
-        this.materializeRuntimeReplayPlan(plan, imageBudget, undefined, checkpoint),
+      materializeRuntimeReplayPlan: (plan, attachmentBudget, checkpoint) =>
+        this.materializeRuntimeReplayPlan(plan, attachmentBudget, undefined, checkpoint),
       canReplayProviderNative: (plan) => this.canReplayProviderNative(plan),
       appendTurnTailPrompt: (content, turnTailPrompt) =>
         this.appendTurnTailPrompt(content, turnTailPrompt),
@@ -1339,7 +1349,12 @@ export class AiSdkBackend implements AgentBackend {
       orchestrationMode: identity.orchestrationMode,
       ...(identity.invocationId ? { invocationId: identity.invocationId } : {}),
       materializeDefaultToolResultOutput: ({ toolCallId, output }) =>
-        this.materializeToolResultOutput(identity.scope().imageBudget, output, false, toolCallId),
+        this.materializeToolResultOutput(
+          identity.scope().attachmentBudget,
+          output,
+          false,
+          toolCallId,
+        ),
       spawnChildAgent: input.spawnChildAgent,
       spawnChildSession: input.spawnChildSession,
       prepareChildAgentResume: input.prepareChildAgentResume,
@@ -1867,7 +1882,7 @@ export class AiSdkBackend implements AgentBackend {
         const currentUserContent = input.continuation
           ? undefined
           : await this.buildCurrentUserContent(
-              scope.imageBudget,
+              scope.attachmentBudget,
               input.text,
               input.attachments,
               input.quotes,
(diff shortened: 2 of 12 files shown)


MoonOld (commented):
One focused blocker before this PR closes #3164:

**P2 — durable replay drops PDF attachments from steering messages.** The live steering path materializes attachments through `appendAttachmentParts`, but `materializeRuntimeReplayItem` returns `item.content` directly whenever `item.steering` is set, even though the replay plan preserves `item.attachments`. A later Turn or restart therefore replays the steering envelope text without the PDF file part.

Please route this branch through `appendAttachmentParts` using the stable `steering:${item.steering.eventId}` decision key, preserve `steeringProviderOptions`, and extend the existing prior-turn steering replay test with a PDF assertion.

I rechecked the scope and narrowed #3164: base-URL allowlisting, the removed Headless path, and encrypted/page-limit/provider-rejection-specific recovery are not blockers for this first slice.

packages/core/src/attachments.ts
```diff
@@ -58,6 +58,20 @@ export const READ_IMAGE_TOO_LARGE_MESSAGE = `Image exceeds the ${MAX_READ_IMAGE_
 export const MAX_PROVIDER_IMAGE_REQUEST_BYTES = 12 * 1024 * 1024;
 export const PROVIDER_IMAGE_BUDGET_EXCEEDED_MESSAGE = `Image was read, but the per-request image budget (${MAX_PROVIDER_IMAGE_REQUEST_BYTES / 1024 / 1024}MB across all images this turn) was exceeded; earlier images were sent and this one was omitted. Read fewer or smaller images.`;
 
+/**
+ * Native PDF requests are Base64 encoded. Sixteen raw MiB expands to roughly
+ * 21.4 MiB, leaving headroom under Anthropic's 32 MiB whole-request limit for
+ * text, tool schemas, JSON framing, and other content.
+ */
+export const MAX_PROVIDER_PDF_REQUEST_BYTES = 16 * 1024 * 1024;
```
Astro-Han, line 90: [P1] Bound the work the request telemetry does on these bytes before raising the cap that makes them routine. `capturePreparedProviderRequest` runs on every provider request — the tracker exists whenever model-call accounting does, so this is not gated on `recordProviderRequestCapture` — and `canonicalize` has no typed-array branch, so a `Uint8Array` is expanded element-wise through `Object.keys().sort()`. Measured against this repository's built `packages/runtime/dist` with capture persistence disabled: an 8 MiB attachment costs 3.5 s per step, 12 MiB (today's image cap) 7.5 s, and 16 MiB (this cap) 9.5 s per step with `requestBytes` reported as 222.4 MiB; across three steps that is 28.5 s during which a 10 ms timer scheduled beforehand never fires, and RSS grows 4.4 to 7.6 GiB. With capture persistence enabled the bytes are also verbatim in `serializedRequest` — I matched the `%PDF-` header inside it — which contradicts this PR's "without copying bytes into transcript text or diagnostics" and #3164's "PDF bytes must never enter transcript text, logs, RuntimeEvents, or diagnostics". This is not introduced here; images take the same path on `main`. But this slice raises the ceiling to 18 MiB combined and makes ten-plus-MiB attachments ordinary, because a 16 MiB PDF is a normal document and a 16 MiB PNG is not. Give `request-shape.ts` a binary branch that represents file data as `{ byteLength, sha256 }` — substituting that summary drops the same call from 9.5 s to 0.4 ms and `requestBytes` from 222.4 MiB to 312 B — and add a regression asserting that a file part's bytes never appear in `serializedRequest`. Landing it as a separate prerequisite is fine; landing native PDF input without it is not.

Astro-Han (commented):
Thanks for taking this on — the provider evidence in #3164 and the wire-lowering tests here made the contract straightforward to check against the real adapters rather than against claims.

Reviewed exact head `303d20a3a295cfa54be20f7aaacaebe0a8707e87`.

The provider/wire gate is the right minimal shape: `resolveModelPdfInputContract` fails closed for relays, subscriptions, and unknown models, an explicit provider inventory outranks generated modality metadata, and the three wire shapes are locked against the real SDK adapters rather than asserted. The turn budget is created once in `TurnScope` and the decision keys line up across current turn, RuntimeEvent replay, and steering. The implementation does not yet satisfy its bounded-work contract, so I recommend revising it before approval.

The concrete current-head findings are inline. At the architecture level the gap is that "bounded" is enforced on what reaches the provider payload, but not on the work done to produce and observe that payload: bytes are read again per step after a decision is already cached, and the always-on request telemetry expands them element-wise on every step. Both belong before materialization, not after.

@MoonOld's steering finding reproduces on this head, and it is pre-existing for images as well — `materializeRuntimeReplayItem` ignores `item.attachments` on the steering branch on current `main` too, while the replay plan does preserve them. A steered PDF replays as envelope text plus an `<attachment>` ref and zero file parts. Routing that branch through `appendAttachmentParts` with `steering:${item.steering.eventId}` fixes it and keeps dedupe aligned with the live drain and the stored-message sidecar.

Verification: I built this head in an isolated worktree and ran `packages/runtime/dist/__tests__/ai-sdk-backend.test.js` — 210/212 pass, the two failures being regressions I wrote for the steering gap and the per-step re-read. A 52-line change (cache-before-read, plus the steering branch) takes it to 212/212, including `enforces the PDF subtype budget from bytes read, not attachment metadata`, so the "charge actual bytes" semantics survive the fix.

Review disclosure: this review was prepared with Claude Code, which read the diff, traced the call paths, and executed the reproductions and measurements quoted inline against this head. The human contributor reviewed the findings and the measurements and chose to publish this COMMENT review. The P1 timings come from a single macOS machine and are order-of-magnitude evidence, not a benchmark. No CI checks are currently reported for this head.

packages/runtime/src/ai-sdk-backend.ts
```diff
@@ -4480,87 +4500,148 @@ export class AiSdkBackend implements AgentBackend {
...
+      if (attachment.kind === 'pdf' && attachment.mimeType !== 'application/pdf') {
+        parts.push({
+          type: 'text',
+          text: `PDF attachment "${attachment.name}" was omitted because its media type is not application/pdf.`,
+        });
+        continue;
+      }
+      let read: Awaited<ReturnType<AttachmentByteReader>>;
+      try {
+        read = await this.input.readAttachmentBytes(attachment.ref);
```
Astro-Han, line 4369: [P2] Consult the decision cache before reading the bytes, not after. The cache is checked inside `chargeAttachmentBudget`, which runs once `readAttachmentBytes` has already returned, so an attachment whose omission was decided on the first step is read in full again on every later step; `materializeToolResultOutput` below already does this in the correct order. Reproduced on a durable turn with `loadTurnRuntimeEvents` wired, as Host composition does: reads equal steps + 1 for the omitted ref — 3 reads over 2 steps, 5 over 4. In production each read is a full `readDurableAttachmentBinary` with no `maxBytes` passed, so up to `MAX_ATTACHMENT_BYTES` (50 MiB), serialized on the ArtifactStore queue and blocking other artifact work in the session. Move the `decisionKey` computation and a non-`keep` early return above the read, and add a multi-step regression counting reads per ref.

packages/runtime/src/ai-sdk-backend.ts
```diff
@@ -4480,87 +4500,148 @@ export class AiSdkBackend implements AgentBackend {
...
-    for (const [index, image] of images.entries()) {
-      const read = await this.input.readAttachmentBytes(image.ref);
+    > = [{ type: 'text', text: fallbackText }];
+    const omitted = { image_limit: 0, pdf_limit: 0, combined_limit: 0 };
+    for (const [index, attachment] of binaryAttachments.entries()) {
+      const isEligible =
+        (attachment.kind === 'image' && this.input.supportsVision === true) ||
+        (attachment.kind === 'pdf' && this.input.pdfInputContract !== undefined);
+      if (!isEligible) continue;
+      if (attachment.kind === 'pdf' && attachment.mimeType !== 'application/pdf') {
```
Astro-Han: [P3] Degrade a mislabeled PDF locally instead of spending a provider round trip on it. This check compares declared metadata only, but `readPreparedBinary` already sniffs magic bytes and returns the true MIME — `%PDF-` is in the allowlist, so a file that is not any allowed binary type does fail closed with `unsupported_mime`. `createAttachmentByteReader` then discards `result.mimeType`, so a PNG named `brief.pdf` is uploaded as `application/pdf` and degrades through a provider 400 rather than the bounded local explanation #3164 asks for. Surface the sniffed MIME through the reader and compare it here; that is smaller than the preflight parsing deferred to #3284.

packages/runtime/src/model-runtime.ts
```diff
@@ -35,6 +39,15 @@ export interface ResolvedModelRuntime {
   applyPatchProfile: ApplyPatchProfile | null;
 }
 
+/**
+ * Provider/wire combinations whose native API contract has been verified to
+ * accept an AI SDK PDF file part. Provider identity is intentional: a relay
+ * using the same adapter or wire does not inherit first-party PDF support.
+ */
+export type ModelPdfInputContract =
```
Astro-Han: [P3] Say provider identity, not endpoint, in this contract's wording, and give the union a consumer or drop it. The gate is `providerType` plus wire; the endpoint is not checked, and this PR's own test authorizes `providerType: 'openai'` with `baseUrl: 'https://provider.invalid/v1'`. Narrowing base-URL allowlisting out of this slice is reasonable and images are in the same position, but "verified first-party provider wire" reads as endpoint verification. Separately, nothing consumes `providerType` or `wire` — `AiSdkBackend` only tests the contract for presence — so either branch on the wire where it matters, such as per-wire caps, or let it be a boolean.

packages/runtime/src/ai-sdk-backend.ts
```diff
@@ -4480,87 +4500,148 @@ export class AiSdkBackend implements AgentBackend {
...
+          attachment.kind === 'image' || attachment.kind === 'pdf',
+      ) ?? [];
+    if (binaryAttachments.length === 0) return textContent;
+    const hasUnsupportedImages =
+      this.input.supportsVision !== true &&
+      binaryAttachments.some((attachment) => attachment.kind === 'image');
+    const fallbackText = hasUnsupportedImages
+      ? appendNonVisionImageFallbackNotice(textContent)
+      : textContent;
+    const eligibleAttachments = binaryAttachments.filter(
```
Astro-Han: [P3] Keep one copy of the eligibility rule. `eligibleAttachments` is computed only to decide the early return and the loop recomputes the same predicate per item, so the two must be kept in sync by hand. Iterating `eligibleAttachments` directly removes the duplicate; note the decision key indexes into `binaryAttachments`, so whichever array you iterate has to stay consistent with the key.

Astro-Han:

- `packages/runtime/src/model-runtime.ts`

```
# resolve, then
git push --force-with-lease
```

---

tianrking:
Rebased onto the current `main` and force-pushed `16e99b580`.

Addressed the review points:
- durable steering replay now materializes authorized attachments through `appendAttachmentParts` with the stable `steering:<eventId>` key while preserving steering identity;
- request capture summarizes every `Uint8Array` as `{ byteLength, sha256 }` before serialization, so raw file bytes cannot enter telemetry/artifacts;
- cached non-keep attachment-budget decisions are checked before another durable read;
- the storage reader returns the sniffed MIME type, and a declared PDF is omitted if its loaded bytes are not `application/pdf`;
- the PDF gate is a consumed boolean based on first-party provider identity plus resolved wire, and relays/subscriptions fail closed without runtime resolution;
- attachment eligibility is derived once with the original attachment index retained for the decision key.

Local verification on this final rebased commit:
- `biome check` for all 10 changed files;
- TypeScript project builds for core, storage, mcp, runtime, and runtime-host;
- focused regression tests for current/replay/steering PDF materialization, budget read caching, MIME rejection, request capture redaction, provider gating, storage authority, and host composition (30 passing tests).

GitHub created [the new CI run](https://github.com/apache/maka/actions/runs/32645778011), but it is currently `action_required` with no jobs started. An authorized maintainer action is needed before it can run and turn green. Please re-review after that completes.

Issue #3218, opened 2026-08-18
chore: audit residue — dead scripts, duplicated /skill: grammar, chat-readiness remnant


Astro-Han:
Found by a repo-wide simplification audit; zero-consumer claims re-verified with `git grep -aI`. Four small independent cleanups bundled here:

1. **`scripts/check-astryx-alignment.mjs`** (130 lines): no execution path — not in root `package.json` scripts, any of the 8 workflows, or `scripts/ci-test-plan.mjs` (which wires the sibling `check-astryx-surface-inventory.mjs` instead). Repo-wide references: the file itself and one doc line. Its `astryx-allow: raw-button` escape-hatch comment still spreads in sources.
2. **`scripts/deepseek-live-cost-baseline.mjs`** (2960 lines): only reference is the npm alias `cost:deepseek-baseline`; zero invocations, zero workflow/doc references outside `docs/archive/` (which declares itself non-current). Couples to Runtime/Storage internals and is the last user of the legacy compaction capability (see the History Compact V1 issue).
3. **`packages/cli/src/skill-token.ts`**: a near-verbatim second authority for the `/skill:` token grammar duplicating `packages/runtime/src/skill-invocation.ts`. The CLI copy's three main symbols have zero consumers; its two importers use only `skillInvocationPrefixAt` and a constant re-export, while the CLI already imports `@maka/runtime/skill-invocation` in 3 places.
4. **`apps/desktop/src/main/chat-readiness.ts`**: the admission gate is residue — the only production importer is `main-window.ts` importing `errorMessage`; `assertSessionCanSend` / `ensureSessionCanSendOrRebind` have no production callers. Collapse to the error helpers actually used.

seanxuu:
Claim this issue #3218. I will only be handling the /skill: duplicate CLI helper task to avoid conflicting with the cleanup of the other three items.

Xinbeok:
Has this issue been resolved? I want to claim this issue.

Astro-Han:
@Xinbeok I checked the current state of this issue on `main` and the related PRs — it's already fully covered by existing work, so I wouldn't claim it:

- **#1** `scripts/check-astryx-alignment.mjs` and **#2** `scripts/deepseek-live-cost-baseline.mjs` — deleted in **PR #3242** "chore: remove audit residue" (open, Closes this issue).
[...]

PR #3267, opened 2026-08-20
chore(cli): remove duplicated skill token helpers

seanxuu:
## Summary

Refs #3218.

Removes the unused CLI copy of skill-token parsing helpers. The remaining autocomplete helper now lives with its only caller, while highlighting imports the shared token grammar from Core.

## Verification

- `npx biome check packages/cli/src/pi-tui-pickers.ts packages/cli/src/skill-highlight-editor.ts`
- `npm --workspace maka-agent test` (315 passing)

## AI use

- [x] No generative tool made a substantive contribution

Tool(s) and scope: Codex assisted with code analysis, patch preparation, and local verification. I reviewed and submitted the change.


packages/cli/src/pi-tui-pickers.ts
```diff
@@ -266,6 +265,18 @@ function slashCommandPrefix(lines: string[], cursorLine: number, cursorCol: numb
...
 }
 
+function skillInvocationPrefixAt(
+  lines: string[],
+  cursorLine: number,
+  cursorCol: number,
+): { prefix: string; query: string } | null {
+  const currentLine = lines[cursorLine] || '';
+  const beforeCursor = currentLine.slice(0, cursorCol);
+  const match = /(?:^|\s)(\/skill:([A-Za-z0-9._-]*))$/.exec(beforeCursor);
```
Astro-Han, line 275: [P3] Derive this prefix pattern from the shared grammar instead of restating it. The deleted module's stated purpose was that `SKILL_INVOCATION_TOKEN_SOURCE` in Core is the single owner of where a `/skill:` token starts and ends, and the highlighter next door still imports it — but this literal re-declares the id charset `[A-Za-z0-9._-]`, so widening the charset in Core would leave autocomplete quietly refusing to offer names the rest of the system accepts. The divergence is pre-existing, not introduced here, but a PR whose whole point is removing the duplicated copy is the right place to close it. Build the regex from `SKILL_INVOCATION_TOKEN_SOURCE` with the trailing quantifier relaxed so an empty query still triggers, or at minimum add a test asserting both accept the same name charset.

Astro-Han (commented):
Thanks — this is a real duplicate, not a cosmetic one. Reviewed exact head `7fdac7fc19d8e75cd11c96184ac6c85dbd226e47`.

I confirmed the deletion is safe: `parseSkillInvocationTokens` and `stripSkillInvocationTokens` already live in `@maka/runtime/skill-invocation` and are what `root-turn-coordinator` and `interactive-turn-coordinator` actually call on the submit path, so the CLI copy had no consumers, and nothing else in the tree still references `skill-token`. Moving the one surviving helper next to its only caller and importing the grammar from Core is the right direction.

One follow-up inline: the module you deleted existed to keep the grammar in one place, and the helper that survived it still hardcodes its own copy of that grammar. No findings above P3. No CI checks are currently reported for this head.

Review disclosure: this review was prepared with Claude Code, which read the diff at this head and traced the removed helpers' consumers across the workspace. The human contributor reviewed it before posting.

Astro-Han (approved):
Approving at `7fdac7fc19d8e75cd11c96184ac6c85dbd226e47`. My review of this head found nothing above P3, so nothing here should hold the merge — the inline note stands as a suggestion, to take or leave.

PR #3277, opened 2026-08-20
refactor(runtime-host): retire client surface identity plumbing

### LINKED ISSUE #3216: chore(runtime-host): ClientSurface is a required handshake field with zero readers
Found by a repo-wide simplification audit; re-verified with `git grep -aI`.

`ClientSurface` is a required handshake field that no Host code reads:

- Type/decode/`requireSurface`: `packages/runtime-host/src/protocol/index.ts:88-95`, `:219`, `:402-414`; threaded into `ConnectionContext` (`operation-dispatcher.ts`), assigned once in `host-kernel.ts`.
- Producers exist in all five clients (desktop, tui, run, activation, capability-provider), but `git grep -aInw surface` across `src/server/` and `src/transport/` shows zero production readers.

Methodology note that likely explains its longevity: `protocol/index.ts` contains NUL bytes, so default `rg`/`grep` treat it as binary and silently skip it — absence checks must use `git grep -aI`.

Disappearance surface: the enum, decode + `requireSurface` validation, the `ConnectionContext` field, and the five client-side pass-throughs. If client identity is wanted later for routing/telemetry, it can be reintroduced with an actual reader.

Sun-GLiang:
## Summary

- Remove the unused `ClientSurface` identity field, Host connection context field, and local/remote/profile/CLI/Desktop pass-throughs.
- Make new Hosts accept hello frames both with and without the legacy `surface` field; decoded `ClientHello` values no longer retain it.
- Keep one private outgoing compatibility shim that sends `surface: "desktop"` so new Clients still connect to old Hosts that require the field.
- Preserve remote TUI SSH behavior through an explicit `interactiveSsh` intent instead of coupling it to handshake identity.
- Keep compatibility epoch 27 because Host decoding becomes more permissive and the Client shim preserves old-Host compatibility.

Related issue: [#3216](https://github.com/apache/maka/issues/3216)
Follow-up removal gate: [#3297](https://github.com/apache/maka/issues/3297)

Closes #3216

## Verification

Passed:

- Affected Runtime Host review suites — 79 passed (73 compatibility/kernel/coordinator tests plus 6 isolated candidate-lifecycle tests)
- `npm --workspace maka-agent test` — 316 passed
- `npm --workspace @maka/desktop test` — 971 passed
- `npm run build`
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `git grep -aI` checks for `ClientSurface`, `requireSurface`, `frame.surface`, and retired connection input plumbing

`npm --workspace @maka/runtime-host test` reports 1022/1023 locally because `owned Host exits promptly after its first connection closes` intermittently fails only in the full parallel run; its file passes 6/6 in isolation. The affected compatibility/kernel/coordinator suites pass 73/73.

`npm test` completed with two unrelated macOS/environment observations:

- Runtime reported one Unix PTY exit-race failure under the full parallel workspace run; isolated `npm --workspace @maka/runtime test` then passed 2963/2963.
- Eval reports 74/75 because `install-preflight.test` compares the same temporary path as `/private/var/...` versus `/var/...`. The identical failure reproduces on the unmodified `main` worktree.

## Rollout

New Hosts accept both legacy hello frames that include `surface` and new hello frames that omit it. New Clients temporarily keep the private fixed-value shim for older Hosts. The shim can be removed only when the minimum supported compatibility epoch is greater than 27; the removal change must bump the epoch so older Hosts take the structured incompatibility path. This gate is tracked in #3297.

## Review focus

This changes the bootstrap protocol and public Runtime Host client input types, so it needs independent human review. Please focus on mixed-version Client/Host behavior, the single private shim boundary, and the removal of all business-level surface plumbing without changing unrelated execution or UI surface concepts.

## AI use

Select exactly one:

Tool(s) and scope: Codex implemented the protocol and client cleanup, added and ran compatibility tests, and analyzed mixed-version rollout behavior.

## Checklist

- [ ] No


packages/runtime-host/src/client/connection.ts
```diff
@@ -1391,6 +1386,15 @@ interface ExchangeRuntimeHostHandshakeInput {
   readonly connectionResource?: RuntimeHostConnectionResource;
 }
 
+interface LegacySurfaceClientHello extends ClientHello {
+  /**
+   * Old Hosts require this field while decoding the bootstrap hello. Keep the
```
Astro-Han: [P2] Restate the shim's removal condition as something a future reader can actually check. The comment says to keep this until a surface-tolerant Host is the minimum supported version, but `RUNTIME_HOST_COMPATIBILITY_EPOCH` deliberately stays at 27, so a pre-PR Host that requires `surface` and a post-PR Host that tolerates its absence advertise the identical epoch and are permanently indistinguishable to a Client. The epoch is the only mixed-version discriminator on the wire, so as written the precondition can never be evaluated. That matters because of how the removal fails: the Host runs `decodeClientFrame` before `#admitHandshake`, so `requireSurface(undefined)` throws `invalidProtocolFrame('Invalid surface')` and the enclosing catch aborts the transport — which bypasses the `staleCompatibility` path that exists precisely so an old Host answers with an `incompatible` frame and the CLI can print its `RUNTIME_HOST_RESTART_REQUIRED` guidance. So the user with a resident old Host gets a bare socket abort instead of the one message that would tell them what to do. Holding the epoch is correct here; what is missing is a checkable gate. Word it as "removable once the minimum supported compatibility epoch is greater than 27", bump the epoch in the PR that removes the shim, and open the tracking issue so the decision is not a judgement call made from release notes.

packages/runtime-host/src/__tests__/handshake-compatibility.test.ts
```diff
@@ -33,7 +33,10 @@ test('rejects an epoch-23 Host before any domain command', async () => {
   let admittedRequest: RequestFrame | undefined;
   await withForgedHandshakePeer(
     async (transport, hostEpoch, rootId) => {
-      const hello = decodeClientFrame(await transport.read(2_000));
+      const rawHello = await transport.read(2_000);
+      assert.ok(rawHello && typeof rawHello === 'object');
+      assert.equal((rawHello as Record<string, unknown>).surface, 'desktop');
```
Astro-Han, line 37: [P3] Give the shim its own test rather than leaving its only guard inside a test about something else. This assertion is the single place in the repository that checks the Client still emits `surface` on the wire, and it lives in `test('rejects an epoch-23 Host before any domain command')`, whose stated subject is epoch rejection. It also survives only incidentally: `encodeProtocolMessage` is `JSON.stringify` and `writeClientFrame` receives a variable rather than an object literal, so TypeScript's excess-property check never fires on the extra field. Two independent cleanups therefore break it silently — retiring the epoch-23 case, or rewriting the encoder as a whitelist mirroring `decodeClientFrame`. Either one leaves every new Client unable to connect to every pre-PR Host, with CI green. A test named for the shim costs three lines and makes the guard intentional.

packages/cli/src/runtime-host-cli-context.ts
```diff
@@ -63,9 +62,9 @@ interface RuntimeHostCliContextDeps {
 export async function connectRuntimeHostCli(
   input: {
     readonly rootPath: string;
-    readonly surface: ClientSurface;
     readonly profileId?: string;
     readonly clientDataRoot?: string;
+    readonly interactiveSsh?: boolean;
```
Astro-Han, line 67: [P3] Cover `interactiveSsh`, since it turned a required discriminator into an optional flag. `surface` was mandatory, so every entry point had to state its identity and `'tui'` structurally implied interactive SSH; omitting `interactiveSsh` is now type-legal and silently yields `'batch'`. Both current TUI callers pass it and all four non-test call sites are converted, so nothing is broken today — the cost is that the next interactive entry point can forget it and the failure surfaces as an opaque SSH auth error when a key needs a passphrase, because batch mode cannot prompt. There was no CLI-level test for `sshInteraction` on `main` either, so this is a pre-existing gap the refactor makes marginally easier to fall into; two assertions in `runtime-host-cli-context.test.ts` — `'inherit'` with the flag and a TTY, `'batch'` without it — close it.

packages/runtime-host/src/__tests__/client-capability-coordinator.test.ts
```diff
@@ -138,7 +138,7 @@ describe('Host Client Capability coordinator', () => {
           },
         ],
       },
-      { ...connectionContext('connection-a'), surface: 'capability-provider' },
+      { ...connectionContext('connection-a') },
```
Astro-Han: [P3] Drop the spread and the stale name now that the override is gone. `{ ...connectionContext('connection-a'), surface: 'capability-provider' }` became `{ ...connectionContext('connection-a') }`, which copies a fresh object for no reason, and the local at line 202 is still called `providerContext` though nothing now makes it a provider. Harmless at runtime, but it leaves the next reader believing capability-provider registration is context-scoped when the real gate is `principal` — which is the opposite of what this PR just established. Pass `connectionContext('connection-a')` directly and rename the local.

packages/runtime-host/src/__tests__/host-kernel.test.ts
```diff
@@ -1786,6 +1759,37 @@ describe('non-serving Runtime Host kernel', () => {
     });
   });
 
+  test('accepts a Client hello without a surface identity', async () => {
```
Astro-Han: [P3] Pin both directions here while you are adding the kernel case. Every raw hello in this file had `surface` stripped and this new test adds the omitted-field case, so after this PR no kernel-level test sends a hello that still carries `surface` — the old-Client-to-new-Host direction is covered only by the decoder unit test in `protocol.test.ts`. The risk is genuinely low, since the kernel routes hello through that decoder and the whitelist rebuild is directly tested, but the two directions are the whole compatibility story of this change and one extra raw hello with `surface: 'tui'` in this same test would assert them together at the level where the connection is actually admitted.

Astro-Han (commented):
Thanks — this is a clean retirement of genuinely dead plumbing, and the part I most wanted to check holds up. Reviewed exact head `5c05c1b33d360a4a7c54f9a1b39404e9c7a7174c`; two independent passes, one on deletion safety and one on the correctness of what remains.

`ClientSurface` really did guarantee nothing. It was written into `ConnectionContext` by `host-kernel.ts` and `execution-composition.ts` and read by no handler, coordinator, log or authority check — connection authority has always come from the listener, never from `surface` — and it was never read at any point inside the epoch-27 window either. Its one real behavioural consumer was Client-local, the `surface === 'tui'` → `sshInteraction: 'inherit'` mapping, and `interactiveSsh` re-expresses that faithfully with all four non-test `connectRuntimeHostCli` callers converted. Zero tests were deleted — the diff contains no file additions or removals at all, and the test-case count goes up by three: two decoder cases and one kernel-level accept. The 87 non-production edits are all the same mechanical shape; I diffed the connect helpers across the UDS, execution-host and desktop suites against each other and found no site updated differently from its peers and no behaviour smuggled into the churn. The mixed-version matrix works in both directions, and holding epoch 27 is right for this PR because the Host change is a pure loosening.

One correction to how the PR is framed: the handshake itself is not retired. `exchangeRuntimeHostHandshake` still writes hello, reads the host frame, and validates epoch, protocol, rootId and compositionId before returning a connection — "the Client is ready" is defined at exactly the same point as before. What was retired is the surface identity plumbing. Worth saying because "retire client surface handshake plumbing" reads like the sequencing point moved, and reviewing it on that assumption would look for a pre-connection window that does not exist.

One P2 and four P3s inline, all about the durability of the compatibility shim and about test residue — none about the deletion being wrong. One thing I am deliberately not filing as a finding: by this PR's own criterion, `ConnectionContext.principal` is also write-only — two writers, zero reads, and `connection-session.ts` explicitly `Omit`s it. Unlike `surface` it is authority-derived and plausibly reserved for future authorization, and removing it is outside this PR's intent; I mention it only so the minimal-final-state claim stays honest.

Review disclosure: this review was prepared with Claude Code, which ran two parallel passes over the diff at this head — one auditing what the deletion removed against `main`, one auditing the correctness and consistency of what remains — and I checked each surviving finding against the source before keeping it. The reference searches and test-count comparisons described above were executed; the mixed-version failure scenarios were not, and are labelled inline. The human contributor reviewed this before posting.

packages/runtime-host/src/client/connection.ts
Sun-GLiang: Updated the private shim comment to use the observable gate: the minimum supported compatibility epoch must be greater than 27, and the removal change must bump the epoch so old Hosts reach the structured incompatibility path. Opened #3297 to track that removal gate.

packages/runtime-host/src/__tests__/handshake-compatibility.test.ts
Sun-GLiang: Moved the raw `surface: "desktop"` assertion into a dedicated `emits the legacy desktop surface shim in the raw Client hello` test. The epoch-23 rejection test now covers only epoch rejection.

packages/cli/src/runtime-host-cli-context.ts
Sun-GLiang: Added CLI-level coverage with TTY streams: explicit `interactiveSsh: true` forwards `inherit`, while omission forwards `batch`.

packages/runtime-host/src/__tests__/client-capability-coordinator.test.ts
Sun-GLiang: Removed the redundant spread and renamed the reused local from `providerContext` to `trustedContext`.

packages/runtime-host/src/__tests__/host-kernel.test.ts
Sun-GLiang: Expanded the kernel admission test to cover both directions: a hello without `surface` and a legacy hello with `surface: "tui"` are both accepted.

Astro-Han (commented):
Re-reviewed at `0dcfe7aa`. **All five findings from my previous round are fixed, and fixed correctly.** No new findings.

I isolated the author's own work from the merged `main` by comparing every file in the PR blob-by-blob against `5c05c1b3`: exactly five files differ, and they are precisely the five my findings named. Nothing else in the 90-file list moved.

Disposition, in the order I raised them:

- **P2, `connection.ts` shim comment** — fixed as asked, and better than asked. The removal condition is now checkable ("minimum supported compatibility epoch is greater than 27") rather than a judgement call, it explicitly requires the removal change to bump the epoch so old Hosts take the structured incompatibility path instead of a bare transport abort, and it links a tracking issue. I confirmed `#3297` exists, is open, and is scoped to exactly this — `chore(runtime-host): retire the legacy surface hello shim after epoch 27`.
- **P3, shim test** — fixed. The new `emits the legacy desktop surface shim in the raw Client hello` asserts `rawHello.surface === 'desktop'` on the **raw frame**, before `decodeClientFrame`. That is the right level: it survives the encoder being rewritten as a whitelist, which was the silent-failure mode I was worried about, and it no longer depends on the epoch-23 case continuing to exist.
- **P3, `interactiveSsh` coverage** — fixed. The new CLI test drives both directions against a real TTY and asserts `deepEqual(sshInteractions, ['inherit', 'batch'])`, with the `isTTY` descriptors properly saved and restored in `t.after`. That closes the gap that existed on `main` too, not just the one this refactor widened.
- **P3, spread and stale name** — fixed. `connectionContext('connection-a')` is passed directly and the local is now `trustedContext`, which names the actual gate.
- **P3, both hello directions at the kernel** — fixed. The kernel test now loops over a hello with and one without `surface`, each on its own transport with `try`/`finally` cleanup, and asserts both are accepted. The two directions are asserted together at the level where the connection is actually admitted, which is what I was asking for.

Nothing further from me on this change.

Reviewed with Claude Opus as an analysis assistant. Every disposition above was verified by reading the source at this head and diffing it against the head I previously reviewed; I did not run the suite.

Not marking this approved only because my standing authorization in this review campaign is comment-only; nothing here blocks it from my side.

Astro-Han (approved):
Approving at `0dcfe7aa`. All five findings from my earlier round are resolved — the shim's removal condition is now a checkable epoch gate with a tracking issue, the shim has its own test asserting `surface` on the raw wire frame, `interactiveSsh` is covered in both directions, and the two test cleanups landed. My detailed disposition is in the previous review; nothing is outstanding from my side.

Issue #3269, opened 2026-08-20
release: add the ASF source candidate workflow


M4n5ter:
<details open>
<summary>English</summary>

Part of #2974 — G2: source release mechanics.

## Outcome

Provide a reviewed source-only Apache Incubator release-candidate workflow that
produces, verifies, and documents the exact bytes used for a vote without
placing a Release Manager's private PGP key in CI.

## Exit criteria

- [x] Build `apache-maka-<version>-incubating-src.tar.gz` from an immutable Git
  commit with one matching top-level directory.
- [x] Produce and verify SHA-512 for the exact archive.
- [x] Provide a local detached-signature command and verify signatures against
  a reviewed `KEYS` file in an isolated keyring.
- [x] Reject missing release documents, unsafe paths, generated/private paths,
  and version mismatches.
- [x] Demonstrate that the same commit and version produce identical bytes.
- [x] Extract the exact candidate in CI and install, audit, build, type-check,
  run release checks, and test from that extracted source.
- [x] Document candidate tags, local signing, `KEYS`, Apache `dist/dev` staging,
  independent verification, podling/IPMC votes, and immutable RC handling.
- [x] Run the manual workflow successfully for a reviewed commit.

## Existing work

- #3278 was merged into `main` as `84fbe05d`. It implements the
  builder/verifier, focused tests, manual GitHub workflow, and Release Manager
  runbook.
- #3332 fixed the stale Runtime Host test input exposed by clean-archive
  validation.
- Manual validation run [32376142358](https://github.com/apache/maka/actions/runs/32376142358)
  completed successfully from reviewed `main` commit
  `98bdd6cf8add2900a3db9cc27a2cb623f1148a5e`. Every clean-archive verification
  step passed and the verified unsigned candidate artifact was uploaded.

## Out of scope

- Desktop installers, npm packages, bundled Git, and convenience binaries.
- Performing authenticated PGP signing or Apache distribution writes in CI.
- Deciding whether provenance or legal documents are release-ready.

## References

- [Incubator release management](https://incubator.apache.org/guides/releasemanagement.html)
- [Incubator distribution guidelines](https://incubator.apache.org/guides/distribution.html)
- [ASF release policy](https://www.apache.org/legal/release-policy.html)

## Ownership

M4n5ter coordinates this gate and owns the current implementation. Link the
reviewed implementation PR here once it is opened; release approval remains
with the PPMC, mentors, and IPMC.

</details>

<details>
<summary>简体中文</summary>

[...]

Issue #3276, opened 2026-08-20
release: publish ASF-compliant Desktop convenience artifacts


M4n5ter:
<details open>
<summary>English</summary>

Part of #2974 — G9: Desktop convenience artifacts.

## Outcome

Publish the supported Maka Desktop artifacts as ASF-compliant convenience
binaries built from the exact IPMC-approved Apache Maka source release.

## Exit criteria

- [ ] Define the first-release platform and artifact matrix, including the
  intended macOS arm64 and Windows x64 installers/archives.
- [ ] Build every Desktop artifact from the approved source archive and bind it
  to the same version, commit, and release identity.
- [ ] Remove bundled GPLv2 Git and all other Category X bytes from every
  convenience artifact; relying on a separately installed tool must follow the
  applicable ASF optional-dependency guidance.
- [ ] Audit each platform artifact's dependency closure and provide
  artifact-specific `LICENSE`, `NOTICE`, `DISCLAIMER-WIP`, and required
  prominent Category B labels/notices.
- [ ] Produce detached PGP signatures and SHA-512 hashes for every downloadable
  artifact, in addition to platform code signing where applicable.
- [ ] Through #3414, provision the ASF Apple Developer Program identity and
  complete macOS Developer ID signing, notarization, stapling, and
  browser-download acceptance with that identity.
- [ ] Through #3414, provision ASF ssl.com eSigner access and sign Windows
  release-candidate artifacts; otherwise obtain an explicit mentor/IPMC
  determination for any proposed unsigned artifact before the vote.
- [ ] Display the full Incubator disclaimer and source-release link on the
  GitHub release/download surface.
- [ ] Keep release candidates marked as prereleases and unadvertised; publish
  convenience artifacts only after the source release is approved.
- [ ] Independently verify installation, launch, upgrade/uninstall, and the
  representative supported functionality on each target platform.

## Existing work

- #3414 is the external prerequisite for ASF Apple Developer Program and ssl.com
  eSigner identities. G9 cannot close until Infra provisions them and the pipeline
  verifies the resulting platform signatures.
- `.github/workflows/release-desktop.yml` builds the current product artifacts,
  but it does not implement the ASF source-release binding, PGP/SHA-512 matrix,
  or Category X exclusion required here.
- #3222 partially unifies Desktop/CLI product release identity, but does not
  remove bundled Git or complete this issue's ASF compliance criteria.
- #3146 and its associated work concern duplicate Desktop dependencies and may
[...]

Issue #3274, opened 2026-08-20
release: resolve post-incubation GitHub and npm distribution status


M4n5ter:
<details open>
<summary>English</summary>

Part of #2974 — G7: existing distribution remediation.

## Outcome

Obtain and execute a mentor/PPMC-approved disposition for releases published
after incubation began and for the bundled-Git binary distribution path before
the first ASF release is announced.

## Exit criteria

- [ ] Inventory GitHub and npm artifacts published after Maka entered
  incubation, including `v0.1.11` and `cli-v0.1.0-beta.1`.
- [ ] Record which artifacts lack an Incubator disclaimer, an explicit non-ASF
  designation, or the expected release documents.
- [ ] Obtain a decision on `dev@maka.apache.org` with mentors about whether each
  artifact should be relabeled, deprecated, superseded, or otherwise remediated.
- [ ] Apply the approved GitHub/npm metadata changes without replacing artifact
  bytes or presenting them as ASF releases.
- [ ] Resolve or explicitly separate the Desktop path that bundles GPLv2 Git
  through Dugite before proposing any ASF convenience binary.
- [ ] Document the distinction between the official source release and the new
  npm/Desktop convenience artifacts built from it.

## Existing work

- #3220 ensures future CLI packages carry `DISCLAIMER-WIP`, but the published
  beta predates that change.
- #1510, #3002, and #3222 concern product distribution and do not resolve this
  policy decision.

## Out of scope

- Deleting or replacing existing releases without a mentor-approved decision.
- Completing Desktop/CLI product release features.
- Building the new compliant npm/Desktop artifacts tracked by G8 and G9.

## References

- [Incubator distribution guidelines](https://incubator.apache.org/guides/distribution.html)
- [ASF third-party license policy](https://www.apache.org/legal/resolved.html)

## Ownership

Leave this issue unassigned until the PPMC and mentors identify a coordinator.

</details>

<details>
<summary>简体中文</summary>

#2974 的一部分——G7：已有分发补救。

## 目标结果

在第一次 ASF release 公告前，就进入孵化器后发布的版本和 bundled Git 二进制分发路径，
取得 mentors/PPMC 批准的处理结论并落实。

## 完成条件

- [ ] 盘点 Maka 进入孵化器后发布的 GitHub 和 npm artifacts，包括 `v0.1.11` 和
  `cli-v0.1.0-beta.1`。
[...]

PR #3278, opened 2026-08-20
feat(release): add ASF source candidate workflow

M4n5ter:
<details open>
<summary>English</summary>

## Summary

- Build Apache Maka Incubator source candidates from an immutable Git commit while isolating every Git subprocess from inherited repository, configuration, template, and attribute controls; archive permissions remain independent of ambient `tar.umask`.
- Verify archive boundaries, required release documents, SHA-512, and detached signatures against a reviewed `KEYS` file without allowing ambient tar or gzip options to alter tool behavior.
- Before signing, require Release Managers to reproduce the canonical uncompressed Git tar payload on trusted hardware. The downloaded workflow archive remains the exact artifact that is signed, so platform-specific gzip encodings cannot change source identity or block cross-platform signing.
- Bind signing to a complete PGP fingerprint; require the actual signing key or subkey to be RSA with at least 2048 bits; create SHA-512 signatures; and accept only SHA-256, SHA-384, or SHA-512 signatures while failing closed on expired, revoked, bad, weak-digest, or missing signatures.
- Give ASF source mechanics a dedicated required CI lane, independent from npm and Desktop convenience-artifact gates, and route it only from inputs that it actually verifies.
- Document candidate tags, local signing, Apache `dist/dev` staging, every binding voter's verification duty, complete podling/IPMC vote thresholds, and immutable RC handling.

Refs #3269
Part of #2974

## Verification

- `npm run check:asf-source` — 9/9 passed, including hostile `GIT_DIR`/`GIT_TEMPLATE_DIR`/Git config/GZIP isolation, GNU `TAR_OPTIONS` bypass rejection, equivalent tar payloads with different gzip bytes, strong-digest enforcement, RSA signing-subkey acceptance, Ed25519/RSA-1024 rejection, and isolated-`KEYS` verification.
- `npm run check:release` — 32/32 passed after separating the ASF source gate.
- `node --test --test-concurrency=1 scripts/ci-test-plan.test.mjs` — 18/18 passed.
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `actionlint` v1.7.7 on the changed CI and candidate workflows.

- An earlier `npm test` run completed the full build and most suites, but remained non-green on two known macOS environment cases: Eval expected `/var/...` instead of resolved `/private/var/...`, and one Runtime `node-pty` lifecycle test timed out after 10 seconds. Neither failing area is changed by this PR.

## Remaining validation

- [ ] Obtain independent human review of the release policy and Release Manager runbook.
- [ ] Run **Prepare ASF source candidate** successfully from a reviewed commit on `main`.

## AI use

Select exactly one:

Tool(s) and scope: Codex prepared and reviewed the source-candidate automation, trust-boundary tests, workflows, and Release Manager runbook under human direction.

## Checklist

- [x] Lint, format, build, and affected suites pass locally

- [ ] No

</details>

<details>
<summary>简体中文</summary>

## 概要

- 从不可变 Git commit 构建 Apache Maka Incubator 源码候选包，同时隔离每个 Git 子进程继承的仓库、配置、模板与 attributes 控制；归档权限仍不受本机 `tar.umask` 影响。
- 验证归档边界、必需发版文件、SHA-512，以及基于已审查 `KEYS` 文件的 detached signature，并阻止本机 tar 或 gzip 选项改变工具行为。
- 签名前要求 Release Manager 在受信硬件上重建并逐字节匹配规范的未压缩 Git tar payload；最终签署的仍是 workflow 下载到的原始归档，因此不同平台的 gzip 编码不会改变源码身份或阻断跨平台签名。
- 将签名绑定到完整 PGP 指纹，要求实际签名 key 或 subkey 至少为 2048-bit RSA，生成 SHA-512 签名，并只接受 SHA-256、SHA-384 或 SHA-512；对过期、撤销、错误、弱 digest 或缺失签名 fail closed。
- 为 ASF 源码发版机制建立独立 required CI lane，与 npm 和 Desktop convenience artifact gate 解耦，并只由它实际验证的输入触发。
- 记录 candidate tag、本地签名、Apache `dist/dev` 暂存、每位 binding voter 的验包义务、完整 podling/IPMC 投票门槛及不可变 RC 处理流程。

关联 #3269
#2974 的一部分

## 验证

- `npm run check:asf-source`——9/9 通过，覆盖恶意 `GIT_DIR`/`GIT_TEMPLATE_DIR`/Git config/GZIP 隔离、GNU `TAR_OPTIONS` 绕过拒绝、不同 gzip 字节下的等价 tar payload、强 digest 强制、RSA signing subkey 正向路径、Ed25519/RSA-1024 拒绝及隔离 `KEYS` 验签。
- `npm run check:release`——ASF source gate 拆分后 32/32 通过。
- `node --test --test-concurrency=1 scripts/ci-test-plan.test.mjs`——18/18 通过。
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- 使用 `actionlint` v1.7.7 检查变更的 CI 与候选包 workflow。

- 较早的一次 `npm test` 完成全量构建及绝大多数 suite，但仍有两个已知 macOS 环境问题：Eval 预期 `/var/...`，实际得到 `/private/var/...`；另有一个 Runtime `node-pty` 生命周期测试在 10 秒后超时。这两个失败区域都未被本 PR 修改。

## 剩余验证

- [ ] 对发版政策和 Release Manager runbook 完成独立人工审查。
- [ ] 从 `main` 上经过审查的 commit 成功运行 **Prepare ASF source candidate**。

## AI 使用

二选一：

- [ ] 没有生成式工具做出实质性贡献
- [x] 生成式工具做出了实质性贡献

工具及范围：Codex 在人工指导下准备并审查了源码候选包自动化、信任边界测试、workflow 和 Release Manager runbook。

## 检查清单

- [x] 测试覆盖此变更，并且缺少此变更时会失败
- [x] lint、format、build 和受影响的测试套件均在本地通过

此 PR 是否导致行为变化？

- [x] 是——已在上述概要中说明
- [ ] 否

</details>


.github/workflows/asf-source-candidate.yml
```diff
@@ -0,0 +1,130 @@
...
+
+      - name: Build and type-check the extracted source
+        working-directory: candidate-source
+        run: |
+          npm run build
+          npm run typecheck
+
+      - name: Run ASF source checks from the extracted source
+        working-directory: candidate-source
+        run: npm run check:asf-source
```
Astro-Han, line 105: [P2] Validate the generated attribution files that ship inside the tarball. The archive carries `packages/cli/THIRD_PARTY_NOTICES.txt` and `apps/desktop/src/renderer/public/THIRD_PARTY_LICENSES.txt`, both generated from `package-lock.json`, but nothing in this job regenerates or compares them: the steps are `lint`, `format:check`, `build`, `typecheck`, `check:asf-source`, and `npm test` — and root `test` is `build:test` plus `run-workspace-tests-parallel.mjs`, which never invokes the notice generators. `check:release`, which does, is not invoked by any workflow in `.github/workflows/`. So a candidate can ship third-party attribution that does not match its own dependency set while passing every gate — wrong on precisely the axis the IPMC reviews. Confirmed by reading the script bodies and grepping every workflow at this head. Add a `working-directory: candidate-source` step running `npm run check:third-party-notices && npm run check:cli-third-party-notices` (or the whole `check:release`) between the build and the upload.

.github/ASF_SOURCE_RELEASE.md
```diff
@@ -0,0 +1,217 @@
...
+to the downloaded candidate after decompression. The original downloaded
+candidate is then signed with SHA-512 after resolving the selected secret key to
+the exact full fingerprint. Comparing the canonical payload avoids treating a
+platform's gzip encoding as source identity. A workflow-produced digest alone
+is not an independent trust check and is insufficient for signing.
+
+Export the matching public key for the podling `KEYS` file when needed:
+
+```sh
+gpg --armor --export <full-pgp-fingerprint> > KEYS
```
Astro-Han: [P2] Fix this command before a Release Manager runs it — it truncates the podling KEYS file. `gpg --armor --export <fpr> > KEYS` overwrites, which directly contradicts the warning four lines below telling the RM to merge reviewed public-key blocks rather than replace the published file. An RM following the copy-pasteable block literally wipes every other committer's public key out of `dist/*/KEYS`, breaking signature verification for previously published releases — the highest blast radius of anything in this runbook, and it costs one character to fix. Confirmed by reading; `>` truncation is not in dispute. Use `>>` against a checked-out copy, and follow ASF convention by prefixing each block with `gpg --list-sigs <fpr>` output.

.github/workflows/asf-source-candidate.yml
```diff
@@ -0,0 +1,130 @@
...
+        run: |
+          if [[ "$RELEASE_REF" != "refs/heads/main" ]]; then
+            echo "ASF source candidates must be dispatched from main; found $RELEASE_REF" >&2
+            exit 1
+          fi
+
+      - name: Check out the candidate commit
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ github.sha }}
```
Astro-Han, line 39: [P2] Name the commit the candidate is built from, or state the abort condition in the runbook. `ref: ${{ github.sha }}` and `--revision "$GITHUB_SHA"` resolve at dispatch time, so if anything merges to `main` between "CI is green for the exact commit" (runbook prerequisite 2) and the dispatch, this produces a fully valid, fully verified candidate **for a different commit** — no error, no warning, and the only signal is the `Commit:` line in the step summary. I read this as deliberate rather than an oversight: `asf-source-workflow-policy.test.mjs:11` asserts `doesNotMatch(workflow, /RELEASE_SHA/)`, and the head commit is titled "remove unsupported mutation defenses". So this is a contract decision, and it needs an owner rather than a patch. Confirmed by reading code at this head. Either restore a commit input, or add an explicit runbook step — abort and re-dispatch if the reported commit is not the commit CI validated — and put the reason for the removal in the PR body, because right now it reads as unexplained.

.github/workflows/asf-source-candidate.yml
```diff
@@ -0,0 +1,130 @@
...
+            --strip-components=1 \
+            -C candidate-source
+
+      - name: Install from the committed lockfile
+        working-directory: candidate-source
+        run: npm ci
+
+      - name: Audit production dependencies
+        working-directory: candidate-source
+        run: npm audit --omit=dev --audit-level=moderate
```
Astro-Han, line 82: [P2] Take the release path off the live advisory feed, or stop downgrading the check that already exists. `npm audit --omit=dev --audit-level=moderate` queries the registry at run time, so the same commit produces a candidate today and hard-fails tomorrow for reasons unrelated to the source bytes, and a registry outage blocks a release vote outright. It also duplicates `.github/workflows/dependency-audit.yml`, which runs the byte-identical command daily, on `main` pushes, and on lockfile PRs — and additionally runs `npm audit signatures --omit=dev`, which this copy omits. Against "extend the closest existing seam", this is the one step in the PR that parallels an existing path rather than extending it. Confirmed by reading both workflow files at this head. Drop the step and make "`Dependency audit` is green for this commit" a runbook prerequisite, or keep it and add `npm audit signatures --omit=dev` so it is at least not weaker than the job it duplicates.

scripts/asf-source-release.mjs
```diff
@@ -0,0 +1,717 @@
...
+  return hash.digest('hex');
+}
+
+function validateArchiveContents(archivePath, identity) {
+  for (const requiredFile of requiredReleaseDocuments) {
+    const contents = execTar(archivePath, ['-xOzf', `${identity.rootDirectory}/${requiredFile}`], {
+      encoding: 'utf8',
+      maxBuffer: maxCommandBuffer,
+    });
+    if (!contents.trim()) throw new Error(`${requiredFile} is empty`);
```
Astro-Han, line 446: [P2] Plan for NOTICE to be checked, not just non-empty. `validateArchiveContents` asserts real content only for `DISCLAIMER-WIP` (that it contains "Apache Maka" and "incubation"); `LICENSE` and `NOTICE` pass on any non-blank text. The repository's current NOTICE lacks both the ASF copyright line and the "developed at The Apache Software Foundation" statement, so this tooling reports a "verified" candidate carrying the single most common cause of an IPMC -1. I am filing this as a process decision rather than a defect, because `DISCLAIMER-WIP` explicitly lists "NOTICE has not yet been reworked into ASF form" and runbook prerequisite 3 delegates LICENSE/NOTICE readiness to the PPMC and mentors — that is a legitimate WIP posture, but the PPMC should choose it knowingly rather than meet it mid-vote. Reproduced: the check passes against the current NOTICE. Once NOTICE is reworked, extend this function to assert the ASF attribution string the same way it already asserts the disclaimer text.

.github/workflows/asf-source-candidate.yml
```diff
@@ -0,0 +1,130 @@
+name: Prepare ASF source candidate
+
+on:
+  workflow_dispatch:
```
Astro-Han, line 4: [P2] Merge this expecting a first-run shakeout, and say so. GitHub exposes `workflow_dispatch` only for workflows already on the default branch — `gh api .../workflows/asf-source-candidate.yml/runs` returns 404 and the workflow is absent from the repository's workflow list — and the `Require main` guard hard-fails any non-`refs/heads/main` ref anyway. So everything from `Extract the exact candidate` through the upload is unexecuted at merge time: the `--strip-components=1` extraction, `npm ci` from the extracted tree, the artifact path globs, and the step summary. The `create` script itself is well covered — I ran its suite and built a real candidate — but the workflow plumbing around it is not, and the PR's own "Remaining validation" section acknowledges this. Reproduced by API query. Nothing to change in the code; state in the PR body that a failed first dispatch is a workflow bug rather than a release blocker, so the first RM is not left guessing.

scripts/asf-source-release.mjs
```diff
@@ -0,0 +1,717 @@
...
+    console.log(`Reproduced commit ${result.commit}`);
+    console.log(`Created ${result.signaturePath}`);
+    return;
+  }
+  throw new Error('Usage: asf-source-release.mjs <create|verify|sign> [options]');
+}
+
+if (
+  process.argv[1] &&
+  realpathSync(fileURLToPath(import.meta.url)) === realpathSync(process.argv[1])
```
Astro-Han, line 711: [P3] Guard the main-module detection so importing this file cannot crash. `realpathSync(process.argv[1])` runs unguarded at module scope, so any embedding context where `argv[1]` is not an existing path throws before a single export is reachable. Reproduced by execution: importing the module with a non-path argument threw `ENOENT: no such file or directory, lstat '<arg>'` from this line. `node --test` and normal CLI use are unaffected, which is why it has not surfaced. Compare against `resolve(process.argv[1])` first, or wrap the `realpathSync` pair in a try/catch.

scripts/asf-source-release.mjs
```diff
@@ -0,0 +1,717 @@
...
+  'EXPSIG',
+  'FAILURE',
+  'KEYEXPIRED',
+  'KEYREVOKED',
+  'NODATA',
+  'NO_PUBKEY',
+  'REVKEYSIG',
+]);
+const allowedGpgHashAlgorithms = new Set([8, 9, 10]);
+const rsaPublicKeyAlgorithms = new Set([1, 2, 3]);
```
Astro-Han, line 43: [P3] Confirm the intended Release Manager's ASF-registered key is RSA before this is relied on. `rsaPublicKeyAlgorithms = new Set([1, 2, 3])` restricts signing to RSA at 2048 bits or more and rejects EdDSA (algorithm 22); the restriction is documented in the runbook and covered by a passing test, so it is a deliberate contract decision rather than an oversight. Worth surfacing anyway: ASF guidance recommends RSA ≥4096 but does not forbid EdDSA, and some ASF projects sign with it — so if the RM's registered key is Ed25519 they simply cannot sign, and that is much better discovered now than during the vote. Confirmed by reading code at this head plus the suite's own Ed25519-rejection case.

Astro-Han (commented):
Reviewed at `ac36d4fc`. **No P0 or P1.** The core mechanics here are unusually solid, and I verified the load-bearing ones by building the candidate rather than by reading: in a throwaway worktree I ran `create`, `verify`, `reproduce`, and `check:asf-source` (9/9), then diffed the extracted tarball against a clean checkout of the same commit.

What that established, so it does not have to be re-litigated during the vote:

- **The archive is deterministic.** uid/gid 0, uname/gname `root`, a single mtime from the commit, modes forced by `-c tar.umask=0002` regardless of ambient umask, and `gzip -n -9` so nothing leaks into the gzip header. The longest tracked path splits cleanly into ustar `prefix`+`name`, so no per-entry pax headers appear; git's `pax_global_header` is in the payload but is not listed by `tar -t` and so does not trip `validateArchiveEntries`.
- **The contents are exactly the committed tree minus the three `export-ignore` paths** — `diff -rq` differs only by `.claude/`, `.maka-shots/`, and `maka-proposal-zh-review.txt`. No binaries beyond five PNGs (README hero images and desktop icons), no symlinks, no gitlinks, no `node_modules`, no `.git`.
- **The `forbiddenSegments` backstop is live, not decorative.** Building from the pre-PR base commit, whose `.gitattributes` lacks the `export-ignore` lines, fails with `Forbidden archive entry: …/.claude/`.
- **The signature covers the published bytes.** `signSourceCandidate` signs `archivePath`, `verifySha512File` hashes the same file, and `publishSourceCandidate` hardlinks it — same inode throughout.
- **The secret surface is genuinely clean.** Zero `secrets.*` references; `workflow_dispatch` only, so no fork-reachable path; `permissions: contents: read`; `persist-credentials: false`; all three third-party actions pinned to full commit SHAs; and every `${{ }}` reaching a `run:` goes through a job-level `env:` and is dereferenced as a quoted shell variable. No private key ever enters CI — signing is local by design.
- **The GPG status parsing is correct**: `VALIDSIG` field indices, rejection of `EXPKEYSIG`/`REVKEYSIG`/`BADSIG`/`NO_PUBKEY`/`NODATA` before the GOODSIG count, a hash allowlist of exactly SHA-256/384/512, and subkey fingerprint resolution. Verification runs in a `mkdtemp` homedir seeded only from the reviewed KEYS file, so an unlisted signer yields `NO_PUBKEY`.
- **The `ci-test-plan.mjs` edit is not a coverage regression.** Dropping five entries from `CLI_PACKAGE_FILES` looked like one; running `planTests` on each shows all five still resolve to `full: true` through the unknown-root-file fallback.

On the seam question: this extends rather than parallels. No existing script produces a checksum file or touches GPG, the ASF source tarball is a different artifact with a different authority from the npm convenience tarball and the DMG, and `asf-source-workflow-policy.test.mjs` mirrors the existing `release-cli-workflow-policy.test.mjs` convention. Splitting `check:asf-source` out of `check:release` is right — the authoritative artifact should not be gated on Desktop and npm convenience checks. The one genuine overlap is the `npm audit` step, inline.

Two of the findings below are **process decisions rather than defects** and need a named human owner rather than a code change: the candidate always builds `main`'s tip at dispatch (deliberate — a test asserts the absence of a SHA input), and NOTICE is not yet in ASF form (deliberate — `DISCLAIMER-WIP` says so). Both are legitimate WIP postures; the risk is that the PPMC discovers them mid-vote rather than deciding them beforehand.

Reviewed with Claude Opus as an analysis assistant. Findings marked "reproduced by execution" were built and run in an isolated worktree that has been removed; the root checkout was not modified and no repository-wide tests were run. Evidence grade is stated per finding.

.github/ASF_SOURCE_RELEASE.md
```diff
@@ -0,0 +1,217 @@
...
+
+The unsigned workflow handoff is identified by its exact source commit, not by
+an RC number. Assign the RC identity only when creating the immutable candidate
+tag and staging directory for the bytes selected by the Release Manager.
+
+## Prerequisites
+
+1. The intended version is committed to the root `package.json` on `main`.
+2. Normal CI is green for the exact commit.
+3. The PPMC and mentors have confirmed that provenance, `LICENSE`, `NOTICE`, and
```
Astro-Han, line 33: [P3] Have the PPMC address the bundled-Git GPLv2 relationship deliberately. The tarball ships `apps/desktop/resources/licenses/git/{LICENSE.txt,NOTICE.txt,SOURCE_OFFER.txt}` — a GPLv2 text and a written offer for Git source. This is **not** a Category-X problem for the source release: I grepped the full extracted tree for the GPL header and only Git's own `LICENSE.txt` matches, so no GPL code ships here. But the root `LICENSE` covers only `trycua/cua` (MIT) and Geist (OFL) and says nothing about the Git arrangement — and the Geist entry itself scopes to material "redistributed in binary artifacts", which is not what this archive contains. A reviewer who opens `SOURCE_OFFER.txt` in an Apache source release will ask whether the product depends on GPL code; better to pre-answer that in `DISCLAIMER-WIP` or in this prerequisite than to answer it mid-vote. Confirmed by reading the files inside the built archive.

.github/ASF_SOURCE_RELEASE.md
```diff
@@ -0,0 +1,217 @@
...
+
+```sh
+npm run release:asf:sign -- \
+  --artifact <candidate-dir>/apache-maka-<version>-incubating-src.tar.gz \
+  --key <full-pgp-fingerprint> \
+  --revision v<version>-incubating-rc<rc>
+```
+
+The signing command first validates the downloaded SHA-512 and archive. It then
+rebuilds the canonical uncompressed Git archive from the specified revision on
```
Astro-Han, line 99: [P3] Make a Git-version skew diagnosable in one step. `reproduceSourceCandidate` byte-compares the uncompressed tar produced by the RM's local `git archive` against the runner's, and the runbook carefully explains gzip encoding differences across platforms while saying nothing about Git's own tar writer. This fails closed — it blocks signing rather than producing a wrong artifact, which is the right direction — but the failure reads as `Candidate source payload does not match <sha> rebuilt on this machine`, which sounds like a provenance incident and could stall a release while the RM hunts for tampering that never happened. Inference, not reproduced: I reproduced successfully with `git 2.50.1 (Apple Git-155)` on macOS against a macOS-built archive and did not cross-test against the `ubuntu-24.04` runner's Git. Emit `git --version` into the existing step summary and name the version in the mismatch error.

.github/workflows/asf-source-candidate.yml
M4n5ter: Thanks — we evaluated this against the supported handoff and signing path and are not treating it as a code finding. The unsigned artifact name and workflow summary record the actual `github.sha`; before signing, the RM creates the RC tag and `release:asf:sign` reproduces the candidate from that tag and byte-compares the canonical source payload with the downloaded artifact. A moved `main` tip therefore cannot silently become a signed RC for the intended revision: reproduction fails unless the RM deliberately adopts the recorded SHA. The remaining consequence is a recoverable extra dispatch, so adding a second revision input and authority is not justified.

.github/workflows/asf-source-candidate.yml
M4n5ter: Thanks. We are keeping this check. It complements rather than replaces `.github/workflows/dependency-audit.yml`; that workflow remains the continuous registry-signature check, while this command audits the production dependency set installed from the exact extracted candidate. This job already requires registry access for `npm ci`, so removing `npm audit` would not make candidate validation offline or immune to registry outages, but it would remove a useful candidate-boundary check.

scripts/asf-source-release.mjs
M4n5ter: Thanks — this is a valid release-readiness concern, but it belongs to G3 rather than the G2 release-mechanics verifier. `DISCLAIMER-WIP` already states that the current `NOTICE` is not yet in ASF form, and the runbook requires PPMC/mentor confirmation of legal-file readiness. #3270 owns the artifact-specific `LICENSE`/`NOTICE` decision and rewrite. We will keep that legal wording authority there instead of hard-coding provisional text in this mechanical source check.

.github/workflows/asf-source-candidate.yml
M4n5ter: Thanks. The GitHub first-run limitation is real, and the required shakeout is already tracked in this PR's Remaining validation section and in #3269's exit criteria: the workflow must complete successfully from a reviewed `main` commit before G2 is complete. We are not adding wording that an initial failure is "not a release blocker" because any real failure must be fixed before the release gate can be considered ready.

scripts/asf-source-release.mjs
M4n5ter: Thanks — the reproduction is valid only for unsupported embedding where `process.argv[1]` names a nonexistent entry point. The supported CLI invocation and the `node:test` paths both provide an existing script path, and this module is not a public package API. Adding alternate entry-point detection for that hypothetical contract would increase complexity without protecting a supported release path, so we are leaving the fail-closed behavior unchanged.

scripts/asf-source-release.mjs
M4n5ter: Thanks. This is a deliberate signing policy rather than an implementation defect. The runbook already makes compatible key selection a pre-release prerequisite, and ASF signing guidance recommends new RSA keys of at least 4096 bits while allowing existing RSA keys of at least 2048 bits. Before closing G2 we will confirm that the intended Release Manager's ASF-registered key is compatible; if it is not, we will choose or create a compatible key, or broaden algorithm support based on that concrete requirement rather than adding speculative policy now.

.github/ASF_SOURCE_RELEASE.md
M4n5ter: Thanks — the review confirms that no GPL-covered Git code is present in the source candidate; the archive contains license, notice, and source-offer resources for the separately distributed bundled Git. The exact source legal-file contents are owned by #3270 (G3), the existing distribution relationship by #3274 (G7), and removal of bundled Git from the Desktop binary by #3276 (G9). Expanding G2's mechanical verifier would duplicate those policy authorities, so we are keeping this work in the existing gates.

.github/ASF_SOURCE_RELEASE.md
M4n5ter: Thanks. We do not have evidence of a Git-version compatibility failure on the supported release path, and the current implementation already fails closed if reproduction differs. Recording a version string would not establish that Git caused a mismatch and would add diagnostic state without resolving one. The required cross-platform G2 shakeout will give us concrete evidence; if it exposes a version-dependent failure, we will add a targeted diagnostic with the corresponding compatibility decision.

.github/workflows/asf-source-candidate.yml
M4n5ter: Fixed in `251728b0a`. The source-candidate workflow now runs both generated attribution inventory checks from `candidate-source`, after extracting and installing the exact archived payload. We kept these as source-specific checks instead of invoking the shared npm/Desktop release gate, and added a workflow-policy assertion so this candidate-boundary behavior cannot disappear silently.

.github/ASF_SOURCE_RELEASE.md
M4n5ter: Fixed in `251728b0a`. The runbook now starts from a reviewed copy of the current podling `KEYS` file and appends the selected key with `(gpg --list-sigs ... && gpg --armor --export ...) >> KEYS`, then verifies the combined file. It also explicitly requires retaining every key previously used to sign an Apache Maka release, matching the ASF key-history model.

Astro-Han (commented):
Re-reviewed at `251728b`. Two of my P2s from `ac36d4f` are resolved and I verified both:

- **Attribution inventories** — the lane now runs `check:third-party-notices` and `check:cli-third-party-notices` against the extracted source, after a full `npm ci`, with `asf-source-workflow-policy.test.mjs` asserting the workflow still contains them. Both scripts exist in the archived `package.json` and need only devDependencies, which the preceding step installs. Correct fix, and pinning it with a policy assertion is the right instinct.
- **KEYS truncation** — `> KEYS` is now `>> KEYS` and the runbook starts from a reviewed copy of the published file. `(gpg --list-sigs <fpr> && gpg --armor --export <fpr>) >> KEYS` produces the conventional ASF KEYS block and cannot destroy existing keys.

One follow-up on the inventory fix below, and my remaining threads from the previous head carry over unchanged — the dispatch-time `$GITHUB_SHA` resolution, `npm audit` against the live advisory feed on the release path, `NOTICE` being checked for existence rather than content, the `workflow_dispatch` first-run shakeout, and the four P3s. None of them are new, so I have not re-posted them.

Still COMMENT rather than approve: the open items are all P2, none of them blocks a release by itself, but this is the lane that decides what an IPMC vote sees, and I would rather the RM path be settled before it becomes the default.

AI disclosure: this review was produced with Claude Code (Opus 5). For this pass I read the delta commit, the workflow and runbook at `251728b`, `asf-source-release.mjs`'s archive boundary set, `.gitattributes`, and the notice-generation scripts and their call sites; I did not run the workflow or the tests. Per AGENTS.md this is not independent human review.

.github/workflows/asf-source-candidate.yml
```diff
@@ -0,0 +1,136 @@
...
+
+      - name: Install from the committed lockfile
+        working-directory: candidate-source
+        run: npm ci
+
+      - name: Audit production dependencies
+        working-directory: candidate-source
+        run: npm audit --omit=dev --audit-level=moderate
+
+      - name: Verify source attribution inventories
```
Astro-Han, line 84: [P2] Add `check:windows-cargo-notices` here — this step now verifies two of the three attribution inventories the tarball ships. `apps/desktop/resources/licenses/cargo/THIRD_PARTY_NOTICES.txt` is a tracked 237 KB file, and nothing in `forbiddenSegments` or `.gitattributes` export-ignores it, so it goes into every source candidate; its inputs `experiments/windows-sandbox/launcher/{Cargo.toml,Cargo.lock}` ship alongside it, so it is verifiable from the archive alone. But `check:windows-cargo-notices` is invoked only from `scripts/package-windows-x64.mjs`, which this lane never reaches — so a stale cargo inventory reaches a vote unchallenged, which is the same failure mode you just closed for the npm inventories and the one an IPMC reviewer is most likely to open the tarball and check. Confirmed by reading the script's output path, the archive boundary set and every call site at `251728b`; not executed. `generate-windows-cargo-notices.mjs` shells out to `process.env.CARGO ?? 'cargo'`, which GitHub's ubuntu image provides, so this should be one more line in the same step. Extend the `asf-source-workflow-policy.test.mjs` assertion you just added to cover the third script, so the lane cannot silently lose an inventory again.

M4n5ter: Fixed in `5d52e8fdf`. The extracted-source attribution step now runs `npm run check:windows-cargo-notices` alongside the two npm inventory checks, so the archived Cargo notice is validated against the archived manifest and lockfile before upload. The existing workflow-policy test now asserts all three inventory commands. I also ran all three checks, `check:asf-source` (9/9), the focused policy test, lint, format, actionlint, and `git diff --check` successfully.

Astro-Han (commented):
The cargo-inventory P2 is fixed at `5d52e8f`, including the part I care about most — extending the `asf-source-workflow-policy.test.mjs` assertion so the lane cannot lose an inventory again without a test noticing.

I checked the one thing that could have made my own suggestion wrong: this job is `runs-on: ubuntu-24.04` with no Rust setup step, so `check:windows-cargo-notices` depends on the runner image's preinstalled toolchain. That is already this repository's established assumption — `windows-sandbox-w0.yml` runs `cargo build --locked` and `cargo test --locked` on `windows-2025` with no setup step either — and `experiments/windows-sandbox/launcher/{Cargo.toml,Cargo.lock}` are neither in `forbiddenSegments` nor export-ignored, so they are present in the extracted candidate. It should hold. Confirmed by reading the workflows and the archive boundary set; not executed, so the first real run is still the proof.

Worth distinguishing from my open `npm audit` thread, since they look similar: `cargo metadata --locked` resolves from the committed lockfile and is deterministic, so this does not add the kind of time-varying external dependency that one is about.

My remaining threads from the previous heads carry over unchanged — the dispatch-time `$GITHUB_SHA` resolution, `npm audit` against the live advisory feed on the release path, `NOTICE` being checked for existence rather than content, the `workflow_dispatch` first-run shakeout, and the four P3s. Nothing new from me on this head, so I have not re-posted them.

Still COMMENT rather than approve: everything open is P2 or below and none of it blocks a release on its own, but this lane decides what an IPMC vote sees, and I would rather the Release Manager path be settled before it becomes the default.

AI disclosure: this review was produced with Claude Code (Opus 5). I diffed `251728b..5d52e8f` and read the workflow, the sandbox workflow and the archive boundary code at this head; I did not run the workflow or the tests. Per AGENTS.md this is not independent human review.

Astro-Han (approved):
Approving at `5d52e8f`, and correcting my own bar while I do it.

I withheld approval on the previous two heads on the grounds that this lane decides what an IPMC vote sees, so I wanted the Release Manager path settled first. That was a standard I invented for this PR rather than the one I apply everywhere else: P0 and P1 block, P2 should be fixed but does not. Everything still open here is P2 or P3, and holding a release-mechanics PR to a stricter rule than the product PRs I approved today is not defensible. My mistake, and it cost you two extra rounds.

What actually landed across those rounds: the attribution inventories are verified inside the extracted candidate for all three sources now — npm, CLI and cargo — each pinned by an assertion in `asf-source-workflow-policy.test.mjs` so the lane cannot quietly lose one; and the KEYS runbook no longer truncates the published file. All of my review threads are resolved.

The four P2s I am approving over, so they are on the record rather than forgotten: the candidate commit resolves at dispatch time rather than being named or asserted; `npm audit` puts a time-varying external feed on the release path; `NOTICE` is checked for existence rather than content; and `workflow_dispatch` only becomes available once this is on the default branch, so the first real run is the first syntax check. None of them blocks a release on its own, and the last one cannot be resolved before merging by construction.

Note on CI: `package` is failing at "Verify automatic update end to end", which is the Windows autoupdate lane. This PR's only workflow change adds an `asf_source` gate and one step to the `test` job and does not touch `package`; that lane is the one #3241 and #3265 are currently fixing. `test` and `audit` are green.

AI disclosure: this approval was produced with Claude Code (Opus 5). Across the three heads I read the workflow, the runbook, `asf-source-release.mjs`'s archive boundary, `.gitattributes`, the notice-generation scripts and their call sites, and confirmed the failing job's step. I have not run the workflow. Per AGENTS.md this does not count as the independent human review this PR still needs.

PR #3280, opened 2026-08-20
fix(desktop): recover Runtime Host reads after reconnect

sunheyi6:
## Summary

- replay transcript range reads against the current Desktop replica after Runtime Host replacement
- make `app:info` a reconnectable read so project metadata queries survive connection loss
- surface reconnect failures, candidate exit details, and top-level Runtime Host lifecycle errors
- add regression coverage for a stale transcript generation request completing on the replacement replica

## Root cause and behavior change

The Runtime Host database was healthy. The visible task-read failures came from an incomplete recovery path after the Host connection was lost: some reads remained bound to the dead connection, while transcript navigation rejected requests solely because the replacement replica had a newer generation.

After this change, eligible reads wait for the replacement connection, and durable transcript sequence requests continue against the current replica. The added diagnostics also make future Host exits and reconnect failures observable instead of leaving only downstream `connection_lost` errors.

## Validation

- Desktop Runtime Host observer/manager/reconnecting IPC tests: 56 passed
- Runtime Host execution candidate tests: 3 passed
- Runtime Host reconnecting connection tests: 9 passed
- `git diff --check`

Known local-only result: the owned-candidate suite passed 5/6; its prompt-exit timing test exceeded the 2-second election deadline on this Windows environment and was terminated by the test harness. The changed launcher code only observes the existing exit event and does not alter shutdown control flow.

---

## 中文说明

本次问题不是 `runtime.sqlite` 损坏，而是 Runtime Host 断线后的恢复链路不完整：部分读取仍然发往旧连接，聊天记录请求也会因为后台换代而直接报错。

修改后，可重试的读取会等待新连接；旧代聊天记录请求会交给当前副本实际执行。同时补充了重连失败、子进程退出码/信号和顶层生命周期异常日志，方便后续定位真正的退出原因。


apps/desktop/src/main/runtime-host-session-observer.ts
```diff
@@ -1302,9 +1302,10 @@ export class RuntimeHostSessionObserver {
     if (targetId !== undefined && consumer.target.id !== targetId) {
       throw new Error('Desktop transcript consumer belongs to another renderer');
     }
-    if (consumer.generation !== request.generation || replica.generation !== request.generation) {
-      throw new Error('Desktop transcript generation changed');
-    }
+    // Durable transcript sequence identities belong to the Session, not to a
```
Astro-Han: [P2] Narrow the guard to the invariant you mean instead of deleting it, and make that invariant testable. The removed check covered two different things — `replica.generation !== request.generation` (the replica was replaced) and `consumer.generation !== request.generation` (the renderer is asking on behalf of a superseded consumer view) — and both are now unenforced, which leaves `DesktopTranscriptRangeRequest.generation` declared in `apps/desktop/src/preload/transcript-contract.ts`, still sent by the renderer, and read by nothing. The comment states the load-bearing claim, that durable sequence identity is Session-scoped rather than replica-scoped, but no code asserts it and the new regression only asserts `rangeLoads === 1` for a stale generation, i.e. that the read is attempted. If a replacement replica ever renumbers, the renderer's `anchorSequence` silently addresses a different slice and the user sees wrong content instead of an error. Either keep a check on what actually must match across replacement — the Session and `hostEpoch`, which `openTranscript` still returns — and drop `generation` from the range request contract, or keep the field and have the test seed the replacement replica with the same durable sequences and assert the returned range matches.

packages/runtime-host/src/client/launcher.ts
```diff
@@ -130,6 +132,17 @@ function spawnedPid(child: ReturnType<typeof spawn>): Promise<{ pid: number }> {
   });
 }
 
+function observeCandidateExit(child: ReturnType<typeof spawn>): void {
```
Astro-Han: [P3] Route candidate-exit diagnostics through a caller-provided sink rather than the console. These three new sites plus the one in `candidate-entry.ts` are nearly all of the unstructured logging in `packages/runtime-host/src`, and this one fires for every candidate exit in every embedder, including the intentional clean shutdowns that the `code === 0` branch exists to recognise — so a library decides for the Desktop app and the CLI alike what lands on stderr. The reconnect path in this same PR shows the alternative: `onReconnectError` is a callback the embedder supplies. Also worth deciding deliberately: `console.error('[runtime-host] lifecycle failed:', error)` in `candidate-entry.ts` prints a whole error object, whose message can carry paths and spawn arguments, which is the kind of thing the repository asks us to keep out of logs. An `onExit` hook next to the existing `hooks.onWon` seam keeps the observability and leaves the sink to the owner.

Astro-Han (commented):
Thanks — separating "the database is healthy" from "the recovery path never rebound these reads" is the right diagnosis, and making `app:info` reconnectable is clearly correct. Reviewed exact head `2f89bcde025db8127dd89f1916e5bd47bd620b7b`.

My concern is with how the transcript fix is expressed rather than with the intent. The guard was removed rather than narrowed, which leaves the invariant the new comment asserts unowned by any code, and the new test asserts that the stale request runs, not that it returns the right slice. Details inline, along with a note on the new logging.

No CI checks are currently reported for this head.

Review disclosure: this review was prepared with Claude Code, which read the diff at this head, traced `generation` through the observer and the preload transcript contract, and checked the surrounding reset path. The reproductions are code reading, not execution. The human contributor reviewed it before posting.

apps/desktop/src/preload/preload.ts
```diff
@@ -1665,7 +1665,8 @@ const makaBridge = {
       ): Promise<void> =>
         ipcRenderer.invoke(operation, consumerScope, {
           consumerId,
-          generation,
+          sessionId,
+          hostEpoch: opened.hostEpoch,
```
likun666661: [P2] Advance the range Host epoch when a reset installs a new Host. This closure captures `opened.hostEpoch` forever. After `open()` has completed, a replacement Host sends reset batches with a new `batch.hostEpoch`; the renderer store accepts that reset, but every subsequent `loadBefore`/`loadAround` still sends the old epoch. The new Main guard therefore rejects every retry with `Desktop transcript host epoch changed`, and the controller does not automatically reopen the handle. Track a mutable Host epoch alongside `generation` (update it from accepted reset batches and initialize it from `opened`), or explicitly recreate the handle on epoch change. Please add a regression where open completes on `host-1`, a reset arrives for `host-2`, the next range request uses `host-2`, and an already-dispatched `host-1` request still fails closed.

likun666661 (dismissed):
Requesting changes on the current head. The `app:info` reconnectable-read change and the diagnosis that durable transcript sequence is not a Desktop replica generation both look sound. There is one blocking recovery gap inline: the new Host-epoch fence is never advanced in the preload handle after `open()` returns, so a real Host replacement turns all later range reads into permanent failures until the whole handle is recreated. Please make the range identity follow reset batches (or explicitly reopen the handle) and cover a post-open Host-epoch change.

Scope note, non-blocking: the candidate-exit/reconnect logging is useful but orthogonal to the read repair and substantially widens this PR. If it stays here, please keep reconnect errors generalized/redacted and isolate embedder-provided exit sinks so diagnostics cannot throw through the child-process event listener.

Review performed against `fc7a9890`.

sunheyi6 (commented):
done

sunheyi6:
CI fix: the `chore: keep workflow files at the branch base` commit left `.github/workflows` stale relative to the CI planning contract tests in `scripts/ci-test-plan.test.mjs` (identical to upstream main): the planner emits an `asf_source` surface, the `Test CI planner` / `Check Windows test inventory` steps must run before dependency setup, the sandbox lane needs its nightly `schedule`, and every checkout must set `persist-credentials: false`.

Restored the workflow files to the PR base versions (98bdd6cf8) — verified byte-identical, so this PR still has zero `.github/workflows` diff. All 21 CI planner contract tests pass locally at the pushed head.

The CI run is waiting on the fork-PR workflow approval gate (triggered by an external contributor). Could a maintainer approve `CI` / `Windows sandbox W0 evidence` runs for `97c45fc44`? cc @Astro-Han

Astro-Han:
@sunheyi6 CI runned~

Astro-Han:

- `packages/runtime-host/src/client/launcher.ts`

```
# resolve, then
git push --force-with-lease
```

---

Astro-Han (commented):
Re-reviewed at exact head `950affb4a499a85991fb46cdbd9cd8f9dbe2cb15`, focused on whether the previously-raised blocking finding is genuinely closed.

**It is closed, and closed at the right layer.** No `[P0]`–`[P2]` in the current scope. One optional test suggestion below.

The earlier concern was that a stale handle could keep reading after the Runtime Host was replaced. The fix draws the distinction that actually matters: durable transcript sequence identity belongs to the **session and Host epoch**, not to a Desktop replica generation.

- `runtime-host-session-observer.ts:1345-1358` admits a request only when the session matches and `replica.hostEpoch === request.hostEpoch`. A genuine Host replacement therefore fails closed on in-flight requests, while recovery that installs a replacement replica under the same host epoch still completes the read — which is the behavior the PR exists to restore.
- `preload/preload.ts:1790-1814` advances `{generation, hostEpoch}` and the target scope atomically inside the reset batch, and the handle range path re-reads the current identity on each call rather than closing over a stale epoch.

I looked specifically for a path where a stale range could be silently served by a new Host and did not find one; the rejection is an explicit throw, not a fallback.

### `[P3]` Suggested bridge-level regression

The added tests are pure identity checks plus a synthetic Main-side observer — they don't exercise the real renderer/preload ↔ Host bridge. That doesn't undermine the reasoning above, since the closure is visible in the code, but a bridge-level case would pin the behavior against future refactors: after `reset(host-2)`, the next range on the same handle should go to host-2, and an already-dispatched host-1 range should be rejected.

### CI

Not green at this head, and not for a code reason: exact-head run `32625500665` was cancelled during checkout, so `package` and `test` are both `completed/cancelled` and every later step was skipped. Correctness is fine from my side; the checks still need a re-run before this can be treated as gated.

Validation ran at this head: build plus preload, focused suite 40/40, owned-candidate 11/11, Biome and diff-check clean.

Astro-Han:
CI is failing on this head, and the cause is small and mechanical.

`Check ASF source headers` fails with two files missing the license header:

```
apps/desktop/src/main/__tests__/transcript-identity.test.ts
apps/desktop/src/preload/transcript-identity.ts
```

Fix:

```sh
npm run write:asf-headers
```

Worth knowing: this step runs **before** the dependency install, so no test has executed on this head — there is no test signal here yet, green or red. Once the headers land, the real suite will run for the first time and may surface something else.

<details>
<summary>中文</summary>

这个 head 上 CI 挂了，原因很小：`Check ASF source headers` 报两个文件缺许可证头——`apps/desktop/src/main/__tests__/transcript-identity.test.ts` 和 `apps/desktop/src/preload/transcript-identity.ts`。跑 `npm run write:asf-headers` 即可。

顺带说一句：这一步在装依赖**之前**，所以这个 head 上一个测试都没跑过，现在既没有绿也没有红。补完头之后真正的测试才会第一次运行，可能还会有别的问题。
</details>

sunheyi6:
CI fix for the new head `625bcc84`:

Fixed the `Check ASF source headers` failure by adding the ASF license header (canonical, `npm run write:asf-headers` equivalent) to:

- `apps/desktop/src/main/__tests__/transcript-identity.test.ts`
- `apps/desktop/src/preload/transcript-identity.ts`

Verified both files now classify as `canonical` under `scripts/asf-license-headers.mjs`, so the header audit should pass and the real suite will finally execute on this head.

The new head's workflow runs (`32705807410`, `32705807431`) are again sitting on the fork-PR approval gate. Could a maintainer approve them? cc @Astro-Han

Astro-Han:
I went through the three blocking points one by one at the current head `625bcc84c`. All three look genuinely addressed to me, not just marked resolved.

**1. Host epoch not advancing in `preload.ts` (P2 from @likun666661)** — fixed. `identity` is now a mutable reference, swapped by `adoptTranscriptIdentity()` when a reset batch arrives, and `range()` reads it at call time via `const currentIdentity = identity` rather than capturing it in a closure at `open()`. So the first `loadBefore`/`loadAround` after a Host replacement carries the new epoch.

**2. Deleted guard in the observer (my earlier P2)** — resolved via the first option I offered: `generation` is gone from the `DesktopTranscriptRangeRequest` contract, replaced by `sessionId` + `hostEpoch`, and both are actually asserted (`runtime-host-session-observer.ts:1345-1357`). No field is left declared-but-unread.

**3. Candidate exit logging (my earlier P3)** — an `onExit` hook was added as suggested. `packages/runtime-host` no longer writes to the console; the sink belongs to the embedder, and the Desktop side decides its own logging in `runtime-host-desktop-manager.ts`. The hook is wrapped in try/catch so a throwing diagnostic cannot propagate through the child-process listener. `candidate-entry.ts` now goes through `generalizedErrorMessage` as well.

Regression coverage: `transcript-identity.test.ts` covers the identity adoption logic, and `runtime-host-session-observer.test.ts` covers the main-side guard — including a stale-epoch request being rejected after a Host replacement (`assert.rejects` with `hostEpoch: 'host-1'`) and a read succeeding across a replica replacement. Together those two halves are exactly the scenario @likun666661 asked for.

I also trial-merged this into the latest main: main's only changes to these two files in the meantime added `updateQueueEntry`, which is unrelated to transcript identity. After merging, runtime-host typechecks cleanly and all 40 tests in those two files pass.

@likun666661 the P2 you raised has been fixed the way you described — could you take another look? Your CHANGES_REQUESTED is still bound to `fc7a9890` and is currently blocking this PR. CI is still running on the current head; I will go over it again once it is green.

<details>
<summary>简体中文</summary>

在当前 head `625bcc84c` 上把三条挡着的意见逐条核过了，我认为都已经真正解决，不是形式上关掉的。

**1. `preload.ts` 的 Host epoch 不推进（@likun666661 提的 P2）** — 已修。`identity` 现在是一个可变引用，`adoptTranscriptIdentity()` 在收到 reset batch 时换成新的身份；`range()` 里是 `const currentIdentity = identity` 在调用时读取，不再是 `open()` 时捕获的闭包。所以替换 Host 之后的第一次 `loadBefore`/`loadAround` 就会带上新的 epoch。

**2. observer 里的守卫被删掉（我之前提的 P2）** — 按我给的第一个方案改了：`generation` 已从 `DesktopTranscriptRangeRequest` 契约里去掉，改成 `sessionId` + `hostEpoch`，并且两者都有实际断言（`runtime-host-session-observer.ts:1345-1357`）。不再有"声明了但没人读"的字段。

**3. candidate 退出日志（我之前提的 P3）** — 按建议加了 `onExit` 钩子，`packages/runtime-host` 里不再有 console 输出，sink 交给 embedder；Desktop 侧在 `runtime-host-desktop-manager.ts` 自己决定怎么打。钩子外包了 try/catch，诊断抛错不会穿透子进程事件监听。`candidate-entry.ts` 那处也改成了 `generalizedErrorMessage`。

回归覆盖：`transcript-identity.test.ts` 覆盖身份采纳逻辑，`runtime-host-session-observer.test.ts` 覆盖 Main 侧守卫——包括替换 Host 后旧 epoch 请求被拒和跨 replica 替换的读继续成功。这两半合起来正是 @likun666661 要求的那个场景。

我把它试合并进最新 main 验证过：main 这期间只在这两个文件里加了 `updateQueueEntry`，和 transcript 身份互不相干；合并后 runtime-host 类型检查干净，上述两个测试文件 40 项全过。

@likun666661 你提的 P2 已经按你说的方式修好了，麻烦你重新看一眼——你那条 CHANGES_REQUESTED 还绑在 `fc7a9890` 上，现在挡着这个 PR。CI 在当前 head 上还在跑，绿了之后我这边也会重新过一遍。

</details>

Astro-Han:
@sunheyi6 there is nothing left blocking from my side on content — I verified all three points against the current head.

The only thing holding this up now is @likun666661's CHANGES_REQUESTED. It is still bound to `fc7a9890` from four days ago, and reviewDecision reads from it, which is why the PR stays BLOCKED. Could you reach out to likun666661 directly and ask for a fresh look at the current head? Dismissing someone else's review is not ours to do — it has to come from them.

I will approve once CI reaches terminal green on the current head.

<details>
<summary>简体中文</summary>

@sunheyi6 内容上我这边已经没有阻塞项了，三条意见我都在当前 head 上核过是真修好的。

现在唯一卡住的是 @likun666661 那条 CHANGES_REQUESTED——它还绑在 4 天前的 `fc7a9890`，而 reviewDecision 取的就是它，所以 PR 一直是 BLOCKED。麻烦你直接联系 likun666661 请他按当前 head 重新看一下；撤别人的 review 不是我们该做的事，得由他本人处理。

我这边等当前 head 的 CI 跑成终态绿之后会给 approve。

</details>

sunheyi6:
> @sunheyi6 内容上我这边已经没有阻塞项了，三条意见我都在当前 head 上核过是真修好的。
> 
> 现在唯一卡住的是 @likun666661 那条 CHANGES_REQUESTED——它还绑在 4 天前的 `fc7a9890`，而 reviewDecision 取的就是它，所以 PR 一直是 BLOCKED。麻烦你直接联系 likun666661 请他按当前 head 重新看一下；撤别人的 review 不是我们该做的事，得由他本人处理。
> 
> 我这边等当前 head 的 CI 跑成终态绿之后会给 approve。

ok

Astro-Han (commented):
CI on this head (`625bcc84`) failed. Three Desktop e2e tests in `apps/desktop/e2e/prompt-rail.spec.ts` are red — run [32705807431](https://github.com/apache/maka/actions/runs/32705807431):

- `the first click of a session lands on its prompt and holds` — the polled `landing()?.offset` stayed `undefined` for the full 10s
- `long transcripts keep a bounded mounted turn window` — expected `> 100`, received `15`
- `evicting a turn returns focus to the transcript`

This is the first run on this PR that reached the e2e step; the earlier ones stopped at the ASF header check, which you have since fixed. The same suite passes on `main` — run [32707461612](https://github.com/apache/maka/actions/runs/32707461612) at `03e0ebc25`, which already contains most of the commits between this branch's base and `main`. So the failures look specific to this change rather than to drift underneath it.

**A likely cause, from reading the diff — I could not run the suite locally, so please treat this as a lead rather than a diagnosis.**

The two sides of the generation check moved apart.

In `runtime-host-session-observer.ts` you deliberately stopped binding a range read to the consumer's generation: `isCurrent()` no longer requires `consumer.generation === request.generation`, and `#requireTranscriptConsumer` now admits a replacement replica as long as the session and host epoch match. That is the point of the PR, and the reasoning in the comment is sound.

But `preload.ts` still filters the delivered batches by generation, and it only adopts a new one on a reset:

```ts
const adopted = adoptTranscriptIdentity(identity, batch);
if (adopted !== identity) { identity = adopted; consumerScope = host; }
if (identity !== undefined && batch.generation === identity.generation) handler(batch);
```

`adoptTranscriptIdentity` returns `current` by reference whenever `current !== undefined && !batch.reset`, so a replica swap that keeps the same `hostEpoch` — exactly the case the observer change was written to allow — produces batches carrying the **new** generation while the preload still holds the **old** one. Main serves the read, the batches arrive, and every one of them is dropped before reaching `handler`.

That would leave range loads resolving successfully while nothing new mounts, which is what the three failures describe: a bounded window stuck at 15 turns, and a clicked prompt that never lands because its turn never arrives.

If that is right, the fix is to make the preload adopt on the same terms the observer now accepts — carry `hostEpoch` as the identity that gates delivery, and let the generation follow the batch rather than gate it. The new `transcript-identity.ts` is the natural place for it, and `adoptTranscriptIdentity` is already shaped to hold both fields.

**Separately, the other blocker has not moved.** The `CHANGES_REQUESTED` review is still bound to the old head `fc7a9890`, so it does not reflect the current code but still blocks. Only likun666661 can clear it — please reach out to them directly.

这个 head（`625bcc84`）上的 CI 挂了。`apps/desktop/e2e/prompt-rail.spec.ts` 里有三个 Desktop e2e 用例是红的——运行记录 [32705807431](https://github.com/apache/maka/actions/runs/32705807431)：

- `the first click of a session lands on its prompt and holds`——轮询的 `landing()?.offset` 在整整 10 秒内始终是 `undefined`
- `long transcripts keep a bounded mounted turn window`——期望 `> 100`，实际收到 `15`
- `evicting a turn returns focus to the transcript`

这是本 PR 第一次跑到 e2e 这一步；之前几次都停在 ASF header 检查上，那个问题你已经修好了。同一套用例在 `main` 上是通过的——运行记录 [32707461612](https://github.com/apache/maka/actions/runs/32707461612)，对应 `03e0ebc25`，它已经包含了本分支 base 到 `main` 之间的大部分提交。所以这几个失败看起来是本次改动特有的，而不是底下的漂移造成的。

**一个可能的原因，来自读 diff——我没能在本地跑这套用例，所以请把它当作线索而不是定论。**

generation 检查的两端分开了。

在 `runtime-host-session-observer.ts` 里你有意不再把范围读取绑定到 consumer 的 generation：`isCurrent()` 不再要求 `consumer.generation === request.generation`，`#requireTranscriptConsumer` 现在只要 session 和 host epoch 匹配就接纳替换后的 replica。这正是本 PR 的目的，注释里的推理也是成立的。

但 `preload.ts` 仍然按 generation 过滤投递的批次，并且只在 reset 时才采纳新的：

```ts
const adopted = adoptTranscriptIdentity(identity, batch);
if (adopted !== identity) { identity = adopted; consumerScope = host; }
if (identity !== undefined && batch.generation === identity.generation) handler(batch);
```

只要 `current !== undefined && !batch.reset`，`adoptTranscriptIdentity` 就按引用返回 `current`。因此，一次保持相同 `hostEpoch` 的 replica 替换——恰恰是 observer 那处改动想要允许的情形——会产生带**新** generation 的批次，而 preload 手里还是**旧**的那个。主进程照常服务这次读取，批次也确实到达了，但每一条都在抵达 `handler` 之前被丢掉。

这会导致范围加载正常返回、却没有任何新内容挂载，而这正是三个失败所描述的：窗口卡在 15 个 turn，点击的 prompt 因为它所属的 turn 从未到达而始终落不下去。

如果这个判断没错，修法是让 preload 采纳的条件与 observer 现在接受的条件一致——用 `hostEpoch` 作为把关投递的身份，让 generation 跟随批次而不是把关批次。新加的 `transcript-identity.ts` 正是放这个逻辑的地方，`adoptTranscriptIdentity` 的形状也已经能同时持有这两个字段。

**另外，另一个阻塞项没有变化。** 那条 `CHANGES_REQUESTED` 仍然绑在旧 head `fc7a9890` 上，因此它并不反映当前代码，却仍然构成阻塞。只有 likun666661 能撤销它，请直接联系他。

</details>

me2seeks:
Reproduction evidence for the reconnect recovery path — Desktop `dev@32a1db0` on Windows, same log signature as this PR's root cause.

**Toast:**
```
Surface: toast
Title: 读取任务失败
Description: 任务操作失败，请稍后重试。
```

**Environment:**
- Maka: 0.2.0 dev @ 32a1db0, Electron 43.4.1 / Chrome 150.0.7871.224 / Node 24.18.1
- OS: win32 10.0.26200 (x64), Locale zh-CN
- Workspace: ~\AppData\Roaming\Maka Dev\workspaces\default (dev)
- Captured: 2026-08-26T07:11:58.558Z, Main uptime 1717s

**Recent main-process logs (trimmed to the relevant chain — matches `Some reads remained bound to the dead connection, while transcript navigation rejected requests solely because the replacement replica had a newer generation`):**

```
[07:10:07.128Z] ERROR [runtime-host] projection refresh failed: RuntimeHostRequestInterruptedError: Runtime Host subscription.open was interrupted: the operation outcome is unknown; do not retry it automatically
  operation: 'subscription.open', mode: 'control', dispatch: 'dispatched', reason: 'connection_lost'
  [cause]: RuntimeHostTransportError: Timed out waiting for Runtime Host host.status response (code: read_timeout)
[07:10:07.196Z] ERROR ... subscription.open was interrupted: the operation was not dispatched (same cause, read_timeout)
# repeats every ~6-17s at 07:10:24, 07:10:30, 07:10:43 — 8× total before 07:11

[07:11:03.629Z] ERROR [runtime-host] MCP capability alignment failed: RuntimeHostRequestInterruptedError: Runtime Host client.capability.replace was interrupted: the operation was not dispatched
  operation: 'client.capability.replace', mode: 'command', dispatch: 'not_dispatched', reason: 'connection_lost'
  [cause]: RuntimeHostTransportError: Runtime Host transport read side ended (code: read_eof)
# repeats at 07:11:12, 07:11:18, 07:11:32, 07:11:55 — 5× read_eof

[07:11:55.875Z] ERROR Error occurred in handler for 'sessions:transcript:load-before': Error: Desktop transcript generation changed
  at #requireTranscriptConsumer (runtime-host-session-observer.ts:1352)
  at #runTranscriptRangeOperation ... at RuntimeHostSessionObserver.loadTranscriptBefore

Diagnostics: Runtime Host is reconnecting (Diagnostics unavailable), Recent local Runtime Host process exits: <none captured>
```

**Correlation:**
- `read_timeout` on `host.status` → `connection_lost` → 8× `subscription.open` failures (07:10:07-07:10:43) → matches this PR's `reads still bound to dead connection`
- `read_eof` on `client.capability.replace` (07:11:03-07:11:55) → transport break / flapping, same family as #3458/#3847
- `Desktop transcript generation changed` on `load-before` (07:11:55) → exact case this PR replays against the current replica (`replay transcript range reads against the current Desktop replica after Runtime Host replacement`)

This was observed while an Agent Graph with 7× `local-read` operators was running (`0/7 settled`, see #3878) and navigating transcript history — no `runtime.sqlite` corruption involved, reproducing the PR description on Windows.

<details><summary>中文复现说明 / Chinese</summary>

与本 PR 描述完全一致：非数据库损坏，而是断连后恢复链路不完整。07:10 起 `host.status` 超时导致 `subscription.open` 连续 8 次 `connection_lost`，随后 `client.capability.replace` 因 `read_eof` 连续失败，最后 `load-before` 触发 `generation changed` 抛到前端 toast。复现于 `dev@32a1db0` Win32 + 7 扇出 Graph 过程中，作此 PR 的 Win 复现证据。

</details>

me2seeks:
Follow-up cross-reference for the same diagnostic chain:

Relates to #3795 and #3847 — same `connection_lost` / `read_eof` flapping chain observed here.

- #3795 covers the `handler waits indefinitely` branch (composer `task-readiness` latched) that this log's 8× `subscription.open` timeouts would hit if user tried to `Send` during the `reconnecting` window.
- #3847 covers the 15s shared-budget for reconnectable reads that this log's 8× `subscription.open` + 5× `client.capability.replace` retries exercised — the pending-argument retention root proved there.

Parent tracking remains #3458 (OOM / heap verification under prolonged flapping). This comment keeps #3280's evidence linked to its sibling fixes.

me2seeks:
Follow-up: closing-stage evidence for the same Windows chain (sibling of #issuecomment-5421983866).

**New screenshot:** `ScreenShot_2026-08-26_153037_200.png` — after the `0/7` Graph stalled (previous `read_timeout`/`read_eof` storm), closing Desktop showed:

```
无法安全退出 Maka
本地 Runtime Host 未能安全停止，Maka 仍在运行。
退出已取消。请重试；如果问题持续存在，请查看诊断信息。
原因: Runtime Host has no current connection to quiesce
```

**Root:** `packages/runtime-host/src/client/reconnect-lifecycle.ts:199`

```ts
quiesce(): RuntimeHostReconnectQuiescence<T> {
  if (!current) throw new Error('Runtime Host has no current connection to quiesce');
}
```

At that moment `reconnect-lifecycle` had no `current` due to the preceding `host.status` `read_timeout` → `subscription.open` `connection_lost` (8×) + `client.capability.replace` `read_eof` (5×) storm — same diagnostic as the prior comment.

`apps/desktop/src/main/runtime-host-desktop-manager.ts:558-584` exit path:

```ts
const quiescence = lifecycle.quiesce();
diagnostics = await quiescence.current.client.queryHostDiagnostics();
result = await quiescence.current.client.prepareHostRetirement(mode);
```

throws → caught and wrapped as `DesktopLocalHostRetirementError` → rendered by `runtime-host-quit-copy.ts:67` (`buildRuntimeHostQuitFailureDialog`, zh `无法安全退出 Maka`), with `app-quit-coordinator.ts: handleBeforeQuit` correctly cancelling quit (`phase='running'` → `focusOrCreateWindow()`), hence "Maka is still running."

On next launch Desktop showed an empty state — expected transient window: the new candidate got a new `hostEpoch` while the old `transcriptConsumer.generation` was stale, so `runtime-host-session-observer.ts:1352` `Desktop transcript generation changed` makes the transcript appear empty until `listEpochs`/`getSnapshot` replay on the new replica completes. Not data loss; re-appears after `Diagnostics: Runtime Host is reconnecting` clears.

This confirms #3280's reconnect-replay scope must also cover the quit path's `quiesce`/`prepareHostRetirement` — same `current === undefined` race as the prior `load-before` failure. Related to #3795 (5s handler window) and #3847 (15s reconnectable-read budget), parent #3458.

<details><summary>中文说明 / Chinese</summary>

与上一条评论同一条链路的“关闭阶段”：Graph `0/7` 卡住时的 `read_timeout`/`read_eof` 已使 `reconnect-lifecycle` 失去 `current`，此时点关闭会走到 `lifecycle.quiesce()` 因 `!current` 抛 `has no current connection to quiesce`，被 `DesktopLocalHostRetirementError` 包后弹出对话框并取消退出（避免丢执行）。重启后空白是 `hostEpoch` 切换后 `generation changed` 的短暂重放窗口，非数据丢失。此证据补全 #3280 对 `quiesce` 路径的覆盖。

</details>

M4n5ter:
@likun666661, the branch is now updated to current `main` at exact head `6dc5767d9f8c5f5c57c86a025041dd63ea8fede6`.

The blocking Host-epoch issue from review `4983376181` is closed on this head: the preload handle adopts `{ generation, hostEpoch }` from accepted reset batches, and each range request uses the current Host epoch together with the Host Session ID. Requests already dispatched with an older Host epoch still fail closed. The old `CHANGES_REQUESTED` review remains active and blocks the pull request even though it targets `fc7a9890`.

Could you re-review the current head and dismiss or replace that old review if the fix satisfies the concern? The exact-head hosted checks are currently running.

---
_Posted by an automated review agent operated by @M4n5ter. This is not an
anything here is wrong._

_本条评论由 @M4n5ter 运行的自动化审查程序发出。它不构成 CONTRIBUTING.md

</details>

Astro-Han (commented):

Uses sessionId+hostEpoch for transcript range, allowing same-Host replica recovery while blocking cross-session reads; reconnectable IPC paths clean. Hosted `windows_recovery: SUCCESS`, `test/package: IN_PROGRESS`.

No P0-P3.

<details><summary>简体中文</summary>该头无阻断。</details>

---

M4n5ter (approved):

PR #3291, opened 2026-08-20
test(desktop): isolate side-chat slash dispatch

M4n5ter:
## Summary

- remove the unrelated steering submission from the `/side` slash-command E2E
- keep the scenario focused on the running-turn command-dispatch contract
- rely on the existing streaming-remount E2E for Desktop steering coverage

Fixes #2916.

## Root cause

The test chained two asynchronous submissions. Under CI load, the second slash-menu interaction could begin while the steering submission still owned the previous draft, producing `/compact explain/` instead of `/` and leaving no `/side` option to select. The same failure appeared on unrelated PRs #3242 and #3278.

The steering behavior is already covered by `streaming-remount.spec.ts`, so removing that duplicate prelude eliminates the race without adding another timing fence or product state solely for a test.

## Verification

- targeted `/side` E2E repeated 10 times: 10/10 passed
- `/side` dispatch and the independent live-turn steering E2E: 2/2 passed
- Biome check and `git diff --check` passed

PR #3294, opened 2026-08-20
test: fix the two intermittent CI lane failures at their causes

### LINKED ISSUE #3289: flaky(e2e): slash-command-menu specs time out intermittently on the Linux e2e lane
## Summary

The desktop Playwright e2e suite's slash-command-menu tests fail intermittently on the Linux `e2e` lane with `expect(locator).toBeVisible()` timing out at 10s (`element(s) not found`), and pass on every rerun. Observed at least four times across three unrelated PRs in the last three days, each time as the only failure in an otherwise green run.

## Evidence (all failures are `toBeVisible` 10s timeouts; every rerun passed)

| Date (UTC) | PR | Failing test | Run/job |
|---|---|---|---|
| 2026-08-18 ~16:27 | #3174 (Windows sandbox — unrelated code) | `slash-command-menu-compacts-the-active-session` | run 32157698387 attempt 1 |
| 2026-08-18 ~10:0x | #3174 | `slash-command-menu-compacts-the-active-session` | earlier attempt, rerun passed |
| 2026-08-19 ~11:03 | #3240 (release scripts — unrelated code) | `slash-command-menu-dispatch…-ring-it-into-a-running-turn` | run 32244987684, job 96044046442 |
| 2026-08-20 ~01:50 | #3265 (installer rollback — unrelated code) | `slash-command-menu-dispatch…-ring-it-into-a-running-turn` | run 32322199338, job 96286486651 |

Playwright artifacts (`test-failed-1.png`, `error-context.md`, `trace.zip`) are attached to each failing run under `e2e/test-results/slash-command-menu-*`.

## Why this matters more since the Apache migration

`gh run rerun` now requires admin rights on the repository, so contributors cannot re-run a flaked lane; the only recovery is pushing an empty commit, which cancels any in-flight lanes for the ref (concurrency group) and re-triggers every bot review. A single flake therefore costs a full CI cycle plus review noise.

## Suggested directions (for the owning team to judge)

- The two failing specs share a shape: open the slash-command menu, then assert a menu item/result is visible within 10s. Either the menu open animation/data load occasionally exceeds 10s on the shared runner, or the click lands before the menu is interactive. A `waitFor` on the menu container before the item assertion, or a raised per-assertion timeout for these two specs, would likely close the gap.
- Alternatively mark the two specs `retries: 1` in the Playwright project config if the timing source is judged environmental.

No repository code owned by the reporting PRs touches this suite; filing so the desktop e2e owners can pick it up with the traces above.

### LINKED ISSUE #3290: flaky(test): computer-use maka-cu-service teardown races temp-root removal (ENOTEMPTY)
## Summary

`packages/computer-use/dist/__tests__/maka-cu-service.test.js` fails intermittently on the Linux `test` lane (`Run affected standard workspace tests`) with a temp-directory cleanup race:

```
✖ /home/runner/work/maka/maka/packages/computer-use/dist/__tests__/maka-cu-service.test.js (14.7ms)
[Error: ENOTEMPTY: directory not empty, rmdir '/tmp/maka-test-computer-use-9VpvvT/maka-cu-service-test-fYcMJy']
  { errno: -39, code: 'ENOTEMPTY', syscall: 'rmdir', path: '…' }
[computer-use] failed with code 1
```

All 971 tests in the affected run otherwise passed (`fail 0` in the desktop workspace summary of the same run); the suite fails only in its cleanup.

## Evidence

- 2026-08-20 ~05:56 UTC, PR #3265 (installer rollback — does not touch `packages/computer-use` at all): run **32337361618**, job `test`, step "Run affected standard workspace tests".

## Analysis

`ENOTEMPTY` on `rmdir` during teardown is the same defect family as the Windows `EBUSY` cleanup failures documented in `docs/windows-support.md` ("Stores, owners, and leases must close deterministically before their temporary root is removed") — a child process or open handle created by the test is still writing into the temp root when the recursive removal runs, or the removal races a concurrently-created file. On Linux this surfaces as `ENOTEMPTY` instead of `EBUSY`.

Likely fixes, for the owning team to judge:
- await/settle whatever the service spawns (or its log/artifact writers) before the temp root is removed, mirroring the deterministic-close work done for the storage suites in Phase 1;
- or use `fs.rm(root, { recursive: true, force: true, maxRetries: N, retryDelay: … })` in the test teardown, which the release verifier scripts already standardize on — though retries paper over a still-running writer, so the deterministic settle is the better first choice.

## Friction note

Since the Apache migration `gh run rerun` requires admin rights, so a contributor hit by this flake can only recover by pushing an empty commit — which cancels in-flight lanes for the ref and re-triggers all review bots.

liugddx:
## Summary

Fixes the two intermittent CI lane failures tracked in #3289 and #3290, each at its cause. No assertion is weakened and no timeout is raised.

Since the Apache migration a single flaked lane costs a full CI cycle plus a re-review round (`gh run rerun` needs admin; the only contributor recovery is an empty-commit push that cancels in-flight lanes and re-triggers every bot), which is why these two are worth fixing at the source rather than retrying around.

## #3290 — computer-use `ENOTEMPTY` teardown race

`maka-cu-service.test.ts`'s `after()` disposed every service and immediately `rm -r`'d the shared work directory. `dispose()` is deliberately fire-and-forget: it SIGTERMs the child and schedules an **asynchronous** image-directory purge on child exit (or on the 3s shutdown-grace SIGKILL), plus one more purge when an in-flight `start()` settles. The recursive `rm` therefore raced concurrent purges — and a child still flushing its ndjson log — and failed with `ENOTEMPTY` (`maka-cu-service.ts` `dispose()`, the purge closure).

Fix (test-side only; the production fire-and-forget contract is intentional and untouched):
- `makeService` now records each per-service image directory.
- `after()` waits (bounded, 10s; grace is 3s) for every image directory to vanish — each one is deleted by its service's own purge, so this is a deterministic "children exited and purges finished" barrier. The wait is tolerant rather than asserting, because purge failure is explicitly permitted by the production contract ("the next spawn purges it again") and is not what this teardown tests.
- The final `rm(workDir)` gets `maxRetries: 10, retryDelay: 100` as a backstop for same-tree stragglers (e.g. a final log flush).

## #3289 — slash-command e2e `toBeVisible` 10s timeouts

dd4b2d0b4 (#3291) already fixed the `/side` spec by removing the racy steering chain and documented the underlying race: *after the composer remounts, `fill()` can land before the contentEditable is focused, so the draft never populates.*

The remaining flaky spec, `compacts the active session`, hits the same race one step later: right after the compact completes (composer clears/remounts), it did `fill('after compact')` + `Enter` — if the fill lands before focus returns, Enter submits nothing and `Fake backend received: after compact` never appears. That is exactly the pinned failure: run 32157698387 attempt 1 fails at **line 83** on `getByText('Fake backend received: after compact')`, 10s `element(s) not found`, after all preceding compact assertions passed.

Fix mirrors dd4b2d0b4 in the same file: `click()` + `pressSequentially('after compact')` + `expect.poll(textContent).toBe('after compact')` before pressing Enter — the dispatch only happens once the draft has demonstrably settled.

## Verification

- **L2 (local, Windows dev machine)**: `@maka/computer-use` builds (`tsc`) and the service test file runs its new teardown cleanly; pass/fail counts are byte-identical to unmodified `main` (4 pass / 13 fail, all pre-existing environmental `spawn EFTYPE` — Windows cannot spawn the `.cjs` mock directly; the Linux `test` lane is the authoritative runner for this suite). Desktop typecheck and biome pass. 
- **L3 (authoritative, CI)**: the Linux `test` and `e2e` lanes on this PR execute both changed suites for real. Local Playwright execution was attempted and is not possible on this machine (Electron `firstWindow` timeout at fixture level, before any test body, for all specs including unmodified ones).
- The flake is intermittent, so a single green lane proves non-regression, not the absence of the race; the determinism argument above is the actual fix rationale.

Closes #3289
Closes #3290

### DIFF
--- a/packages/computer-use/src/__tests__/maka-cu-service.test.ts
+++ b/packages/computer-use/src/__tests__/maka-cu-service.test.ts
@@ -113,6 +113,7 @@ process.stdin.on('data', function (chunk) {
 let workDir = '';
 let mockPath = '';
 const services: MakaCuService[] = [];
+const imageDirs: string[] = [];
 
 function delay(ms: number): Promise<void> {
   return new Promise((resolve) => setTimeout(resolve, ms));
@@ -197,6 +198,7 @@ function makeService(
   };
   const service = new MakaCuService(options);
   services.push(service);
+  imageDirs.push(imageDir);
   return { service, logPath, imageDir };
 }
 
@@ -215,7 +217,24 @@ after(async () => {
       // A disposed service disposing again is not a test failure.
     }
   }
-  if (workDir) await rm(workDir, { recursive: true, force: true });
+  // `dispose()` is deliberately fire-and-forget: it SIGTERMs the child and
+  // schedules an asynchronous image-directory purge on child exit (or on the
+  // shutdown-grace SIGKILL, at most `shutdownGraceMs` = 3s later). Deleting
+  // `workDir` while those purges — and a child still flushing its ndjson log —
+  // are in flight races `rm` against concurrent writers and fails with
+  // ENOTEMPTY (issue #3290). Every `makeService` image directory is purged by
+  // its `dispose()`, so waiting for them to vanish is a "children exited and
+  // purges finished" barrier. The wait is tolerant rather than asserting —
+  // purge failures are permitted by contract ("the next spawn purges it
+  // again") and are not what this teardown tests — and the retrying `rm`
+  // backstop then only absorbs stragglers, not an unbounded race.
+  const purgeDeadline = Date.now() + 10_000;
+  while (imageDirs.some((imageDir) => existsSync(imageDir)) && Date.now() < purgeDeadline) {
+    await delay(25);
+  }
+  if (workDir) {
+    await rm(workDir, { recursive: true, force: true, maxRetries: 10, retryDelay: 100 });
+  }
 });
 
 describe('maka-cu supervisor: what the child may put on stdout', () => {
--- a/apps/desktop/e2e/slash-command-menu.spec.ts
+++ b/apps/desktop/e2e/slash-command-menu.spec.ts
@@ -78,7 +78,14 @@ test('compacts the active session', async ({
   await expect.poll(() => composer.textContent()).toBe('');
   await expect(page.getByText('压缩失败')).toHaveCount(0);
 
-  await composer.fill('after compact');
+  // After the compact completes the composer clears and can remount. `fill()`
+  // can land before the contentEditable is focused again, so the draft never
+  // populates and Enter submits nothing — the flake in issue #3289. Type
+  // through the focused element and require the draft to have settled before
+  // dispatching, mirroring the running-turn spec below.
+  await composer.click();
+  await composer.pressSequentially('after compact');
+  await expect.poll(() => composer.textContent()).toBe('after compact');
   await composer.press('Enter');
   await expect(page.getByText('Fake backend received: after compact')).toBeVisible();
   await expect(page.getByText('Fake backend received: /compact')).toHaveCount(0);


hqhq1025 (approved):
No actionable findings.

Problem and root cause

- The computer-use test lane could fail during teardown because `MakaCuService.dispose()` deliberately completes child shutdown and image-directory cleanup asynchronously, while the suite immediately removed the shared work directory.
- The slash-command E2E could submit an empty draft after `/compact` because the composer is cleared/remounted and `fill()` could race its focus/state transition.

How this revision solves them

- The computer-use teardown now waits on the service-owned image directories to disappear, then performs a bounded recursive cleanup with retries.
- The E2E now clicks the remounted composer, types through the normal keyboard path, and verifies the observable draft text before submitting.

Assessment

- The problem definitions match the captured failures and the relevant production lifecycles.
- Both fixes follow first principles and Occam's razor: they synchronize on the state that must actually become true instead of increasing arbitrary sleeps, weakening assertions, or adding production compatibility paths.
- The current solution is appropriately scoped and optimal for these test-only races.
- Production code to delete: none; this PR does not change production code.
- Low-quality tests to delete or replace: none. The existing shutdown test still strictly verifies image-directory removal, so the more tolerant suite teardown does not become the sole coverage for that behavior.
- Deeper refactor required: no. The production ownership boundaries remain unchanged, and the fixes live at the test synchronization boundary where the races occur.

Verification

- CI `test`: passed at `9269c5fe7de3fc69c891129a77362a5518e23b06`.
- `maka-cu-service.test.js`: 17/17 passed, then the complete file passed 10/10 repeated runs.
- Electron `compacts the active session` E2E: 5/5 repeated runs passed.
- Desktop workspace build, Biome checks for both changed files, and `git diff --check`: passed.

Merge verdict

Ready to merge at the reviewed head.

Residual risk: intermittent failures cannot be disproved absolutely by a finite repetition count. Future CI should still be watched for a different lifecycle race, but this revision directly addresses the two observed failure mechanisms without masking them.

packages/computer-use/src/__tests__/maka-cu-service.test.ts
```diff
@@ -215,7 +217,24 @@ after(async () => {
...
+  // `workDir` while those purges — and a child still flushing its ndjson log —
+  // are in flight races `rm` against concurrent writers and fails with
+  // ENOTEMPTY (issue #3290). Every `makeService` image directory is purged by
+  // its `dispose()`, so waiting for them to vanish is a "children exited and
+  // purges finished" barrier. The wait is tolerant rather than asserting —
+  // purge failures are permitted by contract ("the next spawn purges it
+  // again") and are not what this teardown tests — and the retrying `rm`
+  // backstop then only absorbs stragglers, not an unbounded race.
+  const purgeDeadline = Date.now() + 10_000;
+  while (imageDirs.some((imageDir) => existsSync(imageDir)) && Date.now() < purgeDeadline) {
```
Astro-Han, line 232: [P3] Consider waiting on child exit rather than on directory removal, so a contract-permitted purge failure does not cost ten seconds. The comment is right that purge failure is allowed ("the next spawn purges it again") and right that the tolerant wait plus the retrying `rm` is safe — but the two facts combine into a cost: whenever a purge legitimately fails, this loop cannot distinguish that from a purge still running, so it spins the full ten seconds before the backstop takes over, on a suite whose whole point in this PR is CI time. Waiting on the thing the barrier actually means — the children having exited — would return immediately in both cases, and the `maxRetries` `rm` would still absorb the log-flush stragglers it is there for. Not worth blocking on: the failure is rare and the outcome is slow rather than wrong.

Astro-Han (approved):
Thanks — both fixes are at the cause, and the write-up is the kind I wish more flake fixes had: it names the mechanism, cites the failing run and line, and is explicit about what the verification does and does not prove. Reviewed exact head `9269c5fe7de3fc69c891129a77362a5518e23b06`.

The e2e half is right and consistent with precedent — it mirrors dd4b2d0b4 in the same file, and the diagnosis holds: after a compact the composer clears and can remount, `fill()` can land before the contentEditable regains focus, and Enter then submits an empty draft. Typing through the focused element and polling until the draft has settled makes the dispatch depend on observed state rather than on timing. The teardown half is also the right shape: `dispose()` being fire-and-forget is a deliberate production contract, so fixing this test-side rather than making `dispose()` awaitable is correct, and using each service's own image-directory purge as the barrier is a real signal rather than a sleep. I checked that `existsSync` is already imported, that every `makeService` gets a `randomUUID`-unique image directory so the barrier cannot alias, and that a service whose directory was never created simply passes immediately.

One note on verification, in your favour: the e2e run is a step inside the `test` check gated on the change plan, not a separate reported check — and since this PR touches `apps/desktop/e2e/*.spec.ts`, the green `test` at this head does cover the changed spec. The description reads as though a separate `e2e` lane should have reported; it did run, inside `test`.

One P3 inline. Approving — nothing here should hold the merge.

Review disclosure: this review was prepared with Claude Code, which read the diff at this head, checked the import and image-directory uniqueness against the file at head, and read the CI workflow to confirm how the e2e step is gated. Neither suite was executed locally. I reviewed the diff and the findings myself before posting; this approval is my own judgment.

PR #3295, opened 2026-08-20
refactor(storage): centralize root writer lifecycle

Colafornia:
## Summary

Add a storage writer composition that owns the Runtime Host writer lifecycle.

The composition now:

- validates the root lease before opening writers
- permits only one active composition per root lease and rejects overlapping opens
- opens existing domain writers in one defined order
- closes opened writers in reverse order after partial initialization failures
- keeps the lease unavailable if any writer fails to close
- makes `close()` idempotent
- removes writer lifecycle management from Runtime Host domain coordinators

Domain writer interfaces, authorization checks, and persistence behavior remain unchanged. This PR does not introduce a repository framework or merge domain interfaces.

The Runtime Host still bootstraps runtime policy immediately after opening its stores and before opening the remaining writers.

## Verification

- `npm --workspace @maka/storage run typecheck` — passed
- `npm --workspace @maka/runtime-host run typecheck` — passed
- targeted storage composition and usage-store tests — 12 passed
- Biome format check for changed files — passed
- `git diff --check` — passed
- full storage suite — not rerun after the review fixes; the earlier run passed 832 tests and had 3 unrelated environment or fixture failures

## AI use

Select exactly one:

Tool(s) and scope: Codex implemented the storage writer lifecycle refactor, drafted its tests, and addressed code-review feedback. The human contributor reviewed the final diff and owns the decision to submit it. The commit carries a `Generated-by: Codex` trailer.

### DIFF
--- a/packages/runtime-host/src/server/execution-composition.ts
+++ b/packages/runtime-host/src/server/execution-composition.ts
@@ -48,37 +48,16 @@ import {
 } from '@maka/runtime/shell-detect';
 import { type MakaTool } from '@maka/runtime/tool-runtime';
 import { type RuntimeHostedRootAuthority } from '@maka/runtime/message-authority';
-import {
-  openInteractiveProjectCatalogForWrite,
-  type InteractiveProjectCatalogWriter,
-} from '@maka/storage';
 import { createAgentGraphControlStore } from '@maka/storage/agent-graph-control-store';
 import {
   createArtifactAttachmentResourceReader,
   createReadImageSnapshotter,
-  openInteractiveArtifactStoreForWrite,
 } from '@maka/storage/artifact-stores';
-import { openInteractiveScheduledTaskStoreForWrite } from '@maka/storage/scheduled-task-store';
-import { openInteractiveDeepResearchStoreForWrite } from '@maka/storage/deep-research-authority';
-import { openInteractiveDailyReviewAuthorityForWrite } from '@maka/storage/daily-review-authority';
-import { openInteractiveGoalAuthorityForWrite } from '@maka/storage/goal-authority';
-import { openInteractivePlanStoreForWrite } from '@maka/storage/plan-authority';
-import {
-  isSessionNotFoundError,
-  openInteractiveExecutionStoresForWrite,
-} from '@maka/storage/execution-stores';
+import { isSessionNotFoundError } from '@maka/storage/execution-stores';
 import { createExternalSessionAdapterRegistry } from '@maka/storage/external-sessions';
 import { createGitWorktreeChildExecutor } from '@maka/storage/git-worktree-child-executor';
-import {
-  type InteractiveLongTermMemoryWriter,
-  openInteractiveLongTermMemoryStoreForWrite,
-} from '@maka/storage/long-term-memory-store';
-import { openInteractiveMemoryBundleStoreForWrite } from '@maka/storage/memory-bundle-store';
 import { runWithStorageRootLease } from '@maka/storage/root-authority';
-import { openInteractiveRuntimePolicyStoresForWrite } from '@maka/storage/runtime-policy-stores';
-import { openInteractiveTaskLedgerStoreForWrite } from '@maka/storage/task-ledger-authority';
-import { openInteractiveShellRunStoreForWrite } from '@maka/storage/shell-run-authority';
-import { openInteractiveUsageStoresForWrite } from '@maka/storage/usage-stores';
+import { openStorageWriterComposition } from '@maka/storage';
 import { resolveWorkspaceIdentity } from '@maka/storage/workspace-identity';
 import {
   openManagedWorkspaceOwner,
@@ -214,27 +193,22 @@ export async function createExecutionRuntimeHostComposition(
   options: CreateExecutionRuntimeHostCompositionOptions = {},
   dependencies: ExecutionRuntimeHostCompositionDependencies = {},
 ): Promise<ExecutionRuntimeHostComposition> {
-  const stores = await openInteractiveExecutionStoresForWrite(context.owner.lease);
-  await stores.sessionStore.ready();
+  const storage = await openStorageWriterComposition(context.owner.lease, {
+    afterRuntimePolicyOpened: async (stores) => {
+      if (options.bootstrapRuntimePolicy !== false) {
+        await ensureBootstrapRuntimePolicy({
+          workspaceRoot: context.owner.capability.canonicalPath,
+          stores,
+          onDeferredError: (error) =>
+            console.error(
+              `[runtime-host] optional bootstrap target could not be configured: ${generalizedErrorMessage(error)}`,
+            ),
+        });
+      }
+    },
+  });
+  const stores = storage.execution;
   let graphControlStore: ReturnType<typeof createAgentGraphControlStore> | undefined;
-  let taskLedgerStore:
-    | Awaited<ReturnType<typeof openInteractiveTaskLedgerStoreForWrite>>
-    | undefined;
-  let usageStores: Awaited<ReturnType<typeof openInteractiveUsageStoresForWrite>> | undefined;
-  let artifactStore: Awaited<ReturnType<typeof openInteractiveArtifactStoreForWrite>> | undefined;
-  let shellRunStore: Awaited<ReturnType<typeof openInteractiveShellRunStoreForWrite>> | undefined;
-  let longTermMemoryStore: InteractiveLongTermMemoryWriter | undefined;
-  let scheduledTaskStore:
-    | Awaited<ReturnType<typeof openInteractiveScheduledTaskStoreForWrite>>
-    | undefined;
-  let planStore: Awaited<ReturnType<typeof openInteractivePlanStoreForWrite>> | undefined;
-  let deepResearchStore:
-    | Awaited<ReturnType<typeof openInteractiveDeepResearchStoreForWrite>>
-    | undefined;
-  let dailyReviewStore:
-    | Awaited<ReturnType<typeof openInteractiveDailyReviewAuthorityForWrite>>
-    | undefined;
-  let goalStore: Awaited<ReturnType<typeof openInteractiveGoalAuthorityForWrite>> | undefined;
   let graphClient: HostAgentGraphCoordinator | undefined;
   let sessionEffects: HostSessionEffectCoordinator | undefined;
   let memoryExtraction: HostMemoryExtractionCoordinator | undefined;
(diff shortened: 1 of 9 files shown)


Astro-Han (commented):
Reviewed at `3f3dc4ab`. The consolidation is the right move and the shape is good: one owner for writer open order, reverse-order rollback on partial failure, and a single `storage.close()` replacing twenty-odd individually-tracked `let store: X | undefined` variables and their hand-written cleanup ladder in `execution-composition.ts`. Removing 150 lines of that ladder is a genuine simplification, and the failure path is now correct where it previously depended on every future writer being remembered in three places.

I checked the two things most likely to break in a refactor like this and both are fine: `execution.sessionStore.close()` is the *full* teardown — `closeExecutionStorePersistence` closes `runtimePersistence`, `agentRunStore`, `conversationOperationalStateStore`, `messageReceiptStore` and the interaction-store facade — so closing only `sessionStore` is not a leak; and `runtimePolicy` / `memoryBundle` are the two writers passed without a close callback because they genuinely expose none. Closing all stores after every domain module has closed, rather than interleaved per module, is also an improvement: coordinators now reliably shut down before the stores they write to.

The architectural concern is **ownership**, and it is what all three inline findings share. This PR does not just centralize the lifecycle, it also makes the composition a *process-wide singleton per lease* — `openStorageWriterComposition` caches on the lease object and hands the same frozen object to every caller. Before this change, the execution stores were already lease-cached, but the other thirteen writers were not: two callers each got their own project-catalog writer, and one closing did not disturb the other. Now `close()` is an unconditional teardown of a shared object with no reference count, and the second holder's `close()` silently resolves to the first holder's memoized `closeTask`. The new test asserts `first === second` and then closes twice — which encodes the sharing but never exercises the hazard it creates.

That is worth deciding deliberately rather than inheriting: either the composition is genuinely a singleton whose lifetime the root owner controls (in which case `close()` should not be on the object every caller holds), or it is shared and needs a reference count. Today production reaches it through a single root owner so the window is narrow, but `createExecutionRuntimeHostCompositionSource` is a *source* — it can build a composition more than once over a process lifetime, and any overlap between one teardown and the next open lands exactly in that window.

Reviewed with Claude Opus as an analysis assistant. I verified the store close surfaces, the old and new cleanup paths, and the added test by reading source at this head; the race in the first finding is reasoned from the code path and not reproduced by execution.

packages/storage/src/storage-writer-composition.ts
```diff
@@ -0,0 +1,182 @@
...
+
+const compositions = new WeakMap<object, StorageWriterComposition>();
+const openings = new WeakMap<object, Promise<StorageWriterComposition>>();
+
+export async function openStorageWriterComposition(
+  lease: StorageRootLease<'interactive', 'write'>,
+  options: OpenStorageWriterCompositionOptions = {},
+): Promise<StorageWriterComposition> {
+  await assertStorageRootLease(lease, 'interactive', 'write');
+  const existing = compositions.get(lease);
```
Astro-Han: [P2] Evict the composition from the cache when `close()` starts, not when it finishes. `close()` runs `closeInReverse(closes).finally(() => compositions.delete(lease))`, so the entry stays reachable for the entire teardown — an `openStorageWriterComposition` call that arrives mid-close takes this early return and receives a composition whose SQLite handles are already being closed, with no signal that anything is wrong. The new holder then calls `close()` and gets the memoized `closeTask`, so it silently no-ops while believing it released its writers. Reasoned from the code path at this head, not reproduced. Delete the map entry synchronously at the top of `close()` — before awaiting anything — so an overlapping open builds a fresh composition instead of adopting a dying one. Regression test: start `close()`, do not await it, call `openStorageWriterComposition` on the same lease, assert the returned composition is not the closing one and that its writers are usable.

packages/storage/src/storage-writer-composition.ts
```diff
@@ -0,0 +1,182 @@
...
+}
+
+async function createComposition(
+  lease: StorageRootLease<'interactive', 'write'>,
+  options: OpenStorageWriterCompositionOptions,
+): Promise<StorageWriterComposition> {
+  const closes: Array<() => void | Promise<void>> = [];
+  let composition!: StorageWriterComposition;
+  let closeTask: Promise<void> | undefined;
+  const close = () =>
```
Astro-Han, line 65: [P2] Decide whether a shared composition may be closed by any holder, and make the code say so. This `close` tears down all fourteen writers unconditionally, but the composition it belongs to is handed to every caller holding the same lease — so the first caller to finish revokes writers the others are still using, and their own `close()` resolves against the same memoized `closeTask` without doing or reporting anything. Thirteen of those writers were per-caller before this PR, so this is a new lifetime coupling rather than an existing one being centralized. Confirmed by reading code at this head; production currently reaches it through a single root owner, so the window requires one teardown to overlap the next open — which `createExecutionRuntimeHostCompositionSource` can produce, since it constructs a composition per context rather than once per process. Either reference-count the holders and only tear down on the last release, or remove `close` from the shared object and let the root owner own the teardown. Regression test: open twice on one lease, close the first, assert the second holder's writers still work or that its `close()` reports the composition was already closed — the current test closes the same handle twice and cannot distinguish the two designs.

packages/storage/src/storage-writer-composition.ts
```diff
@@ -0,0 +1,182 @@
...
+  } catch (error) {
+    await failOpen(error);
+  }
+  const projectCatalog = await openWriter(
+    () => openInteractiveProjectCatalogForWrite(lease),
+    closeWriter,
+  );
+  const runtimePolicy = await openWriter(() => openInteractiveRuntimePolicyStoresForWrite(lease));
+  try {
+    await options.afterRuntimePolicyOpened?.(runtimePolicy);
```
Astro-Han, line 106: [P2] Do not let a caller's bootstrap option be decided by which call won the race. `afterRuntimePolicyOpened` runs only for the call that starts a new composition — the comment at line 18 says as much — but `execution-composition.ts` uses this hook to run `ensureBootstrapRuntimePolicy` gated on `options.bootstrapRuntimePolicy !== false`. Two callers on one lease disagreeing about that flag therefore get whichever one opened first: a caller that passed `bootstrapRuntimePolicy: false` can have bootstrap performed anyway, and a caller that wanted it can silently not get it. Documenting an order-dependent side effect does not make it safe; the hook is the one option here whose effect is not idempotent. Confirmed by reading code at this head. Either reject a reuse whose options differ from the composition's, or lift bootstrap out of the open path so it runs per caller. Regression test: open with the hook, then open again with a hook that must not run, assert the mismatch is refused rather than ignored.

packages/runtime-host/src/server/execution-composition.ts
```diff
@@ -1398,24 +1344,20 @@ export async function createExecutionRuntimeHostComposition(
...
+        close: [() => memoryExtraction?.close(), () => memory?.close()],
         releaseConnection: [
           (connectionId) => requireMemory(memory).releaseConnection(connectionId),
         ],
       }),
       createRuntimeHostDomainModule({
         id: 'plan',
         handlers: [plans.handlers],
-        close: [() => openedPlanStore.close()],
+        close: [],
```
Astro-Han: [P3] Drop the domain modules' now-empty `close: []` arrays rather than leaving them as placeholders. Five modules keep `close: []` purely because their store teardown moved into `storage.close()`, which reads as "this module has nothing to close" when the truth is "something else closes it now". The next person adding a store to one of these modules has a ready-made slot inviting them to reintroduce exactly the per-module teardown this PR consolidated. Confirmed by reading code at this head. Omit the key where there is nothing to close. No regression test needed; this is a readability change.

Astro-Han (commented):
Re-reviewed at `a9fd1a14`. **All four findings are fixed**, and the author picked the stronger of the two options I offered rather than the cheaper one.

Replacing the per-lease cache and single-flight map with an `activeCompositions` WeakSet removes the sharing that caused three of the four findings at once:

- **Handing out a closing composition** — impossible now. The lease stays marked active until `close()` settles, so an overlapping open throws instead of adopting a dying composition. The new test pins exactly this: it rejects a second open while the first is alive, rejects again *during* `close()`, and then succeeds once the close resolves.
- **Shared teardown with no reference count** — moot, since two holders can no longer exist. This also removes the silent-no-op second `close()`, which was the part I liked least.
- **Order-dependent `afterRuntimePolicyOpened`** — moot for the same reason, and the doc comment was correctly rewritten from "runs only when this call starts a new composition" to what the hook now unconditionally does.
- **Empty `close: []` placeholders** — removed from all three domain modules.

Choosing "one composition per lease, loudly enforced" over "shared with a reference count" is the right call: it makes the lifetime rule checkable at the boundary instead of distributed across holders, and the failure is an error at open rather than a use-after-close later. The duplicate `activeCompositions.has` check before and after `assertStorageRootLease` is not redundant in a way that matters — the pair is atomic because there is no await between the second check and `add`, and the early one just avoids paying for the lease assertion on a busy lease.

One thing worth confirming rather than a finding: this converts a previously silent situation into a hard throw. On `main`, two callers on one lease each opened their own writers; now the second gets `Storage writer composition is already active for this lease`. `createExecutionRuntimeHostCompositionSource` builds a composition per context and `test-only/desktop-e2e-execution.ts` builds one too — please confirm no path constructs two concurrently against a single owner's lease, because that now fails at startup instead of degrading. Failing loud is the right behaviour; I just want it to be a decision rather than a discovery.

Reviewed with Claude Opus as an analysis assistant; verified by reading the full new file, the rewritten test, and the diff against the head I previously reviewed. Nothing was executed.

packages/storage/src/storage-writer-composition.ts
```diff
@@ -0,0 +1,180 @@
...
+
+async function createComposition(
+  lease: StorageRootLease<'interactive', 'write'>,
+  options: OpenStorageWriterCompositionOptions,
+): Promise<StorageWriterComposition> {
+  const closes: Array<() => void | Promise<void>> = [];
+  let closeTask: Promise<void> | undefined;
+  const close = () =>
+    (closeTask ??= closeInReverse(closes).finally(() => {
+      activeCompositions.delete(lease);
```
Astro-Han, line 68: [P3] Decide what a *failed* close should do to the lease. `closeInReverse` collects every writer's error and throws an `AggregateError` at the end, but this `finally` releases the lease regardless — so after a partial close failure a fresh `openStorageWriterComposition` succeeds and opens the same SQLite files while a writer whose `close()` rejected may still hold its handle. The reopen then looks healthy and the double-open surfaces later as lock contention or a corrupt read rather than as the close failure that caused it. Confirmed by reading code at this head; not reproduced, since it needs a writer close to actually reject. Both answers are defensible — release and accept the risk, or keep the lease marked active so the caller must restart rather than silently double-open — but the code currently picks one without saying so. Whichever you choose, say it in a comment here. Regression test: a writer whose `close()` rejects; assert the chosen behaviour for the subsequent open.

Astro-Han (dismissed):
Approving at `a9fd1a14`. All four findings are resolved, and by the stronger route: one composition per lease, enforced at the boundary, with an overlapping open rejected until `close()` settles. That removes the shared-teardown and use-after-close hazards outright rather than papering over them. The one P3 left on the previous review — what a *failed* close should do to the lease — is a decision to record in a comment, not a blocker.

packages/storage/src/storage-writer-composition.ts
Colafornia: Fixed by enforcing one active composition per lease. An open that overlaps an active or closing composition now rejects, so it cannot adopt a dying object. I did not evict at close start because the execution and usage writer caches could still return handles being closed. The regression test starts `close()` without awaiting it, verifies rejection, then verifies a usable reopen after close completes.

packages/storage/src/storage-writer-composition.ts
Colafornia: P3 is fixed too: a failed close now keeps the lease unavailable, with a regression test covering the behavior.

packages/runtime-host/src/server/execution-composition.ts
Colafornia: Fixed. Removed every empty `close` property present at this head: plan, project-catalog, and session. There were three current occurrences rather than five.

packages/storage/src/storage-writer-composition.ts
Colafornia: Fixed by refusing composition reuse. Every accepted open owns a new lifecycle and runs its own `afterRuntimePolicyOpened` callback; a second call on the same lease is rejected instead of silently inheriting the first option. Reopen is allowed after close completes.

packages/storage/src/storage-writer-composition.ts
Colafornia: Fixed by making the composition exclusive instead of reference-counted. A second open on the same lease now rejects while the first composition is opening, active, or closing; only the sole holder can call `close()`. This matches the production root-owner lifecycle and avoids per-holder facade objects. The regression test covers active reuse, close overlap, and usable reopen.

Astro-Han (approved):
Re-reviewed at `38b816c8` after the force-push dismissed my approval. The P3 from my previous review is fixed, and fixed as the decision I asked for rather than as a patch: `.finally` became `.then`, so a close that fails keeps the lease unavailable, with a comment on the line saying exactly why — a failed close may have left a writer handle open. That closes the double-open hazard.

Two things I checked because they follow from that change:

- Dropping the `try`/`catch` that deleted the lease on open failure is correct, not an omission. `failOpen` already routes through `close()`, so a successful rollback releases the lease and a failed one holds it — the old outer `catch` would have released unconditionally and contradicted the new policy.
- The cost is now explicit: a close failure makes that lease unusable for the rest of the process lifetime, recoverable only by restarting. That is the right side of the trade to land on, since the alternative silently opens the same SQLite files twice, and it is now stated in the code instead of implied.

The added test earns it — it forces a real close failure by closing the operational-state database underneath the composition, asserts `close()` rejects with an `AggregateError`, and then asserts the reopen is rejected. That is the assertion that would catch a future revert to `.finally`.

Approving. Reviewed with Claude Opus as an analysis assistant; verified by diffing both files against the head I previously approved. Nothing was executed.

PR #3299, opened 2026-08-20
feat(desktop): fetch custom relay models before connect

localhost-copilot:
## Summary

Add model discovery for unsaved custom relay providers.

Users can fetch models using the configured endpoint, API key, and request headers, then
select a default model from the returned catalog. If discovery fails or `/models` is
unavailable, manual model entry remains available and required.

The preview uses transient Runtime Host verification and does not persist credentials or
connection data. The Runtime Host compatibility epoch is bumped so older hosts reject the
new preview inputs safely.

## Verification

- `npm --workspace @maka/runtime-host test` — 1022 passed
- `npm --workspace @maka/desktop test` — 972 passed
- `npm run lint`
- `npm run format:check`
- `git diff --check`

UI evidence:
mac_1787214476615


packages/runtime-host/src/protocol/index.ts
```diff
@@ -72,9 +72,10 @@ export const RUNTIME_HOST_REGISTRATION_SCHEMA_VERSION = 1 as const;
 export const RUNTIME_HOST_PROTOCOL_VERSION = 0 as const;
 // Increment when the same protocol version no longer guarantees safe Client-Host
 // interoperability. Mismatches are rejected before domain commands are admitted.
-export const RUNTIME_HOST_COMPATIBILITY_EPOCH = 27 as const;
-// 27: Runtime Policy carries the Host-owned shell preference used by tool,
-// PTY, and prompt composition. Older peers cannot safely preserve that field.
+export const RUNTIME_HOST_COMPATIBILITY_EPOCH = 28 as const;
```
Astro-Han: [P1] Coordinate this epoch bump with #3236, which raises the same constant to the same value for a different reason. #3236 changes 27 → 28 for staged `access.credential.prepare`/`finalize`; this PR changes 27 → 28 for the transient endpoint and request-header fields. Both branches write the literal `28`, so git merges them without a conflict and the second to land silently ships one epoch covering two independent incompatibilities. The concrete failure: a Client built from this branch and a Host built from #3236 both advertise 28 and are admitted by `compatibilityEpoch !== RUNTIME_HOST_COMPATIBILITY_EPOCH`, then the Host's `requireExactRecord` rejects `baseUrl` as an unknown field and aborts the transport — which is precisely the outcome the epoch exists to replace with a structured `incompatible` frame. Confirmed by reading both branches at their current heads; not reproduced by merging. Whichever PR lands second must take 29 and append its own comment line rather than accepting the textual merge. A test that pins the epoch to a literal would turn this silent collision into a failing check; there is currently none.

packages/runtime-host/src/server/connection-effect-coordinator.ts
```diff
@@ -195,7 +195,9 @@ export class HostConnectionEffectCoordinator {
     }
     const slug = deriveConnectionSlug(input.providerType);
     const catalog = await this.#stores.connectionCatalog.getSnapshot();
-    const candidate = catalog.connections.find((connection) => connection.slug === slug);
+    const candidate = input.baseUrl
```
Astro-Han: [P2] Decide explicitly whether a Client may direct the Host's outbound requests at an arbitrary address. `validateConnectionBaseUrl` constrains only scheme and length, so any `http:`/`https:` URL is accepted here — including link-local and private-range addresses such as a cloud metadata endpoint — and `createRequestCustomizationFetch` attaches caller-supplied headers to that request. This is a contract decision, not a defect: the capability already exists via create-connection plus `connection.models.fetch`, and the credential guard immediately below this line correctly prevents a stored key from reaching a supplied endpoint. What changes is that the request now requires no catalog mutation and leaves no persisted trace, arriving as #3236 makes remote Runtime Hosts first-class — so the actor and the Host are increasingly on different machines and networks. Confirmed by reading code at this head; not exercised against a live Host. Either state in `docs/runtime-host-remote-access.md` that a Client may originate arbitrary outbound HTTP from the Host, or constrain preview targets. Regression test: whichever rule you choose, assert it here — a preview against a link-local address should have a defined, tested outcome.

packages/runtime-host/src/protocol/connection-effects.ts
```diff
@@ -254,16 +263,41 @@ export function decodeConnectionOnboardingSaveResult(
 export function decodeConnectionOnboardingVerifyInput(
   value: unknown,
 ): ConnectionOnboardingVerifyInput {
+  const fields = requireRecord(value, 'connection onboarding verification input');
   const input = requireExactRecord(value, 'connection onboarding verification input', [
     'providerType',
     'apiKey',
+    ...('baseUrl' in fields ? ['baseUrl'] : []),
```
Astro-Han: [P3] Pass the full key list to `requireExactRecord` instead of deriving it from the value being validated. `assertExactKeys` only rejects keys that are *not* in the allowlist — it never requires a listed key to be present — so `...('baseUrl' in fields ? ['baseUrl'] : [])` admits exactly the same inputs as listing `'baseUrl'` unconditionally. The conditional and the extra `requireRecord` call above it are therefore dead machinery, and worse, they read as though the allowlist adapts to the payload, which is the one thing an exact-record check must never do. The next reader auditing this decoder for injectable fields has to work out that it is a no-op before they can trust it. Confirmed by reading `codec.ts` at this head. List all four keys directly and drop the `fields` binding.

Astro-Han (commented):
Reviewed at `d20fd8c8`. The feature is well-shaped and the two things most likely to go wrong in a change like this are both handled correctly — I checked them specifically rather than assuming:

- **A user-supplied endpoint never receives a stored credential.** `#discoverOnboarding` sets `candidate = undefined` whenever `input.baseUrl` is present, so `exportCredentialMaterial` is never consulted and `secret` collapses to the supplied key alone. Without that guard, `{ baseUrl: <attacker>, apiKey: null }` would have shipped the saved provider key to an arbitrary host. It reads as deliberate, and it is the right guard.
- **Preview-only fields cannot reach persistence.** Changing `ConnectionOnboardingSaveInput` from `extends ConnectionOnboardingVerifyInput` to a standalone interface means `baseUrl` and `requestHeaders` are structurally absent on the save path, so a transient endpoint cannot be smuggled into a stored connection. That is the correct way to express "preview only", better than a runtime check.

The blocking problem is not in this PR's own logic — it is a collision with a PR that is open right now. **#3236 also bumps `RUNTIME_HOST_COMPATIBILITY_EPOCH` from 27 to 28**, for an unrelated reason (staged `access.credential.prepare`/`finalize`). Both branches write the literal `28`, so a textual merge is clean and silent, and the second one to land ships an epoch that no longer distinguishes two independent, mutually-incompatible protocol changes. Details inline; this needs coordinating before either merges.

The remaining architectural question is one of contract rather than correctness. `validateConnectionBaseUrl` allows any `http:`/`https:` URL with no restriction on the host, so this operation lets a Client make the Runtime Host issue an outbound request to an arbitrary address with arbitrary headers. That capability is not new — a user could already create a connection with any `baseUrl`, call `connection.models.fetch`, and delete it. What changes is that it now requires no catalog mutation and leaves no trace, and it arrives at the same time as #3236 makes remote Runtime Hosts a first-class deployment. Whether a remote Host should accept arbitrary outbound targets from its Client is a decision worth making explicitly rather than inheriting.

Reviewed with Claude Opus as an analysis assistant. Every claim here was verified by reading source at this head — including `assertExactKeys`, `validateConnectionBaseUrl`, the epoch comparison in `client/connection.ts`, and #3236's own diff. Nothing was executed; the epoch collision is a reading of both branches, not an observed merge.

Astro-Han (commented):
Re-reviewed at `f96af70e`. **The P1 is resolved.** The epoch is now 29, and the comment block reserves 28 for #3236's staged access-credential pairing — which is better than just moving your own number, because it makes the next person's collision impossible to create silently. Nothing else moved: the delta is `protocol/index.ts` and `protocol.test.ts` only.

My other two findings still stand at this head and I am not re-filing them inline:

- **P2**, the outbound-target contract: `validateConnectionBaseUrl` still constrains only scheme and length, so this operation lets a Client point the Host's `/models` fetch at any `http:`/`https:` address with caller-supplied headers. The credential guard remains correct — a supplied `baseUrl` still forces `candidate = undefined`, so no stored key can reach a supplied endpoint — this is about whether the Host should originate arbitrary outbound requests at all, which matters more as #3236 makes remote Hosts first-class.
- **P3**, the `requireExactRecord` allowlist computed from the value's own keys, which is a no-op because `assertExactKeys` only rejects unknown keys.

Reviewed with Claude Opus as an analysis assistant; verified by diffing against the head I previously reviewed and re-reading both files at this one.

Astro-Han:

Astro-Han:
Independent review of `4a290b704c6fa26be09e07d06cc6a657352aae0a`. **One [P1] — which is an existing finding that has changed state rather than a new one — plus an argument for closing one of the open threads.**

### [P1] The epoch collision is no longer a risk; it has landed

This branch is based on a `main` where `RUNTIME_HOST_COMPATIBILITY_EPOCH` was 36 and bumps it to 37. **`main` is now at 39** (`packages/runtime-host/src/protocol/index.ts:94`, verified directly). Both sides edit the same constant, so git no longer merges it silently — **the PR is `CONFLICTING` / `DIRTY` right now.**

Resolving the text by hand would not be enough: 37 already means something on `main`, so this branch needs a rebase onto current `main` and a bump to **40**.

The earlier note about this was written when it was still a *risk of collision*. It is now a hard blocker. Reporting the state change rather than re-filing it as a new finding.

### On the "can the renderer point the Host at an arbitrary address" thread — I believe this can be closed, with new evidence

I traced `#discoverOnboarding` in `connection-effect-coordinator.ts` on this head independently:

- **When a `baseUrl` override is supplied, `candidate` is forced to `undefined` (`:197-199`), so `stored = null`.** The preview cannot borrow credentials from an existing connection with the same slug. **A stored key cannot be aimed at an attacker-chosen URL** — that path is closed in code, not by convention.
- The transient `apiKey` the user types is sent only to the URL that same user typed. **That is exactly the authority the existing "save the connection, then fetch models" path already has** — and preview is *narrower*, since it persists nothing. Tests pin zero writes on both the catalog and vault sides, and pin that the response carries no secret.
- Protocol level: `file://` is rejected (with a test), the response frame is forbidden from carrying `apiKey` (with a test), and the epoch bump's semantic comment is accurate.

**So there is no secret-exfiltration channel here.** What remains is a product decision — whether a renderer may cause the Host to reach an arbitrary address at all — and the pre-existing create-then-fetch path already answers it the same way. **My recommendation is to close that thread as decided rather than ask for code changes.**

### Still open from the earlier review, unchanged on this head

`requireExactRecord`'s derived key list (`'baseUrl' in fields ? ['baseUrl'] : []`) is untouched. It is functionally correct — unknown keys are still rejected — but it derives *what is permitted* from the input itself, which is a roundabout way to state an allowlist. **Not blocking; noting only that it has had no response.**

- *Preview is a new SSRF surface* — withdrawn; authority does not exceed the existing create+fetch path, and stored secrets are isolated from it.
- *A 64 KB transient `apiKey` crossing IPC and the protocol* — withdrawn; same path as the existing `onboarding.verify`, with protocol tests pinning the bound.
- *A preview failure showing only a banner could mislead* — withdrawn; it falls back to manual entry with a warning, which is the right semantics.

### Scope

+383/-24 across 20 files for "add a fetch-models button to a form" looks heavy until you see the split: most of it is protocol extension (epoch, codec, boundary tests) and two-sided plumbing. That is the legitimate cost of this seam, with no surplus abstraction.

### CI

**`check-runs` on this head: 0 — never run.** No workflow run exists to approve on this SHA, so CI will only appear after the rebase. **No red mark here has never meant green.**

*Reviewed at 2026-08-23 12:50 UTC. Blind line — provisional judgment sealed before the existing reviews were read. No overall verdict offered; note that this PR currently cannot be merged at all.*

Astro-Han:

```
# resolve, then
git push --force-with-lease
```

---

apps/desktop/src/renderer/settings/provider-add-form.tsx
```diff
@@ -129,6 +142,58 @@ export function AddProviderForm(props: {
...
+        message: providerPanelActionErrorMessage(fetchError, locale),
+      });
+    } finally {
+      submitGuard.finish();
+      if (addProviderMountedRef.current) setFetchingModels(false);
+    }
+  }
+
   async function submit() {
     if (submitGuard.current !== null) return;
```
jackwener, line 405: [P3] `submit()` and `fetchModelOptions()` share one `submitGuard`. While preview is in flight (up to the 10s Host fetch timeout), Save stays visually enabled (`isDisabled={busy || isExperimental}` — it does not include `fetchingModels`) but this early return drops the click with no feedback.

Not blocking: the user can wait, or close the form; a failed preview already degrades to manual model entry. If you touch this again, disable Save while `fetchingModels` (or give the two actions separate guards).

jackwener (approved):
Independent review of `aae5cadea973ea5843ed0de7d2292545c9f2038b`.

**GO. No P0–P2.** Hosted `test` is terminal success on this head. PR is CONFLICTING — I am not merging. After rebase, `RUNTIME_HOST_COMPATIBILITY_EPOCH` will need a fresh bump: this SHA goes 42→43, current `main` is already 44.

## What this solves

Unsaved custom relays had no way to discover `/models` before create. Preview reuses `connection.onboarding.verify` with optional `baseUrl` / `requestHeaders`, does not persist catalog or vault state, and the epoch bump is the right handshake fence for the new optional fields.

## Stall / timeout / degrade (the question for this PR)

Preview is **opt-in**, not on the save path. Save still goes `create` then existing `fetchModels`.

- **Timeout, not hang:** `runConnectionModelDiscoveryEffect` uses `MODEL_FETCH_TIMEOUT_MS = 10_000`. A slow or dead `/models` fails; it does not wait forever.
- **Failure is not silent:** warning banner + `modelsFetchFallback`; the default-model field stays a manual text input. Save remains available.
- **Connect does not wait on preview.** The only coupling is the shared `submitGuard` (inline P3): while preview is in flight, Save looks enabled but the click is dropped.

I independently agree the preview URL is not a new SSRF/exfil surface versus create-then-fetch: a supplied `baseUrl` skips the catalog candidate, so a stored key cannot be aimed at an attacker URL. Not re-filed.

No P0/P1/P2 on this head.

Astro-Han (commented):
### #3299 26f5279 — incremental re-review (bind exact head)

**Gate**: CLEAN, `test` green. 4 threads outstanding as noted.
**Scope**: Re-check 4 prior findings + delta since aae5cade only. Not full 20-file re-review.
**Exclusions**: Full-file audit out of scope unless incidental.

**[P1] Epoch collision — resolved by time**: PR now sets `RUNTIME_HOST_COMPATIBILITY_EPOCH=45`, main is 44. So not 28 vs 28 collision; value is strictly greater than base. Contention with other 45 claimants remains per-merge-order but not regression vs main. No block. Verified via `contents?ref=26f5279` vs `ref=main`.

**[P2] validateConnectionBaseUrl allows any http/https including link-local/private** — still present (scheme+length only). Credential guard in `createRequestCustomizationFetch` still blocks leaking stored keys. Determination remains contract-level capability (building connection + models.fetch is intentional). No expansion of surface observed in delta (not reachable pre-onboarding). Not blocking per prior rationale.

**[P3] requireExactRecord dead-chain + submitGuard sharing** — unchanged in delta, P3 non-blocking.

**Verdict**: COMMENT — no P0-P2 in re-check scope. Prior P1 now moot due to main advancement.

<details><summary>中文</summary>增量复核，旧 P1 随主线推进已失效。</details>

Astro-Han:
A number collision to flag, not a review of the change itself.

This branch declares `RUNTIME_HOST_COMPATIBILITY_EPOCH = 49`, and so does #3651. `main` is now at 48, so 49 is the immediate next number and both branches are asking for it. Whichever merges first takes it; the other moves to 50 with its assertion changed to `> 49`.

This branch is approved and #3651 is not, so you are likely to get there first — but the number is only guaranteed at the moment of merge, so it is worth re-deriving then rather than trusting this message:

```
git fetch origin main
git show origin/main:packages/runtime-host/src/protocol/index.ts | grep COMPATIBILITY_EPOCH
gh pr list --repo apache/maka --limit 200 --json number --jq '.[].number' | while read n; do
  gh api repos/apache/maka/pulls/$n/files --paginate \
    --jq '.[]|select(.filename=="packages/runtime-host/src/protocol/index.ts")|.patch' 2>/dev/null \
  | grep -oP '^\+export const RUNTIME_HOST_COMPATIBILITY_EPOCH = \K\d+' | sed "s/^/#$n /"
done
```

I have told #3651 the same thing, so neither of you is being asked to renumber pre-emptively.

<details>
<summary>简体中文</summary>

提醒一个编号冲突，不是对改动本身的评审。

这个分支声明了 `RUNTIME_HOST_COMPATIBILITY_EPOCH = 49`，而 #3651 也声明了同一个值。`main` 目前是 48，所以 49 是紧接着的下一个号，两个分支都在要它。先合并的那个拿到它，另一个改成 50，并把断言改为 `> 49`。

这个分支已获批准而 #3651 还没有，所以大概率是你先到——但这个号只有在合并那一刻才有保证，因此建议届时重新推导一次，而不是相信这条消息：

```
git fetch origin main
git show origin/main:packages/runtime-host/src/protocol/index.ts | grep COMPATIBILITY_EPOCH
gh pr list --repo apache/maka --limit 200 --json number --jq '.[].number' | while read n; do
  gh api repos/apache/maka/pulls/$n/files --paginate \
    --jq '.[]|select(.filename=="packages/runtime-host/src/protocol/index.ts")|.patch' 2>/dev/null \
  | grep -oP '^\+export const RUNTIME_HOST_COMPATIBILITY_EPOCH = \K\d+' | sed "s/^/#$n /"
done
```

我已经对 #3651 说了同样的话，所以你们两边都不需要现在就抢着改号。

</details>

Astro-Han:
Status note on mergeability, not a new review finding. The code review stands and the `test` check is green on `a1194c840`.

Your rebase on 2026-08-25 worked — the earlier conflict set (`protocol.test.ts` and several `protocol/*` files) is fully resolved. `main` has since moved again, so GitHub shows this as CONFLICTING once more, but the conflict has collapsed to a single file.

I ran the merge against current `main` with `git merge-tree`, which writes nothing and does not touch your branch:

```
Auto-merging apps/desktop/src/main/runtime-host-client.ts
Auto-merging apps/desktop/src/preload/bridge-contract.d.ts
Auto-merging apps/desktop/src/preload/preload.ts
Auto-merging apps/desktop/src/renderer/locales/settings-provider-copy.ts
Auto-merging apps/desktop/stories/settings/provider-settings.stories.tsx
Auto-merging apps/desktop/stories/settings/settings-pages.stories.tsx
Auto-merging docs/astryx-surface-file-inventory.md
Auto-merging packages/runtime-host/src/protocol/index.ts
CONFLICT (content): Merge conflict in packages/runtime-host/src/protocol/index.ts
```

Everything else auto-merges. The one conflict is `RUNTIME_HOST_COMPATIBILITY_EPOCH` again: your branch declares 49, `main` is now at 50.

One thing worth knowing before you pick a number. Do not take 51. I scanned all 127 open PRs for their declared epoch, and four of them already claim 51: #3651, #3704, #3781, #3818. Whichever of those merges first takes it, and anyone else holding 51 conflicts again immediately.

The only thing that actually works here is to rebase, read `RUNTIME_HOST_COMPATIBILITY_EPOCH` off `main` at that moment, set yours to that value plus one, and add your changelog line above the existing entries. Expect it to move again if you sit unmerged for long — that is not something you are doing wrong.

This is the fourth time this PR has been blocked on the epoch number and never on the change itself. The counter is a hand-assigned monotonic integer plus a changelog comment in one file, so every protocol-affecting PR edits the same two lines and serializes against every other one. That is a repository-level problem rather than yours, and I will raise it separately.



---

apps/desktop/src/renderer/settings/provider-add-form.tsx
```diff
@@ -129,6 +152,60 @@ export function AddProviderForm(props: {
...
+      const models = await props.bridge.previewModels({
+        providerType: props.providerType,
+        ...(normalizedBaseUrl ? { baseUrl: normalizedBaseUrl } : {}),
+        ...(normalizedApiKey ? { apiKey: normalizedApiKey } : {}),
+        ...(Object.keys(normalizedRequestHeaders).length > 0
+          ? { requestHeaders: normalizedRequestHeaders }
+          : {}),
+      });
+      if (!addProviderMountedRef.current) return;
+      setActionState((current) => ({ ...current, discoveredModels: models }));
```
hqhq1025: [P2] Discard model previews when their inputs change

`previewModels` captures the endpoint, API key, and request headers before the await, but those controls remain editable while `fetchingModels` is true because they are disabled only by `busy`. Their change handlers clear the current catalog, yet a slow old request can return here afterward and repopulate it with models from the previous inputs. I reproduced this by starting discovery against `old.example`, changing the endpoint to `new.example` before resolution, and then saving: the create call received `{ baseUrl: "https://new.example/v1", defaultModel: "old-model" }`. The post-create catalog fetch does not repair this because reconciliation intentionally preserves user-selected models that the new inventory omits. Please bind each result to an input generation/fingerprint or disable all discovery-affecting controls for the request, and add a deferred-response regression.

hqhq1025 (commented):
Reviewed exact head `8a4fadf501bbc277a450279ba1efac35b3351e41`. **One P2 finding; not ready to merge.**

This change adds unsaved custom-relay model discovery through a new `connections:previewModels` path, carries transient endpoint/API-key/header inputs to the Runtime Host without persisting preview state, and bumps the compatibility epoch.

Validation: clean install, `npm run build:test`, full typecheck, changed-file Biome, `git diff --check`, Runtime Host protocol/coordinator tests (89/89), Desktop connection IPC tests (11/11), and a React deferred-promise reproduction all completed. Hosted `test` and `windows_recovery` are green. The unrelated full-suite MCP OAuth cancellations and sandbox-context failure reproduce unchanged on exact base `9ff2675`.

I did not exercise a real third-party relay or packaged Electron UI.

packages/runtime-host/src/protocol/index.ts
```diff
@@ -100,7 +100,9 @@ export const RUNTIME_HOST_REGISTRATION_SCHEMA_VERSION = 1 as const;
 export const RUNTIME_HOST_PROTOCOL_VERSION = 0 as const;
 // Increment when the same protocol version no longer guarantees safe Client-Host
 // interoperability. Mismatches are rejected before domain commands are admitted.
-export const RUNTIME_HOST_COMPATIBILITY_EPOCH = 100 as const;
+export const RUNTIME_HOST_COMPATIBILITY_EPOCH = 101 as const;
```
Astro-Han: P1: This head publishes epoch 101 while `main` is already at 109, and the branch conflicts with `main` in exactly this file. The conflict means nothing merges silently, which is good, but resolving it as "ours" would move the published epoch from 109 down to 101 and destroy the handshake fence for eight independent incompatibilities. `main`'s own assertions top out at `> 102` (`protocol.test.ts:259`), so CI would not catch it. Smallest fix: rebase onto `main`, take 110, and append the comment line above `main`'s 109 entry. `packages/runtime-host/src/protocol/connection-effects.ts` also conflicts and needs a real resolution, since this decoder changed on `main` too.

packages/runtime-host/src/__tests__/protocol.test.ts
```diff
@@ -426,6 +426,10 @@ describe('Runtime Host bootstrap protocol', () => {
     assert.ok(RUNTIME_HOST_COMPATIBILITY_EPOCH > 99);
   });
 
+  test('publishes a new compatibility epoch for transient onboarding request headers', () => {
+    assert.ok(RUNTIME_HOST_COMPATIBILITY_EPOCH > 100);
```
Astro-Han: P3: `> 100` stops asserting anything once this branch rebases, because `main` is already at 109 and satisfies it on its own. That leaves the bump this PR needs unpinned, which is the thing the assertion exists to prevent. Move it to `> 109` in the same commit that takes epoch 110.

apps/desktop/src/renderer/settings/provider-add-form.tsx
```diff
@@ -645,17 +717,49 @@ export function AddProviderForm(props: {
...
-            onChange={(value) => {
-              setDefaultModel(value);
-              resetManagedVerification();
-            }}
-            placeholder={copy.defaultModelPlaceholder}
-            isDisabled={isExperimental || busy}
-            label={copy.defaultModel}
-            description={copy.defaultModelHelp}
-          />
+          discoveredModels ? (
```
Astro-Han: P2: A successful preview replaces the default-model text field with a closed `Selector`, so a model id the relay's `/models` omits can no longer be entered. Gateways commonly serve a partial catalog, and the only way back to free text is to edit the endpoint (which calls `invalidateDiscoveredModels`) and change it back. That contradicts the PR description's "manual model entry remains available and required". Smallest fix: keep the `TextInput` and offer the discovered ids as suggestions, or give the `Selector` an explicit manual-entry escape.

apps/desktop/src/main/runtime-host-connections-ipc-main.ts
```diff
@@ -343,6 +345,20 @@ export function registerRuntimeHostConnectionsIpc(
     const latest = requireConnectionIdentity(await snapshot(), connectionIdentity(current));
     return { models: [...latest.models], source: result.source };
   });
+  deps.ipcMain.handle('connections:previewModels', async (_event, raw: unknown) => {
```
Astro-Han: P2: This opens a second Desktop path to one Host operation. `connections:previewModels` targets `connection.onboarding.verify`, the same operation `connections:onboardingVerify` already handles a hundred and seventy lines above, and `DesktopRuntimeHostClient.previewConnectionModels` is byte-identical to the existing `verifyConnectionOnboarding`. The new path also hand-rolls `normalizePreviewConnectionModelsInputForIpc` instead of using `CONNECTION_EFFECT_OPERATION_SPECS['connection.onboarding.verify'].decodeInput`, which is the protocol's own decoder, so the two channels can drift on what they accept.

The seam is already inside the component that needs it: `provider-add-form.tsx:272` calls `props.apiKeyOnboardingBridge.verify({ target: { kind: 'create', providerType }, apiKey, baseUrl: null })`. Having `fetchModelOptions` call that with `baseUrl` and `requestHeaders` filled in, and render failures through the existing `onboardingFailureMessage(result)`, would let this PR delete the new core type `PreviewConnectionModelsInput`, the new IPC normalizer, this channel, the duplicate client method, and the `fetchModels<T>` overload in `preload.ts` that discriminates on `'connectionId' in input` behind a conditional return type. Net effect is a smaller diff that reaches the same behavior with no new concept.

Astro-Han (commented):
Re-reviewed at `a645f490`. Every prior inline finding is addressed; what remains is a rebase gate and two things worth fixing before merge.

**Prior threads, checked against this head:**

- P1 epoch collision with #3236: moot. Both branches moved on; see the new P1 below, which is a different problem with the same constant.
- P2 arbitrary outbound target: no longer a finding against this PR. `baseUrl` on `connection.onboarding.verify` is now `main`'s own field, landed with managed onboarding, so this PR does not introduce that capability. Its net Host-side addition is `requestHeaders`, and that goes through `normalizeRequestHeaders`, the same authority as persisted connection headers, with a protected-header denylist (`authorization`, `host`, `x-api-key`, `proxy-authorization`) and count/length/byte caps. If the outbound-target contract is still worth deciding, it belongs on `main`, not here.
- P3 `requireExactRecord` allowlist derived from the value: fixed. `decodeConnectionOnboardingVerifyInput` now uses `requireShapedRecord` with explicit required and optional key lists.
- P3 shared `submitGuard` dropping the Save click: fixed. Save is now `isDisabled={busy || fetchingModels || isExperimental}`.
- P2 slow preview repopulating the catalog after the inputs changed: fixed, and fixed at the right level. The API key, endpoint, and request-header editor are all disabled for the duration of the request, and `AddCustomRelayWhileModelDiscoveryPending` drives the real component to assert it.

I also re-checked the credential boundary and it is now stronger than when I first read it. For a `create` target, `beginConnectionOnboarding` allocates a fresh `connectionId`, so `existingConnection` is undefined and both `storedSecret` and `requestHeadersSecret` are structurally null. No stored credential can reach a supplied endpoint by construction, rather than by a runtime guard. The coordinator test asserts the catalog and vault are untouched.

One cross-cutting note for anyone else reading this: diffing this head against current `main` shows a pile of apparent deletions (`slug_taken`, the Copilot OAuth slug, OAuth subscription token parsing, the `empty` discovery kind). Those are artifacts of a stale base, not changes in this PR. The real delta is against `e1e108a792`, and on the Host side it is about 30 lines.

Reviewed with Claude Opus as an analysis assistant. Verified by reading source at this head, diffing against the PR's own base, and running `git merge-tree` against `main`; nothing was executed.

Astro-Han (commented):
Reviewed current head `f19053f17c15f31cdf1535d559894249da62e298` (OPEN). Technical GO — no P0–P2, one P3 below. This is a user-facing feature; merging is a human decision regardless.

## What it does

The "add provider" form for custom relays gets a "pull models" button: after endpoint and key are filled, it queries the relay's model list once so the default model can be picked from a dropdown instead of typed by hand. For this, the host protocol's `connection.onboarding.verify` gains an optional `requestHeaders` field, letting the probe carry user-configured custom headers (e.g. `X-Tenant`).

## The security question that matters most: clean

This opens a user-controlled HTTP header channel. Two concerns, both closed: the implementation reuses the existing `normalizeRequestHeaders` (`packages/core/src/request-customization.ts:63`) instead of inventing new validation — header names go through the HTTP token charset regex (`request-customization.ts:43`), which excludes CR/LF, so header injection is impossible; and no new credential channel is created. The one confirmed new behavior — a `modelDiscovery` error field rendered in exactly one place (`provider-add-form.tsx:897`) — cannot double-display.

## P3 — passing an explicit empty object differs from omitting, and it is undocumented

`packages/runtime-host/src/server/connection-effect-coordinator.ts:290-292` uses nullish coalescing, so an explicitly passed empty headers object behaves differently from an omitted one. Please document the difference.

## Gate note

The only APPROVED on record binds an older head and does not cover this one; a fresh human approval is still needed.

---

本条结论全部来自 @Opus-Qronos-AstroHan 的审查。我自己没有读这份 diff；我核的是当前 head 有没有漂移。当前 head 是 f19053f1，未关闭。新增拉取模型按钮，安全问题已排除。P3 是显式空对象和不传行为不同但没写下来。旧批准不覆盖当前 head，还需人类重新批准。

</details>

packages/runtime-host/src/protocol/index.ts
localhost-copilot: Rebased onto current main and bumped the epoch from 121 to 122

packages/runtime-host/src/__tests__/protocol.test.ts
localhost-copilot: Updated.

apps/desktop/src/renderer/settings/provider-add-form.tsx
localhost-copilot: Updated.

apps/desktop/src/main/runtime-host-connections-ipc-main.ts
localhost-copilot: Updated.

me2seeks (commented):

**P2 (Should-Fix) — the new failure path has no test that runs by default, and the verification command in the description does not exist on this head.**

The probe's request shaping and its failure handling live inline in the component:

```tsx
// apps/desktop/src/renderer/settings/provider-add-form.tsx:235-289
async function fetchModelOptions(...)      // builds the verify input, handles the result
// :191-197
function invalidateDiscoveredModels(...)
```

so the branch that matters — what the user sees when the fetch fails or the host rejects it — is only exercised by a Storybook play function, which runs under the plan-gated `storybook` job rather than the package unit tests. The sibling module in this feature documents the opposite convention precisely because inline decisions here are unobservable from a test. Concretely uncovered: an unreachable host or a `failed` verify result (the banner plus "you can still enter a model ID manually" path), a thrown bridge error, and the omitted-versus-forwarded request-headers decision — the latter being the reason the compatibility epoch moved.

Related: the description cites `npm --workspace @maka/desktop test`, but the desktop package at this head has only `test:dist`, which is also what the repo runner invokes, so that command cannot have produced the reported output. There is no CI run on this SHA.

Smallest sound fix: extract the request shaping into the sibling module (matching its stated purpose) and add cases there for no-headers, headers-forwarded, a `failed` result, and a rejection.

**P2 (Should-Fix) — the branch conflicts and its epoch is behind `main`.** The head declares epoch 122 while `main` is at 141, so the protocol file conflicts and the epoch guard fails on a decreasing epoch. Rebase, renumber, and re-run before merge.

**P3 (Nice-to-have) — stale auto-filled model id.** A successful fetch auto-fills the default model, but the invalidation path (triggered when the endpoint, key or headers change) clears the list without clearing that id, so repointing the connection at a different relay can create a connection whose only enabled model id was never observed there. Smallest fix: remember the auto-filled id and clear it during invalidation while it still equals the field.

**P3 (Nice-to-have) — an unreachable branch and its copy.** The empty-catalog branch in the component cannot fire: the host converts an empty discovery into an invalid-response failure and the wire rejects an empty list, so that copy is unreachable from this path.

**P3 (Nice-to-have) — one new test asserts the wrong reason.** The decode case rejecting `requestHeaders: { Authorization: 42 }` passes because `authorization` is a reserved name (checked before the value type), so it would still pass if numeric values were accepted; the value-type rejection and the size bound are untested. Use a non-reserved name and add an over-limit case.

**Checked and clean (recorded because this is the security-relevant part):** the credential cannot be aimed at an unintended host. The renderer always sends a create target, and the host assigns stored vault material only in the existing-connection branch, so a create probe reads no vault secret and a blank key is rejected rather than substituted. The probe is bounded (per-request timeout, response size cap, header count/size/value limits, URL length, scheme allowlist, reserved headers refused), persists nothing, and the response frame cannot carry the key — all pinned by tests. The fetch list is component state, cleared when the key, base URL or headers change, and the new button cannot submit the surrounding form.

**Review-relevant risks.** The compatibility-epoch change is a public contract and needs independent human review under `CONTRIBUTING.md`. The model list preview is advisory only — the persisted catalog is fetched again after create — so a preview/save divergence would be invisible to the user.

**Required conclusion.**
1. Optimal for the actual problem? Mostly — reusing the existing verify operation with one optional field is the smallest coherent seam. Not optimal in that the request shaping sits inline where it cannot be tested, and the auto-fill adds behavior without matching invalidation.
2. Production code that can be deleted? The unreachable empty-catalog branch and its copy.
3. Low-quality tests to delete or replace? None to delete; one to strengthen (the reserved-name false positive), plus the missing failure-path cases.
4. Deeper refactor required? No architectural change — move the request shaping next to its sibling and test it there.
5. Ready to merge? No: the branch conflicts, the epoch is behind, and the new failure path is untested where it counts.
6. Residual risks / verification gaps: I did not run the suite. Whether cross-origin redirects forward custom headers is a repo-wide pre-existing question, not introduced here, and I did not verify it.

jackwener:
**[kabi-opus-dev]** — Stale-approval notice. **The existing `jackwener` approval on this PR does not cover the current code.**

- That approval was submitted `2026-08-23` and explicitly scoped itself to `aae5cadea973ea5843ed0de7d2292545c9f2038b` ("Independent review of `aae5cadea`").
- The head has since been rewritten to `9de694849` (`2026-09-06`). Nothing at `9de694849` has been reviewed by this account.
- GitHub still reports `reviewDecision: APPROVED` for this PR, so **the review gate is currently green off a SHA no one reviewed.** The PR is `CONFLICTING`, which is what is holding it today — but that is a separate, fixable blocker. Once it is rebased, the merge gate would already be satisfied by a signal that is not valid.

Treat the approval as void from here. **I am deliberately not dismissing it** — dismissing moves a merge gate, and that is a maintainer decision, not an automated reviewer's. A human should either re-request review or dismiss it explicitly.

One concrete thing the drift already invalidates: the original review flagged that after a rebase `RUNTIME_HOST_COMPATIBILITY_EPOCH` would need a fresh bump. At the time that was `42 → 43` against a `main` of `44`. As of now the current head carries **`122`** while `main` is at **`165`**, so the handshake fence is far staler than what was reviewed and needs re-deciding on the rebased result, not carried over.

Disclosure: `jackwener` is shared by more than one automated review seat. This notice is from a different seat than the one that left the original approval, and it is a procedural correction, not an independent re-review of the code — no one has reviewed `9de694849`.

PR #2445, opened 2026-08-07
fix(storage): serialize operational schema migrations

Astro-Han:
## Summary

Refs #2357.

Released operational databases need one migration owner: classify every schema scope, recheck under the write lock, migrate, and publish the new registry as one SQLite transaction.

This PR:

- runs operational schema upgrades under one `BEGIN IMMEDIATE` transaction and rolls back every scope on failure;
- migrates released v0.1.6-v0.1.10 Plan Reminder records into the current Scheduled Task authority when every persisted fact is representable;
- fails closed before mutation when released Automation exists or Reminder state cannot be represented losslessly, leaving the old database available to the previous Maka release;
- preserves released Automation provenance in Session messages, admissions, AgentRuns, and RuntimeEvents after #2742 removed the execution authority;
- canonically settles orphaned legacy Automation executions during Runtime Host recovery and shows their provenance as read-only history;
- reports migration blockers and permanent storage failures through stable Candidate exit codes to actionable CLI and Desktop errors.

The upgrade is one-way after a successful migration. It does not restore historical backups, preserve Automation as a second authority, convert legacy Automation into Scheduled Tasks, or support an old process continuing to write after migration.

## Verification

- Built `@maka/core`, `@maka/storage`, `@maka/runtime`, `@maka/runtime-host`, Desktop main, and `@maka/ui` from clean artifacts.
- Focused Operational migration and concurrency tests: 32 passed.
- Focused persisted-codec and Core execution tests: 27 passed.
- Focused Runtime Host production-composition tests: 11 passed.
- Focused Runtime Host kernel and Candidate lifecycle tests: 47 passed.
- Focused Desktop owner tests: 16 passed.
- Focused CLI context tests: 5 passed.
- `npm run windows:inventory`.
- `npm run format:check`.

Coverage includes concurrent initialization, transaction rollback, a frozen real v0.1.6 database, successful Reminder migration and reopen, Automation fail-closed behavior, unrepresentable Reminder history preservation, legacy Automation message recovery, canonical closure of an orphaned legacy admission, Candidate failure transport, and single actionable CLI/Desktop presentation.

## Review focus

Please verify the released database lifecycle from classification through migration or rollback and reopen. In particular:

- successful migration preserves every representable Reminder fact;
- Automation and unrepresentable Reminder state fail before mutation while their authority still exists;
- databases already opened by #2742 retain readable legacy Automation history and recover without restoring that authority;
- migration and storage failures reach CLI/Desktop once with actionable guidance.

## Checklist

[...]

PR #1994, opened 2026-08-03
refactor(storage): make SQLite the sole operational authority

jackwener:
## Summary

- make `runtime.sqlite` the only operational authority for Sessions, execution ledgers, workflows, usage, Artifacts metadata, Automations, Daily Review, and Headless TaskRuns
- remove File/JSONL stores, startup cutovers, legacy imports, compatibility facades, Session JSONL transfer/maintenance, and redundant parity tests
- keep filesystem storage only for user configuration, Artifact bytes, foreign inputs, and explicit benchmark/protocol outputs
- export and back up canonical state as SQLite plus Artifact bytes, with Artifact writer locking, SQLite/content validation, owner-only modes, fsync discipline, and atomic publication
- address review blockers: seed fresh pricing authority, repair legacy-schema rewind fixtures, delete stale legacy-import tests, remove the obsolete admission path assertion, restore ShellRun `ENOENT`, and align the health-notice E2E with the surviving fixture

## Compatibility decisions

Backward compatibility and old File/JSONL data migration are intentionally not retained.

- Existing conversation history that exists only in legacy transcript files is not backfilled into `session_messages`; affected sessions can retain metadata/title but open with an empty thread. This irreversible upgrade boundary is accepted and documented in the README.
- Pre-version or Electron `safeStorage` credential/token files are not imported. `credentials.json` is the sole authority and affected users must re-authenticate; README and SECURITY now state this explicitly.
- The broad legacy/cutover parity suites removed by this PR are deliberately deferred rather than restored. The retained SQLite tests cover current authority behavior, schema upgrades, admission, and backup integrity; additional race/keyset/projection depth can return as focused follow-up coverage without preserving deleted file-store contracts.

## Verification

- Storage typecheck/build
- Runtime, Runtime Host, Desktop main, and CLI typecheck on the original change
- Storage targeted tests: 66 passed on the original change
- Desktop targeted tests: 77 passed on the original change
- Headless TaskRun targeted tests: 8 passed on the original change
- Latest review-fix commit: `npm --workspace @maka/storage run typecheck`
- Latest review-fix commit: `git diff --check`

No additional local test suite was run for the final review pass; CI is the merge gate.

PR #3301, opened 2026-08-20
refactor(storage): drop the barrel and publish narrow entrypoints

### LINKED ISSUE #1257: fix(cli): avoid node:sqlite ExperimentalWarning on non-SQLite storage imports
## Summary

Importing the top-level `@maka/storage` package eagerly evaluates the SQLite runtime store and writes Node's `ExperimentalWarning` to stderr, even when the caller only uses a non-SQLite store.

## Reproduction

Environment: Node v22.22.0, current `main` at `6274189092ca7771427671cc48aafb63edcb9057`.

```bash
node --input-type=module -e "await import('@maka/storage')"
node --input-type=module -e "await import('./packages/storage/dist/session-store.js')"
```

## Actual behavior

The top-level package import prints:

Image

```text
ExperimentalWarning: SQLite is an experimental feature and might change at any time
```

The direct non-SQLite module import remains clean.

## Evidence

The screenshot below captures both commands in the same VS Code integrated terminal:

## Impact

- CLI subprocess tests that require clean stderr fail.
- Commands such as `maka inspect` / `maka run` can expose unrelated SQLite warnings to users.
- Non-SQLite consumers pay a runtime side effect for a feature they did not request.

## Root cause evidence

- `packages/storage/src/index.ts` re-exports `./sqlite-runtime-store.js`.
- `packages/storage/src/sqlite-runtime-store.ts` statically imports `DatabaseSync` from `node:sqlite`.
- Importing the barrel therefore evaluates `node:sqlite` eagerly.

## Acceptance criteria

- Importing `@maka/storage` without constructing a SQLite store writes nothing to stderr.
- Constructing and using the SQLite runtime store still works.
- Existing CLI, storage, build, typecheck, lint, and regression tests pass.

childrentime:
## Summary

`maka --help` printed Node's SQLite `ExperimentalWarning` to stderr, on a command that never opens a database. Every other CLI entry did the same. This removes the structure that caused it: `@maka/storage` no longer publishes a barrel, and every consumer imports the narrow entrypoint that owns what it needs.

Fixes #1257

## Why the warning was there

Node evaluates a builtin the moment it enters a module graph. A static value import is therefore not a declaration of intent — it *is* the load:

```ts
import { DatabaseSync } from 'node:sqlite'; // evaluated here
```

Exactly three modules do this: `operational-target-schema.ts`, `operational-state-backup.ts` and `session-bundle-policy.ts`.

`operational-target-schema.ts` is the amplifier. `operational-state-store.ts` imports it — and `operational-state-store.ts` is imported by roughly forty modules, which is how three import statements reached 45 of the package's modules. The last link is the CLI: `cli-core.ts` imports `@maka/storage` to reach `resolveMakaDataRoots`, a pure path helper, and the barrel's `export *` surface hands it the whole storage layer, SQLite included, before `--help` prints a single line.

## The alternative this PR chooses against

Review correctly pointed out that removing the barrel is not the only fix, and the smaller one deserves to be stated so the choice here is deliberate. The three static imports above all construct only inside function bodies, and this package already owns a lazy-load seam twice over (`loadDatabaseSync()` in `operational-state-store.ts`, `loadSqliteModule()` in `sqlite-session-metadata-store.ts`). Converting the three imports to `import type` plus that seam is a three-file fix that keeps the barrel and silences the warning.

This PR prefers the structural fix because #1257 was already fixed once the small way, and it came back: its guard test asserted an empty stderr, #2710 removed that assertion in a bulk test cleanup, #1994 and #2445 then added static imports, and the warning shipped again. Lazy loading keeps the boundary as a convention each future edit must remember; removing the barrel makes it structural — widening the SQLite surface now requires editing a published entrypoint list in a test and saying why. The two designs buy different things (three files of churn vs. cost made legible at every import site), and a maintainer preferring the minimal fix is a legitimate outcome of this review; I'd then close this and submit the three-file version.

## Change

- Delete `packages/storage/src/index.ts` and drop `.`, `main` and `types` from the manifest.
- Publish the modules consumers actually reach as narrow subpaths — the map has 49 entrypoints, and every one of them has a consumer outside the package (six consumer-less entrypoints predate this change and are allowlisted exactly; a *new* one fails the guard outright).
- Where the barrel exported hand-picked names rather than a whole module, the **newly published** subpath carries a facade with exactly those picks, following the existing `interaction-store-public` idiom. That is now one entrypoint, `operational-state-store`, where the schema-migration internals stay private. Every other hand-picked module was already surface-neutral, verified by diffing the barrel's picks against each module's runtime exports.
- No pre-existing entrypoint changes target. The export map only drops `.` and adds 15 subpaths, which `git diff upstream/main...HEAD -- packages/storage/package.json` shows directly.
- Rewrite every consumer onto those subpaths (95 files against current `main`; the touched files in `packages/runtime` are all tests, that package's production code is untouched).
- Two tests moved off the barrel's *shape* rather than its contents. `managed-workspace-baseline` now loads every published entrypoint and asserts the union of reachable symbols excludes the internals — a claim that survives a future re-export, which an assertion on the `exports` map's targets would not. `provider-request-capture-artifact` reaches its subject directly.

## Regression guard

`public-entrypoints.test.ts` loads all 49 published entrypoints and asserts:

- the package publishes no `.` entrypoint and no `main`/`types` field;
- every published target is a file the build actually emits;
- no source file in the repository imports the bare `@maka/storage` specifier, side-effect form included — this is the test that would have caught the writer-composition consumer #3295 landed on `main` while this branch was in review, which merged cleanly and then failed to build;
- every imported subpath is published, and the set of published-but-unconsumed entrypoints is exactly the six that predate this change;
- the entrypoints whose module graph reaches `node:sqlite` are **exactly** the 29 listed in the test, leaving 20 that are free of it. Reaching `node:sqlite` is observed through a `module.registerHooks` resolve hook rather than by matching the warning's text, which Node owns and has already reworded once; probe children are capped at four concurrent.

That list is the package's SQLite boundary written down. Widening it later means editing the list and saying why. (`./storage-writer-composition` is on it because it statically imports `execution-stores` and thirteen other SQLite-backed modules.)

Scripts import workspace modules by `dist/` file path, outside both the export map and the typechecker — the release smoke script through installed `node_modules`, the computer-use scripts through relative `packages/*/dist` paths. `release-cli-file-policy.test.mjs` walks every script for both forms and asserts each target is a file the build emits.

## Rebase onto current `main`

Rebased from `84fbe05` onto `23c9214` and force-pushed. Most conflicts were `main` adding the ASF header to files this branch re-imports, and took `main`'s file with the narrow import. Three needed a decision:

- `main` grew four new bare `@maka/storage` importers while this sat in review. All four are on subpaths now — none of them conflicted, `public-entrypoints.test.ts` is what found them, which is the guard doing the job the #3295 case was added for.
- **`./artifact-store` is no longer published.** Its only consumer outside the package was the desktop test `main` deleted, and a new consumer-less entrypoint is what this PR's own guard rejects. The module and its lease-gated write authority stay package-private; its facade is gone, so two facades remain rather than three. `sanitizeArtifactName` is still reachable through `./artifact-stores`, as it already was.
- **`./model-call-ledger` joined the consumer-less allowlist** untouched by this branch: `repairPendingModelCallProjections` lost its last caller when `canonical-usage-reader` was rewritten on `main`. Retiring a subpath that already shipped is its own compatibility call, so it is listed with that reason recorded in the test.

`main`'s own new entrypoints — `./process-lifetime-file-update-lock` and `./stable-storage` — both have consumers and carry through unchanged.

## Follow-up after review

`a2c64e3` answers @M4n5ter's P2. `./credential-store` was published before this branch and mapped straight to `credential-store.js`, so `withCredentialFileLock` was reachable through it; routing it through a facade made that symbol `undefined`. That is a contract change to a pre-existing entrypoint and unrelated to removing the barrel — the facade was applying the barrel's picks to an entrypoint the barrel never owned, a rule that belongs to subpaths published here for the first time.

Preserved rather than retired: the subpath points back at `credential-store.js`, the facade is deleted, and `withCredentialFileLock` leaves the reachable-symbol denylist because it is public again. Whether the file lock should become package-private stays open as its own change.

## Verification

Node v24.11.1, npm 11.19.0 (the pinned `packageManager`), macOS arm64, at this head. The surfaces are the ones `scripts/ci-test-plan.mjs --base 23c9214 --head HEAD` selects for this PR: `code`, `astryx_surface`, `release_contract`, `cli_package`, `runtime_host`, `runtime_sandbox`, `e2e`, `storybook`, plus workspaces `storage, runtime, eval, computer-use, cli, desktop`.

| Check | Result |
| --- | --- |
| `ci-test-plan` · `verify-windows-harness` · `protocol-epoch-check` · `ax-tree-audit` tests | 92 pass |
| `windows:inventory` · `check:asf-npm` · `check:asf-headers` | pass (2801 covered / 137 excluded files) |
| `protocol-epoch-check --base HEAD^1` | pass, no protocol change (epoch 44) |
| `lint` · `format:check` | pass, 2667 / 1611 files |
| `astryx:surface-inventory` · `astryx:theme -- --check` | pass |
| `build` · `typecheck` · `knip` (desktop, ui) | pass |
| `check:release` | 110 pass |
| workspace `test:dist`, via CI's own commands | runtime 3003 (2990 pass, 13 skipped) · desktop 1364 · runtime-host 1120 · storage 925 (910 pass, 14 skipped, 1 below) · cli 423 · computer-use 118 · eval 75; `core`, `mcp` and `ui` pass in a full sweep |
| `release:cli:pack` | pass — 13.4 MiB, 6346 files. The tarball ships `node_modules/@maka/storage/dist/workspace-root.js` and no `dist/index.js`, which is exactly what the smoke script's changed import needs |
| `release:cli:smoke` | pass — installed the tarball offline and validated bins, Eval assets, native PTY and file locks, the TUI setup path, the managed Runtime Host lifecycle and a controlled model turn |
| `maka --version` / `--help` / `run --help` stderr | 0 bytes, no `ExperimentalWarning` |

Not run: **Desktop e2e**, **Storybook smoke** and the **Linux sandbox smoke**. CI runs all three on Linux under `xvfb`/`bubblewrap`; this PR touches no renderer, Electron or sandbox code, and the desktop files it does touch are import rewrites covered by the 1364 passing desktop tests, `typecheck` and `knip`.

One failure remains, and it is not from this change: `@maka/storage` `managed-dependency-environment-crash` → *rejects a second authority for the same storage root in another process*. That test treats any child stderr as an error and Node prints the SQLite warning there. Verified to fail identically on `23c9214` in a clean worktree, and CI is green on this head with that suite included, so it is this machine only. The branch alters no dependency manifest either — `git diff upstream/main...HEAD -- package.json package-lock.json` is empty.

## Remaining warning sites

The paths that still emit the warning all genuinely use SQLite, so the warning is expected there:

- `maka eval` and `maka activate` — including their `--help`, since the command module loads before argument handling;
- the 29 SQLite-backed entrypoints of `@maka/storage`;
- the processes that actually open databases (the Runtime Host child, the Electron main process).

Both existing warning suppressors (`loadDatabaseSync`, `loadSqliteModule`) are on `main` today and survive this PR unchanged.

## Compatibility — maintainer sign-off wanted

Two deliberate contract changes, called out for an explicit decision rather than buried in the diff:

1. **Removing `.` (and `main`/`types`) is a breaking change** to the package's public surface. `@maka/storage` is not published independently and every consumer lives in this repository, so the blast radius today is zero — but if the package is ever published on its own, this belongs in the release notes.
2. **Fourteen modules drop from publicly reachable to package-private**: `artifact-attachments`, `artifact-store`, `operational-state-backup`, `plan-store`, `provider-request-capture-artifact`, `session-bundle-canonical-tree`, `session-bundle-contract`, `session-bundle-file-service`, `session-bundle-manifest`, `session-bundle-ustar`, `sqlite-artifact-metadata`, `sqlite-usage-store`, `task-ledger-store`, `telemetry-repo`. Nothing outside the package imports them; re-publishing any of them later is a one-line `exports` addition. (`artifact-store` is the one the rebase moved into this list — `sanitizeArtifactName` stays reachable through `./artifact-stores`, as it already was.)
3. **Four symbols the barrel withheld are still not publicly reachable**: `migrateOperationalStateDatabaseInternal`, `inspectOperationalStateSchema` and `OperationalStateMigrationBlockedError`, because `./operational-state-store` is new here and its facade publishes only the barrel's picks; and `createSqliteArtifactStoreWriteAuthority`, because `artifact-store` is not published at all. The reachable-symbol union test names each one. `withCredentialFileLock` is deliberately **not** on this list — see the follow-up below.

## AI use

Select exactly one:

Tool(s) and scope: **Claude Code.** Claude Fable 5 traced the regression to the barrel, performed the mechanical rewrite, wrote the regression tests, ran the measurements and applied the review follow-ups. Claude Opus 5 rebased onto `23c9214`, made the three decisions in the rebase section above, re-ran the verification and updated this description. Each commit carries a `Generated-by: Claude Code` trailer — please keep it in the squash commit.

A human contributor of record has reviewed the diff, verified the evidence above and decided to submit it.

## Checklist

- [ ] No


Astro-Han (commented):
Making the SQLite cost visible at the entrypoint is a real improvement, and `public-entrypoints.test.ts` stating the boundary out loud is the right instinct — a list a reader can check beats a property nobody can see. I also verified the part that is easiest to get wrong and hardest to check: I resolved every one of the 52 published targets against every `@maka/storage` import in the tree, symbol by symbol, and found no missing subpath and no missing export. That work is sound.

Three things below will fail CI as filed, so I cannot approve this head. Two of them are invisible from this diff, which is what makes them worth flagging rather than leaving to a red run.

**The scope question, which is yours to decide rather than mine.** The PR argues that dropping the barrel is what fixes #1257. I do not think that premise holds. There are exactly three static value imports of `node:sqlite` in the tree — `operational-target-schema.ts:1`, `operational-state-backup.ts:16`, `session-bundle-policy.ts:4` — and all three only construct inside function bodies. This package already owns the seam for that, twice: `loadDatabaseSync()` in `operational-state-store.ts` and `loadSqliteModule()` in `sqlite-session-metadata-store.ts`, both lazy-require-plus-suppress-warning. A subagent reports restoring the barrel wholesale, converting those three to `import type` plus the existing loader, rebuilding, and measuring the barrel no longer pulling in `node:sqlite`. If that holds, #1257 is a three-file fix and this is an 87-file one.

That does not make this PR wrong. The two designs buy different things: dropping the barrel makes cost legible at the import site, lazy-loading keeps the change surface at three files. AGENTS.md's first principle points at the existing seam, and the PR body's claim that "no lazy-load indirection and no warning suppression are involved" is not accurate about this repository either way — both suppressors are on `main` today and survive this PR. What I am asking for is that the justification match the actual alternative, so whoever approves this is choosing legibility deliberately rather than believing it was forced.

Two contract decisions a maintainer should sign off on rather than a reviewer: removing `.` is a breaking change to the package's public surface, and eight modules drop from publicly reachable to package-private, which the title does not mention.

AI disclosure: this review was produced with Claude Code (Opus 5) with a subagent covering security, correctness, integration and simplification. I independently re-derived every finding published here: I confirmed the deep `dist/index.js` import in the smoke script and the bare `@maka/storage` specifier on `main` by reading both files, and I reproduced the formatting failure myself by running this repository's pinned Biome over the two files at this head and diffing its output. The subagent additionally reports building the branch and executing the merge; I did not re-run those. Per AGENTS.md this is not independent human review.

packages/storage/package.json
```diff
@@ -7,37 +7,57 @@
   "main": "./dist/index.js",
   "types": "./dist/index.d.ts",
   "exports": {
-    ".": "./dist/index.js",
-    "./credential-store": "./dist/credential-store.js",
-    "./file-update-lock": "./dist/file-update-lock.js",
+    "./activation-secret-injector": "./dist/activation-secret-injector.js",
```
Astro-Han, line 8: [P1] Removing `src/index.ts` breaks `scripts/smoke-release-cli-package.mjs`, which reaches past the export map. Line 706 of that script does `importInstalled(packageRoot, 'node_modules/@maka/storage/dist/index.js')` — a raw `import(pathToFileURL(...))` on a file path, so neither the `exports` map nor typecheck protects it, and deleting `src/index.ts` means `tsc` stops emitting `dist/index.js`. It is called from two scenarios in that script, and the CI path is closed: `scripts/ci-test-plan.mjs` lists `packages/storage` in `CLI_PACKAGE_WORKSPACES`, which sets `cli_package=true`, which runs `npm run release:cli:smoke`. `resolveMakaDataRoots` now lives in `workspace-root.ts`, so the fix is to point that import at `dist/workspace-root.js` — note lines 474, 713 and 780 of the same script already use exactly that deep form. Confirmed by reading the script on `main` and the file removal in this PR; a subagent reports building the workspace and getting `ERR_MODULE_NOT_FOUND` on the resulting `dist/index.js`. Regression test: extend `scripts/release-cli-file-policy.test.mjs` to walk every `importInstalled(..., 'node_modules/...')` literal in the smoke script and assert the target exists under `packages/*/dist`, since this class of import is protected by neither the export map nor the typechecker.

packages/storage/package.json
```diff
@@ -7,37 +7,57 @@
...
+    "./agent-graph-control-store": "./dist/agent-graph-control-store.js",
+    "./agent-run-store": "./dist/agent-run-store.js",
+    "./artifact-store": "./dist/artifact-store.js",
```
Astro-Han: [P1] This branch is stale against `main` in a way that merges cleanly and then fails to build. `#3295` landed on `main` about an hour ago and its consumer imports through the **bare** specifier: `packages/runtime-host/src/server/execution-composition.ts:60` is `import { openStorageWriterComposition } from '@maka/storage';`. That line is new on `main`, does not overlap anything this PR edits, and will therefore merge with no conflict marker anywhere near it — while this PR removes the `.` entry it depends on. The result is `@maka/runtime-host` failing with `ERR_PACKAGE_PATH_NOT_EXPORTED`, from a merge that looked clean. Worth knowing that `main`'s export map already carries `./storage-writer-composition`, so the fix on this side is a one-line specifier change, and the visible conflicts in `packages/storage/package.json` and the `src/index.ts` modify/delete are the ones that will draw attention — not this. Confirmed by reading `execution-composition.ts` and `packages/storage/package.json` on `main` at `a6f33c9`. Also check whether `SQLITE_BACKED_ENTRYPOINTS` needs `storage-writer-composition` added, since it statically imports `./execution-stores.js`. Regression test, and the one that would have caught this without a rebase: assert in `public-entrypoints.test.ts` that no source file in the repository imports or re-exports the bare `@maka/storage` specifier.

packages/storage/src/__tests__/public-entrypoints.test.ts
```diff
@@ -0,0 +1,91 @@
...
+
+test('only the declared entrypoints load node:sqlite', { timeout: 120_000 }, async () => {
+  const exports = await publishedEntrypoints();
+  const results = await Promise.all(
+    Object.entries(exports).map(async ([subpath, target]) => ({
+      subpath,
+      sqlite: await loadsSqlite(target),
+    })),
+  );
+  const actual = results.filter((entry) => entry.sqlite).map((entry) => entry.subpath).sort();
```
Astro-Han: [P1] `npm run format:check` fails on this line, and on `provider-request-capture-artifact.test.ts:13`. Biome wants this chain broken across lines — `const actual = results` then `.filter(...)`, `.map(...)`, `.sort()` each on their own — and wants the `Reflect.get(...)` call in the other file split across three argument lines. `.github/workflows/ci.yml` runs `format:check` whenever `code == 'true'`, which this PR obviously satisfies. Reproduced by execution: I ran this repository's pinned `@biomejs/biome@2.5.6` over both files at `0124d97` via `--stdin-file-path` and diffed its output against the committed bytes; these are the only two differences in either file. The PR's verification table lists `npm run lint` — which I also ran clean — but not `format:check`, and AGENTS.md calls out exactly that gap: workspace typecheck and lint alone do not cover CI's format check. `npm run format` fixes both.

packages/storage/package.json
```diff
@@ -7,37 +7,57 @@
...
+    "./sqlite-runtime-store": "./dist/sqlite-runtime-store.js",
+    "./sqlite-session-metadata-store": "./dist/sqlite-session-metadata-store.js",
+    "./state-root-composition": "./dist/state-root-composition.js",
     "./task-ledger-authority": "./dist/task-ledger-authority.js",
-    "./work-board-store": "./dist/work-board-store.js",
-    "./model-call-ledger": "./dist/model-call-ledger.js",
     "./usage-stores": "./dist/usage-stores.js",
+    "./work-board-store": "./dist/work-board-store.js",
     "./workspace-identity": "./dist/workspace-identity.js",
+    "./workspace-root": "./dist/workspace-root.js",
```
Astro-Han, line 55: [P2] Drop `main` and `types`, or repoint them — they still name a file this PR stops emitting. `"main": "./dist/index.js"` and `"types": "./dist/index.d.ts"` survive the removal of the `.` export. Nothing in this repository breaks, because Node prefers `exports` and every internal consumer resolves through it. The exposure is at publish time: `scripts/release-cli-package.mjs`'s `copyInternalPackage` lists `main` in `allowedFields` and copies it verbatim into the released manifest, so the published `@maka/storage` advertises an entry point that does not exist, and any consumer that does not honour `exports` — an older bundler, some editor resolvers — fails to resolve the package at all. Since the whole point of this PR is that the package's public surface should be exactly what it says, a dangling `main` is the one field that contradicts it. Confirmed by reading the manifest at this head and the release script on `main`. Regression test, which also covers the finding above: assert in `public-entrypoints.test.ts` that every `main`, `types` and `exports` target named by the manifest exists in `dist` after a build.

packages/storage/src/__tests__/managed-workspace-baseline.test.ts
```diff
@@ -61,11 +60,18 @@ test('does not expose baseline receipt issuance on the public Git workspace serv
...
-  assert.equal('inspectManagedWorkspaceExecutionScopeInternal' in publicStorage, false);
+  // The package has no barrel: reachability is decided by the `exports` map.
+  // These internals live in modules it does not publish, so no consumer can
+  // import them at all.
+  const { readFile } = await import('node:fs/promises');
+  const manifest = JSON.parse(
+    await readFile(new URL('../../package.json', import.meta.url), 'utf8'),
+  ) as { exports: Record<string, string> };
+  const published = new Set(Object.values(manifest.exports));
+  assert.equal(published.has('./dist/git-workspace-service.js'), false);
```
Astro-Han: [P2] This assertion is weaker than the one it replaces, not stronger. The previous check asked whether the internal symbol was reachable on the barrel object; this one asks whether the module *file* appears as a target in the export map. Those differ precisely where it matters: if any published entrypoint later gains `export * from './git-workspace-service.js'`, the file still never appears as a target, this assertion still passes, and the internal symbol is public again. The PR body describes it as a stronger claim, which holds for direct exposure but not for re-export — and re-export is the failure mode a barrel-removal PR is most likely to reintroduce. No leak exists today; a subagent reports walking the transitive re-export graph of all 52 entrypoints and finding all three internal modules unreachable. The stronger form is to `import()` each published entrypoint and assert the union of their named exports excludes these symbols — asserting on reachable symbols rather than on module paths.

packages/storage/src/__tests__/public-entrypoints.test.ts
```diff
@@ -0,0 +1,91 @@
...
+];
+
+async function loadsSqlite(target: string): Promise<boolean> {
+  const specifier = pathToFileURL(resolve(packageRoot, target)).href;
+  const { stderr } = await run(
+    process.execPath,
+    ['--input-type=module', '--eval', `await import(${JSON.stringify(specifier)})`],
+    { encoding: 'utf8' },
+  );
+  return stderr.includes('SQLite is an experimental feature');
```
Astro-Han: [P3] Do not detect the SQLite boundary by matching Node's warning text. `stderr.includes('SQLite is an experimental feature')` ties this guard to a string Node owns and has already changed: a subagent reports running this test on Node 26.7.0 and getting `actual: []` — all 29 entries fall through and the test fails outright. `engines` allows `>=22.19.0` and CI runs 24, so this is green today and red for any contributor on a newer Node, and it will silently stop guarding anything the day CI bumps. A `module.registerHooks` resolve hook that observes whether `node:sqlite` enters the module graph is version-independent and tests the actual property; the subagent reports that recomputing the set that way reproduces your 29 entries exactly, which is a good sign for the list itself. Worth fixing here because this test is the entire enforcement mechanism for the boundary this PR exists to establish.

childrentime:
@Astro-Han Thanks for the review — every finding was real. All six are addressed at 2d5c2cf, and I've resolved the threads accordingly:

- Rebased onto `main`; the writer-composition consumer #3295 added now imports through `@maka/storage/storage-writer-composition`, and `public-entrypoints.test.ts` asserts no source file in the repository imports the bare specifier — the test you suggested, and the one that would have caught this without a rebase.
- The smoke script's deep import points at `dist/workspace-root.js`, and `release-cli-file-policy.test.mjs` now walks every `importInstalled(..., 'node_modules/@maka/...')` literal and asserts the build emits the target. (Asserting against `src` twins turned out wrong in one case — `@maka/runtime`'s filesystem worker is emitted by a bundling script, not tsc — so the test checks `dist`, which `check:stale` keeps honest in `check:release`.)
- `main`/`types` are dropped, with a test asserting they stay absent and that every `exports` target is a file the build emits.
- The internal-symbol assertion now loads all 53 published entrypoints and checks the union of reachable symbols against the internals — your re-export scenario fails it.
- The SQLite boundary is observed through a `module.registerHooks` resolve hook; the probe errors loudly if it produces no verdict rather than falling through. `./storage-writer-composition` is entry 30 on the list — you were right that it belongs there, it statically imports `execution-stores` and thirteen other SQLite-backed modules.
- `npm run format` fixed both files; `format:check` is in the validation table now.

On scope: you're right that a three-file lazy-load fix exists, and the PR body now states it explicitly, along with why I still prefer the structural version — #1257 was fixed the small way once and regressed when its guard was deleted in #2710. The breaking `.` removal and the eight modules going package-private are called out in the description under "maintainer sign-off wanted". If the maintainers prefer the minimal fix, I'll close this and submit the three-file version.

AI disclosure: these follow-ups were implemented with Claude Code (Claude Fable 5); I reviewed the diff and the validation results before pushing.

scripts/release-cli-file-policy.test.mjs
```diff
@@ -54,3 +56,37 @@ describe('CLI release file policy', () => {
...
+    );
+    const workspaceDirByPackageName = new Map(
+      readdirSync(join(repoRoot, 'packages')).flatMap((directory) => {
+        const manifestPath = join(repoRoot, 'packages', directory, 'package.json');
+        if (!existsSync(manifestPath)) return [];
+        return [[JSON.parse(readFileSync(manifestPath, 'utf8')).name, directory]];
+      }),
+    );
+    const literals = [
+      ...smokeScript.matchAll(/'node_modules\/(@maka\/[^/']+)\/(dist\/[^']+\.js)'/gu),
```
Astro-Han: [P1] Widen this scan beyond the smoke script — one file it does not cover has a reach-in that this PR breaks. `scripts/computer-use/real-model.mjs:13` still reads `import { createSqliteAgentRunStore } from '../../packages/storage/dist/index.js';`. Deleting `packages/storage/src/index.ts` (removed here, -160) means `tsc` stops emitting `dist/index.js`, so that import throws `ERR_MODULE_NOT_FOUND` at module load and `npm run computer-use -- real-model` — the command `docs/computer-use-evidence-classes.md:53` documents for producing real-model provider evidence — no longer starts. Lines 17 and 18 of the same file already use the surviving deep form, so the author of that file understood the pattern; line 13 is the one that was missed. The correct target is `dist/agent-run-store.js` (`packages/storage/src/agent-run-store.ts:265`). Nothing in the tree catches it: it is a relative path so the `exports` map is bypassed, it is `.mjs` so no tsconfig typechecks it, Knip runs only over `apps/desktop` and `packages/ui`, the new bare-specifier guard in `public-entrypoints.test.ts` matches only the literal `@maka/storage`, and this scan reads exactly one file. Evidence: confirmed by reading code at this head — I grepped the whole repo for `packages/storage/dist/` and this is the only surviving reach-in to a deleted emit. Regression test: replace the single-file read here with a walk over `scripts/**/*.mjs` collecting every `packages/*/dist/*.js` literal as well as the `node_modules/@maka/*/dist/*.js` form, asserting each target exists — that one change covers this and the smoke-script case with the same code.

Astro-Han (commented):
Re-reviewed at `2d5c2cf2d`. I dispatched two scoped sub-reviews — integration and simplification/test-quality; I did not run security or resource-bounds dimensions, because an import rewrite has no meaningful surface for either. Everything below I re-derived at this head myself.

**All five findings from the previous round are fixed.** `execution-composition.ts:60` now imports `@maka/storage/storage-writer-composition`; the smoke script points at `dist/workspace-root.js`; `main`/`types` are gone from the manifest; `managed-workspace-baseline.test.ts` asserts on the reachable-symbol union rather than a weaker property; and the SQLite probe uses `module.registerHooks` instead of text matching. I also re-ran the repo-pinned Biome over every changed TS/JS/MJS file outside the `apps/desktop/**` and `packages/ui/**` exclusions — byte-identical output, so `format:check` passes.

I checked the two things that would make a rewrite of this size unsafe, and both are clean. There are zero remaining bare `@maka/storage` import specifiers anywhere in `packages/`, `apps/` or `scripts/` — the four textual hits in `scripts/` are package-name list entries, not specifiers. And every subpath any file imports resolves against the new `exports` map; the set of used subpaths minus the 53 published keys is empty. The one-intent rule holds too: every hunk across the 89 files is an import rewrite, the manifest, the deleted barrel, or a guard for the barrel's absence.

The P1 below is the same mechanism as the smoke-script finding you already accepted and fixed, one file over — a relative reach-in to a `dist` emit that this PR removes. It is the last one; I grepped the tree to be sure.

Two smaller things that do not warrant their own threads:

`apps/desktop/tsconfig.storybook.json:13` still maps `@maka/storage` to `../../packages/storage/src/index.ts`, a file this PR deletes. That config is live in CI — `apps/desktop`'s `typecheck` runs `tsc -p tsconfig.storybook.json --noEmit` — but it does not fail today, because nothing in its include set imports `@maka/storage`. So it is dead config rather than a break. It is worth deleting in this PR anyway: the mapping is non-wildcard, so it would silently shadow the `exports` map for any future bare import from a story, which is precisely the failure mode this PR exists to remove. The file is not in the diff, hence no inline thread.

The new bare-specifier guard at `public-entrypoints.test.ts:120` is, for `.ts` files, redundant — `moduleResolution: "Bundler"` already honours `exports`, so a bare specifier fails `npm run typecheck` before the guard runs. Its real value is over `.mjs`/`.js`, and there it misses the side-effect form `import '@maka/storage';` as well as the relative reach-ins the P1 describes. The regression test I suggest on that thread fixes both.

Carrying a P1 → COMMENT. The rest is a clean, well-scoped change and I expect to approve it once the `real-model.mjs` import is repointed.

*AI disclosure: reviewed with Claude Code, including two scoped sub-reviews used as leads. Every finding published here I verified at `2d5c2cf2d` myself — I read the deleted barrel's named export lists against the modules they came from, counted external importers per published subpath on the branch, and grepped the tree for surviving `packages/storage/dist/` reach-ins. Sub-review findings I could not confirm are not included.*

packages/storage/package.json
```diff
@@ -4,41 +4,59 @@
...
-    "./encrypted-file-managed-secret-store": "./dist/encrypted-file-managed-secret-store.js",
+    "./goal-authority": "./dist/goal-authority.js",
+    "./interaction-store": "./dist/interaction-store-public.js",
     "./long-term-memory-store": "./dist/long-term-memory-store.js",
+    "./managed-secret-store": "./dist/managed-secret-store.js",
+    "./managed-workspace-owner": "./dist/managed-workspace-owner.js",
+    "./mcp-config-store": "./dist/mcp-config-store.js",
+    "./memory-bundle-store": "./dist/memory-bundle-store.js",
+    "./model-call-ledger": "./dist/model-call-ledger.js",
+    "./operational-state-store": "./dist/operational-state-store.js",
```
Astro-Han: [P2] Narrow this entrypoint, or move the internals behind their owning module — publishing `operational-state-store.ts` wholesale is the one place where dropping the barrel is not surface-neutral. The deleted barrel re-exported a hand-picked four names plus two types from this module (`index.ts:131-140`: `acquireOperationalStateDatabase`, `OPERATIONAL_STATE_DATABASE_NAME`, `OPERATIONAL_STATE_SCHEMA_VERSION`, `resolveOperationalStateDatabasePath`). A whole-module subpath also publishes `migrateOperationalStateDatabaseInternal` (`:407`), `inspectOperationalStateSchema` (`:279`) and `OperationalStateMigrationBlockedError` (`:128`), none of which any entrypoint reached before — the migration function runs `BEGIN IMMEDIATE` and applies five schema migrations against a caller-supplied `DatabaseSync`, and its only in-repo caller is line 260 of its own module. The same shape with a sharper edge sits at `./artifact-store`: the barrel exported six names and deliberately withheld `createSqliteArtifactStoreWriteAuthority` (`artifact-store.ts:231`), whose only caller, `artifact-stores.ts:78`, binds it to a storage-root lease via `assertAuthority` and `leaseBoundWriterLockAuthority`. Both of those options are optional, so an external consumer importing `@maka/storage/artifact-store` can now construct an artifact write authority with no lease at all — exactly the gate `artifact-stores.ts` exists to enforce. Evidence: confirmed by reading code at this head; I diffed the barrel's named export lists against the modules' actual exports. Regression test: extend the reachable-symbol union in `managed-workspace-baseline.test.ts:74-85`, which already loads all 53 entrypoints and has the right shape, to include `migrateOperationalStateDatabaseInternal` and `createSqliteArtifactStoreWriteAuthority` — so any future widening has to be declared rather than inherited.

packages/storage/package.json
```diff
@@ -4,41 +4,59 @@
...
-    "./storage-writer-composition": "./dist/storage-writer-composition.js",
+    "./runtime-event-persistence": "./dist/runtime-event-persistence.js",
     "./runtime-policy-stores": "./dist/runtime-policy-stores.js",
     "./scheduled-task-store": "./dist/scheduled-task-store.js",
+    "./session-bundle-canonical-tree": "./dist/session-bundle-canonical-tree.js",
+    "./session-bundle-contract": "./dist/session-bundle-contract.js",
+    "./session-bundle-file-service": "./dist/session-bundle-file-service.js",
+    "./session-bundle-manifest": "./dist/session-bundle-manifest.js",
+    "./session-bundle-policy": "./dist/session-bundle-policy.js",
+    "./session-bundle-ustar": "./dist/session-bundle-ustar.js",
```
Astro-Han: [P2] Drop the five `session-bundle-*` entrypoints that no consumer reaches — they invert the narrowing this PR is executing. `./session-bundle-canonical-tree`, `./session-bundle-contract`, `./session-bundle-file-service`, `./session-bundle-manifest` and `./session-bundle-ustar` are all added here and imported by zero files outside `packages/storage`; the package's own tests reach them by relative specifier, as the rewritten fixtures in this diff show. `./session-bundle-policy` has exactly one external importer and is the only one of the six that is earned. This is the same rule the PR itself applies to demote other modules to package-private, just not applied here, and each published subpath is a compatibility promise from here on. Nothing will ever flag them, since Knip runs only over `apps/desktop` and `packages/ui`. Evidence: confirmed by reading code at this head — I counted external importers per subpath across the branch. Regression test: assert in `public-entrypoints.test.ts` that every published subpath has at least one importer outside `packages/storage`. That would also surface the five consumer-less entrypoints that predate this PR (`write-queue`, `work-board-store`, `managed-secret-store`, `activation-secret-injector`, `encrypted-file-managed-secret-store`), which are not this PR's problem but are the same class.

packages/storage/src/__tests__/public-entrypoints.test.ts
```diff
@@ -0,0 +1,159 @@
...
+  assert.deepEqual(
+    offenders,
+    [],
+    'bare `@maka/storage` imports resolve to the removed `.` entrypoint and fail at runtime',
+  );
+});
+
+test('only the declared entrypoints load node:sqlite', { timeout: 120_000 }, async () => {
+  const exports = await publishedEntrypoints();
+  const results = await Promise.all(
```
Astro-Han: [P3] Batch these probes rather than spawning all 53 at once. `Promise.all` over `Object.entries(exports)` gives one unbounded `execFile` child per entrypoint under a 120s timeout, while `managed-workspace-baseline.test.ts:69` separately imports all 53 into the test process and the outer `node --test "dist/**/*.test.js"` is already running test files concurrently. AGENTS.md calls out this exact shape ("set `--test-concurrency` to at most 4"). This is a flake budget rather than a correctness bug — the assertions are right and the count is fixed — but it is a large instantaneous load on a CI runner for a test whose value does not depend on running in parallel. Evidence: the concurrency is confirmed by reading code at this head; the CI impact is inference, not measured. Regression test: none needed — batch in groups of 4-8 and the existing assertions still cover the same surface.

childrentime:
@Astro-Han All four findings and both of the no-thread notes are addressed at 049e93f:

- `real-model.mjs:13` points at `dist/agent-run-store.js`, and the file-policy scan now walks every script in `scripts/` for both import forms — installed `node_modules/@maka/*/dist` and relative `packages/*/dist` — asserting each target is still emitted. I verified the guard by reverting the fix locally: it fails on exactly that line.
- The five consumer-less `session-bundle-*` entrypoints are gone; `session-bundle-policy` stays. `public-entrypoints.test.ts` now asserts the exact set of published-but-unconsumed entrypoints, allowlisting only the five that predate this PR (your list matched my recount), and separately that every imported subpath is published. The bare-specifier guard also covers the side-effect form now.
- `operational-state-store`, `artifact-store` — and `credential-store` — publish facades re-exporting exactly the names the barrel picked, following the existing `interaction-store-public` idiom, and the reachable-symbol union test names all five withheld symbols. Credential-store is one you didn't flag: I re-derived the barrel-picks-vs-module-exports diff for every published hand-picked module, and `withCredentialFileLock` was the only other widened symbol — its own doc comment says it is deliberately not on the public surface. Everything else was surface-neutral.
- SQLite probes run at most four concurrent.
- The dead storybook path mapping is deleted (its `tsc -p tsconfig.storybook.json --noEmit` still passes), and knip over `apps/desktop` and `packages/ui` is clean.

Counts in the PR body are updated: 48 entrypoints, 30 SQLite-backed, 18 free, 13 modules package-private.

AI disclosure: implemented and self-reviewed with Claude Code (Claude Fable 5); I reviewed the diff and validation before pushing.

Astro-Han (approved):
Approving at `049e93fdf`. I reviewed only the delta from `2d5c2cf2d` — one commit, nine files — and every finding from the last round is fixed, several of them better than what I asked for.

The P1 is closed at its cause rather than at the symptom. `real-model.mjs:13` now points at `dist/agent-run-store.js`, and `release-cli-file-policy.test.mjs` no longer reads one file for one literal shape: it walks every script for both the `node_modules/@maka/*/dist/*.js` and the relative `packages/*/dist/*.js` forms and asserts each target exists. That closes the whole class, which is what makes it a fix rather than a patch.

The surface-widening P2 got the right treatment. Rather than narrowing by convention, `artifact-store-public.ts`, `operational-state-store-public.ts` and `credential-store-public.ts` re-export exactly the sets the deleted barrel published, and `package.json` repoints the three subpaths at them — I diffed each against the barrel's old export lists and they match name for name. `managed-workspace-baseline.test.ts` then asserts the withheld internals stay unreachable, so the narrowing is enforced rather than documented. You also found `withCredentialFileLock`, which I had missed; that is a fourth internal on the same footing as the three I named.

The consumer-less entrypoints are gone, and the guard that replaces them is stronger than the assertion I suggested: `PREEXISTING_UNCONSUMED_ENTRYPOINTS` is exact in both directions, so gaining a consumer forces removing the entry and publishing a new consumer-less subpath fails outright, with the comment stating the list may only shrink. That converts a one-time cleanup into an invariant. The dead `tsconfig.storybook.json` mapping is deleted, and `mapWithConcurrency` caps the probe children at 4 — the number AGENTS.md actually names.

I re-verified rather than taking the delta on trust: every external `@maka/storage/*` import in this branch falls inside the narrowed surfaces, `./session-bundle-policy` keeps its one real consumer, and the repo-pinned Biome produces byte-identical output for all seven changed TS/JS/MJS files outside the `apps/desktop/**` and `packages/ui/**` exclusions.

No findings at this head.

*AI disclosure: reviewed with Claude Code. I diffed each new `*-public.ts` against the deleted barrel's export lists myself, re-scanned the branch for external imports of the repointed subpaths, and ran the read-only Biome check against the head content. The approval is mine and rests on those checks.*

Astro-Han:

- `apps/desktop/src/main/client-settings-effects.ts`
- `apps/desktop/src/main/mcp-ipc-main.ts`
- `apps/desktop/src/main/new-session-project.ts`
- `apps/desktop/src/main/runtime-host-boot.ts`
- `…(more)`

```
# resolve, then
git push --force-with-lease
```

---

zhiiw (approved):
Incremental re-review at exact head `d35c1c3615707bdfeb9786e469ab637b1af7892e` (previous approval was on `049e93fd`; this pass covers only what changed between them, plus re-verifying the load-bearing gate).

**The boundary is enforced, not transcribed.** `public-entrypoints.test.ts` does not pin a copied count: `only the declared entrypoints load node:sqlite` loads *every* published entrypoint in a child process with a `module.registerHooks` resolve hook and asserts the measured set equals the declared list exactly, both directions. A pure-type entrypoint that accidentally drags in `node:sqlite` tomorrow fails this test. `published entrypoints and their consumers match exactly` separately fails on any new consumer-less entrypoint. I ran the suite at this head: 5/5 green.

**Import hygiene:** scanned every published entrypoint's source for re-exports — all within-package, or explicit Core type flows (`deep-research-store` → `@maka/core/deep-research-run`, `project-catalog` → `@maka/core/project`). No consumer resolves a type through a package that doesn't own it. Residual honesty: a *semantically* wrong-but-typecheck-compatible import can't be ruled out mechanically; I found none in the changed files.

**Number note:** the first commit's message says "exactly 28 of the 54 published entrypoints load `node:sqlite`". At this head the map publishes 49 entrypoints and the declared SQLite set holds 29 — the message numbers describe the first commit's state, not the head's. The test is the authority and it is enforced; the prose is stale. Cosmetic, not blocking.

**Gate:** `package` / `audit` / `test` all `completed/success` on the exact head; head re-polled at review time, unchanged, MERGEABLE.

No new findings in the incremental surface.

<details>
<summary>简体中文</summary>

增量复审：边界测试是真门禁（子进程实测每个 entrypoint 是否加载 node:sqlite，与声明列表双向精确相等），不是把现状抄成期望值；我本地在 exact head 跑了 5/5 绿。重导出扫描无跨包类型偷渡。首条提交信息里的 28/54 数字是旧状态的快照，与 head（49/29）不符——纯文案陈旧，不阻断。增量面无新 finding。

</details>

M4n5ter (commented):
I reviewed exact head `d35c1c3615707bdfeb9786e469ab637b1af7892e`. The root-barrel removal is structurally sound: runtime reachability contracts from 272 exported names to 216, the automatic root-to-SQLite evaluation path disappears, the dependency-cycle count does not increase, and a clean synthetic merge with current `main` passes the Storage, Desktop, CLI, entrypoint, and release-policy checks I ran.

One existing subpath is silently contracted, so my result is **COMMENT with 1×P2**.

<details>
<summary>简体中文</summary>

我审查了精确提交 `d35c1c3615707bdfeb9786e469ab637b1af7892e`。删除根 barrel 的整体结构是成立的：运行时可达导出从 272 个收缩到 216 个，根入口自动加载 SQLite 的路径消失，依赖环数量没有增加；与当前 `main` 的干净合并结果也通过了我运行的 Storage、Desktop、CLI、入口点及发布策略检查。

但一个既有子路径被静默收缩，因此本次结论为 **COMMENT，1×P2**。

</details>

packages/storage/package.json
```diff
@@ -4,43 +4,55 @@
...
-    "./credential-store": "./dist/credential-store.js",
-    "./file-update-lock": "./dist/file-update-lock.js",
-    "./process-lifetime-file-update-lock": "./dist/process-lifetime-file-update-lock.js",
-    "./stable-storage": "./dist/stable-storage.js",
+    "./activation-secret-injector": "./dist/activation-secret-injector.js",
+    "./agent-graph-control-store": "./dist/agent-graph-control-store.js",
+    "./agent-run-store": "./dist/agent-run-store.js",
     "./artifact-stores": "./dist/artifact-stores.js",
+    "./config-transfer": "./dist/config-transfer.js",
+    "./credential-store": "./dist/credential-store-public.js",
```
M4n5ter: **[P2] Preserve or explicitly retire `withCredentialFileLock` from this existing entrypoint**

`./credential-store` was already a published subpath before this PR and mapped directly to `credential-store.js`. Importing `@maka/storage/credential-store` on the parent commit exposes `withCredentialFileLock` as a function; at this head the same import resolves through the new facade and the symbol is `undefined`. This contract change is independent of deleting the root barrel.

The new guards do not catch it: the reachable-symbol denylist expects this symbol to be absent, while the unconsumed-subpath guard checks path presence rather than symbol compatibility. Please either preserve the old target/symbol for this pre-existing entrypoint, or explicitly record and obtain agreement for retiring it as a compatibility change.

<details>
<summary>简体中文</summary>

**[P2] 保留或明确退役该既有入口中的 `withCredentialFileLock`**

`./credential-store` 在本 PR 之前已经是发布的子路径，并直接指向 `credential-store.js`。在父提交上导入 `@maka/storage/credential-store` 时，`withCredentialFileLock` 是函数；在当前提交上，同一导入经过新 facade 后该符号为 `undefined`。这项契约变化与删除根 barrel 无关。

现有守卫拦不住它：可达符号 denylist 本身要求该符号不存在，而未消费子路径守卫只检查路径是否存在，不检查符号兼容性。请保留这个既有入口原来的目标/符号，或者将其作为兼容性变更明确记录并取得维护者认可。

</details>

childrentime: Fixed at `a2c64e3` — preserved rather than retired.

You're right that this is independent of removing the barrel, and my description had it backwards: I listed `withCredentialFileLock` among "symbols the barrel deliberately withheld", but the barrel is not what made it reachable. `./credential-store` published the whole module and predates this PR, so the symbol was public regardless of what the barrel picked.

The facade was applying the barrel's picks to an entrypoint the barrel never owned. That rule belongs to the subpaths this branch publishes for the first time, where the surface is still being chosen; it is not a reason to narrow one that already shipped. So `./credential-store` points back at `credential-store.js`, the facade is deleted, and `withCredentialFileLock` also leaves the reachable-symbol denylist in `managed-workspace-baseline`, since it is publicly reachable again. Verified at this head: `typeof withCredentialFileLock === 'function'` through the subpath, as on the parent commit.

On the guard gap you named — it is real, and I did not close it. Catching this class properly needs a symbol-level baseline built from the previous commit, which is more machinery than this PR should carry. What changed instead is that the weaker property is now trivially checkable, because the export map only drops `.` and adds subpaths, with no pre-existing entrypoint retargeted:

```
git diff upstream/main...HEAD -- packages/storage/package.json
```

Whether the file lock should become package-private is still open — worth its own change if you want it, and it would then need the compatibility note this one no longer does.

Re-ran after the fix: storage 925 (910 pass, 14 skipped, 1 pre-existing `managed-dependency-environment-crash` failure that reproduces on `main` locally and passes in CI), `lint`, `format:check`, `typecheck` and `check:asf-headers` all pass.

M4n5ter (approved):
Re-reviewed exact head `a2c64e319d739d1c37f7d4a70bf4d073c794f15b`. The previous credential-store compatibility finding is closed: the pre-existing `./credential-store` target is restored, `withCredentialFileLock` is again reachable as a function, the narrowing facade is deleted, and the baseline test no longer encodes the incompatible contraction. I reviewed the complete one-commit delta and found no new issue.

The validated merge with `main` at `f7d4957fe866c303588355bfa7c069f26325a691` passed the Storage entrypoint and boundary checks (27/27), direct subpath import, and the full Storage suite apart from one assertion coupled to Node's experimental-warning stderr; that assertion passes in isolation with warnings suppressed. Current `main` has since advanced only in four unrelated UI/CLI files, with no overlap with this PR's Storage changes, and GitHub reports the PR mergeable. Exact-head hosted `package`, `audit`, and `test` checks are all terminal green.

No remaining P0–P3 finding.

<details>
<summary>简体中文</summary>

已复审精确提交 `a2c64e319d739d1c37f7d4a70bf4d073c794f15b`。此前 credential-store 兼容性问题已经关闭：既有 `./credential-store` 入口恢复到原目标，`withCredentialFileLock` 再次以函数形式可达，收窄 facade 已删除，基线测试也不再把不兼容的收缩编码为期望行为。我完整审查了这一个新增提交，没有发现新问题。

与 `main` 的已验证合并结果通过了 Storage 入口和边界检查（27/27）、直接子路径导入及完整 Storage 测试；其中只有一条依赖 Node 实验性警告 stderr 的断言失败，该断言在屏蔽警告后单独运行通过。此后当前 `main` 只改动了 4 个无关的 UI/CLI 文件，与本 PR 的 Storage 变更没有重叠，GitHub 也报告可合并。精确 head 的托管 `package`、`audit`、`test` 均已终态成功。

没有剩余的 P0–P3 问题。

</details>

M4n5ter (approved):
Verified at exact head fbb2a5561c98ea69bfef89047ac10f66240f7563 after rebasing onto the updated main history. The conflict resolution preserves main's preferred-path project-catalog registration coverage while retaining this PR's narrow Storage subpath imports. Storage passed 945 tests (929 passed, 16 skipped), the focused Storage entrypoint/boundary suite passed 27/27, the Runtime Host conflict-focused suite passed 3/3, and the Storage and Runtime Host builds completed successfully. All three exact-head hosted checks are terminal and successful, the current main check is terminal and successful, the current merge tree is clean, and there are no unresolved review threads.

Issue #2188, opened 2026-08-04
tracking(workbar): replace the hand-written tab strip with TabList, and close the panel gaps that are left


Astro-Han:
## Why this issue is being rewritten

The original body was written just after #2158 and described the workbar as "eight apps in a tab strip". Most of what it asked for has since landed through the panel PRs, so I re-checked every work item against `main` before rewriting it. Six of the eleven items were already done or rested on a premise that no longer held, so each of them would have opened as an empty PR.

**Updated after #4789.** PR 1 has landed and its four measured defects are closed. PR 2 shrank: the Task face is retired, so half of it no longer exists. PR 4's floor moved with the panel's. What is left is below, with the measurements that back it.

Everything marked *measured* comes from a live renderer or the built Storybook rather than from reading the CSS.

## Problem

The workbar column is composed from Astryx primitives — panels already use `Section`, `Banner`, `EmptyState`, `Toolbar`, `Collapsible`, and the resize edges are Astryx `ResizeHandle`. The tab strip was the exception: `.maka-workbar-tab*` in `styles/workbar/shell.css` plus roughly 200 lines of `workbar-surface.tsx` re-implemented a tab strip that `@astryxdesign/core/TabList` already ships, and four measured defects were all consequences of that re-implementation.

That is done — see **PR 1** below for what each defect measured before and after. The panel gaps that remain are the ones this tracker still exists for.

## Goal

One tab strip, owned by `TabList`, and no panel left drawing chrome the design system already ships.

This supersedes the direction recorded as A3 in `docs/frontend-architecture-astryx-review-2026-08-09.md` ("keep custom tab strip (dnd + `role=tab` is justified)"). Both halves of that justification are gone: `role=tab` is `TabList`'s, and drag-to-reorder was dropped rather than rebuilt — faces are opened from one menu and the strip's order is the order they were opened in.

## Plan

Three PRs left. They open flat off `main` and touch disjoint files, so none blocks another. Each carries BEFORE/AFTER Storybook captures (light and dark) from the `Product/Session Workbar` story group.

### PR 1 — Shell: the tab strip becomes `TabList` — **done (#4789)**

`workbar-surface.tsx`, `styles/workbar/shell.css`, `workbar-toggle.tsx`. `WorkbarTabStrip` and `SortableWorkbarTab` are gone, `.maka-workbar-tab*` with them, and dnd-kit left the dependency tree. Net deletion.

[...]

PR #3305, opened 2026-08-20
refactor(desktop): align browser panel with Astryx

orangeCatDeveloper:
## Summary

Refs #2188

First panel interior from the #2188 list. Auditing the browser chrome against the design system found three defects, fixed as one pass:

- The address bar never stretched (~154px at every width): the flex rule targeted the panel's own wrapper, not the Astryx Toolbar slot.
- The empty state restated `EmptyState` defaults as magic numbers (32px/39px, a 768px breakpoint used nowhere else). Deleted.
- `favicon`/`secure`/`title` crossed IPC with no consumer. `favicon` is removed end to end (the renderer CSP can never load it); `secure` now shows a warning status on http pages; `title` feeds the panel's `aria-label`.

New stories pin every state the E2E fixture cannot reach — empty, loading, loaded, insecure — at the 320px floor, 400px, and the sub-990px stacked layout, mounted in the real AppShell grid.

Behavior change: the http warning and the title-bearing accessible name above; the empty-state Globe icon also moves from `--muted-foreground` to the inherited foreground, converging with the terminal/artifacts empty states.

## Verification

lint · format:check · build · typecheck · knip (desktop, ui) pass; browser main-process tests 34/34; Storybook render smoke 143 stories; `session-workbar.spec.ts` (4/4) including a new width-tracking assertion verified to fail with the slot rule removed; `audit-alignment.mjs` clean including the new `turn-narrative-browser` fixture.

```
before  browser-loaded          addressInput: 154
before  browser-at-400          addressInput: 154
before  browser-at-column-floor addressInput: 154

after   browser-loaded          addressInput: 310
after   browser-at-400          addressInput: 230
after   browser-at-column-floor addressInput: 150
```

|  | before | after |
| --- | --- | --- |
| 480px column | before-browser-loaded | after-browser-loaded |
| http page | before-browser-insecure | after-browser-insecure |
| 400px | before-browser-at-400 | after-browser-at-400 |

At the 320px floor the field yields to the four controls either way (154px → 150px), so before/after are visually identical there by design.

## AI use

Select exactly one:

Tool(s) and scope: Claude Code (Opus) — panel/CSS changes and stories, human-directed and reviewed.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/src/renderer/styles/chat-detail.css
+++ b/apps/desktop/src/renderer/styles/chat-detail.css
@@ -930,12 +930,11 @@
   border-bottom: var(--border-width-hairline) solid var(--border);
 }
 
-.maka-browser-toolbar-start {
-  display: flex;
-  min-width: 0;
+/* Astryx sizes its toolbar slots to content; only the product knows the address
+   field, not the nav buttons, should eat the free width. */
+.maka-browser-toolbar div:has(> .maka-browser-address-field) {
   flex: 1 1 auto;
-  align-items: center;
-  gap: var(--space-1);
+  min-width: 0;
 }
 
 .maka-browser-address-field {
@@ -944,20 +943,9 @@
 }
 
 .maka-browser-strip {
-  position: relative;
   flex: 1 1 auto;
   min-height: 0;
-}
-
-.maka-browser-empty {
-  position: absolute;
-  inset: 0;
-  padding-block: 32px;
-  color: var(--muted-foreground);
-}
-
-@media (min-width: 768px) {
-  .maka-browser-empty { padding-block: 39px; }
+  display: grid;
 }
 
 .maka-artifact-pane {
--- a/apps/desktop/stories/session-workbar.stories.tsx
+++ b/apps/desktop/stories/session-workbar.stories.tsx
@@ -1,5 +1,7 @@
+import type { CSSProperties } from 'react';
 import type { Meta, StoryObj } from '@storybook/react-vite';
 import type { ArtifactRecord } from '@maka/core/artifacts';
+import type { BrowserState } from '@maka/core/browser';
 import type { GitReviewSnapshot } from '@maka/core/git-review';
 import type { SessionSummary } from '@maka/core/session';
 import type { Task } from '@maka/core/task-ledger';
@@ -37,7 +39,31 @@ import { withScopedMakaBridge } from './maka-bridge';
 // 1280, above it.
 
 const SESSION_ID = 'session-workbar';
+const STACKED_WINDOW_VIEWPORT = {
+  makaStackedWindow: {
+    name: 'Maka window below the 990px stack point',
+    styles: { width: '900px', height: '900px' },
+    type: 'desktop' as const,
+  },
+};
 const NOW = Date.UTC(2026, 6, 31, 10, 30, 0);
+const EMPTY_BROWSER_STATE: BrowserState = {
+  url: '',
+  title: '',
+  canGoBack: false,
+  canGoForward: false,
+  loading: false,
+  secure: false,
+  hasPage: false,
+};
+const LOADED_BROWSER_STATE: BrowserState = {
+  ...EMPTY_BROWSER_STATE,
+  url: 'https://maka.apache.org/docs/getting-started',
+  title: 'Getting started — Apache Maka',
+  canGoBack: true,
+  secure: true,
+  hasPage: true,
+};
 const TOOL_PICKER_SOURCE_SESSION: SessionSummary = {
   id: SESSION_ID,
   name: '工作栏组件审查',
@@ -515,7 +541,9 @@ function bridge(options: {
   /** The context snapshot the composition block reads (#2323). */
   context?: ContextDiagnosticsResult;
   recordFilePath?: string;
+  browserState?: BrowserState;
 } = {}) {
+  const browserState = options.browserState ?? EMPTY_BROWSER_STATE;
   return withScopedMakaBridge({
     tasks: {
       list: async () => {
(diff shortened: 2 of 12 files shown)


apps/desktop/src/renderer/styles/chat-detail.css
```diff
@@ -930,12 +930,11 @@
   border-bottom: var(--border-width-hairline) solid var(--border);
 }
 
-.maka-browser-toolbar-start {
-  display: flex;
-  min-width: 0;
+/* Astryx sizes its toolbar slots to content; only the product knows the address
+   field, not the nav buttons, should eat the free width. */
+.maka-browser-toolbar div:has(> .maka-browser-address-field) {
```
Astro-Han, line 935: [P2] Nothing pins the fix this PR exists to make. This rule reaches into Astryx's internals: `Toolbar` renders `startContent` into an unclassed slot div, so the selector holds only while `.maka-browser-address-field` remains a *direct* child of that slot. Astryx already wraps some children in `display: contents` shells — `Tooltip` does exactly this — so an upstream slot-wrapper change would silently return the field to its content width with nothing failing anywhere. And no lane measures it: `storybook-visual-smoke.mjs` forces 1280x900, disables `play`, and asserts only that the root is non-empty with no console error; no Playwright spec opens this panel; and `audit-alignment.mjs` never sees this row because the only fixture that sets a workbar tab, `e2e-fixture.ts:142`, selects `'tasks'`. So the PR checklist's claim that tests cover the change and fail without it does not hold for the stretch fix — the new stories are render-smoke only. A subagent reports reconstructing the Astryx DOM in Chromium and measuring 328/248/168px at 480/400/320px with this rule against a flat 147px without it, matching the numbers in your PR body, so the fix itself is not in question. Adding `workbarTab: 'browser'` to a fixture scenario and asserting the input's width tracks the column in `session-workbar.spec.ts` is one line of fixture plus one assertion — and it also puts this four-control-plus-input row under `alignment_audit`, which is the lane for exactly the defect class this PR fixes.

apps/desktop/stories/session-workbar.stories.tsx
```diff
@@ -558,22 +586,47 @@ function bridge(options: {
         snapshot: gitReviewSnapshot,
       }),
     },
+    browser: {
```
Astro-Han, line 589: [P3] These stories throw on unmount. `withScopedMakaBridge` deletes `window.maka` in a `useLayoutEffect` cleanup, while `BrowserPanel`'s viewport effect calls `window.maka.browser.setViewport({ rect: null })` in a `useEffect` cleanup. Layout-effect destroys run in the commit phase and passive destroys after it, so the bridge is always gone first and the panel dereferences `undefined`. These are the first stories to mount a component that touches `window.maka` at teardown, which is why the PR surfaces it rather than causes it. No user impact — preload never deletes the bridge — and CI never trips it, because the smoke harness loads one story per page and closes it. But switching stories or hot-reloading in the Storybook UI produces `TypeError: Cannot read properties of undefined (reading 'browser')`, which is exactly the workflow these stories exist to support. A subagent reports reproducing it under React 19.2.8 with `createRoot`, arriving as an uncaught `pageerror`. Capturing `const { browser } = window.maka` in the effect body and using that in the cleanup fixes it, and is more correct in the app too.

apps/desktop/stories/session-workbar.stories.tsx
```diff
@@ -641,6 +694,56 @@ export const TasksLoadFailed: Story = {
...
+export const BrowserAt400: Story = {
+  decorators: [bridge({ browserState: LOADED_BROWSER_STATE })],
+  render: () => <Workbar tab="browser" width={400} />,
+};
+
+// Below 990px the grid stacks the same right-placement column under the
+// conversation: full width, capped at 42dvh.
+export const BrowserStacked: Story = {
+  parameters: { viewport: { options: STACKED_WINDOW_VIEWPORT } },
+  globals: { viewport: { value: 'makaStackedWindow', isRotated: false } },
```
Astro-Han, line 743: [P3] `BrowserStacked` does not render stacked under `smoke:storybook`. In Storybook 10 the viewport preview annotation registers only `initialGlobals`; the manager tool is what actually resizes the iframe, and `storybook-visual-smoke.mjs` loads `iframe.html` directly at 1280x900 — above the 990px stack point. So in CI this story renders identically to `BrowserLoaded`. It still demonstrates the stacked layout for a human browsing Storybook, which is most of its value; the comment saying it pins the behaviour is what overstates it. Confirmed by reading the smoke harness and the Storybook viewport annotation.

apps/desktop/src/renderer/styles/chat-detail.css
```diff
@@ -944,20 +943,9 @@
...
-.maka-browser-empty {
-  position: absolute;
-  inset: 0;
-  padding-block: 32px;
-  color: var(--muted-foreground);
-}
-
-@media (min-width: 768px) {
-  .maka-browser-empty { padding-block: 39px; }
+  display: grid;
```
Astro-Han, line 948: [P3] One undisclosed visual delta rides along with this hunk. The deleted `.maka-browser-empty` block also carried `color: var(--muted-foreground)`, and Astryx's `EmptyState` sets colors on the title and description but renders the icon in a bare `<div aria-hidden>` that inherits — so the `Globe` icon moves from `--muted-foreground` to the inherited `--foreground`. I read this as an improvement, since it converges with the sibling terminal, projects and artifacts empty states, which all pass icons with no color wrapper. Raising it only because the PR body enumerates its behaviour changes and this one is not among them, and no before/after screenshot covers the empty state. Confirm it is intended and it needs no code change.

Astro-Han (commented):
This one is sound. The three defects it claims are real, the fixes do what they say, and it is a net deletion in the stylesheet — 5 added lines against 17 removed, replacing hand-rolled positioning and spacing with the design system's own primitives. I found no P0 or P1.

Worth recording what I checked rather than assumed, since token-alignment PRs are easy to wave through: the tokens are the right ones, not lookalikes. Dropping `.maka-browser-toolbar-start`'s `gap: var(--space-1)` is safe because maka's `--space-1` is 4px and Toolbar's default `gap={1}` resolves to Astryx's `--spacing-1`, also 4px, with no override in the maka theme. The removed `padding-block: 32px` is exactly `EmptyState`'s `--spacing-8` default. The only genuine loss is the `@media (min-width: 768px)` bump to 39px, which keyed a panel-width concern to window width — losing it is a fix, not a regression. Replacing absolute positioning with a single-child grid is also equivalent, since `EmptyState`'s own centering still applies.

Accessibility holds up too: `statusVariant="tooltip"` renders a real focusable button whose message is wired to the input through `aria-describedby`, Toolbar's caret guard releases arrow keys at the field boundary so it is not a keyboard trap, and `warning` correctly does not set `aria-invalid` on an address bar. Localization is complete in both zh and en, including the new status button's accessible name, which already resolves through the existing Astryx override in `packages/ui/src/astryx-i18n.tsx` — Astryx ships no zh catalog, so that would otherwise have leaked English.

The findings below are all P2 or P3 and none of them blocks. I am leaving this as COMMENT rather than approving for one reason: the checklist claims tests cover the change and fail without it, and for the stretch fix that is not true at this head — the new stories are render-smoke only. That is a small, mechanical thing to resolve, and I would rather it be resolved than discovered by the next reviewer. Fix the claim or add the fixture line and I expect this to be approvable.

AI disclosure: this review was produced with Claude Code (Opus 5) with a subagent covering security, correctness, integration and simplification. The subagent reports reconstructing the Astryx DOM and StyleX declarations in Chromium to measure the selector's effect, and reproducing the story unmount error under React 19.2.8; I did not re-run those, so they are its evidence rather than mine. I confirmed the file list, the anchors and the CI lane coverage myself. Per AGENTS.md this is not independent human review.

orangeCatDeveloper:
All four addressed in 61d24f9d1: the stretch rule is now pinned by a Playwright width assertion (verified to fail with the rule removed) plus a `turn-narrative-browser` fixture under `alignment_audit`; the viewport cleanup captures the bridge; the `BrowserStacked` comment now says it is Storybook-UI only; and the icon-color delta is intended and now listed in the description.

Astro-Han (approved):
APPROVE.

All four findings from my previous review are closed at `61d24f9d1`, and I checked each one rather than taking the delta at its word:

- **P2, nothing pinned the fix.** Now pinned behaviourally: `session-workbar.spec.ts` drives the panel to 480px and 320px and asserts the address input crosses 250px and drops below 220px. That measures the outcome the slot rule exists to produce, so an upstream Astryx slot-wrapper change breaks the test rather than the product silently. Both widths sit inside the panel's `min-width: 320px` / `max-width: 600px` clamp (`chat-detail.css:86-90`), and the panel is `box-sizing: border-box` with no border, so `toBeCloseTo(width, 0)` is exact rather than a rounding coin-flip. The new `turn-narrative-browser` fixture and its `audit-alignment.mjs` entry are a real addition on top of what I asked for.
- **P3, the stories threw on unmount.** Hoisting `const { browser } = window.maka` to effect setup is the right fix and it is correct for the ordering that caused it: the decorator's layout-effect cleanup can only run after the panel's passive effect has already captured the reference.
- **P3, `BrowserStacked` does not stack under `smoke:storybook`.** Documented at the story rather than papered over, which is what I suggested — the story still earns its place as a Storybook-UI check.
- **P3, the undisclosed Globe colour delta.** Now stated in the Summary as a behaviour change, with the rationale that it converges with the terminal and artifacts empty states.

`biome format` is clean on all six files in the delta.

Nothing further from me.

AI disclosure: this review was assisted by Claude (Opus) for code search and cross-checking. Everything above I re-derived myself against the source at `61d24f9d1`.

PR #2641, opened 2026-08-10
fix(desktop): read Usage from Runtime Host

me2seeks:
## Summary

- move Settings → Usage off the legacy session scanner and onto the reconnectable Runtime Host client
- preserve missing token and pricing basis through the Host protocol, then show clean known totals with one coverage notice and precise request-row labels
- keep aborted tool calls separate from successful calls and remove the legacy Storage read model

The Desktop adapter uses one explicit time window and retries when paged Host projections do not describe the same snapshot. Coverage stays page-level because legacy, unreadable, and pending records cannot be attributed to individual provider or model buckets without guessing.

Fixes #2128

## Verification

- `npm run lint`
- `npm run format:check`
- `npx knip --workspace apps/desktop`
- `npm --workspace @maka/core test` (821 passed)
- `npm --workspace @maka/storage test` (795 passed, 16 skipped)
- `npm --workspace @maka/runtime-host test` (838 passed)
- `npm --workspace @maka/desktop test` (1158 passed)
- `npm --workspace @maka/desktop run build:renderer`
- Storybook render smoke (129 stories), including incomplete, narrow, detailed, tool, and legacy Usage states

### Request log

Usage request log with one coverage notice and precise missing, partial, and unpriced labels

### Provider totals

Usage provider totals showing clean known subtotals with a page-level coverage notice

### Tool totals

Usage tool totals showing aborted calls separately

## Checklist

- [x] Tests cover the change and fail without it
- [x] Lint, format, typecheck and the affected suites pass locally

Does this PR entail a change in behavior?

- [x] Yes — described under Summary above
- [ ] No

<details>
<summary>简体中文</summary>

## 摘要

- 将“设置 → 使用统计”从旧的 Session 扫描器切换到可重连的 Runtime Host client
- 在 Host 协议中保留 Token 与价格缺失的依据，以干净的已知数值配合一条覆盖情况提示展示，并在请求明细中准确区分未上报、部分数据、未定价和未记录
- 将中断的工具调用与成功调用分开，并删除旧的 Storage 读模型

Desktop adapter 使用同一个明确时间窗口；如果分页读取到的 Host 投影不属于同一快照，则会重新读取，而不是拼接后返回。覆盖情况保持在页面级，因为历史、无法读取和待修复的记录不能在不猜测的情况下归到某个供应商或模型。

## 验证

- Core、Storage、Runtime Host 与 Desktop 全量测试通过
- lint、format、Desktop knip、typecheck 与 renderer build 通过
[...]

PR #3309, opened 2026-08-20
feat(core,runtime,desktop): add Fast mode for OpenAI Responses relays

hbw00111:
## Summary

Adds an optional Fast mode to custom OpenAI relay profiles. It maps to OpenAI `service_tier: "fast"` on the Responses wire.

Fast is deliberately limited to the OpenAI Responses adapter and OpenAI-named model families that the pinned SDK recognizes for priority processing. Arbitrary relay model IDs and OpenAI Chat-compatible relays do not receive a misleading option that the SDK would strip.

Refs #3300

## Verification

- core and runtime builds passed
- focused provider-options tests passed
- real Responses request-body capture asserts `service_tier: "fast"`
- arbitrary relay model IDs are covered as unsupported and emit no Fast option
- full repository build remains blocked by existing `packages/ui` type drift

## AI use

Tool(s) and scope: Codex performed codebase research, implementation, test authoring, and verification.

## Checklist

- [x] Lint, format, typecheck and affected suites pass locally

- [ ] No

### DIFF
--- a/apps/desktop/src/renderer/settings/provider-connection-detail.tsx
+++ b/apps/desktop/src/renderer/settings/provider-connection-detail.tsx
@@ -15,6 +15,7 @@ import { isRelayProviderType, PROVIDER_DEFAULTS } from '@maka/core/llm-connectio
 import {
   DECLARABLE_RELAY_THINKING_LEVELS,
   THINKING_LEVELS,
+  supportsRelayFastServiceTier,
   type RelayModelProfile,
   type ThinkingLevel,
 } from '@maka/core/model-thinking';
@@ -152,6 +153,7 @@ function ConnectionDetailInner(props: ConnectionDetailProps) {
     setDraftThinkingLevels,
     setDraftVision,
     setDraftContextWindow,
+    setDraftServiceTier,
     saveRelayProfiles,
     runTest,
     refreshModels,
@@ -526,6 +528,11 @@ function ConnectionDetailInner(props: ConnectionDetailProps) {
                       ? 'disabled'
                       : 'auto';
                 const draftLevels = declared?.thinkingLevels ?? [];
+                const serviceTierValue = declared?.serviceTier ?? 'auto';
+                const showsFastMode = supportsRelayFastServiceTier(
+                  connection.providerType,
+                  modelId,
+                );
                 // The menu offers the five declarable levels PLUS anything
                 // the stored table already claims — a level saved while it
                 // was still declarable (or hand-written into the document)
@@ -617,6 +624,25 @@ function ConnectionDetailInner(props: ConnectionDetailProps) {
                         }
                       />
                     </CapabilityRow>
+                    {showsFastMode && (
+                      <CapabilityRow label={copy.fastMode} description={copy.fastModeHelp}>
+                        <Selector
+                          label={`${copy.fastMode} — ${modelId}`}
+                          isLabelHidden
+                          size="sm"
+                          width={132}
+                          options={[
+                            { value: 'auto', label: copy.fastAuto },
+                            { value: 'fast', label: copy.fastEnabled },
+                          ]}
+                          value={serviceTierValue}
+                          onChange={(value) =>
+                            setDraftServiceTier(modelId, value === 'fast' ? 'fast' : undefined)
+                          }
+                          isDisabled={allActionsBusy}
+                        />
+                      </CapabilityRow>
+                    )}
                   </VStack>
                 );
               })}
--- a/apps/desktop/src/renderer/settings/provider-connection-detail.tsx
+++ b/apps/desktop/src/renderer/settings/provider-connection-detail.tsx
@@ -11,7 +11,7 @@ import {
   Text,
   VStack,
 } from '@astryxdesign/core';
-import { PROVIDER_DEFAULTS } from '@maka/core/llm-connections';
+import { isRelayProviderType, PROVIDER_DEFAULTS } from '@maka/core/llm-connections';
 import {
   DECLARABLE_RELAY_THINKING_LEVELS,
   THINKING_LEVELS,
@@ -158,12 +158,12 @@ function ConnectionDetailInner(props: ConnectionDetailProps) {
     remove,
     refreshAfterRelogin,
   } = useConnectionDetail(props);
-  // Capability switches only exist for openai-compatible relays: built-in
+  // Capability switches only exist for custom OpenAI relays: built-in
   // providers declare their thinking support in model metadata, a custom
   // relay's backing model is unknown until the user says what it can do. The
   // declaration is per model — a relay can front both a reasoner and a plain
   // instruct model.
-  const showsCapabilities = connection.providerType === 'openai-compatible';
+  const showsCapabilities = isRelayProviderType(connection.providerType);
   // Rows are the enabled models, exactly — the store prunes a model's profile
   // the moment it is disabled, so no declaration can ever belong to a row
   // this list does not show. The editor edits the per-model draft; 保存
--- a/packages/core/src/model-thinking.ts
+++ b/packages/core/src/model-thinking.ts
@@ -123,6 +123,8 @@ export interface RelayModelProfile {
   readonly thinkingLevels?: readonly ThinkingLevel[];
   readonly vision?: boolean;
   readonly contextWindow?: number;
+  /** Use OpenAI's low-latency service tier for this relay model. */
+  readonly serviceTier?: 'fast';
 }
 
 export type RelayModelProfiles = Readonly<Record<string, RelayModelProfile>>;
@@ -137,6 +139,7 @@ function normalizeRelayModelProfile(entry: unknown): RelayModelProfile | undefin
     thinkingLevels?: readonly ThinkingLevel[];
     vision?: boolean;
     contextWindow?: number;
+    serviceTier?: 'fast';
   } = {};
   if (Array.isArray(entry.thinkingLevels)) {
     // Declared levels are filtered to the declarable vocabulary, not merely
@@ -168,6 +171,7 @@ function normalizeRelayModelProfile(entry: unknown): RelayModelProfile | undefin
   ) {
     declared.contextWindow = entry.contextWindow;
   }
+  if (entry.serviceTier === 'fast') declared.serviceTier = 'fast';
   return Object.keys(declared).length > 0 ? (declared as RelayModelProfile) : undefined;
 }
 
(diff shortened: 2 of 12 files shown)


apps/desktop/src/renderer/settings/provider-connection-detail.tsx
```diff
@@ -617,6 +619,23 @@ function ConnectionDetailInner(props: ConnectionDetailProps) {
                         }
                       />
                     </CapabilityRow>
+                    <CapabilityRow label={copy.fastMode} description={copy.fastModeHelp}>
```
Astro-Han: [P1] Gate the Fast row on the same condition the runtime uses, or name the constraint in `fastModeHelp`. This `CapabilityRow` renders for every id in `capabilityModelIds` whenever `showsCapabilities` is true — that is, for both relay provider types and for any model id — while `buildFamilyWire` only emits `serviceTier` when `wire === 'openai-responses'` and `reasoningReplay.contract.adapter === 'openai'` and `supportsFastServiceTier(modelId)` (`packages/runtime/src/model-factory.ts:525-531`). On an `openai-compatible` (Chat Completions) relay the selector appears, the choice is normalized, persisted into `relayModelProfiles`, and then silently dropped at request build; your own test asserts exactly that, `buildProviderOptions(chat, 'fast-model')` deep-equals `{}`. The PR body says Chat-compatible relays "do not receive a misleading option that the SDK would strip" — the runtime does not, but the settings UI does, and the help copy states no constraint, so the user sees a saved setting that never reaches the wire. Confirmed by reading the source at this head. Regression test needed: a renderer test asserting the Fast row is absent for an `openai-compatible` connection and for a Responses relay model id outside the supported families.

packages/core/src/runtime-policy/connection-catalog-codec.ts
```diff
@@ -244,13 +249,14 @@ export function decodeRelayModelProfilesTable(
     const entry = exactRecord(
       rawEntry,
       `relay model profile for ${modelId}`,
-      ['thinkingLevels', 'vision', 'contextWindow'],
+      ['thinkingLevels', 'vision', 'contextWindow', 'serviceTier'],
```
Astro-Han, line 251: [P1] Bump `RUNTIME_HOST_COMPATIBILITY_EPOCH` for this new Runtime Policy field. `RelayModelProfile` crosses the Host protocol as `ConnectionCatalogPageItem.relayProfile` (`packages/runtime-host/src/protocol/runtime-policy.ts`), and this key list is the admission check on the Host's write path through `normalizeConnectionCatalogEntryUpdateForProvider` -> `packages/storage/src/runtime-policy/connection-catalog-document.ts:194`. `exactRecord` rejects unknown keys outright (`packages/core/src/runtime-policy/domain-codec.ts:27-29`), so a new Client paired with an epoch-27 Host — a pair the handshake still admits, because this PR leaves the epoch at 27 — fails the entire connection update with `relay model profile for <id> contains unknown field 'serviceTier'` the first time a user saves Fast, and the same decode makes an already-written document unreadable after a downgrade. This is the case epoch 27's own comment describes, "Older peers cannot safely preserve that field", and the precedent is direct: #2463 introduced this exact key list and bumped the epoch, #2565 bumped it for connection fields. Note that 28 is currently contested by #3236, #3199 and #3133, all taking 28 from a base at 27 — take 28 and re-check it at merge time. Confirmed by reading the source at this head. No new regression test is needed beyond the constant itself; `runtime-policy-codec.test.ts` already covers the key.

Astro-Han (commented):
COMMENT.

Splitting Fast out of #3300 is the right call, and the concern I raised there is genuinely addressed: `service_tier: "fast"` is now asserted against a captured Responses request body rather than an intermediate `providerOptions` object, and `gpt-5-relay` does satisfy the pinned SDK's priority-processing rule, so that capture proves what it claims. I re-derived it at `2a66ac7a6`.

Three findings inline; the two P1s I think need to land before merge.

Two things above the line level:

1. This PR is not only Fast. It re-implements #3300's relay taxonomy wholesale — `isRelayProviderType`, `PROVIDER_DEFAULTS.relayModelProfiles`, and the codec/UI admission of `openai-responses-compatible` — from `main` rather than from #3300. Two open PRs each export the same `isRelayProviderType` from the same file: whichever lands first makes the other conflict, and neither is a single revertable intent any more. Stacking #3309 on #3300 leaves the taxonomy with exactly one owner.

2. `mergeable_state` is `dirty` at this head, so the branch needs a rebase before any CI result on it means much.

AI disclosure: this review was assisted by Claude (Opus) for code search and cross-checking. Every finding below I re-derived myself against the source at `2a66ac7a6` and the pinned `@ai-sdk/openai@4.0.42`; nothing unverified is included.

packages/runtime/src/model-factory.ts
```diff
@@ -627,3 +646,8 @@ function openAiCompatibleProviderOptionsKey(
     ? toCamelCase(connection.slug)
     : connection.providerType;
 }
+
+/** The pinned OpenAI SDK strips fast for model ids outside these families. */
+function supportsFastServiceTier(modelId: string): boolean {
```
Astro-Han: [P2] Mirror `@ai-sdk/openai@4.0.42`'s `supportsPriorityProcessing` here instead of approximating it — this predicate diverges from it in both directions. The SDK's rule (`node_modules/@ai-sdk/openai/dist/index.js:52`) is `modelId.startsWith('gpt-4') || gptVersion.major >= 5 && !isGptNanoModel && !isGptChatModel || oSeriesVersion >= 3`, with `gptVersion` parsed by `/^gpt-(\d+)(?:\.(\d+))?(?:-(.+))?$/`. False positives: `gpt-5-nano` and `gpt-5-chat-latest` match this regex but the SDK excludes them, so `service_tier` is deleted at `dist/index.js:6353` and the toggle becomes the silent no-op this gate exists to prevent. False negatives: `gpt-4o`, `gpt-4o-mini`, `gpt-4.1`, `gpt-5.1` and `gpt-5.1-codex` all satisfy the SDK but fail `gpt-4(?:-|$)` / `gpt-5(?:-|$)`, so Fast is refused for models that would have accepted it — `gpt-5.1` most notably. `getOpenAILanguageModelCapabilities` is not exported from the package, so the predicate does have to be restated by hand, which is what makes pinning it worthwhile. Confirmed by reading the pinned SDK in `node_modules`. Regression test needed: a table over `gpt-4o`, `gpt-4.1`, `gpt-5`, `gpt-5.1`, `gpt-5-nano`, `gpt-5-chat-latest`, `o3-mini`, `o4-mini` and a plain relay id, asserting the emitted providerOptions for each.

hbw00111:
Addressed the review findings in 0854f7c48:

- rebased the Fast commit onto #3300 head edd5321e3, so relay taxonomy has one owner
- gated the renderer row and runtime emission through shared `supportsRelayFastServiceTier`
- mirrored the pinned `@ai-sdk/openai@4.0.42` priority-processing rule, including nano/chat exclusions and gpt-4o/gpt-4.1/gpt-5.1 support
- bumped `RUNTIME_HOST_COMPATIBILITY_EPOCH` to 28 for the new Runtime Policy field
- added the requested model-family table coverage and retained the captured Responses body assertion for `service_tier: "fast"`

Verified core/runtime/runtime-host builds, focused Fast tests (5 passed), Biome on changed files, and `git diff --check`.

packages/core/src/model-thinking.ts
```diff
@@ -231,12 +237,31 @@ export function relayModelProfile(
...
+ * runtime share this gate so a saved Fast declaration always reaches the wire.
+ */
+export function supportsRelayFastServiceTier(providerType: ProviderType, modelId: string): boolean {
+  if (providerType !== 'openai-responses-compatible') return false;
+  const oSeriesVersion = /^o(\d+)(?:-|$)/.exec(modelId)?.[1];
+  const gptMatch = /^gpt-(\d+)(?:\.(\d+))?(?:-(.+))?$/.exec(modelId);
+  const gptMajor = gptMatch?.[1] === undefined ? undefined : Number(gptMatch[1]);
+  const gptVariant = gptMatch?.[3];
+  const isGptNanoModel = gptVariant?.startsWith('nano') ?? false;
+  const isGptChatModel = gptVariant?.startsWith('chat') ?? false;
```
Astro-Han, line 255: [P3] Add the `minor == null` condition the SDK's `isGptChatModel` carries, or this gate is narrower than the rule it mirrors. `@ai-sdk/openai@4.0.42` computes it as `gptVersion.minor == null && variant.startsWith('chat')` (`dist/index.js:49`) — a *chat* variant only disqualifies a model that has no minor version. This line drops the minor check, so `gpt-5.1-chat` is refused Fast here while the SDK would accept it and send `service_tier`. That is a false negative on one hypothetical id rather than the silent-no-op class the earlier finding was about, so it costs a capability rather than misleading anyone, and I would not hold the merge for it. Everything else in this function matches the SDK verbatim, including `startsWith('gpt-4')`, the `/^gpt-(\d+)(?:\.(\d+))?(?:-(.+))?$/` parse and the `/^o(\d+)(?:-|$)/` o-series rule. Confirmed by reading the pinned SDK and the source at this head. Regression test needed: add `gpt-5.1-chat` to whatever table pins this predicate.

Astro-Han (approved):
APPROVE. Both P1s from my last review are fixed, and fixed at the right level.

- **The UI/runtime split.** `supportsRelayFastServiceTier` is now one exported predicate that both sides call — `provider-connection-detail.tsx` gates the whole `CapabilityRow` on `showsFastMode`, and `model-factory.ts:530` gates the wire on the same function with the same arguments. That is stronger than what I asked for: the previous shape had two independent conditions that could drift, and now there is one, so a saved Fast declaration cannot exist without a path to the wire.
- **The epoch.** `RUNTIME_HOST_COMPATIBILITY_EPOCH` is 28 with a comment that names the actual reason ("Relay model profiles carry the Fast service-tier declaration"). That closes the mixed-version hole where a new Client and an epoch-27 Host would shake hands and then fail the first save with an unknown-field error.
- **The predicate itself.** It now mirrors `@ai-sdk/openai@4.0.42`'s `supportsPriorityProcessing` rather than approximating it — `gpt-5-nano` and `gpt-5-chat-latest` are correctly excluded, and `gpt-4o`, `gpt-4.1` and `gpt-5.1` are correctly included. I checked each against the pinned SDK. One narrow residual is inline as a P3.

Two things to flag that are not findings against this PR:

**Epoch 28 is now contested by five open PRs** — #2521, #2641, #3236, #3322 and this one — against a `main` at 27. The `= 28` line merges cleanly because every branch writes identical text and each PR's own lower-bound assertion still passes, so whichever lands second silently shares an epoch with the first and the mixed-version guarantee is lost without any conflict to warn you. This needs a human check of the constant at merge time, every time, until #3313 removes the class.

**The taxonomy overlap with #3300 is resolved from my side.** This head no longer re-implements it in a way that would conflict — `isRelayProviderType` and the `relayModelProfiles` registry flag are still here, so whichever of the two lands first, the other will need a rebase, but that is ordinary and no longer a competing definition.

AI disclosure: this review was assisted by Claude (Opus) for code search and cross-checking. Everything above I re-derived myself against the source at `0854f7c48` and the pinned `@ai-sdk/openai@4.0.42` in `node_modules`.

PR #3310, opened 2026-08-20
fix(cli): stop steering known slash commands into a running turn

### LINKED ISSUE #3308: fix(cli): /recap and /resume typed during a running turn are steered into the model as prompt text
## What happened

In the TUI, typing `/recap` or `/resume` while a turn is running does not execute the command. The literal text is injected into the running turn as a steering message (the transcript shows `Steering: /resume`), so the model sees `/recap` / `/resume` as ordinary user text and may react to it — derailing the current task, polluting context, and burning tokens. The recap is never generated and the safe-boundary resume never runs.

Root cause: in `packages/cli/src/pi-tui-runner.ts`, `editor.onSubmit` special-cases only exit forms, `/swarm`, `/graph`, and `/goal` while `turnRunning`; every other input — including known slash commands — goes to `steerRunningTurn()`.

The behavior is also timing-dependent: if the turn ends between submit and enqueue, the runtime returns `fallback` and the CLI re-routes through `submitPrompt()`, where the command *does* run. Same command, two outcomes, depending on whether the turn was still alive.

Expected:
- `/recap` answers locally mid-turn — it uses the independent `session.recap.generate` call with its own in-flight lock and never touches the running turn.
- `/resume` and other session-mutating commands (`/compact`, `/rewind`, `/model`, `/session`, `/new`, `/rename`, `/thinking`, `/permissions`, `/move`, `/skill`, `/setup`, `/context`, `/help`) refuse with a clear message instead of steering — today they would silently no-op via the `runControl` busy gate even if they did reach the handler.
- Unknown slash-prefixed text keeps steering unchanged (it may be intended prompt text).

## How to reproduce

1. Run the TUI from source (`npm run cli:dev`), start a session, submit a prompt that keeps the agent working.
2. While the turn is running, type `/recap` and press Enter.
3. Observe: transcript shows `Steering: /recap`; no recap notice appears; the model responds to the literal text.
4. Repeat with `/resume`: no safe-boundary resume; the text is injected instead.

## Environment

- Maka version or commit: 9d4392c85 (main)
- OS and version: Windows (also reproducible from source on Linux/macOS)
- Surface: TUI
- Node.js version: 22

me2seeks:
## Summary

Typing `/recap` or `/resume` in the TUI while a turn is running never ran the command: `editor.onSubmit` only special-cased exit forms, `/swarm`, `/graph`, and `/goal`, so every other input — including known slash commands — went to `steerRunningTurn()` and the model received the literal text (`Steering: /resume`) as a user message. Whether the command executed at all depended on whether the turn was still alive when the enqueue landed (the `fallback` path re-routes through `submitPrompt`).

Mid-turn submissions of known slash commands now route explicitly: `/goal` and `/recap` answer locally (both are independent of the running turn — recap uses the separate `session.recap.generate` call behind its own in-flight lock); every other known command is refused with a clear message ("Cannot run /X while a turn is running — interrupt it (Esc) or wait for it to finish.") instead of being steered or silently no-opping on the `runControl` busy gate; unknown slash-prefixed text (skill invocations like `/skill:<name>`, paths) still steers unchanged.

Fixes #3308

## Verification

- `npm --workspace maka-agent test`: 320 pass / 0 fail, including 4 new tests under `slash commands during a running turn` (`/recap` answers locally, `/resume` and `/model` refuse with a message, unknown slash text still steers).
- The new tests fail without the fix: reverting `pi-tui-runner.ts` and rebuilding makes the `/recap`, `/resume`, and `/model` tests fail while the unknown-slash test still passes.
- `npm run lint`, `npm run format:check`, `npm run typecheck`: pass.
- Not run: desktop e2e (no desktop change).

## AI use

Select exactly one:

Tool(s) and scope: Maka (AI agent) located the routing defect, implemented the fix and tests, and ran the verification; the commit carries the `Generated-by: Maka` trailer. Human contributor of record reviewed the diagnosis and the routing design and decided to submit.

## Checklist

- [ ] No

### DIFF
--- a/packages/cli/src/pi-tui-runner.ts
+++ b/packages/cli/src/pi-tui-runner.ts
@@ -1084,13 +1084,34 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
         }
         return;
       }
-      // Read-only status commands answer locally even mid-turn instead of
-      // steering into the model as prompt text: an autonomous goal loop keeps
-      // a turn running almost by definition, and that is exactly when the
-      // user reaches for `/goal` (review finding on turnRunning routing).
-      if (prompt.trim().split(/\s+/, 1)[0] === '/goal') {
+      // Known slash commands typed mid-turn must not steer into the model as
+      // prompt text (review finding on turnRunning routing). `/goal` and
+      // `/recap` answer locally: both are independent of the running turn —
+      // `/goal` is read-only status, and `/recap` uses the separate
+      // session.recap.generate call behind its own in-flight lock. Every other
+      // known command either mutates session state behind the turn's back or
+      // opens a picker the turn would race (and would silently no-op on the
+      // runControl busy gate even if it ran), so refuse it with a clear
+      // message. Unknown slash-prefixed text still steers: it may be intended
+      // prompt text (a skill invocation such as `/skill:<name>`, or a path).
+      const commandToken = prompt.trim().split(/\s+/, 1)[0] ?? '';
+      const knownCommand = slashCommands.find(
+        (candidate) =>
+          `/${candidate.name}` === commandToken ||
+          candidate.aliases?.some((alias) => `/${alias}` === commandToken),
+      );
+      if (knownCommand) {
         editor.addToHistory(prompt);
-        handleSlashCommand(prompt, 0);
+        if (knownCommand.name === 'goal' || knownCommand.name === 'recap') {
+          handleSlashCommand(prompt, 0);
+        } else {
+          state.entries.push({
+            kind: 'notice',
+            level: 'error',
+            text: `Cannot run /${knownCommand.name} while a turn is running — interrupt it (Esc) or wait for it to finish.`,
+          });
+          requestRender();
+        }
         return;
       }
       steerRunningTurn(prompt);
--- a/packages/cli/src/__tests__/pi-tui-runner.test.ts
+++ b/packages/cli/src/__tests__/pi-tui-runner.test.ts
@@ -4930,6 +4930,159 @@ describe('Maka Pi TUI runner', () => {
     });
   });
 
+  describe('slash commands during a running turn', () => {
+    test('/recap answers locally instead of steering into the model', async () => {
+      const terminal = new FakeTerminal();
+      const driver = new SteeringTurnDriver();
+      driver.rewindTargets = [{ turnId: 'turn-1', label: 'first' }];
+      let calls = 0;
+      const run = runMakaPiTui({
+        title: 'Maka',
+        driver,
+        cwd: '/repo',
+        model: 'm',
+        connectionSlug: 'c',
+        permissionMode: 'bypass',
+        terminal,
+        recap: {
+          generate: async () => {
+            calls += 1;
+            return { ok: true, text: 'mid-turn recap', raw: 'mid-turn recap' };
+          },
+        },
+      });
+
+      terminal.input('start the work');
+      terminal.input('\r');
+      await waitFor(() => terminal.progressStates.at(-1) === true);
+
+      // The recap uses the independent session.recap.generate call behind its
+      // own in-flight lock, so it can answer while the turn is running;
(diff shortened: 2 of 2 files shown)


packages/cli/src/pi-tui-runner.ts
```diff
@@ -1084,13 +1084,34 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
...
+      const commandToken = prompt.trim().split(/\s+/, 1)[0] ?? '';
+      const knownCommand = slashCommands.find(
+        (candidate) =>
+          `/${candidate.name}` === commandToken ||
+          candidate.aliases?.some((alias) => `/${alias}` === commandToken),
+      );
+      if (knownCommand) {
         editor.addToHistory(prompt);
-        handleSlashCommand(prompt, 0);
+        if (knownCommand.name === 'goal' || knownCommand.name === 'recap') {
```
Astro-Han, line 1105: [P3] Consider putting the mid-turn disposition on the command spec instead of matching two names here. `slashCommandHandlers` is already a `Record<TuiSlashCommandId, TuiSlashCommandHandler>` keyed by the same catalog ids (`pi-tui-runner.ts:2943-2949`), so a `midTurn: 'local' | 'refuse'` field on each handler would keep "safe while a turn is running" next to the handler that has to be safe, and a newly added command would have to state its answer rather than inherit `refuse` from a string comparison at the call site. The default here is the safe one, so drift costs correctness nothing — this is a seam observation, not a defect, and I would not hold the merge for it. Confirmed by reading the source at this head. No regression test needed.

Astro-Han (approved):
APPROVE.

The routing is right and the tests pin it at the observable boundary: `driver.steered` stays empty for `/recap`, `/resume` and `/model`, and `/skill:review` still steers, so both directions are covered rather than just the fixed one.

I checked the three claims the new comment rests on, and all three hold at `5c4ee2284`:

- `runAgentTurn` sets `busy` and `turnRunning` together (`pi-tui-runner.ts:1128-1130`), so `runControl`'s `if (busy) return` at `:633` really would swallow `/help`, `/context`, `/new` and the rest without a word. Refusing loudly is strictly better than that silent no-op, and the rationale in the comment is accurate.
- `recap.run` goes straight to `runRecap('manual')` (`:2857-2861`) and never enters `runControl`, so answering `/recap` mid-turn is genuinely safe rather than merely untested.
- `/goal <action>` already carries its own `if (busy)` notice (`:2713-2722`), so exempting `/goal` from the refusal does not reintroduce a silent no-op on the mutating path. That was my first suspicion and the code refutes it.

`biome format` is clean on both changed files at this head.

One non-blocking P3 inline. Separately, and not a review finding: `statusCheckRollup` is empty on this head — the `test` check has not run at all — so this approval covers the diff, not CI.

AI disclosure: this review was assisted by Claude (Opus) for code search and cross-checking. Everything above I re-derived myself against the source at `5c4ee2284`.

packages/cli/src/pi-tui-runner.ts
me2seeks: Agreed — a midTurn disposition on the handler spec is the better seam: keeping "safe while a turn is running" next to the handler that has to be safe, and forcing a newly added command to state its answer, beats the string comparison at the call site.

I'd like to keep this PR at the minimal fix rather than fold the refactor in — a new commit would dismiss the approval already given, and the current default (refuse) is the safe one, so deferring costs nothing. Happy to do the spec-field refactor as a small follow-up after this lands. Thanks for re-deriving the three claims against the source.

me2seeks: Follow-up is up: #3379. `midTurn` landed as a required field on `MakaSlashCommand` with three dispositions — `'local'`, `'refuse'`, plus `'intercepted'` for /exit, /swarm and /graph since their dedicated checks claim them ahead of generic routing and the field would otherwise be a lie. Omitting the field is now a compile error (verified: removing one declaration fails with TS2741).

PR #3311, opened 2026-08-20
fix(desktop): align project session titles with the project name

Astro-Han:
## Summary

Project-grouped session rows were nested a full Astryx SideNav step (24px). That step is for icon-less children so their text meets a parent title after a 16px icon + 8px gap. Session rows already spend 8px on StatusDot in that slot, so the extra 24px pushed titles too far right of the project name.

image

Keep the remaining 8px nest (`--spacing-2`) so titles share one x and the row still sits inside the project.

## Verification

- Ran `node --experimental-strip-types --test apps/desktop/src/main/__tests__/session-project-hierarchy-contract.test.ts` in the worktree (pass).
- Did not run Playwright locally: the geometry spec opens a visible Electron window and would steal focus from an active `maka dev` session. Leave `sidebar-project-row.spec.ts` to CI.

## AI use

Select exactly one:

Tool(s) and scope:

Maka diagnosed the indent, chose the 8px remaining nest, and authored the CSS, contract test, e2e assertion, and commit.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/e2e/sidebar-project-row.spec.ts
+++ b/apps/desktop/e2e/sidebar-project-row.spec.ts
@@ -42,7 +42,10 @@ test('project navigation and actions follow their visual keyboard order', async 
   ]);
   expect(projectNavigationBox).not.toBeNull();
   expect(firstSessionBox).not.toBeNull();
-  expect(firstSessionBox!.x - projectNavigationBox!.x).toBeGreaterThanOrEqual(20);
+  // Remaining nest after StatusDot occupies 8px of Astryx's 24px icon column.
+  const sessionInset = firstSessionBox!.x - projectNavigationBox!.x;
+  expect(sessionInset).toBeGreaterThanOrEqual(6);
+  expect(sessionInset).toBeLessThan(16);
 
   await expect(projectRow.locator('button button')).toHaveCount(0);
   await expect(navigation).toHaveAttribute('aria-expanded', 'true');
--- a/apps/desktop/e2e/sidebar-project-row.spec.ts
+++ b/apps/desktop/e2e/sidebar-project-row.spec.ts
@@ -36,16 +36,27 @@ test('project navigation and actions follow their visual keyboard order', async 
 
   await expect(navigation).toBeVisible();
   await expect(firstSessionControl).toBeVisible();
-  const [projectNavigationBox, firstSessionBox] = await Promise.all([
+  const projectTitle = navigation.getByText(LONG_SIDEBAR_PROJECT_NAME, { exact: true });
+  const sessionTitle = firstSessionControl.getByText('任务 00', { exact: true });
+  await expect(projectTitle).toBeVisible();
+  await expect(sessionTitle).toBeVisible();
+  const [projectNavigationBox, firstSessionBox, projectTitleBox, sessionTitleBox] = await Promise.all([
     navigation.boundingBox(),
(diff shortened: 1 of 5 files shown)


likun666661:
Overall, the direction is sound and the `spacing-6 → spacing-2` geometry checks out. Before marking this PR ready, however, I suggest tightening both the problem definition and the E2E contract.

The current geometry is:

- Project title: 8px row padding + 16px Folder icon + 8px gap = 32px
- Session title before this PR: 24px child inset + 8px row padding + 8px StatusDot + 8px gap = 48px
- Session title after this PR: 8px child inset + 8px row padding + 8px StatusDot + 8px gap = 32px

So the old layout does place session titles 16px too far to the right, and `--spacing-2` aligns them exactly. The production change itself is minimal and appropriate.

I see three points worth addressing:

1. **The problem statement presents a product choice as a design-system fact.** The claim that the 24px SideNav nesting step is specifically for icon-less children is not established by the current Astryx implementation; Astryx applies `spacing-6` to children unconditionally. A more precise statement would be: “The parent and child rows have leading content of different widths. Combined with the fixed nesting inset, this misaligns their titles. The product intent is to align the titles while retaining an 8px hierarchical inset.” This matters because #3175 only recently defined the full 24px nesting step as the visual contract. This PR is selecting a different contract, not merely correcting arithmetic. A before/after screenshot or design reference would make that intent explicit.

2. **The E2E assertion does not directly verify the stated result.** It currently checks that the project and session button left edges differ by 6–16px, rather than checking that their title x-coordinates match. If the StatusDot width, icon size, or SideNav gap changes later, the titles could become misaligned while the test still passes. I suggest directly asserting `abs(projectTitle.x - sessionTitle.x) <= 1–2px`. If preserving hierarchy is also a contract, separately assert that the session button begins about 8px inside the project button.

3. **The visible effect is broader than the title.** This change moves the left edge of the entire session button, including its hover/selected background and hit area, 16px to the left. The right edge should remain stable. That is likely acceptable, but it should be included in the behavior description.

From an Occam’s razor perspective, I would keep the current `spacing-2` implementation. A `calc()` expression or a new component abstraction would add complexity without improving the result. The main improvement is to describe the actual product decision and test the visual outcome—title alignment—rather than pinning a CSS token or internal DOM mechanism.

AI-assisted review disclosure: This comment was prepared with Codex assistance after statically reviewing the PR diff, the Astryx SideNav/StatusDot layout, and the history of #3175.

Astro-Han:
Accepted points 1–3. Kept `--spacing-2`; that is still the smallest implementation.

1. The 24px SideNav nest is a fixed child inset, not an icon-less-child alignment rule. This PR chooses a different product contract from #3175: keep an 8px hierarchical inset so session titles share the project title's x. Comments in `sidebar.css`, the source contract, and the ProjectGroups story now say that.

2. The sidebar e2e now asserts `abs(projectTitle.x - sessionTitle.x) <= 2` and still checks that the session button starts about 8px inside the project button.

3. The whole session button, including hover/selected fill and hit area, moves 16px left; the right edge stays on the rail. That is now in the CSS comment.

Before/after for the title column is on the desktop compare image from the earlier thread.

hqhq1025 (approved):
Reviewed the exact current revision against the stated layout contract, production DOM/CSS path, prior maintainer feedback, and focused geometry/Electron coverage. No actionable defect survived the policy v7 materiality review.

PR #3315, opened 2026-08-20
chore(ui): show signed-in mark in model provider list

zwpaper:
In model list, after signed in a provider, there is no indidcator to mark it as signed in.

this pr add a mark to show it.

Before:
image

After:
image

### DIFF
--- a/apps/desktop/src/renderer/settings/provider-catalog-page.tsx
+++ b/apps/desktop/src/renderer/settings/provider-catalog-page.tsx
@@ -8,7 +8,7 @@ import {
   Selector,
   VStack,
 } from '@astryxdesign/core';
-import { ICON_SIZE, ChevronRight, Search } from '@maka/ui/icons';
+import { ICON_SIZE, Check, ChevronRight, Search } from '@maka/ui/icons';
 import {
   CATALOG_PROVIDER_TYPES,
   RECOMMENDED_PROVIDER_TYPES,
@@ -144,7 +144,17 @@ export function ProviderCatalogPage(props: {
               startContent={<ProviderLogo type={card.providerType} />}
               label={/* a11y-allow: this label names the ROW, not the span. Astryx's Item puts consumer props on its outer wrapper and renders a separate invisible <button> for the click target, so an aria-label on the Item never reaches that button — measured. The button is named from its content, and this span is how the status reaches that name. Removing it drops the runtime error from the row's accessible name (settings.spec:226).*/ <span aria-label={providerCopy.oauthSection.cardAria(card.name, card.status, card.description)}>{card.name}</span>}
               description={card.description}
-              endContent={<ChevronRight size={ICON_SIZE.chrome} aria-hidden="true" />}
+              endContent={(
+                <HStack gap={2} vAlign="center">
+                  {card.isLoggedIn && (
+                    <span className="settingsStatus" aria-hidden="true">
+                      <Check size={ICON_SIZE.chrome} />
+                      <span>{providerCopy.oauthSection.signedIn}</span>
+                    </span>
+                  )}
+                  <ChevronRight size={ICON_SIZE.chrome} aria-hidden="true" />
+                </HStack>
+              )}
               onClick={() => props.onPick({
                 method: 'account',
                 cardId: card.id,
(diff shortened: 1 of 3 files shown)


apps/desktop/src/renderer/settings/provider-catalog-page.tsx
```diff
@@ -144,7 +144,17 @@ export function ProviderCatalogPage(props: {
               startContent={<ProviderLogo type={card.providerType} />}
               label={/* a11y-allow: this label names the ROW, not the span. Astryx's Item puts consumer props on its outer wrapper and renders a separate invisible <button> for the click target, so an aria-label on the Item never reaches that button — measured. The button is named from its content, and this span is how the status reaches that name. Removing it drops the runtime error from the row's accessible name (settings.spec:226).*/ <span aria-label={providerCopy.oauthSection.cardAria(card.name, card.status, card.description)}>{card.name}</span>}
               description={card.description}
-              endContent={<ChevronRight size={ICON_SIZE.chrome} aria-hidden="true" />}
+              endContent={(
+                <HStack gap={2} vAlign="center">
+                  {card.isLoggedIn && (
```
Astro-Han, line 149: Contract observation, not a defect, so no P level. `isLoggedIn` in `provider-oauth-section.tsx:148-152` is true for `authenticated`, `refreshing`, `quota_unavailable` **and** `provider_rejected`, so a provider whose credentials the provider itself has rejected now carries an affirmative check mark reading "Signed in". That classification predates this PR and already fed `card.status` and the row's accessible name, so the change is consistent rather than newly wrong — but a check mark is a stronger claim than an aria string, and `provider_rejected` is precisely the state a user would open this list to notice. Worth deciding deliberately whether the mark should track `authenticated | refreshing` only and leave the broken states to the description. Confirmed by reading the source at this head.

apps/desktop/src/renderer/settings/provider-catalog-page.tsx
```diff
@@ -144,7 +144,17 @@ export function ProviderCatalogPage(props: {
...
+                    <span className="settingsStatus" aria-hidden="true">
+                      <Check size={ICON_SIZE.chrome} />
```
Astro-Han, line 151: Test gap rather than a defect, so no P level. Nothing in this PR is covered by a test, and the row this touches already has desktop e2e coverage — the a11y comment two lines up cites `settings.spec:226` for the accessible name of exactly this `ListItem`. An assertion in that spec that a signed-in card's row renders the mark, and a signed-out one does not, would be a few lines and would pin the one thing a future refactor of `endContent` could silently drop. Not blocking from my side, but the PR template's "Tests cover the change and fail without it" is currently unmet.

Astro-Han (approved):
APPROVE.

Small and correct. I checked every symbol the diff depends on at `eceba28e2` and all of them already exist — `HStack` is imported in `provider-catalog-page.tsx`, `isLoggedIn` is already on the card projection (`provider-oauth-section.tsx:159`), `copy.signedIn` is defined in both locales, and `settingsStatus` is the same class the other settings pages use. `biome format` is clean on all three changed files.

Two things I suspected and could not sustain, recorded so nobody re-raises them:

- I expected the new mark to go stale after a sign-out, since neither `ClaudeSubscriptionCard.logout` nor `useOAuthLoginFlow.logout` calls `onLoginSuccess`. It does not: `providers-panel.tsx:300-314` renders the catalog and the setup level as alternatives, so entering the login panel unmounts `ProviderCatalogPage`, and coming back remounts it and re-runs `useOAuthCards`'s mount-time refresh.
- `aria-hidden="true"` on the badge is right, not an a11y regression: `useOAuthCards` already puts `copy.signedIn` into `card.status`, which `cardAria` folds into the row's accessible name, so exposing the badge too would announce it twice.

One thing above the line level: threading `onLoginSuccess` through `ClaudeSubscriptionCard` and `GitHubCopilotLoginPanel` is a genuine fix — those two panels previously left the providers panel's `reload()` unfired after a successful sign-in, unlike `SubscriptionLoginPanel` — but it is a second intent, not something the check mark needs (the remount above already covers the catalog). It is small enough that I would not ask you to split it; noting it because one revert now undoes two fixes.

Also minor: the description does not follow `.github/pull_request_template.md`, so the AI-use declaration and the checklist are missing.

Two non-blocking notes inline, neither carrying a P level.

AI disclosure: this review was assisted by Claude (Opus) for code search and cross-checking. Everything above I re-derived myself against the source at `eceba28e2`.

Issue #3313, opened 2026-08-20
Derive the Runtime Host compatibility epoch instead of hand-editing it


Astro-Han:
### Problem

`RUNTIME_HOST_COMPATIBILITY_EPOCH` in `packages/runtime-host/src/protocol/index.ts` is a hand-edited integer, and every contributor who changes the wire contract has to guess a value no one else has taken. Right now `main` is at 27 while three independent open PRs each set it to 28 for different, mutually incompatible wire changes — #3236 (access credential `prepare`/`finalize`), #3199 (`goal.arm`), #3133 (session trace cursor pages) — and #3299 takes 29 on the assumption that exactly one 28 lands.

What makes this worth fixing rather than just coordinating around is that the collision is silent by construction:

- Every branch writes the **same text** to that line, so git's three-way merge resolves it with no conflict. Only the adjacent comment block conflicts, and keeping both comments is the natural resolution.
- Per-PR assertions of the form `assert epoch > 27` still pass on all of them.
- Nothing compares the merged value against the merge base.

So two incompatible protocols end up advertising one epoch. `client/connection.ts` then *admits* the peer because the epochs match, and the unknown operation fails decode and tears down the transport — bypassing the structured `incompatible` handshake the epoch exists to provide. The failure surfaces to a user as a connection that dies instead of a version mismatch that explains itself.

This is not the first time; it recurs whenever two protocol changes are open at once, and rebasing does not surface it.

### Desired outcome

Two peers advertise the same compatibility discriminator if and only if they actually share a wire contract, without any contributor having to know what other branches are doing.

A useful constraint for whoever picks this up: **the epoch is never ordered at runtime.** Every comparison is equality — `client/connection.ts:1287`, `:1328`, `:1448`, and the corresponding check in `host-kernel.ts` — and the `- 1` occurrences are test fixtures constructing "some other epoch". So the value does not have to be a monotonically increasing integer. It only has to differ whenever the contract differs.

### Alternatives or workarounds

**Derive it from the protocol surface.** Replace the hand-edited constant with a digest computed at build time over the canonical contract — the aggregated `*_OPERATION_SPECS` in `protocol/operations.ts` together with the frame and codec key lists. Two branches that add different operations then produce different digests on their own, which makes the collision impossible rather than merely detectable, and removes the "what number is free?" question entirely. It also gets the converse right: a refactor that preserves the contract yields the same digest, whereas the manual scheme encourages a bump whenever someone is unsure.

Costs worth weighing before committing to this:

[...]

PR #2521, opened 2026-08-08
fix(runtime): distinguish usage limits from auth errors

me2seeks:
## Summary

After #2675, Maka already has one safe path for provider diagnostics: ModelAdapter keeps an allowlisted, redacted, bounded summary; the terminal RuntimeEvent remains canonical; Runtime Host projects the message through failed Turn snapshots and recovery; and Desktop can include the exact execution trace in copied reports.

This PR fills the remaining classification gap on top of that plumbing:

- structured provider identifiers take precedence over ambiguous HTTP status codes — a structured context-overflow code is evaluated before abort wording and before numeric 402/429/401, so a real `context_length_exceeded` envelope reaches compaction/recovery instead of being misread as Abort or RateLimit;
- a bare HTTP 403 is no longer assumed to mean invalid credentials;
- the observed Kimi `permission_error` remains unknown and non-retryable, so the bounded provider explanation reaches TUI/Desktop instead of being replaced with `Authentication failed`;
- stable identifiers such as `permission_denied`, `usage_limit_reached`, and `insufficient_quota` produce distinct non-retryable account meanings;
- existing bare-429 rate-limit behavior remains unchanged, while structured account limits are not retried as transient throttling.

Scope, accurately at the current head:

- `packages/core`: two new fields on `ErrorEvent`/`RuntimeEventErrorContent` for the structured provider failure, a new `provider-failure` module, and its export-map entry.
- `packages/runtime`: the classification precedence fix at the shared ModelAdapter seam, provider-failure provenance on connection-effect outcomes, and conformance/classification regression coverage.
- `packages/runtime-host`: Turn snapshots and connection effects carry the structured provider failure code and the bounded provider message (protocol, canonical turn snapshot, session projector); the compatibility epoch advances to **29** because older peers cannot safely preserve those fields.
- `apps/desktop`: main-process and renderer presentation of the account-limit vs. authentication vs. unknown states, including the allowance/plan/reset recovery guidance, in both locales.
- `packages/cli`: `pi-transcript` renders the bounded provider message through a small zh/en copy table.

The branch was rebuilt on current `main`. Older Runtime Host, Headless, continuity-schema, and diagnostics changes were removed earlier because #2675 now owns those responsibilities.

Fixes #2516

## Verification

- `npm --workspace @maka/core test` — all pass
- `npm --workspace @maka/storage test` — all pass
- `npm --workspace @maka/runtime test` — all pass (provider classification, conformance, telemetry)
[...]

PR #3316, opened 2026-08-20
fix(runtime-host): forward the durable steering echo to session subscribers

### LINKED ISSUE #3304: fix(tui): consumed steering message is never projected into the transcript
## Summary

When a steering message is sent while the TUI is mid-turn (Enter during "Working…"), the model receives it (visible in thinking), but the message is **never projected into the TUI transcript** — the pending `Steering: …` bar clears and no user message is rendered in its place.

## Expected

A consumed steering message is rendered in place as a user turn in the transcript (this is what `pi-transcript.ts` does for `steering_message` events: `packages/cli/src/pi-transcript.ts` → `applyMakaSessionEventToTranscript` → `appendUserPrompt`).

## Actual

Nothing is rendered. After the run consumes the steering, the transcript shows no trace of the user's interjection (until a full session reload).

## Root cause

The host-backed TUI has exactly one live render path for steering, and it is racy by construction:

1. The kernel echoes a **durable** `steering_message` session event when the run drains steering at a step boundary (`packages/runtime/src/ai-sdk-backend.ts` → `drainSteeringInto`, ~4755). This is what makes the ledger/transcript correct.

2. But the Runtime Host event pump **drops that event for subscribers**: `root-turn-coordinator.ts` (~2099) only forwards `isRuntimeSessionTransientEvent` types (text/thinking/tool/provider_retry) to `acceptRuntimeEvent`, and `RuntimeSessionTransientEvent` (`session-continuity-coordinator.ts:63`) does not include `steering_message`. It doesn't even trigger `enqueueCanonicalRefresh`.

3. So the client's only live source is the projector **synthesizing** `steering_message` from queue entries observed in the transient `in_flight` state (`packages/runtime-host/src/adapter/session-projector.ts` → `accept()` → `newlyInFlight`, ~344).

4. That `in_flight` window (message-coordinator `#pull` → `#ack`, ~1116–1150) is milliseconds, and the projection reaches subscribers only through an **async, coalesced** canonical refresh (`#mutated` → `onProjectionChanged` → `enqueueCanonicalRefresh` → session lane → `#readCanonicalProjection`). If no refresh observes the intermediate state — easily the common case under load — the subscriber sees the queue jump `queued → []` and the projector emits only `queue_update` (empty): no `steering_message` is ever emitted.

## Reproduction (projector level)

Feeding `RuntimeHostSessionProjector` the two snapshots a subscriber actually receives in the racy case:

- prev: `queue.steering = [{ state: 'queued', … }]`, `queueRevision: 2`
- next: `queue.steering = []`, `queueRevision: 4` (in_flight state never observed)

produces only:

```json
[{ "type": "queue_update", "steering": [], "followup": [] }]
```

— no `steering_message`. Feeding the intermediate `{ state: 'in_flight' }` snapshot does produce `steering_message`, confirming the render depends entirely on observing a transient queue state.

## Fix direction

Make the authoritative, durable `steering_message` a first-class event for session subscribers instead of relying on observing the transient `in_flight` queue state — e.g. extend the host event pump / `RuntimeSessionTransientEvent` to forward `steering_message` (it has no `toolUseId`, so the session-event frame shape needs a small generalization), and keep the queue projection responsible only for the pending `Steering:`/`Queued:` bar. The client reducer (`pi-transcript.ts:752`) already renders it; the desktop projector (`packages/ui/src/live-turn-projection.ts:169`) would benefit from the same fix.

## Environment

- maka TUI against Runtime Host, observed on main (609520352)

me2seeks:
## Summary

Fixes #3304 — a steering message sent while the TUI is mid-turn was delivered to the model but never rendered in the transcript.

Root cause: the Runtime Host event pump dropped the durable `steering_message` (it is not a transient tool/text event), leaving observation of the **transient `in_flight` queue state** as the only live render path. The async, coalesced canonical refresh can skip that state entirely (`queued → consumed`), so subscribers saw only an emptying `queue_update` and the interjection vanished from the UI.

## Changes

- **Protocol** (`session-continuity.ts`): new `SessionSteeringEvent` carried on the existing `subscription.session_event` frame (no `toolUseId`), strict exact-keys decode with bounded message content; wire schema 4 → 5 (mismatched clients already fail fast at subscription open).
- **Host** (`root-turn-coordinator.ts`, `session-continuity-coordinator.ts`): the pump now forwards the durable `steering_message`; `projectToolEvent` → `projectSessionEvent` passes it through verbatim to subscribers.
- **Client projector** (`adapter/session-projector.ts`): projects the forwarded echo as a first-class `steering_message` SessionEvent, and dedups per `messageId` against the existing queue in-flight synthesis (`#renderedSteeringMessageIds`) — whichever authoritative path arrives first renders exactly once. The synthesis stays for the attach/rejoin window where the durable event has not landed yet.

TUI reducer and desktop live projection already render `steering_message`, so no surface changes were needed there.

## Tests

- New: projector regression (in-flight state never observed → echo still renders; bidirectional exactly-once dedup; rejoin seed), protocol round-trip + rejection cases, coordinator forwarding, and a full-stack test (real host + fake backend + subscribed client receives the steering frame).
- `packages/runtime-host`: 1024/1024 pass; `packages/cli` steering/transcript/driver suites: 202/202; `@maka/ui`: 183/183; biome check clean; desktop typecheck clean.

## Compatibility

Wire schema bump 4 → 5: old clients connecting to a new host fail the handshake at subscription open instead of mis-decoding the new event variant (the established closed wire-schema convention).

## AI use


packages/runtime-host/src/server/root-turn-coordinator.ts
```diff
@@ -2787,6 +2787,7 @@ function isRuntimeSessionTransientEvent(
     event.type === 'tool_progress' ||
     event.type === 'tool_result_preview' ||
     event.type === 'tool_result' ||
+    event.type === 'steering_message' ||
```
Astro-Han, line 3001: [P3] Rename this predicate — it now returns true for the one event this PR's own protocol comment calls durable. `isRuntimeSessionTransientEvent` gates exactly one thing at `:2099`, `continuity.acceptRuntimeEvent`, so what it really answers is "forward this live to subscribers instead of waiting for the canonical refresh", and adding `steering_message` to it is the correct behaviour. But the name now says the opposite of `SessionSteeringEvent`'s doc comment ("The durable mid-turn user interjection ... forwarded verbatim"), and the next person deciding whether a new event belongs in this list will read the name, not the call site. Something like `isRuntimeSessionForwardedEvent` would carry the actual rule. The same drift shows up on the client side: `adapter/session-projector.ts:480`'s `projectToolEvent` now handles a non-tool event, while its Host-side counterpart was renamed to `projectSessionEvent` in this PR — worth renaming both together. Confirmed by reading the source at this head. No regression test needed.

Astro-Han (dismissed):
APPROVE.

The fix is at the right seam. The root cause is stated precisely and the change addresses it rather than the symptom: the pump now forwards the durable `steering_message`, so rendering no longer depends on a subscriber happening to observe a transient `in_flight` queue state that the coalesced canonical refresh is free to skip.

I checked the two things that could have made this wrong, and neither does:

- **The transient classification does not affect persistence.** Adding `steering_message` to `isRuntimeSessionTransientEvent` looked like it might drop the message from durable storage. It cannot: `:2099` is the predicate's only call site, and it gates `continuity.acceptRuntimeEvent` — live forwarding — while persistence happens upstream in the run's own event stream. Only the name is now misleading, which is the P3 inline.
- **The mixed-version story holds without an epoch bump.** `SESSION_CONTINUITY_SCHEMA_VERSION` goes 4 to 5, and the check lives inside `decodeSessionContinuitySnapshot` (`session-continuity.ts:521`), which the `subscription.open` result decode runs — so a schema-4 client fails at open and never reaches a frame carrying the new variant. The precedent is real and recent: #3074 took 3 to 4 the same way without touching the epoch. Worth stating explicitly since #3236, #3199, #3133 and #2521 are currently all contending for epoch 28 — this PR correctly stays out of that.

The dedup is the part I looked at hardest and it holds up. `#renderedSteeringMessageIds` is set on all three render paths — the rejoin seed at `:154`, the newly-in-flight synthesis at `:356`, and the forwarded echo at `:329` — and the tests pin the race in both orders, plus the rejoin case, plus the original bug's shape (in-flight state never observed). The protocol test covers a round trip, the `toolUseId` rejection and the oversize-content rejection. The full-stack test in `execution-host-message.test.ts` asserts against a real host with a subscribed client rather than an intermediate object.

`biome format` is clean on all eight files at this head.

One P3 inline, purely naming. Nothing blocking.

AI disclosure: this review was assisted by Claude (Opus) for code search and cross-checking. Everything above I re-derived myself against the source at `18e2905f9`.

Astro-Han:
The `test` failure here is not caused by anything in this PR. The build step fails at `src/__tests__/execution-host-message.test.ts(101,54): error TS2554: Expected 1 arguments, but got 2.`

#3277 (`a6f33c952`, "retire client surface identity plumbing") landed after this head and removed `connectClient`'s second `surface` parameter, while the new test added here still calls `connectClient(fixture.root, 'tui')`. A rebase onto `main` and dropping the `'tui'` argument should be the whole fix.

My approval stands — the diff I reviewed is unaffected.

*Written with Claude Code; the CI log, #3277's diff and the current `connectClient` signature were read at this head and verified by me.*

packages/runtime-host/src/server/root-turn-coordinator.ts
me2seeks: Done in 2225aea39 — renamed to `isRuntimeSessionForwardedEvent` / `RuntimeSessionForwardedEvent` (with a comment stating the forwarding rule and why the durable `steering_message` belongs), and renamed the client-side `projectToolEvent` to `projectSessionEvent` to match the Host side. Tests re-run: 103/103 across the four affected suites.

packages/runtime-host/src/__tests__/execution-host-message.test.ts
```diff
@@ -95,6 +95,64 @@ import {
   withTimeout,
 } from './fixtures/execution-host-suite.js';
 
+test('subscribed Clients receive the durable steering echo as a session event', async () => {
+  await withExecutionRoot(async (fixture) => {
+    const host = await fixture.startHost();
+    const client = await connectClient(fixture.root, 'tui');
```
Astro-Han: [P1] Rebase onto `main` and drop the `'tui'` argument — this line is why the required `test` check cannot pass. `connectClient` now takes one parameter (`execution-host-suite.ts:1043`, `connectClient(rootPath: string)`); #3277 (`a6f33c952`, "retire client surface identity plumbing") removed the second `surface` parameter after this branch was cut. The build step fails at `src/__tests__/execution-host-message.test.ts(101,54): error TS2554: Expected 1 arguments, but got 2.` I flagged this as external to your diff on the previous head, and that was fair then; at this head the branch has moved without picking up `main`, so it is now this PR's to resolve, and it is the only thing standing between the change and a green check. Evidence: confirmed by reading code at this head — I compared the call site here against the fixture's current signature on `main`. No new regression test is needed; the existing suite covers the behaviour once it compiles.

Astro-Han (commented):
Reviewed the delta from `18e2905f9` — one commit, three files, a pure rename. It is the right response to the naming concern and I have no objection to it.

`RuntimeSessionTransientEvent` → `RuntimeSessionForwardedEvent` and `isRuntimeSessionTransientEvent` → `isRuntimeSessionForwardedEvent` make the predicate say what it decides instead of implying a durability class it never controlled, and the new comment states the actual rule — forward live to subscribers via the continuity coordinator rather than letting the canonical refresh carry it, with persistence orthogonal and handled upstream in the run's own event stream. That is exactly the distinction that made `steering_message`'s membership look wrong under the old name, and it is now settled by the name rather than by having to re-derive it. `projectToolEvent` → `projectSessionEvent` in the projector is the same correction one layer down; the function has never been tool-specific.

I checked the rename is complete rather than partial: zero occurrences of either old identifier remain in the three touched files at this head, and the two module-local `projectSessionEvent` functions live in different modules at different layers, so there is no shadowing.

The one blocker is unchanged and is now inside this PR's scope — see the thread below. My earlier approval was dismissed by the new commit, so this is a fresh review; the P1 is the only thing keeping it a COMMENT.

Worth flagging separately: CI has not run on this head at all. The run is sitting at `action_required` (id `32367571694`), which is the fork-contributor workflow-approval gate, not a failure. A committer has to release it before any check result exists — so the rebase alone will not produce a green check without that.

*AI disclosure: reviewed with Claude Code. I read the delta and the current `connectClient` signature at this head myself, confirmed the rename left no stale identifiers, and checked the workflow-run state directly. The conclusions are mine.*

packages/runtime-host/src/adapter/session-projector.ts
```diff
@@ -320,8 +328,16 @@ export class RuntimeHostSessionProjector {
       return emptyUpdate(events);
     }
     if (frame.kind === 'subscription.session_event') {
-      const event = projectToolEvent(frame);
-      if (event) events.push(event);
+      const event = projectSessionEvent(frame);
+      if (event) {
```
likun666661: [P1] Dedup the forwarded echo against the transcript bootstrap as well. The runtime persists a `steering_message` before the Host forwards it, so `subscription.open` can bootstrap message X and install the subscriber before `acceptRuntimeEvent` publishes the live frame for X. A fresh projector starts `#renderedSteeringMessageIds` empty; `#transcriptIds` only suppresses queue synthesis, and this branch does not consult it. The pending live frame therefore emits X again, while the TUI reducer unconditionally appends another user prompt. This is possible even when X is still `in_flight`: `seedActive` skips it because it is durable but does not mark it rendered. Please seed/check the durable steering identities (or establish a non-overlapping transcript/live high-water boundary), and add a regression where the bootstrap already contains X and the first pending frame is the forwarded echo for X.

packages/runtime-host/src/__tests__/execution-host-message.test.ts
me2seeks: Done — branch rebased onto main (`5d8605ab0..6397a5677`) and the new e2e test updated to the single-argument `connectClient` from #3277. Full rebuild of the dependency chain is clean, and the affected suites pass: projector/protocol/coordinator/root-turn 134/134, execution-host message/queue e2e 14/14.

packages/runtime-host/src/adapter/session-projector.ts
me2seeks: Fixed in 6397a5677 — dedup now covers the bootstrap path too, not just the two live paths.

`RuntimeHostSessionProjectionSeed` gained `durableSteeringMessageIds`: the messageIds of transcript user messages carrying `steeringEventId` (the read model assigns the submitted steering messageId as the stored message id via `stableMessageId` → `refs.providerEventId`). The projector seeds `#renderedSteeringMessageIds` from them in the constructor, so all three render paths — bootstrap render, queue in-flight synthesis, and the forwarded echo — share one dedup set keyed by messageId, and whichever projects the message first wins.

Regression added exactly on the requested shape (`session-projector.test.ts`): a projector seeded with a transcript containing X, with X still `in_flight` — `seedActive` emits no synthesis for X, the first pending frame is the forwarded echo for X, and it is suppressed; a different steering message still renders once. All suites pass (134/134 unit + 14/14 e2e).

packages/runtime-host/src/__tests__/execution-host-message.test.ts
me2seeks: Fixed — rebased onto current main (`be9d908f0`); the branch now compiles against the single-parameter `connectClient` from #3277, and the e2e suite passes on the rebuilt head.

packages/runtime-host/src/adapter/session-projector.ts
me2seeks: Also fixed at the rebased head — `durableSteeringMessageIds` now seeds `#renderedSteeringMessageIds` from transcript user messages carrying a steering event id, so the forwarded echo is suppressed whenever the bootstrap already rendered it, and the regression (bootstrap contains X + first pending frame is X's echo) is covered in `session-projector.test.ts`. Full suites pass on the rebuilt head: runtime-host 1031/1031, desktop main 988/988.

packages/runtime-host/src/protocol/session-continuity.ts
```diff
@@ -50,7 +55,7 @@ import {
   type SessionTranscriptBootstrap,
 } from './session-transcript.js';
 
-export const SESSION_CONTINUITY_SCHEMA_VERSION = 4 as const;
+export const SESSION_CONTINUITY_SCHEMA_VERSION = 5 as const;
```
Astro-Han, line 58: **[P1] `test` is red on this head, and the gate is right.**

```
Protocol files changed but RUNTIME_HOST_COMPATIBILITY_EPOCH is still 39
```

This PR takes `SESSION_CONTINUITY_SCHEMA_VERSION` from 4 to 5 and makes `decodeSessionContinuitySnapshot` reject anything that isn't 5, so an older client fails at open. That's a deliberate compatibility break and the right call — but `RUNTIME_HOST_COMPATIBILITY_EPOCH` in `protocol/index.ts:94` is still `39`, the same value as `main`.

Worth noting this isn't an assertion mismatch, so the usual "`+` is actual, `-` is expected" reading doesn't apply — the gate is asking for an epoch the current base hasn't seen, and it hasn't got one.

Bumping the epoch alongside the schema is all that's needed. It's the piece that lets the host tell an incompatible client apart from a merely older one, which is exactly the situation this schema bump creates.

Astro-Han (commented):
The diagnosis here is the good kind — the durable ledger always had the steering message, so nothing was lost at rest; what was missing was the subscriber hop. Forwarding the durable event over the existing `subscription.session_event` channel rather than inventing a new one is the right shape, and de-duplicating by `messageId` across bootstrap, in-flight synthesis and live echo is what makes the three paths safe to keep. Keeping the synthesis path is also correct — it still covers the attach/rejoin window where no echo arrives.

One blocker, and it's the housekeeping that a protocol change always drags along. Inline.

I checked the two earlier P1s against this head and both are closed: `connectClient` is single-argument now, and the bootstrap de-duplication landed with test coverage. I didn't re-raise them.

On the three things I'd normally worry about for a forwarding change — delivery failure, reconnect, and double delivery — the state passes through queue lease → kernel persist → RuntimeEvent with `refs.providerEventId = messageId` → pump → subscription frame → projector, and the `messageId` identity survives end to end, which is what makes the de-duplication sound. With no subscriber the event simply isn't enqueued and stays in the run stream. `acceptRuntimeEvent` does throw when the canonical root turn doesn't match, which interrupts the pump — but that's the same door `text_delta` already goes through, so it isn't a seam this PR opens.

me2seeks:
Rebased onto current `main` (`6ada6b546`+, 3 commits replayed cleanly). The CI failure you flagged is resolved by the rebase itself: the branch's `execution-host-message` test no longer passes a surface argument to `connectClient`, and the full clean rebuild + runtime-host suite is green on the new head (1090/1090, biome clean).

packages/runtime-host/src/protocol/session-continuity.ts
Astro-Han, line 58: **[P1] The epoch gate is red on this exact head, and it is the same open item from the 2026-08-22 review.**

The CI failure on this head:

```
Protocol files changed but RUNTIME_HOST_COMPATIBILITY_EPOCH is still <base parent's value>
```

This PR changes a protocol file (`session-continuity.ts`, schema 4 -> 5, new event variant) but leaves `RUNTIME_HOST_COMPATIBILITY_EPOCH` at the base's value. `main` has since moved to **44**, so after rebasing onto current `main` the epoch must be set past 44 (i.e. 45), with a comment naming this change - not merely rebased onto the old number. The guard exists precisely because same-number sibling bumps merge cleanly and then advertise one epoch for two incompatible protocols (#3313).

Everything else in the wire change looks right (strict exact-keys decode, bounded content, reject-at-open on schema mismatch), so this is the one blocking item.

Astro-Han (commented):
Reviewed at exact head `323fb3220eb6722853f9ac185ec491deef75911f`. **NO-GO at this head**: one `[P1]` inline (the epoch gate), and `test` is terminal **failure** on this exact head, so the approve condition fails on both findings and checks.

This PR has two prior review rounds (2026-08-20 approved at `18e2905f`, later dismissed on head movement; 2026-08-22 commented at `be9d908f`). The 08-22 round's open item was exactly the epoch bump; the head has since been rebased (now on base `6ada6b54`) but the epoch is still untouched, and the red check is the proof.

This pass re-verified the substance independently, on this head:

- **The identity chain behind the de-duplication holds.** Transcript `user` message id comes from `stableMessageId` = `refs.storedMessageId ?? refs.providerEventId` (`runtime-event-read-model.ts`), the durable echo carries `messageId = lease.messageId`, and the queue in-flight entries are keyed by the same `messageId` - so the three-way dedup (`#renderedSteeringMessageIds`) compares like with like across bootstrap, synthesis, and echo.
- **The new projector tests cover the interesting races**: echo when the coalesced refresh skipped `in_flight`, exactly-once in both directions, rejoin seed, and bootstrap suppression of a late echo. I read them against the production id sources and they match.
- **The wire change is bounded**: `SessionSteeringEvent` decodes with exact keys, ids/counts validated, and content through `decodeMessageContent` (text + whole-content byte caps). `SESSION_CONTINUITY_SCHEMA_VERSION` 4 -> 5 with reject-at-open is the established convention.
- **The rename** (`RuntimeSessionTransientEvent` -> `RuntimeSessionForwardedEvent`) with the corrected comment (forwarding vs persistence being orthogonal) reads accurately against the pump code.

**Overlap with #3633** (asked by the orchestrator): the two PRs target the same symptom (#3304 rendering of mid-turn steering) and the same files (`session-projector.ts`, `root-turn-coordinator.ts`, `execution-host-message.test.ts`), but in opposite directions - this PR is additive (forward the durable echo, keep the in-flight synthesis as the rejoin fallback), while #3633 is subtractive (make admission durable, remove the synthesis, remove Desktop's steering IPC). They are complementary rather than competing: whichever lands first, the other must rebase over it. Concretely, if this PR lands first, #3633's rebase would delete the synthesis leg of this PR's three-way dedup (leaving bootstrap + echo, which is exactly #3633's world), and #3633 would inherit this PR's wire schema 5 and epoch. The reverse order leaves this PR's echo as a redundant-but-deduped second render path over #3633's admission durability. Neither order invalidates the other PR's architecture; both orders force a nontrivial rebase of the later one. The bigger practical point: this PR's remaining blocker is mechanical (epoch + rebase), while #3633's NO-GO is substantive (auto-recovery on an unreachable path), so landing this one first does not strand any fix #3633 needs.

Gate: `test` failed on this exact head (protocol epoch guard). Path-filtered: `test` is the applicable check; it must be green on the eventual rebased head.

me2seeks:
Rebased onto current `main` (`c79e9eb4`). Epoch is now **45** with the continuity-schema rationale recorded in the epoch history (`591d5ff7`); the epoch-guard step passes against the new base. The steering-echo suppression tests from the earlier review round are retained and green (session-projector: 11/11).

jackwener (commented):
Independent review of `591d5ff70c6d2d224c0eb0d2de310781c1cb46c8`.

**GO. No P0–P2.** MERGEABLE / BLOCKED. I am not approving: this head has no terminal-green `test` run (`CI` is `action_required`, typical of a fork PR). Current `main` `test` is green, so this is not a main-is-red false alarm.

## What this solves

Steering sent mid-turn could be consumed by the model and still vanish from the UI, because the Host only forwarded transient tool/text events and the coalesced canonical refresh can skip `in_flight`. Forwarding the durable `steering_message` on `subscription.session_event`, and deduping it against queue synthesis **and** transcript bootstrap (`durableSteeringMessageIds`), is the right cut.

## Epoch

Schema 4→5 is a closed-shape change: older peers reject an unknown `schemaVersion` at admission. That **must** bump the compatibility epoch; a `protocol-compatible-changes/` declaration would be wrong. This head does bump it. After any rebase onto current main, read `RUNTIME_HOST_COMPATIBILITY_EPOCH` there and take a number greater than that — several open PRs currently share the same next slot.

Prior P1s on this PR (epoch gate, bootstrap dedup, `connectClient` arity) independently re-judged **closed** on this head. No inline findings. No merge from me.

Astro-Han (commented):
Review at `591d5ff7`. **No P0–P3 on the code.** Not approving only because this head has no current-base CI evidence.

The ordering question this change turns on is answered correctly: the steering echo is made durable before it is forwarded (`packages/runtime/src/ai-sdk-backend.ts:4532-4547`), and the host projector dedupes by `messageId` across the live, replay and transcript paths (`packages/runtime-host/src/adapter/session-projector.ts:361-401`), so a reconnect does not double-deliver.

On the red: it is **inherited, not attributable to this PR**. The failure is `packages/storage/src/__tests__/codex-session-adapter.test.ts:267` referencing `decodeStoredMessage`, which no longer exists. This PR's merge-base is `c79e9eb4`, which predates #3656 (`3bb645e9`, *test(storage): use the canonical message decoder*); current `main` has zero references to that symbol while the merge-base still has one. The file is outside this PR's diff.

Note that a re-run cannot clear it: re-running replays the same merge ref rather than recomputing against the repaired base. A rebase onto current `main` is what this needs, after which the checks should speak for the change itself.

<details>
<summary>中文</summary>

在 `591d5ff7` 上审，**代码面无 P0–P3**。不 approve 的唯一原因是本 head 缺少基于当前 base 的 CI 证据。

这个改动最关键的时序问题答对了：steering echo 先落持久化再转发（`ai-sdk-backend.ts:4532-4547`），且 host projector 在实时、重放、transcript 三条路径上都按 `messageId` 去重（`session-projector.ts:361-401`），重连不会重复投递。

关于红：它是**继承来的，不归本 PR**。失败点是 `codex-session-adapter.test.ts:267` 引用了已不存在的 `decodeStoredMessage`。本 PR 的 merge-base 是 `c79e9eb4`，早于 #3656（`3bb645e9`，test(storage): use the canonical message decoder）；当前 `main` 上该符号出现 0 次，而 merge-base 上仍有 1 次，且该文件不在本 PR 的 diff 内。

注意重跑清不掉它：重跑只会重放同一个 merge ref，不会针对修好的 base 重算。需要 rebase 到当前 `main`，之后 checks 才能真正说明这个改动本身。
</details>

Astro-Han (commented):
I reviewed this head. No blocking P0-P2.

The durable steering echo direction is correct — reusing ordered `session_event` forwarding with `messageId` deduplication. One non-blocking P3: `session-projector.ts:103` `renderedSteeringMessageIds` is never reclaimed across Turns and will grow with continuous steering; should be bounded by queue/root lifetime.

Machine gate: `test` is `failure` due to 3 PR-owned TypeScript errors in tests (`SteeringMessageSnapshot` imports / renamed `RuntimeSessionTransientEvent`), so the head does not build — not green.

<details><summary>简体中文</summary>该头无阻断，P3 为内存增长小项，但编译未通过。</details>

Astro-Han (approved):
LGTM — per @AstroHan merge authorization (verified admin). Code GO with test/windows_recovery green; approving exact head d50a046.

---

PR #3255, opened 2026-08-19
feat(providers): add Alibaba Token Plan Responses compatibility

MoonOld:
## Summary

Adds Alibaba Token Plan Responses support for `qwen3.8-max` while keeping the other Token Plan models on Chat Completions.

Core owns only protocol facts and explicit delegation:

- `openAiAdapterApiProtocol` declares the Alibaba Token Plan `qwen3.8-max → openai-responses` wire;
- both Token Plan registry entries reference the stable `alibaba-token-plan` Runtime profile;
- the adapter type forbids declaring both a Core `responses` contract and a Runtime profile.

Runtime owns the concrete execution profile:

- selects `@ai-sdk/open-responses@2.0.34`;
- forces `store: false`, so continuation never depends on a provider-stored response ID;
- preserves Alibaba's documented `required` and single-tool `allowed_tools` choices while rejecting unsupported forced shapes;
- consumes both the official `reasoning_summary_text` and regional `reasoning_text` carriers through the upstream SDK mapping;
- treats `output_item.done.item.summary` as the durable identity and summary-boundary authority;
- persists bounded local item identity and summary parts for restart-safe replay;
- isolates invalid, late, interrupted, or mismatched reasoning from finalized durable state;
- preserves provider failure classifications without widening retry eligibility;
- leaves DeepSeek on its existing plaintext-content replay path.

Fixes #3162

## Upstream resolution

The temporary Maka raw-event normalizer is no longer needed:

- vercel/ai#19659 reported the missing official summary-delta mapping;
- vercel/ai#19661 fixed both `reasoning_summary_text.delta` and the regional `reasoning_text.delta` carrier;
- the fix shipped in `@ai-sdk/open-responses@2.0.34`.

This head upgrades to 2.0.34 and deletes the workaround, including the 328-line production normalizer and its 241-line dedicated test. The functional cleanup commit is 65 insertions and 633 deletions; the follow-up commit refreshes the generated Desktop/CLI legal notices for the new dependency graph.

## Verification

Exact head `bac63a56b`, rebased onto `main` at `a956b1ae0`:

- fresh root `npm ci` completed, reapplied repository patches, and reported 0 vulnerabilities;
- full workspace `build:test` and full workspace typecheck passed;
- `npm run format:check`, `npm run lint`, `npm run check:asf-headers`, and `git diff --check` passed;
- the complete `npm run check:release` contract, including regenerated third-party notices, passed (110 release tests);
[...]

PR #3321, opened 2026-08-20
fix(runtime): emit camelCase OpenAI-compatible provider option keys

### LINKED ISSUE #1430: fix(runtime): use camelCase OpenAI-compatible provider option keys
## Problem

After the AI SDK v7 upgrade in #1198, requests to Z.AI Coding Plan emit repeated deprecation warnings:

```text
AI SDK Warning (zai-coding-plan.chat / glm-5.2): Deprecated: "providerOptions key 'zai-coding-plan'". Use 'zaiCodingPlan' instead.
```

The request still works because `@ai-sdk/openai-compatible@3.0.12` currently accepts both forms, but it emits one warning for every model invocation. Agent/tool loops therefore produce the warning repeatedly.

## Cause

`buildProviderOptions` currently emits the raw provider name as the options namespace:

```ts
{ 'zai-coding-plan': { reasoningEffort: 'high' } }
```

`@ai-sdk/openai-compatible` now expects camelCase provider option keys:

```ts
{ zaiCodingPlan: { reasoningEffort: 'high' } }
```

The current regression test explicitly pins the deprecated dashed namespace:

- `packages/runtime/src/model-factory.ts`
- `packages/runtime/src/__tests__/model-factory-thinking.test.ts`

This may also affect other OpenAI-compatible providers whose configured names contain hyphens.

## Expected behavior

OpenAI-compatible provider options use the SDK's current camelCase namespace without changing the provider ID, model routing, or wire-level request fields.

## Acceptance criteria

- `zai-coding-plan` provider options are emitted under `zaiCodingPlan`.
- `reasoningEffort` still reaches the request as `reasoning_effort`.
- Dashed OpenAI-compatible provider names are handled consistently rather than special-casing only Z.AI.
- Regression coverage exercises the current AI SDK parsing path and confirms that no deprecation warning is produced.

yihanzhu:
## Summary

Every doGenerate through a built-in dashed OpenAI-compatible provider (`zai-coding-plan`, `volcengine-ark`, …) logs an AI SDK deprecation warning: options are keyed by the raw provider name, while the SDK's canonical key is its camelCase alias. This derives the key via the #2463 `toCamelCase` helper for all openai-compatible identities and fixes the three direct dashed emissions (`volcengine-ark`, `cloudflare-workers-ai`, the `github-copilot` chat lane). The open-responses path is unchanged — that SDK resolves only the raw name.

Fixes #1430

## Verification

- `packages/runtime` suite: 2977 tests pass; with the fix reverted, 8 fail (the updated namespace pins plus a new doGenerate capture test asserting `reasoning_effort` reaches the request body with zero `deprecated` warnings)
- Root `typecheck`, `lint`, and `format:check` pass
- Unrelated: `node-pty-write-lifecycle` times out on my machine (spawnSync ETIMEDOUT) — reproduced identically on unmodified main

## AI use

Select exactly one:

Tool(s) and scope: Claude Code (Fable 5) — implementation and tests, under my direction and review. The commit carries a Generated-by trailer.

## Checklist

- [ ] No

### DIFF
--- a/packages/runtime/src/__tests__/model-factory-thinking.test.ts
+++ b/packages/runtime/src/__tests__/model-factory-thinking.test.ts
@@ -642,5 +642,53 @@ describe('buildProviderOptions: openai-compatible namespace', () => {
       false,
       JSON.stringify(result.warnings),
     );
+    // The options key also selects the SDK's response metadata namespace:
+    // metadata must come back under the camelCase alias, not the dashed name.
+    assert.deepEqual(Object.keys(result.providerMetadata ?? {}), ['zaiCodingPlan']);
+  });
+
+  test('passthrough provider options reach the chat request body without deprecation', async () => {
+    // reasoningEffort above travels the SDK's schema lane, which parses both
+    // spellings. Volcengine Ark's `thinking` object is not in the schema and
+    // travels the passthrough spread instead — pin that lane at the wire too.
+    const bodies: Record<string, unknown>[] = [];
+    const captureFetch: typeof globalThis.fetch = async (_input, init) => {
+      bodies.push(JSON.parse(String(init?.body)) as Record<string, unknown>);
+      return new Response(
+        JSON.stringify({
+          id: 'chatcmpl-1',
+          object: 'chat.completion',
+          created: 1,
+          model: 'doubao-seed-2-0-pro-260215',
+          choices: [
+            {
+              index: 0,
+              message: { role: 'assistant', content: 'ok' },
+              finish_reason: 'stop',
+            },
+          ],
+          usage: { prompt_tokens: 1, completion_tokens: 1, total_tokens: 2 },
+        }),
+        { status: 200, headers: { 'content-type': 'application/json' } },
+      );
+    };
+    const connection = conn('volcengine-ark', 'volcengine-ark');
+    const model = getAIModel({
+      connection,
+      apiKey: 'ark-key',
+      modelId: 'doubao-seed-2-0-pro-260215',
+      fetch: captureFetch,
+    });
+    const result = await model.doGenerate({
+      prompt: [{ role: 'user', content: [{ type: 'text', text: 'hi' }] }],
+      providerOptions: buildProviderOptions(connection, 'doubao-seed-2-0-pro-260215'),
+    });
+    assert.equal(bodies.length, 1);
+    assert.deepEqual(bodies[0]?.thinking, { type: 'enabled' });
+    assert.equal(
+      (result.warnings ?? []).some((warning) => warning.type === 'deprecated'),
+      false,
+      JSON.stringify(result.warnings),
+    );
   });
 });
--- a/packages/runtime/src/__tests__/model-factory-thinking.test.ts
+++ b/packages/runtime/src/__tests__/model-factory-thinking.test.ts
@@ -264,7 +264,7 @@ describe('buildProviderOptions: thinking level', () => {
     });
     // Copilot defaults to its OpenAI-compatible chat wire without a protocol hint.
     assert.deepEqual(buildProviderOptions(conn('github-copilot'), 'gpt-5.4', 'high'), {
-      'github-copilot': { reasoningEffort: 'high' },
+      githubCopilot: { reasoningEffort: 'high' },
     });
   });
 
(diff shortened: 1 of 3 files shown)


packages/runtime/src/model-factory.ts
```diff
@@ -612,18 +612,15 @@ function toCamelCase(name: string): string {
...
+ * every doGenerate result), so the camelCase alias is the canonical key.
  */
 function openAiCompatibleProviderOptionsKey(
   adapter: ProviderRuntimeAdapter,
   connection: RuntimeExecutionConnection,
 ): string {
-  return adapter.kind === 'openai-compatible' && adapter.name === 'connection'
-    ? toCamelCase(connection.slug)
-    : connection.providerType;
+  return toCamelCase(openAiCompatibleProviderName(adapter, connection));
```
Astro-Han, line 631: [P3] Note in this docstring that the change also moves the SDK's *response* metadata namespace, because that is the one consequence the PR body does not mention. `@ai-sdk/openai-compatible@3.0.30` derives `metadataKey = resolveProviderOptionsKey(rawName, providerOptions)` (`dist/index.js:540`), which returns the camelCase alias whenever the camelCase key is the one present. So once options are keyed `zaiCodingPlan`, the provider metadata the SDK writes back — `acceptedPredictionTokens`, `rejectedPredictionTokens`, `thoughtSignature` (`dist/index.js:635-652`, `:764`) — moves from `providerMetadata['zai-coding-plan']` to `providerMetadata.zaiCodingPlan` too. I checked and this breaks nothing today: the only metadata readers in the tree key off fixed literals — `anthropic` and `openai` in `model-adapter.ts:714-731`, `zenmux` in `reasoningDetailsFromMetadata`, and `openaiCompatible` in `withReasoningDetails` — so no reader is keyed by `connection.providerType`. It is worth one sentence anyway, because the natural way to write a future metadata reader is to key it by the provider type, and that reader would silently return undefined for exactly the dashed providers this PR targets. Evidence: confirmed by reading the pinned SDK and grepping every `providerMetadata` access in `packages/runtime` and `packages/runtime-host` at this head. Regression test: assert the round trip in the new capture test — that `result.providerMetadata` carries the camelCase key — so the coupling is pinned rather than implied.

packages/runtime/src/__tests__/model-factory-thinking.test.ts
```diff
@@ -599,4 +599,48 @@ describe('buildProviderOptions: openai-compatible namespace', () => {
       JSON.stringify(result.warnings),
     );
   });
+
+  test('built-in dashed provider effort reaches the chat request body without deprecation', async () => {
```
Astro-Han, line 603: [P3] Add a wire-level case for a passthrough field, not just for `reasoningEffort` — the passthrough providers are the riskier half of this change and they are the half without request-body coverage. `reasoningEffort` is a *schema* option: the SDK parses `providerOptions` three times, under `"openaiCompatible"`, the raw name and the camelCase alias (`dist/index.js:505-519`), and merges them, so this test exercises the well-covered path. `volcengine-ark`'s `thinking: { type }` and `cloudflare-workers-ai`'s `chat_template_kwargs` are not in the schema and travel a different path — the spread at `dist/index.js:569-577`, which merges both spellings and then filters out anything matching the chat-options shape. Both spellings do work there; I read that spread at the pinned version to confirm it, and that is the only reason the two direct emissions this PR rewrites are safe. But the tests for those two providers only pin the intermediate `buildProviderOptions` object, so nothing in the suite would catch the SDK dropping the camelCase arm of that spread in a future minor. Evidence: confirmed by reading the pinned SDK at this head; not executed. Regression test: one more capture in this shape against `volcengine-ark`, asserting `thinking` reaches the request body and that no `deprecated` warning is emitted.

Astro-Han (dismissed):
Approving. The change is correct, and it is correct for a reason worth stating: it draws the raw-versus-camelCase line exactly where the two SDKs actually differ, rather than applying one rule everywhere.

I verified the substance rather than taking the premise on trust:

- The `toCamelCase` added here is semantically identical to the SDK's own (`@ai-sdk/openai-compatible@3.0.30`, `dist/index.js:19-21`) — same `/[_-]([a-z])/g` regex, same replacement. So the emitted key always matches the alias the SDK resolves, including the edge cases where they *both* decline to transform, such as a relay slug like `relay-1` where the character after the dash is not a lowercase letter.
- Both option paths resolve either spelling, so nothing is dropped. Schema options are parsed three times and merged — under `"openaiCompatible"`, the raw name and the camelCase alias (`dist/index.js:505-519`). Passthrough fields are spread from both keys and then filtered against the chat-options shape (`dist/index.js:569-577`). That second one is what makes `volcengine-ark`'s `thinking` and `cloudflare-workers-ai`'s `chat_template_kwargs` safe to rekey, and it is the less obvious of the two.
- The deprecation warning fires precisely when the raw key is present and differs from its alias (`warnIfDeprecatedProviderOptionsKey`, `dist/index.js:29-40`), so removing the dashed key removes the warning at its cause rather than suppressing it.
- Leaving `model-factory.ts:543` on the raw name is right, not an oversight. `@ai-sdk/open-responses@2.0.28` sets `providerOptionsName: providerName` verbatim (`dist/index.js:936`) and never camelCases, so the Responses lane must stay raw. The comment already says so; I confirmed it against the pinned package.
- The three providers whose namespaces change are all constructed through `createOpenAICompatible` — `github-copilot` explicitly at `model-factory.ts:116-117`, and `volcengine-ark`, `zai-coding-plan` and the rest via the generic path at `:189` with `name: openAiCompatibleProviderName(...)`, since their registry adapters are `{ kind: 'openai-compatible', name: 'provider' }`. So every rekeyed identity is one the SDK aliases.

The new capture test is the right kind: it asserts `reasoning_effort` in the actual request body *and* that no `deprecated` warning comes back, which is the defect this PR closes rather than a proxy for it.

Two P3s below — a docstring note and a test gap. Neither blocks; both are about making the coupling to the SDK explicit rather than incidental.

One coordination note, not a finding. `packages/runtime/src/model-factory.ts` is contested right now: #3255 and #3300 both touch `buildFamilyWire`'s `openai-compatible` case, and #3255 introduces a third provider-key resolver in `model-runtime.ts` that must stay raw for the Responses lane. Whichever of these lands last should re-check that the raw and camelCase resolvers are still distinguished on purpose — the refactor here makes `openAiCompatibleProviderName` the single raw source, which is the right base for that reconciliation.

*AI disclosure: reviewed with Claude Code. I read the pinned `@ai-sdk/openai-compatible@3.0.30` and `@ai-sdk/open-responses@2.0.28` sources myself to confirm the alias resolution, the passthrough spread and the deprecation trigger, checked each rewritten provider's registry adapter, and grepped every `providerMetadata` reader in the runtime packages. The approval is mine and rests on those checks.*

packages/runtime/src/model-factory.ts
yihanzhu: Added the docstring note and pinned providerMetadata under zaiCodingPlan in 4d98710.

packages/runtime/src/__tests__/model-factory-thinking.test.ts
yihanzhu: Added the volcengine-ark passthrough capture in the same commit.

likun666661 (approved):
Reviewed the current head. The issue is correctly scoped to the deprecated OpenAI-compatible provider-options namespace; the Chat camelCase alias and Responses raw-name paths remain intentionally distinct. Wire-level coverage preserves both schema and passthrough fields, and the providerMetadata namespace change has no current raw-provider consumer. Approved.

PR #3322, opened 2026-08-20
test(runtime-host): guard v0.1.11 handshake compatibility

### LINKED ISSUE #3297: chore(runtime-host): retire the legacy surface hello shim after epoch 27
## Summary

PR #3277 removes `ClientSurface` from the Runtime Host protocol and connection plumbing, while retaining one private outgoing hello shim that sends `surface: "desktop"` for pre-#3277 Hosts.

Track removal of that shim once the compatibility boundary is observable on the wire.

## Removal gate

Remove the shim only when all of the following are true:

- the minimum supported Runtime Host compatibility epoch is greater than 27;
- the removal change bumps the Client compatibility epoch accordingly;
- mixed-version coverage verifies that unsupported older Hosts reach the structured `incompatible` / restart-guidance path rather than failing with a bare transport abort.

Epoch 27 cannot distinguish a pre-#3277 Host that requires `surface` from a post-#3277 Host that tolerates its absence, so release notes alone are not a safe removal signal.

## Related

- Follow-up to #3216
- Compatibility shim introduced by #3277
- Released-Host compatibility guard added by #3322

AI assistance disclosure: Codex helped analyze the mixed-version failure mode and draft this tracking issue.

Sun-GLiang:
## Summary

- Retain the private outgoing `surface: "desktop"` hello sentinel because the released v0.1.11 Host (compatibility epoch 25) requires it before compatibility negotiation can run.
- Add a frozen v0.1.11 decoder/admission regression that drives the raw socket and Client handshake through the structured `incompatible` path.
- Pin CLI restart-guidance coverage to the released epoch-25 boundary.
- Leave #3297 open until the minimum supported tagged Host has the tolerant decoder required by its removal gate.

Closes #3297.

## Verification

- `npm run typecheck`
- `npm run lint`
- `npm run format:check -- --diagnostic-level=error`
- `npm --workspace @maka/runtime-host test` (1037 passed)
- `npm --workspace maka-agent test` (336 passed)
- `git diff --check`

## AI use

Select exactly one:

Tool(s) and scope: Codex analyzed the released mixed-version failure, retained the compatibility sentinel, added the version-pinned regression coverage, and verified the final diff.

## Checklist

- [x] No


packages/runtime-host/src/protocol/index.ts
```diff
@@ -72,7 +72,10 @@ export const RUNTIME_HOST_REGISTRATION_SCHEMA_VERSION = 1 as const;
 export const RUNTIME_HOST_PROTOCOL_VERSION = 0 as const;
 // Increment when the same protocol version no longer guarantees safe Client-Host
 // interoperability. Mismatches are rejected before domain commands are admitted.
-export const RUNTIME_HOST_COMPATIBILITY_EPOCH = 27 as const;
+export const RUNTIME_HOST_COMPATIBILITY_EPOCH = 28 as const;
+// 28: Clients stop sending the retired surface identity in their bootstrap
```
Astro-Han: [P1] Establish that no epoch-27 Host predating #3277 can still be running, or keep the shim one more epoch — this comment's guarantee holds only for the tolerant half of epoch 27. Epoch 27 has two distinct Host flavours, because #3277 (`a6f33c952`, merged today) removed the surface plumbing without bumping the epoch. Against a **post**-#3277 epoch-27 Host the guarantee is real, and the new test proves it. Against a **pre**-#3277 epoch-27 Host it is not: that Host's `decodeClientFrame` calls `requireSurface(frame.surface)` while decoding the bootstrap hello, and `requireSurface` throws `invalidProtocolFrame` on `undefined`. That throw happens in `#serveConnection` at the first statement of the `try`, strictly before `#admitHandshake` — which is where the `hello.compatibilityEpoch !== RUNTIME_HOST_COMPATIBILITY_EPOCH` comparison lives — so the structured `kind: 'incompatible'` frame is never constructed. The `catch { transport.abort(); }` fires instead, the Client sees a transport abort with no response frame, and the CLI restart guidance that `runtime-host-cli-context.test.ts` covers never runs. The user gets an opaque protocol error on exactly the upgrade path the epoch mechanism exists to make legible. This is also precisely the precondition the shim you are deleting stated for its own removal — "keep the sentinel private until the minimum supported compatibility epoch is greater than 27" — and bumping the Client to 28 raises the Client's epoch, not the minimum Host epoch that can still be running; the Host check is a strict equality, so there is no minimum-supported ladder to raise. Evidence: confirmed by reading code — I diffed the decoder at `a6f33c952^` against `main` and traced the decode-before-admit ordering in `host-kernel.ts`. Reachability is a release-timing question I cannot settle from the repo: if no build between the epoch-27 bump and #3277 was ever shipped or is ever left running, this is a non-issue and the comment should say so. Regression test: the mirror of the new tolerant-Host test — a pinned `a6f33c952^` intolerant epoch-27 Host — asserting whichever outcome you decide is correct.

packages/runtime-host/src/protocol/index.ts
```diff
@@ -72,7 +72,10 @@ export const RUNTIME_HOST_REGISTRATION_SCHEMA_VERSION = 1 as const;
 export const RUNTIME_HOST_PROTOCOL_VERSION = 0 as const;
 // Increment when the same protocol version no longer guarantees safe Client-Host
 // interoperability. Mismatches are rejected before domain commands are admitted.
-export const RUNTIME_HOST_COMPATIBILITY_EPOCH = 27 as const;
+export const RUNTIME_HOST_COMPATIBILITY_EPOCH = 28 as const;
```
Astro-Han: [P2] Expect to renumber this before merge — five open PRs currently claim epoch 28. `main` is at 27, and #2521, #2641, #3236, #3309 and this PR all take 28, while #3299 takes 29. I verified each at its current head just now. The failure mode is quiet rather than loud: every branch writes the identical `= 28` line, so Git merges them without a conflict, and each PR's own lower-bound assertion (`EPOCH > 26` and similar) still passes afterwards. Whichever of the five lands second therefore ships a distinct incompatibility under a shared epoch number, which is exactly the condition the epoch exists to prevent — two Client/Host pairs that both say 28 while disagreeing about the wire. Nothing in CI catches it, because no check compares this constant against the other open branches. Evidence: confirmed by reading each PR's head at this moment. This is not a defect in your diff and it should not hold the PR up; it does mean the number here is not yet settled. #3313 tracks removing the class of problem.

Astro-Han (commented):
Reviewed at `a4e747d68`. Small, well-shaped change, and the fixture design is the right instinct — pinning a historical Host by revision and keeping its decoder independent of the current implementation is how a compatibility test should be written, and it is why the tolerant-Host case is genuinely proven rather than asserted.

The P1 below is about which half of epoch 27 the new test covers. Epoch 27 has two Host flavours because #3277 removed the surface plumbing without bumping the epoch, and the constant your fixture pins — `a6f33c9522ee2d4366046b84e5ed442aa1aaafe2` — is #3277's own merge commit. So the covered case is the tolerant Host, which was never the one at risk; the intolerant pre-#3277 Host is the one the shim's comment was written about, and against it the epoch bump does not produce structured incompatibility. I traced that end to end rather than inferring it: the decoder at `a6f33c952^` requires `surface`, the decode runs before the epoch comparison, and the failure path is a bare `transport.abort()`.

That may still be the right call — if no epoch-27 build predating #3277 was ever shipped or can still be running, the concern is empty and the epoch comment should say so explicitly, since that fact is what licenses the removal. That is a release-history question I cannot answer from the repo, which is why the finding asks you to establish it rather than asserting the change is wrong.

Everything else checks out. The Client's `ClientHello` no longer carries `surface`, the current protocol and server have no reference to it (#3277 removed both sides), and the raw-wire test asserts absence on the actual frame rather than on the typed object, which is the assertion that would catch a re-introduction through a spread. Pinning the CLI restart-guidance test to a literal `27` instead of `EPOCH - 1` is a real improvement: the old form silently followed every future bump and would have stopped testing the boundary it was named for.

One coordination note in a second thread — the epoch number itself is contested by five open PRs and will likely need renumbering, but that is not a defect in this diff.

Carrying a P1 → COMMENT rather than approval. I expect to approve once the pre-#3277 Host question is settled either way.

*AI disclosure: reviewed with Claude Code. I verified the load-bearing claims myself at this head — I diffed `decodeClientFrame` at `a6f33c952^` against `main`, traced the decode-before-admit ordering and the abort path in `host-kernel.ts`, confirmed no surviving `surface` reference in the protocol or server, and re-read every open PR's epoch constant at its current head. The conclusions are mine and rest on those checks.*

packages/runtime-host/src/protocol/index.ts
Sun-GLiang: Verified against the release history. The epoch-27 bump (9aea2c26d) entered main on 2026-08-20 at 03:25 +08:00 and #3277 (a6f33c952) followed at 18:36, while both v0.1.11 and cli-v0.1.0-beta.1 were released on 2026-08-18; no tag contains 9aea2c26d. Therefore no tagged product release shipped the surface-requiring epoch-27 decoder.

Updated in ecc20880e:
- the epoch-28 history now states that removal relies on this unshipped window, not on the epoch mismatch alone;
- a fixture pinned to 0d7d174be (#3277 parent) mirrors decode-before-admit and asserts that the missing surface aborts the transport and reaches Client handshake_failed;
- the existing a6f33c952 fixture continues to prove that the post-#3277 epoch-27 Host returns structured incompatible.

This makes the behavior and support boundary for both epoch-27 variants explicit.

packages/runtime-host/src/protocol/index.ts
Sun-GLiang: Confirmed after fetching upstream/main at 52c0e3275: main remains at epoch 27, so no competing epoch-28 change has landed yet. I am keeping this PR at 28 instead of reserving 29 or 30 speculatively. If another epoch-changing PR lands first, this PR must be rebased and assigned the next epoch before merge. #3313 remains the durable coordination fix for this class of collision.

packages/runtime-host/src/client/connection.ts
```diff
@@ -1405,10 +1394,9 @@ async function exchangeRuntimeHostHandshake(
   | { kind: 'draining' }
 > {
   const helloProtocol = input.helloProtocol ?? input.protocol;
-  const hello: LegacySurfaceClientHello = {
+  const hello: ClientHello = {
```
hqhq1025: [P1] Keep the private `surface` sentinel for currently released Hosts. The release-history argument here only rules out a strict **epoch-27** release, but the latest stable release, `v0.1.11` (published August 18, 2026), is epoch 25 and its `decodeClientFrame` still calls `requireSurface(frame.surface)`. With this hello, a normally upgrading Client reaches that released Host without `surface`; the Host aborts in `#serveConnection` before `#admitHandshake` can compare epochs and return structured incompatibility. The current Client then classifies the connection as `handshake_failed`, and local election eventually reports `host_unresponsive`, bypassing the restart/wait guidance intended to protect resident work. This is a supported latest-release upgrade path, not the unshipped epoch-27 gap documented by the new fixture. Please retain `surface: 'desktop'` until the oldest supported tagged Host has a tolerant decoder (or provide an equivalent pre-decode compatibility path), and pin the regression to `v0.1.11` rather than only the unreleased epoch-27 revision.

hqhq1025 (changes requested):

packages/runtime-host/src/client/connection.ts
Sun-GLiang: Confirmed against tagged v0.1.11 (commit a3c4d0b2a6ca0c87bebebff135d40017558ae5b8): its epoch-25 decoder calls requireSurface before Host admission, so the removal was unsafe. Commit 664422579 restores the private `surface: "desktop"` sentinel, reverts the unsupported epoch-30 removal claim, and adds a frozen v0.1.11 decoder/admission regression that proves the raw hello reaches structured `incompatible`. The CLI guidance test is now pinned to epoch 25. Runtime Host (1037/1037), CLI (336/336), typecheck, lint, format, and diff-check pass. The PR no longer closes #3297; that issue remains open until its release-support gate is satisfied.

hqhq1025 (approved):
Verified the prior released-v0.1.11 compatibility defect is fixed on exact head `66442257`: the private bootstrap sentinel is retained, and the pinned release fixture reaches structured incompatibility before any domain command. No material findings remain.

Sun-GLiang:
CI has not started for the current fork head (`66442257908355e3efe8f2ef6be675a16218087c`). Could a maintainer please approve and run CI for this PR? The Runtime Host and CLI suites, typecheck, lint, format check, and diff check pass locally.

Astro-Han (approved):
Thanks — this is the right shape for a compatibility test, and the reason is worth naming because it is easy to get backwards.

The fixture reimplements the v0.1.11 hello decoder, negotiation and admission independently instead of reusing the current implementation. That is what makes the test meaningful: a compatibility test written against today's code proves the code agrees with itself, and would keep passing through exactly the kind of change it exists to catch. Pinning the fixture to a released revision (`a3c4d0b2`) rather than to a moving target is the same instinct applied to the other axis.

`assert.ok(RUNTIME_HOST_COMPATIBILITY_EPOCH > V0_1_11_HOST_COMPATIBILITY_EPOCH)` is the detail that makes the whole thing safe to leave in the tree. Without it, the day the current epoch stops being ahead of 25 the test would quietly become a test of nothing rather than failing. Guarding the premise of a test, and not only its conclusion, is not something I see often.

On the CLI test switching from `RUNTIME_HOST_COMPATIBILITY_EPOCH - 1` to the literal 25: this narrows it from "whatever the previous epoch happens to be" to "the specific released Host users actually have installed". I think that is the better question to be asking — the relative form drifts silently as epochs advance, and nobody runs "current minus one", they run v0.1.11. The added premise assertion keeps it a genuine incompatible case either way.

No product code changes; no P0–P3 findings. CI is terminal green on this exact head.

Issue #1584, opened 2026-07-29
fix(providers): stop static snapshots from constraining model discovery


Astro-Han:
## Problem

PR #1572 made provider model discovery remote-first, but checked-in fallback snapshots still constrain model availability in two ways.

### Live discovery is bounded by the snapshot

Seven providers use protocol discovery with `filter: 'fallback-models'`:

- `xai-oauth`
- `zenmux`
- `nvidia`
- `deepinfra`
- `groq`
- `openrouter`
- `alibaba`

For these providers, `filterDiscoveredModels` intersects the live response with `fallbackModels`. Refresh can observe that a known model disappeared or is unavailable to the current account, but it can never surface a new model ID. A successful provider response is therefore less authoritative than the bundled recovery snapshot.

### Static-only providers have no escape hatch

A provider with `modelDiscovery.kind: 'fallback'` correctly hides remote refresh when its saved inference credential has no callable model-list endpoint. However, the settings UI also offers no supported way to add an exact model ID manually. Users must wait for a Maka release whenever the provider adds a model that is not yet in the bundled snapshot.

Volcengine Ark direct API and Volcengine Ark Agent Plan currently fall into this category. Agent Plan was checked with a valid dedicated key: both `/api/plan/v3/models` and `/api/plan/v1/models` pass gateway authentication and then return 404. A permanently failing refresh button would not restore the missing capability.

The combined result is a capability regression: users can have valid access to a new model while Maka prevents them from selecting it.

Refs #1571 and #1572.

## Desired outcome

Make fallback inventory, live discovery, capability filtering, and explicit user choices independent concerns.

- Treat `fallbackModels` as bootstrap, offline recovery, recommendation, and display metadata—not as an allowlist over a successful live catalog.
- Audit every `filter: 'fallback-models'` provider and replace snapshot membership checks with provider-backed language/tool compatibility filtering where the response exposes enough information.
- Preserve a returned model with explicit unknown capabilities when the provider cannot describe them, rather than dropping it solely because its ID is new to Maka.
- Continue excluding models that are explicitly known to be incompatible with Maka's chat runtime.
- Give verified static-only providers a manual exact-model-ID path. Persist those entries as explicit user choices, not as provider-fetched facts.
- Preserve last-known-good catalogs on discovery failure and reject empty or malformed responses.
- Keep model IDs exact through discovery, persistence, selection, and inference.
- Add contracts proving that:
  - a successful live response can introduce a compatible ID absent from the bundled snapshot;
[...]

PR #3326, opened 2026-08-20
fix(providers): add GLM-5.3 to Agent Plan catalog

logicrw:
## Summary

Adds `glm-5.3` to Maka's built-in Volcengine Ark Agent Plan fallback catalog, including its published 1M-token context window and 128K maximum output.

Agent Plan's dedicated inference API key has no usable runtime model-list endpoint, so this intentionally preserves fallback-based discovery rather than adding a `/models` refresh request.

Refs #1584

Sources:

- [Volcengine Ark model list](https://console.volcengine.com/ark/region:cn-beijing/docs/82379/1330310?lang=zh)
- Official Agent Plan console: `glm-5.3` is present in the `ark-code-latest` routing catalog.

## Verification

- `npm --workspace @maka/core test` — passed (564 tests) before the targeted assertion was removed at maintainer feedback.
- `npx biome lint packages/core/src/__tests__/model-catalog.test.ts` — passed after the test removal.
- `git diff --check` — passed before committing the removal.
- Not run: full repository test suite, format check, or typecheck.

## AI use

Tool(s) and scope: Maka prepared the catalog and metadata change, and this PR description. The human contributor reviewed the scope and authorized submission. The commits include `Generated-by: Maka` trailers.

## Checklist

- [ ] No

### DIFF
--- a/packages/core/src/__tests__/model-catalog.test.ts
+++ b/packages/core/src/__tests__/model-catalog.test.ts
@@ -132,27 +132,6 @@ test('connection catalogs preserve user-choice provenance without inventing avai
   assert.deepEqual(entries[2]?.provenance.sources?.userChoice, ['session_model']);
 });
 
-test('Volcengine Agent Plan offers GLM-5.3 with its published model limits', () => {
-  const entries = buildConnectionModelCatalogEntries({
-    connection: {
-      slug: 'volcengine-agent-plan',
-      providerType: 'volcengine-agent-plan',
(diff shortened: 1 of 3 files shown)


Astro-Han (dismissed):
Nice Clean Fix! My only advice would be to remove the test part as the model's information is a fact in the model-metadata.ts, add a test would not help prevent a real regression as the it use the fallback modelsource to connect.

Astro-Han (approved):

PR #3265, opened 2026-08-20
feat(win): Abort-path installer rollback with backup retention

liugddx:
## Summary

Adds **Abort-path rollback with backup retention** to the one-click NSIS installer, a Windows release gate that exercises the boundary scenario by scenario, and EN/zh documentation for the supported recovery path.

The rollback patch was rebased onto `main` `04680fe0b`, then maintainer-rebased without content changes onto `ce0e5931a`; `git range-diff` reports both commits as exact `=` matches. Fresh GitHub checks exercise the current head against the live base. The diff remains rollback-scoped: the NSIS include and verifier, electron-builder/script/docs wiring, the pull-request and unified release workflow call sites, and the direct workflow/ASF-header policy tests required by current `main`.

Part of #2142.

## Mechanism

Before the destructive upgrade boundary, `customInit` snapshots the required uninstall registration, creates a same-volume sibling backup, verifies its executable witness, writes a version-bound completeness marker and `RECOVERY-README.txt`, and persists the snapshot outside the keys deleted by the old uninstaller.

Admission fails closed with exit 101 when:

- the required registration or `DisplayVersion` is missing and no adoptable persisted snapshot exists
- backup creation or its executable witness fails
- a leftover backup marker does not exactly match the current/persisted snapshot version
- a marker matches but the sibling backup lacks `Maka.exe`

A complete leftover backup is adopted only when both its executable witness and exact version marker match the current or persisted snapshot.

On the NSIS Abort path, `.onInstFailed` performs a two-step same-volume swap: move the extracted new tree aside, move the verified backup into place, write the registry snapshot, then read back and exactly verify `DisplayName`, `UninstallString`, `NoModify=1`, and `NoRepair=1`. Exit 102 is reported only after the restored executable and required registry values are verified; cleanup happens after that proof.

A registry read-back mismatch returns 103 before cleanup. The restored tree, aside tree, and persisted snapshot remain. Rebuilding the sibling backup is best effort: when it succeeds, the version marker and README permit a later recovery rerun; otherwise a later installer fails closed and the interactive message directs manual repair/reinstall instead of promising automatic adoption.

`customInstall` removes backup/aside/snapshot residue only after a successful install.

## Windows gate scenarios

1. Abort after extraction, before registry creation: exit 102; previous tree byte-identical by per-entry SHA-256 (including empty directories), registered, and launchable.
[...]

Issue #3279, opened 2026-08-20
runtime-host: intermittent 'stopped responding during startup' after a Windows upgrade over an existing profile


Joob1n:
On the Windows upgrade-lifecycle check, the app installed by an upgrade intermittently fails to start: its own stderr reports

```
[runtime-host] fatal: Error: Runtime Host stopped responding during startup
    at runtimeHostStartupError (…/app.asar/node_modules/@maka/runtime-host/dist/client/startup-error.js:25:20)
    at RuntimeHostDesktopManagerImpl.connect (…/dist/main/runtime-host-desktop-manager.js:317:19)
    at async RuntimeHostReconnectLifecycleImpl.start (…/reconnect-lifecycle.js:62:45)
```

and the renderer never mounts, so the packaged smoke times out.

**Observed:** [run 32324991998](https://github.com/apache/maka/actions/runs/32324991998/job/96295423079) on #3241, step `Exercise pinned-version upgrade and uninstall`. The sequence that failed: the pinned 0.1.9 baseline installed and verified clean → upgraded to the current 0.1.11 build → the upgraded app's Runtime Host stopped responding during startup. The standalone smoke of the same 0.1.11 build passed minutes earlier in the same job, so the build starts fine against a fresh profile.

**What makes the upgrade path different:** the lifecycle deliberately reuses one isolated HOME and user-data directory across both versions — that is the point of an upgrade test — so the 0.1.11 Runtime Host starts against whatever state 0.1.9's run left under that HOME (host socket/lock/config). Whether the hang is caused by that leftover state or is an unlucky cold-runner slowdown is exactly what the log cannot yet say: the fatal is a client-side startup timeout, and the Host process's own stderr is not captured in the verifier output.

**Attribution note:** this surfaced through #3241's new CDP diagnostics (the app wrote `DevToolsActivePort`, the poll named the bound port, and the endpoint never answered — which is what pointed at the main process rather than the port plumbing). Some of the historical `did not expose CDP within 30 seconds` failures in this step may share this cause, but that cannot be established retroactively from the old logs.

**Suggested next steps:**
1. Capture the Runtime Host child's stderr (or its log file) into the lifecycle verifier's failure output, so a startup hang is attributable to the Host's own last words rather than the client's timeout.
2. Reproduce locally: install 0.1.9's profile state, then start a current build against it, in a loop.
3. If leftover state is the trigger, the fix likely belongs in Host startup's handling of a predecessor's socket/lock remnants.

[...]

PR #3327, opened 2026-08-20
test(windows): harden the release-verification harness and pin its contracts

liugddx:
## Summary

Windows release-verification harness hardening, split out of #3265 at its reviewer's request ("land the CI-stability fixes first, as their own PR") so these fixes merge on their own evidence and #3265 stands or falls on the installer transaction alone. All changes are under `scripts/` plus one workflow test-step line; no packaged-product code is touched.

## What's fixed

**CDP attach and renderer readiness** (four observed failures in this family across #3161/#3240/#3265 lanes):
- `waitForDevToolsPort` reads the port from the DevTools stderr announcement (`--remote-debugging-port=0`) instead of pre-reserving a port that Electron may race.
- Renderer discovery: 90s deadline, per-probe `AbortSignal` bounds (a hung connect once turned a "90 seconds" failure into ~6 minutes), errno cause chains in the failure message.
- New shared `waitForUsableRenderer`: polls the renderer-usable state with the deadline as the sole authority — a single stalled `Runtime.evaluate` used to fail the whole gate (run 32352924376, `CDP renderer evaluation timed out.` while the renderer was still booting). The WebSocket handshake now has its own bound, so a port that accepts TCP but never speaks fails one probe instead of hanging the lane (#3265 review finding 3821101776).

**Uninstall registration reads** (the cause of the deterministic red in run 32363159920):
- The scan matched `DisplayName -eq 'Maka'`, but electron-builder's default `uninstallDisplayName` is `"${productName} ${version}"` (NsisTarget.js:473), so the registered name is `Maka 0.1.11` and the filter matched nothing — every reader was blind. Now matches the versioned name.
- `waitForUninstallRegistrationToClear`: an NSIS uninstaller launched without `_?=` detaches and deletes its registry keys tens of seconds after the files disappear; anything installing into that window gets its fresh registration deleted. The wait's comment states the residual window precisely (one registry call wide — the uninstall key at uninstaller.nsh:250, the install key at :254; finding 3820922176).

**Bounded probes under polling deadlines** (findings 3820232602 / 3820922173 / 3821101780 — the anti-pattern #3241 names; kept as explicit per-site bounds rather than a `runCommand` default, per the accepted argument that codesign/notarization have no honest upper bound):
- Every PowerShell probe that runs inside a wait loop now passes `timeoutMs`, including the evidence capture that runs on an already-wedged machine.
- Waits tolerate a failed probe and retry until their deadline; a failed enumeration is never treated as "no processes".

**Evidence and coverage** (finding 3819778535):
- `directoryTreeManifest` / `diffTreeManifests` (shared exports the #3265 gate consumes) now record empty directories so their loss is visible.
- The table-driven contract tests exist in the repository as `scripts/verify-windows-harness.test.mjs` (manifest shape/sorting/empty-dir/symlink-fail-closed, diff table, runCommand timeout kill, DevTools port announcement, usable-renderer deadline semantics, both waits' tolerate-and-deadline contracts) and run in the CI planner test step on every PR — 19 pass locally on Windows (1 skip: symlink creation needs privilege), and the step runs them on Linux too.
- On a relaunch version mismatch, the autoupdate harness captures process command lines, backup-directory presence, uninstall registrations and executable timestamps before failing, so the next occurrence of the nondeterministic "relaunched but still old" state is attributable.

## Verification

- L1/L2 (this machine): `node --check` on all touched files; biome clean; `verify-windows-harness.test.mjs` 19 pass / 1 skip; `ci-test-plan.test.mjs` 17 pass (step-name pin updated to the renamed step, contract unchanged: planner tests still run before dependency setup, unconditionally).
- L3 (CI): the Windows lanes execute these helpers for real; this PR's own `test` lane runs both test files.

Relates to #3265 (review findings 3819778548, 3820232602, 3820922173, 3820922176, 3821101776, 3821101780, 3819778535) and #2142.

### DIFF
--- a/scripts/verify-windows-installer-lifecycle.mjs
+++ b/scripts/verify-windows-installer-lifecycle.mjs
@@ -8,6 +8,7 @@ import { powerShellLiteral, verifyPackagedWindowsApp } from './verify-windows-x6
 const uninstallExecutableName = 'Uninstall Maka.exe';
 const temporaryCleanupRetries = 20;
 const temporaryCleanupRetryDelayMs = 250;
+const pollingProbeTimeoutMs = 10_000;
 
 export function installerVersion(path) {
   const match = basename(path).match(/^Maka-(\d+\.\d+\.\d+)-win-x64\.exe$/u);
@@ -21,7 +22,10 @@ function delay(milliseconds) {
   return new Promise((resolvePromise) => setTimeout(resolvePromise, milliseconds));
 }
 
-export async function listInstalledProcesses(installDirectory, { run = runCommand } = {}) {
+export async function listInstalledProcesses(
+  installDirectory,
+  { run = runCommand, timeoutMs = pollingProbeTimeoutMs } = {},
+) {
   const root = `${resolve(installDirectory)}${sep}`;
   const script = String.raw`
 $root = [IO.Path]::GetFullPath(${powerShellLiteral(root)})
@@ -44,7 +48,7 @@ $matches | ConvertTo-Json -Compress
     'powershell',
     ['-NoProfile', '-NonInteractive', '-Command', script],
     {
-      timeoutMs: 30_000,
+      timeoutMs,
     },
   );
   if (!stdout.trim()) return [];
@@ -69,7 +73,9 @@ export async function waitForInstalledProcessesToExit(
     // polling and fail at the deadline, not pass or die on one hiccup.
     let processes;
     try {
-      processes = await listProcesses(installDirectory);
+      processes = await listProcesses(installDirectory, {
+        timeoutMs: remainingProbeBudget(deadline),
+      });
       lastProbeError = undefined;
     } catch (error) {
       lastProbeError = error;
@@ -115,7 +121,11 @@ export async function waitForInstalledProcessAppearance(
   let lastProbeError;
   for (;;) {
     try {
-      const matched = (await listProcesses(installDirectory)).filter(
+      const matched = (
+        await listProcesses(installDirectory, {
+          timeoutMs: remainingProbeBudget(deadline),
+        })
+      ).filter(
         (processInfo) => basename(processInfo.path).toLowerCase() === executableName.toLowerCase(),
       );
       lastProbeError = undefined;
@@ -150,9 +160,31 @@ export async function terminateInstalledProcesses(
     run = runCommand,
     listProcesses = listInstalledProcesses,
     waitForExit = waitForInstalledProcessesToExit,
+    enumerationTimeoutMs = 60_000,
+    pollIntervalMs = 1_000,
+    sleep = delay,
   } = {},
 ) {
-  const processes = await listProcesses(installDirectory);
+  const deadline = Date.now() + enumerationTimeoutMs;
+  const enumerate = (directory, options = {}) => listProcesses(directory, { run, ...options });
+  let processes;
+  for (;;) {
+    try {
+      processes = await enumerate(installDirectory, {
+        timeoutMs: remainingProbeBudget(deadline),
+      });
+      break;
+    } catch (error) {
+      if (Date.now() >= deadline) {
+        throw new Error(
+          `Could not enumerate installed Maka processes within ${enumerationTimeoutMs}ms: ` +
+            `${error.message}`,
+          { cause: error },
+        );
+      }
+      await sleep(pollIntervalMs);
+    }
+  }
   for (const processInfo of processes) {
     try {
       await run('taskkill', ['/PID', String(processInfo.processId), '/T', '/F'], {
@@ -165,7 +197,7 @@ export async function terminateInstalledProcesses(
       }
     }
   }
-  await waitForExit(installDirectory, { listProcesses });
+  await waitForExit(installDirectory, { listProcesses: enumerate });
 }
 
 export async function waitUntilMissing(
@@ -192,7 +224,10 @@ export async function waitUntilMissing(
  * with commas ('' when none). One PowerShell scan, shared by the harnesses
  * that assert on — or wait out — the uninstall registration.
  */
-export async function readUninstallDisplayVersions({ run = runCommand } = {}) {
+export async function readUninstallDisplayVersions({
+  run = runCommand,
+  timeoutMs = pollingProbeTimeoutMs,
+} = {}) {
   // electron-builder's default uninstallDisplayName is
   // "${productName} ${version}" (NsisTarget: UNINSTALL_DISPLAY_NAME), so the
   // registered DisplayName is "Maka 0.1.11", never the bare product name. The
@@ -212,7 +247,7 @@ $entries = Get-ChildItem 'HKCU:\Software\Microsoft\Windows\CurrentVersion\Uninst
     'powershell',
     ['-NoProfile', '-NonInteractive', '-Command', script],
     {
-      timeoutMs: 30_000,
+      timeoutMs,
     },
   );
   return stdout.trim();
@@ -248,7 +283,10 @@ export async function waitForUninstallRegistrationToClear({
     // end a wait that owns two minutes of budget.
     let versions;
     try {
-      versions = await readUninstallDisplayVersions({ run });
+      versions = await readUninstallDisplayVersions({
+        run,
+        timeoutMs: remainingProbeBudget(deadline),
+      });
       lastProbeError = undefined;
     } catch (error) {
       lastProbeError = error;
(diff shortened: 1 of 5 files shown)


liugddx:
All lanes are green at head d6f443f48: the unified CI job (harness contract tests included, running pre-install as pinned by #3317's step contract) and the full `release-windows-check` Windows lane (18m14s, through the end-to-end autoupdate gate — run 32380790549). This is the split requested in #3265's review so the harness fixes merge on their own evidence; #3265 is stacked on it and rebases cleanly. Ready for review/merge whenever convenient.

liugddx:
@Astro-Han @M4n5ter PTAL

hqhq1025 (changes requested):
[P1] `scripts/verify-windows-installer-lifecycle.mjs:47` — Retry WMI failures inside the relaunch deadline. This new 30-second bound correctly turns a wedged `Get-CimInstance` into a rejection, but `scripts/verify-windows-autoupdate.mjs:410` directly awaits `listInstalledProcesses()` inside a 120-second relaunch loop. Run 32340493254 failed on exactly this path after one transient WMI timeout even though the upgraded process may already have relaunched. Catch failed rounds, continue until the outer deadline, and report the last probe error only if no later enumeration succeeds; use a shorter or remaining-budget probe timeout and add a failure-then-success regression.

[P2] `scripts/verify-windows-autoupdate.mjs:492` — Reuse this termination policy in cleanup. The main path now treats exit 128 and timed-out `taskkill` as mechanism failures before relying on `waitForInstalledProcessesToExit`, but the `finally` block at lines 607-613 still calls `taskkill` directly. Run 32378497920 recorded the same stale-PID exit 128 from cleanup. After a later smoke or verification failure, that exception skips the authoritative exit wait and the uninstaller, potentially leaving a stale uninstall registration on the runner. Extract one stop-and-verify helper for both paths and complete the uninstall/registration barrier during cleanup.

Astro-Han:
A data point for the bounded-probe half of this PR, from #3225's lane.

`Verify automatic update end to end` went red there four times: once `did not expose CDP within 30 seconds: fetch failed`, twice `taskkill /PID … /T /F failed with exit code 128`, and on the last rerun it just stopped — 47 minutes on that single step, no output, heading for the 75-minute job timeout. I cancelled it.

That last one is the shape you describe in the PR body: with no bound the step has no upper limit at all, so the cost is a wedged runner rather than a fast red. Good to see `timeoutMs` pushed down to every probe inside a wait loop.

Reference: https://github.com/apache/maka/actions/runs/32387807797 (job 96534477439). Nothing needed from you — just adding evidence in case it's useful for the review.

Joob1n:
@liugddx — this subsumes #3241, so I am closing that one rather than making anyone review the same file twice. Two things from its review and its CI history that are worth carrying over, since neither is in this PR yet.

**The probe timeout is too large for the budget it sits in.** `listInstalledProcesses` here takes `timeoutMs: 30_000`, and `waitForInstalledProcessesToExit`'s default budget is 60 seconds — so a stalled probe leaves room for exactly one more attempt. @likun666661 made this point on #3241 (where I had the same 30 seconds) and it is the sharpest review either of these PRs has had: a loop that can only try twice cannot survive a stall, which is the entire reason to retry. #3241 ended at **10 seconds**, giving about six attempts inside a 60-second wait and twelve inside the 120-second relaunch wait, still far above a healthy WMI query. The retry logic you have written is right; the constant is what decides whether it can do anything.

**Measured at the relaunch wait, same step, across two revisions of my branch:**

| probe bound | loop on probe failure | result |
| --- | --- | --- |
| none | — | hung 59 minutes past a 120s deadline; runner cancelled the job |
| 30s | aborts the wait | failed in 74 seconds naming the stalled query |
| 10s | retries until the deadline | **relaunch detected in 27 seconds** |

Being precise about what that proves: the last row changed both the bound and the loop behaviour at once, so it does not isolate the constant — the argument for 10s over 30s is the arithmetic above, not this table. What the table does establish is that the stall is real, reproducible at that exact step, and that a tolerant loop gets through it.

**Two faults still ahead of you on that lane, neither mine to fix:**

- With the relaunch wait fixed, my last run died at the next step: `taskkill /PID … did not finish within 30000ms` while stopping the relaunched instance, plus a second `taskkill … exit code 255` during cleanup. Bounded, so it fails cleanly — but the same process-table churn, one step later.
- #3279 tracks a product-side intermittent: the upgraded install's own Runtime Host stops responding during startup, which this lane shows as a renderer that never appears. Worth recognising so it does not get filed as a harness bug.

The uninstall `DisplayName` finding is a good catch — `Maka 0.1.11` versus a bare `Maka` would make every reader silently blind, and that class of bug survives exactly because the check still "passes".

liugddx:
@hqhq1025 @Joob1n Addressed at head `22ad8be62` within the verifier-only scope.

- WMI/registry polling rounds are capped at 10 seconds and additionally capped to the enclosing wait's remaining budget.
- The upgraded-process appearance wait now retries transient enumeration failures until its 120-second deadline and reports the last probe failure only if no later round succeeds.
- Main-path and `finally` cleanup now share one `terminateInstalledProcesses` stop-and-prove policy. Exit 128 and an overrun `taskkill` remain mechanism failures only; the authoritative process-exit wait still runs.
- Both normal and cleanup uninstall paths share the same completion barrier: installation directory absent, then uninstall registration absent.
- Renderer CDP evaluation now receives the outer wait's remaining budget, so a half-open WebSocket cannot overrun the declared deadline.

Regression coverage includes failure-then-success WMI enumeration, transient cleanup enumeration failure, tolerated kill failures followed by the exit proof, both uninstall barriers (including an already-missing detached uninstaller), a half-open WebSocket, and proof that a timed-out command's child no longer executes.

Local verification: 29 passed, 1 symlink-privilege skip, 0 failed; Node syntax checks, Biome 2.5.6 formatting, and `git diff --check` all pass.

scripts/verify-windows-installer-lifecycle.mjs
```diff
@@ -53,19 +66,140 @@ export async function waitForInstalledProcessesToExit(
...
+    }
+  }
+  for (const processInfo of processes) {
+    try {
+      await run('taskkill', ['/PID', String(processInfo.processId), '/T', '/F'], {
+        timeoutMs: 30_000,
+      });
+    } catch (error) {
+      const message = String(error?.message);
+      if (!/exit code 128/.test(message) && !/did not finish within/.test(message)) {
```
hqhq1025: [P2] Always run the authoritative exit proof before classifying `taskkill` failures. This branch states that the kill is only the mechanism, but any error other than exit 128 or a timeout is rethrown here before `waitForExit`; the live PR discussion records exit 255 on this same cleanup lane. If the process exited between enumeration and kill, cleanup still sets `exited = false` and skips the uninstaller plus both directory and registration barriers, recreating the stale-runner state this fix is meant to prevent. Record kill errors, run `waitForExit` unconditionally, and let that proof decide: ignore mechanism errors when no process remains, or report them together with a failed exit proof. The test that currently expects exit 1 to bypass the proof should be updated accordingly.

hqhq1025 (commented):
The previous WMI appearance-loop P1 is fixed, and the cleanup paths now share the intended stop-and-prove helper. Exact-head harness and CI-planner tests passed 51/51; syntax, lint, format, diff, and merge-tree checks also passed. One cleanup-path P2 remains inline.

liugddx:
Follow-up from the first Windows run at `22ad8be62`: the relaunch and cleanup fixes worked far enough to expose the next transient probe, a one-shot PowerShell `ProductVersion` read that timed out after 30 seconds.

Head `913fbb99d` keeps this narrowly in the verifier: the installed-version read now owns a 60-second outer budget, caps each PowerShell round at 10 seconds (or the remaining budget), and retries failures before reporting the last error at the deadline. Added a failure-then-success regression.

Local verification: 30 passed, 1 symlink-privilege skip, 0 failed; Biome 2.5.6, syntax checks, and `git diff --check` pass. The new CI run is the authoritative Windows evidence.

scripts/verify-windows-installer-lifecycle.mjs
liugddx: Addressed in `d9fe44361`. `terminateInstalledProcesses` now records every `taskkill` error and always runs the authoritative exit proof. If the proof finds no installed process, mechanism errors including exit 255 do not block uninstall/cleanup. If residue remains, an `AggregateError` reports the exit-proof failure together with all recorded kill diagnostics.

The prior test that expected exit 1 to bypass the proof was replaced with two contract cases: arbitrary kill failure + successful proof, and kill failure + failed proof with both errors retained. Local harness result: 31 passed, 1 symlink-privilege skip, 0 failed.

liugddx:
@hqhq1025 PTAL

hqhq1025 (approved):
Approved at exact head `d9fe443614bed143d84da4692f9eba3cc2dc9458`.

No actionable findings remain. The two issues from my previous review are fixed at the shared ownership boundary:

- Process-appearance polling now tolerates transient WMI failures until the enclosing deadline.
- Main-path and cleanup termination both use one stop-and-prove helper. Every `taskkill` error is retained as mechanism evidence, but the authoritative installed-process exit check always runs; kill errors are ignored only when that proof confirms no process remains, and are aggregated with the exit failure when residue remains.

The later ProductVersion fix follows the same rule: each PowerShell read is capped to ten seconds or the remaining outer budget, transient probe failures retry for up to sixty seconds, and the version assertion remains strict. This improves observability and boundedness without converting a real old-version relaunch into success.

Problem definition and mechanism: the PR addresses verifier instability caused by unbounded or one-shot observations, not product failures. It removes the CDP port reservation race, bounds HTTP/WebSocket/CDP and PowerShell probes, retries transient observation failures under one outer deadline, waits for both uninstall files and registration state, and retains hard assertions on renderer usability, installed process exit, product version, directory contents, and uninstall cleanup.

First principles and Occam's razor: yes. Unreliable probes are mechanisms; durable state and final process/file/registry assertions are the verdict. The shared renderer wait, process lifecycle helpers, and uninstall barrier remove duplicated policies rather than adding fallback paths. I found no production code to delete because the PR touches verifier code only. The table-driven tests cover distinct contracts and failure sequences; I found no low-quality test that should be removed. No deeper refactor is required.

Verification:

- Harness and CI planner contracts passed 53/53 locally.
- Node syntax checks, Biome, and `git diff --check` passed.
- The merge tree against current `main` is clean; the current-main semantic delta remains the same five verifier/workflow files.
- Live CI is green, including the full Windows release, installer, and end-to-end automatic-update job.

Merge verdict: ready to merge.

Residual risk: this hardens the verifier but does not claim to fix product-side Windows startup or update defects such as #3279. Those states should now fail within bounded time with enough evidence to attribute them, rather than hanging or being misclassified.

PR #3183, opened 2026-08-18
feat(runtime): retire the Claude subscription OAuth provider

Joob1n:
## Summary

Maka could sign in with a Claude Pro/Max subscription and send inference through it. Anthropic's Consumer Terms permit programmatic access to the consumer Claude services only through an API key or explicit permission, and this path had neither — it presented itself as Claude Code (borrowing that `client_id`, its User-Agent, its beta header set, and an `x-app: cli` marker) to get requests accepted. The account carrying that risk is the user's, not Maka's.

This removes the capability rather than gating it further.

`claude-subscription` keeps its registry entry, and the entry is marked `retired` — deliberately distinct from a provider that was never wired, since both end up with no Runtime adapter but only one of them used to work and needs to explain itself.

Keeping it registered is a product choice, not a technical necessity. Unregistering is possible: `isRetiredGeminiCliConnection` is the precedent, filtering by raw `providerType` string before canonical decode. It is avoided here because a user who signed in has a credential on this machine, and the connection is what makes that visible and deletable — unregister it and onboarding reports `needs_connection`, with the credential still on disk and nothing pointing at it. (`decodeProviderType` does throw on unknown types and `connection-catalog-codec.ts` would fail the whole document, so unregistering *without* the filtering precedent would cost the user every other connection — but that is a reason to use the precedent, not a reason the choice was forced.)

Retirement is refused at each authority that could otherwise admit the connection, so no single revert makes it sendable again:

| Authority | Behavior |
| --- | --- |
| `deriveProviderAuthContract` | every action `hidden` — this is what makes the storage layer refuse a model fetch or a connection test |
| `resolveExecutionConnection` | returns `provider_retired`, a first-class result, so a retained credential can no longer read as `ready` |
| `SessionCatalogCoordinator` | refuses before a Session is committed — the path Bot, CLI and scheduled runs take, which never reaches the Desktop readiness gate |
| `isConnectionReady` | `provider_retired`, before a Desktop send is admitted |
| `buildConnectionModelCatalogEntries` | every model resolves to `provider_removed`, so the pickers stop offering them |
| `resolveModelRuntime` | refuses ahead of the per-model override lookup, so a generated override row cannot hand a retired provider a working adapter |
| storage interactive-login allow list | no longer names it |
| `OAUTH_LOGIN_PROVIDERS` (Host wire) | no longer names it |
| `getAIModel` | throws, as the last backstop |

[...]

PR #3330, opened 2026-08-20
fix: stop treating a model list as an allowlist

### LINKED ISSUE #1584: fix(providers): stop static snapshots from constraining model discovery

Astro-Han:
## Summary

A Volcengine Ark Agent Plan key cannot select the models its own plan serves (#1584). The plan's key reaches only the inference data plane, so Maka cannot enumerate the account and falls back to the array this build ships — and every gate read that array as the set of models the account has.

The array is not that set, and no array Maka holds is. A `/models` response describes one call at one moment, scoped to whatever the provider chose to report; a shipped snapshot describes the provider, not the account; four providers have no list endpoint at all. Only the user knows what their account can run, and only the provider can settle it — by answering the request. If the user is wrong, the provider's own error says so, which is strictly better than Maka pre-refusing from a list that may be stale, filtered, or nonexistent.

So this PR removes the allowlist rather than adding another exception to it. `authorizeConnectionModel` is the one seam that answers "can this connection run this model", and the user's selection is the authorization. An observed row still wins wherever it exists — it carries wire metadata and capabilities a synthesized entry cannot — but its absence authorizes nothing and vetoes nothing. A fetch no longer prunes what the user chose; `filterDiscoveredModels` no longer intersects a live response with the shipped snapshot (which was silently dropping models from six providers whose real catalogs outgrew it); readiness, the connection probe, the hosted execution target and the catalog coordinator all consult that one seam, and `connection-model-admission.ts` with its hardcoded DeepSeek exception is deleted.

What the picker *says* is unchanged in spirit and now purely advisory: a model the live response omitted is annotated `not_in_live_list`, a warning the user can act on rather than a refusal.

The second commit is the other half: nothing could put a model into such a connection. An "Add model" action now sits alongside "Update models" instead of replacing it, and the `relayModelProfiles` gate splits by field rather than by provider — `contextWindow`/`vision` are facts about a model and legal everywhere, `thinkingLevels`/`serviceTier` name an OpenAI-compatible wire feature and stay relay-only.

Fixes #1584

image

image

## History

An earlier revision of this branch repaired the read path and kept the allowlist. Adversarial review found a P0 in it; the redo built a read-side authority, and review found two more P1s. Both rounds were the same shape — every gate that consulted a list Maka does not own was another exit — so the branch was rewritten against the conclusion above and the concept was removed. The commits are new; the previous ones described a design that no longer exists.

## Verification

`@maka/core` 584, `@maka/storage` 848, `@maka/runtime` 3002, `@maka/runtime-host` 1042, `@maka/desktop` 1017, `maka-agent` 339, `@maka/mcp` 115, Storybook smoke 144 stories, `format:check`, the Astryx surface inventory and `typecheck` clean across every workspace. Rebased onto current `main` (past the Claude subscription retirement) and re-run there.

The first two commits were typechecked in isolation, so reverting the later ones leaves a working tree. The third commit answers the review's P2: the add-model dialog now closes only after the write lands, holding the typed id and context window when it does not.

The behavioural contracts that changed were rewritten rather than deleted: a discovery run that stops listing a selected model now asserts the selection survives, the connection probe asserts it tests the user's model even when the last inventory came back empty, and the workspace default target asserts it stays put. The end-to-end test drives a real `volcengine-agent-plan` connection through `testConnection` and asserts the wire body names the user's model.

## Known gap

The CLI onboarding wizard still offers only what discovery returned, so declaring a model by hand is reachable from desktop connection settings only. CLI `/setup` no longer deletes models added there, but it cannot add one.

## AI use

Tool(s) and scope: Claude Code (Opus 5) wrote the implementation, tests and this description under my direction, and ran the suites above. I reviewed the final diff and commit messages and own the result.

## Checklist

- [ ] No


packages/core/src/llm-connections.ts
```diff
@@ -174,6 +174,86 @@ export function connectionEnabledModelIds(connection: {
...
+  readonly defaultModel?: string;
+  readonly models?: readonly ModelInfo[];
+  readonly modelSource?: ModelDiscoverySource;
+}
+
+export function classifyConnectionModelInventory(
+  connection: ConnectionModelAuthorityInput,
+): ConnectionModelInventory {
+  if (connection.models === undefined || connection.modelSource === undefined) return 'absent';
+  if (!providerSupportsModelDiscovery(connection.providerType)) return 'snapshot';
```
hqhq1025, line 219: [P1] Preserve provider-returned models that are absent from the fallback snapshot

This classifies every `modelDiscovery.kind !== "fallback"` result with `modelSource: "fetched"` as an account-authoritative allowlist, but ZenMux, NVIDIA, DeepInfra, Groq, OpenRouter, and Alibaba still run successful provider responses through `filter: "fallback-models"`. On a normal OpenRouter `/models` response containing one bundled ID and one new compatible ID, discovery persisted only the bundled ID; this authority then rejected the returned, user-enabled model as `not_in_live_list`. These providers expose refresh instead of manual addition, so the first desired outcome in #1584 remains unreachable. Remove snapshot-membership filtering for successful live discovery, or persist an explicit partial/non-authoritative inventory classification before using absence as a veto.

apps/desktop/src/renderer/settings/use-connection-detail.ts
```diff
@@ -478,6 +479,66 @@ export function useConnectionDetail(props: ConnectionDetailProps) {
...
+    const lifecycle = connectionDetailLifecycleRef.current;
+    const releaseSaveModels = connectionDetailActionGuard.begin('save-enabled-models');
+    if (!releaseSaveModels) return;
+    setSavingEnabledModels(true);
+    setEnabledModelIds(next);
+    let saved = false;
+    try {
+      await props.bridge.update(connection.slug, {
+        enabledModelIds: next,
+        relayModelProfiles: { ...(savedRelayProfiles ?? {}), [modelId]: { contextWindow } },
```
hqhq1025, line 522: [P1] Keep manually added models across snapshot refresh

This write establishes the user-selected ID and its context-window declaration, but `ConnectionCatalogDocumentOwner.writeModelFetchResult` still reconciles every `source: "fetched"` result as an allowlist. A fallback provider such as `volcengine-agent-plan` labels its replayed shipped snapshot `fetched`; refreshing after adding `deepseek-v4-pro-beta` removed that ID from `enabledModelIds` and pruned its persisted profile, so the model became unusable after restart. This violates #1584’s explicit survival-across-refresh contract. Use the provider-aware inventory classification at the storage boundary: only a genuinely live inventory should rewrite selection/default/profile state; a snapshot refresh should update observed metadata while preserving user authority.

hqhq1025 (changes requested):
Two merge-blocking authority inconsistencies remain on the reviewed revision. Both reproduce on the exact head despite green required CI and focused local verification.

Astro-Han:
Both findings landed. The conclusion I drew from them is that fixing either one in place would have left the next exit open, so the branch is rewritten and force-pushed rather than amended — the commits are new, and the old ones described a design that no longer exists.

**P1-1** is fixed at its root. `filterDiscoveredModels` intersected every live response with the shipped snapshot, so ZenMux, NVIDIA, DeepInfra, Groq, OpenRouter and Alibaba lost models their real catalogs had outgrown. `filter: 'fallback-models'` is removed from the registry and the filter now does only what its name says.

**P1-2** is fixed, though not quite where the comment points. `fetch_models` is gated to `hasSecret && supportsModelDiscovery` in `provider-auth.ts`, so a `kind: 'fallback'` connection has no refresh action in settings and that specific path does not run. The same loss does reach the user through onboarding, which rewrites the selection; I reproduced it there. Both are gone now — `reconcileConnectionAfterModelFetch` no longer prunes, and onboarding merges models it never displayed instead of dropping them.

Beyond the two findings: your suggested remedy — persist an explicit partial/non-authoritative classification before using absence as a veto — is what the reviewed revision already was, and it was the second revision to fail in that shape. So I took the question a step back. No list Maka holds is the set of models an account can run. A `/models` response describes one call at one moment, scoped to whatever the provider chose to report; a shipped snapshot describes the provider, not the account; four providers have no list endpoint at all. When a user selects a model the provider did not list, sending the request and letting the provider answer is strictly better than refusing from a list that may be stale, filtered, or absent.

So the allowlist is removed rather than reclassified. `authorizeConnectionModel` is the single seam, the user's selection is the authorization, and an observed row still wins where it exists — for its wire metadata and capabilities — but its absence neither authorizes nor vetoes. Classification survives for display only: a model the live response omitted is annotated, not refused.

Details in the updated description. CI is green on the new head.

Claude Code (Opus 5) did the rewrite and drafted this reply under my direction; I reviewed the diff and own the result.

packages/core/src/llm-connections.ts
Astro-Han: Fixed by the rewrite — see the summary comment.

apps/desktop/src/renderer/settings/use-connection-detail.ts
Astro-Han: Fixed by the rewrite — see the summary comment.

Astro-Han:
Ready for another look. Since the review above:

- The branch was rewritten rather than patched — the allowlist is removed instead of reclassified. Both P1 threads are answered in the [summary comment](https://github.com/apache/maka/pull/3330#issuecomment-5365822247) and the description is rewritten to match.
- Rebased onto current `main`, past the Claude subscription retirement (#3183). Two conflicts, both resolved in favour of `main`'s new behaviour: the `provider_retired` step keeps its place at the head of the readiness order, and the two new retired-provider coordinator cases are kept alongside the one this PR rewrites.
- Green on the new head: core 584, storage 848, runtime 3002, runtime-host 1042, desktop 1017, cli 339, mcp 115, Storybook 144 stories, plus `format:check`, the Astryx surface inventory and `typecheck` across every workspace. Required CI is green.

@hqhq1025 re-requested since the finding was yours; anyone else on the list is welcome. Note that the commits are new — reviewing the previous ones would describe a design that no longer exists.

apps/desktop/src/renderer/settings/provider-add-model-dialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+    props.onOpenChange(false);
+  }
+
+  function submit(event: FormEvent) {
+    event.preventDefault();
+    setSubmitAttempted(true);
+    if (idError || !contextWindow) return;
+    props.onSubmit(trimmedId, contextWindow);
+    close();
+  }
```
hqhq1025, line 71: [P2] Keep the draft until the model write succeeds

onSubmit is declared as synchronous and the parent deliberately discards addDeclaredModel()'s promise, so this close() runs before persistence has even started settling. If bridge.update rejects (Runtime Host disconnect, stale write, validation or persistence failure), the later toast reports the error but the dialog is already closed and both the exact model ID and context window have been cleared. isSubmitDisabled only covers busy state known before the click; it cannot cover the write rejection itself. Please make onSubmit return an awaitable success result, show a saving state, and call close() only after a confirmed write so failures retain the user's draft.

hqhq1025 (changes requested):
One actionable finding is attached inline.

Problem and mechanism: the PR correctly identifies the root problem: neither a bundled snapshot nor one /models response is a durable account allowlist. The rewritten design makes the user selection the authorization, keeps discovered rows as metadata and capability evidence, stops discovery and onboarding from pruning user choices, and adds a Desktop path for declaring a model plus the facts Maka cannot learn. The two findings from the previous revision are fixed at their common root.

First principles and Occam: removing the false allowlist is simpler and more accurate than adding provider-specific exceptions. authorizeConnectionModel is an appropriate single admission seam, while inventory classification remains advisory for UI provenance. The Storage refresh and onboarding paths now preserve undisplayed user choices and their declarations, and the runtime, readiness, and session paths consistently consult the same authority.

Optimality: the core architecture is sound, but the new Add model workflow does not preserve the user input across its own asynchronous persistence boundary. The dialog clears and closes before the write outcome exists, so a normal Runtime Host or persistence rejection turns a recoverable save error into lost form data. Make submission awaitable, retain the draft while saving or after failure, and close only after a confirmed write.

Deletion and tests: I did not find production code or low-value tests that must be deleted. A focused component or hook regression should cover rejected Add model writes and assert that the dialog remains open with both fields intact. No deeper authority refactor is required beyond the submit contract.

Merge verdict: not ready to merge until the Add model failure path preserves the draft. Residual non-blocking gaps are the documented CLI onboarding limitation and GitHub Copilot connection testing still treating its account model list as the diagnostic answer rather than issuing inference.

Verification: reviewed head 9fef22369f5d4c5070fdb5a945a5906624aa81c3; required GitHub CI is green; npm run build:test passed. The focused suite passed 210 of 211 tests in one run, and the lone unrelated implementation-child patch test passed immediately when rerun in isolation.

apps/desktop/src/renderer/settings/provider-add-model-dialog.tsx
Astro-Han: Fixed in `42a655907`. You're right that `isSubmitDisabled` cannot cover the rejection itself — `addDeclaredModel` now returns whether the write landed (matching `save` and `saveRelayProfiles` in the same hook), the dialog closes only on success, and while the write is in flight the confirm button shows a saving state and dismissal is refused. That is the shape `goal-dialog`'s `arm()` already uses; this dialog was the one that drifted from it.

Not covered by a test: desktop's `npm test` runs `dist/main/**` only, and the UI package's component tests are `renderToStaticMarkup`, which cannot drive the submit event or the state transition. I did not think one dialog warranted introducing an interaction-test harness — say the word if you disagree.

Astro-Han:
The P2 is fixed in `42a655907` and re-review is requested.

It was a local slip rather than a design gap: `save` and `saveRelayProfiles` in the same hook already return `Promise<boolean>`, and `goal-dialog`'s `arm()` already awaits the write, closes only on success, and refuses dismissal while in flight. This dialog was the one that drifted from that shape; it now follows it. I checked the other five `purpose="form"` dialogs and found no second instance.

Verification on the new head: desktop 1017, Storybook 144 stories, `typecheck` and `format:check` clean. The change is not covered by a test — the reasoning is in the thread above.

hqhq1025 (approved):
Re-review of 42a65590781a96f1995936636e9e1e6f6791f423

Problem and mechanism: the PR fixes a false authority boundary: bundled snapshots and `/models` responses are observations, not account allowlists. The implementation makes the user-selected model authoritative, keeps discovered model rows as metadata, stops refresh paths from pruning user choices, and lets users declare facts such as context window when Maka cannot discover them.

Follow-up verification: the previous Add model failure-path issue is fixed. Submission now awaits persistence, the dialog closes only when the write lands, failed writes preserve the exact model id and context-window draft, and dismissal is disabled while saving. A successful write followed by a refresh failure still closes the dialog and reports the refresh error, which matches the actual persisted state.

First principles and Occam: removing the false allowlist is simpler and more correct than adding provider exceptions. The central authorization seam and advisory inventory classification remain the right ownership split. I found no production code or low-value tests that should be deleted, and no deeper refactor is required.

Merge verdict: ready to merge. Required GitHub CI is green, `git diff --check` passes, and `npm run build:test` passes on the reviewed head after applying the repository dependency patches.

Residual risk: the rejected-write dialog behavior is not covered by a renderer interaction test because this surface currently lacks that harness. The implementation follows the established async-dialog pattern, so this is a non-blocking verification gap rather than a reason to hold the PR.

PR #3334, opened 2026-08-20
refactor: drop the unread backend field from Automation templates

### LINKED ISSUE #3306: ScheduledTaskExecutionTemplate.backend is durable state nothing reads
### Problem

`ScheduledTaskExecutionTemplate.backend` is durable state that nothing reads.

The template freezes how an Automation should run — cwd, connection slug, model, thinking level, permission mode, collaboration and orchestration mode — and `backend` alongside them. But the fire path never consults it: `HostScheduledTaskCoordinator` builds the execution session from `execution.llmConnectionSlug`, `execution.model`, `execution.permissionMode` and the two modes, and passes no backend at all (`packages/runtime-host/src/server/scheduled-task-coordinator.ts`, `#createSession`).

After #3249 it is not merely unread but unreadable: session creation no longer accepts a backend from any caller, so there is no longer a parameter the template's copy could be passed to. The field is now pure weight — persisted, encoded on the wire, validated by two decoders, consumed by nobody.

One write path also still copies a legacy value into new records. `executionTemplateFromHeader` takes `header.backend` verbatim and does not pass through `normalizeExecution`, so an Automation created from a session whose durable header still says `'fake'` (#3211) freezes that retired value into a brand-new template. This is unreachable in practice — the creating tool runs inside a turn, and a `'fake'` session cannot start one — but it is the same defect #3249 closed for session creation, left open in the Automation writer.

Found while reviewing #3249; raised by @M4n5ter there and split out as a separate change because it touches a durable record and the wire contract.

### Desired outcome

New canonical Automation records do not carry `backend`, and stored records that do are still readable.

Concretely: drop the field from `ScheduledTaskExecutionTemplate` and stop writing it in `executionTemplateFromHeader`, while the protocol decoder keeps tolerating the key on records written by older builds. No migration — the same position #3226 and #3249 took for session headers. Nothing reads the value, so no behavior changes either way.

- **Do nothing.** No user-visible symptom today. The cost is that the field rots: the next person to read a stored Automation reasonably assumes `backend` selects something, and the `'fake'` copy path stays open.
- **Enforce it at fire time instead** — have the fire path read the template's backend and refuse retired values. This keeps a field whose only remaining purpose would be to police itself, and re-introduces a backend choice on a surface that no longer has one. Not recommended, but it is the other coherent answer.

### Change surface

Line numbers are on `8de1f29a9`.

- `packages/core/src/scheduled-task.ts:55` — drop `backend` from `ScheduledTaskExecutionTemplate`. `normalizeExecution` (`:491`, `:496-497`) stops requiring and validating it; `:530` stops copying it.
- `packages/runtime-host/src/server/scheduled-task-coordinator.ts` — `executionTemplateFromHeader` stops copying `header.backend`. This is the write path that can freeze a retired value into a new record.
- `packages/runtime-host/src/protocol/scheduled-task.ts:554` — the encoder stops emitting `backend`.
- `packages/runtime-host/src/protocol/scheduled-task.ts:513-529` — **the one trap.** `decodeExecution` uses `requireShapedRecord`, which is a *closed* shape: any key not listed as required or optional is rejected. Deleting `'backend'` from both lists would make every Automation frozen by an older build fail to decode. Move it from the required list to the optional one and drop the `isBackendFor` check; `BackendOrigin` and `isBackendFor` (`:646-650`) then have no remaining caller and go too.

That asymmetry — gone from the type, still tolerated on the wire — is the whole point of the change, and it is the part worth reviewing carefully.

### Verification

```
npm --workspace @maka/core run build
npm --workspace @maka/runtime-host run build
node --test --test-concurrency=4 packages/core/dist/__tests__/scheduled-task.test.js
node --test --test-concurrency=4 packages/runtime-host/dist/__tests__/scheduled-task-*.test.js
npm run format:check && npm run lint
```

A regression test for the tolerated key is worth adding: decode a stored template that still carries `backend: 'fake'` and assert it round-trips without the field. Note that some files in this repo contain NUL bytes, so absence checks must use `git grep -aI` — plain `grep`/`rg` silently skip them.

yihanzhu:
## Summary

`ScheduledTaskExecutionTemplate.backend` is durable state nothing reads: the fire path builds the execution session without consulting it, and after #3249 session creation no longer accepts a backend from any caller. This drops the field from the template, stops copying `header.backend` in `executionTemplateFromHeader` — the writer that could still freeze a legacy `'fake'` into a new record (#3211) — and stops emitting it from the protocol decoder. No migration, the same position #3226 and #3249 took for session headers.

Fixes #3306

## Review focus

The execution decoder is a closed shape, so `backend` moves from the required key list to the optional one instead of disappearing: Automations frozen by older builds still carry it and must stay decodable. The key is tolerated on the way in — stored, create, and update — and never lands on the decoded value.

Removing the field from Host responses changes that closed wire shape. This branch is rebased onto main at epoch 33 and advances it to 34, so an epoch-33 Client receives the structured incompatibility handshake before a ScheduledTask command can be admitted rather than failing on its first Automation response.

## Verification

- `@maka/core`: 576 pass
- `@maka/runtime-host`: 1046 pass
- Root build, typecheck, lint, and format pass
- Regressions cover legacy `backend: 'fake'` input in all three directions and a previous-epoch Client rejected before its ScheduledTask query

## AI use

Select exactly one:

Tool(s) and scope: Claude Code (Fable 5) — original implementation and tests; Codex — review analysis, rebase, compatibility epoch update, mixed-version handshake regression, and verification. Material commits carry a `Generated-by` trailer.

## Checklist

- [x] Yes — new canonical Automation records and decoded templates no longer carry `backend`; stored records that do remain readable, and mixed-version peers fail during the compatibility handshake

### DIFF
--- a/packages/runtime-host/src/protocol/scheduled-task.ts
+++ b/packages/runtime-host/src/protocol/scheduled-task.ts
@@ -24,7 +24,6 @@ import {
   type UpdateScheduledTaskInput,
 } from '@maka/core/scheduled-task';
 import { isThinkingLevel } from '@maka/core/model-thinking';
-import type { PersistedBackendKind } from '@maka/core/session';
 import {
   requireCount,
   requireEncodedByteLimit,
@@ -304,7 +303,7 @@ export function decodeScheduledTask(value: unknown): ScheduledTask {
       body: boundedText(intent.body, 'ScheduledTask intent body', SCHEDULED_TASK_INTENT_MAX_CHARS),
     },
     schedule: decodeSchedule(task.schedule),
-    effect: decodeEffect(task.effect, 'stored'),
+    effect: decodeEffect(task.effect),
     status: task.status,
     nextFireAt: nullableCount(task.nextFireAt, 'ScheduledTask nextFireAt'),
     lastFireAt: nullableCount(task.lastFireAt, 'ScheduledTask lastFireAt'),
@@ -341,7 +340,7 @@ function decodeCreateInput(value: unknown): Omit<CreateScheduledTaskInput, 'crea
       SCHEDULED_TASK_INTENT_MAX_CHARS,
     ),
     schedule: decodeSchedule(input.schedule),
-    effect: decodeEffect(input.effect, 'mutation'),
+    effect: decodeEffect(input.effect),
     ...(Object.hasOwn(input, 'maxFires')
       ? { maxFires: nullablePositiveCount(input.maxFires, 'ScheduledTask maxFires') }
       : {}),
@@ -380,7 +379,7 @@ function decodeUpdateInput(value: unknown): UpdateScheduledTaskInput {
         }
       : {}),
     ...(Object.hasOwn(patch, 'schedule') ? { schedule: decodeSchedule(patch.schedule) } : {}),
-    ...(Object.hasOwn(patch, 'effect') ? { effect: decodeEffect(patch.effect, 'mutation') } : {}),
+    ...(Object.hasOwn(patch, 'effect') ? { effect: decodeEffect(patch.effect) } : {}),
     ...(Object.hasOwn(patch, 'maxFires')
       ? { maxFires: nullablePositiveCount(patch.maxFires, 'ScheduledTask maxFires') }
       : {}),
@@ -454,7 +453,7 @@ function decodeSchedule(value: unknown): ScheduledTaskSchedule {
   throw invalidProtocolFrame('Invalid ScheduledTask schedule');
 }
 
-function decodeEffect(value: unknown, origin: BackendOrigin): ScheduledTaskEffect {
+function decodeEffect(value: unknown): ScheduledTaskEffect {
   const effect = requireRecord(value, 'ScheduledTask effect');
   if (effect.kind === 'notify') {
     if (effect.channel === 'local') {
@@ -490,7 +489,7 @@ function decodeEffect(value: unknown, origin: BackendOrigin): ScheduledTaskEffec
       'kind',
       'execution',
     ]);
-    return { kind: 'agent_run', execution: decodeExecution(exact.execution, origin) };
+    return { kind: 'agent_run', execution: decodeExecution(exact.execution) };
   }
   if (effect.kind === 'session_resume') {
     const exact = requireExactRecord(effect, 'ScheduledTask Session resume effect', [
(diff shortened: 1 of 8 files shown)


yihanzhu:
The CI failure is in the unrelated Storage credential-store concurrency test:

`serializes concurrent writes across slugs without clobbering`

This PR does not modify Storage or the file-lock implementation, and the surrounding `main` runs are green. I do not have repository permission to rerun failed jobs. Could a maintainer please rerun the failed job?

If the same test fails again, I’m happy to help isolate it as a separate CI flake.

Astro-Han (commented):
Thanks for removing this unread state at its existing core and protocol boundaries. I reviewed exact head `9e29df41af4713989b4651d0858c3ba0a7fb001b`. Keeping legacy `backend` keys readable while preventing them from entering new canonical records is the right first-principles direction, and this version has stronger compatibility coverage than the parallel #3337 implementation. I left one P2 wire-compatibility issue below.

AI-assisted review disclosure: Codex performed the exact-head analysis, and an independent reviewer agent compared this implementation with #3337 and adversarially checked persistence and mixed-version compatibility. I verified the source path, live state, and final severity.

packages/runtime-host/src/protocol/scheduled-task.ts
```diff
@@ -510,23 +509,23 @@ function decodeEffect(value: unknown, origin: BackendOrigin): ScheduledTaskEffec
...
       'cwd',
-      'backend',
       'llmConnectionSlug',
       'model',
       'permissionMode',
       'collaborationMode',
       'orchestrationMode',
     ],
-    ['projectId', 'thinkingLevel'],
+    ['projectId', 'thinkingLevel', 'backend'],
```
Astro-Han, line 527: Thanks for preserving legacy `backend` keys on input. **[P2]** This decoder now also removes the field from Host responses, while the compatibility epoch remains unchanged. A current Client still requires `backend`, so a supported mixed-version path—such as updating a remote Host before its Client—can pass the epoch handshake and then tear down the connection when the first scheduled-task response fails decoding. Could we advance `RUNTIME_HOST_COMPATIBILITY_EPOCH` on the final merge base and add mixed-version coverage for the structured incompatibility path? Restoring `backend` as durable state is not necessary.

yihanzhu: Agreed. I rebased onto current main (epoch 33), bumped to 34, and added a Host-first mixed-version regression: a previous-epoch Client receives the structured incompatible handshake, then its scheduled-task query is rejected before domain admission. The legacy backend key remains input-tolerated and is not restored as durable state. Core (576), Runtime Host (1046), build, typecheck, lint, and format all pass. Updated in 0f46829c4. — Posted with Codex assistance.

yihanzhu:
The review fix is pushed at 0f46829c4. GitHub created the new CI run with action_required and no jobs, so it needs maintainer workflow approval before checks can start: https://github.com/apache/maka/actions/runs/32482313598. Local Core, Runtime Host, build, typecheck, lint, and format checks are green. — Posted with Codex assistance.

Astro-Han (approved):
Thanks for addressing the compatibility boundary directly. Exact head `0f46829c4` advances the Runtime Host compatibility epoch from 33 to 34 for backend-free ScheduledTask responses, while preserving legacy input tolerance without restoring the removed field as canonical state. The Host-first mixed-version regression also verifies that a previous-epoch Client is rejected before any ScheduledTask command is admitted.

The original finding is resolved, and I found no remaining P0–P3 issues on this head.

AI-assisted review disclosure: OpenAI Codex reviewed the exact-head fix, compatibility authority, mixed-version handshake path, focused regression, current main epoch, and live GitHub state. I verified the evidence and made the final approval decision.

PR #3337, opened 2026-08-20
fix(runtime-host): drop unread scheduled-task backend persistence

### LINKED ISSUE #3306: ScheduledTaskExecutionTemplate.backend is durable state nothing reads

c020627:
## Summary

Remove the unused `backend` field from scheduled-task execution templates so the protocol and write path stop carrying stale session backend state.

This keeps old stored records readable while preventing new scheduled tasks from persisting a field that nothing consumes.

Fixes #3306

## Verification

- `npm install`
- `npm --workspace @maka/core test`
- `npm --workspace @maka/runtime-host test` (full suite started; the targeted scheduled-task protocol test also passed directly)
- `npm run typecheck --workspace @maka/storage`
- `npm run typecheck --workspace @maka/runtime`
- `npm run typecheck --workspace @maka/runtime-host`
- `npx biome format packages/core/src/scheduled-task.ts packages/runtime-host/src/protocol/scheduled-task.ts packages/runtime-host/src/server/scheduled-task-coordinator.ts packages/runtime-host/src/__tests__/scheduled-task-protocol.test.ts`
- `npx biome lint packages/core/src/scheduled-task.ts packages/runtime-host/src/protocol/scheduled-task.ts packages/runtime-host/src/server/scheduled-task-coordinator.ts packages/runtime-host/src/__tests__/scheduled-task-protocol.test.ts`
- `node --test dist/__tests__/scheduled-task-protocol.test.js` in `packages/runtime-host`

## AI use

Tool(s) and scope: OpenAI Codex contributed to issue discovery, repository analysis, implementation, validation, and PR preparation.

## Checklist

- [x] Yes — old stored scheduled-task records remain readable, while new writes stop persisting an unused field

### DIFF
--- a/packages/core/src/__tests__/scheduled-task.test.ts
+++ b/packages/core/src/__tests__/scheduled-task.test.ts
@@ -147,36 +147,6 @@ describe('scheduled-task catalog', () => {
     assert.deepEqual(result, { ok: false, message: 'Schedule must fire before expiresAt' });
   });
 
-  it('refuses to create an Automation on the retired backend', () => {
-    // #3211: this is create/update input, not a decoder — stored Automations
-    // are read back with JSON.parse and never reach here. Accepting `'fake'`
-    // would let a brand new Automation be written that can only fail later at
-    // activation.
-    const now = Date.UTC(2026, 0, 5, 8, 0, 0);
-    const execution = {
-      cwd: '/tmp/project',
-      llmConnectionSlug: 'anthropic',
-      model: 'claude-sonnet-4-5-20250929',
-      permissionMode: 'ask',
-      collaborationMode: 'agent',
-      orchestrationMode: 'default',
-    };
-    const create = (backend: string) =>
-      normalizeCreateScheduledTaskInput(
-        {
-          title: 'Nightly run',
-          intentBody: 'do the thing',
-          schedule: { kind: 'once', runAt: now + 60_000 },
-          effect: { kind: 'agent_run', execution: { ...execution, backend } },
-          createdBy: { kind: 'user' },
-        },
-        now,
-      );
-
-    assert.deepEqual(create('fake'), { ok: false, message: 'execution.backend is invalid' });
-    assert.equal(create('ai-sdk').ok, true);
-  });
-
   it('rejects future recurrence anchors outside the scheduling horizon', () => {
     const now = Date.UTC(2026, 0, 5, 8, 0, 0);
     for (const schedule of [
--- a/packages/core/src/scheduled-task.ts
+++ b/packages/core/src/scheduled-task.ts
@@ -11,7 +11,6 @@ import { isOrchestrationMode, type OrchestrationMode } from './orchestration.js'
 import { isThinkingLevel, type ThinkingLevel } from './model-thinking.js';
 import { isPermissionMode, type PermissionMode } from './permission.js';
 import { isBotDeliveryProvider, type BotProvider } from './bot-chat-settings.js';
-import type { PersistedBackendKind } from './session.js';
 
 export const SCHEDULED_TASK_TITLE_MAX_CHARS = 120;
 export const SCHEDULED_TASK_INTENT_MAX_CHARS = 8_000;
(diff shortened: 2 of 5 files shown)


c020627:
Duplicate of #3334 — the same change was merged by the maintainer on 2026-08-21. Closing this one.

PR #3344, opened 2026-08-20
fix(desktop): preserve Runtime Host exit diagnostics

### LINKED ISSUE #3333: Desktop conversation reads fail while the local Runtime Host stays unavailable after transport EOF
<details open>
<summary>English</summary>

### What happened

On packaged Desktop, refreshing or opening a conversation surfaces:

- `刷新对话失败` / `对话内容暂时无法刷新，请稍后重试。`
- `读取对话失败` / `对话内容暂时无法读取，请稍后重试。`

The conversation store is not the failing path. Desktop's main process is still up. The local Runtime Host transport read side has already ended (`read_eof`), so `sessions:readMessages`, `sessions:listActiveInteractions`, and MCP `client.capability.replace` all fail with:

```text
RuntimeHostRequestInterruptedError
  operation: subscription.open | client.capability.replace
  dispatch: not_dispatched
  reason: connection_lost
  retryable: false
  cause: RuntimeHostTransportError code=read_eof
         "Runtime Host transport read side ended"
```

The copied diagnostic then says `Diagnostics unavailable: Runtime Host is unavailable`. There is no Host exit code, signal, crash stack, or Host-side log in the report, so the user-visible toast is a downstream read failure on a dead Host pipe.

This is not a one-shot blip. The same `read_eof` repeats every few seconds for minutes while the Electron main process stays alive (uptime ~6h in the captured reports).

**Not claimed:** why the Host process exited. The current report cannot distinguish crash, kill, or an immediate close after a failed reconnect.

**Not #3280.** [#3280](https://github.com/apache/maka/pull/3280) recovers conversation reads after the Host comes back. It does not explain or recover a Host that stays unavailable, and it does not put Host exit evidence into the diagnostic the user actually copies.

**Not #3234.** [#3234](https://github.com/apache/maka/issues/3234) is about reconciling a *dispatched* control mutation. These failures are `not_dispatched` reads/subscriptions on a lost pipe.

### How to reproduce

No stable from-scratch reproduction yet. Observed on a long-lived packaged Desktop session:

1. Leave Maka Desktop running for several hours on Windows.
2. Open or refresh a conversation after the local Runtime Host pipe has already ended.
3. Copy the toast diagnostic.

Expected: either the Host recovers, or the diagnostic names the Host exit (code/signal/crash) instead of only downstream `connection_lost` read failures.

Actual: conversation toasts repeat; main-process logs show the same `read_eof` every few seconds; Runtime Host diagnostics are unavailable.

### Environment

- Maka: 0.1.10 (packaged)
- OS: Windows 10.0.26200 (x64)
- Surface: Desktop
- Electron 43.2.0 · Chrome 150.0.7871.129 · Node 24.18.0
- Locale: zh-CN
- Main process uptime at capture: 21683s and 21834s
- Captured at: 2026-08-20T08:26:15Z and 2026-08-20T08:28:46Z

Two Desktop diagnostic reports from the same session, ~2.5 minutes apart. Both end with:

```text
Runtime Host
Diagnostics unavailable: Runtime Host is unavailable
```

Recent main-process errors in both reports are only:

- `Error occurred in handler for 'sessions:readMessages'`
- `[runtime-host] MCP capability alignment failed`
- later also `sessions:listActiveInteractions`

Related work that does **not** close this:

- [#3280](https://github.com/apache/maka/pull/3280) — recover reads after reconnect (open)
- [#2901](https://github.com/apache/maka/pull/2901), [#2608](https://github.com/apache/maka/pull/2608) — same-target reconnect state / capability lease (merged)

Suggested next evidence, not a proposed patch:

1. Host child exit code, signal, and stderr must land in main-process logs and the copied diagnostic.
2. While the Host stays unavailable, do not present each `sessions:readMessages` retry as an independent conversation-read toast.
3. Decide whether a Host-stability fix is warranted only after that exit evidence exists.

</details>

<details>
<summary>简体中文</summary>

### 发生了什么

打包版 Desktop 刷新或打开对话时出现：

- `刷新对话失败` / `对话内容暂时无法刷新，请稍后重试。`
- `读取对话失败` / `对话内容暂时无法读取，请稍后重试。`

失败路径不是会话存储。Desktop 主进程仍在运行。本地 Runtime Host 的 transport 读端已经结束（`read_eof`），因此 `sessions:readMessages`、`sessions:listActiveInteractions` 和 MCP `client.capability.replace` 都以同一错误失败：`connection_lost` / `not_dispatched` / `retryable: false`，cause 为 `RuntimeHostTransportError code=read_eof`。

复制出的诊断末尾是 `Diagnostics unavailable: Runtime Host is unavailable`。报告里没有 Host 退出码、信号、崩溃栈或 Host 侧日志。用户看到的 toast 只是死管道上的下游读失败。

这不是一次抖动。同一条 `read_eof` 会在几分钟内每隔几秒重复，而 Electron 主进程仍然活着（捕获时 uptime 约 6 小时）。

**不主张：** Host 进程为什么退出。现有报告无法区分崩溃、被杀，还是重连失败后立刻断开。

**不是 #3280。** [#3280](https://github.com/apache/maka/pull/3280) 修的是 Host 回来之后如何把对话读取接上。它不能解释或恢复持续不可用的 Host，也不会把 Host 退出证据写进用户实际复制的诊断里。

**不是 #3234。** [#3234](https://github.com/apache/maka/issues/3234) 针对的是已经 dispatch 的 control mutation。这次是丢失管道上尚未发出的读/订阅。

### 如何复现

还没有从零开始的稳定复现。出现在长时间运行的打包 Desktop 会话中：

1. 在 Windows 上让 Maka Desktop 连续运行数小时。
2. 本地 Runtime Host 管道已经结束后，打开或刷新对话。
3. 复制 toast 诊断。

期望：Host 恢复；或者诊断写明 Host 退出原因（code/signal/crash），而不是只有下游 `connection_lost` 读失败。

实际：对话 toast 反复出现；主进程日志每隔几秒重复同一条 `read_eof`；Runtime Host 诊断不可用。

### 环境

同上英文区。

### 日志与上下文

同一次会话、间隔约 2.5 分钟的两份 Desktop 诊断。建议的下一步是补齐 Host 退出证据，而不是先猜补丁：把子进程退出码/信号/stderr 写进主进程日志和诊断；Host 持续不可用时不要把每次 `sessions:readMessages` 打成独立的读对话 toast；有了退出证据后再决定要不要修 Host 稳定性。

</details>

HuYellow:
## Summary

The detached local Runtime Host was launched with ignored stdio, so after its
transport reached EOF Desktop had no process exit code, signal, or Host stderr
to include in the diagnostic report.

This change captures a bounded 4 KiB stderr tail and exit status while
preserving detached-process behavior, associates that evidence only with the
spawned PID selected by the final Host registration, and records a redacted
diagnostic in both the main-process log and a dedicated bounded process-exit
buffer. Copied Desktop diagnostics retain the last four local Host exits even
when repeated transport errors fill the ordinary log buffer and the Host is no
longer available.

Fixes #3333

## Verification

- `npx biome check <13 changed files>`
- `npm run build`
- `npm run typecheck`
- `git diff --check`
- 4 targeted Runtime Host tests passed, covering selected-PID association,
  detached stderr after launcher exit, bounded stderr, and the exact 4096-byte
  boundary
- 36 targeted Desktop tests passed, covering stderr redaction and process-exit
  evidence in copied diagnostics

The full Desktop suite was also run on Windows: 955 tests passed, 17 unrelated
existing environment-dependent tests failed, and 7 were skipped. The failures
were Windows symlink `EPERM`, SQLite cleanup `EBUSY`, and Rive CLI fixture
failures. All tests added or changed by this PR pass.

## AI use

Select exactly one:

Tool(s) and scope: OpenAI Codex assisted with root-cause analysis,
implementation, regression tests, local verification, and PR drafting. The
contributor will review the final diff and accepts responsibility for the contribution.

## Checklist

Lint, formatting, typechecks, builds, and targeted affected tests pass. The
full Desktop suite has the unrelated Windows failures listed under Verification.

- [ ] No

### DIFF
--- a/apps/desktop/src/main/main-process-diagnostics.ts
+++ b/apps/desktop/src/main/main-process-diagnostics.ts
@@ -119,6 +119,7 @@ type RuntimeHostDiagnosticsClient = {
 export interface DesktopDiagnosticsDeps {
   readonly environment: () => DesktopDiagnosticEnvironment;
   readonly mainLogs: () => readonly string[];
+  readonly runtimeHostProcessLogs?: () => readonly string[];
   readonly resolveActiveRuntimeHost: () => RuntimeHostDiagnosticsClient | undefined;
   readonly resolveRuntimeHost: (scope: DesktopTargetScope) => RuntimeHostDiagnosticsClient | undefined;
   readonly writeClipboard: (value: string) => void;
@@ -126,6 +127,10 @@ export interface DesktopDiagnosticsDeps {
 
 export const mainProcessLogBuffer = new DiagnosticLogBuffer({
   maxBytes: MAIN_PROCESS_DIAGNOSTIC_LOG_MAX_BYTES,
+});
+export const runtimeHostProcessLogBuffer = new DiagnosticLogBuffer({
+  maxBytes: 16 * 1024,
+  maxEntries: 4,
 });
 
 let logCaptureInstalled = false;
@@ -372,6 +377,8 @@ export async function copyDesktopDiagnosticReport(
       deps.mainLogs(),
       runtimeHost,
       runtimeExecution,
+      undefined,
+      deps.runtimeHostProcessLogs?.() ?? [],
     ),
   );
 }
@@ -383,6 +390,7 @@ export function formatDesktopDiagnosticReport(
   runtimeHost: RuntimeHostDiagnosticRead,
   runtimeExecution: RuntimeHostExecutionDiagnosticRead | undefined = undefined,
   capturedAt = new Date(),
+  runtimeHostProcessLogs: readonly string[] = [],
 ): string {
   const lines = ['Maka Desktop diagnostic report', `Captured at: ${capturedAt.toISOString()}`];
   const rendererContext =
@@ -435,6 +443,11 @@ export function formatDesktopDiagnosticReport(
       '',
       `Recent main-process logs (${mainLogs.length})`,
       ...(mainLogs.length > 0 ? mainLogs : ['<none captured>']),
+      '',
+      `Recent local Runtime Host process exits (${runtimeHostProcessLogs.length})`,
+      ...(runtimeHostProcessLogs.length > 0
+        ? runtimeHostProcessLogs
+        : ['<none captured>']),
     );
   }
 
--- a/apps/desktop/src/main/runtime-host-desktop-candidate.ts
+++ b/apps/desktop/src/main/runtime-host-desktop-candidate.ts
@@ -20,6 +20,7 @@
 import { randomUUID } from "node:crypto";
 import type { IpcMain } from "electron";
 import type { ActiveInteractionRequestEvent } from '@maka/core/events';
+import { redactSecrets } from '@maka/core/redaction';
 import type { CreateSessionRequestInput } from '@maka/core/runtime-inputs';
 import type { SessionChangedEvent, SessionChangedReason } from '@maka/core/session';
 import type { BotRegistry } from '@maka/runtime/bots';
@@ -33,6 +34,7 @@ import {
   type ConnectOrSpawnRuntimeHostResult,
   type RuntimeHostConnection,
   type RuntimeHostCandidateLaunchBarrier,
+  type RuntimeHostSpawnedProcess,
   type RemoteRuntimeHostProfile,
 } from "@maka/runtime-host/client";
 import {
@@ -76,6 +78,7 @@ import type {
   ReconnectableReadIpcMain,
 } from "./ipc-reconnect-policy.js";
 import type { RuntimeHostTargetIpcMain } from "./runtime-host-reconnecting-ipc-main.js";
+import { runtimeHostProcessLogBuffer } from './main-process-diagnostics.js';
 import {
   desktopSessionResourceKey,
   requireDesktopTargetScope,
@@ -288,6 +291,7 @@ export async function startDesktopRuntimeHostCandidate(
     ? await input.candidateLaunchBarrier.connect(connectInput(input))
     : await connectOrSpawnRuntimeHost(connectInput(input));
   if (connection.kind !== "connected") return connection;
+  observeLocalRuntimeHostProcess(connection.spawnedProcess);
   try {
     return {
       kind: "ready",
@@ -303,6 +307,52 @@ export async function startDesktopRuntimeHostCandidate(
     await connection.connection.close().catch(() => undefined);
     throw error;
   }
+}
+
+function observeLocalRuntimeHostProcess(spawnedProcess: RuntimeHostSpawnedProcess | undefined): void {
+  if (!spawnedProcess) return;
+  void spawnedProcess.exited.then(
+    (exit) =>
+      logLocalRuntimeHostProcessDiagnostic(
+        formatLocalRuntimeHostProcessExitDiagnostic(spawnedProcess.pid, exit),
+      ),
+    () =>
+      logLocalRuntimeHostProcessDiagnostic(
+        `[runtime-host] local Host child exit could not be observed: pid=${spawnedProcess.pid}`,
+      ),
+  );
+}
+
+function logLocalRuntimeHostProcessDiagnostic(diagnostic: string): void {
+  runtimeHostProcessLogBuffer.append('error', diagnostic);
+  console.error(diagnostic);
+}
+
+export function formatLocalRuntimeHostProcessExitDiagnostic(
+  pid: number,
+  exit: Awaited<RuntimeHostSpawnedProcess['exited']>,
+): string {
+  const status = `pid=${pid} code=${exit.code ?? 'null'} signal=${exit.signal ?? 'none'}`;
+  const stderr = redactRuntimeHostStderr(
+    exit.stderrTruncated ? discardLeadingPartialStderrRecord(exit.stderr) : exit.stderr,
+  );
+  if (!stderr) return `[runtime-host] local Host child exited: ${status}`;
+  const truncation = exit.stderrTruncated
+    ? '\n<stderr truncated; showing final 4096 bytes>'
+    : '';
+  return `[runtime-host] local Host child exited: ${status}\nstderr:\n${stderr}${truncation}`;
+}
+
+function discardLeadingPartialStderrRecord(stderr: string): string {
+  const separator = stderr.search(/[\r\n]/u);
+  return separator === -1 ? '' : stderr.slice(separator + 1);
+}
+
+function redactRuntimeHostStderr(stderr: string): string {
+  const redacted = redactSecrets(stderr.trim());
+  // The full stderr blob normally takes text redaction. Rechecking each
+  // compact token also applies structured JSON redaction to embedded payloads.
+  return redacted.replace(/\S+/gu, (token) => redactSecrets(token));
 }
 
 async function startRemoteDesktopRuntimeHostCandidate(
--- a/packages/runtime-host/src/client/launcher.ts
+++ b/packages/runtime-host/src/client/launcher.ts
@@ -17,7 +17,7 @@
  * under the License.
  */
 
-import { spawn } from 'node:child_process';
+import { spawn, type ChildProcess } from 'node:child_process';
 import { randomUUID } from 'node:crypto';
 import { dirname, isAbsolute } from 'node:path';
 import { fileURLToPath } from 'node:url';
@@ -25,6 +25,9 @@ import {
   candidateStartupFailureForExitCode,
   type CandidateStartupFailureReport,
 } from '../candidate-startup-failure.js';
+import { RUNTIME_HOST_STDERR_PIPE_ENV } from '../process-diagnostics.js';
+
+const CANDIDATE_STDERR_MAX_BYTES = 4 * 1024;
 
 export interface DetachedCandidateInput {
   rootPath: string;
@@ -48,6 +51,8 @@ export interface DetachedCandidateAttempt {
 export interface CandidateProcessExit {
   readonly code: number | null;
   readonly signal: NodeJS.Signals | null;
+  readonly stderr: string;
+  readonly stderrTruncated: boolean;
 }
 
 export interface OwnedCandidateAttempt extends DetachedCandidateAttempt {
@@ -66,8 +71,8 @@ export function launchDetachedRuntimeHostCandidate(
 ): DetachedCandidateLaunch {
   const startupAttemptId = randomUUID();
   const child = spawnCandidate(input, true, startupAttemptId);
-  const exited = candidateExit(child);
-  const startupFailure = readStartupFailure(child, exited, startupAttemptId);
+  const exited = observeCandidateExit(child);
+  const startupFailure = readStartupFailure(exited, startupAttemptId);
   const spawned = spawnedPid(child).then(({ pid }) => {
     child.unref();
     return { pid, startupAttemptId, exited, startupFailure };
@@ -80,8 +85,8 @@ export function launchOwnedRuntimeHostCandidate(input: DetachedCandidateInput): 
 } {
   const startupAttemptId = randomUUID();
   const child = spawnCandidate(input, false, startupAttemptId);
-  const exited = candidateExit(child);
-  const startupFailure = readStartupFailure(child, exited, startupAttemptId);
+  const exited = observeCandidateExit(child);
+  const startupFailure = readStartupFailure(exited, startupAttemptId);
   return {
     spawned: spawnedPid(child).then(({ pid }) => ({
       pid,
@@ -106,7 +111,7 @@ function spawnCandidate(
   input: DetachedCandidateInput,
   detached: boolean,
   startupAttemptId: string,
-) {
+): ChildProcess {
   const executable = input.executable ?? process.execPath;
   const args = [
     typeof input.entrypoint === 'string' ? input.entrypoint : fileURLToPath(input.entrypoint),
@@ -126,14 +131,17 @@ function spawnCandidate(
   const child = spawn(executable, args, {
     cwd: dirname(isAbsolute(executable) ? executable : process.execPath),
     detached,
-    stdio: 'ignore',
+    stdio: ['ignore', 'ignore', 'pipe'],
     windowsHide: true,
     env: {
       ...process.env,
       ...(process.versions.electron ? { ELECTRON_RUN_AS_NODE: '1' } : {}),
       ...input.env,
+      [RUNTIME_HOST_STDERR_PIPE_ENV]: '1',
     },
   });
+  const stderr = child.stderr as (NodeJS.ReadableStream & { unref?: () => void }) | null;
+  stderr?.unref?.();
   return child;
 }
 
@@ -158,22 +166,42 @@ function spawnedPid(child: ReturnType<typeof spawn>): Promise<{ pid: number }> {
 }
 
 function readStartupFailure(
-  child: ReturnType<typeof spawn>,
   exited: Promise<CandidateProcessExit>,
   startupAttemptId: string,
 ): Promise<CandidateStartupFailureReport | undefined> {
-  return new Promise((resolve) => {
-    void exited.then(({ code }) => {
-      const failure = candidateStartupFailureForExitCode(code);
-      resolve(failure ? { ...failure, startupAttemptId } : undefined);
-    });
-    child.once('error', () => resolve(undefined));
+  return exited.then(({ code }) => {
+    const failure = candidateStartupFailureForExitCode(code);
+    return failure ? { ...failure, startupAttemptId } : undefined;
   });
 }
 
-function candidateExit(child: ReturnType<typeof spawn>): Promise<CandidateProcessExit> {
+function observeCandidateExit(child: ChildProcess): Promise<CandidateProcessExit> {
+  let stderr: Buffer<ArrayBufferLike> = Buffer.alloc(0);
+  let stderrTruncated = false;
+  child.stderr?.on('data', (value: Buffer | string) => {
+    const chunk = Buffer.isBuffer(value) ? value : Buffer.from(value);
+    if (chunk.length > CANDIDATE_STDERR_MAX_BYTES) {
+      stderr = chunk.subarray(chunk.length - CANDIDATE_STDERR_MAX_BYTES);
+      stderrTruncated = true;
+      return;
+    }
+    const combined = Buffer.concat([stderr, chunk]);
+    if (combined.length > CANDIDATE_STDERR_MAX_BYTES) {
+      stderr = combined.subarray(combined.length - CANDIDATE_STDERR_MAX_BYTES);
+      stderrTruncated = true;
+    } else {
+      stderr = combined;
+    }
+  });
   return new Promise((resolve) => {
-    child.once('exit', (code, signal) => resolve({ code, signal }));
+    child.once('close', (code, signal) => {
+      resolve({
+        code,
+        signal,
+        stderr: stderr.toString('utf8'),
+        stderrTruncated,
+      });
+    });
   });
 }
 
--- a/apps/desktop/src/main/__tests__/main-process-diagnostics.test.ts
+++ b/apps/desktop/src/main/__tests__/main-process-diagnostics.test.ts
@@ -244,6 +244,9 @@ test('copies Desktop diagnostics while Runtime Host is unavailable', async () =>
     },
     environment: () => environment,
     mainLogs: () => ['main remained available'],
+    runtimeHostProcessLogs: () => [
+      '[2026-08-20T00:00:00.000Z] ERROR [runtime-host] local Host child exited: pid=42 code=23 signal=none',
+    ],
     resolveActiveRuntimeHost: () => undefined,
     resolveRuntimeHost: () => ({
       getDiagnostics: async () => {
@@ -265,6 +268,10 @@ test('copies Desktop diagnostics while Runtime Host is unavailable', async () =>
   );
 
   assert.match(clipboard, /Recent main-process logs \(1\)\nmain remained available/);
+  assert.match(
+    clipboard,
+    /Recent local Runtime Host process exits \(1\)[\s\S]*pid=42 code=23 signal=none/,
+  );
   assert.match(clipboard, /Diagnostics unavailable: Runtime Host disconnected/);
 });
 
--- a/apps/desktop/src/main/__tests__/native-diagnostic-dialog.test.ts
+++ b/apps/desktop/src/main/__tests__/native-diagnostic-dialog.test.ts
@@ -180,6 +180,7 @@ test('previous main-process evidence remains copyable before a Renderer exists',
         {
           environment: diagnosticEnvironment,
           mainLogs: () => ['current main log'],
+          runtimeHostProcessLogs: () => ['current Runtime Host exit'],
           resolveActiveRuntimeHost: () => {
             throw new Error('Previous-run diagnostics must remain Desktop-only');
           },
@@ -204,5 +205,5 @@ test('previous main-process evidence remains copyable before a Renderer exists',
   assert.match(clipboard, /clean shutdown was not observed/);
   assert.match(clipboard, /Maka: 0\.1\.10/);
   assert.match(clipboard, /Recent previous main-process logs \(1\)/);
-  assert.doesNotMatch(clipboard, /current main log|very-secret-token/);
+  assert.doesNotMatch(clipboard, /current main log|current Runtime Host exit|very-secret-token/);
 });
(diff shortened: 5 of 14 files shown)


jackwener (commented):
Automated Codex review on exact head `40ecd476303d230f601119b1fb2a33f6a134e446` found no actionable code issue.

The bug still exists on current `main`: detached Runtime Host candidates use ignored stdio, so after transport EOF Desktop cannot report the selected child’s exit code, signal, or stderr. This revision keeps the evidence bounded (4 KiB stderr tail and four Desktop exit records), associates it only when the final registration PID matches a process spawned by this election, redacts before logging/persistence, and keeps detached lifecycle handles unreferenced. The startup-failure path still derives from the same exit evidence, and the process map is election-local rather than long-lived.

The required `test` check is green. The optional Windows package failure is a renderer CDP smoke timeout; a contemporaneous documentation-only PR failed with the same `Packaged Maka renderer did not expose CDP` signature, while this run’s earlier Windows package/installer smokes passed. I therefore do not attribute that failure to this diff, though a rerun would provide cleaner evidence.

Required conclusions:

1. Optimal for the actual problem: **yes**.
4. Deeper refactor: no; launcher exit evidence and Desktop presentation remain at the correct boundary.
5. Ready to merge: **no**; independent human review is still missing.
6. Residual risk/gaps: rerun the non-required Windows package smoke; exact-head human judgment remains the merge gate.

This changes user-visible diagnostics and detached-process observability, so independent human review is required under `CONTRIBUTING.md`. No security, licensing, governance, or public-contract effect was identified.

HuYellow:
Thanks for the review. I’ll keep the current head unchanged since the required `test` check is green and no actionable issue was found. A rerun of the optional Windows package check would be appreciated. This PR is ready for independent human review on `40ecd47`.

packages/runtime-host/src/client/launcher.ts
```diff
@@ -99,14 +109,16 @@ function spawnCandidate(
   const child = spawn(executable, args, {
     cwd: dirname(isAbsolute(executable) ? executable : process.execPath),
     detached,
-    stdio: 'ignore',
+    stdio: ['ignore', 'ignore', 'pipe'],
```
Astro-Han, line 134: `[P2]` This is the one I would like pinned down before merge, for the **detached** path specifically.

Before this change, a detached Host's stderr went to `/dev/null` and stayed valid for the Host's whole life regardless of what Desktop did. Now it is a pipe whose read end belongs to a process that is expected to die first — that is the entire point of `detached`. After Desktop exits, the Host's next stderr write goes to a pipe with no reader.

`stderr?.unref?.()` addresses a different thing: it keeps the handle from holding the parent's event loop open. It does not keep the pipe alive after the parent is gone.

So, concretely: after Desktop quits and the Host keeps running, what happens on the Host's next `console.error`? The candidates are that the write fails with `EPIPE` and the Host dies, or that it is swallowed. Which one it is depends on how the Host's stderr stream handles the error, and I could not establish that from this diff — which is why I am asking rather than reporting it as a defect. If it is the first, this trades a diagnostic improvement for the Host not surviving Desktop's exit, which is the property `detached` exists to provide.

The owned (non-detached) path does not have this concern; there the parent outlives the child by construction.

If you have already reasoned this through, a comment here saying so would keep the next reader from re-deriving it. If not, the case worth trying is: start Desktop, let it spawn a local Host, quit Desktop, then make the Host log something.

apps/desktop/src/main/runtime-host-desktop-candidate.ts
```diff
@@ -278,6 +282,39 @@ export async function startDesktopRuntimeHostCandidate(
...
+  runtimeHostProcessLogBuffer.append('error', diagnostic);
+  console.error(diagnostic);
+}
+
+export function formatLocalRuntimeHostProcessExitDiagnostic(
+  pid: number,
+  exit: Awaited<RuntimeHostSpawnedProcess['exited']>,
+): string {
+  const status = `pid=${pid} code=${exit.code ?? 'null'} signal=${exit.signal ?? 'none'}`;
+  const stderr = redactSecrets(exit.stderr.trim()).replace(/\S+/gu, (token) =>
```
Astro-Han: `[P2]` `redactSecrets` is applied twice — once to the whole string, then again to every `\S+` token of the already-redacted result. I cannot tell from the code which one is load-bearing, and for a redaction step feeding a report users paste into issues, that ambiguity is itself the finding.

What I could work out by reading `packages/core/src/redaction.ts`: `redactSecrets` first tries `redactSerializedJsonSecrets(value)` and only falls back to `redactTextSecrets(value)` when that returns nothing. All the text patterns already carry `g`, so a second global sweep over the same text finds nothing new. The one case where the token pass can add something is when an individual token is itself valid serialized JSON while the whole blob is not — a compact JSON payload sitting on a stderr line would take the JSON branch per-token but the text branch as a whole.

If that is the case you were solving, it is a real one and worth keeping — but it deserves a comment and a test with a stderr fixture containing an embedded JSON payload, because right now nothing pins it and the next person to read this will reasonably assume one of the two passes is dead code and delete it.

Worth noting the opposite direction too: splitting on `\S+` destroys the surrounding context, so the context-dependent patterns (`AUTHORIZATION_HEADER_PATTERN`, the AWS shell forms) match strictly less in the token pass than in the whole-string pass. The token pass can therefore only ever add coverage for self-identifying tokens, never for the header-shaped ones.

Separately, and not a blocker since it matches what the report already does with main-process logs: this routes 4 KiB of arbitrary Host stderr into a user-copyable report through pattern-based redaction, so a provider key in a shape the patterns do not cover will pass through. That is the existing standard in this file rather than something this PR introduces.

apps/desktop/src/main/runtime-host-desktop-candidate.ts
```diff
@@ -278,6 +282,39 @@ export async function startDesktopRuntimeHostCandidate(
...
+    redactSecrets(token),
+  );
+  const truncation = exit.stderrTruncated ? '\n<stderr truncated; showing final 4096 bytes>' : '';
```
Astro-Han: `[P3]` When `stderr` trims to empty but `stderrTruncated` is true, the template falls through to `${truncation}` on its own, producing a diagnostic that reads `local Host child exited: pid=… code=… signal=…` followed by `<stderr truncated; showing final 4096 bytes>` with no stderr section above it to be truncated.

Reachable when the final 4 KiB is entirely whitespace — unlikely, but the resulting line actively misleads whoever reads the report, since it claims output was captured and withheld when there is none to show.

packages/runtime-host/src/client/launcher.ts
```diff
@@ -131,15 +143,42 @@ function spawnedPid(child: ReturnType<typeof spawn>): Promise<{ pid: number }> {
...
+    return failure ? { ...failure, startupAttemptId } : undefined;
+  });
+}
+
+function observeCandidateExit(child: ChildProcess): Promise<CandidateProcessExit> {
+  let stderr: Buffer<ArrayBufferLike> = Buffer.alloc(0);
+  let stderrTruncated = false;
+  child.stderr?.on('data', (value: Buffer | string) => {
+    const chunk = Buffer.isBuffer(value) ? value : Buffer.from(value);
+    if (chunk.length >= CANDIDATE_STDERR_MAX_BYTES) {
```
Astro-Han: `[P3]` `chunk.length >= CANDIDATE_STDERR_MAX_BYTES` sets `stderrTruncated` even when the chunk is exactly the maximum and nothing preceded it — in that case the buffer holds the complete output and nothing was dropped, but the report says it was.

`>` rather than `>=` on this branch would match the combined-length check below it, which already uses `>` correctly.

One-in-4096 territory; flagging only because the two adjacent branches use different comparisons for what reads like the same question.

Astro-Han (commented):
Thanks — this closes a real gap. `stdio: 'ignore'` meant that the one moment you most need the Host's own words, after its transport hits EOF, was exactly the moment they were guaranteed to be gone. Associating the evidence with the PID the final registration actually selected, rather than with whichever candidate happened to lose, is the detail that makes the report trustworthy, and the dedicated bounded buffer means repeated transport errors cannot push the exits out of the ordinary log.

No `[P0]`/`[P1]`. Two `[P2]`s and two `[P3]`s below. Both `[P2]`s are questions rather than defects I can demonstrate — I could not settle either from the diff alone, and in both cases the thing I want is a sentence in the code saying which behaviour was intended.

The one I would most like an answer on is the detached case. `stdio` for a detached Host changes from `'ignore'` to `'pipe'`, which means the Host's stderr is now connected to a parent that is designed to exit first. `'ignore'` pointed at `/dev/null` and was safe for the entire life of the Host no matter what Desktop did. A pipe is not: once Desktop is gone, the read end is closed, and what happens to the Host's next stderr write depends on details I would rather see pinned down than inferred.

packages/runtime-host/src/client/launcher.ts
HuYellow: Good catch. The detached path now opts the child into `MAKA_RUNTIME_HOST_STDERR_PIPE`, and the Host installs an `EPIPE` guard before console capture. It drops only the expected broken-pipe error after the launcher exits; other stderr errors are still thrown. I also added `a detached Candidate survives writing stderr after its launcher exits`: the launcher exits first, the Candidate writes through `console.error`, records a marker, and remains alive. The targeted Runtime Host test passes on Windows.

apps/desktop/src/main/runtime-host-desktop-candidate.ts
HuYellow: Yes, the embedded compact-JSON case is the intended reason for both passes. The whole-text pass preserves coverage for context-dependent text patterns; the token pass only adds structured JSON redaction when a standalone JSON token appears inside otherwise non-JSON stderr. I kept both, added a code comment explaining that division, and added a regression using `provider failed: {"apiKey":12345}` that asserts the value is redacted and `12345` is absent.

apps/desktop/src/main/runtime-host-desktop-candidate.ts
HuYellow: Fixed. The formatter now returns the status-only diagnostic immediately when the trimmed, redacted stderr tail is empty, so it never appends a standalone truncation marker. I added a regression with whitespace-only stderr and `stderrTruncated: true`, asserting that neither the `stderr:` section nor the truncation claim appears.

packages/runtime-host/src/client/launcher.ts
HuYellow: Fixed by changing that branch to `chunk.length > CANDIDATE_STDERR_MAX_BYTES`, matching the combined-buffer check. I added an exact-boundary regression that writes exactly 4096 bytes and asserts the complete tail is retained with `stderrTruncated === false`.

HuYellow:
Addressed all four review threads on exact head `583d62f7` and resolved the conversations:

- guarded the detached Host's stderr pipe against the expected `EPIPE` after its launcher exits, with a survival regression;
- documented why whole-text and token redaction are both required, with an embedded compact-JSON regression;
- omitted the stderr/truncation section for a blank trimmed tail;
- fixed the exact 4096-byte boundary so a complete tail is not reported as truncated.

The branch is rebased onto current `main` with no conflicts. Local verification is green: full build and typecheck, Biome on all 13 changed files, `git diff --check`, 4 targeted Runtime Host tests, and 36 targeted Desktop tests.

@Astro-Han, I have re-requested review for this exact head. Could a maintainer also approve the two workflows currently awaiting approval so the required `test` check can run?

Astro-Han (commented):
Thanks — you answered all four points from the previous round, and this comment exists mostly to confirm that on the record. Re-reviewed at exact head `583d62f7408805e0ac820249810ff58b92ef7454` against base `4acfa26934ce4b2b385b76f8a11048bb83fab861`. None of the six existing reviews are bound to this head, so everything below was re-checked against the current code rather than carried forward.

**Each of our four items, verified here:**

- **[P2] detached stderr pipe lifetime — fixed.** `launcher.ts` now spawns with `['ignore','ignore','pipe']` behind the `MAKA_RUNTIME_HOST_STDERR_PIPE` opt-in, and the EPIPE guard in `process-diagnostics.ts` swallows only `EPIPE` and rethrows everything else. That distinction matters and it is drawn correctly — a blanket catch here would have hidden the next real write failure. The detached-survival regression covers it.
- **[P2] `redactSecrets` two-pass semantics — resolved by explanation, which was the right resolution.** The comment now states what each pass is for: full-text pattern matching, then per-token structured redaction to catch compact JSON the first pass steps over. The ambiguity was in the reader's head, not the code, so documenting it was the proportionate fix.
- **[P3] orphaned truncation marker on empty stderr — fixed.** Empty stderr returns status-only before the marker is attached, with a whitespace-only regression alongside it.
- **[P3] off-by-one at exactly 4096 — fixed.** `chunk.length > MAX` rather than `>=`, with a regression pinning the exact boundary.

**Also checked, and sound:** `observeCandidateExit` listens on `'close'` rather than `'exit'`, so the stderr tail is complete before the record is built — using `'exit'` would have truncated exactly the evidence this PR exists to preserve. Diagnostics associate with the finally-registered PID, so a connect-or-spawn race attributes output to the process that actually won. The log buffer is bounded at 16KiB across four entries.

**[P3, ordinary path] The survival regression has a read race in the fixture, not in the product.**

`stderr-after-launcher-exit` writes its marker with `writeFile(marker, 'alive')`, which truncates before it writes, while `waitForFileText` polls the file every 20ms. A poll landing between truncate and write reads an empty string and the assertion fails on `'' !== 'alive'`. We reproduced this once in five isolated runs, so it will go red on CI at some rate.

The product behaviour is correct — the marker being written at all is what demonstrates the EPIPE guard works and the process survived. Writing to a temp file and renaming it into place would make the read atomic; retrying on empty would also do it.

**On CI:** this head has **zero check runs**. The description reports a passing suite, and we do not doubt it, but nothing has executed against `583d62f7` itself, so there is no independent verification to point at. A push to trigger CI is the remaining gap — it is also what would let this be approved rather than commented on.

Two reviewers went over this independently, and both reached the same conclusion on all four items. Our own run of the host-kernel suite was 48/51, with the three failures being Electron-dependent tests that cannot start on a headless machine (missing `libatk-1.0.so.0`) — an environment limit, not a result about this change.

---

packages/runtime-host/src/client/launcher.ts
```diff
@@ -118,14 +129,17 @@ function spawnCandidate(
   const child = spawn(executable, args, {
     cwd: dirname(isAbsolute(executable) ? executable : process.execPath),
     detached,
-    stdio: 'ignore',
+    stdio: ['ignore', 'ignore', 'pipe'],
```
Astro-Han, line 134: [P2] This is the line I'd check first against the failing `Release Windows check`.

Going from `stdio: 'ignore'` to `['ignore', 'ignore', 'pipe']` gives the detached child a stderr pipe whose read end lives in the parent. On Windows, once the parent goes away, writes to that pipe behave differently than to an ignored handle — and the failing step is precisely a detached spawn surviving an automatic update.

The EPIPE guard covers the write side inside our own process, so this may well be unrelated. But the check passes on neighbouring branches and fails here, and this is the one line in the PR that touches detached-process plumbing, so it seems the most likely place for the two facts to meet.

If it turns out to be unrelated, saying so is enough for me — I just don't want to approve past a red Windows check without either of us having looked.

Astro-Han (commented):
One follow-up now that the full check set has run on this head.

`test` is green, but **`Release Windows check` is failing** — `Verify automatic update end to end` times out with `Maka.exe did not appear among installed processes within 120000ms`.

I looked at whether that check is just flaky, and it doesn't appear to be: of the twelve most recent runs, it succeeded on eight branches around the same window and failed on two, one of which is this branch. So it isn't a broken runner.

I have **not** proven this PR causes it, and I'd rather say so than assert it. But the inline note below is where I'd start looking — the failing step is a detached-spawn scenario, and this is the PR that changes how that spawn's stdio is wired.

Everything from my previous review still stands: all four earlier points are resolved, and the only other open item is the P3 test race in the fixture.

---

packages/runtime-host/src/client/launcher.ts
```diff
@@ -150,15 +164,42 @@ function spawnedPid(child: ReturnType<typeof spawn>): Promise<{ pid: number }> {
...
+  child.stderr?.on('data', (value: Buffer | string) => {
+    const chunk = Buffer.isBuffer(value) ? value : Buffer.from(value);
+    if (chunk.length > CANDIDATE_STDERR_MAX_BYTES) {
+      stderr = chunk.subarray(chunk.length - CANDIDATE_STDERR_MAX_BYTES);
+      stderrTruncated = true;
+      return;
+    }
+    const combined = Buffer.concat([stderr, chunk]);
+    if (combined.length > CANDIDATE_STDERR_MAX_BYTES) {
+      stderr = combined.subarray(combined.length - CANDIDATE_STDERR_MAX_BYTES);
```
Astro-Han, line 190: **[P3] stderr is truncated *before* it is redacted, so a secret straddling the 4 KiB boundary can leak its tail.**

```ts
if (combined.length > CANDIDATE_STDERR_MAX_BYTES) {
  stderr = combined.subarray(combined.length - CANDIDATE_STDERR_MAX_BYTES);
}
```

The retained buffer is the **last** 4096 bytes. Redaction runs later, on that already-cut buffer:

```ts
const stderr = redactRuntimeHostStderr(exit.stderr);
```

`redactSecrets` recognizes a secret by its surrounding shape — `key=value`, `"apiKey":"…"`. When the cut lands in the middle of such a pair, the half that survives is the **value tail with its key prefix gone**, so there is no longer anything for the matcher to key on, and the fragment is copied into the diagnostic verbatim.

Reachability: a Host that dies during startup emitting more than 4 KiB of stderr is entirely ordinary; the secret landing exactly across the boundary is the low-probability part. That combination is why this is P3 and not higher. **The cost when it does happen is that a secret fragment ends up in a diagnostic artifact users are explicitly encouraged to copy and share.**

Fix, in preference order: redact the full stderr and truncate the redacted result, or — if buffering the whole stream is unacceptable — keep a small overlap window across the cut and re-scan it so a straddling pair is still matchable.

*This is a different concern from the existing note about redaction being applied twice. That one is about* how thoroughly *the redactor runs; this one is about* what it is handed.

---

**A second candidate, stated with its weakness up front rather than dressed up.** `stdio` for the candidate moves from `'ignore'` to `'pipe'`, with the EPIPE guard living behind `MAKA_RUNTIME_HOST_STDERR_PIPE=1` — i.e. it exists only in a Host that already has this change. A **new client spawning an older Host** would give that Host a piped stderr with no guard on the far side; after the client exits, the Host's next write to stderr raises EPIPE with no listener attached, which is an uncaught error in the Host. Launch does not check Host version, so the window is real but narrow, and Desktop normally spawns a matching Host. **Reachability is weak enough that I would not argue against dropping this one** — recorded so the next reviewer does not have to re-derive it.

---

**Two hypotheses raised and withdrawn**, so they need not be walked again:

- **Removing `child.once('error')` leaving spawn failures hung.** Measured: on `ENOENT` the `close` event still fires (`code=-2, signal=null`), the promise resolves, and `candidateStartupFailureForExitCode(-2)` returns `undefined` — behaviorally identical to the old `error → undefined`. Withdrawn.
- **`exited` promises leaking for non-winning attempts.** `spawnedProcesses` only retains the final registration's pid; the other attempts' listeners are released with the child, and the stderr stream has a `data` listener so its content is already collected before `close`. Withdrawn.

**On the change's direction:** what actually disappears in the `-18` is `stdio: 'ignore'` plus the `error` catch-all — stdout/stderr are no longer blindly discarded, and an untestable fallback is replaced with `close` + exit-code semantics that can be asserted. Bounded buffering (4 KiB per entry, 16 KiB across 4) plus a forensic format plus both sides of the pipe is the minimal closed loop for "the process died, leave a trail." **The truncation-before-redaction ordering is the one real seam in it.**

---

**Two CI facts for the record, no verdict attached:** this head is `CONFLICTING` against `main`, and while `test` is green, `package` is **failing** on the Windows packaging job.

*Independent line, bound to `583d62f7`. Verified at the gate 2026-08-23 13:02 UTC.*

Astro-Han:

- `apps/desktop/src/main/main-process-diagnostics.ts`
- `packages/runtime-host/src/__tests__/host-kernel.test.ts`
- `packages/runtime-host/src/client/connect-or-spawn.ts`
- `…(more)`

```
# resolve, then
git push --force-with-lease
```

---

HuYellow:
@Astro-Han Rebased onto current `main` (`729839ed8ada3e5498b0ad27e64fb42dfd283ae7`) and force-pushed exact head `2f7c10dbab4bab45856a5e5e57eaaad843843bfb`.

I manually resolved the five conflicts, preserving the new election diagnostics and keeping current-run Runtime Host exits out of previous-main-process reports. I also addressed the two follow-ups:

- A truncated stderr tail now drops its potentially partial leading record before redaction. The regression constructs a secret straddling the 4096-byte cut, confirms the retained raw tail still contains the secret fragment, and verifies that the formatted Desktop diagnostic does not.
- The survival fixture now writes a unique sibling temporary file and atomically renames it into place, removing the truncate/write polling race.

Local verification on Windows is green:

- `npm run build`
- `npm run typecheck`
- Biome on all 15 changed files
- `npm run check:stale`
- 13 targeted Runtime Host election/detached stderr tests
- 9 owned-candidate tests
- 41 Desktop candidate/diagnostics/recovery tests

The exact detached-process regression where the launcher exits and the Host later writes stderr passes on this head. GitHub currently shows the new head without source conflicts; `test` is pending and two workflows are awaiting maintainer approval. Please take another look at `2f7c10d` when CI is available.

packages/runtime-host/src/client/launcher.ts
HuYellow: Fixed on `2f7c10d`. The formatter now treats the leading record of a truncated raw tail as potentially partial and drops it before redaction. This keeps stderr buffering bounded while preventing a value fragment whose key fell before the cut from reaching copied diagnostics.

The new regression constructs a secret across the 4096-byte cut, confirms the retained raw tail still contains `secret-value`, then asserts the formatted Desktop diagnostic does not. The full Desktop candidate/diagnostics/recovery selection passes (41 tests).

packages/runtime-host/src/client/launcher.ts
HuYellow: Checked this on rebased head `2f7c10d` on Windows. The exact detached-process regression where the launcher exits and the Host later writes stderr passed, together with the full build/typecheck and all 13 targeted election/launcher tests. The fixture now publishes its marker by atomic rename, so the prior polling race cannot mask that survival result.

I have not reproduced the previous automatic-update timeout locally. The new-head Release Windows workflow is still awaiting maintainer approval, so I am leaving this thread open until that independent result is available.

HuYellow: Follow-up on exact head `2f7c10d`: the independent `Release Windows check / package` is now green (11m), alongside the required `CI / test` check (10m). This confirms the automatic-update detached-spawn scenario passes with the piped stderr change and the EPIPE guard on the rebased head. Resolving this thread.

HuYellow:
@Astro-Han All requested follow-ups are now complete on exact head `2f7c10dbab4bab45856a5e5e57eaaad843843bfb`: the branch is rebased onto current `main`, all review threads are resolved, and both GitHub checks are green (required `CI / test` in 10m and `Release Windows check / package` in 11m). The PR is now blocked only on the required approving review. Please complete the pending review when you have a chance.

apps/desktop/src/main/main-process-diagnostics.ts
```diff
@@ -127,6 +128,10 @@ export interface DesktopDiagnosticsDeps {
 export const mainProcessLogBuffer = new DiagnosticLogBuffer({
   maxBytes: MAIN_PROCESS_DIAGNOSTIC_LOG_MAX_BYTES,
 });
+export const runtimeHostProcessLogBuffer = new DiagnosticLogBuffer({
+  maxBytes: 16 * 1024,
```
jackwener, line 132: [P3] The promised four-exit history only retains three maximum-size records. This buffer combines `maxBytes: 16 * 1024` with `maxEntries: 4`, but each formatted 4096-byte stderr-tail diagnostic is larger than 4 KiB once the timestamp, level, exit status, truncation marker, and JSON string encoding are included. Reproduction on this production buffer: append four such records; `snapshot()` returns 3 entries (4206 encoded bytes each), retaining PIDs 2/3/4 and evicting PID 1. The current diagnostic test injects only one short record, so it cannot catch the contract gap. Please either size the total cap for four worst-case formatted entries or impose a per-entry bound that makes the stated four-entry retention true, and add a four-maximum-entry regression.

jackwener (approved):
Representative review under @WAWQAQ's standing apache/maka review authorization, bound to exact head `2f7c10dbab4bab45856a5e5e57eaaad843843bfb`.

GO: no P0–P2 findings. I verified the bounded stderr-tail collection, final-registration PID association, redaction/truncation path, detached-process behavior, and the dedicated copied-diagnostics path. The affected Runtime Host tests passed 5/5, Desktop diagnostics/candidate tests passed 41/41, and removing the selected-process handoff made its targeted regression fail before restoration. The exact-head `test` and Windows `package` check-runs are both completed/success; the PR is open, non-draft, and mergeable, with no current live hold.

One non-blocking P3 is recorded inline. This agent-executed approval does not claim to satisfy any separate repository human-review requirement. No merge performed.

PR #3348, opened 2026-08-20
fix(desktop): clear registered Runtime Host residue before update handoff

### LINKED ISSUE #3340: windows: auto-update silently fails when a Runtime Host process survives the app quit (installer cannot clear it)
## Summary

`quitAndInstall` hands off to the NSIS installer without draining the Runtime Host processes the app spawned. A `Maka.exe` host/candidate process running **from the install directory** survives the app's quit, the upgrade installer bails out on its app-running handling, and the update silently never applies — the user keeps the old version with no error surfaced. Observed twice on the `release-windows-check` autoupdate gate; the second occurrence carries full state evidence.

## Evidence (run [32382283646](https://github.com/apache/maka/actions/runs/32382283646), head 6b50c6308, step "Verify automatic update end to end")

The harness drove a real in-app update (check → download → `installUpdate` handoff at 15:05:13). The app process exited. The upgraded app never appeared; at the 90s mark the state was:

```
relaunched processes: [{"processId":4648,"name":"Maka.exe","path":"...\installed\Maka.exe"}]
process command lines: "...\installed\Maka.exe ...\app.asar\node_modules\@maka\runtime-host\dist\execution-candidate-main.js
  --root ...\user-data\workspaces\default --expected-root-id b7a77... --startup-attempt-id af63d71b-...
  --initial-connection-timeout-ms 37453 --generation 0.1.11"   CreationDate 15:04:53
backup directory ...\installed.pre-upgrade-backup: present
uninstall registrations: "0.1.11"
Maka.exe on disk: ProductVersion 0.1.11.0
```

Reading the state:
- The only `Maka.exe` alive is a **Runtime Host execution candidate** (`--generation 0.1.11`), created during the old app's startup and still running **106 seconds** after creation — long past its own `--initial-connection-timeout-ms 37453` — and well after its parent app quit for the update.
- The upgrade installer **did start**: the pre-upgrade backup hook (from #3265's installer include, present on this branch) ran and left its backup directory. But the uninstall registration, the on-disk executable, and the uninstaller are all untouched old-version state — so the installer exited on a hookless template path **before anything destructive**, i.e. in the app-running-check region between `.onInit` and `uninstallOldVersion`, with the surviving candidate as the process it could not clear (electron-builder's silent-mode `CheckAppRunning`/`MB_RETRYCANCEL /SD IDCANCEL` path ends in `Quit`).
- Net user-visible behavior: the update downloads, the app quits and relaunches nothing; on next launch the user is still on the old version, with no error anywhere.

A first occurrence with the same signature (relaunch wait finds a `Maka.exe`, disk still old) was run 32359725716, before the evidence capture existed.

## Likely relationship to #3279

An orphaned host/candidate from a previous app generation holding the workspace named-pipe endpoint is also a plausible producer of #3279's `host_unresponsive` startup failures ("stopped responding during startup" after an upgrade over an existing profile): the new app's election keeps probing an endpoint owned by a stale process. Worth checking both against the same lifecycle fix.

## Suggested direction (product decision, not prescribing)

Either `quitAndInstall` drains Runtime Host children (and any host it owns) before spawning the installer, or host/candidate lifetime is bounded by parent death on Windows (Phase 4 of #2142 already lists "cancellation, parent-death, concurrency, process-drain" evidence as open). The 37s candidate timeout also did not bound this process's lifetime, which may be its own defect.

Evidence grade: CI run log with per-process command lines + template source read (app-running check ordering); not locally reproduced.

liugddx:
## Summary

Fixes the update-drain gap behind #3340: a Runtime Host election winner the desktop manager never adopted — a late winner, or a survivor from a previous app generation — keeps running from the installation directory `quitAndInstall` is about to replace. Windows delivers no cross-process SIGTERM, electron-builder's NSIS upgrade quits when it cannot clear such a process, and the update silently never applies. CI run 32382283646 captured the orphan's full command line (`execution-candidate-main.js --generation 0.1.11`, alive 106s after creation, well past its parent's quit), untouched files and registration, and the retained pre-upgrade backup.

## Mechanism

The existing tracked drain is kept as-is: `installUpdate` → `prepareForUpdate` → `host.upgrade.prepare` wire verb → kernel drain → `waitForHostExit(pid)`. **After** it, the manager now reads the root-scoped host registration (`registration.json`, the authoritative live-host record election winners write) through a boot-injected capability closure (`resolveExistingStorageRootControlDirectory` + `readHostRegistration`, newly exported from `@maka/runtime-host/client`), and:

- an **explicitly ephemeral** registration naming a **live** pid that is **not** the one just drained is terminated outright and awaited — its executable image is about to be replaced and its committed state is crash-safe by the platform's recovery evidence, so termination is strictly better than a doomed installer handoff;
- a registration that is service-mode, **unlabeled** (predates `lifecycleMode`, could be a deployment-owned service host — destructive polarity requires the explicit label), dead, unreadable, or absent leaves the sweep inert, with the installer's own app-running handling as the remaining line;
- `ESRCH` on the kill is tolerated (the idle ephemeral residue can self-terminate between the liveness probe and the kill); `EPERM` stays fatal;
- a residue that survives termination rejects, surfacing `install_failed` instead of handing off to an installer that cannot succeed.

Normal-quit host survival (session persistence) is untouched: the sweep runs only on the update path.

## Stated boundaries and accepted residual risk

- **pid reuse (accepted, explicit):** a host that crashes leaves a stale registration; if Windows reuses that pid before the next election overwrites the file **and** the user updates in that window, the sweep would terminate an unrelated same-user process. The platform's identity-checking alternative (endpoint handshake + `expectedHostEpoch` probe before the kill) is noted as a follow-up hardening; the trigger requires a crashed registered host, pid reuse, and an update in the same window, and no lane can deterministically exercise it. Flagged by the local review gate and accepted here rather than silently.
- If the tracked local target never became ready, `prepareForUpdate` throws before the sweep (pre-existing behavior); a winner elected after the registration read is inherently uncovered (millisecond window, installer handling remains).
- The orphan-genesis question ("why did a candidate outlive its 37s connection timeout") resolved to by-design ephemeral-host lifetime (idle grace after an accepted connection), not a defect — recorded in #3340.

## Verification

- **L2 (local)**: `@maka/runtime-host` + `@maka/desktop` build and typecheck clean; biome clean; `runtime-host-desktop-manager` suite 19 pass / 0 fail, including new pins: kill-before-await ordering for the orphan, inert sweep for service/unlabeled/dead/unregistered registrations (each case constructed so removing its specific guard turns it red), unkillable residue → rejection, unreadable registration → degrade to installer handling. An independent review gate ran before push; its three P2s (ESRCH tolerance, destructive polarity on unlabeled registrations, red-signal coverage for the service guard) are fixed in this commit, and its pid-reuse finding is the accepted risk above.
- **L3 (CI)**: the Windows `release-windows-check` autoupdate gate exercises the real handoff; the orphan race it removes was hitting roughly one in three recent runs.

Closes #3340. Relates to #3279 (a stale endpoint owner is a plausible shared root) and #2142 Phase 4 (process-drain evidence).


jackwener (commented):
Automated Codex review — blocking finding on exact head `27adb845a288a080a49a156ad7898ee321991f3b`.

**P1: prove process identity before terminating the registered PID.** In `RuntimeHostDesktopManagerImpl.#drainRegisteredHostResidue`, the registration may be stale after a Host crash. The current guards prove only that the old registration said `ephemeral` and that *some* process now occupies its PID; `process.kill(pid)` then terminates that process. If Windows has reused the PID before the update, this kills an unrelated same-user process. The PR description acknowledges exactly this execution path, so it is not merely theoretical.

The smallest sound fix is to authenticate the live endpoint immediately before termination using the registration's expected `hostEpoch` and root identity (the existing connection handshake already validates those values), and fail safe when identity cannot be proved. Add a regression case where a stale registration names a live reused PID and assert that `killHost` is not called. A liveness probe alone cannot close this gap.

Required conclusions:

1. Optimal for the actual problem: **no**; the post-drain sweep addresses the orphan, but its destructive identity boundary is incomplete.
3. Tests to delete/replace: none; add the stale-registration/PID-reuse case above.
4. Deeper refactor: no; reuse the existing authenticated Host identity check at this boundary.
5. Ready to merge: **no**.
6. Residual risk/gaps: unrelated-process termination during update; independent human review is also still required.

This affects process safety and update behavior, so it is a protected-area change requiring independent human review under `CONTRIBUTING.md` after the blocker is fixed.

liugddx:
@jackwener Addressed at head `3bae05d78` with a stricter fail-closed boundary.

After implementing the suggested authenticated handshake, I re-reviewed the destructive boundary and found that `handshake -> close connection -> process.kill(pid)` still leaves a PID-reuse TOCTOU window: the authenticated Host can exit after the handshake and before the PID kill. Closing that correctly would require a process-handle-bound identity/termination primitive or a protocol self-shutdown operation, both outside this PR's narrow scope.

So this revision removes the untracked registration/PID residue sweep entirely. Update preparation now drains only the Host authenticated by the manager's existing connection and waits for that tracked PID to exit; any untracked process residue remains fail-closed at the installer's app-running gate. The sweep-only boot wiring, injectable kill/liveness surface, and its tests were removed as well (net `2 insertions, 200 deletions` from the reviewed head).

Local verification:
- `@maka/runtime-host` build
- `@maka/desktop` main build
- runtime-host desktop manager tests: 14/14
- Biome formatting/check and `git diff --check`

This avoids terminating any process from stale registration evidence without introducing the broader process-handle/protocol architecture deferred from this slice. Independent human security review is still required before merge.

liugddx:
Rebased in substance by merging current `main` at `126d9941b`, which includes #3327 via `3dcaa8893`. I also removed the now-unused `readHostRegistration` client export and boot import churn left behind after the untracked PID sweep was retired.

The net diff against current main is now intentionally one two-line boundary comment in `prepareForUpdate`: only the Host authenticated by the manager's existing connection is drained; untracked process residue remains the installer's responsibility. There is no registration-based PID termination or supporting API left in this PR.

Post-merge local verification:
- `@maka/core`, `@maka/runtime`, `@maka/runtime-host`, and desktop main builds passed
- runtime-host desktop manager tests: 19/19
- Biome and `git diff --check` passed

Given the now comment-only net diff, this PR can either merge as the explicit boundary documentation or be closed as superseded/retired; no functional fix remains to carry.

liugddx:
Closing as intentionally retired after review.

The registration-based residue sweep cannot safely terminate by PID alone: even an authenticated handshake immediately before `process.kill(pid)` does not atomically bind the later kill to that Host, because the Host may exit and Windows may reuse the PID in between. Closing that boundary correctly requires either a stable Windows process handle or an authenticated Host self-shutdown protocol, both outside this PR's narrow scope.

The unsafe sweep and its supporting API were therefore removed. Current `main` already retains the safe part: drain the Host authenticated by the manager's existing connection and wait for that tracked process to exit. The Windows verifier reliability work that exposed/handled the surrounding CI failures was merged separately in #3327.

After merging current main, this PR's net diff is only a two-line explanatory comment, which does not justify a standalone PR. A future implementation should start from the stable-identity primitive rather than reviving registration-based PID termination.

PR #3350, opened 2026-08-20
fix(ui): show one spinner throughout live turns

Colafornia:
## Summary

During a live turn, a completed tool row could appear above the working status while no animated indicator showed that output was still in progress.

The expected behavior is one active spinner—not zero or two—throughout a live turn.

| Live-turn phase | Tool row | Turn status | Spinner owner |
|---|---|---|---|
| The model works before or between tool calls | No running tool | Shows the working phrase, elapsed time, and spinner | Turn status |
| A tool call runs | Shows its running state and spinner | Keeps the working phrase and elapsed time | Tool row |
| The tool settles and the model continues | Shows its completed state | Resumes the spinner with the working phrase and elapsed time | Turn status |

When the turn reaches a terminal state, the running status disappears and no spinner remains.

The previous implementation removed the spinner from `TurnRunningStatus` to prevent duplicate animation during tool calls. This also removed the only animated indicator between tool calls.

This change detects in-flight tools with the shared `isInFlightToolStatus` helper. The turn status renders the existing Astryx `Spinner` only when no tool owns the active spinner.

The working copy, elapsed timer, timeline folding, and tool-row behavior remain unchanged. The spinner is decorative for assistive technology, while the status row keeps its accessible name.

| Before | After | 
|---|---|
|image|image|

## Verification

- `npm --workspace @maka/ui test` — 184 tests passed
- The new regression test failed before the production change and passed after it
- `git diff --cached --check` — passed
- Real-window visual verification was not run

## AI use

Select exactly one:

Tool(s) and scope: Codex diagnosed the missing activity state, implemented conditional spinner ownership, added regression coverage, and drafted this pull request.

## Checklist

- [ ] No

### DIFF
--- a/packages/ui/src/chat-turn.tsx
+++ b/packages/ui/src/chat-turn.tsx
@@ -17,6 +17,7 @@ import {
   ChatTokenizedText,
   HStack,
   IconButton as UiIconButton,
+  Spinner,
   Thumbnail,
   Timestamp,
   Token,
@@ -41,7 +42,7 @@ import { foldTimeline, type FoldedTimelineChild, type FoldedTimelineEntry } from
 import { AttachmentKindIcon } from './attachment-kinds.js';
 import { QuoteRefChip } from './quote-ref-chip.js';
 import { Marker, markerVariants } from './primitives/chat.js';
-import { ToolTrow } from './tool-activity.js';
+import { ToolTrow, toolTrowHasVisibleSpinner } from './tool-activity.js';
 import { formatBytes } from './tool-activity/preview-utils.js';
 import { useUiLocale } from './locale-context.js';
 import { getConversationCopy } from './conversation-copy.js';
@@ -400,6 +401,9 @@ export const TurnView = memo(function TurnView(props: {
     () => splitTimelineAtUserMessages(foldedTimeline, showAssistantMessage),
     [foldedTimeline, showAssistantMessage],
   );
+  const toolSurfaceOwnsSpinner = turn.timeline.some(
+    (item) => item.kind === 'tools' && toolTrowHasVisibleSpinner(item.items),
+  );
   return (
     <section
       className="maka-turn"
@@ -620,7 +624,10 @@ export const TurnView = memo(function TurnView(props: {
                     <ModelProviderRetryIndicator retry={props.liveStreaming.providerRetry} />
                   ) : (
                     props.liveStreaming.runningStatus && (
-                      <TurnRunningStatus startedAt={turn.startedAt} />
+                      <TurnRunningStatus
+                        startedAt={turn.startedAt}
+                        showSpinner={!toolSurfaceOwnsSpinner}
+                      />
                     )
                   )}
                 </>
@@ -901,7 +908,7 @@ const ELAPSED_TICK_MS = 1_000;
  * rare fallback path where streaming beat the user turn into the transcript;
  * the phrase then stands alone.
  */
-export function TurnRunningStatus(props: { startedAt?: number }) {
+export function TurnRunningStatus(props: { startedAt?: number; showSpinner?: boolean }) {
   const copy = getConversationCopy(useUiLocale()).messages;
   const phrases = copy.workingPhrases;
   const { startedAt } = props;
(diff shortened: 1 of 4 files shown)


jackwener (commented):
Automated Codex review on exact head `e0c7eeadaefb5f2a243a0381663041b33ff3d2c2`.

The product fix is first-principles and appropriately small: current `main` removed the turn-status spinner to avoid duplication during an active tool, which also left the before/between-tool phases with no spinner. Computing ownership from the existing `isInFlightToolStatus` contract restores exactly one spinner, and the focused test protects the running-tool → status-off / completed-tool → status-on transition.

There is one deterministic merge blocker: the required `test` check fails because adding Astryx `Spinner` to `packages/ui/src/chat-turn.tsx` changes the generated surface inventory, but `docs/astryx-surface-file-inventory.md` was not refreshed. This is not fixed merely by rebasing: run `npm run astryx:surface-inventory:write`, commit the generated inventory update, and let required CI rerun.

Required conclusions:

1. Optimal for the actual problem: **yes**.
4. Deeper refactor: no; the existing tool-status source of truth is the right boundary.
5. Ready to merge: **no**; required `test` is red and independent human review is missing.
6. Residual risk/gaps: refresh the generated inventory; the UI behavior is material enough that the supplied before/after evidence and independent human judgment remain the acceptance boundary.

No security, licensing, release, governance, or public-contract effect was identified; this is a user-visible UI behavior change and therefore still requires independent human review under `CONTRIBUTING.md`.

Colafornia:
@jackwener Thanks for the review. I’ve refreshed and committed the Astryx surface inventory. The inventory check and all 184 UI tests now pass locally.

packages/ui/src/chat-turn.tsx
```diff
@@ -400,6 +402,9 @@ export const TurnView = memo(function TurnView(props: {
     () => splitTimelineAtUserMessages(foldedTimeline, showAssistantMessage),
     [foldedTimeline, showAssistantMessage],
   );
+  const hasInFlightTool = turn.timeline.some(
```
hqhq1025: [P2] Keep the turn spinner when the collapsed tool surface does not expose the in-flight call. This predicate suppresses the status spinner when **any** raw tool is `pending` or `running`, but Astryx `ChatToolCalls` collapses multi-call groups and surfaces only the last call. A normal parallel batch can therefore be `[running, completed]`: the earlier call's spinner exists only inside the collapsed content, the visible summary shows the later completed icon, and this line also removes the turn spinner. Rendering that exact state produces zero visible spinners, returning the user to the no-animation state this PR is meant to fix while work is still active. Please base ownership on whether the actual visible tool surface owns a spinner (including the collapsed-group aggregate), and add a mixed parallel-status regression such as `[running, completed]`.

hqhq1025 (commented):
The current revision was reviewed completely and the focused local verification passed, but the current CI run concluded `action_required` with no jobs, so the required `test` check has not passed and this revision cannot be approved yet.

packages/ui/src/chat-turn.tsx
Colafornia: Fixed in 135b2ce84. Spinner ownership now follows the last visible standard call in each ChatToolCalls segment, so a collapsed mixed group such as [running, completed] leaves the spinner on the turn status. Added the mixed-status regression; all 185 @maka/ui tests and the Astryx surface inventory check pass locally.

hqhq1025 (commented):
Re-review of the current head found no remaining actionable defect; the prior collapsed parallel-tool spinner finding is resolved. Exact-head root build, 185 UI tests, repository lint and format checks, Astryx inventory validation, and the current-main merge tree all passed. GitHub Actions run 32439079216 concluded `action_required` with zero jobs, so CI has not executed on this revision and I cannot approve it yet.

Astro-Han (approved):
Thanks @Colafornia — reviewed at exact head `a74a272f6b0f37db82b5d5c1c57101a0be7a361e`. No findings; approving.

**The single-spinner invariant holds**

`toolSurfaceOwnsSpinner` is computed once from the turn timeline and passed straight through as `showSpinner={!toolSurfaceOwnsSpinner}`, so the tool surface and `TurnRunningStatus` are strictly mutually exclusive — a turn cannot render two spinners, and the "who owns it" question has exactly one answer.

**The earlier collapsed-group gap is closed**

The `some(isInFlight)` formulation would have handed ownership to the tool surface for a `[running, completed]` group whose running entry is hidden by folding, leaving zero visible spinners. `toolTrowHasVisibleSpinner` now requires the in-flight item to be the last entry in its segment (or to be followed by a linked-agent row), so `[running, completed]` correctly falls back to the turn-level spinner. The regression test `keeps the turn spinner when a collapsed group hides the running tool` pins that case.

**Terminal states**

The block is not rendered at all unless the turn is `running` with a truthy `liveStreaming.runningStatus`, so a failed or cancelled turn cannot leave a spinner stranded on screen — the case most visible to users if it were wrong.

Linked-agent sub-sessions signal progress with a pulsing `StatusDot` rather than a `Spinner`, so ownership does not blur across that boundary either.

**Generated inventory**

Re-ran `generate-astryx-surface-inventory.mjs` at this head: working tree clean, and the new `Spinner` usage in `chat-turn.tsx` is reflected in the committed `.md` — no hand-edited drift.

**One architectural note, not a finding**

`toolSurfaceOwnsSpinner` reads `turn.timeline` while the rendered rows come from the folded timeline. Today the two agree, because "last item in the segment" is precisely the position folding preserves. It is worth being aware that the ownership predicate and the visibility it stands for are derived from different sources, so a future change to folding rules would need to keep them in step. Nothing to change here.

---

PR #3359, opened 2026-08-20
refactor(desktop): share the dev userData profile across worktrees

Astro-Han:
## Summary

Use one `Maka Dev` AppProfile for plain and TCC development across worktrees. Released Maka keeps its separate profile, and the Runtime Host state root remains independent from the AppProfile.

Electron's `requestSingleInstanceLock()` is the sole authority for which development client may use the shared profile. Plain Electron consumes its child exit code; the LaunchServices-detached TCC launcher consumes a private one-shot winner/loser verdict. No command-line process scan participates in the decision.

The TCC app and its launcher now have deliberately separate lifecycles. The launcher owns only its Vite server, log follower, and per-launch files. Ctrl-C stops those resources but does not quit the TCC Electron app; the user quits that app with Cmd-Q. Accordingly, the verdict carries no PID, and the launcher performs no PID liveness polling or TERM/KILL teardown.

The launch attempt is one cancellable ownership boundary. A tested launch session owns signal registration, aborts the attempt, and joins its eventual result; all asynchronous preparation finishes before launch files or processes are created, so cancellation cannot be followed by a late Electron spawn or leaked launch files.

The TCC bundle identity remains worktree-scoped because macOS permissions are tied to that signed bundle identity. Existing per-worktree profile data is not deleted; the launcher warns when it finds the legacy location.

Refs #3539

## Verification

- `npm --workspace @maka/core test` — 639/639 passed
- `npm --workspace @maka/desktop test` — 1265/1265 passed
- `node --test apps/desktop/scripts/dev-app-runtime.test.mjs` — 20/20 passed
- `npm --workspace @maka/desktop run typecheck` — passed

## Migration

Previous TCC data remains under `~/Library/Application Support/Maka Dev-<WORKTREE_ID>`. The app now reads `~/Library/Application Support/Maka Dev` unless `--user-data-dir` is supplied. No data is copied or removed automatically.

## AI use

Select exactly one:

Tool(s) and scope: Codex contributed architecture analysis, implementation, tests, documentation, and review remediation. The commit retains a `Generated-by: Codex` trailer.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/scripts/dev-app-runtime.mjs
+++ b/apps/desktop/scripts/dev-app-runtime.mjs
@@ -30,10 +30,10 @@
  * a small bootstrap injected. Everything the bootstrap needs is a build-time
  * constant, so a launch with no arguments and no environment — the Dock,
  * Spotlight, or the system's Screen Recording "Quit & Reopen" — reproduces a
- * correct app. That is why there is no session protocol, pid file, or
- * supervisor here: the app instance and the dev session are separate
- * lifecycles, and nothing about the app's correctness depends on who started
- * it.
+ * correct app. The launcher's one-shot result file reports only the Electron
+ * lock verdict; it does not supervise the app or carry application state. The
+ * app instance and the dev session remain separate lifecycles: stopping the
+ * launcher never terminates the TCC app.
  *
  * `app.isPackaged` is native and computed as
  * `basename(process.execPath) !== 'electron'`, so the invariant that keeps
@@ -46,9 +46,16 @@
  * cannot shadow a real packaged app laid down at the standard location.
  */
 import { spawn, spawnSync } from 'node:child_process';
-import { createHash } from 'node:crypto';
-import { existsSync, mkdirSync, readFileSync, renameSync, rmSync, writeFileSync } from 'node:fs';
-import { homedir } from 'node:os';
+import { createHash, randomUUID } from 'node:crypto';
+import {
+  existsSync,
+  mkdirSync,
+  readFileSync,
+  renameSync,
+  rmSync,
+  writeFileSync,
+} from 'node:fs';
+import { homedir, tmpdir } from 'node:os';
 import { dirname, join, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
 
@@ -64,13 +71,10 @@ const MARKER = join(DEV_RUNTIME_DIR, 'runtime.json');
 const ELECTRON_PACKAGE = join(REPO_ROOT, 'node_modules', 'electron', 'package.json');
 const SOURCE_APP = join(REPO_ROOT, 'node_modules', 'electron', 'dist', 'Electron.app');
 const WORKTREE_ID = createHash('sha256').update(REPO_ROOT).digest('hex').slice(0, 12);
-const DEV_USER_DATA_DIR = join(
-  homedir(),
-  'Library',
-  'Application Support',
-  `Maka Dev-${WORKTREE_ID}`,
-);
-
+// Deliberately NOT worktree-scoped, unlike the bundle identifier above: plain
+// Electron and the TCC bundle must share the same single-instance authority.
+export const DEV_USER_DATA_DIR = join(homedir(), 'Library', 'Application Support', 'Maka Dev');
+const DEV_ENV_SCHEMA_VERSION = 1;
 /**
  * Per-worktree, and deliberately so. An ad-hoc signature designates a bare
  * `cdhash`, and TCC keys its rows on the bundle identifier — so a shared
@@ -90,17 +94,40 @@ const DEV_USER_DATA_DIR = join(
  */
 const DEV_BUNDLE_ID = `com.maka.dev.${WORKTREE_ID}`;
 const RUNTIME_SCHEMA_VERSION = 7;
-const DEV_ENV_SCHEMA_VERSION = 1;
 
 export const developmentAppPath = DEV_APP;
-const developmentExecutablePath = DEV_EXECUTABLE;
 
-export async function resolveMacosDevelopmentLaunch(env = process.env) {
+async function prepareMacosDevelopmentLaunch(env = process.env) {
   if (!shouldUseMacosDevelopmentApp(process.platform, env)) return null;
   const appPath = await prepareDevelopmentApp();
-  // A leftover app would absorb this launch through the single-instance lock.
-  await ensureNoRunningDevelopmentApp();
-  return createMacosDevelopmentLaunch(appPath, developmentLogFile);
+  const resultFileArgPrefix = (await devSingleInstanceConstants())
+    ?.DEV_LAUNCH_RESULT_FILE_ARG_PREFIX;
+  if (!resultFileArgPrefix) {
+    throw new Error('Maka Dev launch-result contract is unavailable; rebuild @maka/core and retry.');
+  }
+  return { appPath, resultFileArgPrefix };
+}
+
+/**
+ * The TCC bundle previously wrote per-worktree userData roots
+ * (`Maka Dev-<WORKTREE_ID>`); sharing the profile means this launch reads the
+ * common `Maka Dev` root, so sessions and settings that lived in the legacy
+ * directory are no longer read. They are NOT deleted — say where they are.
+ */
+export function warnAboutLegacyTccDataRoot(options = {}) {
+  if ((options.platform ?? process.platform) !== 'darwin') return;
+  const effectiveUserDataDir = options.effectiveUserDataDir ?? DEV_USER_DATA_DIR;
+  const legacy = options.legacyUserDataDir ??
+    join(homedir(), 'Library', 'Application Support', `Maka Dev-${WORKTREE_ID}`);
+  const exists = options.exists ?? existsSync;
+  const warn = options.warn ?? console.warn;
+  if (effectiveUserDataDir !== DEV_USER_DATA_DIR || legacy === DEV_USER_DATA_DIR || !exists(legacy)) {
+    return;
+  }
+  warn(
+    `[maka-dev] shared profile: your earlier TCC data (sessions, settings) lives in ${legacy} ` +
+      'and is no longer read. It is not deleted — copy it back if needed, or remove it.',
+  );
 }
 
 /**
@@ -116,17 +143,38 @@ export function shouldUseMacosDevelopmentApp(platform, env = process.env) {
   return optIn === '1' || optIn === 'true';
 }
 
-export function createMacosDevelopmentLaunch(appPath, logFile) {
+export function createMacosDevelopmentLaunch(
+  appPath,
+  logFile,
+  resultFile,
+  resultFileArgPrefix,
+) {
   // LaunchServices must own the launch so macOS TCC attributes the running
   // executable to Maka Dev rather than to its parent terminal.
   const args = ['-n', '-a', appPath];
   // Without this the detached app's output only reaches Console.app, which is
   // also where a bootstrap or boot failure would go silent.
   if (logFile) args.push('--stdout', logFile, '--stderr', logFile);
+  // The private result path must come LAST: everything after `open --args` is
+  // passed to the app, and the last launcher-owned value wins over any
+  // user-supplied lookalike argument.
+  if (resultFile && resultFileArgPrefix) {
+    args.push('--args', `${resultFileArgPrefix}${resultFile}`);
+  }
   return { command: 'open', args };
 }
 
-export const developmentLogFile = join(DEV_RUNTIME_DIR, 'app.log');
+export function createDevelopmentLaunchFiles(options = {}) {
+  const directory = options.directory ?? join(tmpdir(), 'maka-dev-launches');
+  const id = options.id ?? randomUUID();
+  const logFile = join(directory, `${id}.log`);
+  const resultFile = join(directory, `${id}.result.json`);
+  const mkdir = options.mkdir ?? mkdirSync;
+  const write = options.write ?? ((target) => writeFileSync(target, '', { mode: 0o600 }));
+  mkdir(directory, { recursive: true, mode: 0o700 });
+  write(logFile);
+  return { logFile, resultFile };
+}
 
 /**
  * `pkill -f` / `pgrep -f` match an EXTENDED REGEX against the whole command
@@ -139,41 +187,48 @@ export function toProcessMatchPattern(executable) {
 }
 
 /**
- * Terminates this worktree's development app. The bundle path is unique per
- * worktree, so matching on it is precise without tracking a pid: concurrent
- * worktrees own different bundles and are unaffected.
+ * Electron's lock is authoritative. The plain launcher owns its child directly
+ * and reads its exit code, while the detached TCC launcher consumes a private,
+ * one-shot winner/loser result without taking ownership of the app process.
+ * Constants are imported lazily so this module loads before libs are built.
  */
-export async function quitMacosDevelopmentApp(options = {}) {
-  const platform = options.platform ?? process.platform;
-  const executable = options.executable ?? DEV_EXECUTABLE;
-  const graceMs = options.graceMs ?? 3_000;
-  const signal = options.signal ?? sendSignalToExecutable;
-  const delay = options.delay ?? ((ms) => new Promise((done) => setTimeout(done, ms)));
-  if (platform !== 'darwin') return false;
-  if (!signal('TERM', executable)) return false;
-  // Main-process cleanup runs on before-quit and can outlive a plain SIGTERM.
-  await delay(graceMs);
-  signal('KILL', executable);
-  return true;
+export async function devSingleInstanceConstants(loader = () => import('@maka/core/dev-single-instance')) {
+  // Explicit failure contract: the import can fail on a fresh clone before
+  // libs are built. Every consumer must handle undefined explicitly — a
+  // throw here would crash the plain launcher. One warning, then undefined.
+  try {
+    return await loader();
+  } catch (error) {
+    console.warn(`[maka-dev] dev single-instance constants unavailable: ${String(error)}`);
+    return undefined;
+  }
 }
 
-function sendSignalToExecutable(name, executable) {
-  const status = spawnSync('pkill', [`-${name}`, '-f', toProcessMatchPattern(executable)]).status;
-  // 0 = signalled, 1 = no match. Anything else is a usage or pattern error and
-  // must not be read as "nothing was running".
-  if (status !== 0 && status !== 1) {
-    throw new Error(`pkill -${name} failed for ${executable} (exit ${status})`);
+export async function readDevelopmentLaunchResult(resultFile, loader) {
+  const constants = await devSingleInstanceConstants(loader);
+  if (!constants) return undefined;
+  try {
+    return constants.parseDevelopmentLaunchResult(readFileSync(resultFile, 'utf8'));
+  } catch {
+    return undefined;
   }
-  return status === 0;
+}
+
+/** Exit-code check for the plain child (child is the app process). */
+export async function plainLoserExitCode(code, loader) {
+  const constants = await devSingleInstanceConstants(loader);
+  if (!constants) return false; // cannot attribute; do not invent a loser
+  return code === constants.DEV_LOSER_EXIT_CODE;
 }
 
 export function isDevelopmentAppRunning(options = {}) {
-  const executable = options.executable ?? DEV_EXECUTABLE;
+  const bundle = options.executable ?? DEV_EXECUTABLE;
   const probe = options.probe ?? defaultLivenessProbe;
-  return probe(executable);
+  return probe(bundle);
 }
 
 function defaultLivenessProbe(executable) {
+  if (process.platform === 'win32') return false;
   const status = spawnSync('pgrep', ['-f', toProcessMatchPattern(executable)]).status;
   if (status !== 0 && status !== 1) {
     throw new Error(`pgrep failed for ${executable} (exit ${status})`);
@@ -181,38 +236,10 @@ function defaultLivenessProbe(executable) {
   return status === 0;
 }
 
-/**
- * Takes ownership of this worktree's app instance before launching or
- * rebuilding. Electron's single-instance lock is keyed on userData, so an app
- * left over from a hard-killed session would absorb the new launch: the new
- * process exits 0, the OLD window is raised, and it stays pointed at a dead
- * Vite URL while a liveness probe still reports success.
- */
-export async function ensureNoRunningDevelopmentApp(options = {}) {
-  const running = options.isRunning ?? isDevelopmentAppRunning;
-  const quit = options.quit ?? quitMacosDevelopmentApp;
-  const delay = options.delay ?? ((ms) => new Promise((done) => setTimeout(done, ms)));
-  const attempts = options.attempts ?? 10;
-  const pollMs = options.pollMs ?? 200;
-  // Forward each callee only what it accepts. Passing this whole object through
-  // would make `delay` double as the SIGTERM grace period, and would let a
-  // caller that stubs `probe` still fall through to a real `pkill`.
-  const liveness = { executable: options.executable, probe: options.probe };
-  const shutdown = {
-    platform: options.platform,
-    executable: options.executable,
-    graceMs: options.graceMs,
-    signal: options.signal,
-  };
-  if (!running(liveness)) return false;
-  await quit(shutdown);
-  for (let attempt = 0; attempt < attempts; attempt += 1) {
-    if (!running(liveness)) return true;
-    await delay(pollMs);
-  }
-  throw new Error(
-    'A previous Maka Dev.app is still running and could not be stopped; quit it manually (Cmd-Q) and retry',
-  );
+export function assertDevelopmentAppNotRunning(options = {}) {
+  const isRunning = options.isRunning ?? isDevelopmentAppRunning;
+  if (!isRunning(options)) return;
+  throw new Error('Maka Dev.app is running. Quit it (Cmd-Q) and retry the rebuild.');
 }
 
 /**
@@ -301,8 +328,8 @@ export function writeDevelopmentEnvironment(content, options = {}) {
 /**
  * The Vite URL a previous launcher published, if any.
  *
- * `npm start` does not run a dev server, but it may be reclaiming an app from a
- * live `npm run dev`. Republishing without the URL would drop the app to the
+ * `npm start` does not run a dev server, but it may reuse the URL published by
+ * the previous `npm run dev`. Republishing without the URL would drop the app to the
  * prebuilt renderer on disk — stale, or absent entirely on a fresh checkout.
  */
 export function readPublishedViteUrl(file = DEV_ENV_FILE) {
@@ -336,33 +363,65 @@ function resolveElectronBinary() {
  * Returns a handle rather than a child process because the two paths are not
  * comparable: on the macOS bundle path `open` exits at the LaunchServices
  * handoff, so its exit code says nothing about the app.
+ *
+ * All asynchronous preparation finishes before launch artifacts or processes
+ * are created. Once `signal` is aborted, this attempt cannot publish either.
  */
 export async function startDevelopmentApp(options = {}) {
+  const prepareMacosLaunch = options.prepareMacosDevelopmentLaunch ?? prepareMacosDevelopmentLaunch;
+  const loadSingleInstanceConstants = options.devSingleInstanceConstants ?? devSingleInstanceConstants;
+  const createLaunchFiles = options.createDevelopmentLaunchFiles ?? createDevelopmentLaunchFiles;
+  const spawnProcess = options.spawn ?? spawn;
+  const signal = options.signal;
+  signal?.throwIfAborted();
   const argv = options.argv ?? [];
+  const { userDataDir } = splitDevelopmentCliArgs(argv);
+  warnAboutLegacyTccDataRoot({ effectiveUserDataDir: userDataDir ?? DEV_USER_DATA_DIR });
   // Read before preparing: a rebuild republishes the runtime directory and
   // takes the previous environment file with it.
   const viteUrl = options.viteUrl ?? readPublishedViteUrl();
-  const launch = await resolveMacosDevelopmentLaunch();
-  if (!launch) {
-    const child = spawn(resolveElectronBinary(), [DESKTOP_DIR, ...argv], {
+  const preparedLaunch = await prepareMacosLaunch();
+  signal?.throwIfAborted();
+  if (!preparedLaunch) {
+    const launcherFlag = (await loadSingleInstanceConstants())?.DEV_CONFLICT_HANDLED_BY_LAUNCHER_FLAG;
+    signal?.throwIfAborted();
+    const child = spawnProcess(resolveElectronBinary(), [DESKTOP_DIR, ...argv, ...(launcherFlag ? [launcherFlag] : [])], {
       cwd: DESKTOP_DIR,
       stdio: 'inherit',
       env: viteUrl ? { ...process.env, VITE_DEV_SERVER_URL: viteUrl } : process.env,
     });
+    child.once('exit', (code) => {
+      void plainLoserExitCode(code).then((loser) => {
+        if (!loser) return;
+        console.error(
+          'Maka Dev could not start: another instance holds the shared profile lock ' +
+            `(loser exit code ${code}). Quit it (Cmd-Q) and retry.`,
+        );
+      });
+    });
     return { child, isMacosBundle: false, stop: () => terminateProcessTree(child) };
   }
 
+  const files = createLaunchFiles();
+  const launch = {
+    ...createMacosDevelopmentLaunch(
+      preparedLaunch.appPath,
+      files.logFile,
+      files.resultFile,
+      preparedLaunch.resultFileArgPrefix,
+    ),
+    ...files,
+  };
   writeDevelopmentEnvironment(
     createDevelopmentEnvironmentFile({ argv, env: process.env, viteUrl }),
   );
   // LaunchServices detaches the app from this terminal's stdio, so `open`
-  // redirects its output here and we follow it. The log carries whatever the
-  // app prints, so it gets the same mode as the environment file beside it.
-  writeFileSync(developmentLogFile, '', { mode: 0o600 });
-  const logStream = spawn('tail', ['-n', '+1', '-F', developmentLogFile], {
+  // redirects its output here and we follow it. Logs are observation only;
+  // the lock verdict travels through launch.resultFile instead.
+  const logStream = spawnProcess('tail', ['-n', '+1', '-F', launch.logFile], {
     stdio: ['ignore', 'inherit', 'inherit'],
   });
-  const child = spawn(launch.command, launch.args, {
+  const child = spawnProcess(launch.command, launch.args, {
     cwd: DESKTOP_DIR,
     stdio: 'inherit',
     env: process.env,
@@ -370,11 +429,71 @@ export async function startDevelopmentApp(options = {}) {
   return {
     child,
     isMacosBundle: true,
-    stop: async () => {
-      logStream.kill();
-      await quitMacosDevelopmentApp();
-    },
+    logFile: launch.logFile,
+    resultFile: launch.resultFile,
+    stop: () => cleanupDevelopmentLaunch({
+      logStream,
+      logFile: launch.logFile,
+      resultFile: launch.resultFile,
+    }),
   };
+}
+
+/**
+ * Owns the dev launch from process signal through final resource cleanup.
+ * Keeping this coupling here makes the cancellation wire itself testable: a
+ * signal aborts the same attempt `start()` created, then joins and stops any
+ * handle that attempt was already publishing.
+ */
+export function createDevelopmentLaunchSession(options = {}) {
+  const startApp = options.startDevelopmentApp ?? startDevelopmentApp;
+  const signals = options.signals ?? process;
+  const close = options.close ?? (() => {});
+  const exit = options.exit ?? ((code) => process.exit(code));
+  const controller = new AbortController();
+  let launchPromise = null;
+  let stopPromise = null;
+
+  const stop = (code = 0) => {
+    if (stopPromise) return stopPromise;
+    controller.abort();
+    stopPromise = (async () => {
+      const app = await launchPromise?.catch(() => null);
+      await app?.stop();
+      try {
+        await close();
+      } catch {
+        // Shutdown is best-effort; the process must still reach its exit path.
+      }
+      exit(code);
+    })();
+    return stopPromise;
+  };
+
+  signals.on('SIGINT', () => stop(0));
+  signals.on('SIGTERM', () => stop(0));
+  signals.on('SIGHUP', () => stop(0));
+
+  return {
+    start(startOptions = {}) {
+      launchPromise = Promise.resolve(startApp({
+        ...startOptions,
+        signal: controller.signal,
+      }));
+      return launchPromise.catch((error) => {
+        if (controller.signal.aborted && error === controller.signal.reason) return null;
+        throw error;
+      });
+    },
+    stop,
+    isStopping: () => stopPromise !== null,
+  };
+}
+
+export function cleanupDevelopmentLaunch({ logStream, logFile, resultFile }) {
+  logStream.kill();
+  rmSync(logFile, { force: true });
+  rmSync(resultFile, { force: true });
 }
 
 function terminateProcessTree(child) {
@@ -393,33 +512,52 @@ function terminateProcessTree(child) {
 }
 
 /**
- * Watches the detached bundle for the whole session, because `open` exits 0 at
- * the handoff and reports nothing afterwards.
- *
- * Waiting a fixed interval and checking once cannot work in either direction: a
- * first launch of the freshly signed bundle can still be starting, and quitting
- * the app is a normal thing to do at any moment. So this waits for the app to
- * appear, then reports the eventual disappearance as an ordinary session end.
+ * Waits only for Electron's lock verdict. After a winner is known, the detached
+ * TCC app owns its own lifecycle; the launcher neither polls nor terminates it.
  */
-export async function monitorDevelopmentApp(options = {}) {
-  const isRunning = options.isRunning ?? isDevelopmentAppRunning;
+export async function waitForDevelopmentLaunchVerdict(options = {}) {
+  const readLaunchResult = options.readLaunchResult ??
+    (() => readDevelopmentLaunchResult(options.resultFile));
   const delay = options.delay ?? ((ms) => new Promise((done) => setTimeout(done, ms)));
   const stopped = options.stopped ?? (() => false);
   const pollMs = options.pollMs ?? 250;
   const startupAttempts = options.startupAttempts ?? 120;
-  let appeared = false;
-  for (let attempt = 0; attempt < startupAttempts && !appeared; attempt += 1) {
+  for (let attempt = 0; attempt < startupAttempts; attempt += 1) {
+    const result = await readLaunchResult();
+    if (result?.status === 'loser') return 'absorbed';
+    if (result?.status === 'winner') return 'started';
     if (stopped()) return 'stopped';
-    if (isRunning()) appeared = true;
-    else await delay(pollMs);
-  }
-  if (!appeared) return 'never-started';
-  while (!stopped()) {
     await delay(pollMs);
-    if (stopped()) break;
-    if (!isRunning()) return 'exited';
   }
-  return 'stopped';
+  return 'never-started';
+}
+
+/**
+ * Launch-outcome decision, fully contained: absorbed/never-started are failures,
+ * stopped is a normal launcher exit, and started deliberately leaves the
+ * launcher's own dev resources running without supervising Electron.
+ */
+export function handleDevelopmentLaunchOutcome(outcome, effects = {}) {
+  const log = effects.log ?? console.error;
+  const exit = effects.exit ?? ((code) => { process.exitCode = code; });
+  switch (outcome) {
+    case 'started':
+      return;
+    case 'absorbed':
+      log('Maka Dev was absorbed by another instance holding the profile lock; quitting.');
+      exit(1);
+      return;
+    case 'never-started':
+      log('Maka Dev did not start within the startup window.');
+      exit(1);
+      return;
+    case 'stopped':
+      exit(0);
+      return;
+    default:
+      log(`Unknown Maka Dev monitor outcome: ${String(outcome)}`);
+      exit(1);
+  }
 }
 
 export async function prepareDevelopmentApp() {
@@ -433,9 +571,9 @@ export async function prepareDevelopmentApp() {
   const electronVersion = JSON.parse(readFileSync(ELECTRON_PACKAGE, 'utf8')).version;
   if (isCurrentRuntime(electronVersion)) return DEV_APP;
 
-  // Rebuilding unlinks the bundle an already-running app was launched from,
-  // and deletes the environment file it would read on relaunch.
-  await ensureNoRunningDevelopmentApp();
+  // Rebuilding unlinks the bundle an already-running app was launched from.
+  // Refuse rather than terminating a process the launcher did not create.
+  assertDevelopmentAppNotRunning();
 
   try {
     await rebuildDevelopmentRuntime({
@@ -503,6 +641,9 @@ export function createRuntimeMarker(electronVersion) {
     electronVersion,
     bundleId: DEV_BUNDLE_ID,
     desktopDir: DESKTOP_DIR,
+    // Burned into the generated bootstrap, so a change here must invalidate
+    // the cached bundle (isDevelopmentRuntimeCurrent compares every field).
+    userDataDir: DEV_USER_DATA_DIR,
   };
 }
 
--- a/apps/desktop/README.md
+++ b/apps/desktop/README.md
@@ -44,14 +44,17 @@ or this repository's path changes; run
 This workflow is opt-in because it costs a codesign rebuild and puts an extra
 app in your Dock and in System Settings. Developers who are not touching OS
 permissions should not pay for it. Main-process logs are still streamed to the
-terminal — `open` redirects them to `.maka-dev/app.log`, which the launcher
-follows.
+terminal. Each TCC launch gets separate private log and result files: logs are
+observation only, while the one-shot result carries the single-instance verdict.
 
 Everything the bundle needs is fixed when it is built, so a launch with no
 arguments and no environment — the Dock, Spotlight, or Screen Recording's
-“Quit & Reopen” — produces a correct app. There is no session protocol or
-supervising process: the app instance and the dev session are separate
-lifecycles. Application-control variables (API keys, `MAKA_*`, the Vite URL)
+“Quit & Reopen” — produces a correct app. There is no long-lived session
+protocol or supervising process: the app instance and the dev session are
+separate lifecycles. Ctrl-C stops the launcher and, for `npm run dev`, its Vite
+server; it does not quit a TCC app launched through LaunchServices. Quit that
+app with Cmd-Q. Likewise, quitting the TCC app does not stop the dev server.
+Application-control variables (API keys, `MAKA_*`, the Vite URL)
 are published to an ignored `0600` file at `.maka-dev/dev-env.json` rather than
 a command line. That file, not the shell, is what makes a Dock or “Quit &
 Reopen” launch work, since those have no parent shell at all. `PATH` is not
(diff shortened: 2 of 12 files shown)


hqhq1025 (approved):
Reviewed the current revision across the TCC bootstrap, the plain Desktop and repository CLI profile contracts, and runtime cache invalidation. The shared `Maka Dev` data root now matches the other development entry points while the worktree-scoped bundle identity remains unchanged; the previous marker is invalidated by the new burned-in path. The script syntax and focused marker/bootstrap assertions passed, and the required CI check is green.

jackwener (commented):
Automated Codex review — blocking finding on exact head `c17e155f47d8ff5cae02cb20a41f5a9f91950b01`.

**P2: the shared profile invalidates the launcher's worktree-local stale-owner recovery.** `resolveMacosDevelopmentLaunch()` still calls `ensureNoRunningDevelopmentApp()`, but that function probes and kills only this worktree's `DEV_EXECUTABLE`. After this PR, a TCC app from worktree A and a launch from worktree B use the same `Maka Dev` single-instance lock while having different executable paths. If A is still running—or was orphaned after its terminal/Vite process died—B cannot see or stop it. B's Electron instance is absorbed by A's lock and exits, while B's monitor only watches B's path and eventually reports `never-started`. This is the same stale-window failure the ownership step's comment says it prevents, now moved to the cross-worktree case.

The README partly exposes the mismatch: it says another worktree's app survives and holds the shared lock, then says a launch reclaims an app left by a hard-killed session. The latter is no longer true across worktrees.

Before merge, make launch ownership profile-scoped as well as `userData` profile-scoped (or fail immediately with an explicit current-owner contract), and add a two-worktree regression that proves a stale owner cannot absorb the new launch. Keep `DEV_BUNDLE_ID` worktree-scoped for TCC as this PR already does.

Required conclusions:

1. Optimal for the actual problem: **not yet**; the data-root correction is sound, but ownership follows the old isolation boundary.
3. Tests to delete/replace: none; current marker smoke coverage misses the cross-worktree lock behavior.
4. Deeper refactor: no; align the existing stale-owner recovery with the new shared-profile boundary.
5. Ready to merge: **no**, despite green `test` and a current-head committer approval.
6. Residual risk/gaps: stale/other-worktree app absorption and 30-second false `never-started` failure.

This is a developer-workflow behavior change; the existing independent human review remains necessary under `CONTRIBUTING.md` after the blocker is resolved.

jackwener (dismissed):
Re-review of unchanged exact head `c17e155f47d8ff5cae02cb20a41f5a9f91950b01` against current `main@d62857a8357e9160926726a2a13096bc2dc2b91d`. Converting the existing blocking automated comment into an explicit changes-requested gate because the head has not changed.

**P2 — shared `userData` leaves stale-owner recovery worktree-local.** `resolveMacosDevelopmentLaunch()` still asks `ensureNoRunningDevelopmentApp()` to find/kill only this worktree's `DEV_EXECUTABLE`. After the PR, worktree A and B use one `Maka Dev` single-instance lock but retain different executable paths. A running or orphaned app from A is invisible to B's recovery; B is absorbed by A's lock, exits, and its own-path monitor eventually reports `never-started`. Make launch ownership profile-scoped too, or fail immediately with an explicit current-owner contract, and add a two-worktree regression.

Required conclusions remain:

1. Optimal: **not yet**; data ownership and process ownership use different boundaries.
3. Tests to delete/replace: none; add cross-worktree ownership coverage.
4. Deeper refactor: no; align the existing recovery seam with the shared profile.
5. Ready to merge: **no**, despite green `test` and the earlier human approval.
6. Residual risks/gaps: stale-window absorption and a false 30-second `never-started` failure; developer-workflow behavior needs fresh independent human review after repair.

Astro-Han:
Pushed a fix for the blocking finding — new head `21e165b87`, rebased onto `4acfa2693`.

@jackwener your P2 was right: sharing the `userData` root made Electron's single-instance lock cross-worktree while `ensureNoRunningDevelopmentApp()` still probed and killed only this worktree's `DEV_EXECUTABLE`. We verified the absorption path end to end before changing anything.

We took the second of the two options you offered — **fail immediately with an explicit current-owner contract** — rather than making the kill profile-scoped. The reason: a profile-scoped kill would mean worktree B's launch script terminates a window someone may be actively using in worktree A. Sharing a data root does not confer disposal rights over another worktree's process, so naming the owner and stopping is both smaller and more honest than expanding what we are willing to kill.

What changed:

- `assertNoCrossWorktreeOwner()` runs *before* `ensureNoRunningDevelopmentApp()` — establish that no foreign owner holds the lock first, then clean up our own leftovers.
- `sharedDevelopmentAppCommandLines()` probes via `pgrep -af` and treats exit codes 0 and 1 as the only valid answers; anything else throws rather than being read as "nothing is running", so a probe failure cannot be mistaken for a clear lock.
- The error names the owner's full command line and says what to do about it (`Quit it (Cmd-Q)`), because the failure this replaces was exit-0 plus a `never-started` report 30 seconds later — indistinguishable from a normal launch.
- `apps/desktop/README.md` now says same-worktree reclaim still works while cross-worktree fails fast with the owner named, replacing the claim you flagged as no longer true.

The two-worktree regression you asked for is in `apps/desktop/scripts/dev-app-runtime.test.mjs` — 7 cases covering: another worktree's command line reported as owner, our own not reported, none running, a foreign owner producing a throw whose message carries both the guidance and the owner path, and a path that merely shares a suffix still classified as foreign. Wired into the desktop `test:dist` chain so it runs in CI rather than only locally.

`DEV_BUNDLE_ID` stays worktree-scoped for TCC, as you asked.

One note on the diff you will see: the branch's merge base predated the ASF header sweep, so the rebase onto current `main` picked up headers that already exist upstream. The only substantive commit is `21e165b87`.

---
*This work was AI-assisted. The finding was reproduced and the fix verified against the exact head above; please push back where we got it wrong.*

jackwener (dismissed):
Independent review of `21e165b87`. Did not treat Astro-Han authorship as evidence.

1) Problem. TCC dev `userData` was `Maka Dev-<worktreeHash>` while CLI and plain `npm run dev` use `Maka Dev`. Enabling TCC silently switched the database and broke attaching to the same host.

2) Solution. Share `~/Library/Application Support/Maka Dev`. Keep `DEV_BUNDLE_ID` worktree-scoped for TCC. Burn `userDataDir` into the runtime marker so old cached apps rebuild.

3) The earlier P2 (worktree-local stale-owner recovery vs shared lock) is fixed: `assertNoCrossWorktreeOwner` pgrep's every `Maka Dev.app` Electron, treats a non-own path as owner, and fails immediately instead of being absorbed. Own-path prefix is not confused with a sibling suffix. Tests cover that. Own-worktree leftovers still go through `ensureNoRunningDevelopmentApp`.

Approve. Merge stays with the author.

jackwener (commented):
Follow-up after independently checking Darwin `pgrep` on macOS 26. My earlier APPROVE was too early.

`pgrep -af` on Darwin is **not** Linux procps. Here `-a` means include ancestors, and without `-l` stdout is **PIDs only**. `-f -l` is what prints PID + full argv. The production probe therefore returns bare PIDs, `startsWith(ownExecutable)` never matches, and this worktree's leftover TCC app is treated as a foreign owner — fail-fast runs **before** `ensureNoRunningDevelopmentApp()`. That is a real P1 on this head.

Second P1: the probe only matches `Maka Dev.app/.../Electron`. Plain `npm run dev` is `Electron.app/.../Electron` with `app.setName('Maka Dev')`, which still holds the shared profile lock. That process is invisible to the new owner check, so the original absorption/`never-started` hole is not actually closed.

P2: `--user-data-dir` is applied after the owner probe, so an isolated launch can be blocked by an unrelated default-profile owner.

The shared `Maka Dev` root and worktree-scoped `DEV_BUNDLE_ID` are still the right data-plane cut. The process-owner probe has to key on the **profile**, with Darwin `pgrep -fl`, not on one bundle suffix.

Not converting this to REQUEST_CHANGES (comment-only). I no longer consider this head merge-ready.

apps/desktop/scripts/dev-app-runtime.mjs
```diff
@@ -139,21 +148,197 @@ export function toProcessMatchPattern(executable) {
...
+
+/**
+ * Absolute root of the worktree a dev-app command line came from, by scanning
+ * the argv tokens for a path that reaches into a repository checkout
+ * (`/apps/desktop` or `/node_modules/`). The shim, the resolved Electron
+ * binary, the bundle, and the positional desktop-dir argument all carry such
+ * a token.
+ */
+export function worktreeFromCommandLine(commandLine) {
+  const tokens = commandLine.split(/\s+/);
```
jackwener: P1: `split(/\s+/)` drops any marker that does not stay an absolute token. `Maka Dev.app` happens to still work because `/apps/desktop` is before that space; `/Users/me/Dropbox (Personal)/maka/apps/desktop/...` returns `undefined`. Same class this file already escapes for pgrep. Search the raw line for `/apps/desktop` / `/node_modules/`, then walk left to a ` /` argv boundary.

apps/desktop/scripts/dev-app-runtime.mjs
```diff
@@ -139,21 +148,197 @@ export function toProcessMatchPattern(executable) {
...
+  const commandLines = options.commandLines ?? sharedDevelopmentAppCommandLines(options);
+  const target = targetProfile ?? DEV_USER_DATA_DIR;
+  return commandLines.find((line) => {
+    // Own worktree: any dev-app shape (bundle, node shim, resolved Electron
+    // binary) carries this worktree's absolute root in an argv token, so the
+    // same root-recovery used for the profile gives an exact comparison —
+    // `includes` would also swallow a deeper path that merely begins with our
+    // root (e.g. a sibling checkout nested under it).
+    if (worktreeFromCommandLine(line) === ownRoot) return false;
+    const profile = resolveProcessUserDataDir(line, options) ?? DEV_USER_DATA_DIR;
```
jackwener: P1: `resolveProcessUserDataDir(...) ?? DEV_USER_DATA_DIR` plus the broad Electron pgrep filter attributes every unmatched npm-electron process to the shared Maka Dev lock. A missing env file is not evidence that the process is using this profile. Default only for the unique `Maka Dev.app` shape, or when the env file actually exists and omits `userDataDir`.

jackwener (commented):
Independent re-review of `459995ac`. The pgrep-then-ps split is the right Darwin fix (I reproduced `pgrep -f` PIDs + `ps -p pid -o command=` on this Mac; BSD `-a` is not involved). The owner *judgment* still has two P1s. Do not merge this head. GitHub `APPROVED` is a leftover hqhq1025 review on `c17e155f`, not a current GO.

**P1: `worktreeFromCommandLine` splits on whitespace.** Paths contain spaces. This file already documents `~/Dropbox (Personal)` as a pgrep-regex case. On that command line the `/apps/desktop` and `/node_modules/` markers no longer start with `/`, so the function returns `undefined`. The owner then does not recognise self and `?? DEV_USER_DATA_DIR` treats the process as a foreign holder of the shared lock — self-blocking, same shape as `pgrep -af`. `Maka Dev.app` still parses only because `/apps/desktop` appears *before* that space; that is luck. Scan the raw string for the marker, then walk left to a ` /` (space+slash) argv boundary; do not tokenise.

**P1: the scan is any npm Electron, not Maka Dev.** The rough filter matches `node_modules/.bin/electron` and `Electron.app/Contents/MacOS/Electron`. Another repo's Electron recovers a worktree, has no maka `dev-env.json`, and is attributed to the shared Maka Dev profile. Launching then fail-fasts as a cross-worktree owner. Missing env file ≠ "uses the Maka default profile". Only `Maka Dev.app`, or a worktree that actually published a maka env file, should participate in the lock judgment.

Tests still use synthetic argv and say real Darwin `pgrep`/`ps` samples are "still being collected". That is the same self-proving fixture failure as the previous `startsWith(node_modules/.bin/electron)` head. Cover a space-containing root and an unrelated `node_modules/electron` process.

apps/desktop/scripts/dev-app-runtime.mjs
```diff
@@ -342,8 +486,15 @@ export async function startDevelopmentApp(options = {}) {
...
-  const launch = await resolveMacosDevelopmentLaunch();
+  const launch = await resolveMacosDevelopmentLaunch(process.env, argv);
   if (!launch) {
+    // The plain shape uses the same shared profile, so the same owner gate
+    // applies before we spawn — otherwise another worktree's app (TCC or
+    // plain) would absorb this launch through the single-instance lock.
+    const targetProfile = splitDevelopmentCliArgs(argv).userDataDir;
+    assertNoCrossWorktreeOwner({ targetProfile });
+    // A leftover own dev app would absorb this launch through the lock.
+    await ensureNoRunningDevelopmentApp();
```
jackwener: P1 on Linux: this new plain-path reclaim calls `isDevelopmentAppRunning`, which now pgrep's the npm shim. `quitMacosDevelopmentApp` returns false when `platform !== 'darwin'`, so a leftover same-worktree shim throws instead of being ignored (main) or killed. Windows is skipped in the probe; Linux is not.

jackwener (commented):
Integration pass on `cbdd2cc2` (not an approve).

TCC and plain both hit `assertNoCrossWorktreeOwner` then `ensureNoRunningDevelopmentApp`. Data-root drop of `WORKTREE_ID` plus the legacy-dir warn matches the README shutdown/owner text. CI runs `test:dist`, which includes the new script tests.

**P1:** Linux `startDevelopmentApp` now probes this worktree's `node_modules/.bin/electron` shim, but `quitMacosDevelopmentApp` still no-ops off darwin. A leftover same-worktree `npm run dev` therefore throws `could not be stopped`. On main, liveness only looked at the Darwin bundle path, so this path did not fail-closed. Either teach quit to signal the Linux shim, or keep shim/real reclaim Darwin-only.

GitHub `APPROVED` is still the stale hqhq1025 review on an older head.

apps/desktop/scripts/dev-app-runtime.mjs
```diff
@@ -413,11 +640,17 @@ export async function monitorDevelopmentApp(options = {}) {
...
+  if (!appeared) {
+    recheck(recheckInput);
+    return 'never-started';
+  }
   while (!stopped()) {
     await delay(pollMs);
     if (stopped()) break;
-    if (!isRunning()) return 'exited';
+    if (!isRunning()) {
+      recheck(recheckInput);
```
jackwener: P1: this is the Cmd-Q / normal end path (`appeared` was true). Recheck here will print an absorption error if anything else still matches `holdsProfile` (including a foreign `apps/desktop` Electron). Recheck belongs on `never-started` only.

apps/desktop/scripts/dev-app-runtime.mjs
```diff
@@ -342,13 +545,30 @@ export async function startDevelopmentApp(options = {}) {
...
+    // A leftover own dev app would absorb this launch through the lock.
+    await ensureNoRunningDevelopmentApp();
     const child = spawn(resolveElectronBinary(), [DESKTOP_DIR, ...argv], {
       cwd: DESKTOP_DIR,
       stdio: 'inherit',
       env: viteUrl ? { ...process.env, VITE_DEV_SERVER_URL: viteUrl } : process.env,
     });
+    // Post-spawn recheck: an exit-0 without our app appearing means the lock
+    // was absorbed by a foreign holder — name it instead of exiting quietly.
+    child.once('exit', (code) => {
```
jackwener, line 393: P1: Electron second-instance absorption AND a normal Cmd-Q both exit 0. Rechecking every exit-0 will false-alarm a finished session whenever another holder is still visible. Gate this on a short-lived child (startup window), not every zero exit.

jackwener (commented):
Integration re-review of `4f6e838d` (not an approve).

**P1:** post-spawn recheck on `appeared → exited` and plain `child exit 0` conflates absorption with a normal session end. Cmd-Q of a TCC app takes the appeared-then-exited path; a finished `npm run dev` takes exit-0. `recheckAfterAbsence` then reports any remaining foreign/false-positive holder as "absorbed this launch". That is only valid when this launch never became the lock owner (`never-started`, or an immediate second-instance exit). After the app has appeared, the lock was ours; a later process is not this launch being absorbed.

Tests currently require the appeared→exited path to print the conflict. That encodes the false positive.

Fix: recheck only on `never-started` (TCC) and on a short-lived exit-0 (plain). Do not recheck after a session that actually ran.

Astro-Han:
Heads-up: this branch was force-pushed twice just now, so anyone with a local copy will need to re-fetch. Current head is `8d3e52f70bfb4e3dd74ad77d5dff078a987e066c`.

**What changed in the code.** The launcher's pre-flight owner probe is gone. It tried to answer "which worktree currently owns this dev data directory?" by matching `pgrep -f` output against an expected bundle path and argv shape. That premise does not hold: the TCC profile is not on the command line (it lives in a mutable `dev-env.json`), plain values contain spaces, the npm shim reports `argv[0]` as `node`, and helper, GPU, and about-to-exit processes all match the same pattern. A LaunchServices-started app never presents the argv shape the probe expected at all. Rather than keep patching the heuristic, we removed it and rely on the single-instance lock, which is the actual authority for that question. `dev-app-profile.mjs` and its test are deleted; net effect on the branch is roughly −480 lines.

Two smaller fixes went in with it: the monitor's `default` branch is now fail-closed instead of exiting `0` on an unrecognized outcome, and `splitDevelopmentCliArgs` now strips the launcher flag unconditionally. The latter was conditional before, which meant a user passing the flag by hand could persist it into `dev-env.json` and permanently silence the conflict dialog for Dock launches.

Some context on the rejected approach is in #3539, including what we verified about Chromium's `NotifyOtherProcessOrCreate` behaviour and where our conclusions are inferences rather than direct observations.

**Why the second force-push.** Twenty-eight of the thirty commits were authored as `Ubuntu <ubuntu@localhost.localdomain>` — the default git identity on the machine the work ran on, not the actual author. They are now correctly attributed. The tree is byte-identical across the rewrite (`git diff` between the pre- and post-rewrite heads is empty) and the commit count is unchanged; only author and committer metadata differ.

@hqhq1025 — sorry for the churn. Your approval is on `c17e155f` from Aug 20, which predates all of this, so it no longer reflects what is here. Whenever you have time, a fresh look at `8d3e52f7` would be appreciated. No rush from our side.

apps/desktop/scripts/dev-app-runtime.mjs
```diff
@@ -403,23 +524,60 @@ function terminateProcessTree(child) {
  */
 export async function monitorDevelopmentApp(options = {}) {
   const isRunning = options.isRunning ?? isDevelopmentAppRunning;
+  const logFile = options.logFile ?? developmentLogFile;
   const delay = options.delay ?? ((ms) => new Promise((done) => setTimeout(done, ms)));
   const stopped = options.stopped ?? (() => false);
   const pollMs = options.pollMs ?? 250;
```
Astro-Han, line 523: **[P2] Two concurrent launches in the same worktree make the lock *winner* consume the *loser's* marker, and both stop.**

`developmentLogFile` is a single fixed path per worktree (`.maka-dev/app.log`), truncated on every launch, and each launcher starts its own `tail -F` on it. So when two TCC launchers in the same worktree both pass `ensureNoRunningDevelopmentApp()` before either Electron becomes visible, they `open -n` concurrently, share one log, and both monitor it. The lock loser writes the marker — and **both** monitors read it.

**Reproduced** by driving two production `monitorDevelopmentApp` instances against one marker log: `{"outcomes":["absorbed","absorbed"],"exits":[1,1]}`. Both launchers then call `app.stop()`, so the process that actually won the single-instance lock is terminated too.

This is recoverable by retrying, but it is worth fixing at the mechanism rather than at the symptom: the marker is an *authoritative* verdict about one specific launch, and it is currently being attributed to whichever launcher happens to be reading. Replacing a heuristic with a real signal (which this PR does, correctly) only pays off if the signal is addressed to its owner.

**Why the new tests do not catch it:** the marker carries a profile path but no launch nonce or PID, and all thirteen new tests run a single monitor at a time — so nothing in the suite can distinguish two consumers of one log.

**Smallest fix:** give each launch its own log/marker, or write a launch nonce into the marker and have a launcher consume the loser verdict only when the nonce is its own.

apps/desktop/scripts/dev-app-runtime.mjs
```diff
@@ -168,12 +245,36 @@ function sendSignalToExecutable(name, executable) {
...
   const probe = options.probe ?? defaultLivenessProbe;
-  return probe(executable);
+  // The shared-profile lock is a macOS concept (plain dev and the TCC bundle
+  // both reach it only there). On other platforms keep main's behavior: only
+  // the bundle is probed — the bundle never exists off macOS, so liveness is
+  // false and the launch proceeds to the single-instance lock. Probing the
+  // shim/Electron shapes off-darwin would hard-fail a plain `npm run dev`
+  // that main allowed (the shim exists on Linux; quit is darwin-only).
+  if ((options.platform ?? process.platform) !== 'darwin') return probe(bundle);
+  return probe(bundle) || probe(shim) || probe(real);
```
Astro-Han: **[P2] Ownership is decided by executable path alone, so a plain dev launch kills this worktree's own E2E / smoke Electron.**

`isDevelopmentAppRunning` probes the bundle plus the Electron shim and the real native binary for this worktree, and `quitMacosDevelopmentApp` then sends TERM/KILL to that same binary path. But a Playwright Electron E2E run (`apps/desktop/e2e/fixtures.ts:380-392`) and `electron scripts/browser-observe-act-smoke.mjs` execute **the same binary** — `node_modules/electron/dist/Electron.app/Contents/MacOS/Electron` in this checkout.

**Verified:** feeding that real binary path to the production helper as a non-dev probe hit returns `running = true`. So running E2E or the smoke script and then starting `npm run dev` / `npm start` on macOS terminates the local verification that is in progress, not a stale Maka Dev.

**Why the guards do not cover it:** `REPO_ROOT` isolates *other* worktrees, which is a different axis — it cannot separate two Electron workloads inside the same worktree. The tests assert that all three paths get probed, but there is no counter-example with a non-dev command line.

This predates the earlier review rather than arriving with the recent commits.

**Smallest fix:** match on the desktop app's positional path or an explicit dev marker in the command line, not on the executable alone.

apps/desktop/scripts/dev-app-runtime.mjs
```diff
@@ -90,17 +97,33 @@ const DEV_USER_DATA_DIR = join(
...
 const DEV_BUNDLE_ID = `com.maka.dev.${WORKTREE_ID}`;
 const RUNTIME_SCHEMA_VERSION = 7;
-const DEV_ENV_SCHEMA_VERSION = 1;
 
 export const developmentAppPath = DEV_APP;
 const developmentExecutablePath = DEV_EXECUTABLE;
 
-export async function resolveMacosDevelopmentLaunch(env = process.env) {
+export async function resolveMacosDevelopmentLaunch(env = process.env, argv = []) {
   if (!shouldUseMacosDevelopmentApp(process.platform, env)) return null;
```
Astro-Han, line 101: **[P2] On the first launch after upgrading, a plain start returns here before the legacy-root warning can run — so the migration notice the description promises is never shown.**

`warnAboutLegacyTccDataRoot()` is called on the next line, but this early return fires first whenever `shouldUseMacosDevelopmentApp` is false — which is exactly the plain path with `MAKA_DEV_TCC` unset.

**The scenario is the supported upgrade path, not an edge case:** an older TCC build left `Maka Dev-<WORKTREE_ID>`, the developer upgrades and starts plainly. They land in the shared `Maka Dev` profile with none of their sessions or settings, and receive no explanation. The warning text is well written — it says where the data is and that it is not deleted — but a developer on this path never sees it, and the natural conclusion is that the data was lost.

**Why the tests do not catch it:** nothing covers the warning at all, and no test drives both the plain and TCC entry points against a real legacy directory.

**Smallest fix:** emit the notice before the TCC gate, or from a single `startDevelopmentApp` entry point, deciding on the effective `--user-data-dir` so it does not fire spuriously.

Astro-Han (commented):
Independent re-review at exact head `8d3e52f70bfb4e3dd74ad77d5dff078a987e066c`. All three check-runs (`audit`, `test`, `package`) are `completed / success` on that SHA.

**Disclosure first, because it changes how you should read this.** This PR is authored under the `Astro-Han` account, which is the same shared account our review line publishes from — so we are the responsible party here and cannot approve our own work. This review was therefore run as a **separate, blind line**: the reviewer reached a provisional judgment on the head before reading any of the existing GitHub review threads, and only then reconciled against them. It is posted as a comment, not an approval, by design.

**Every earlier finding is closed, and closed properly.** The last outstanding `[P1]` — the post-spawn owner recheck reporting false absorption on `appeared→exited` and on any exit-0 — is gone because the heuristic itself was removed. The launch now consumes a real loser marker / exit 42 from the main process, so the finding's precondition no longer exists rather than being papered over. The same is true for the Darwin `pgrep -af` tokenization, the space-in-path problem, foreign-Electron misidentification, and the Linux plain-reclaim conflict: the roughly 480 lines of heuristic ownership probing that carried them are deleted. **Removing a second source of truth rather than patching it is the right shape**, and it is worth saying so plainly.

**The audit was run across the whole range from the previously reviewed tree `4f6e838d`, not just the last commit** — which is what surfaced the split below.

**Three `[P2]`s and one `[P3]`.** Of these, **only Finding 1 arrived with the replacement loser-contract**; Findings 2 and 3 predate the earlier review and were missed by it. They are marked as such rather than presented as regressions, because the distinction matters when you decide what to fix and where.

**Two candidates were deliberately withdrawn** rather than left in: `splitDevelopmentCliArgs` preserving a canonical flag (only reachable if an unsupported caller deliberately substitutes the contract value; every production caller omits the override), and the core-constants import failure (it fails toward the native dialog, and plain's original non-zero exit code survives — not a silent success).

*This review is AI-assisted. Under `CONTRIBUTING.md` §Review it does not count as the required independent human review; and since the PR is authored from the same shared account, an approval from this line would not be meaningful here regardless.*

apps/desktop/scripts/dev-app-runtime.mjs
```diff
@@ -370,9 +441,11 @@ export async function startDevelopmentApp(options = {}) {
   return {
     child,
     isMacosBundle: true,
-    stop: async () => {
+    logFile: launch.logFile,
+    stop: async (stopOptions = {}) => {
       logStream.kill();
-      await quitMacosDevelopmentApp();
+      if (stopOptions.terminateApp !== false) await quitMacosDevelopmentApp();
```
M4n5ter: **[P2] Please do not terminate the bundle until this invocation has positively proved ownership.**

The new `terminateApp: false` guard only runs after `monitorDevelopmentApp()` has classified the private loser marker as `absorbed`. Before that classification, both launcher scripts still use the default `stop()` here: a `SIGINT`/`SIGTERM`/`SIGHUP`, or a non-zero/error exit from `open`, calls this path with `terminateApp` unset.

Reproduction: keep a TCC `Maka Dev.app` from this worktree running on the shared profile; start a second TCC launch from the same worktree; interrupt the second launcher before its private log exposes the loser marker (or make its `open` handoff fail). `stop()` calls `quitMacosDevelopmentApp()`, whose `pkill -f` target is the worktree bundle executable path, so it terminates the already-running lock owner even though this launch never proved it acquired the lock.

The current guards cannot distinguish these processes: `isDevelopmentAppRunning()` proves only that *some* process uses this bundle path, the private log has a loser marker but no positive winner/ownership marker, and the two script couplers are explicitly outside the focused tests. This contradicts the replacement contract that the single-instance loser must not dispose of the owner.

Could you make termination conditional on positive ownership tied to this invocation (for example, a private winner marker/PID), and fail closed while ownership is unknown? Absence of a loser marker or bundle-path liveness is not ownership.

<details>
<summary>中文</summary>

新的 `terminateApp: false` 只有在 monitor 已把私有日志里的 loser marker 判成 `absorbed` 后才生效。在此之前，两个 launcher 遇到信号或 `open` 非零/错误仍会调用默认 `stop()`；它按本 worktree 的 bundle 可执行路径执行 `pkill -f`，会杀掉已经持有共享 profile 锁的旧实例，即使本次启动从未证明自己拿到锁。

请把终止动作绑定到本次 invocation 的正向所有权证据（例如私有 winner marker/PID），所有权未知时 fail closed。没有 loser marker、或仅看到 bundle 路径存活，都不能证明所有权。

</details>

M4n5ter (commented):
Technical review of exact head `e9f1232170d594355ae646c5fa8f0ede17017b61`. I found one development-only P2 below, so the implementation is not yet GO: an unclassified TCC invocation can still terminate the existing same-worktree lock owner before proving ownership. Exact-head hosted `audit`, `test`, and `package` checks are terminal green; the focused launcher suite passes 19/19, core passes 606/606, and the current-main merge-tree is clean.

<details>
<summary>中文</summary>

本次技术审查绑定精确 head `e9f1232170d594355ae646c5fa8f0ede17017b61`。当前实现仍有下方 1 个仅开发态的 P2，因此暂未达到 GO：尚未完成归类的 TCC invocation 在证明所有权前，仍可能终止同一 worktree 中已经持有锁的实例。精确 head 的 `audit`、`test`、`package` 均终态绿；launcher 聚焦测试 19/19、core 606/606 通过，且与当前 main 的 merge-tree 干净。

</details>

jackwener (commented):
Independent incremental review of exact head `e9f1232170d594355ae646c5fa8f0ede17017b61`, performed by the prior Kabi reviewer with the new-commit judgment sealed before reading current review conclusions. Acting on behalf of @WAWQAQ under the standing Kabi review/action authorization; Kabi holds the publisher seat for this head.

**Conclusion: COMMENT — one P2 remains, plus one P3. No APPROVE.**

The previous findings are properly closed or moot on this head: executable-path ownership probing, the shared launch log, the plain-start legacy warning gap, and the Linux preflight reclaim path were removed or moved to the correct authority boundary. The selected product contract is also implemented consistently: both development entry points use the shared `Maka Dev` profile.

**[P2] An unclassified TCC launcher can still terminate the existing lock owner before proving that this invocation owns the bundle.**

`apps/desktop/scripts/start-dev-app.mjs:25-52` and `apps/desktop/scripts/dev.mjs:156-185` send signals and non-zero `open` exits through the default stop path. That calls `app.stop()` with `terminateApp` unset; `apps/desktop/scripts/dev-app-runtime.mjs:445-448` therefore calls `quitMacosDevelopmentApp()`, whose `pkill -f` match is the worktree bundle executable path, not this launch identity. Only the already-classified `absorbed` path passes `terminateApp: false`.

Reachable reproduction: keep the same-worktree TCC bundle running on the shared profile, start another TCC launcher, then interrupt that launcher before its private log reports the loser marker (or make the LaunchServices handoff fail). The second invocation has no positive ownership evidence, yet its teardown kills the existing owner. This corroborates the exact-head inline finding already posted by M4n5ter; it was independently traced after provisional sealing rather than independently discovered.

The clean fix is to make termination depend on positive ownership tied to this invocation (winner marker/PID/nonce) and fail closed while ownership is unknown. Add signal and non-zero-handoff production-seam tests with a pre-existing owner.

**[P3] Per-launch logs can accumulate after abnormal launcher death.**

`apps/desktop/scripts/dev-app-runtime.mjs:162-169` creates a UUID log under `tmpdir()/maka-dev-launches`; cleanup exists only in `stop()` at `:448`. SIGKILL, a crash, or power loss bypasses cleanup and leaves the full session output indefinitely until OS temp cleanup. Add bounded startup cleanup (age-based at minimum; launcher-PID ownership is cleaner). This remains non-blocking because it is Darwin-only, explicitly opt-in, and has OS temp cleanup as a fallback.

Fresh gate immediately before publication: OPEN, non-draft, MERGEABLE/CLEAN; exact-head hosted `audit`, `test`, and `package` are all `completed/success`. GitHub's aggregate `APPROVED` includes stale/other-account history and does not override this live P2. No merge action was taken. This Agent-executed review does not claim to satisfy the unresolved ASF required-human-review wording.

M4n5ter (commented):
<details open>
<summary>English</summary>

Independent re-review of exact head `18539a8cf13b3c66b5dd82883c81f423edfe9b1e`: **COMMENT — one P2 remains.**

**[P2] `apps/desktop/scripts/dev-app-runtime.mjs:207-216` trusts the last winner-marker-shaped substring in the app's mixed stdout log as the PID to terminate.**

The new per-launch log separates launches, but it is still the complete stdout/stderr stream of the winning Electron app. `devLogWinnerPid()` uses `lastIndexOf(DEV_WINNER_LOG_MARKER)` and accepts the following positive integer from anywhere in that stream. Supported application data can contain the same bytes later: for example, Runtime Host profile names reject control characters but allow a name such as `[dev] single-instance winner: 222`; an unavailable profile is logged verbatim by `runtime-host-profile-service.ts:826`.

Reproduction: a private launch log containing the legitimate first line `... winner: 111`, followed by `[runtime-host] ... winner: 222 is unavailable`, makes `stopDevelopmentLaunch()` send both TERM and KILL to PID 222. PID 111 — the app this launcher actually won — remains running. The current test covers only one uncontested marker, so file permissions and per-launch uniqueness do not block this path.

Please keep the ownership result outside the mixed application log (for example, a dedicated one-shot result file/pipe), or at minimum accept only the first exact, line-anchored winner record emitted before full boot and ignore/reject later marker-shaped output.

The earlier bundle-wide `pkill -f` finding is otherwise genuinely fixed: an unclassified launch no longer terminates another worktree, and the normal classified path targets a PID rather than a bundle pattern. Exact-head `audit`, `test`, and `package` are all completed/success; the focused launcher suite passed 18/18 after rebuilding Core. No other new P0–P3 was found in the incremental surface.

</details>

<details>
<summary>中文</summary>

对 exact head `18539a8cf13b3c66b5dd82883c81f423edfe9b1e` 的独立增量复核结论：**COMMENT——仍有一条 P2。**

**[P2] `apps/desktop/scripts/dev-app-runtime.mjs:207-216` 把应用混合 stdout 日志中最后一个形似 winner marker 的子串当成要终止的 PID。**

新的每次启动独立日志隔离了不同 launcher，但它仍承载获胜 Electron 应用的全部 stdout/stderr。`devLogWinnerPid()` 使用 `lastIndexOf(DEV_WINNER_LOG_MARKER)`，并接受该子串后任意位置的正整数。受支持的应用数据可以在启动后再次产生同样字节：例如 Runtime Host profile 名只禁止控制字符，允许命名为 `[dev] single-instance winner: 222`；profile 不可用时，`runtime-host-profile-service.ts:826` 会把名称原样写入日志。

复现：私有 launch log 第一行是真实的 `... winner: 111`，之后出现 `[runtime-host] ... winner: 222 is unavailable`，`stopDevelopmentLaunch()` 会向 PID 222 同时发送 TERM 与 KILL；launcher 实际拥有的 PID 111 继续运行。现有测试只覆盖单个无争议 marker，因此文件权限与每次启动唯一日志都拦不住该路径。

请把 ownership 结果移出混合应用日志（例如使用专用的一次性结果文件/pipe）；最低限度也应只接受 full boot 前写入的第一条、整行锚定的 winner 记录，并忽略或拒绝之后形似 marker 的输出。

除此之外，上一轮 bundle-wide `pkill -f` finding 已真正关闭：未分类 launch 不再终止其他 worktree，正常已分类路径也改为按 PID 而非 bundle pattern 终止。exact-head `audit`、`test`、`package` 均 completed/success；重建 Core 后定向 launcher 测试 18/18。增量范围内未发现其他新 P0–P3。

</details>

jackwener (commented):
Independent incremental review of exact head `18539a8cf13b3c66b5dd82883c81f423edfe9b1e`, performed by the prior Kabi reviewer with the new-commit judgment sealed before reading current reviewer conclusions. Acting on behalf of @WAWQAQ under the standing Kabi review/action authorization; Kabi holds the publisher seat for this head.

**Conclusion: COMMENT — one P2 remains, plus one P3. No APPROVE.**

The prior P2 is genuinely closed at the mechanism level: bundle-wide `pkill -f` termination is deleted; an unclassified launch sends no signal; a classified winner targets only the PID recorded in its private launch result. Old executable-path ownership heuristics remain deleted, and the selected shared `Maka Dev` AppProfile contract is implemented consistently.

**[P2] Mixed application logs can override the winner PID used for termination.**

`apps/desktop/scripts/dev-app-runtime.mjs:207-218` uses `lastIndexOf(DEV_WINNER_LOG_MARKER)` over the entire per-launch stdout/stderr log. A supported Runtime Host profile name may contain printable text such as `[dev] single-instance winner: 222` (`packages/runtime-host/src/client/host-profile.ts:697-705`), and an unavailable startup profile is logged verbatim to stderr after the real winner marker (`apps/desktop/src/main/runtime-host-profile-service.ts:819-827`). Because `open` redirects both stdout and stderr into the same file, `stopDevelopmentLaunch()` then reads PID 222, sends TERM and later KILL to it, and leaves the actual owner PID 111 running. If PID 222 is not signalable, the teardown can instead throw `EPERM`.

This reachability candidate first appeared in the same-head review already published by M4n5ter after the Kabi provisional was sealed; Kabi independently traced and confirmed it. It is corroborated evidence, not an independent discovery.

Minimum lift: accept only the first exact, line-anchored winner record emitted before boot; add a production-seam log containing a legitimate winner line followed by the marker bytes inside an unavailable profile name, and assert only the real PID receives signals. Clean lift: move ownership data out of mixed stdout/stderr into a dedicated one-shot PID/result file or pipe. Treat `EPERM` as “not owned / do not signal,” not an unhandled teardown error.

**[P3] Per-launch logs still accumulate after abnormal launcher death.**

`apps/desktop/scripts/dev-app-runtime.mjs:464` removes the UUID log only from normal `stop()`. SIGKILL, crash, or power loss leaves full-session output under `tmpdir()/maka-dev-launches` until OS cleanup. Add bounded startup scavenging; a launcher-PID-owned cleanup record is cleaner than age alone.

Fresh gate before publication: head unchanged, OPEN/non-draft/MERGEABLE/CLEAN; exact-head `audit`, `test`, and `package` are all `completed/success`. GitHub's aggregate APPROVED comes from `hqhq1025@c17e155f`, not this head, and does not override the live P2 hold. No merge action was taken. This Agent review does not claim to satisfy the unresolved ASF required-human-review wording.

Astro-Han:
@M4n5ter @jackwener

Addressed at exact head `221985ae068720c7b1235687bc0f93b019edeae0`.

We did not tighten the mixed-log marker parser. The marker protocol has been removed entirely.

Electron remains the sole single-instance authority and now publishes exactly one structured `winner + PID` or `loser` verdict to this invocation’s private result file. The file is created with exclusive-write semantics; stdout/stderr remains a separate observation-only log. The detached TCC launcher waits for that result, monitors only the reported winner PID, and sends no signal while ownership is unknown.

The previous `DEV_WINNER_LOG_MARKER`, `DEV_LOSER_LOG_MARKER`, whole-log scanning, and log-driven monitor branches are deleted. The regression reproduces ordinary application output containing a later marker-shaped PID and proves that only the PID from the dedicated result is signalled.

Verification on this exact head: local Desktop `test:dist` 1284/1284, Desktop typecheck and Biome passed; hosted `audit`, `test`, and `package` are all completed/success. AI-assisted implementation; please re-review the exact head.

apps/desktop/scripts/dev-app-runtime.mjs
Astro-Han: Resolved by deletion at `221985ae0`. The broad Electron process scan and fallback attribution to the Maka Dev profile are gone; unrelated Electron processes are no longer classified as profile owners.

apps/desktop/scripts/dev-app-runtime.mjs
Astro-Han: Resolved by deletion at `221985ae0`. The command-line ownership probe and `worktreeFromCommandLine` no longer exist, so no path tokenization participates in launch ownership.

apps/desktop/scripts/dev-app-runtime.mjs
Astro-Han: Resolved at `221985ae0`. Each TCC launch now has separate private log and result files. Electron writes the winner/loser verdict once to the result file; stdout/stderr is observation-only and cannot be consumed by another launch as authority.

apps/desktop/scripts/dev-app-runtime.mjs
Astro-Han: Resolved by deletion at `221985ae0`. Exit code 0 is no longer reinterpreted through a process-owner scan; TCC lifecycle comes from the private winner/loser result and exact winner PID.

Astro-Han:
Re-reviewed at exact head `221985ae068720c7b1235687bc0f93b019edeae0`. Exact-head `audit`, `test`, and `package` are all `completed/success`.

**The previously-raised P2 is genuinely closed, and closed well.** `DEV_WINNER_LOG_MARKER` / `DEV_LOSER_LOG_MARKER`, the whole-log scan, and the log-driven lifecycle branches are all gone. The Electron lock now writes `winner+pid` or `loser` exactly once to a result file private to that invocation, with `wx` making the first writer unique; stdout and stderr are observed but no longer carry authority. Removing the mixed log as a second authority is a real simplification, and the regression that plants a fake `222` marker in ordinary log text while the legitimate winner is `111` pins it properly.

One new finding on the cancellation path.

### `[P2]` Cancelling before the verdict exists can orphan an app that later wins the lock — `apps/desktop/scripts/start-dev-app.mjs:23-40`, `apps/desktop/scripts/dev-app-runtime.mjs:465-473`

The signal handlers are deliberately registered before `await startDevelopmentApp()`, and the comment explains exactly why — that part is right. The gap is that `stop()` is one-shot:

```js
async function stop(code = 0, options = {}) {
  if (stopping) return;
  stopping = true;
  await app?.stop({ terminateApp: options.terminateApp });
```

**Path A — signal arrives before the handle exists.** If prepare/codesign/handoff hasn't returned, `app` is still `null`. `stop()` sets `stopping = true` and awaits nothing, because there is no handle to stop. The pending `startDevelopmentApp()` then completes normally, spawns `tail` and `open`, and returns a live app. Any later stop returns immediately at the `stopping` guard, so that launch is never adopted or terminated. In the `dev.mjs` path the Vite server has already been closed by then, so what survives is a blank-window app holding the shared `Maka Dev` profile lock — and the lingering `tail` can keep the launcher itself from exiting.

**Path B — signal arrives after spawn but before the result is written.** `app.stop()` calls `stopDevelopmentLaunch()`, which reads the winner pid, finds nothing, and returns `false` without signalling. Control then reaches `rmSync(launch.resultFile, { force: true })`. If Electron wins the lock immediately afterwards it recreates its own result file and keeps running, and no second stop is ever attempted.

**Why the existing guard doesn't cover this.** The test `an unclassified TCC invocation cannot terminate another lock owner` proves the right thing — an unknown invocation must not signal anyone. It does not establish that the *same* unknown invocation won't become the winner a moment later, which is the case here. The `stopping` boolean is what prevents the compensating cleanup once the verdict finally arrives.

**Suggested shape:** make cancellation a durable request on the launch handle rather than a single attempt. If the signal precedes the handle, the handle must execute the cancellation as soon as it returns; if the signal lands after spawn but before the verdict, don't delete the result file first — wait for this invocation's one-shot verdict, then terminate only that PID when it's the winner and signal nobody when it's a loser. That reuses the result authority this PR just established, and it does not mean going back to a bundle-wide `pkill` or to log parsing. A production-seam regression for "stop with no result, delayed winner afterwards" would pin it.

If there is a structural convergence worth doing now, it's letting start/stop/cancel share one launch-handle lifecycle — that's the same change, and I found no second independent simplification worth raising.

### Validation

Ran: workspace dependency build; Desktop `build:main`, four typecheck passes, Biome over 9 files; the dedicated result/launcher/core tests 22/22; Desktop `test:dist` 1284/1284; `git diff --check` and a clean merge-tree against current `main`.

Not run: this was a Linux host, so there was no real macOS LaunchServices/TCC hand-launch. The finding above comes from running the production functions directly plus static control flow across both entry points — it is not claimed as a reproduction on a real machine.

Astro-Han:
Two `[P3]` addenda at the same head `221985ae068720c7b1235687bc0f93b019edeae0`, from a second reviewer working independently of the earlier comment. Neither blocks.

### `[P3]` Winner identity is a bare PID, so a reused PID can be followed or killed — `apps/desktop/scripts/dev-app-runtime.mjs:240-252, 497-531`

The result file records `{winner, pid}`, and everything downstream trusts the number alone. `monitorDevelopmentApp` polls liveness with `kill(pid, 0)`, and `quitMacosDevelopmentApp` sends `TERM`, waits a fixed 3 s, then sends `KILL`:

```js
if (!signal('TERM', pid)) return false;
await delay(graceMs);
signal('KILL', pid);
```

There is no identity re-check between those two signals. If Electron exits inside the polling interval or the grace window and the OS reuses that PID, the launcher will either follow an unrelated process as if it were the app, or `KILL` it.

This needs dev-only conditions plus fast PID reuse, hence P3. The clean minimal fix is to make liveness and signalling verify identity rather than existence: before each call, confirm the process's argv still contains this invocation's unique `--maka-dev-launch-result-file=<path>`, or record an equivalent OS process-birth identity. A regression covering "PID reused after TERM" would pin it.

### `[P3]` No scavenger for the per-invocation log and result files

The UUID log/result pair is removed only in `startDevelopmentApp().stop()` (`:466-472`). A launcher that dies to `SIGKILL`, a crash, or power loss leaves both files behind, with OS temp cleanup as the only fallback. Same finding as before, still P3.

### Note on the earlier `[P2]`

This reviewer's pass concluded GO with no P0–P2, but it examined the marker/result authority rather than the cancellation path, so it does not bear on the `[P2]` raised in the preceding comment about cancelling before the verdict exists. That finding was verified independently against the source and still stands — treat the GO as scoped to the authority question.

Independent validation at this head: `@maka/core` 638/638, Desktop main 1265/1265, main-result plus launcher focused 20/20. One note for anyone reproducing: an initial `npm ci --ignore-scripts` skips the three dependency patches and fails the UI prebuild; a normal `npm ci` applies them and everything passes. That first failure was environmental, not a code regression.

Astro-Han:
Automated follow-up from Codex on behalf of AstroHan, at exact head `93695090fef9f6f6d50b049788199b4306636442`.

The cancellation P2 and bare-PID P3 shared one architectural cause: the detached TCC launcher was still treated as an owner of the Electron process. This head removes that ownership instead of adding another cancellation state machine.

- A TCC winner verdict is now only `{ "status": "winner" }`; it carries no PID.
- `developmentLaunchWinnerPid`, `stopDevelopmentLaunch`, `quitMacosDevelopmentApp`, PID liveness polling, and TERM/KILL teardown are deleted.
- After the one-shot lock verdict, the launcher stops observing Electron. Ctrl-C cleans up only launcher-owned resources (the log follower, result/log files, and Vite server); the TCC app remains open until the user quits it with Cmd-Q.
- The pre-handle signal guard in `start-dev-app.mjs` is no longer needed and is deleted, so the reported Path A one-shot cancellation state also disappears.

This closes the failure paths in [the cancellation P2](https://github.com/apache/maka/pull/3359#issuecomment-5385946595) and [the PID-reuse P3](https://github.com/apache/maka/pull/3359#issuecomment-5385957752) by removing their preconditions. The temp-file scavenging P3 remains a nonblocking, separate observation; this change does not add a scavenger state machine for it.

Local verification on this head: Core 639/639, Desktop 1265/1265, launcher 16/16, Desktop typecheck, and repository format check. Hosted CI is still running; this note does not claim it is green.

jackwener (commented):
Technical review of exact head `93695090fef9f6f6d50b049788199b4306636442`, acting on behalf of @WAWQAQ under the standing Kabi review/publisher authorization.

**Conclusion: COMMENT — one P2 remains, plus the acknowledged nonblocking P3. No APPROVE and no merge.**

The two requested historical candidates are mostly closed. `start-dev-app.mjs` now installs signal handlers only after its launch handle exists, and the detached TCC path has deleted winner PID reporting, PID liveness polling, and the fixed TERM→KILL sequence entirely. The private one-shot result file is the sole lock-verdict channel; ordinary logs carry no authority.

### [P2] `npm run dev` can still create launcher resources and Electron after cancellation

`apps/desktop/scripts/dev.mjs:158-177` retains the old one-shot shape: `shutdown()` sets `shuttingDown = true`, calls `app?.stop()` while `app` may still be `null`, then closes Vite; the concurrent `app = await startDevelopmentApp(...)` is not cancelled or joined. `startDevelopmentApp()` has real asynchronous boundaries before it returns the handle (`apps/desktop/scripts/dev-app-runtime.mjs:100-108,376-416`): both launch modes lazily import the Core contract, and the TCC mode prepares the bundle first. After those awaits it can still spawn plain Electron, or `tail` plus `open`, and only then return the cleanup handle.

Reachable path: run `npm run dev`, deliver SIGINT while the lazy contract import/start preparation is pending, and let that start finish before Vite's asynchronous close finishes. The first `shutdown()` skips `app.stop()` because `app === null`; the launch then creates its child/resources and assigns `app`, but every later cleanup request returns at `if (shuttingDown) return`. The process can therefore start Electron after the user cancelled, and the TCC path can leave the log follower plus launch files uncleaned; its app is pointed at the Vite server that cancellation just closed. The plain child is likewise never terminated by the launcher.

The current guard does not cover this: moving handlers below the await fixed `npm start`, not `npm run dev`, and the 16 launcher tests exercise exported helpers but not either top-level signal coupler. Join cancellation to the in-flight launch (or pass an AbortSignal) so shutdown awaits/adopts the eventual handle and runs `stop()` exactly once before closing/exiting. Add a seam test with a deferred `startDevelopmentApp()` that resolves after the signal and assert its returned handle is stopped.

### [P3] Abnormal launcher death still leaves per-launch output

`apps/desktop/scripts/dev-app-runtime.mjs:176-185,430-433` creates UUID log/result paths and removes them only through normal handle cleanup. SIGKILL, crash, or power loss still leaves full-session logs under the temp directory until OS cleanup. This remains nonblocking; bounded startup scavenging would close it.

Validation: Core 639/639, Desktop main 1265/1265, launcher 16/16; worktree clean. Fresh machine state before publication: OPEN, non-draft, MERGEABLE/CLEAN; exact-head `audit`, `test`, and `package` are all `completed/success`. GitHub's aggregate APPROVED is only stale old-head history, not a current-head GO.

Astro-Han:
Automated architecture decision note from Codex on behalf of AstroHan, at exact head `93695090fef9f6f6d50b049788199b4306636442`.

The repeated P3 about abnormal-death leftovers is accepted and deliberately deferred for this PR.

Current scope and impact:

- It is reachable only in the opt-in macOS TCC development path after SIGKILL, process crash, or power loss bypasses normal cleanup.
- The launch directory is created with mode `0700`, and launch files use `0600`; they are local developer artifacts under the OS temporary directory.
- Normal launcher shutdown removes the files. Abnormal shutdown relies on the OS temporary-file lifecycle and may retain full-session output longer than desired.

Why this PR does not add a scavenger:

- A safe scavenger must distinguish an active launch from an abandoned one before deleting its live log or one-shot verdict.
- Age alone can delete a valid long-running session. Launcher-PID ownership or a lease recreates the process-identity and lifecycle state that this head deliberately removed. A more elaborate lease/registry would cost more authority and failure modes than this dev-only residual currently justifies.
- This PR's architectural goal is to leave Electron as the sole app-lifecycle owner and keep the launcher contract one-shot. Adding another persistent cleanup authority would work against that goal.

Therefore this remains a nonblocking P3, not an unimplemented merge requirement. It should be reconsidered if there is evidence of material disk growth, sensitive log retention outside the existing private permissions, or inadequate OS cleanup in supported development environments. Unless one of those premises changes, repeating the same observation on later heads will not change the decision.

Separately, the new `npm run dev` cancellation P2 is valid and is not covered by this deferral; its fix belongs at the launch-attempt owner and must guarantee that cancellation becomes observable before any later `spawn`.

jackwener (commented):
Technical incremental review of exact head `dbd84a6ac7d2a9263890f61e5c1abeca31d5cdec`, acting on behalf of @WAWQAQ under the standing Kabi review/publisher authorization.

**Conclusion: COMMENT — one P2 remains. No APPROVE and no merge.**

The runtime fix itself closes the prior cancellation P2: `apps/desktop/scripts/dev.mjs:158-185` makes the in-flight launch promise the ownership boundary, aborts before joining it, and passes the same signal into `startDevelopmentApp()`. `apps/desktop/scripts/dev-app-runtime.mjs:370-428` checks that signal after each real asynchronous preparation boundary and before any spawn. The two TCC spawns have no await between them, and once the handle is published `shutdown()` joins that same promise before calling `stop()`. I found no reachable post-spawn orphan window in this implementation.

### [P2] The new regression tests do not cross the top-level signal coupling they are meant to protect

`apps/desktop/scripts/dev-app-runtime.test.mjs:22-70` calls `startDevelopmentApp({ signal })` directly. It proves the helper honors a signal already supplied by its caller, but it never exercises `dev.mjs` creating the controller, passing `launchController.signal`, receiving a real signal, joining the launch promise, and stopping the eventual handle.

This is load-bearing rather than merely missing extra coverage: deleting the single `signal: launchController.signal` line at `apps/desktop/scripts/dev.mjs:185` reintroduces the exact old failure — shutdown closes Vite while the uncancelled helper can later spawn Electron — yet both newly-added cancellation tests still pass 2/2. The tests therefore cannot distinguish this fixed head from the broken wiring they were added to prevent.

Please add a regression that crosses the `dev.mjs` top-level coupler with a deferred launch and real signal timing, asserting that cancellation prevents any later spawn and that an already-published handle is stopped exactly once. The mutation criterion is simple: remove the `signal:` wiring and that regression must fail.

Local validation: launcher/runtime suite 18/18 and Core build passed; the two new tests remained 2/2 green under the wiring-deletion mutation, then the worktree was restored clean. A full Desktop pretest on this checkout was blocked by untouched UI build/type skew after `npm ci --ignore-scripts`; this does not affect the focused result. Exact-head hosted `audit`, `test`, and `package` are all `completed/success`; fresh state is OPEN, non-draft, MERGEABLE/CLEAN. GitHub's aggregate APPROVED is stale old-head history and does not override this live P2. The previously accepted abnormal-death temp-file P3 remains nonblocking and unchanged.

jackwener (approved):
Incremental technical re-review of exact head `19061b50c0a03901c8565539961db61ae250bf27`, executed by the Kabi review agent on behalf of @WAWQAQ under the standing Kabi review/publisher authorization.

**GO — no live P0–P2; the previously accepted abnormal-death temp-file P3 remains unchanged. APPROVE; no merge action.**

The real increment from reviewed head `dbd84a6ac` is 3 files (`+126/-38`). It moves the top-level AbortController, launch-promise ownership, signal handlers, join/adopt/stop sequence, server close, and exit latch from `dev.mjs` into `createDevelopmentLaunchSession()` without changing the production ownership order.

The old P2 test gap is closed with mutation evidence. Baseline launcher/core/main focused tests passed 24/24. Deleting only `signal: controller.signal` from the new production coupler made the launcher suite 19/20: the sole failure was `a launcher signal cancels preparation before Electron can spawn`, and execution crossed the deferred preparation boundary into the forbidden launch-file creation. The regression therefore guards the exact top-level coupling that was previously untested.

The shared `Maka Dev` profile and Electron single-instance contract were already present on the reviewed old head and are unchanged in this increment. Current loser paths exit before store or Runtime Host boot: the plain child exit and the TCC private one-shot verdict each close only their launcher/Vite resources, while the winning Electron process remains the sole shared-profile lock owner. No new cross-worktree write or lock competition was found beyond the documented existing Chromium timeout and stale per-worktree dev-env limitations.

Three discussion endpoints were reconciled after provisional sealing; no current-head P0–P2 remains. Fresh action gate: OPEN, non-draft, MERGEABLE/CLEAN; exact-head hosted `audit`, `test`, and `package` are all `completed/success`. This Agent-executed approval does not claim to satisfy any separate unresolved human-review wording.

PR #3364, opened 2026-08-21
fix: keep relative timestamps on just now for the first minute

Astro-Han:
## Summary

Relative-time suffixes under one minute were counting seconds (`1秒钟前`, `2秒钟前`, …), which made sidebar task rows and other `RelativeTime` surfaces flicker. Keep a single just-now label (`刚刚` / `just now`) for the whole first minute, then switch to minute buckets, and delay the ticker until that boundary.

## Verification

- `packages/core` build + `node --test dist/__tests__/relative-time.test.js` (pass)
- `biome format` / `biome lint` on the touched files (pass)
- Did not run repository-wide tests; leave full coverage to CI

## AI use

Select exactly one:

Tool(s) and scope:

Maka implemented the formatter/ticker change, tests, and this PR.

## Checklist

- [ ] No

### DIFF
--- a/packages/core/src/relative-time.ts
+++ b/packages/core/src/relative-time.ts
@@ -18,24 +18,11 @@
  */
 
 /**
- * PR-RELATIVE-TIME-0: a single, locale-aware relative-time formatter
- * shared by every Maka surface. The existing `formatRelativeTimestamp`
- * sat inside `packages/ui/src/components.tsx` as a private helper,
- * which meant it could not be unit-tested and could not be reused by
- * sidebar / settings panels.
- *
- * The formatter is pure (takes an optional `now` so tests do not have
- * to monkey-patch `Date.now`). The first minute stays on a single
- * just-now label so scan-level rows do not tick through "1秒钟前",
- * "2秒钟前", …. After that the buckets widen from minute to hour to
- * day. Anything older than ~7 days falls back to an absolute date so
- * we do not produce misleadingly round numbers like "5 个月前".
- *
- * The threshold is a deliberate divergence from the previous
- * implementation, which used `.format(-Math.round(diffHours / 24), 'day')`
- * for ALL timestamps older than a day and produced things like
- * "300 天前" for messages a year old. That was strictly less useful than
- * the locale date string.
+ * Locale-aware relative-time formatter shared across Maka surfaces. Pure
+ * (optional `now`) so tests can pin a clock. The first minute stays on one
+ * just-now label instead of counting seconds, then buckets widen from minute
+ * to hour to day; past ~7 days we fall back to an absolute date, which is more
+ * useful than a relative label like "300 天前".
  */
 
 import { uiLocaleToIntlLocale, type UiCatalog, type UiLocale } from './ui-locale.js';
@@ -79,12 +66,9 @@ function getAbsoluteFormat(uiLocale: UiLocale): Intl.DateTimeFormat {
 }
 
 /**
- * Returns a localized relative label for `ts` (e.g. "刚刚", "1 分钟前")
- * when within `RELATIVE_HORIZON_MS`, otherwise the absolute date string.
- *
- * `now` is injectable so tests can pin a deterministic clock. The first
- * minute — and future timestamps from clock skew — stay on one just-now
- * label so rows do not count seconds or show "in 2 minutes".
+ * Localized relative label for `ts` within the 7-day horizon, otherwise the
+ * absolute date string. `now` is injectable so tests pin a deterministic clock;
+ * future timestamps (clock skew) snap to the just-now label.
  */
 export function formatRelativeTimestamp(
   ts: number,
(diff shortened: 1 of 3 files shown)


jackwener (commented):
Automated review of exact head `b2a51c66b208b2da10a690d51c641deaf62ff71f` against current `main@d62857a8357e9160926726a2a13096bc2dc2b91d`.

The defect remains on main: sub-minute labels count seconds and `RelativeTime` schedules a render every second. This revision puts the first-minute boundary in the shared pure formatter, reuses it for compact and wide labels, and schedules exactly one timer at the boundary instead of adding a UI-local special case. Boundary coverage at 0/59,999/60,000ms would fail under main. I found no actionable P0-P2 defect; required `test` is green.

Required conclusions:

1. Optimal for the actual problem: **yes**; one shared `JUST_NOW_MS` authority drives both formatting and refresh cadence.
2. Production code to delete: none beyond the superseded second-level branch/comments already removed.
3. Tests to delete/replace: none identified; the two focused tests are behavioral and protect the original defect.
4. Deeper refactor: no.
5. Ready to merge: **code-wise yes**, but the PR is still Draft and this automated review is not approval.
6. Residual risks/gaps: no rendered React timer test, though the pure delay contract covers the load-bearing behavior; independent human review is still required for this user-visible behavior change.

jackwener (dismissed):
Independent review of exact head `b2a51c66b208b2da10a690d51c641deaf62ff71f`.

No P0–P3. First-minute just-now is a single `JUST_NOW_MS` authority for both labels and ticker delay. Boundaries 0 / 59,999 / 60,000 ms and future/clock-skew (`diffMs < JUST_NOW_MS`) hold.

Authorization: Raft #PR-Review-Kabi:8bb877df via dm:@Opus-Qronos-AstroHan:c1beef44 (AstroHan-delegated). Approve only; not merging.

jackwener (commented):
Independent review of `b2a51c66b`.

1) Problem. Sub-minute labels counted seconds, so sidebar rows flickered every second.

2) Solution. One just-now label (`刚刚` / `just now`) for the first 60s, then minute buckets. Ticker waits until that window ends (`JUST_NOW_MS - diffMs`) instead of ticking every second. Compact path uses the same formatter.

3) Occam. This is the right cut. Tests pin the first minute and the delay.

Merge caveat after #3397: `packages/core/src/__tests__/relative-time.test.ts` is a new file with no ASF header. The listed `test` SUCCESS is from 2026-08-21, before that gate. Rebase onto current `main` and run `npm run write:asf-headers` before merge, or `check:asf-headers` will fail.

Not requesting changes on the behavior.

jackwener:
Correction on the header caveat: only the **new** test file `packages/core/src/__tests__/relative-time.test.ts` lacks an ASF header.

`relative-time.ts` and `relative-time.tsx` already have headers on current `main` (added by #3397). This branch is from before that. On rebase, keep main's header and put the new JSDoc under it — do not overwrite the header with the branch's first-line JSDoc.

jackwener (dismissed):
Re-review of `e65809320` after the header fix.

Behavior still holds: first minute is one just-now label; ticker waits for that boundary. The new test file now has the ASF header. Required `test` on this head started 18:42Z, after #3397's header gate, and is SUCCESS.

Approve. Merge stays with the author.

packages/core/src/relative-time.ts
```diff
@@ -174,15 +154,14 @@ export function resetRelativeTimeFormatters(): void {
...
- * every 10 minutes after that. Past the horizon we never re-render.
+ * Next tick delay (ms) for the `<RelativeTime>` ticker: once when the just-now
+ * window ends, every minute for the first hour, then every 10 minutes; null
+ * past the horizon (never re-render).
  */
 export function nextRelativeRefreshDelay(ts: number, now: number = Date.now()): number | null {
   const diffMs = now - ts;
   if (diffMs > RELATIVE_HORIZON_MS) return null;
-  if (diffMs < 60_000) return 1_000;
+  if (diffMs < JUST_NOW_MS) return JUST_NOW_MS - diffMs;
```
jackwener: `JUST_NOW_MS - diffMs` assumes `diffMs >= 0`. Future `ts` makes this unbounded and can overflow Chromium's 32-bit timer.

Minimal/clean fix is the same line: `const diffMs = Math.max(0, now - ts)` so age is a non-negative duration. Then this branch returns `JUST_NOW_MS` for future timestamps, which matches the documented just-now display and cannot overflow.

jackwener (commented):
Withdrawing the previous APPROVE on this head.

**P1:** `nextRelativeRefreshDelay` computes `JUST_NOW_MS - diffMs` for every `diffMs < JUST_NOW_MS`. That formula is only valid for `0 <= diffMs < 60s`. Future timestamps (`diffMs < 0`) — already documented as clock skew → just-now — produce an unbounded delay. Around 25 days ahead it exceeds INT32_MAX; Chromium/Node `setTimeout` overflow-fires in ~1ms. `<RelativeTime>`'s effect has no deps and reschedules every render, so that becomes a tight loop.

This is a regression from `main`, which used a constant `1000` in the same branch (1Hz tick, no overflow).

Do not clamp at `setTimeout`. Age should never be negative:

```ts
const diffMs = Math.max(0, now - ts);
```

Future records stay on the just-now label and refresh on the existing 60s cadence until the clock catches up. That also survives clock correction; a 24-day timer would not.

Please add tests for `ts > now` and `ts = now + 30d` asserting the delay stays in the cadence table, not `JUST_NOW_MS - diffMs`.

jackwener (approved):
Independent re-review of a59cbb00.

The future-timestamp timer overflow is closed: one private relativeAgeMs() = Math.max(0, now - ts) feeds all three consumers. Delay for +30d is 60s, not an unbounded remaining window. Required test on this head is SUCCESS. Non-blocking: the new test name still says "finite" while the sample includes NaN/±Infinity.

Approve. Merge stays with the author.

PR #3370, opened 2026-08-21
feat(provider): add Amazon Bedrock IAM Identity Center SSO

keyuchen21:
## Summary

- register `amazon-bedrock` as a first-class provider with typed IAM Identity Center configuration and a dedicated `aws_sso` credential kind
- add a Host-owned `bedrock.sso.*` device-authorization lifecycle with account/role selection, bounded model discovery, manual model validation, and atomic onboarding recovery
- resolve and CAS-refresh SSO sessions in Runtime Host, while keeping temporary AWS role credentials memory-only and singleflight refreshed
- add Bedrock Converse/ConverseStream execution through `@ai-sdk/amazon-bedrock`, including auxiliary model calls, connection tests, inference-profile capability inheritance, and source-model pricing
- add the Desktop account setup flow and make the resulting connection available to Desktop, CLI, TUI, bots, and other Runtime Host clients
- route AWS SDK control-plane traffic through Maka's scoped fetch/proxy boundary and add permission/configuration error classification

## Security properties

- OIDC client secrets, access/refresh tokens, device codes, and temporary role credentials never cross the Runtime Host protocol into renderer state
- temporary AWS role credentials are never persisted
- uncommitted login attempts are cancelled and cleared after a bounded TTL
- credential refresh uses canonical generation binding and CAS reconciliation, so logout/reauth cannot revive an old credential generation
- Bedrock connections reject custom request headers and body overlays to avoid post-signature mutation
- Desktop configuration exports omit IAM Identity Center sessions and require reauthorization after import

## Validation

- `npm run build`
- `npm run lint`
- Core, Storage, Runtime, Runtime Host, and Desktop test suites
- Bedrock contract, protocol-secret-boundary, SSO session, provider conformance, pricing, and error-classification tests
- Desktop and CLI third-party notice checks
- live IAM Identity Center device authorization against a test tenant
- live Bedrock foundation-model and inference-profile discovery through the new Smithy-to-scoped-fetch handler
- live `ConverseStream` request with tool call, tool result replay, and final text

`npm test` completed every JavaScript/workspace suite, but the local `@maka/eval` Python continuation stopped because the machine has Python 3.9 and the existing test uses Python 3.10 union syntax (`int | None`). Its 75 Node tests passed before that environment-only failure.


Astro-Han:

- `apps/desktop/src/main/runtime-host-client.ts`
- `apps/desktop/src/preload/bridge-contract.d.ts`
- `apps/desktop/src/renderer/settings/provider-connection-detail.tsx`
- `…(more)`

```
# resolve, then
git push --force-with-lease
```

---

Astro-Han (commented):
Independent review at exact head `c87e304c41c233e75cdca9086a4cee0064d0f7e5`. Thanks for taking on Bedrock SSO — it's a genuinely wanted provider path, and the Host-side credential handling here is careful work. Three `[P1]`s stand between it and landing, and the first one is the important one.

---

### `[P1]` The SSO sign-in flow cannot complete — `apps/desktop/src/renderer/settings/bedrock-sso-setup.tsx:31-35`

```tsx
const mounted = useRef(true);

useEffect(() => () => {
  mounted.current = false;
  if (attemptId) void window.maka.amazonBedrockSso.cancel(attemptId, host).catch(() => undefined);
}, [attemptId, host]);
```

A cleanup-only effect with dependencies runs its cleanup on **every dependency change**, not only on unmount. `login()` calls `setAttemptId(started.attemptId)` at `:42`, which changes a dependency, so the cleanup fires mid-sign-in and sets `mounted.current = false`. The `if (attemptId)` guard covers only the `cancel` call — the sentinel is cleared unconditionally. Nothing ever restores it, because the effect body is just `() => cleanup`.

The polling loop at `:45` is `while (mounted.current && ...)`, and `:50` is `if (!mounted.current) return`. So the first sign-in — necessarily the render where `attemptId` goes from undefined to a value — always returns early and never reaches accounts or roles. This isn't a race; it's unconditional.

Suggested shape: separate the mount-lifetime effect (empty deps, owns the sentinel) from the attempt-cancellation effect (deps on `attemptId`/`host`, owns only cancel), so changing the attempt id can't retire the component's liveness flag.

Worth noting alongside it: there is no renderer behaviour test over `BedrockSsoSetup`, so nothing in 69 files and +21,862 lines could have caught this. A test that drives `login()` through a fake bridge and asserts roles appear would pin the whole flow.

### `[P1]` Ambient `AWS_BEARER_TOKEN_BEDROCK` silently overrides the configured SSO identity

`@ai-sdk/amazon-bedrock`, pinned to 5.0.58 by this PR's lockfile:

```js
const rawApiKey = loadOptionalSetting({
  settingValue: options.apiKey,
  environmentVariableName: "AWS_BEARER_TOKEN_BEDROCK"
});
const apiKey = rawApiKey && rawApiKey.trim().length > 0 ? rawApiKey.trim() : void 0;
const fetchFunction = apiKey
  ? createApiKeyFetchFunction(apiKey, options.fetch)
  : createSigV4FetchFunction(async () => { /* ... options.credentialProvider ... */ });
```

`credentialProvider` lives inside the SigV4 branch. `model-factory.ts:99-103` passes it correctly, but with a non-empty ambient bearer token that branch is never taken and the provider is never constructed, let alone called. A fake-only probe confirmed Bearer auth on the wire with zero calls into the explicit SSO provider.

The harm is **wrong identity, not failed auth**: a user who configured SSO gets inference, connection tests and manual probes running as whoever the stale environment token belongs to, and Settings still reports the connection as authenticated because that display is derived from the vault entry alone. Nothing on the running path discloses which credential source is actually in use.

One trap when fixing: passing `apiKey: undefined` does **not** suppress the lookup — `loadOptionalSetting` falls through to the environment variable exactly when the setting value is absent. Suppressing it requires an explicit empty string, or an upstream option that genuinely disables the env bearer.

### `[P1]` Cancel only aborts device authorization, not what follows it

The cancel path aborts the device-authorization polling loop, but the post-auth calls — `listAccounts`, `listRoles`, model discovery, the manual Converse probe — take no abort signal and continue after a cancel returns. The Cancel button also remains clickable while `busy`, so a user who cancels can still have credentials in use afterwards, including model probes that cost money. Host close doesn't wait on these either.

---

**Credential lifecycle — no findings.** Reporting this explicitly because it is the part most likely to be assumed rather than checked: the SSO session lands in the Host workspace's `runtime-policy-onboarding.json` crash journal and finally in `credential-vault.json`, written atomically at `0600`; role credentials stay in Host memory with a 5-minute skew refresh and never touch disk or the protocol. No token or role-key exposure was found in logs or protocol frames within the reviewed surface. With no ambient bearer present, the explicit provider is invoked and refresh failures classify correctly as Auth — so there is no access-key or default-chain fallback defect.

**Deletion axis — no closed candidate.** +21,862 lines invites the question of how much duplicates the existing OAuth credential path. Merging the two generation/CAS authorities is a *probe*, not a recommendation: AWS's two-level session→role expiry and OAuth's uncertain-commit semantics are different enough that a shared owner might add branching rather than remove it. Not claiming a net deletion without proving that.

**Validation limits.** Core→Storage→MCP→Runtime→Runtime Host clean build; 17/17 targeted tests; two fake-only probes. `astryx:surface-inventory` reproduces a required test failure on this head (the new setup file is not in the inventory). Full repo suite, renderer/Electron E2E and Windows packaging were **not** run, and **no real AWS credentials or live AWS calls were used at any point** — every probe used fakes and placeholder values.

packages/runtime-host/src/server/bedrock-sso-coordinator.ts
```diff
@@ -0,0 +1,459 @@
...
+            ssoRegion: attempt.ssoRegion,
+            region: attempt.region,
+            accountId: attempt.accountId!,
+            roleName: attempt.roleName!,
+          },
+        }),
+      );
+      if (committed.kind !== 'committed') {
+        return failure('persistence_failed', 'Amazon Bedrock connection could not be committed');
+      }
```
Astro-Han: [P1] A durable commit can be reported to the client as a persistence failure, and activation is left unpoisoned.

`commitConnectionOnboarding` returns `committed` — the connection is durably persisted. The very next statement is `await this.invalidateBackends()`, and the enclosing `catch` maps *any* throw to `failure('persistence_failed', …)`. If invalidation throws:

- the client is told persistence failed while the connection is in fact persisted;
- `this.#committed.set(attemptId, result)` is never reached, so there is no replay record for the retry to find;
- `#attempts` still holds the attempt;
- activation is **not** poisoned, so a later Turn can proceed against backends whose invalidation never completed.

This file's own sibling already encodes the correct rule. `runtime-policy-coordinator.ts:282-299` wraps the same shape and comments it explicitly:

> The durable outcome is authoritative, but no later Turn may activate against a backend whose invalidation did not complete.

…followed by `this.activation.poison()`. That is the invariant this path drops.

Suggest preserving the durable outcome (record the replay result before invalidation, or report success-with-degraded-activation) and poisoning or fataling activation when invalidation fails, matching the policy coordinator.

<details>
<summary>中文</summary>

`commitConnectionOnboarding` 返回 `committed` 时连接已经持久化，紧接着的 `await this.invalidateBackends()` 若抛出，外层 `catch` 会把它一律映射成 `persistence_failed`。后果有四点：客户端被告知持久化失败，而数据其实已经落盘；`#committed.set(attemptId, result)` 根本没执行到，重试找不到 replay 记录；`#attempts` 仍留着该 attempt；并且 activation **没有**被 poison，后续 Turn 可以在"失效流程未完成"的 backend 上继续跑。

同一仓库里的 `runtime-policy-coordinator.ts:282-299` 已经把正确规则写下来了——注释明确说"持久结果是权威的，但任何后续 Turn 都不得在失效流程未完成的 backend 上激活"，随后调用 `this.activation.poison()`。这条路径丢掉的正是这个不变量。

建议：保住持久结果（在失效之前先记录 replay 结果，或返回"成功但激活降级"），并在失效失败时按 policy coordinator 的做法 poison 或 fatal 掉 activation。
</details>

packages/runtime-host/src/server/bedrock-sso-coordinator.ts
```diff
@@ -0,0 +1,459 @@
...
+        initiatingConnectionId,
+        OAUTH_PRESENTATION_SERVICE_ID,
+        OAUTH_PRESENTATION_SERVICE_VERSION,
+      )
+    ) {
+      return failure('operation_unavailable', 'Desktop cannot present IAM Identity Center login');
+    }
+    for (const previous of this.#attempts.values()) {
+      this.#cancelAttempt(previous);
+      await previous.settlement.catch(() => undefined);
```
Astro-Han, line 162: [P2] `#start` has no admission gate, so a concurrent start can orphan an in-flight AWS device flow.

The only early return is `this.#attempts.get(input.attemptId)`, which two starts carrying *different* attempt ids both pass. Each then cancels the other's attempts and awaits `previous.settlement`.

The problem is when `settlement` becomes real: it is initialised to `Promise.resolve()` at `:137` and only replaced with the actual `#poll` promise at `:168`. The awaited AWS device-authorization start sits between those two points, so a second start arriving in that window awaits an already-resolved promise, clears `#attempts`, and proceeds. The first flow then completes, opens a browser URL and begins polling with no attempt record owning it.

Suggest serialising admissions through a lease and awaiting the genuinely in-flight start before superseding or clearing.

<details>
<summary>中文</summary>

`#start` 唯一的提前返回是 `this.#attempts.get(input.attemptId)`，两个携带**不同** attemptId 的并发 start 都能通过，随后各自取消对方的 attempt 并 await `previous.settlement`。

关键在于 `settlement` 何时变成真的：它在 `:137` 被初始化为 `Promise.resolve()`，直到 `:168` 才被替换成真正的 `#poll` promise，而那段被 await 的 AWS device authorization 启动过程正好位于两者之间。因此落在这个窗口里的第二个 start 会 await 一个已经 resolve 的 promise，清空 `#attempts` 然后继续；第一个流程随后完成、打开浏览器 URL 并开始轮询，却已经没有任何 attempt 记录持有它，成为孤儿。

建议用 lease 把 admission 串行化，并在顶替或清空之前 await 真正在飞的那个 start。
</details>

Astro-Han (commented):
Review at `c87e304c`. One **[P1]** and one **[P2]**, both inline on `bedrock-sso-coordinator.ts`. Holding.

Both are lifecycle rather than cryptography: the login flow's happy path reads fine, but the commit and start paths each lose an invariant the rest of the codebase already keeps.

This review was performed statically with fakes and placeholders only — no AWS calls were made and no real credentials were used, so anything that would require live IAM Identity Center access is explicitly unverified here rather than passed.

Gate status: `audit` passed; `test` fails on Astryx stale surface inventory (the new `bedrock-sso-setup.tsx` is not listed) and `package` fails on an unrelated Windows packaged-app CDP timeout. The PR is also currently dirty.

<details>
<summary>中文</summary>

在 `c87e304c` 上审，一条 **[P1]** 加一条 **[P2]**，均行内提在 `bedrock-sso-coordinator.ts`，暂缓。

两条都属生命周期问题而非加密问题：登录流程的正常路径没问题，但提交路径和启动路径各自丢掉了一个仓库其他地方已经维持的不变量。

本次审查仅为静态审查，全程使用 fake 与占位符——没有发起任何 AWS 调用，也没有使用真实凭证；因此凡是需要真实 IAM Identity Center 访问才能确认的部分，这里明确标为未验证，而不是视为通过。

门禁状态：`audit` 通过；`test` 因 Astryx surface inventory 过期而失败（新增的 `bedrock-sso-setup.tsx` 未登记），`package` 因与本 PR 无关的 Windows 打包应用 CDP 超时而失败。此外该 PR 当前处于冲突状态。
</details>

keyuchen21:
@Astro-Han Thanks for the detailed reviews, Astro-Han — especially for spelling out the lifecycle and identity invariants and the limits of fake-only validation.

All reported blockers are addressed on the current rebased head `47fb80977`:

### Coordinator lifecycle findings

- **Durable commit / invalidation (P1):** the durable onboarding result is recorded and replayable before backend invalidation. Invalidation stays inside the activation mutation fence; if it fails, durable success is preserved and activation is poisoned instead of returning `persistence_failed`.
- **Concurrent starts (P2):** starts pass through a serialized admission lease. A superseding start waits for the genuinely in-flight start/settlement, and `close()` waits for admission, so a device flow cannot continue without an owning attempt.

### Independent-review P1s

- **Renderer cleanup / first login:** mount lifetime and attempt cancellation now use separate effects, so changing `attemptId` no longer marks the component unmounted. The login sequence was moved into a testable flow with coverage through account and role selection and cancellation before post-auth calls.
- **Ambient `AWS_BEARER_TOKEN_BEDROCK`:** both model execution and AWS SDK model discovery now explicitly prefer SigV4, suppressing bearer-token environment fallback and preserving the Host-owned SSO identity. Regression tests set the environment variable and prove the explicit credential provider is called and requests are SigV4-signed rather than Bearer-authenticated.
- **Post-auth cancellation / Host close:** account listing, role listing, role credential resolution, model discovery, and manual probes share the attempt abort signal. Admitted post-auth work is tracked; Cancel aborts it, and Host close waits for settlement even after attempt removal.
- **Astryx inventory:** regenerated; `bedrock-sso-setup.tsx` is listed and the inventory check passes.

The branch is rebased onto current `main`, adapted to the current multi-account onboarding target/ticket API, and GitHub reports it mergeable.

Validation on the current head:

- Bedrock Runtime tests: **5/5 passed**
- Bedrock Runtime Host lifecycle tests: **4/4 passed**
- Bedrock renderer login-flow tests: **2/2 passed**
- Runtime Host suite: **1382 passed, 9 skipped, 0 failed**
- Runtime + Runtime Host typechecks, Desktop main/renderer typechecks, Biome, diff check, notice inventory, and Astryx inventory checks passed

Live AWS SSO smoke validation was also completed with a real, locally configured IAM Identity Center session: STS identity resolution, 50-model AWS CLI discovery, 78-model Runtime discovery, and a low-cost Nova Converse call all passed with usage returned. Repeating Runtime discovery and Converse with an intentionally invalid ambient bearer token also passed through the explicit SSO/SigV4 provider. No credential, account, role, token, request ID, model response, or secret material was printed or persisted by the test harness.

Ready for re-review when convenient.

keyuchen21:
Follow-up: the three P1s from the independent review are now addressed in `62cbddf0f`.

- **Renderer cleanup / first login:** split mount lifetime from attempt cancellation, and moved the renderer login sequence into a testable flow. Updating `attemptId` no longer marks the component unmounted. Added coverage that drives authorization through account and role selection, plus cancellation before post-auth calls.
- **Ambient `AWS_BEARER_TOKEN_BEDROCK`:** the Bedrock provider is now constructed with an explicit empty `apiKey`, suppressing the SDK environment fallback and preserving the Host-owned SSO identity. Added a regression test with the environment variable set that proves the explicit credential provider is called and the request is SigV4-signed rather than Bearer-authenticated.
- **Post-auth cancellation / Host close:** account listing, role listing, role credential resolution, model discovery, and manual probes now share the attempt abort signal. Admitted post-auth work is tracked; Cancel aborts it, and Host close waits for settlement even after the attempt is removed. Added a regression test covering both abort and close waiting.
- Regenerated the Astryx surface inventory; `bedrock-sso-setup.tsx` is now listed and the inventory check passes.

Validation on this head:

- Bedrock runtime tests: **4/4 passed**
- Bedrock Runtime Host lifecycle tests: **4/4 passed**
- Bedrock renderer login-flow tests: **2/2 passed**
- Runtime Host suite: **1382 passed, 9 skipped, 0 failed**
- Runtime + Runtime Host typechecks, Desktop main/renderer typechecks, Biome, diff check, and Astryx inventory check passed

The cancellation and identity tests use fakes/placeholders only; no live AWS credentials or calls were used.

keyuchen21:
Rebased the follow-up onto the current `main`; the three-P1 fix is now `ab20534e0` (same changes as the previously referenced `62cbddf0f`). Targeted tests and typechecks were rerun after the rebase and remain green.

Astro-Han (dismissed):
Thanks for taking on the Bedrock provider and for the substantial work already done on the Runtime Host-owned SSO flow. I support adding this provider, and I verified that the earlier renderer-lifetime, ambient bearer-token, post-auth cancellation, durable-commit ordering, and concurrent-start findings have been addressed on exact head `47fb8097`.

There are still three reachable issues that need to be resolved before this can land:

1. **[P1] Keep the AWS SSO session out of generic credential export.** This PR adds `aws_sso` to the generic credential locator, while `configuration.credentials.export` is available to a Remote Owner and the Host handler exports the encoded secret without a kind-specific denial. The encoded `BedrockSsoSession` contains the client secret, access token, and refresh token, so this contradicts the PR's stated boundary that SSO secrets do not cross the Runtime Host protocol. The simplest design is to make `aws_sso` Host-only and reject it at the Host credential-export owner, with a protocol-level regression test. If portable SSO-session export is intentionally supported instead, it needs an explicit authorization/encryption contract; I do not think this PR needs to take on that larger scope.

2. **[P1] Changing the AWS role can persist the previously loaded role.** A normal sequence is: select role A, load models, change the selector to role B, then save. The renderer updates only its displayed `roleName`, while Save submits only `attemptId + enabledModelIds`; the Host persists the role captured by the last model fetch, which is still role A. The UI can therefore show one permission identity while the connection stores another. The smallest fix is to clear the model inventory and require a reload whenever the account or role changes, or to send the selected identity at commit and atomically validate it against the inventory.

3. **[P1] Bedrock connection commit is not tracked by Host shutdown.** `close()` waits for start admission, attempt settlement, and tracked in-flight operations, but `commit()` runs the onboarding mutation outside those tracked operations. If the Host drains while Save is in progress, `close()` can report completion while connection/vault persistence and backend invalidation are still running. Please capture an immutable commit snapshot, register the full mutation/invalidation promise as in-flight, and make shutdown wait for it.

One additional non-blocking lifecycle issue is worth addressing:

- **[P2] Bound and cancel SSO credential refresh.** Refresh-token and `GetRoleCredentials` requests do not currently have a clear operation deadline or Host-owned cancellation boundary. A stalled AWS request can delay a Turn, a connection test, or Host shutdown. This is recoverable and does not justify P1, but it should reuse the existing operation/connection deadlines and be abortable during Host close.

The architecture is otherwise moving in the right direction: Runtime Host remains the execution and credential authority, temporary role credentials stay in memory, and the Desktop flow uses Astryx primitives rather than introducing a parallel UI system. Please also rebase the current conflict, restore ASF headers for the newly added source/generated files, and add before/after or walkthrough screenshots for the new Settings flow so the UI/UX can be reviewed.

This was an AI-assisted review using Codex to inspect the exact-head production paths; I independently checked the reachability, ownership boundaries, and severity of the findings above. No live AWS credentials or calls were used in this review.

<details>
<summary>中文对照</summary>

感谢你推动 Bedrock Provider，也感谢你已经对 Runtime Host 托管的 SSO 流程做了大量工作。我支持加入这个 Provider，并且已经确认之前关于 renderer 生命周期、环境 bearer token、认证后取消、durable commit 顺序和并发 start 的问题，在当前 exact head `47fb8097` 上已经修复。

合并前还有三个真实可达的问题需要解决：

1. **[P1] 不要让通用凭据导出暴露 AWS SSO session。** 当前 `aws_sso` 被加入通用 credential locator，而 Remote Owner 可以调用 `configuration.credentials.export`，Host handler 也没有按凭据类型拒绝。导出的 `BedrockSsoSession` 包含 client secret、access token 和 refresh token，这与 PR 声明的“SSO secret 不跨 Runtime Host protocol”相矛盾。最简单的方案是明确 `aws_sso` 只归 Host 所有，在 Host 的 credential-export owner 处直接拒绝并补协议测试。如果确实要支持跨 Host 导出，则需要完整的授权和加密契约；我认为当前 PR 没必要承担这部分更大的范围。

2. **[P1] 切换 AWS role 后可能保存上一次加载模型时的 role。** 正常操作是：选择 role A、加载模型、把下拉框切到 role B、保存。renderer 只更新了界面中的 `roleName`，Save 只提交 `attemptId + enabledModelIds`；Host 持久化的仍是上一次加载模型时记录的 role A。结果是界面显示的权限身份和实际保存的身份不一致。最小修复是 account/role 改变时清空模型并要求重新加载，或者在 commit 时提交所选身份并与 inventory 做原子校验。

3. **[P1] 保存 Bedrock 连接的操作没有被 Host shutdown 等待。** `close()` 会等待 start admission、attempt settlement 和已登记的 in-flight operation，但 `commit()` 的 onboarding mutation 没有被登记。如果保存过程中 Host 开始 drain，`close()` 可能已经返回完成，而连接、vault 持久化或 backend invalidation 仍在后台运行。请在 commit 开始时捕获不可变快照，把完整 mutation/invalidation promise 纳入 in-flight，并让 shutdown 等待它。

另外有一个非阻塞但值得处理的生命周期问题：

- **[P2] 给 SSO credential refresh 加上明确的超时和取消边界。** refresh token 和 `GetRoleCredentials` 请求目前没有清晰的 operation deadline 或 Host-owned cancellation。AWS 请求卡住时可能拖慢 Turn、连接测试或 Host 关闭。这个问题可以恢复，因此不应定为 P1，但应复用现有 operation/connection deadline，并允许 Host close 中止它。

除此之外，整体架构方向是对的：Runtime Host 仍是执行和凭据 authority，临时 role credential 只留在内存，Desktop 也复用了 Astryx 原语，没有新建平行 UI 体系。还请处理当前与 main 的冲突，为新增源文件和生成文件补齐 ASF header，并在 PR 正文加入新 Settings 流程的前后对比或操作截图，方便完成 UI/UX 人工确认。

本次评审由 Codex 辅助检查 exact-head 生产路径；上述问题的可达性、职责边界和分级由我独立核验。本次没有使用真实 AWS 凭据，也没有发起真实 AWS 调用。

</details>

packages/core/src/runtime-policy/credential-vault-codec.ts
```diff
@@ -51,7 +51,12 @@ export function decodeCredentialLocator(value: unknown): CredentialLocator {
       'connectionId',
       'kind',
     ]);
-    if (item.kind !== 'api_key' && item.kind !== 'oauth_token' && item.kind !== 'request_headers') {
+    if (
+      item.kind !== 'api_key' &&
+      item.kind !== 'oauth_token' &&
+      item.kind !== 'aws_sso' &&
```
Astro-Han, line 57: **[P1, reachability ③] Keep `aws_sso` out of generic credential export.** Adding the kind here makes it a valid locator for `configuration.credentials.export`; that operation is available to a Remote Owner, and the Host export handler currently returns the encoded secret without a kind-specific denial. The serialized session contains the client secret, access token, and refresh token, so this crosses the protocol boundary the PR says should remain Host-only. The smallest solution is to reject `aws_sso` at the Host credential-export owner and add a protocol regression test. If portable SSO export is intentionally supported, it needs a separate explicit authorization/encryption design; deleting that promise from this PR is preferable to inheriting it accidentally.

apps/desktop/src/renderer/settings/bedrock-sso-setup.tsx
```diff
@@ -0,0 +1,213 @@
...
+            value={accountId}
+            onChange={(value: string) => void loadRoles(value)}
+            options={accounts.map((account) => ({
+              value: account.accountId,
+              label: `${account.accountName ?? account.emailAddress ?? 'AWS'} · ••••${account.accountId.slice(-4)}`,
+            }))}
+            isDisabled={busy}
+          />
+          {roles.length > 0 && (
+            <Selector label={zh ? '角色' : 'Role'} value={roleName} onChange={(value: string) => setRoleName(value)} options={roles.map((role) => ({ value: role, label: role }))} isDisabled={busy} />
```
Astro-Han, line 178: **[P1, reachability ①] Changing the displayed role does not invalidate the inventory loaded for the previous role.** A user can select `ReadOnlyAccess`, load models, switch this selector to `AdministratorAccess`, and save. Save submits only `attemptId + enabledModelIds`, while the Host persists the role captured by the last model fetch, so the UI can show one AWS permission identity while the connection stores another. Please clear `models` and `enabledModelIds` and require a reload whenever account/role changes, or send the selected identity at commit and atomically validate it against the inventory. Add the four-step interaction as a renderer regression test.

packages/runtime-host/src/server/bedrock-sso-coordinator.ts
```diff
@@ -0,0 +1,570 @@
...
+    }
+    const available = new Set(attempt.models.map((model) => model.id));
+    if (enabledModelIds.length === 0 || enabledModelIds.some((id) => !available.has(id))) {
+      return failure(
+        'invalid_request',
+        'Enabled Bedrock models are not in the validated inventory',
+      );
+    }
+    try {
+      return await this.activation.runMutation(async () => {
```
Astro-Han, line 391: **[P1, reachability ②] Track the full commit under Host shutdown.** `close()` waits for start admission, attempt settlement, and `#inFlightOperations`, but this onboarding mutation is not registered in any of them. If drain begins while Save is running, `close()` can report completion while connection/vault persistence and backend invalidation are still executing. Please capture an immutable attempt snapshot before the mutation, register the entire mutation/invalidation promise as in-flight, and make `close()` wait for it; once durable commit begins, ordinary cancel should not mutate the captured inputs.

packages/runtime-host/src/server/bedrock-sso-execution-authority.ts
```diff
@@ -0,0 +1,247 @@
...
+      let session = state.session ?? parseBedrockSsoSession(state.raw);
+      if (!session) {
+        throw new BedrockSsoCredentialError(
+          'credential_unavailable',
+          'Stored IAM Identity Center session is invalid',
+        );
+      }
+      if (session.expiresAt - this.now() <= REFRESH_SKEW_MS) {
+        try {
+          const refreshed = await refreshBedrockSsoSession({
```
Astro-Han, line 133: **[P2, reachability ②] Give credential resolution an explicit deadline and Host-owned cancellation boundary.** Both session refresh here and the later `GetRoleCredentials` call use the transport without an operation signal/deadline owned by this authority. A stalled AWS request can delay a Turn, a connection test, or Host shutdown. This is recoverable, so P2 is the appropriate severity, but please reuse the existing operation/connection timeout budget and let Host close abort an in-progress resolution; add a stalled-fetch test for cancellation and bounded completion.

Astro-Han (commented):
Thanks again for taking on the Bedrock provider. I support landing this capability, and the end-to-end owner direction is sound: Runtime Host owns the SSO session and execution identity, temporary role credentials stay in memory, and the Desktop flow reuses Astryx. I dismissed my previous Request Changes review and am replacing it with this non-blocking review plus line-level findings.

The headline `+44,656/-3,912` diff is less alarming than it looks: roughly 40.5k added lines are generated metadata/pricing, dependency lock data, and third-party notices; the handwritten implementation and tests are about 4.1k added lines. A complete provider necessarily crosses Core, Storage, Runtime, Runtime Host, protocol, and Desktop, so I would not split the coherent end-to-end slice merely to make the number smaller.

There is still useful scope contraction available inside this PR:

- Remove the unrelated Claude subscription surface carried into this branch (`claudeSubscription`, `claudeDeviceId`, `oauth-subscription.ts`, and `subscription-auth.ts` additions) unless the author can point to a Bedrock dependency. It adds public bridge/runtime surface without completing a Bedrock requirement.
- Collapse the duplicate and conflicting credential-authority text added to `SECURITY.md`; there should be one current description and one numbering sequence.
- Connect Bedrock metadata/pricing generation to the existing `sync/refresh/check:model-metadata` authority rather than leaving a second standalone generator path that can drift. The generated data itself is justified; the parallel maintenance entry point is not.
- Align the duplicated `ai` dependency version with the workspace version where possible.

I do not see evidence that the Bedrock behavior tests themselves are the source of the size problem, so I would keep tests that pin the SSO, SigV4, lifecycle, and execution boundaries. The best reduction is removing unrelated production surface and duplicate authorities, not deleting useful regression evidence.

Please also rebase the current protocol-epoch conflict, restore ASF headers in the new source/generated files, and add a short screenshot or walkthrough of the new Settings flow for UI/UX review.

This was an AI-assisted review using Codex to inspect the exact-head production paths; I independently checked the reachability, ownership boundaries, severity, and scope recommendations. No live AWS credentials or calls were used.

<details>
<summary>中文对照</summary>

再次感谢你推动 Bedrock Provider。我支持加入这项能力，整体 owner 方向也是对的：Runtime Host 持有 SSO session 和执行身份，临时 role credential 只留在内存，Desktop 复用 Astryx。我已经撤销之前的 Request Changes，改为这份不阻拦式 Review，并把具体问题分别放到对应代码行。

`+44,656/-3,912` 看起来很大，但约 40.5k 新增行来自生成的 metadata/pricing、依赖 lock 数据和第三方 notices；手写实现与测试约为 4.1k 行。完整 Provider 本来就需要贯穿 Core、Storage、Runtime、Runtime Host、protocol 和 Desktop，因此不建议为了缩小数字把一个完整端到端切片机械拆开。

但当前 PR 内仍有明确可以收敛的内容：

- 删除与 Bedrock 无关的 Claude subscription surface，包括 `claudeSubscription`、`claudeDeviceId`、`oauth-subscription.ts` 和 `subscription-auth.ts` 的相关新增；除非能说明它们是 Bedrock 的必要依赖。
- 合并 `SECURITY.md` 中重复、相互冲突的凭据 authority 描述，只保留一套当前事实和连续编号。
- 把 Bedrock metadata/pricing 生成接入现有 `sync/refresh/check:model-metadata` authority，不要留下会漂移的第二套独立生成入口。生成数据本身有必要，平行维护路径没有必要。
- 能统一时，将重复的 `ai` 依赖版本与 workspace 当前版本对齐。

目前没有证据表明 Bedrock 行为测试是体积问题，因此应该保留约束 SSO、SigV4、生命周期和执行边界的测试。最有价值的减法是删除无关生产 surface 和重复 authority，而不是删除有效回归证据。

另外请处理当前 protocol epoch 冲突，为新增源文件和生成文件补齐 ASF header，并在 PR 正文加入一段新 Settings 流程的截图或操作说明，方便完成人工 UI/UX 确认。

本次评审由 Codex 辅助检查 exact-head 生产路径；可达性、职责边界、分级和 scope 建议由我独立核验。本次没有使用真实 AWS 凭据，也没有发起真实 AWS 调用。

</details>

packages/runtime-host/src/server/bedrock-sso-coordinator.ts
```diff
@@ -0,0 +1,570 @@
...
+        const begun = await this.stores.operations.beginConnectionOnboarding({
+          target: { kind: 'create', providerType: 'amazon-bedrock' },
```
Astro-Han, line 393: [P1] Suggestion (category ① — normal re-authorization/second setup): could this commit update the one canonical Bedrock connection instead of always creating another entry? The signed-in Bedrock card remains clickable, so a user can complete setup a second time. This `create` target reaches `deriveConnectionSlug()`, which turns an existing `amazon-bedrock` into `amazon-bedrock-2`; onboarding then writes that entry without re-running the canonical provider decoder. On the next catalog read, `decodeCanonicalConnectionCatalogEntry()` rejects every Bedrock slug except exactly `amazon-bedrock`, leaving the persisted connection document unreadable (`invalid_document`) and the new credential orphaned. Returning the first provider match at lines 414–416 also hides which entry was actually committed. I suggest making the storage authority update/fail closed on the single canonical Bedrock identity, returning the committed connection directly, and adding a first-setup → second-setup/re-auth regression that reads the catalog afterward. Please push back if a UI or Host admission invariant already makes this path unreachable.

Astro-Han (commented):
Thanks for moving the Bedrock SSO flow behind the Runtime Host authority. I reviewed exact head `47fb809772c05b0f9bc2c7624b89947feb4a15f7` and found one additional canonical-storage issue that is separate from the existing current threads. The branch is currently conflicting and has no exact-head checks, so I have not treated it as merge-ready. Please do push back if Bedrock setup is intentionally single-use or another invariant prevents a second onboarding attempt—the author may have context that is not visible from this lane.

感谢把 Bedrock SSO 流程收进 Runtime Host 权威。当前 head 还存在一个与已有线程不同的 canonical storage 问题；分支目前冲突且没有 exact-head checks。若产品明确禁止第二次配置，或另有不变量阻止该路径，也欢迎直接反驳并补充上下文。

</details>

packages/runtime-host/src/server/bedrock-sso-coordinator.ts
keyuchen21: Fixed in f22fdbefb. Storage now treats repeated Amazon Bedrock create onboarding as re-authorization of the existing canonical connection, the coordinator returns the exact committed candidate by connection ID, and the regression performs first setup followed by second setup and successfully reads back one canonical `amazon-bedrock` catalog entry. I also merged current `main` to resolve the branch conflicts.

Issue #1067, opened 2026-07-16
fix(runtime): emit contextRemaining in token_usage events so the TUI ctx segment renders


Astro-Han:
## Problem

The TUI statusline ctx segment (`ctx 32k/128k 25%`) never renders. The entire pipeline is wired but the source never produces data:

- `TokenUsageEvent.contextRemaining` is defined (`packages/core/src/events.ts:487`)
- `ai-sdk-flow.ts:465-466` forwards it to `StoredMessage`
- `pi-transcript.ts:478` stores it on `state.usage`
- `pi-transcript.ts:946` gates the ctx segment on `usage.contextRemaining !== undefined`

But `ai-sdk-backend.ts:1770–1790` constructs the `TokenUsageEvent` without ever setting `contextRemaining`. The field is always `undefined`, so the gate is always `false`, and the ctx segment is always omitted.

#1064 identified the symptom and #1066 addressed the chrome side (color thresholds, cwd shortening), but deliberately omitted a fallback using `token_usage.input` because that field is a billing-cumulative sum across tool-loop steps — using it as "used context" would produce misleading percentages (potentially >100%). The correct fix is to set `contextRemaining` at the source.

## Root cause

`ai-sdk-backend.ts` has both pieces of data needed to compute `contextRemaining`:

1. `contextWindow` — from `resolveSelectedModelContextWindow(connection, modelId)` (already used for mid-turn capacity compaction at line 2558)
2. Last step's `inputTokens` — from the `finish-step` chunk's `stepUsage.inputTokens` (captured at line 1525 as `stepUsage`, accumulated into `completedStepUsage`)

The last step's `inputTokens` is the actual prompt token count of the final API request = how much context is in use. `contextRemaining = contextWindow - lastStepInputTokens`.

`tokenUsage.inputTokens` (the `totalUsage` used for the event's `input` field) is NOT suitable — it is cumulative across all tool-loop steps, which is why #1066 correctly rejected it as a fallback.

## Desired outcome

- Capture the last step's `inputTokens` from the `finish-step` handler (not the cumulative `totalUsage`).
- When constructing the `TokenUsageEvent`, compute `contextRemaining = max(0, contextWindow - lastStepInputTokens)` when both values are available, and include it on the event.
- The TUI ctx segment renders without any further changes — all downstream plumbing already exists.
- Desktop's context display (if any) also benefits if it consumes the same field.

## Non-goals

[...]

PR #3373, opened 2026-08-21
feat(cli): degrade status-line ctx segment to ?/window when usage is unavailable (#3371)

### LINKED ISSUE #3371: feat(cli): status line should degrade gracefully when ctx/token usage is unavailable
## Context

The TUI status line already implements a token/context segment (#1067): `ctx used/window pct%` (yellow >80%, red >95%), plus `$cost` and `cache %` in `renderMakaPiStatusLine` (packages/cli/src/pi-transcript.ts).

However, the segment is **silently absent** unless ALL of these hold:
- the connection's selected model has a resolvable `contextWindow` (declared via relay profile, connection model list, or static metadata), and
- the provider reports per-step input tokens so the runtime can emit `contextRemaining` in `token_usage` (ai-sdk-backend.ts), and
- at least one turn has completed.

Observed on an `opencode-go` connection: `Maka · Full access · muse-spark-1.2-contributor · opencode-go · ~/project/maka-agent` — no ctx, no cost. The user cannot tell whether the connection doesn't report usage, the model lacks a declared context window, or just no turn has finished yet.

## Reference: pi's footer design (badlogic/pi-mono)

pi's interactive footer (`packages/coding-agent/src/modes/interactive/components/footer.ts`) always renders a context segment and degrades explicitly:
- `pct%/window` with color thresholds (warning >70%, error >90%)
- fallback `?/200k` when usage is unknown but the window is known
- cumulative `↑in ↓out R/W CH% $cost` stats alongside

## Proposal

1. When `modelContextWindow` is known but `contextRemaining` is not yet available (fresh session / provider doesn't report usage), render `ctx ?/<window>` (dim) instead of nothing.
2. When neither is known, keep the segment hidden (current behavior) — no window to anchor a fallback.
3. Follow-up: verify whether the opencode-go relay reports per-step input tokens; if it does but `contextRemaining` still isn't emitted, fix that path.

## Acceptance criteria

- A session on a connection with a known context window shows `ctx ?/<window>` before the first completed turn, and `ctx used/window pct%` after.
- Existing color thresholds and the omit-when-unknown-window behavior are unchanged.

me2seeks:
## Summary

The TUI status-line ctx segment (#1067) only rendered when **both** the model context window and a `token_usage` `contextRemaining` were present. On connections whose provider does not report per-step input tokens (e.g. `opencode-go`), the status line shows no context indicator at all — the user cannot distinguish "provider doesn't report usage" from "not measured yet" from "window unknown".

In `renderMakaPiStatusLine` (packages/cli/src/pi-transcript.ts), the ctx segment is lifted out of the `if (usage)` block:

- window known + `contextRemaining` available → unchanged: `ctx 20k/500k 4%` with the existing 80%/95% color thresholds
- window known + no usage yet → **new**: dim `ctx ?/500k` fallback, following pi-mono's footer design (`?/200k` when context usage is unknown)
- window unknown → unchanged: no segment

The follow-up from the issue — verifying whether the opencode-go relay reports per-step input tokens — is not in scope here.

Fixes #3371

## Verification

- pi-transcript.test.js 59/59 (new unit test covering all four combinations: no usage / usage without contextRemaining / unknown window stays hidden / measured rendering unchanged)
- pi-tui-runner.test.js 119/119 (including the existing `ctx 20k/500k 4%` integration assertion)
- biome format clean; full dependency chain + cli build (typecheck) pass

## AI use

Tool(s) and scope: Maka (AI coding agent) authored the design, implementation, and tests; the diff was human-reviewed before push. `Generated-by: Maka` trailer is present on the commit and retained on this branch.

## Checklist

- [ ] No

### DIFF
--- a/packages/cli/src/__tests__/pi-transcript.test.ts
+++ b/packages/cli/src/__tests__/pi-transcript.test.ts
@@ -118,6 +118,55 @@ describe('Maka Pi TUI transcript', () => {
     assert.doesNotMatch(stripAnsi(renderMakaPiStatusLine({ ...meta(), goal: null }, 120)), /goal/);
   });
 
+  test('status line degrades to ctx ?/window when the window is known but usage is not (#3371)', () => {
+    // No usage object at all: the window is known, so degrade explicitly.
+    assert.match(
+      stripAnsi(renderMakaPiStatusLine({ ...meta(), modelContextWindow: 500_000 }, 120)),
+      /ctx \?\/500k/,
+    );
+
+    // Usage exists but contextRemaining has not been reported yet.
+    assert.match(
+      stripAnsi(
+        renderMakaPiStatusLine(
+          {
+            ...meta(),
+            modelContextWindow: 500_000,
+            usage: { costUsd: 0, cacheHitInput: 0, cacheMissInput: 0 },
+          },
+          120,
+        ),
+      ),
+      /ctx \?\/500k/,
+    );
+
+    // Window unknown: the segment stays hidden, even with usage present.
+    assert.doesNotMatch(
+      stripAnsi(
+        renderMakaPiStatusLine(
+          { ...meta(), usage: { costUsd: 0.01, cacheHitInput: 0, cacheMissInput: 0 } },
+          120,
+        ),
+      ),
+      /ctx/,
+    );
+
+    // contextRemaining present: the measured segment is unchanged.
+    assert.match(
+      stripAnsi(
+        renderMakaPiStatusLine(
+          {
+            ...meta(),
+            modelContextWindow: 500_000,
+            usage: { costUsd: 0, cacheHitInput: 0, cacheMissInput: 0, contextRemaining: 480_000 },
+          },
+          120,
+        ),
+      ),
+      /ctx 20k\/500k 4%/,
+    );
+  });
+
   test('keeps assistant text after a tool call visible after the tool block', () => {
     const state = createMakaPiTranscriptState();
     appendUserPrompt(state, 'inspect the package');
--- a/packages/cli/src/pi-transcript.ts
+++ b/packages/cli/src/pi-transcript.ts
@@ -1359,22 +1359,29 @@ export function renderMakaPiStatusLine(metadata: MakaPiTranscriptMetadata, width
     );
   }
   const usage = metadata.usage;
+  // ctx segment: only show "used" when contextRemaining is available, since
+  // token_usage.input is a billing-cumulative sum across tool-loop steps,
+  // not the last request's context size. Using it as a proxy for "used"
+  // would produce misleading percentages (potentially >100%).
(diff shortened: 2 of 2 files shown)


Astro-Han (commented):
Thanks — this is a small change that gets a distinction right, and the distinction is the part worth naming: the status line now separates "the window is unknown" from "the window is known but nothing has been measured yet." Those are different states and they were collapsing into the same blank.

Reviewed at exact head `2a7230c0249c2e555bbb793f37cc58bcbd3314d4` against base `a2f4e6f808c8647c62194aa41bb683f6829b08f1` (real merge-base `57e08d83497d1d7ace7d6eff88e4e5267a0345b5`). **No findings — no P0, P1, P2 or P3.** No checks have run on this head, so this stays a COMMENT; it is not an approval.

**On whether `?` degrades or hides.** This was the question we most wanted to answer against you, because a placeholder is exactly how a real regression gets papered over. It does not hide anything here: the runtime clears `contextRemaining` when the latest step carries no input tokens, so `?` cannot be a stale measurement wearing a fresh label. The condition is also correctly asymmetric — `window && remaining` renders the measured value, `window` alone renders `?/window`, and an unknown window still renders nothing at all, because `?/unknown` would be noise rather than information. `cost` and `cache` stay behind `if (usage)`, so nothing else moved.

**On whether the problem is real.** It is, and on the ordinary path rather than a failure path: a session before its first step has no usage by construction, and a provider that does not report per-step input tokens has none persistently. That is a normal state being rendered as if the window were unknown.

**On scope.** The residual ambiguity — "not measured yet" versus "this provider never reports" — is not resolvable from the data available at this layer, and your description defers it explicitly rather than papering over it. We think that is the right call and the right way to say it.

**Verification, stated plainly.** Two of us reviewed this independently without visibility into each other's conclusion, and a third checked the factual claims underneath both. The targeted suite (`pi-transcript`) was run against a merge of this head with current `main` — the CI-equivalent tree rather than the bare branch — and passed 60/60, including the four-way matrix your tests add (no usage, usage without `contextRemaining`, unknown window, and the measured case unchanged at `20k/500k 4%`). We separately confirmed those four cases exist in the test source rather than taking the run at its word.

**The one thing missing is not yours to fix from the code side:** this head has zero check runs — not green, not red, never executed. Everything above is our verification, not the project's. A push to trigger CI would put that right.

---

Astro-Han (approved):
Following up now that CI has run on this head — `test` is green.

Still at `2a7230c0249c2e555bbb793f37cc58bcbd3314d4`, unchanged since my earlier review, so that conclusion still stands: no findings.

The distinction this makes — "window unknown" versus "window known, nothing measured yet" — is the part worth keeping. And `?` doesn't hide a stale reading, because the runtime clears `contextRemaining` when the latest step carries no input tokens.

Approving.

---

PR #3376, opened 2026-08-21
fix(runtime-host,ui,cli): name ordinary tool calls during the live window

### LINKED ISSUE #3336: bug(desktop): collapsed tool rows show no invocation target — ordinary tool cards read empty
## Symptom

In the Desktop transcript, an ordinary collapsed tool row and collapsed group header can have no invocation target during the live window. A row may read `● Bash` instead of `● Bash  $ git status --porcelain`, so the user cannot tell what a long-running call is doing.

After durable args are available, the shared invocation formatter can derive a command, path, pattern, title, or URL. The remaining cross-surface gap is that Runtime Host live `tool_start` frames intentionally omit full args, while Desktop's collapsed target previously depended primarily on `intent`.

## Scope

This issue covers the ordinary-tool collapsed-row **what** slot:

- carry a bounded, redacted invocation preview through the live Runtime Host event;
- render `intent ?? invocation preview` in the single-call row and group header;
- keep durable args/history as the full-data authority;
- preserve the existing expanded detail panel and result deduplication.

## Scope split

Task Ledger tools are not ordinary invocation summaries. `task_create` needs committed IDs and the complete created set; `task_update` needs the subject and mutation snapshot owned by the Session Task Ledger. Their semantic timeline is tracked separately in #4179 and is intentionally excluded from this issue and #3376.

## Acceptance

- An ordinary tool with a safe structured target shows one line describing what it is doing in the collapsed row and collapsed group header during the live window.
- Preview data is allowlist-shaped, bounded, and redacted; full file bodies and arbitrary provider/MCP payloads do not enter the live wire.
- A settled or historical row continues to use durable args.
- Tools without a safe preview keep the current name-only fallback.
- Task Ledger presentation remains unchanged and is handled by #4179.

### LINKED ISSUE #3338: bug(cli): compact tool rows read empty for quiet or nested-args tools — no 'what' without expanding
## Symptom

In the TUI transcript, an ordinary compact tool row can be name-only during the live `tool_start` window, and a quiet successful call can degrade to an unhelpful `● Bash (no output)`. The user has no useful middle ground between the compact row and expanding the full output card.

Current main already renders commands and other invocation details after durable args/results arrive. This issue covers the earlier live window and the compact-row relationship between **what ran** and **what happened**.

## Scope

- Let ordinary tools consume a bounded, redacted live invocation preview.
- Keep command/path/query information visible independently of the result annotation.
- Do not show the dim `no output` placeholder when the row already names a successful quiet call.
- Preserve fixed-shape result statistics, exit status, truncation priority, and the existing `Ctrl+O` expanded-card contract.
- Empty or unavailable args remain a name-only fallback rather than `input: {}` noise.

## Scope split

Task Ledger tools are not generic nested-args tools. A useful create row needs the complete committed task set and real IDs; a useful update row needs the subject and exact mutation state, not merely `T1 → completed`. Their Host-owned semantic timeline is tracked separately in #4179 and is intentionally excluded from this issue and #3376.

## Acceptance

- During the live window, a quiet ordinary Bash call can show `● Bash  $ git add -A (running)` rather than `● Bash (running)`.
- After success, a named quiet call does not use `(no output)` as its only outcome signal.
- Preview strings are redacted and bounded before crossing the Runtime Host wire.
- Durable args replace the preview after reconciliation without changing tool identity.
- Task Ledger presentation remains unchanged and is handled by #4179.

me2seeks:
## Summary

Closes #3336. Closes #3338.

During the live interval between `tool_start` and the first result/durable transcript reconcile, Runtime Host omitted full tool args, so compact TUI and collapsed Desktop rows could render only a name such as `● Bash (running)`. Settled rows already receive durable args and are not the gap addressed here.

This change gives both surfaces the same bounded live target path:

```text
Before: ● Bash (running)
After:  ● Bash  $ git status --porcelain (running)
```

- Core projects a small allowlisted `argsPreview`: each string is bounded and display-redacted, the complete preview is capped, free-form shapes are restricted to their owning built-in tools, and file bodies or generic input payloads are excluded.
- Runtime Host sends optional bounded `intent` and `argsPreview` fields on live `tool_start` frames, never full args. Because older strict clients reject the new keys, the compatibility epoch advances from 64 to 65.
- Desktop and TUI both prefer runtime-authored intent, then format args/argsPreview, then fall back to the bare tool name. Both use the shared display redactor.
- Task Ledger tools are deliberately excluded from generic live previews. They need committed task identity and durable mutation facts; that semantic timeline is tracked separately in #4179.

This does not add result previews or change the Ctrl+O expansion contract.

## Verification

- `npm test` — all workspace tests passed on exact head `3accefbe63123b5e9411720a7e02d059bbe25b03`.
- `npm run build` — passed.
- `npx biome check` for all files changed during review repair — passed.
- `git diff --check origin/main...HEAD` — passed.
- Two independent final reviews of the exact head reported no P0/P1 findings.

## Review focus

- `argsPreview` is display-redacted before publication; `intent` is a bounded Runtime-authored pass-through and is display-redacted by each surface.
- The strict decoder accepts only the two new optional `tool_start` keys and enforces the wire budgets.
- The four Task Ledger tools (`task_create`, `task_update`, `task_list`, `task_get`) fail closed at the preview projection entry.

## AI use

Select exactly one:

Tool(s) and scope: Codex implemented and tested the live tool target projection, resolved the rebase and review findings, and split Task Ledger UX into #4179. Generated-by trailers are present on the branch commits.

## Checklist

- [ ] No

### DIFF
--- a/packages/cli/src/pi-transcript.ts
+++ b/packages/cli/src/pi-transcript.ts
@@ -814,14 +814,16 @@ export function applyMakaSessionEventToTranscript(
       const suppressed =
         (event.toolName === 'Read' || event.toolName === 'StopBackgroundTask') &&
         !!ref &&
-        !!findShellRunParent(state, ref, event.toolUseId);
-      state.entries.push({
+        !!findShellRunParent(state, ref, event.toolUseId);      state.entries.push({
         kind: 'tool',
         turnId: event.turnId,
         toolUseId: event.toolUseId,
         toolName: event.toolName,
         ...(event.displayName ? { title: event.displayName } : {}),
-        input: projectToolActivityArgs(event.toolName, event.args),
+        // Live Runtime Host frames omit full args; the bounded wire preview
+        // still lets the compact row name the call. The turn-end reconcile
+        // replaces it with the durable full args.
+        input: projectToolActivityArgs(event.toolName, event.args ?? event.argsPreview),
         resultVersion: 0,
         progress: createProgressBuffer(),
         outputDeltas: createOutputBuffer(),
--- a/packages/cli/src/pi-transcript.ts
+++ b/packages/cli/src/pi-transcript.ts
@@ -173,6 +173,8 @@ export type MakaPiTranscriptEntry =
       toolUseId: string;
       toolName: string;
       title?: string;
+      /** Runtime-authored, bounded description of the live call's purpose. */
+      intent?: string;
       input: unknown;
       /** Structured result returned by the tool. */
       result?: ToolResultContent;
@@ -821,6 +823,7 @@ export function applyMakaSessionEventToTranscript(
         toolUseId: event.toolUseId,
         toolName: event.toolName,
         ...(event.displayName ? { title: event.displayName } : {}),
+        ...(event.intent ? { intent: event.intent } : {}),
         // Live Runtime Host frames omit full args; the bounded wire preview
         // still lets the compact row name the call. The turn-end reconcile
         // replaces it with the durable full args.
--- a/packages/cli/src/pi-transcript.ts
+++ b/packages/cli/src/pi-transcript.ts
@@ -814,7 +814,8 @@ export function applyMakaSessionEventToTranscript(
       const suppressed =
         (event.toolName === 'Read' || event.toolName === 'StopBackgroundTask') &&
         !!ref &&
-        !!findShellRunParent(state, ref, event.toolUseId);      state.entries.push({
+        !!findShellRunParent(state, ref, event.toolUseId);
+      state.entries.push({
         kind: 'tool',
         turnId: event.turnId,
         toolUseId: event.toolUseId,
--- a/packages/core/src/tool-quiet-preview.ts
+++ b/packages/core/src/tool-quiet-preview.ts
@@ -397,12 +397,16 @@ function previewListSubjects(
   listKey: 'tasks' | 'questions',
   itemKey: 'subject' | 'question',
 ): { items: Record<string, unknown>[]; total: number } | undefined {
+  if (isSensitiveKey(listKey) || isSensitiveKey(itemKey)) return undefined;
   const list = record[listKey];
   if (!Array.isArray(list) || list.length === 0) return undefined;
   const items: Record<string, unknown>[] = [];
   for (const entry of list.slice(0, ARGS_PREVIEW_LIST_MAX_ITEMS)) {
     const text = previewStringField(asRecord(entry) ?? {}, itemKey);
-    if (text !== undefined) items.push({ [itemKey]: text });
+    if (text === undefined) continue;
+    const redacted = redactSecrets(text);
+    if (redacted.trim().length === 0) continue;
+    items.push({ [itemKey]: redacted });
   }
   return items.length > 0 ? { items, total: list.length } : undefined;
 }
--- a/packages/core/src/tool-quiet-preview.ts
+++ b/packages/core/src/tool-quiet-preview.ts
@@ -47,8 +47,6 @@ interface QuietPreviewStrings {
   written: string;
   /** Format a byte count suffix, e.g. `共 7 字节` / `7 bytes`. */
   bytes: (n: number) => string;
-  /** Suffix for a task list previewed by its first entry, e.g. `等 3 项` / `+2 more`. */
-  moreTasks: (total: number) => string;
   /** Suffix for a question list previewed by its first entry, e.g. `等 2 问` / `+1 more`. */
   moreQuestions: (total: number) => string;
 }
@@ -62,7 +60,6 @@ const STRINGS_BY_LOCALE: Record<UiLocale, QuietPreviewStrings> = {
     replacements: (n) => `${n} 处`,
     written: '已写入',
     bytes: (n) => `共 ${n} 字节`,
-    moreTasks: (total) => (total > 1 ? ` 等 ${total} 项` : ''),
     moreQuestions: (total) => (total > 1 ? ` 等 ${total} 问` : ''),
   },
   en: {
@@ -73,7 +70,6 @@ const STRINGS_BY_LOCALE: Record<UiLocale, QuietPreviewStrings> = {
     replacements: (n) => (n === 1 ? '1 replacement' : `${n} replacements`),
     written: 'written',
     bytes: (n) => `${n} bytes`,
-    moreTasks: (total) => (total > 1 ? ` +${total - 1} more` : ''),
     moreQuestions: (total) => (total > 1 ? ` +${total - 1} more` : ''),
   },
 };
@@ -271,33 +267,6 @@ export function formatToolInvocationLine(
     }
   }
 
-  // Session task ledger and other structured-args tools: the generic key:value
-  // fallback renders their nested payloads as `tasks:` / `questions:` dumps, so
-  // give the row the one fact a reader needs — the first subject / question,
-  // the goal condition — with a count suffix when more entries exist.
-  if (name === 'task_create') {
-    const tasks = Array.isArray(args.tasks) ? args.tasks : undefined;
-    const firstSubject = tasks
-      ?.map((task) => stringField(asRecord(task), 'subject'))
-      .find((subject) => subject !== undefined);
-    if (firstSubject) {
-      // A wire args preview caps the list; `tasksTotal` keeps the real count.
-      const total = numberField(args, 'tasksTotal') ?? tasks!.length;
-      return redactSecrets(`${firstSubject}${s.moreTasks(total)}`);
-    }
-  }
-
-  if (name === 'task_update') {
-    const subject = stringField(args, 'subject');
-    if (subject) return redactSecrets(subject);
-    const id = stringField(args, 'id');
-    if (id) {
-      const status = stringField(args, 'status');
-      const shortId = id.length > 12 ? `${id.slice(0, 8)}…` : id;
-      return redactSecrets(status ? `${shortId} → ${status}` : shortId);
-    }
-  }
-
   if (name === 'GoalSet') {
     const condition = stringField(args, 'condition');
     if (condition) return redactSecrets(condition);
@@ -356,8 +325,10 @@ export function formatToolInvocationLine(
 const ARGS_PREVIEW_STRING_MAX_CHARS = 240;
 /** Whole-preview cap; lowest-priority fields drop until the preview fits. */
 const ARGS_PREVIEW_MAX_CHARS = 2048;
-/** List fields (tasks / questions) keep only their leading entries. */
+/** Question lists keep only their leading entries. */
 const ARGS_PREVIEW_LIST_MAX_ITEMS = 4;
+
+const TASK_LEDGER_TOOL_NAMES = new Set(['task_create', 'task_update', 'task_list', 'task_get']);
 
 /**
  * Whitelist of scalar args keys {@link formatToolInvocationLine} can read, in
@@ -404,21 +375,18 @@ function previewStringField(record: Record<string, unknown>, key: string): strin
   return typeof raw === 'string' && raw.trim().length > 0 ? boundPreviewString(raw) : undefined;
 }
 
-function previewListSubjects(
+function previewQuestions(
   record: Record<string, unknown>,
-  listKey: 'tasks' | 'questions',
-  itemKey: 'subject' | 'question',
 ): { items: Record<string, unknown>[]; total: number } | undefined {
-  if (isSensitiveKey(listKey) || isSensitiveKey(itemKey)) return undefined;
-  const list = record[listKey];
+  const list = record.questions;
   if (!Array.isArray(list) || list.length === 0) return undefined;
   const items: Record<string, unknown>[] = [];
   for (const entry of list.slice(0, ARGS_PREVIEW_LIST_MAX_ITEMS)) {
-    const text = previewStringField(asRecord(entry) ?? {}, itemKey);
+    const text = previewStringField(asRecord(entry) ?? {}, 'question');
     if (text === undefined) continue;
     const redacted = redactSecrets(text);
     if (redacted.trim().length === 0) continue;
-    items.push({ [itemKey]: redacted });
+    items.push({ question: redacted });
   }
   return items.length > 0 ? { items, total: list.length } : undefined;
 }
@@ -467,6 +435,10 @@ export function projectToolArgsPreview(
 ): Record<string, unknown> | undefined {
   const record = asRecord(args);
   if (!record) return undefined;
+  // Task rows need committed IDs and the exact Task Ledger mutation snapshot;
+  // args alone cannot identify the user-facing task reliably. Keep them out of
+  // the generic live preview until the Host-owned semantic timeline (#4179).
+  if (TASK_LEDGER_TOOL_NAMES.has(toolName)) return undefined;
 
   // Apply the canonical activity projection first so WriteStdin's inputPreview
   // shape (bounded, display-safe) is what the whitelist picks up.
@@ -495,12 +467,7 @@ export function projectToolArgsPreview(
     const size = previewSize(projected.size);
     if (size) picked.set('size', size);
   }
-  const tasks = previewListSubjects(projected, 'tasks', 'subject');
-  if (tasks) {
-    picked.set('tasks', tasks.items);
-    if (tasks.total > tasks.items.length) picked.set('tasksTotal', tasks.total);
-  }
-  const questions = previewListSubjects(projected, 'questions', 'question');
+  const questions = previewQuestions(projected);
   if (questions) {
     picked.set('questions', questions.items);
     if (questions.total > questions.items.length) picked.set('questionsTotal', questions.total);
@@ -515,8 +482,6 @@ export function projectToolArgsPreview(
     ...ARGS_PREVIEW_NUMBER_KEYS,
     'inputPreview',
     'size',
-    'tasks',
-    'tasksTotal',
     'questions',
     'questionsTotal',
   ];
--- a/packages/core/src/tool-quiet-preview.ts
+++ b/packages/core/src/tool-quiet-preview.ts
@@ -467,10 +467,14 @@ export function projectToolArgsPreview(
     const size = previewSize(projected.size);
     if (size) picked.set('size', size);
   }
-  const questions = previewQuestions(projected);
-  if (questions) {
-    picked.set('questions', questions.items);
-    if (questions.total > questions.items.length) picked.set('questionsTotal', questions.total);
+  // Only the built-in interaction tool owns this free-text shape. Arbitrary
+  // third-party tools may use the same field names for private payloads.
+  if (toolName === 'AskUserQuestion') {
+    const questions = previewQuestions(projected);
+    if (questions) {
+      picked.set('questions', questions.items);
+      if (questions.total > questions.items.length) picked.set('questionsTotal', questions.total);
+    }
   }
 
   if (picked.size === 0) return undefined;
--- a/packages/core/src/tool-quiet-preview.ts
+++ b/packages/core/src/tool-quiet-preview.ts
@@ -30,7 +30,7 @@
  */
 
 import { redactSecrets } from './display-redaction.js';
-import { readWriteStdinInputPreview } from './tool-activity-args.js';
+import { projectToolActivityArgs, readWriteStdinInputPreview } from './tool-activity-args.js';
 import type { UiLocale } from './ui-locale.js';
 
 // ── Locale ───────────────────────────────────────────────────────────────
@@ -47,6 +47,10 @@ interface QuietPreviewStrings {
   written: string;
   /** Format a byte count suffix, e.g. `共 7 字节` / `7 bytes`. */
   bytes: (n: number) => string;
+  /** Suffix for a task list previewed by its first entry, e.g. `等 3 项` / `+2 more`. */
+  moreTasks: (total: number) => string;
+  /** Suffix for a question list previewed by its first entry, e.g. `等 2 问` / `+1 more`. */
+  moreQuestions: (total: number) => string;
 }
 
 const STRINGS_BY_LOCALE: Record<UiLocale, QuietPreviewStrings> = {
@@ -58,6 +62,8 @@ const STRINGS_BY_LOCALE: Record<UiLocale, QuietPreviewStrings> = {
     replacements: (n) => `${n} 处`,
     written: '已写入',
     bytes: (n) => `共 ${n} 字节`,
+    moreTasks: (total) => (total > 1 ? ` 等 ${total} 项` : ''),
+    moreQuestions: (total) => (total > 1 ? ` 等 ${total} 问` : ''),
   },
   en: {
     backgroundTerminal: 'Background terminal interaction',
@@ -67,6 +73,8 @@ const STRINGS_BY_LOCALE: Record<UiLocale, QuietPreviewStrings> = {
     replacements: (n) => (n === 1 ? '1 replacement' : `${n} replacements`),
     written: 'written',
     bytes: (n) => `${n} bytes`,
+    moreTasks: (total) => (total > 1 ? ` +${total - 1} more` : ''),
+    moreQuestions: (total) => (total > 1 ? ` +${total - 1} more` : ''),
   },
 };
 
@@ -255,6 +263,49 @@ export function formatToolInvocationLine(
     return parts.join(' · ');
   }
 
+  // Session task ledger and other structured-args tools: the generic key:value
+  // fallback renders their nested payloads as `tasks:` / `questions:` dumps, so
+  // give the row the one fact a reader needs — the first subject / question,
+  // the goal condition — with a count suffix when more entries exist.
+  if (name === 'task_create') {
+    const tasks = Array.isArray(args.tasks) ? args.tasks : undefined;
+    const firstSubject = tasks
+      ?.map((task) => stringField(asRecord(task), 'subject'))
+      .find((subject) => subject !== undefined);
+    if (firstSubject) {
+      // A wire args preview caps the list; `tasksTotal` keeps the real count.
+      const total = numberField(args, 'tasksTotal') ?? tasks!.length;
+      return redactSecrets(`${firstSubject}${s.moreTasks(total)}`);
+    }
+  }
+
+  if (name === 'task_update') {
+    const subject = stringField(args, 'subject');
+    if (subject) return redactSecrets(subject);
+    const id = stringField(args, 'id');
+    if (id) {
+      const status = stringField(args, 'status');
+      const shortId = id.length > 12 ? `${id.slice(0, 8)}…` : id;
+      return redactSecrets(status ? `${shortId} → ${status}` : shortId);
+    }
+  }
+
+  if (name === 'GoalSet') {
+    const condition = stringField(args, 'condition');
+    if (condition) return redactSecrets(condition);
+  }
+
+  if (name === 'AskUserQuestion') {
+    const questions = Array.isArray(args.questions) ? args.questions : undefined;
+    const firstQuestion = questions
+      ?.map((question) => stringField(asRecord(question), 'question'))
+      .find((questionText) => questionText !== undefined);
+    if (firstQuestion) {
+      const total = numberField(args, 'questionsTotal') ?? questions!.length;
+      return redactSecrets(`${firstQuestion}${s.moreQuestions(total)}`);
+    }
+  }
+
   if (name === 'Grep' || (pattern && (name === 'Glob' || path))) {
     if (pattern) {
       const scope = path ? ` in ${path}` : '';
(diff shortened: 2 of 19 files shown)


Astro-Han:
Heads up — this is currently conflicting with `main`, so I can't review or merge it as-is. **Your CI is fully green**, so it really is just the base that's behind.

One thing worth knowing: [#3397](https://github.com/apache/maka/pull/3397) landed on 2026-08-22 and added ASF license headers across ~2685 files, so the rebase will touch more than you'd expect, and any file you add now needs a header (`npm run write:asf-headers`).

Ping me once it's rebased and I'll pick it up.

Astro-Han:
Independent review of `61391a6eb498539dc20a69e8bbf638c1ebfbf009`. **Three [P2]s, all reproduced against this head.** Posting them here rather than inline: each defect's decisive line sits in unchanged context, so GitHub cannot anchor a review comment to it.

### [P2] A Bash row that was redacted while live shows the original value again once the turn settles

`packages/cli/src/pi-transcript-tools.ts:727-743` — the `Bash` case returns early:

```ts
return `$ ${firstRealLine ?? command.split('\n')[0]!.trim()}`;   // before the shared formatter
```

Two things combine. `projectToolArgsPreview` redacts only the **live** copy — and then `pi-transcript.ts:385-388` does `entry.input = structuredClone(durable.input)`, replacing it with the **full, unredacted** durable args. Meanwhile this `Bash` branch returns before reaching the shared `formatToolInvocationLine`, so the shared redactor never sees the text.

Reproduced with a placeholder secret: the live row renders `Authorization: Bearer <redacted>`; after `reconcileToolsWithStoredMessages` runs over that same row, it renders the placeholder's original value.

**Why this is worse than "not redacted at all": the row was already redacted in front of the user, and then un-redacted itself.** Someone who watched it go by has no reason to look again, and anyone reading the scrollback later sees the raw value with nothing indicating it was ever meant to be hidden. A redactor that reverses itself breaks a promise it already made.

Existing tests cover the live preview only — nothing exercises the live → durable reconcile, which is why the round trip passes.

Minimal fix: keep the "first non-comment command line" presentation (it is genuinely the useful part), but route the final text through the shared redactor instead of returning early, plus a regression asserting a placeholder secret does not reappear across reconcile.

### [P2] A synthesized empty `args = {}` hides the live `WriteStdin` preview

`packages/ui/src/live-turn-projection.ts:329-341` with `packages/core/src/tool-activity-args.ts:163-179`. `projectToolActivityArgs('WriteStdin', undefined)` is the only path returning `{}`. Given a Host `WriteStdin tool_start` with no full `args` but a populated `argsPreview` (`inputPreview={text:"ls -la\n",bytes:7,truncated:false}`, `size={cols:80,rows:24}`), the projected item carries **both** a complete `argsPreview` and `args = {}`.

The wrapper picks with `args ?? argsPreview`. **`{}` is non-nullish, so it wins**, and the row collapses to `Background terminal interaction` — the typed input and the `80x24` geometry both vanish, in precisely the case this PR exists to improve.

**Why the tests miss it:** the UI tests construct items with `args: undefined` directly and the core tests format the preview directly — **neither goes through the real `applyLiveTurnEvent`**, which is the only place that empty object is manufactured. Both sides test around it.

Fix: do not synthesize `args` when the event has no full args, and add a `WriteStdin` regression that runs *through* live projection.

### [P2][design] The new live-preview allowlist omits current built-in structured tools

`packages/core/src/tool-quiet-preview.ts:318-361,429-490`. `projectToolArgsPreview('deep_research_start', {objective:'…', scope_level:'standard'})` returns `undefined`, so the live row shows the bare tool name — while the durable formatter, given the same args, prints `objective: …` and `scope_level: standard`. `deep_research_record_step` keeps only `status` and drops the objective.

The durable generic formatter reads arbitrary structured fields; the new preview authority permits **18 scalar keys**, and neither `objective` nor `scope_level` is among them. The tests enumerate the headline-shaped tools that were picked, without cross-checking the live preview against the current built-in schema.

**This is not a hypothetical MCP tool** — `deep_research_start` is a reachable built-in today (`packages/runtime/src/deep-research-tools.ts:124-148`). Fix by admitting a bounded/redacted `objective` (and stating the policy for the other current structured tools), with a built-in-coverage test.

### Where the entropy actually is

**The file count is not the problem** — 12 production and 7 test files, 352 production additions against 437 test. Propagation across core, wire, and two clients is a real contract boundary.

The problem is that **"the invocation name" now has two authorities in each of two places**: one shared core formatter plus **8 tool-specific branches in the CLI** (Bash / Read / WriteStdin / Write / Edit / Grep / Glob / maka_computer), and separately the `ARGS_PREVIEW_*` allowlist versus the durable generic formatter's wider field access. **Findings 1 and 3 are not two unrelated bugs — they are the observable drift of those two pairs.**

Convergence worth aiming for: the shared redacted invocation formatter is the single semantic authority, surface-specific code decorates only, and preview coverage has a checkable contract against the current tool catalog rather than a hand-maintained parallel field table.

- `ExploreAgent` / legacy `agent_swarm` objective — the current call name is `ExploreAgent` and `describeToolIntent` already supplies a bounded, redacted intent. **Withdrawn.**
- Arbitrary MCP `{owner,repo}` unnamed while live — behavior confirmed, but withdrawn as a separate finding since `deep_research_start` already demonstrates the same design gap via an in-repo path.
- `Read` / `StopBackgroundTask` live polls still reading `event.args` — pre-existing folding, not a regression within this PR's naming charter. **Withdrawn.**
- Removing `(no output)` once a target exists — an explicit product choice; the status dot and expanded card still carry the outcome. **Withdrawn.**
- The decoder validating only byte bounds, not exact preview shape — the Host is the production authority; no trust-boundary bypass established. **Withdrawn.**

### Gate state

- `CONFLICTING / DIRTY`. Against `origin/main=2e5fe9c5`, `merge-tree` gives 2 content conflicts: `packages/runtime-host/src/protocol/index.ts` and `packages/ui/src/tool-activity.tsx`.
- **`check-runs` on this head: 0. `gh run list --commit`: 0. `statusCheckRollup`: empty.** There is an older maintainer comment stating CI was green at the time — **that is a historical description, not a binding gate on this head**, and I am not treating it as one.
- All three review endpoints read separately: 1 issue comment (a rebase request, no findings), 0 reviews, 0 inline comments. Nothing to deduplicate against.
- Local on this exact head: clean tree, `npm ci` fine, Desktop dependency build fine, CLI + eval build fine, 8 focused test files **202/202** green. Three targeted probes reproduce all three findings above stably.

*Blind line — provisional judgment sealed before the endpoints were read. Reviewed 2026-08-23 13:35 UTC. No overall verdict offered.*

me2seeks:
Rebased onto current `main` (`2e5fe9c55`). Three conflict resolutions:

1. **Compatibility epoch** (`protocol/index.ts`): main had advanced to 39; this PR's wire change (optional `intent`/`argsPreview` on live `tool_start` frames) is now epoch **40** with both comment lists preserved.
2. `ui/tool-activity.tsx`: merged main's `inferredTarget` parameter (accumulated Computer-Use target) with this PR's args-derived naming — precedence is runtime `intent` → caller-inferred target → first invocation line. `collapsedToolTarget` gains an optional preferred value; group-header callers are unchanged.

Verified on the new head: full root build, runtime-host 1092/1092 (incl. handshake-compatibility), CLI 385/385, biome clean. UI suite is 218/219 — the one failure (`composer-plus-menu › a loading catalog holds the row still and marks the held state`) reproduces identically on a clean checkout of current `main`, so it is pre-existing and unrelated to this branch.

Astro-Han (commented):
Review at exact head `c05b843952f1808417560c0a6e72b27c70c6c5df`.

**No P0–P2.** The mechanism here is well chosen and the security-sensitive part is built the right way round.

The problem is real: live `tool_start` frames from the Runtime Host omit full args, so a compact row could only print `● Bash` with nothing to say *what* ran until the durable transcript arrived at turn end. Rather than widen the wire to carry full args — a `Write` call would drag an entire file across — this adds a bounded, redacted `argsPreview` and has the display path read `args ?? argsPreview`, so durable replay keeps priority and never sees the preview.

Three details worth calling out as correct, because each is a place this kind of change usually goes wrong:

1. **`projectToolArgsPreview` uses an allowlist, not a denylist.** `ARGS_PREVIEW_SCALAR_KEYS` enumerates the keys `formatToolInvocationLine` can actually render; file contents, option payloads and provider blobs have no path into the preview because they are simply not on the list. A denylist here would have needed updating every time a tool grew a new field.
2. **Redaction happens before truncation.** `boundPreviewString` calls `redactSecrets(value)` and *then* slices to 240 chars. The other order is the classic bug — slicing first can leave a partial secret that the redactor no longer recognizes as one, and the truncated remainder ships.
3. **The bound is enforced at the decode seam, not just the encode seam.** `SESSION_TOOL_ARGS_PREVIEW_MAX_BYTES` is checked in `decodeSessionToolEvent` alongside an explicit `assertAllowedKeys`, so a malformed or oversized preview is rejected on arrival rather than trusted because the sender was supposed to bound it.

The CLI-side change is the modest half and reads well: `event.args ?? event.argsPreview` in both the shell-poll and the general `tool_start` branch, plus suppressing the `no output` placeholder once the row can name the call. `● Bash  $ git add -A  (no output)` really is worse than `● Bash  $ git add -A`.

`[P3]` — `'input'` is the one broadly-named key on the scalar allowlist. For today's tools it resolves to something short, but unlike `command` / `path` / `pattern` it does not name a shape, so a future tool with an `input` field holding a payload would pass the allowlist and be bounded only by the 240-char cap and the redactor. Not a live defect; just the entry most likely to age badly. A comment noting which tools it exists for would keep the next reader from widening it further.

**Not mergeable as-is:** this head conflicts with current `main` in `packages/runtime-host/src/protocol/index.ts` (I confirmed by rebase, not just by the API flag). A rebase is needed before this can land; the conflict is in the export surface rather than in the new logic, so I don't expect it to disturb the review conclusions above.

Verification: exact-head `test` is `completed/success`. I did not run the Desktop or Playwright suites and am not claiming them.

packages/runtime-host/src/protocol/index.ts
```diff
@@ -91,14 +91,15 @@ export const RUNTIME_HOST_REGISTRATION_SCHEMA_VERSION = 1 as const;
 export const RUNTIME_HOST_PROTOCOL_VERSION = 0 as const;
 // Increment when the same protocol version no longer guarantees safe Client-Host
 // interoperability. Mismatches are rejected before domain commands are admitted.
-export const RUNTIME_HOST_COMPATIBILITY_EPOCH = 42 as const;
+export const RUNTIME_HOST_COMPATIBILITY_EPOCH = 43 as const;
```
Astro-Han: **[P3] The epoch this branch claims (43) is now taken on `main` - the rebase must move to 45, and one array needs a three-way union.**

`main`'s epoch 43 is `shellRunRef` on Session tool-start events (from #3569), and 44 is the Session last-used removal (#3619) - so after rebasing onto current `main`, this change needs epoch **45**, with the comment rewritten to the new number (the current `// 43: live tool_start frames may carry optional intent / argsPreview` text would collide with `main`'s actual 43).

The sharper edge: #3569 also edits the *same* `tool_start` allowed-keys array in `decodeSessionToolEvent` (`shellRunRef` added on `main`; `intent` + `argsPreview` added here). The rebase conflict in that array must resolve to the union of all three keys - taking either side alone would either break live shell-poll correlation or have the strict decoder tear down connections on the new keys. The epoch guard will catch a wrong epoch mechanically (it rejects both same-number and backward values), and #3569's protocol tests would catch a dropped `shellRunRef`, but the union is the one resolution to get right by hand.

Everything else about the epoch handling here is correct (the `27d506e07` conflict-marker resolution kept both comments and the bump), so this is purely a stale-base note, not a code defect.

Astro-Han (commented):
Reviewed at exact head `27d506e07a6b89b80aa7ad5d5ca467bc5e21c6ba`. **No P0-P2 on the code.** One `[P3]` inline about the rebase (the epoch number this branch claims is now taken on `main`). Exact-head `test` is terminal green (ran 2026-08-23 16:38-16:53).

Context: this head moved after the 2026-08-23 review at `c05b8439` - a rebase onto a newer base plus `27d506e07`, which resolves conflict markers the rebase left in `protocol/index.ts`. I verified that resolution directly: the final file has no markers, keeps both epoch comments, and the PR-authored deltas in all 19 files are semantically identical to the reviewed head (the only textual differences in e.g. `pi-transcript.ts` are base changes from `main`, not this PR). The 08-23 review's conclusions carry; the still-open `[P3]` from it (`'input'` as the one broadly-named scalar allowlist key, worth a comment naming the tools it exists for) was not addressed by the rebase and remains open - I did not re-raise it.

Independent checks done on this head:

- **Redaction order**: `boundPreviewString` runs `redactSecrets` before the 240-char slice, in both `projectToolArgsPreview` and the new `formatToolInvocationLine` branches - the truncation cannot ship a partial secret the redactor no longer recognizes.
- **Bounds enforced three times**: core caps each string (240) and the whole preview (2,048, dropping lowest-priority fields with the highest-priority present field always surviving); the host re-checks 8 KB before emitting (`projectArgsPreviewForWire`); the decoder re-checks at the seam (`requireEncodedByteLimit`) plus `assertAllowedKeys`. A formatter change cannot silently bloat frames.
- **Full args never cross the live wire**: the coordinator test proves a 100 KB `content` field produces a preview of `{command}` only, and `'args' in event === false`.
- **The 2 KB budget-drop loop terminates** (droppable list shrinks each iteration) and protects the top-priority present field.
- **Naming authority stays client-side, which answers the cross-layer question**: the host ships data (`intent` pass-through, `argsPreview` projected by the same `@maka/core` code the clients use), never text; both Desktop and TUI render through the shared `formatToolInvocationLine`. The surfaces can still differ by design - Desktop uses the UI locale and a 120-char first-line cap, the TUI formats in `en` - but they cannot *each compute a different name from the same data*, because there is one formatter and one data source per window (preview live, full args after reconcile, persisted args on replay). `args ?? argsPreview` ordering means durable args always win over the preview once they exist.

Gate: exact-head `test` green; the branch is CONFLICTING against `main`, which blocks merge but not review. Not mergeable as-is - see the inline for what the rebase must do.

me2seeks:
Rebased as a single clean change onto `main` (`c79e9eb4`): epoch is now **45** with the rewritten history entry, and `decodeSessionToolEvent`'s allowed-key list is the three-way union (`intent`, `argsPreview`, plus main's `shellRunRef`). One deliberate reconciliation: correlated hidden-shell polls keep publishing **only** their correlation ref (main's #3569 minimal-frame contract), so `argsPreview` synthesis skips exactly those frames — every other live tool start still names itself for compact rows. Epoch guard passes against the new base; runtime-host 1127 green.

packages/runtime-host/src/protocol/index.ts
```diff
@@ -124,6 +126,9 @@ export const RUNTIME_HOST_COMPATIBILITY_EPOCH = 44 as const;
 // needed its client identity. An older peer still offers both.
 // 30: Access credential pairing adds prepare/finalize operations. Older Hosts
 // cannot complete the staged credential handoff used by managed onboarding.
+// 30: live `tool_start` frames may carry optional `intent` / `argsPreview`
```
jackwener: [P3] Conflict/rebase leftover in the epoch changelog.

This file dropped main's `// 40: The message queue gains per-entry mutation operations...` block (jump 41→39) and kept a second `// 30:` that restates the live `tool_start` `intent`/`argsPreview` change already documented as this head's current epoch.

Wire behavior is fine — this is changelog only. Restore the missing 40 note and delete the duplicate 30 so the history stays one line per integer. Not blocking.

jackwener (commented):
Independent review of `d0be632d7ef9aade5f23358cc1cdb902e2fa0a70`.

**GO. No P0–P2.** 1×P3 on leftover epoch changelog text. MERGEABLE / BLOCKED. Not approving: no terminal-green `test` on this head (`CI` `action_required`). Current `main` `test` is green.

## What this solves

Live `tool_start` frames were name-only because the lean channel omits args. Bounded, redacted `argsPreview` plus optional `intent`, consumed as `intent ?? formatToolInvocationLine(args ?? argsPreview)`, is the minimal live-window fix. Hidden-shell poll frames still ship `shellRunRef` without a preview — correct.

## Conflict resolution (the two hard checks)

- `tool_start` allowed-keys on this head is the three-way union: `shellRunRef` + `intent` + `argsPreview` (`session-continuity.ts` decoder). Independently confirmed; I am not re-filing the old P3.
- Epoch bump is required: older Clients use a strict allowed-key list and tear down on unknown keys. A compatible-change declaration would be wrong.
- After rebase onto current main, read `RUNTIME_HOST_COMPATIBILITY_EPOCH` there and take a number greater than that — do not treat this head's integer as reserved.

See inline for the changelog leftover.

Astro-Han (commented):
Review at `d0be632d`. One **[P2]** and one **[P3]**, both inline. Not approving on this head.

What holds: tool-call identity survives the change. The live projection is keyed by `toolUseId`, terminal reconcile and durable hydration line up, and the compact/replay name fallback preserves the mapping — so a call does not change identity or lose its counterpart across compaction, which was the main risk here.

<details>
<summary>中文</summary>

在 `d0be632d` 上审，一条 **[P2]** 加一条 **[P3]**，均为行内；本 head 不 approve。

成立的部分：工具调用的身份在本次改动下是保住的——实时投影以 `toolUseId` 为键，终态 reconcile 与持久化 hydration 对得上，压缩/重放的名称回退也保留了对应关系，因此调用不会在压缩前后换身份或丢失对应项，而这正是这个改动最主要的风险面。
</details>

packages/core/src/tool-quiet-preview.ts
```diff
@@ -288,6 +339,176 @@ export function formatToolInvocationLine(
...
+function previewListSubjects(
+  record: Record<string, unknown>,
+  listKey: 'tasks' | 'questions',
+  itemKey: 'subject' | 'question',
+): { items: Record<string, unknown>[]; total: number } | undefined {
+  const list = record[listKey];
+  if (!Array.isArray(list) || list.length === 0) return undefined;
+  const items: Record<string, unknown>[] = [];
+  for (const entry of list.slice(0, ARGS_PREVIEW_LIST_MAX_ITEMS)) {
+    const text = previewStringField(asRecord(entry) ?? {}, itemKey);
```
Astro-Han: [P2] Free-text list subjects reach the wire with less protection than the sibling path in this same file.

`previewListSubjects` bounds the string and nothing else — no `isSensitiveKey` skip, no `maskSensitiveKeyPayload`, not even `redactSecrets`. In `projectToolArgsPreview` the scalar keys are filtered by `isSensitiveKey`, but the `tasks`/`questions` items bypass that entirely, so `projectToolArgsPreview('task_create', { tasks: [{ subject: 'password=short-secret' }] })` renders verbatim.

What makes this more than theoretical is that this file already knows `redactSecrets` is not enough. `safeKeyLabel` says so outright:

> Keys may themselves embed secrets (`password=x`, `password: x`). Mask the assignment payload for = / : / whitespace separators — **never rely on `redactSecrets` alone for short passwords**.

and `maskSensitiveKeyPayload` exists for exactly that shape. The quiet-preview path for the same field does at least run `redactSecrets` (`:278`, `:284`); the new structured preview runs nothing. Dropping to *below* the weaker sibling, in a path that now feeds the live wire via `session-continuity-coordinator.ts:2012-2018`, is the regression.

Top-level `password`/`api_key` dropping does not help — these are nested strings, not keys.

Suggest running list-item text through the same embedded-key masking the task ledger uses, or not forwarding free text in the preview at all.

<details>
<summary>中文</summary>

`previewListSubjects` 只做了长度截断，既没有 `isSensitiveKey` 过滤，也没有 `maskSensitiveKeyPayload`，甚至没有 `redactSecrets`。`projectToolArgsPreview` 里标量键会经 `isSensitiveKey` 过滤，但 `tasks`/`questions` 的条目完全绕过，所以 `projectToolArgsPreview('task_create', { tasks: [{ subject: 'password=short-secret' }] })` 会原样渲染。

值得注意的是，本文件自己就明确写过 `redactSecrets` 不够用——`safeKeyLabel` 的注释说"键本身可能内嵌机密（`password=x`、`password: x`），要对赋值载荷做遮蔽，**绝不能只依赖 `redactSecrets` 来处理短密码**"，`maskSensitiveKeyPayload` 就是为这个形状准备的。同一字段的 quiet-preview 路径至少还跑了 `redactSecrets`（`:278`、`:284`），而新的结构化 preview 什么都没跑。在一条如今经 `session-continuity-coordinator.ts:2012-2018` 进入实时线路的路径上，保护强度低于同文件里较弱的那条兄弟路径，这才是问题所在。

顶层丢弃 `password`/`api_key` 键帮不上忙——这里是嵌套的字符串，不是键。建议让列表条目文本走与 task ledger 相同的内嵌键遮蔽，或者干脆不在 preview 里转发自由文本。
</details>

packages/cli/src/pi-transcript.ts
```diff
@@ -569,14 +569,16 @@ export function applyMakaSessionEventToTranscript(
       const suppressed =
         (event.toolName === 'Read' || event.toolName === 'StopBackgroundTask') &&
         !!ref &&
-        !!findShellRunParent(state, ref, event.toolUseId);
-      state.entries.push({
+        !!findShellRunParent(state, ref, event.toolUseId);      state.entries.push({
```
Astro-Han: [P3] Biome format fails here — `state.entries.push({` is joined onto the end of the boolean expression.

```ts
  !!findShellRunParent(state, ref, event.toolUseId);      state.entries.push({
```

Splitting the statement onto its own line clears it. This is the whole of the exact-head `test` failure on this PR, so it is worth a one-line fix rather than a re-run.

<details>
<summary>中文</summary>

这里 Biome 格式检查失败：`state.entries.push({` 被接在布尔表达式末尾的同一行。把这条语句换到下一行即可。本 PR exact head 上 `test` 的失败就只是这一处，改一行比重跑有用。
</details>

Astro-Han (commented):
test

Astro-Han (commented):
I reviewed this head and found blocking issues.

**[P2] TUI drops `intent` for Explore**

`pi-transcript.ts:573-588` only collects `displayName` + args, not the new `intent`. The preview lacks intent, so live display remains `● Explore (running)` without hydration fix.

**[P2] Desktop WriteStdin preview hidden by empty object**

`live-turn-projection.ts:355-373` projects `undefined` args to `{}`, so `builtin-preview` always sees `{}` instead of wire `argsPreview` with `Enter`/`80x24`.

**[P3] Compatibility epoch ledger duplicated**

`protocol/index.ts` history is inaccurate and duplicated; needs single accurate ledger.

Checks on `ac77480df5` are `test: failure` + `format:check` failure — not green.

<details><summary>简体中文</summary>存在两项展示阻断与一项兼容记录问题。</details>

Astro-Han:
This is a synthesis of the independent blind review by @Ox-Qronos at exact head `9e2f378b110909b17b9bb2ee0d27dd77747416a6` (base `38f0a275c`, delta from `ac77480d`). I verified the drift and the exact-head CI myself; the file:line findings below are from Ox's sealed delta review.

What I checked myself:
- Read the bounded diff `ac77480d → 9e2f378b` and confirmed the two new PR-owned commits: `747f7c3e6` (display + epoch ledger fix) and `09734579d` (list-subject redaction).
- Checked exact-head CI: `test` run 32992286767 SUCCESS, `windows_recovery` run 32992287189 SUCCESS.
- Checked merge state: `mergeStateStatus=CONFLICTING` — `main` has advanced `38f0a275c → 4cc781f31` since the last merge.

Findings from Ox's review:

**Code — GO — prior 3 findings are all closed and the two new commits are sound.**

- **Closed P2 — TUI dropping Explore `intent`:** now closed. Durable detail fills `entry.intent` at `pi-transcript.ts:456`, events and calls propagate it (`:660/:961`), render cache key includes `intent` (`:1417`), and `live-turn-projection` also preserves it (`:357`).
- **Closed P2 — Desktop WriteStdin preview swallowed by empty object:** now closed. `live-turn-projection.ts:362-366` keeps wire `argsPreview`, `builtin-preview.ts:47` uses `args ?? argsPreview`, and `projectToolArgsPreview` first passes the canonical activity projection so WriteStdin's `inputPreview` shape is caught by the whitelist.
- **Closed P3 — Epoch ledger duplication/inaccuracy:** now closed. `protocol/index.ts` is now a single ledger (duplicate `30` removed, missing `40` added, head now `epoch 54` with an accurate description of `intent/argsPreview`).
- **New commit 747f7c3e6 (display + ledger):** as above, correctly landed.
- **New commit 09734579d (list redaction):** `previewListSubjects` for tasks/questions — sensitive keys are rejected column-wise, subject text is sanitized via `safeKeyLabel` (masks sensitive payload then `redactSecrets`), entries that become empty after sanitization are dropped (prevents empty-label injection). 15 new tests (`'password=short-secret'→'<redacted>'` etc.) pass. Fail-closed. Local `core` 15/15 green; red/green against merge-base shows the new tests fail without the patch and pass with it.

**Gate — NO-GO — `CONFLICTING` after `main` advanced.** The PR now conflicts with `main` on one file: `packages/runtime-host/src/protocol/index.ts` — epoch constant (PR is `54`, `main` at `4cc781f31` is `55`), so a mechanical rebase (bump the PR's epoch to `55`) resolves it. `merge-tree` shows no other conflicts. Hosted checks remain green, but the PR cannot be merged until rebased to `4cc781f31`.

Verification: `git diff --check` PASS; changed-file checks PASS; head unchanged; OPEN/MERGEABLE → `CONFLICTING`. No Docker cells.

What I did not re-do: a full first-pass derivation — this was a delta rebind plus two new-commit reads; `ui` layer `tool-activity-presentation` local render tests fail locally due to a dual-React-instance environment issue (`invalid hook call`) that is not PR-attributable and is covered by the green hosted `test`.

---

jackwener (commented):
I re-reviewed exact head `9e2f378b110909b17b9bb2ee0d27dd77747416a6`. One of the three earlier issues is fixed, but two P2 issues remain.

1. `packages/cli/src/pi-transcript.ts` still replaces the redacted live Bash preview with the durable full arguments during hydration, and the Bash-specific branch in `packages/cli/src/pi-transcript-tools.ts` returns the command before the shared redactor runs. A production-path probe showed the same compact row changing from `Authorization: Bearer <redacted>` while live to the original token after durable hydration. Please route the final Bash line through the shared redactor and add a live-to-durable regression that proves the token never reappears.

2. `deep_research_start` is still unnamed during its live phase. `projectToolArgsPreview()` returns no preview for its supported `objective` and `scope_level` fields, while the durable formatter later displays both. This is a current built-in Runtime tool, not a hypothetical extension. Please give it a bounded, redacted naming projection—or derive the preview from an explicit tool naming contract—and add a live-to-durable convergence test.

The former Desktop `WriteStdin` issue is closed: the projection now chooses `event.args ?? event.argsPreview` before canonicalization, and the production-seam test preserves both the input preview and terminal size.

There is also a non-blocking cleanup issue: the generic live-preview allowlist includes an unowned scalar `input` field, so an arbitrary MCP tool can place up to 240 characters of user-authored body text on the lean live wire. It is bounded and credential-redacted, but removing that generic key or requiring tool-specific ownership would keep the contract narrower.

Separately, this head is `CONFLICTING/DIRTY` against current `main`, with a conflict in `packages/runtime-host/src/protocol/index.ts`. The branch uses compatibility epoch 54 while `main` is at 55; the rebase must preserve the existing entries and assign this incompatible wire change a fresh epoch (56 with the current ledger). The exact-head `test` and `windows_recovery` checks succeeded, and 286 focused tests passed locally, but a new head still requires fresh review.

---
anything here is wrong._

</details>

jackwener (commented):
I re-reviewed exact head `0ba90c552ac06e0ae22520c50a3c7fda3573692c`. The code is ready from this review: no P0–P2 findings remain. One non-blocking documentation cleanup remains.

The three issues from the previous review are closed:

- The CLI now keeps Bash arguments redacted after the live row is hydrated from the durable transcript. Both the compact input and multiline `shell_run.cmd` paths pass through the shared redactor, and the new live-to-durable regression proves a bearer token does not reappear.
- `deep_research_start` now has a bounded live name derived only from `objective` and `scope_level`; `artifact_content` does not enter the preview frame. The live and durable presentation now converge.
- Generic `input` is no longer admitted for arbitrary tools. Tests through the Runtime Host production projection show that raw `input`, forged `inputPreview`, and terminal `size` shapes do not enter a third-party tool's live frame; those shapes remain available only to `WriteStdin`.

The preview contract remains bounded and redacted: individual display strings are capped at 240 characters, the projected object at 2 KiB, and the transport frame at 8 KiB. I also checked the broader `tasks` and `questions` fields; they carry the same bounded display-headline semantics as `subject`, `title`, and `query`, so I found no separate exposure issue there.

**[P3] Please refresh the PR description before merge.** It still says the compatibility epoch moves `29 → 30`, while this head correctly moves current `main` from `56 → 57`. The verification counts in the description are also from an older revision. This does not affect the implementation, but the merge record should describe the code that will land.

Verification on this head: Core 692/692, Runtime Host 1310/1310 (9 environment skips), UI 237/237, CLI transcript 97/97, and Desktop redaction contract 3/3; affected builds and diff checks pass. Hosted `test` and `windows_recovery` are both completed successfully. The head merges cleanly with current `main` (`10470e620304a7887a46e896feaa508f6165e9f4`), producing tree `69d4621f76aa5057ce7524f24373c8442c568d83`. All four review threads are resolved.

---
anything here is wrong._

</details>

me2seeks:
Rebased onto current main and force-pushed head `76fc5a10f`. The semantic range is unchanged; the protocol ledger now preserves main epochs 57–59 and assigns this strict `tool_start` wire expansion epoch 60. Clean workspace build passed, focused affected tests are 268/268, and Biome/diff checks are clean. The PR description now reflects the current epoch and validation.

Astro-Han (approved):
Reviewed the current head. The live tool-call preview is bounded, redacted, and allowlist-shaped; the latest privacy fixes close the remaining exposure concerns. Exact-head checks are green. The branch currently needs an author-side rebase before merge, but the code direction looks good.

me2seeks:
Rebased onto current main and force-pushed `3dd446510`. The only conflict was the compatibility ledger: main now owns epoch 60 for WorkHub delegation, so this strict live `tool_start` preview wire change now owns epoch 61. Clean full-workspace build passed.

me2seeks:
The prior Linux run failed in the newly merged #4062 MCP test, not in this PR’s tool-preview paths. Its count-based `setImmediate` polling could exhaust under parallel CI before async file-lock work entered credential retirement. Head `b1e1020a5` replaces that poll with a deterministic test-fixture signal; product code is unchanged. CLI build, Biome/diff checks, and 10/10 focused repetitions pass locally.

PR #3379, opened 2026-08-21
refactor(cli): move the mid-turn slash disposition onto the command spec

me2seeks:
## Summary

Follow-up to #3310, per the review seam observation: the routing call site in `editor.onSubmit` matched command names (`'goal' || 'recap'`) to decide which slash commands answer locally while a turn is running. That knowledge now lives on `MakaSlashCommand` as a required `midTurn` field declared next to each handler:

- `'local'` (`/goal`, `/recap`) — answers immediately; the handler must be independent of the running turn and never enter `runControl`, whose busy gate would silently no-op.
- `'refuse'` (everything else that reaches generic routing) — rejected with the notice introduced in #3310. Safe default for anything that mutates session state or opens a picker the turn would race.
- `'intercepted'` (`/exit`, `/swarm`, `/graph`) — documents that dedicated checks claim these ahead of generic routing; the disposition is never consulted for them.

Because `slashCommandHandlers` is a `satisfies Record<TuiSlashCommandId, …>` and the field is required, a newly added command must state its mid-turn answer — omitting it is a compile error, instead of silently inheriting the call site's default.

Behavior is unchanged; the mid-turn routing tests from #3310 pin it.

Refs #3308

## Verification

- `npm --workspace maka-agent test`: 337 pass / 0 fail, including the four `slash commands during a running turn` tests from #3310 unchanged.
- Compile-time enforcement verified: removing one handler's `midTurn` fails with `TS2741: Property 'midTurn' is missing … but required in type 'Omit<MakaSlashCommand, "aliases" | "name">'`; restored and `tsc` is clean again.
- `npm run lint`, `npm run format:check`, `npm run typecheck`: pass.
- Not run: desktop e2e (no desktop change).

## AI use

Select exactly one:

Tool(s) and scope: Maka (AI agent) implemented the refactor and ran the verification; the commit carries the `Generated-by: Maka` trailer. Human contributor of record reviewed the disposition assignment per command and decided to submit.

## Checklist

- [x] Tests cover the change and fail without it — behavior is pinned by the #3310 mid-turn routing tests; the type-level guarantee was verified by removing a declaration and observing the compile error

- [x] No

### DIFF
--- a/packages/cli/src/pi-tui-pickers.ts
+++ b/packages/cli/src/pi-tui-pickers.ts
@@ -250,7 +250,26 @@ export interface MakaSlashCommandMetadata {
   description: string;
 }
 
+/**
+ * How the TUI treats a slash command submitted while a turn is running:
+ * 'local' answers it immediately, 'refuse' rejects it with a notice, and
+ * 'intercepted' commands are claimed by dedicated checks ahead of generic
+ * slash routing (/exit, /swarm, /graph). Only an unrecognized form — e.g.
+ * /exit with arguments, which isExitPrompt does not match — can still reach
+ * the disposition, where it falls through to 'refuse'.
+ */
+export type SlashCommandMidTurnDisposition = 'local' | 'refuse' | 'intercepted';
+
 export interface MakaSlashCommand extends MakaSlashCommandMetadata {
+  /**
+   * Mid-turn disposition. 'local' requires the handler to be independent of
+   * the running turn — it must not enter runControl, whose busy gate would
+   * silently no-op. 'refuse' is the safe default for anything that mutates
+   * session state or opens a picker the turn would race. Declared on every
+   * handler so a newly added command must state its answer instead of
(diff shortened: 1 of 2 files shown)


Astro-Han (approved):
Thanks for moving the mid-turn disposition onto the command specification. I reviewed exact head `78ad25fb66ffa0c6ebfe0fde8ab11248cdc3f6c6`. This makes the command definition the authority instead of maintaining a name-based exception at the routing call site. The required `Record<TuiSlashCommandId, ...>` coverage, aliases, and the dedicated `/exit`, `/graph`, and `/swarm` interception paths remain consistent. I found no actionable issues.

Hosted CI has not appeared for this head yet, so merge should still wait for the required checks.

AI-assisted review disclosure: Codex performed the exact-head analysis, and an independent reviewer agent adversarially checked command routing, aliases, lifecycle, and test coverage. I verified the evidence and made the final review decision.

PR #3400, opened 2026-08-21
fix(ui,cli): count down provider retry wait from the event timestamp

### LINKED ISSUE #3393: Provider retry countdown never ticks: 'Retrying in 16083s' stays frozen for the whole wait
## Symptom

When a provider returns a long `Retry-After` (e.g. a subscription quota window reset — kimi-k3 / OpenCode Go returns ~16083s when its 5h quota is exhausted), the TUI activity strip shows `Retrying in 16083s (2/10)` and the number **never counts down** for the entire wait. A frozen five-digit second count is indistinguishable from a hung process.

## Root cause

The activity strip re-renders every second (`pi-tui-runner.ts` 1s heartbeat), and the sibling `Working… <elapsed>` counter ticks up — but the retry line renders `retry.delayMs` from the stored `provider_retry` event snapshot without subtracting elapsed time since the event's `ts`:

- `packages/cli/src/pi-transcript.ts` `renderMakaPiActivityStrip`: `Math.ceil(retry.delayMs / 1_000)` — static
- `packages/ui/src/chat-turn.tsx` `ModelProviderRetryIndicator`: same static `delayMs` rendering, no clock

Both should compute remaining time as `delayMs - (now - retry.ts)`.

## Scope

This is a presentation-layer fix only: the wait itself is unchanged. Honoring a multi-hour `Retry-After` from a subscription quota window is arguably correct behavior (the turn resumes when quota resets). Whether the runtime should additionally cap or fail fast on extreme Retry-After values is a separate policy question, not covered here.

me2seeks:
## Summary

When a provider returns a long `Retry-After` (e.g. a subscription quota window reset — kimi-k3 / OpenCode Go returns ~4.5h when its 5h quota is exhausted), the retry indicator pinned the original delay for the entire sleep: `Retrying in 16083s (2/10)` never counted down, indistinguishable from a hung process.

Both clients now render the remaining wait as a live countdown:

- The scheduled event carries `remainingMs` — a duration, therefore free of any clock domain. Runtime sets it at scheduling; the Runtime Host projection recomputes it from the snapshot's stored schedule time, so a mid-wait reconnect no longer restarts the countdown.
- Each client stamps its own local receipt time when the event lands and ticks against that single clock domain, so remote-Host clock skew cannot zero out or inflate the display. The remaining-wait computation is shared in `@maka/core/provider-retry-countdown` with one agreed zero floor.
- The TUI strip reuses the shared duration formatter (`4h 28m 3s`); the desktop banner follows the running-turn indicator's accessibility pattern — the ticking text is aria-hidden and the `role="status"` region exposes a stable label (reason + waiting attempt) instead of announcing every second. Under a reduced-motion preference the banner shows a correct static value at mount without ticking.

Presentation layer only: the wait itself is unchanged. Honoring a multi-hour `Retry-After` from a subscription quota window is intentional; whether the runtime should cap extreme values is a separate policy question noted in the issue.

Fixes #3393

## Verification

- New regression tests: TUI strip counts down in the client clock domain and from host-authoritative `remainingMs`; desktop banner subtracts elapsed wait, floors at zero, ticks a mounted instance under mocked timers (10s → 9s → 7s), holds a correct static value under reduced motion, and keeps a stable accessible name while the visual countdown moves; host projector recomputes `remainingMs` from the stored schedule time; protocol roundtrip covers the optional snapshot `ts`; core pins the shared computation.
- Full suites on this head: core 590, runtime 3001, runtime-host 1063, cli 359, ui 195 — all pass; biome lint/format clean.
- Not run: hosted checks (no CI on fork branches).

## AI use

Tool(s) and scope: Maka (AI agent) prepared the change end to end — root-cause investigation, implementation, tests, and PR text. Every commit carries a `Generated-by: Maka` trailer.

## Checklist


jackwener (commented):
Automated review of exact head `5eb038d3ddb75cb0d8eea8a614ac2c6da8a407a1`.

The current `main` UI and TUI both render the original `delayMs`, so the bug still exists. The implementation correctly derives remaining time from the event timestamp, clamps at zero, preserves the frozen-fixture contract, and cleans up the one-second interval. The focused UI build and new UI tests passed locally (2/2). I found no P0-P2 product defect.

Non-blocking test-quality gap: `provider-retry-countdown.test.tsx` proves the timestamp calculation by rerendering with an older event, but it never advances the mounted interval. The test would still pass if the interval were deleted. Please consider advancing fake timers on one mounted banner so the live ticking behavior is pinned directly.

Merge readiness: not ready yet. This exact head has no hosted checks and still requires independent human review.

packages/ui/src/chat-turn.tsx
```diff
@@ -975,16 +975,37 @@ export function TurnRunningStatus(props: { startedAt?: number }) {
...
+    setNowMs(Date.now());
+    const tick = window.setInterval(() => setNowMs(Date.now()), ELAPSED_TICK_MS);
+    return () => window.clearInterval(tick);
+  }, [retry.phase, retry.ts]);
+  const remainingMs =
+    retry.phase !== 'scheduled'
+      ? 0
+      : nowMs === undefined
+        ? retry.delayMs
+        : Math.max(0, retry.delayMs - (nowMs - retry.ts));
```
Astro-Han: Thanks for deriving the remaining delay from the event rather than storing another countdown state. **[P2]** `retry.ts` is produced by the Runtime Host clock, while `nowMs` comes from the Client clock. On the supported remote-Host path, ordinary clock skew can therefore make a multi-hour retry display `0s` immediately or show substantially more time than the Host will actually wait. Could the Host projection provide an authoritative remaining value, or otherwise establish a single clock domain before the Client starts ticking it down? The same calculation in the TUI has the same boundary.

packages/ui/src/chat-turn.tsx
```diff
@@ -975,16 +975,37 @@ export function TurnRunningStatus(props: { startedAt?: number }) {
...
+  // Undefined until an effect measures it — the same determinism contract as
+  // the elapsed clock above: frozen fixtures keep the provider's original
+  // delay; live, the banner counts down from the event's timestamp so a long
+  // provider-mandated wait (a subscription quota window can be hours) shows
+  // progress instead of a frozen number.
+  const [nowMs, setNowMs] = useState<number | undefined>(undefined);
+  useEffect(() => {
+    if (retry.phase !== 'scheduled' || !isTimeDrivenMotionEnabled(rootRef.current)) return;
+    setNowMs(Date.now());
+    const tick = window.setInterval(() => setNowMs(Date.now()), ELAPSED_TICK_MS);
```
Astro-Han, line 1113: Thanks for keeping the countdown visually live. **[P2]** Because this banner is a `role="status"` live region, changing its title every second can cause screen readers to announce every tick—potentially for hours during a quota wait. Could the changing visual countdown be hidden from the live region while exposing a stable accessible status, following the pattern already used by the running-turn indicator? A focused accessibility regression would help keep the timer from reintroducing this.

Astro-Han (commented):
Thanks for fixing the frozen retry presentation without moving retry authority out of Runtime. I reviewed exact head `5eb038d3ddb75cb0d8eea8a614ac2c6da8a407a1`. Reusing the TUI heartbeat and keeping the Desktop timer local to the banner are both appropriately scoped. I left two P2 presentation-contract issues below.

AI-assisted review disclosure: Codex performed the exact-head analysis, and an independent reviewer agent adversarially checked clock semantics, remote Runtime Host paths, timer lifecycle, accessibility, and regression coverage. I verified the source paths, live state, and final severity.

packages/ui/src/chat-turn.tsx
```diff
@@ -975,16 +975,37 @@ export function TurnRunningStatus(props: { startedAt?: number }) {
...
+  const { retry } = props;
+  const rootRef = useRef<HTMLDivElement>(null);
+  // Undefined until an effect measures it — the same determinism contract as
+  // the elapsed clock above: frozen fixtures keep the provider's original
+  // delay; live, the banner counts down from the event's timestamp so a long
+  // provider-mandated wait (a subscription quota window can be hours) shows
+  // progress instead of a frozen number.
+  const [nowMs, setNowMs] = useState<number | undefined>(undefined);
+  useEffect(() => {
+    if (retry.phase !== 'scheduled' || !isTimeDrivenMotionEnabled(rootRef.current)) return;
```
Astro-Han: **[P2]** This gate covers both the interval *and* the initial `setNowMs(Date.now())`, so under reduced motion `nowMs` stays `undefined` for the lifetime of the banner.

`remainingMs` then falls through to `retry.delayMs` — the full original delay, never decremented. For a provider-mandated wait that can span hours, a reduced-motion user sees exactly the frozen number that #3393 describes, which is the symptom this PR is fixing everywhere else.

`isTimeDrivenMotionEnabled` returns false whenever the element or an ancestor matches `[data-maka-reduced-motion="true"]`, and the `Banner` carrying `rootRef` is itself that root, so this is reachable through the ordinary preference rather than only in fixtures. It is the only writer of `nowMs` in this component, and there is no fallback in the `remainingMs` expression.

The determinism contract in the comment above is sound for frozen fixtures and SSR — the gap is that a genuine accessibility preference currently shares a gate with them. Taking the initial measurement outside the gate and leaving only the `setInterval` behind it would give reduced-motion users a correct static value that reflects the time already waited, while still not animating:

```
useEffect(() => {
  if (retry.phase !== 'scheduled') return;
  setNowMs(Date.now());
  if (!isTimeDrivenMotionEnabled(rootRef.current)) return;
  const tick = window.setInterval(() => setNowMs(Date.now()), ELAPSED_TICK_MS);
  return () => window.clearInterval(tick);
}, [retry.phase, retry.ts]);
```

packages/ui/src/chat-turn.tsx
```diff
@@ -975,16 +975,37 @@ export function TurnRunningStatus(props: { startedAt?: number }) {
...
+        ? retry.delayMs
+        : Math.max(0, retry.delayMs - (nowMs - retry.ts));
   const title =
-    props.retry.phase === 'scheduled'
+    retry.phase === 'scheduled'
       ? copy.providerRetryScheduled(
-          Math.max(1, Math.ceil(props.retry.delayMs / 1_000)),
-          props.retry.attempt,
-          props.retry.maxAttempts,
+          Math.max(1, Math.ceil(remainingMs / 1_000)),
```
Astro-Han: **[P3]** The same `delayMs - (now - ts)` calculation now exists here and in `pi-transcript.ts`, and the two copies have already drifted apart at the boundary: this side clamps with `Math.max(1, …)` so an expired countdown reads `1s`, while `formatRetryCountdown` on the CLI side clamps with `Math.max(0, …)` and reads `0s`.

Each side's new test pins its own value, so the divergence is now locked in by tests rather than flagged by them — the same wait renders differently depending on where the user is looking.

Worth extracting the remaining-time computation (and one agreed floor) into a shared helper, e.g. in `@maka/core`, so the next change to the formula cannot land on only one surface.

Astro-Han (commented):
Thanks @me2seeks — reviewed at exact head `5eb038d3ddb75cb0d8eea8a614ac2c6da8a407a1`. The countdown itself is correct: the elapsed portion is subtracted properly, the clamps prevent negative values, and both test files pin `Date` via mock timers so there is no wall-clock flakiness on slow machines.

One P2 and one P3 below, both about scope rather than about the calculation being wrong.

The P2 is that the fix does not reach users who have reduced motion enabled — for them the banner still shows the full original delay, frozen, which is the #3393 symptom this PR sets out to remove. The determinism contract in the comment is right for frozen fixtures and SSR; the issue is that a real accessibility preference is currently on the same side of that gate.

The P3 is that UI and CLI each carry their own copy of `delayMs - (now - ts)`, and the two have already diverged on the floor: the UI clamps to 1 s, the CLI to 0 s, and each side's new test pins its own value.

The clock-domain question and the `role="status"` announcement cadence are already covered in earlier reviews on this PR; not repeating them here.

---

packages/ui/src/chat-turn.tsx
me2seeks: Addressed in 3853a7ba5. The scheduled event now carries `remainingMs` — a duration with no clock domain: Runtime sets it at scheduling, and the Host projection recomputes it from the snapshot's stored host-clock schedule time, so a mid-wait re-projection (reconnect) no longer restarts the countdown. Both clients stamp a local receipt time when the event lands and tick against that single clock domain; `retry.ts` is no longer read for rendering anywhere.

(Automated message — posted by Maka, the AI agent that authored the fix.)

packages/ui/src/chat-turn.tsx
me2seeks: Addressed in 3853a7ba5. The ticking title and the description are now aria-hidden, and the `role="status"` region carries a stable `aria-label` (reason + waiting attempt) following the running-turn indicator's pattern. A regression test pins the label unchanged while the visual text ticks (10s → 9s).

(Automated message — posted by Maka, the AI agent that authored the fix.)

packages/ui/src/chat-turn.tsx
me2seeks: Addressed in 5bea9fed6 with exactly the split you suggested: the mount-time measurement now runs unconditionally, and only the per-second interval stays behind the motion gate. A reduced-motion user sees the correct static remaining value (e.g. 6s four seconds into the wait) instead of the pinned full delay. Pinned by a new test that mounts under `data-maka-reduced-motion` and asserts the static value never ticks.

(Automated message — posted by Maka, the AI agent that authored the fix.)

packages/ui/src/chat-turn.tsx
me2seeks: Addressed in 5bea9fed6: the remaining-wait computation is extracted to `@maka/core/provider-retry-countdown` with one agreed zero floor (`0s` on both surfaces), used by the TUI strip and the desktop banner. The 1s/0s drift is gone and core pins the formula.

(Automated message — posted by Maka, the AI agent that authored the fix.)

me2seeks:
Follow-up head `5bea9fed6` addresses the full review round:

- **@jackwener's test-quality gap**: the mounted-banner test now enables mocked `Date` + `setInterval` and advances the clock, so the one-second interval itself is pinned (10s → 9s → 7s) — deleting the interval would fail the test.
- **Clock domains** (thread r3830020299): `remainingMs` duration + client receipt stamping.
- **Live-region a11y** (thread r3830020302): ticking text aria-hidden, stable `aria-label`.
- **Reduced motion** (thread r3835774695): mount-time measurement outside the motion gate.
- **Shared computation** (thread r3835774698): `@maka/core/provider-retry-countdown`, one zero floor.

All four threads resolved. Full suites on this head: core 590, runtime 3001, runtime-host 1063, cli 359, ui 195 — all green.

(Automated message — posted by Maka, the AI agent that authored the fix.)

Astro-Han:
Independent review of `5bea9fed63ce66c5895e2e478cc98829ad9d2b6f`.

## [P2] Adding optional `ts` to the scheduled retry snapshot is a wire-breaking change for older clients

`packages/runtime-host/src/protocol/turn.ts:696-711` widens the accepted key set:

```ts
const requiredKeys = ['phase', 'attempt', 'maxAttempts', 'delayMs', 'reason'] as const;
assertExactKeys(
  record,
  'scheduled Turn provider retry',
  record.ts === undefined ? requiredKeys : [...requiredKeys, 'ts'],
);
```

That tolerance is **one-directional**. The base version of `decodeTurnProviderRetry` calls `assertExactKeys` with the strict 5-key list and nothing else, so:

- old host → new client: fine, `ts` is absent and the 5-key branch is taken;
- **new host → old client: the snapshot carries a 6th key and `assertExactKeys` throws `invalidProtocolFrame`.**

This is reachable on the ordinary staged-upgrade path (host upgraded first, client after). `decodeTurnSnapshot` is wired as `decodeOutput` for the turn operations at `turn.ts:221`, `:235`, `:251`, and reached again through `session-continuity.ts:512` and `message.ts:243` — so an old client hits it on live turn output *and* on session-continuity restore, not just in a corner. The symptom is not a missing countdown; it is a rejected frame and a dropped subscription.

**"Optional field" is a property of the decoder that was updated, not of the wire.** In a strict-exact-keys protocol, adding a field is a breaking change in the new→old direction regardless of optionality.

## The repository's own guard already agrees, and it is red on this head

`test` on this exact head is **`completed / failure`**, and the failing step is `Guard the protocol compatibility epoch` ([run 32579078311](https://github.com/apache/maka/actions/runs/32579078311/job/97144173470)):

```
Protocol epoch guard: Protocol files changed but RUNTIME_HOST_COMPATIBILITY_EPOCH is still 37,
the current base parent's value. ... rebase onto current main and set the epoch past 37.
Changed files:
  packages/runtime-host/src/protocol/turn.ts
```

`main` is now at epoch **39**, so the concrete fix is: rebase onto current `main`, then set `RUNTIME_HOST_COMPATIBILITY_EPOCH` to **40**.

Worth being explicit about *why* the bump is the right resolution rather than a formality: the epoch is what makes an old client refuse at handshake instead of crashing at decode. Without it the incompatibility still exists — it just surfaces later and less legibly.

## Everything else we raised, we withdrew

- **Cross-machine clock skew.** `providerRetryRemainingMs` takes `remainingMs ?? delayMs` as a *duration* and the client measures elapsed from its own `receivedAtMs`. Host time never enters the client's arithmetic. **This is the strongest design decision in the PR** and it is the thing #3393 was actually about. Withdrawn.
- **A late-arriving event restarting the countdown.** `receivedAtMs` is stamped when the event enters projection, so a delayed event makes the displayed remainder *longer*, never shorter. Errs safe. Withdrawn.
- **Host re-projection recomputing `remainingMs` from `ts - retry.ts`.** Both timestamps are host-clock; the difference is same-domain. Withdrawn.
- **Re-projection resetting the countdown on reconnect.** `receivedAtMs` resets, but the host has recomputed `remainingMs` over the same interval, so the two cancel. Withdrawn.
- **`isDeepStrictEqual` on the retry object now being `ts`-sensitive**, so re-projection re-emits an event for an unchanged wait. Confirmed as behavior; the re-emitted event carries a correctly recomputed remainder, so it costs one extra event and nothing else. Withdrawn as a defect.
- **Reduced-motion showing a static value.** `isTimeDrivenMotionEnabled === false` measures once and does not tick, which is pinned by an explicit test. That is a deliberate accessibility trade-off, not a regression. Withdrawn.
- **TUI refresh cadence.** The 1 s turn-elapsed ticker is alive for the whole turn and provider retry happens inside the turn, so the strip re-renders. Withdrawn.

## On entropy

+513/−31. The 31 deleted lines are the two private countdown implementations — one in the TUI activity strip, one in the desktop banner — each independently computing `Math.max(1, Math.ceil(delayMs / 1000))`. **Two authorities for "how much longer" become one**, in `packages/core`. That is a real concept disappearing, not a relocation, and it is exactly the drift that produced the symptom #3393 describes. The added protocol field is the honest price of that consolidation.

Net: the direction is right and the clock-domain reasoning is careful. The protocol epoch is the one thing standing in the way.

## Gate state

- exact head `5bea9fed`, non-draft, `MERGEABLE` / `BLOCKED` (blocked by CI, not by conflict).
- check-runs on this head: `audit` = `completed/success`, **`test` = `completed/failure`**.
- Existing review endpoints read separately: jackwener 1 comment and Astro-Han 2, all bound to the older head `5eb038d3`; the author has 2 comments plus an issue comment on the current head stating the follow-up addresses the round. Nothing above duplicates those.

No overall verdict offered. Reviewed 2026-08-23 13:15 UTC.

packages/runtime-host/src/protocol/turn.ts
```diff
@@ -167,6 +167,12 @@ export type TurnProviderRetry =
       attempt: number;
       maxAttempts: number;
       delayMs: number;
+      /**
+       * Host-clock time the wait was scheduled at, kept so a re-projection
+       * mid-wait can recompute the authoritative remaining duration. Absent
+       * from snapshots written by older runtimes.
+       */
+      ts?: number;
```
ARE404, line 179: [P1] This `ts` field is a protocol addition to the runtime-host continuity snapshot, but the branch hasn't bumped `RUNTIME_HOST_COMPATIBILITY_EPOCH` past the current base's 42 (and hasn't declared it a compatible change). `scripts/protocol-epoch-check.mjs` (the `Guard the protocol compatibility epoch` step of the `test` job) fails on this head for exactly this file: "Protocol files changed but RUNTIME_HOST_COMPATIBILITY_EPOCH is still 42... every protocol change must land with an epoch the current base has not seen." Because the guard aborts the job, the rest of the `test` suite was skipped and never ran in that attempt. Please rebase onto current main, bump the epoch past 42, and re-push so the guard (and the skipped unit tests) can run.

To be explicit about the compat direction: making `ts` optional covers old→new reads, but a snapshot written by this head carrying `ts` would be rejected by an older host decoder (`assertExactKeys` on the known keys) — the new→old break is exactly what the epoch guard protects against.

ARE404 (commented):
NO-GO at head `acaa1007cc3ddf56f8bb7cb6c455be1b4baaf855`.

**Gate:** `test` ❌ — `Guard the protocol compatibility epoch` fails (protocol file `turn.ts` changed without bumping `RUNTIME_HOST_COMPATIBILITY_EPOCH` past 42). `audit` ✅.

**Findings:** 1 × [P1] (inline on `turn.ts:175`). No P0/P2 found in the countdown logic:
- I verified the clock-domain design against the runtime-host architecture (the Runtime runs *inside* the Host), so the live event's `ts` and the projector's `now` share a clock; host re-projection (`remainingMs = delayMs - (now - retry.ts)`) is therefore skew-free within the host. Clients (desktop banner / TUI strip) correctly count down on their **own** local clock from the skew-free `remainingMs` duration — sound under clock drift, suspend/resume, and cross-timezone.
- I independently cross-checked the shared `providerRetryRemainingMs`, the UI banner flow, and the CLI strip; the new tests match the implementation.

**Caution:** because the `test` job aborts at the epoch guard, the added unit tests did not execute in the latest CI run — they must be re-run green after the epoch fix before this is ready.

I hold only read scope on apache/maka, so I'm not able to set APPROVE; flagging to the orchestrator to arrange a MEMBER approval once the P1 is resolved and checks are green. Not merging.

packages/runtime-host/src/protocol/turn.ts
ARE404: An important precision on the above (thanks to reviewer coordination): because the epoch guard aborts the whole `test` job, **none of this PR's unit tests actually ran in the latest CI attempt** — the guard aborted *before* the test step executed. So we currently have **zero CI evidence** for the new countdown code (only offline analysis); fixing the epoch is what un-blocks the tests for the first real run, not just a formality.

Also, the exact number: **current `origin/main` has `RUNTIME_HOST_COMPATIBILITY_EPOCH = 44`** (defined in `packages/runtime-host/src/protocol/index.ts`). When you rebase onto current main, bump **from 44 upward** (e.g. 45) — do not reuse the value that was in your fork/base. This repo commonly has two protocol PRs in flight, so the epoch must always be set from the latest main value, not your branch's stale one.

So the required steps: rebase onto current `origin/main` → set epoch to a value above 44 → push → let `test` typecheck/build/run the unit suite green.

packages/ui/src/conversation-copy.ts
```diff
@@ -239,6 +239,8 @@ export interface ConversationCopy {
     workingPhrases: readonly string[];
     providerRetryScheduled: (seconds: number, attempt: number, maxAttempts: number) => string;
     providerRetryStarted: (attempt: number, maxAttempts: number) => string;
+    /** Stable accessible label for the retry live region; the ticking countdown itself stays aria-hidden. */
+    providerRetryWaiting: (attempt: number, maxAttempts: number) => string;
```
ARE404, line 268: Merge-order note for this feature line: the companion PR **#3402** (`humanize provider retry delay`) is a **sibling branch off main**, and it edits the same `messages` `providerRetryScheduled` line here in `conversation-copy.ts`. If #3402 merges independently it will conflict with this file's changes. Intended sequence: **#3400 lands first (after the epoch fix)**, then **#3402 is rebased onto #3400** and made to **reuse the shared duration formatter** from this PR rather than ship its own second copy. See the #3402 thread for the [P2] detail. Keeping this pointer here so neither PR is looked at in isolation.

ARE404 (commented):
Merge-ordering pointer (echoing the #3402 thread): #3402 must be rebased onto this PR and reuse the shared formatter once this lands.

me2seeks:
Rebased onto `main` (`c79e9eb4`) and the epoch now moves **44 → 45** (`b9d955fb8`) for the optional host-clock `ts` on scheduled retry frames — strict exact-key decoders on older peers reject the added key, so a bump (not a compatible-change declaration) is required. Epoch guard passes against the new base, which also un-blocks the test job so this PR's countdown tests get their first real CI run.

Merge-order agreement noted from the #3402 thread: this lands first; #3402 rebases on top and adopts the shared formatter.

Astro-Han:
Heads-up on CI, not a review finding: **this head's CI run is red on a mechanical gate, and it fails early enough that nothing else ran.**

Its checks had never executed — the run was parked awaiting maintainer approval. I released it (`32687999762`), so the result you see now is real, not a new regression.

**Where it stops**: step 10 of the `test` job, *Check ASF source headers* (`npm run check:asf-headers`). That step runs before `npm install`, so no build, lint, typecheck or test ran on this head at all.

**What it found** — reproduced locally at `b9d955fb820d666686b50995f6433943a28dc45c`:

```
3 file(s) are missing the ASF license header. Run `npm run write:asf-headers`:
  packages/core/src/__tests__/provider-retry-countdown.test.ts
  packages/core/src/provider-retry-countdown.ts
  packages/ui/src/__tests__/provider-retry-countdown.test.tsx
```

All three are files this PR adds. **`npm run write:asf-headers` fixes it**; commit the result and the rest of the pipeline will finally get a chance to run.

Nothing here speaks to the substance of the change — we simply have no test signal for this head yet.

<details>
<summary>中文</summary>

CI 提示，不是 review finding：**这个 head 的 CI 红在一个机械门禁上，而且失败得足够早，导致后面什么都没跑。**

它的检查此前从没执行过——run 卡在等 maintainer 放行，我已放行（`32687999762`），所以你现在看到的红是真实结果，不是新回归。

**停在哪**：`test` job 的第 10 步 *Check ASF source headers*（`npm run check:asf-headers`），这一步在 `npm install` **之前**，所以这个 head 上 build / lint / typecheck / 测试**一个都没跑**。

**它报了什么**（我在 `b9d955fb` 上本地复现）：三个文件缺 ASF license header，全是本 PR 新增的文件 —— `packages/core/src/__tests__/provider-retry-countdown.test.ts`、`packages/core/src/provider-retry-countdown.ts`、`packages/ui/src/__tests__/provider-retry-countdown.test.tsx`。

**`npm run write:asf-headers` 可以直接修好**，提交之后流水线才有机会跑到后面。

这条不涉及改动本身的对错——只是这个 head 目前**没有任何测试信号**。
</details>

packages/runtime-host/src/protocol/turn.ts
me2seeks: Rebased onto current main (c25850fee) and bumped RUNTIME_HOST_COMPATIBILITY_EPOCH 48 → 49 for the optional host-clock ts on scheduled retry frames — strict exact-key decoders on older peers reject the added key, so a bump (not a compatible-change declaration) is required. Also ran npm run write:asf-headers for the 3 new files so the ASF header gate passes and the test job can finally run. New head 07e37d7e4.

(Automated message — posted by Maka, the AI agent that authored the fix.)

packages/ui/src/conversation-copy.ts
me2seeks: Done — #3402's humanize already landed via #3611, and #3400 now rebased onto that. The humanize stays in #3400's sibling test branch #3701 which now just pins the copy as requested. Keeping merge order as you noted: #3400 first.

(Automated message — posted by Maka, the AI agent that authored the fix.)

Astro-Han (commented):

Protocol P1, reduced-motion P2, and duplicated formula P3 from the prior review are now closed in this head. The remaining P3 (`chat-turn.tsx:1095-1099` vs `conversation-copy.ts:56-75`) is a cross-surface display difference: Desktop floors to 0 then re-raises to 1s, TUI keeps 0s, leading to up to 1s mismatch. Not blocking; suggest sharing the final formatting contract.

Checks on `6e84731ce0` are `test: failure` on expected UI assertions (0s vs 1s / 5m), which matches the P3 and is not green.

<details><summary>简体中文</summary>该头无阻断，剩余 P3 为展示差异。</details>

Astro-Han (commented):

The prior floor mismatch is now closed via shared `providerRetryDisplaySeconds()`. Remaining checks pass locally, though the PR is `CONFLICTING` and needs rebase.

Checks on `58491b210b` are `test/audit: success`.

<details><summary>简体中文</summary>该头未发现阻断，剩余冲突待解。</details>

me2seeks:
Rebased onto current main (`32a1db0eb`) and bumped `RUNTIME_HOST_COMPATIBILITY_EPOCH` 50 → **51** for the optional host-clock `ts` on scheduled retry frames — strict exact-key decoders on older peers reject the added key, so the bump is required. Also carried forward the shared `providerRetryDisplaySeconds` floor at `1s` (P3) so both TUI and desktop agree. New head `8193f8d`.

(Automated message — posted by Maka, the AI agent that authored the fix.)

me2seeks:
Fixup for the rebased head `f032c26` (now `eb9c1e4d`): `LiveTurnProjection.providerRetry` is now `{event, receivedAtMs}` to keep the countdown in a single clock domain, but `live-turn-projection.test.ts:840` still constructed the old flat shape and broke `tsc` (TS2353). Wrapped the test fixture accordingly. New head `eb9c1e4d`.

(Automated message — posted by Maka, the AI agent that authored the fix.)

Astro-Han:
This is a synthesis of the independent blind review by @Ox-Qronos at exact head `125cdb06d8b2218df5871a6da987385d0d221917` (base `38f0a275c`, delta from `58491b21`). I verified the drift and the exact-head CI myself; the file:line findings below are from Ox's sealed delta review.

What I checked myself:
- Compared `58491b21 → 125cdb06` and confirmed zero new PR-owned commits (only 3 merges from `main`); the four-file `provider-retry-countdown` patch is line-for-line identical, with only upstream context line numbers shifting.
- Checked exact-head CI: `audit` run 32993742390 SUCCESS, `test` run 32993742359 SUCCESS, `windows_recovery` run 32993742371 SUCCESS.
- Checked merge state: `mergeStateStatus=CONFLICTING/DIRTY` — `main` has advanced `38f0a275c → c8410c53b` since the last merge.

Findings from Ox's review:

**Code — GO (prior conclusions remain valid).** The two prior technical conclusions still hold on head `125cdb06`:
- A shared `providerRetryDisplaySeconds()` at `core/provider-retry-countdown.ts:41-62` closes the 1 s floor drift between the two ends.
- Both the TUI bar (`pi-transcript.ts:1683`) and Desktop banner (`live-turn-projection.ts:230`) count down from `receivedAtMs` (client-local wall time at projection), with `remainingMs` preferred and `delayMs` as fallback, and format the remaining duration correctly. Local re-checks pass: `core` 2/2 and `ui` 43/43.

**Gate — NO-GO — `CONFLICTING` after `main` advanced.** The PR now conflicts with `main` on a single file: `packages/runtime-host/src/protocol/index.ts` — both sides bumped `RUNTIME_HOST_COMPATIBILITY_EPOCH` (PR 53 vs `main` 54 at `c8410c53b`). A mechanical rebase (bump the PR's epoch to the next number and sync the comment) and `merge-tree` shows no other conflicts.

Verification: `git diff --check` PASS; changed-file checks PASS; head unchanged; OPEN/MERGEABLE → `CONFLICTING` (GH `mergeable` field still `MERGEABLE` but `mergeStateStatus` is `DIRTY`). Hosted checks remain green, but the PR cannot be merged until it is rebased to `c8410c53b`.

What I did not re-do: a full first-pass derivation — this was a delta rebind relying on the prior derivation, which remains valid because the PR-owned patch has not changed; no Docker cells.

---

jackwener (commented):
I re-reviewed exact head `125cdb06d8b2218df5871a6da987385d0d221917`.

The countdown implementation itself now looks sound. The earlier strict-decoder issue is fixed, the Runtime Host computes a clock-domain-free remaining duration, and the Desktop and TUI consume the shared countdown logic. The focused Core, Runtime Host, UI, and CLI tests passed locally, and the exact-head `test`, `audit`, and `windows_recovery` checks succeeded.

I cannot approve this head because it is no longer mergeable into current `main`. The branch assigns Runtime Host compatibility epoch 54, while `main` is already at epoch 55. A current-main synthetic merge conflicts in `packages/runtime-host/src/protocol/index.ts`; choosing the branch side would incorrectly move the epoch backward and reuse an epoch with a different meaning. Please rebase onto current `main`, advance this protocol change to a fresh epoch (56 with the current ledger), preserve the intervening epoch entries, and rerun the exact-head checks. I will then re-review the new head.

---
anything here is wrong._

</details>

me2seeks:
Rebased onto latest main (`ff226aff2`) — resolved new conflicts in `protocol/index.ts` (epoch 56 → **57** keeping 51/ts) and `session-projector.test.ts` (kept both the context-budget test and the remainingMs test). Fixed leftover conflict markers that broke `tsc` on the rebased head. New head `74b4b5e`.

(Automated message — posted by Maka, the AI agent that authored the fix.)

jackwener (commented):
I re-reviewed the rebased exact head `74b4b5e1ab6651bb03e2425455e71d20db4281b0`. I found no remaining P0–P3 issue in the provider-retry countdown patch.

The rebase preserves the previously reviewed behavior: the Runtime Host recomputes a clock-domain-free `remainingMs`, both Desktop and TUI count down from their local receipt time, and both surfaces use the shared one-second display floor. The branch correctly advances the compatibility epoch to 57 while retaining the intervening epoch-56 contract.

I also checked the rebase-specific repair. The final commit removes the accidentally retained conflict-marker block from `session-projector.test.ts` while preserving both the upstream context-budget regression and this PR's scheduled-retry projection regression. The affected Core, Runtime Host, CLI, and UI builds passed, along with 257 focused tests. Exact-head `test`, `audit`, and `windows_recovery` checks succeeded. A synthetic merge with current `main` is clean.

This automated review does not constitute the independent committer approval required for merge.

---
anything here is wrong._

</details>

likun666661 (approved):

PR #3410, opened 2026-08-21
feat(runtime): add Markdown response guidance

### LINKED ISSUE #3392: feat(runtime): add shared Markdown guidance for main chat responses
## Problem

Main chat responses do not have a shared default format. Their structure can vary across models and providers: simple answers may be over-formatted, longer answers may be harder to scan, and code, literal values, or sources may not be presented consistently.

Maka supports GitHub-Flavored Markdown, but the shared main chat prompt does not give models neutral guidance for when and how to use it. Users can request a format themselves, but that does not provide a consistent product default.

## Desired outcome

Add a small set of shared default rules for main chat responses:

- Use GitHub-Flavored Markdown when it improves readability.
- Keep simple answers simple. Do not force headings or lists.
- Use short headings and flat lists when they make longer answers easier to scan.
- Use fenced code blocks for multiline code and backticks for inline commands, paths, identifiers, and literal values.
- Prefer descriptive link text for external sources when it is available.
- Follow a more specific format requested by the user or task.

These defaults should be neutral enough for a general-purpose agent. They should be the same across models and providers, remain stable across turns, and avoid changing response content or link safety behavior.

Focused coverage should verify that the guidance appears once in main chat prompts and is not added to child-agent prompts. Representative model checks should cover a simple answer, a longer structured answer, multiline code, and a response with a source.

Users can request a format in each conversation or add project instructions. This is repetitive and does not provide a consistent product default.

A single “use Markdown” instruction is shorter, but it does not discourage unnecessary structure or give models useful defaults for longer answers and code.

Rewriting responses after generation is not suitable. The display layer cannot reliably infer the intended structure and could change code, logs, URLs, or exact output.

## Non-goals

- Adding a full response style guide.
- Requiring a fixed syntax for links or citations.
- Adding different format rules for each provider.
- Rewriting model output after generation.
- Replacing free-form responses with JSON or another structured protocol.
- Changing the existing link safety policy.

## AI use

Codex drafted and revised this issue from a read-only review of Maka’s main chat prompt, Markdown response behavior, and reference agent prompts.

Colafornia:
## Summary

Main-chat responses do not have shared formatting guidance, so their presentation depends on each model's defaults. This is most visible with weaker models and source-heavy answers, which can return raw URLs instead of readable Markdown links.

This change adds one static response-format fragment to the shared main-session assembler. The guidance:

- uses GitHub-Flavored Markdown as the default response format;
- keeps simple answers free of unnecessary headings and lists;
- uses short headings and flat lists to organize longer answers;
- formats multiline code and inline commands, paths, identifiers, and literal values consistently;
- prefers descriptive text for external links; and
- allows more specific user or task formats to take priority.

The fragment applies only to regular main-session prompts. Child prompts and frozen execution profiles keep their existing prompts. Output rendering and link safety are unchanged.

The unused exported identity opt-out is removed because interactive main chat is the assembler's only remaining consumer.

Fixes #3392

## Verification

- `npm --workspace @maka/runtime test` — 3,003 passed, 11 skipped
- `node --test packages/runtime-host/dist/__tests__/execution-model-composition.test.js` — 24 passed
- scoped Biome check for the two changed files
- the focused Host test failed before the prompt update because the shared rules were absent, then passed after the update

### Model behavior

I ran one-time controlled checks with `gpt-5.4-mini`. Each comparison used the same input and runtime path. The response-format fragment was disabled for the Before run and enabled for the After run.

| Scenario | Before | After |
| --- | --- | --- |
| Simple answer | Returned `4` without extra structure | Returned `4` without extra structure |
| Longer explanation | Already used a flat numbered list | Used short paragraphs for the 180-word prompt; a longer variant used headings but still produced nested lists |
| Multiline code | Already used a fenced `javascript` block | Used a fenced `js` block and backticks for identifiers |
| External source | Returned the source as a raw URL | Returned a descriptive Markdown link |

In a longer API-comparison prompt, the baseline produced five Markdown headings. The guided run used short paragraphs and flat lists. Both runs already formatted code, commands, and paths correctly. Neither repeated the supplied documentation URLs because the prompt did not request a source list.

The comparison does not show improvement in every formatting dimension because the model already handled some Markdown cases well. It shows that the guidance preserves simple answers, can reduce unnecessary structure, and improves the source-link case from the issue.

These model calls are one-time PR evidence rather than automated tests because model output is nondeterministic and provider-dependent.

## AI use

Select exactly one:

Tool(s) and scope: Codex implemented the prompt guidance and tests, reviewed the change, ran the model comparisons, and prepared this PR description. The commit includes `Generated-by: Codex`.

## Checklist

- [ ] No

### DIFF
--- a/packages/runtime-host/src/__tests__/execution-model-composition.test.ts
+++ b/packages/runtime-host/src/__tests__/execution-model-composition.test.ts
@@ -1623,6 +1623,7 @@ test('production Host executes a durable runnable child with an exact tool ceili
     // agent-permission tools cannot be the thing that decides whether the child
     // can read back a result the runtime itself pruned.
     assert.deepEqual(toolNames(requests[2]?.body), ['ArchiveRead', 'Glob', 'Grep', 'Read']);
+    assert.doesNotMatch(JSON.stringify(requests[2]?.body), /## Response format/u);
     assert.ok(toolNames(requests[3]?.body).includes('agent_spawn'));
 
     const sessions = await execution.sessionStore.listForRecovery();
@@ -2573,6 +2574,18 @@ test('one turn shares one canonical Skill inventory across prompt and lazy tools
   assert.match(nextPrompt ?? '', /NEW_DESCRIPTION/);
   assert.doesNotMatch(nextPrompt ?? '', /OLD_DESCRIPTION/);
   assert.equal(inventoryReads, 2);
+
+  for (const prompt of [firstPrompt, nextPrompt]) {
+    assert.match(prompt ?? '', /^## Response format$/mu);
+    assert.equal(prompt?.match(/Use GitHub-Flavored Markdown for responses\./gmu)?.length, 1);
+    assert.match(
+      prompt ?? '',
+      /Keep simple answers simple; do not add headings or lists to simple answers\./u,
+    );
+    assert.match(prompt ?? '', /Use short headings and flat lists to organize longer answers\./u);
+    assert.match(prompt ?? '', /inline commands/u);
+    assert.match(prompt ?? '', /descriptive link text/u);
+  }
 });
 
 test('one composer freezes Runtime Policy while each Run freezes its remaining prompt sources', async () => {
--- a/packages/runtime/src/system-prompt/main-session-prompt.ts
+++ b/packages/runtime/src/system-prompt/main-session-prompt.ts
@@ -6,17 +6,18 @@
  * product identity. This closes that gap while leaving sub-agent identities to
  * the agent catalog.
  *
- * The assembler is deliberately order-agnostic: it only prepends the identity,
- * drops empty fragments, and joins. It does NOT impose a fixed fragment order,
- * because several fragments are position-semantic and must stay where the host
- * put them — e.g. the Side Chat isolation boundary and the Deep Research mode
- * contract are trailing assertions that constrain everything before them, so
- * moving them earlier weakens them. Each host keeps its existing fragment order.
+ * The assembler is deliberately order-agnostic: it only prepends the shared
+ * defaults, drops empty fragments, and joins. It does NOT impose a fixed
+ * fragment order, because several fragments are position-semantic and must stay
+ * where the host put them — e.g. the Side Chat isolation boundary and the Deep
+ * Research mode contract are trailing assertions that constrain everything
+ * before them, so moving them earlier weakens them. Each host keeps its existing
+ * fragment order.
  *
- * The identity line is pure static text: constant across turns, so it never
- * churns the systemPromptHash / prefix-cache (see request-shape.ts). It is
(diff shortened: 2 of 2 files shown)


Astro-Han (commented):
Thanks for keeping this guidance in the shared main-session assembler. I reviewed exact head `f20fd2813aa189620d7f0b43c4ea23e4c4fe5e6f`. This is the correct authority for provider-independent main-chat defaults; child prompts, frozen profiles, renderer link safety, and more specific task instructions remain outside or later than this fragment. I left one P2 behavior-validation gap below.

AI-assisted review disclosure: Codex performed the exact-head analysis, and an independent reviewer agent adversarially checked prompt precedence, provider boundaries, frozen profiles, rendering safety, and verification coverage. I verified the evidence and final severity.

packages/runtime-host/src/__tests__/execution-model-composition.test.ts
```diff
@@ -2573,6 +2574,18 @@ test('one turn shares one canonical Skill inventory across prompt and lazy tools
   assert.match(nextPrompt ?? '', /NEW_DESCRIPTION/);
   assert.doesNotMatch(nextPrompt ?? '', /OLD_DESCRIPTION/);
   assert.equal(inventoryReads, 2);
+
+  for (const prompt of [firstPrompt, nextPrompt]) {
```
Astro-Han, line 2578: Thanks for pinning the shared prompt and child-agent exclusion here. **[P2]** These assertions verify prompt placement, but the intended outcome is the model’s response behavior. Issue #3392 calls for representative simple, structured, code, and source checks, while the PR notes that the available credentials were rejected before inference, so none of those cases has run yet. Since this fragment reaches every normal main-session turn across providers, could we complete and record that small matrix before treating the behavior as verified? No code change is necessarily needed if those checks pass.

Colafornia: Thanks for pointing this out. I ran the simple, structured, code, source, and longer comparison cases with gpt-5.4-mini, using controlled before/after runs with the shared fragment disabled and enabled. I recorded the results and limitations in the PR description.

Astro-Han (approved):
Thanks for completing the behavior validation. The controlled `gpt-5.4-mini` before/after matrix now covers simple, structured, multiline-code, external-source, and longer comparison cases through the same runtime path. It shows that the shared fragment preserves simple answers, avoids unnecessary structure in the longer comparison, and improves the source case from a raw URL to a descriptive Markdown link, while honestly recording the dimensions the model already handled correctly.

The original validation gap is resolved, and I found no remaining P0–P3 issues on exact head `f20fd2813`.

AI-assisted review disclosure: OpenAI Codex reviewed the exact-head prompt boundary, recorded model evidence, test coverage, provider and child-session exclusions, and live GitHub state. I verified the evidence and made the final approval decision.

PR #3412, opened 2026-08-21
fix(runtime): recover hydrated image context overflow

### LINKED ISSUE #3372: fix(runtime): context budget measures stored events, so hydrated image payloads count as zero
## What happened

A Desktop session on `grok-4.6` (`xai-oauth`) entered a loop it cannot leave: every turn fails with `context_overflow` ("上下文窗口已超出限制") plus the recovery hint "没有执行工具，可直接重试", and automatic history compaction reports "Context summary failed" and keeps failing open. Retrying reproduces it exactly — the prompt is byte-identical on every attempt, so "retry directly" can never succeed.

The session had read three PNG screenshots with `Read`. From the captured provider request for the failing turn:

- prompt: 93 messages, **889,002 bytes** serialized; tool schemas 27,154 bytes
- one single tool result: **554,762 bytes** of base64 PNG; two more at 54,962 bytes each
- images are therefore ≈665 KB of an ≈889 KB prompt
- the same turn's `contextBudget` diagnostic: `estimatedTokensBefore: 67,193` against `maxHistoryEstimatedTokens: 500,000` — the budget sees no pressure at all
- provider outcome: `400 invalid-argument`, "Context window exceeded" → `errorClass: ContextLength`

The control that isolates the images: the `history_compact` call on the **same model and connection**, which does not carry the image parts, completed normally at `inputTokens: 71,082`. Only the image-bearing `main` call is rejected.

## Root cause

**The budget does not measure what is sent.**

Large payloads are stored by reference — the tool result persists as `{"kind":"image","mimeType":"image/png","ref":{"kind":"session_file",...}}`, about 360 bytes — and are hydrated into full base64 only when the provider request is built. That two-representation split is legitimate: `session_message_chunks` caps a chunk at 65,536 bytes, so a 554 KB payload has no business in the row.

The defect is which representation the budget authority reads:

- `estimateRuntimeEventsTokens` (`packages/runtime/src/context-budget-helpers.ts:45`) is the authority, and it runs on stored `RuntimeEvent`s.
- `estimateRuntimeEventChars` (`context-budget-helpers.ts:27`) is an if/else allowlist over `content.kind`: `text`, `thinking`, `function_call`, `function_response`, `error`. **`kind: "image"` matches no branch and returns 0** — not an underestimate, exactly zero.
- `estimateModelMessagesChars` (`context-budget.ts:344`) is the only estimator that runs on the *outgoing* `ModelMessage[]`, and it feeds `promptSegments` diagnostics only (`context-budget.ts:333`) — never a gate. In the failing turn `promptSegments` was `null`.

So no gate on this path has ever looked at the bytes that leave the process. Trimming never triggers, and compaction cannot help either: the images sit in the recent retained turns, and compaction folds historical text.

This stayed latent because for text the two representations are nearly the same size, so measuring the wrong one had no consequence. The budget carries an unwritten, never-verified assumption that stored ≈ sent. Images take the gap from ~0 to 554 KB.

## Proposed boundary

Not a new `else if` for `image`. Adding one concedes the allowlist and the next referenced payload kind — audio, native PDF inputs (#3164), any future blob — reproduces this incident unchanged.

1. **Measure the request, not the store.** The budget's input must be the hydrated payload, after hydration and before dispatch — the only point past which nothing changes the size. That seam already exists and is already captured per item: `PreparedRequestSegment.bytes` is `Buffer.byteLength(stableStringify(value))` at the provider seam (`packages/runtime/src/request-shape.ts:58`, `:461`). #2323 consumes those segments for the trace panel; the budget should consume the same numbers as a gate.
2. **Unknown parts fall back to bytes, not zero.** Whatever estimator survives must cost an unrecognized part by its serialized size, so an omission over-estimates instead of vanishing.

## How to reproduce

1. Open a Desktop session on a vision-capable model (observed on `grok-4.6` via `xai-oauth`).
2. `Read` a large PNG — roughly 400 KB on disk, ~550 KB once base64-encoded — so the tool result is stored as a `session_file` ref.
3. Continue the turn.
4. The provider rejects the request; the turn fails as `context_overflow`, while the `contextBudget` diagnostic reports a comfortable `estimatedTokensBefore` that excludes the image entirely.
5. Every retry and every automatic compaction attempt reproduces it — the session cannot be recovered from the UI.

## Environment

- Maka commit: `57e08d834`
- OS: macOS 26.5.2 (Darwin 25.5.0)
- Surface: Desktop + Runtime Host
- Node.js: v26.7.0
- Provider/model: `xai-oauth` / `grok-4.6`, `contextWindow: 500000`

## Out of scope (separate defects seen in the same session)

- **Compaction fail-open.** Three `history_compact` attempts timed out at 300 s; two more completed but produced unusable summaries — `grok-4.6` returned 57/62 output tokens with 922 reasoning tokens, failing the `orderedSectionsPresent` check from #3029 (`malformed_summary_missing_section`), and a relayed `deepseek-v4-flash` finished with `finishReason: tool_calls` and an empty body (`empty_summary`). Real defects, but not on this causal chain: correct budgeting would have handled the images before compaction was ever needed.
- **Misleading "retry directly" guidance** on a failure whose next request is byte-identical. Same shape as #3341, which already tracks the recovery-hint criterion.

## Related

- #2323 — instruments the same `PreparedRequestSegment` capture for the trace panel (observability; this issue asks for a gate on the same data)
- #2371 — an oversized prior turn defeating the budget, where the estimate itself was accurate (~244k estimated vs ~1 MB sent); here the estimate is zero
- #3164 — native PDF inputs, the next referenced payload kind that would hit this

Sun-GLiang:
## Summary

- omit hydrated historical image tool results only after a provider context overflow, then preserve the exact omission across later requests
- keep the recovery to one shared retry and prevent the generic retry path from looping context overflow
- show dedicated Desktop guidance when the shortened request still exceeds the context window

Fixes #3372

## Verification

- `npm test`
- `npm run lint`
- `npm run format:check`
- `npm run build`
- `npm run typecheck`
- `git diff --check`

### Visual evidence

Before — generic direct-retry guidance:

Context overflow before: generic direct-retry guidance

After — dedicated context-overflow guidance:

Context overflow after: reduce attachments or start a new task

## AI use

Tool(s) and scope: Codex implemented the Runtime recovery, Desktop presentation, tests, documentation, and review fixes.

## Checklist

- [ ] No


Astro-Han (approved):
Thanks for the careful provider-projection work. I reviewed the exact head and verified that the omission stays confined to historical tool-result images, remains durable across subsequent provider projections, and does not create a generic retry loop. The current test check is green, and I found no P0–P2 issues.\n\nAI-assisted review disclosure: Codex delegated two independent exact-head reviews; I verified their conclusions against the current PR state and approved this exact head.

PR #3419, opened 2026-08-21
feat(desktop): manage remote runtime host credentials

### LINKED ISSUE #3418: feat(desktop): manage access credentials for remote Runtime Hosts
<details open>
<summary>English</summary>

## Problem

Desktop can onboard and maintain an SSH-managed Runtime Host, but access credentials still require lower-level CLI operations. Users cannot inspect which Client identities retain access, rotate the credential used by the current Desktop Profile, or revoke an obsolete credential from the per-computer management surface.

Credential secrets must remain one-time values. Desktop should present authoritative metadata and safe maintenance actions without exposing stored secrets or turning an ordinary Runtime Host Client credential into operator authority.

## Scope

- Expose bounded, non-secret credential metadata through a Host-local operator command.
- Identify the credential used by the current Desktop Profile.
- Rotate that credential through replacement verification before retiring the old value.
- Revoke an obsolete credential through the SSH management plane with explicit confirmation.
- Keep newly delivered secrets inside the existing local credential and pairing-recovery path.
- Keep ordinary remote Client grants and remotely available operations unchanged; privileged access-management inputs remain available only to the Host-local operator.

## Implementation

- Draft pull request: #3419

## Non-goals

- Revealing or exporting credential secrets.
- A general RBAC or arbitrary operation-grant editor.
- Package updates, service replacement, or automatic update policy.

## Completion criteria

- Metadata contains no secret material and is bounded before crossing SSH.
- Rotation cannot retire the old credential until the replacement has been verified.
- The current Desktop credential cannot be revoked through the revoke action.
- Direct connection Profiles remain usable without claiming SSH credential management support.
- CLI, Desktop, and a managed Linux Runtime Host exercise the same contract.

</details>

<details>
<summary>简体中文</summary>

## 问题

Desktop 已能完成 SSH-managed Runtime Host 的 onboarding 与服务维护，但 access credential 仍需通过底层 CLI 操作。用户无法在 per-computer 管理界面中查看仍保留访问权的 Client identity、轮换当前 Desktop Profile 使用的 credential，或撤销废弃 credential。

Credential secret 必须继续保持一次性交付。Desktop 只应展示权威 metadata 和安全维护操作，不能暴露已保存的 secret，也不能把普通 Runtime Host Client credential 变成 operator authority。

## 范围

- 通过 Host-local operator command 提供有界且不含 secret 的 credential metadata。
- 识别当前 Desktop Profile 实际使用的 credential。
- 在 replacement 验证通过后再退役旧 credential。
- 通过 SSH management plane 明确确认并撤销废弃 credential。
- 新交付的 secret 只进入现有本地 credential 与 pairing recovery 路径。
- 普通 remote Client grants 与远程可用 operations 保持不变；扩展后的 privileged access-management inputs 仅供 Host-local operator 使用。

## 实现

- Draft pull request：#3419

## 非目标

- 展示或导出 credential secret。
- 通用 RBAC 或任意 operation-grant 编辑器。
- Package 更新、service replacement 或自动更新策略。

## 完成标准

- Metadata 不含 secret，并在跨越 SSH 前受到大小限制。
- Replacement 未验证前不能退役旧 credential。
- 当前 Desktop credential 不能通过 revoke action 撤销。
- Direct connection Profile 继续可用，但不声称支持 SSH credential 管理。
- CLI、Desktop 与 managed Linux Runtime Host 使用并验证同一套契约。

</details>

AI disclosure: OpenAI Codex assisted with implementation and drafting under maintainer direction. The maintainer decides whether to submit and accepts responsibility for the contribution.

M4n5ter:
<details open>
<summary>English</summary>

## Summary

Add credential maintenance to the per-computer management surface for SSH-managed Runtime Hosts. Desktop can inspect non-secret credential metadata, identify its own credential, rotate it through the existing verified pairing path, and explicitly revoke an obsolete credential.

Credential authority remains in Runtime Host. Desktop invokes Host-local operator commands over SSH, secrets never enter renderer state or terminal output, and privileged access-management mutations remain restricted to the Host-local operator; ordinary remote Client grants and remotely available operations are unchanged.

Fixes #3418

## Verification

- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- affected Desktop and CLI tests
- isolated CLI release package build
- Desktop end-to-end list, revoke, and rotate against a managed Linux Runtime Host

## AI use

Tool(s) and scope: Codex implemented and verified the change under maintainer direction and review

## Checklist

- [ ] No

</details>

<details>
<summary>简体中文</summary>

## 概要

为 SSH 托管的 Runtime Host 增加 per-computer credential 维护界面。Desktop 可以查看不含 secret 的 credential metadata、识别自身 credential、通过现有已验证 pairing 路径安全轮换，并明确撤销废弃 credential。

Credential authority 仍属于 Runtime Host。Desktop 只通过 SSH 调用 Host-local operator command；secret 不进入 renderer state 或 terminal output，privileged access-management mutation 仍仅限 Host-local operator；普通 remote Client grants 与远程可用 operations 保持不变。

Fixes #3418

## 验证

- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- 受影响的 Desktop 与 CLI tests
- 隔离环境中的 CLI release package 构建
- 在 managed Linux Runtime Host 上通过 Desktop 完成 list、revoke 与 rotate 端到端验证

## AI 使用

- [ ] 没有生成式工具作出实质贡献
- [x] 生成式工具作出了实质贡献

工具与范围：Codex 在维护者指导与审核下实现并验证了本次变更

## Checklist

- [x] 测试覆盖本次变更，且缺少变更时会失败
- [x] lint、format、typecheck 与受影响 suites 已在本地通过

本 PR 是否改变行为？

- [x] 是——已在概要中说明
- [ ] 否

</details>


Astro-Han (approved):
Approving `205f6572144729867a900a85ddbb86de406dcbf8`. Both required checks are terminal green and bound to that exact SHA (`package` run 32613643801, `test` run 32613643756 — `completed / success`, `head_sha` matches).

**Independent lines run against this head:** permissions & protocol, failure & lifecycle, persistence & protocol compatibility. **Zero P0–P2.**

- **Permissions:** the new rotation operations are not reachable from a remote Client — the operation is absent from the remote authorization list, `operationAllowsRemoteOwner` gates it a second time, and the shipped tests pin both the desktop and terminal shapes. Seven attack directions were enumerated and individually excluded.
- **Failure & lifecycle:** rotation splits into prepare → local rebind → final confirm, and each segment survives a crash because the old credential stays valid until the final step. There is a way back from every intermediate state.
- **Persistence & compatibility:** no format version bump and no migration rewrite; the new operations are local-owner-only, so an older remote Client can neither receive nor emit them. The compatibility matrix is closed.

**One [P3], spun out to an issue rather than held against this PR:** https://github.com/apache/maka/issues/3546 — in `runtime-host-ssh-terminal.ts`, a management frame that has already been fully received and decoded is discarded if the SSH process then fails to exit within `MANAGEMENT_TIMEOUT_MS`. The caller reports failure while the Host-side effect may have succeeded. **The shape pre-exists on main** (confirmed against the merge base) and is idempotent/retryable; this PR only makes it more visible by routing side-effecting credential operations through the same path. Assigned to @M4n5ter at his request.

**One thing I am flagging as a team question, not as a defect in this PR:** `protocol-compatible-changes/` is a human-written semantic assertion — a declaration file stating that a change is compatible, which then lets the epoch stay put. **This PR's assertion was checked by all three lines and it holds.** But the mechanism's correctness now depends on every future review actually reading that declaration. That is a governance decision for the project, and I am raising it separately rather than blocking here.

**Coverage disclosure, stated plainly.** Two further lines — the *functional contract surface* (preload bridge / CLI surface / renderer state machine) and the *untrusted-input & secret-leakage surface* — did not complete: one was voided because I leaked a severity into its dispatch brief and destroyed its independence, and one lost its executor to a runtime failure. **Both are being re-run right now by reviewers who have not seen any of the above.** The three lines that did complete each incidentally touched the credential-leakage question and found nothing, but incidental coverage is not a dedicated line and I am not going to present it as one. **If either re-run surfaces anything, I will post it here immediately** — including on an already-approved PR.

Approved on the explicit go-ahead of @M4n5ter (repo owner, author) and @AstroHan.

apps/desktop/src/renderer/settings/runtime-host-management-dialog.tsx
```diff
@@ -189,29 +263,157 @@ export function RuntimeHostManagementDialog(props: {
...
+                    variant="secondary"
+                    label={copy.cancel}
+                    isDisabled={loading}
+                    onClick={() => setConfirmation(undefined)}
+                  />
+                  <Button
+                    variant="primary"
+                    label={copy.rotateCredentialConfirm}
+                    isDisabled={loading}
+                    onClick={() => void rotateCredential().then(() => setConfirmation(undefined))}
```
Astro-Han, line 396: **[P3] Rotate clears its confirmation even when it failed; revoke does not.**

```tsx
onClick={() => void rotateCredential().then(() => setConfirmation(undefined))}
```

`rotateCredential` (`:134-149`) catches the IPC rejection, sets the error banner, and **does not rethrow** — so the promise resolves on failure too, and `.then()` clears the confirmation regardless of outcome. `revokeCredential` (`:150-172`) puts `setConfirmation(undefined)` **inside the `try`, after** the snapshot comes back (`:163`), so a failed revoke keeps its confirmation.

**Post-approval finding.** This PR is already approved; I am posting it because I said I would post whatever the re-run lanes surfaced, including afterwards. It does not change the approval.

Repro: Manage access → pick the current Desktop credential → confirm Rotate → have the bridge reject with `new Error('<failure>')`. The old access snapshot and the error banner both survive, but the rotation confirmation is gone, so the user has to re-open and re-confirm to retry the thing that just failed.

**Why it is P3 and not higher:** nothing is lost or half-applied — the rotation genuinely did not happen, the snapshot is untouched, and the error is visible. It costs the user one extra confirmation on a path they were already retrying.

**Why it is still worth fixing:** the two mutations sit next to each other and disagree about what a failure means for the confirmation state, and the *safer* of the two shapes is the one revoke uses. Making rotate match revoke — clear only after a successful snapshot — is a one-line move and removes the divergence rather than documenting it.

`:381` (`run('uninstall').then(...)`) has the same `.then()`-after-swallowed-rejection shape, so this is a small pattern rather than a one-off. Worth a look while you are in here, though uninstall's failure mode is less annoying than rotate's.

*Independent contract-surface line, blind to all other review on this PR, bound to `205f6572`. Verified at the gate: 2026-08-23 12:35 UTC.*

packages/cli/src/cli-core.ts
```diff
@@ -123,6 +123,7 @@ function helpText(cliCommand: string): string {
     `  ${cliCommand} runtime-host service status|start|stop|restart|logs|uninstall [--json]`,
     `  ${cliCommand} runtime-host access issue --principal <id> --grant <operation>`,
     `  ${cliCommand} runtime-host access issue --principal <id> --preset <desktop-client|terminal-client>`,
+    `  ${cliCommand} runtime-host access list`,
```
Astro-Han, line 126: **[P3] `access list` / `access revoke` accept `--root` and `--expected-root`, but `--help` never says so — and `--expected-root` is absent from the access help entirely.**

The parser accepts both for every access action (`runtime-host-cli.ts:805-818`), and both are threaded into the results:

```ts
return { kind: 'runtime-host-access-list',  ...(rootPath ? { rootPath } : {}),
         ...(expectedRootId ? { expectedRootId } : {}), framed };
return { kind: 'runtime-host-access-revoke', ...(rootPath ? { rootPath } : {}),
         ...(expectedRootId ? { expectedRootId } : {}), credentialId, ... };
```

But the help shows `access list` and `access revoke --credential <id>` bare, puts `--root` under **"Runtime Host access issue options"**, and lists `--expected-root` only under `profile set` and the capability provider — never under access at all.

**Post-approval finding**, same as the sibling comment: already-approved PR, posting because I committed to reporting what the re-run lanes found.

Repro: `maka runtime-host access list --root <root-path> --expected-root <root-id>` parses and runs. Nothing in `maka --help` lets a user discover that root pinning is available on `list` / `revoke`.

**Why it matters more than a typical help nit:** `--expected-root` is the flag that pins which Host root you are talking to. On a machine with more than one root, a user reading only the help would reasonably conclude that `access list` and `access revoke` cannot be pinned — and would then run them unpinned. The flag exists precisely so that operations against the wrong root fail loudly.

Suggested shape: promote `--root` / `--expected-root` to a shared **"Runtime Host access options"** block and keep the per-action flags (`--principal` / `--grant` / `--preset` under issue, `--credential` under revoke) where they are. `prepare`, `--framed` and `--current-fingerprint` can stay out of public help as the operator-only surface they are.

*Independent contract-surface line, blind to all other review on this PR, bound to `205f6572`. Verified at the gate: 2026-08-23 12:35 UTC.*

Astro-Han:
**Closing the coverage gap I disclosed in the approval.** Both missing lines have now run, on `205f6572144729867a900a85ddbb86de406dcbf8`, by reviewers who had seen none of the earlier review on this PR. **The approval stands; nothing found rises above P3.**

### Untrusted-input & secret-leakage surface — zero findings

- `accessSnapshot` is a **field-by-field allowlist**, not a spread: only `credentialId / principalKind / principalId / status / createdAt / expiresAt / isCurrentDesktop` leave the main process. Even `credentialFingerprint` does not — it is consumed to compute the `isCurrentDesktop` boolean and dropped. The new secret never crosses IPC at all; it goes `response.credential → profiles.rotateManagedCredential(...)` inside main.
- The framed-output filter was checked on its **failure** paths, not just the happy one: an over-long frame sets `pending=''` **and** `discardReservedLine=true` so the rest of the line is also dropped; a stream ending mid-frame discards rather than flushing to visible; `markerSuffixLength` retains a partial prefix across chunk boundaries. Command lines carry only `--credential <id>` and `--current-fingerprint <hash>` — neither is a secret, which matters because argv is readable by other processes on the same machine.
- Decode is fail-closed and **still distrusts what it decoded**: length cap → base64url → `JSON.parse` → `safeParse` against a `.strict()` schema at every level, all inside try/catch returning `undefined`. Then `rotateCredential` re-checks that current is active / `remote_owner` / may publish / **may not** use host paths, replacement is pending, and `sameCredentialAuthority(current, replacement)` requires identical grant sets — **a remote owner cannot quietly widen its own authority through a rotation.**

That last check is the one I would call the load-bearing defense on this surface, and it is correct.

**One stated boundary, resolved at the gate rather than left open.** That line flagged that it could not see the Host-side CLI's own stdout. The file is in fact in this diff (`packages/cli/src/runtime-host-access-command.ts`), so I checked it: `runRuntimeHostAccessPrepareCli` writes **only** the encoded frame; the sole plaintext-credential stdout path is `runRuntimeHostAccessIssueCli`, where plaintext **is** the delivery mechanism for a human operator; and `runtimeHostAccessManagementRemoteCommand` puts `--framed` in the argv **unconditionally** while its action type admits only `prepare | revoke | list`. **Desktop cannot construct the plaintext command.** Closed by type and by unconditional flag, not by convention.

### Functional contract surface — two P3s, both inline

- https://github.com/apache/maka/pull/3419#discussion_r3837684876 — rotate clears its confirmation on failure; revoke does not.
- https://github.com/apache/maka/pull/3419#discussion_r3837684902 — `access list` / `access revoke` accept `--root` / `--expected-root`, but the help never mentions them.

Also checked and clean: preload argument/return/rejection shapes match renderer consumption; `list` / `rotate` / `revoke` all replace the snapshot **after** the backend succeeds, so there is no optimistic commit; the `loading` flag and confirmation gate stop an ordinary double-click from firing a second mutation; the current-credential fingerprint mapping, the synthesized post-pairing snapshot, and the guarded current-credential revoke contract all agree with what the UI displays.

**Why I am posting this on an approved PR rather than quietly closing the loop:** the approval said two surfaces had no dedicated coverage and promised the results either way. A disclosure that only gets updated when it turns out well is not a disclosure.

Astro-Han (commented):
Follow-up on `205f6572144729867a900a85ddbb86de406dcbf8` — the same head this PR was already approved on. **No new findings; this adds a security line that finished after that approval, so the record shows what it actually covers.**

This line examined two claims the change depends on. Both hold, and not in a "found no counter-example" way — the surfaces were walked layer by layer.

**Secrets do not reach renderer state or terminal output.** The credential's whole life cycle has six layers and each one closes: 32 random bytes held only in host memory; the store persists a sha256 hex, never the secret; hand-off to the CLI goes through a one-shot delivery file (0600, UUID name, 60s TTL, deleted on read, residue purged at authority startup, credentialId-checked); the only CLI→Desktop exit is a framed stdout line under a zod schema with size caps, and all three Desktop actions hard-code `--framed` while the remote command itself carries only paths, rootId, fingerprint and credentialId; the PTY filter matches incrementally across chunks and swallows whole lines carrying the prefix — including invalid, oversized and incomplete frames — keeping the marker suffix so a half-prefix cannot leak.

**Verified independently at the gate:** `accessSnapshot` (`runtime-host-management.ts:184`) is an explicit whitelist — `credentialId`, `principalKind`, `principalId`, `status`, `createdAt`, optional `expiresAt`, `isCurrentDesktop`. No credential, no grants. The `credentialFingerprint` is *read* but never emitted: it is compared against the current fingerprint and only the resulting boolean crosses the boundary. That is stronger than omission, because it cannot be reintroduced by a careless spread.

**Privileged mutation stays with the Host-local operator.** Also verified at the gate: this PR's diff against `protocol/operations.ts` is **empty**, and neither `access.credential.rotation.prepare` nor `.revoke` appears in `REMOTE_OWNER_OPERATION_GRANTS`. `authorizeRuntimeHostOperation` gates every request, and `local_owner` cannot be forged over a websocket — authentication compares hashes with `timingSafeEqual`, and `principalKind` comes from the stored record rather than from anything the client says. Transport-wise the host CLI reaches a 0600 local IPC socket. The TOCTOU window is closed server-side: `prepareRotation` requires the credential being replaced to still be active, and `revokeRotation` requires `requiredActive` to still be active and refuses self-revocation.

**Privilege cannot move in either direction:** the replacement's grants are copied by the server from the current credential — the client cannot specify them — and the Desktop side checks exact set equality via `sameCredentialAuthority`. A malicious host cannot escalate, and a malicious desktop cannot request more.

**Excluded rather than reported, with reasons, since a bare "looks fine" is not useful:**

- `access revoke` with an unknown credentialId and no fingerprint hits `undefined === undefined` and shows a misleading *"Rotate this Desktop credential"* (`runtime-host-access-command.ts` ~276). Desktop always sends a fingerprint and the path fails closed — cosmetic, below the bar.
- Non-framed `access list` prints metadata and fingerprint to the host terminal: host-local, no secret. By design.
- The delivery file is readable for 60s by the same user — and the same user *is* the operator. Inside the trust model.
- Two credentials are live during the rotation window (old active, new pending, ≤15min): inherent to a two-phase rotation, grants are identical and the pending one expires on its own. No privilege expansion.

**One retraction, recorded because it matters more than the finding would have.** This line initially reported that the framed filter lacked negative-path tests. That was wrong — wrong-action, invalid, oversized, incomplete and multiple-frame cases are all covered. The original claim came from an incomplete search, and it is withdrawn rather than quietly dropped.

**Observation, not a finding:** the three frame files carry ~30 lines of near-identical envelope encode/decode. A generic envelope would converge them. Below the threshold for a finding, and mentioned only because it is the kind of triplication that grows a fourth member.

**Not covered by this line:** the two `[P3]`s already filed above are unaffected, and UI/UX is out of scope here except for one observation — the snapshot deliberately omits `operationGrants`, so the renderer cannot show per-credential authority scope. That is a product choice rather than a defect.

*This review is AI-assisted. Under `CONTRIBUTING.md` §Review it does not count as the required independent human review.*

Issue #3354, opened 2026-08-20
Persisted connection catalogs keep model ids that the snapshot has retired


Astro-Han:
When a models.dev refresh removes a model id from a provider's fallback list, existing users keep it. `ensureConnection` (`packages/runtime-host/src/server/bootstrap-runtime-policy.ts`) returns an existing connection unchanged, so `enabledModelIds` is never reconciled against the current registry. A persisted default target stays valid while the old id is still enabled (`packages/storage/src/runtime-policy/connection-catalog-document.ts`), and Host admission trusts the persisted `models[]` row (`packages/runtime-host/src/server/connection-model-admission.ts`). The id is still shown as available and default-capable, and still admitted for execution, after the provider has retired it.

Providers with `modelDiscovery: { kind: 'protocol' }` self-heal on the next successful discovery. Two cases do not:

- `opencode-free` is `{ kind: 'fallback' }` — its own reason string states the anonymous endpoint has no model-list contract, so no discovery pass ever repairs the state.
- `google` preserves the old catalog when discovery fails.

## Why this is not tied to one refresh

This is standing behaviour, not a regression. #2050 removed `hy3-free` from the same OpenCode Free list for the same reason — an upstream deprecation — and touched no persistence: `git show 41a983804 --name-only` matches zero files under `packages/storage/` or `packages/runtime-host/`. There is no model-catalog migration or tombstone mechanism; the tombstones in `sqlite-session-metadata-store.ts` serve session metadata only.

Reverting a refresh does not fix it either: the ids return to the fallback lists but remain retired upstream, so a persisted default still fails at call time.

## Scope

A targeted upgrade migration that, for providers without a self-healing discovery contract, removes retired ids from persisted inventories and enabled sets and clears or repairs a matching default target, with base-state upgrade tests.

Found during review of #3324 by @hqhq1025.

mikemikimike:
[...]

Issue #2165, opened 2026-08-04
fix(opencode-free): replace unhealthy big-pickle bootstrap default


liugddx:
## Problem

A fresh Maka install seeds the anonymous `opencode-free` connection with `big-pickle` as its default model. That model currently does not complete within the 15-second connection-test timeout, so the built-in zero-credential provider appears offline even though the OpenCode Zen endpoint and other free models are healthy.

Existing installations retain the same stale default in `llm-connections.json`, so changing only the provider fallback order would not repair them.

## Reproduction

1. Start Maka with a fresh workspace and no provider API keys.
2. Open Settings → Connections.
3. Test the preconfigured OpenCode Free connection.

Expected: the zero-config connection test succeeds and a first chat can use the selected default.

Actual: the request using `big-pickle` times out and the connection is reported as failed.

## Live endpoint check (2026-08-04)

- `GET https://opencode.ai/zen/v1/models`: 200
- anonymous `mimo-v2.5-free`: 200
- anonymous `nemotron-3-ultra-free`: 200
- anonymous `laguna-s-2.1-free`: 200
- anonymous `big-pickle`: timed out
- anonymous `north-mini-code-free`: 500 `Router.Unavailable`

## Proposed fix

- Move the bootstrap default to a currently responsive free model and keep the provider fallback order consistent.
- Migrate only the untouched historical auto-seed (`opencode-free` + old default/selection), without changing customized connections.
- Add regression coverage for fresh bootstrap and migration behavior.
- Verify the selected default against the live anonymous endpoint.

PR #3429, opened 2026-08-21
fix(core): derive the OpenCode Free model set from the models.dev snapshot

### LINKED ISSUE #3409: bug(core): OpenCode Free pinned model list has rotted — 3 of 4 pinned models fail (429/400) and new free models never appear
### What happened

OpenCode Free (the zero-credential default) is largely broken because its free-model set is a hardcoded pin that has rotted, and new free models never reach users.

The pin lives in `packages/core/src/provider-registry.ts:604-609` (`opencodeFreeModelIds`) and the default enabled set at `packages/core/src/provider-registry.ts:8-13`:

- pinned: `nemotron-3-ultra-free`, `mimo-v2.5-free`, `big-pickle`, `deepseek-v4-flash-free`
- default enabled: `nemotron-3-ultra-free`, `mimo-v2.5-free`, `deepseek-v4-flash-free`

Live probes against `https://opencode.ai/zen/v1/chat/completions` (anonymous, 2026-08-21):

| model | result |
| --- | --- |
| `nemotron-3-ultra-free` | 200 OK |
| `mimo-v2.5-free` | **429** `FreeUsageLimitError` — "Rate limit exceeded. Please try again later." |
| `big-pickle` | **429** `FreeUsageLimitError` — same |
| `deepseek-v4-flash-free` | **400** `server_error` — "Upstream request failed: Model is unavailable." |

So 3 of the 4 pinned models are dead: users get stuck in `Error: Rate limit exceeded` loops, and the TUI even logs confusing notices like `Model changed: mimo-v2.5-free → mimo-v2.5-free` when re-electing within the stale set.

At the same time, the free catalog upstream has moved on. models.dev currently lists ~29 free `opencode` models; several work anonymously today but are absent from the pin, e.g.:

- `x-preview-f-free` — displayed as **"Ox Alpha Free (Unlimited)"**, tool-capable, answers 200 anonymously
- `laguna-s-2.1-free` — 200
- `nemotron-3.5-lightning-free` — 200

Because the pin is `as const`, none of these ever appear in the model picker until someone edits the list by hand. The repo's own generated snapshot lags too: as of today's refresh (#3324) `model-metadata.generated.ts` still has no `x-preview-f-free`.

One more wrinkle for whoever fixes this: the anonymous-served set is strictly smaller than the models.dev free set. Many snapshot-listed free ids (`glm-5-free`, `kimi-k2.5-free`, `grok-code`, `minimax-m3-free`, …) answer `401 ModelError: Model … is not supported` anonymously, so deriving the free list from `cost.input === 0` alone would over-offer. Validation needs to hit the live endpoint (or at least treat 401 ModelError as "not anonymously served").

Suggested directions (maintainer's call):

- stop hardcoding the free set; derive candidates from the models.dev snapshot and health-check them (the `testConnection` fallback-probe machinery already does most of this)
- or keep a pin but add a scheduled check that fails CI/lint when a pinned id stops answering 200 anonymously
- treat 429 `FreeUsageLimitError` / 400 "Model is unavailable" as rotation signals so a dead enabled model doesn't trap the session

### How to reproduce

1. Start Maka fresh with no credentials — the bootstrap seeds the `opencode-free` connection with enabled models `[nemotron-3-ultra-free, mimo-v2.5-free, deepseek-v4-flash-free]` (`packages/runtime-host/src/server/bootstrap-runtime-policy.ts:76-83`).
2. Switch the model to `mimo-v2.5-free` (or exhaust the per-IP quota on the default), then send any message.
3. Turn fails with `Error: Rate limit exceeded`; picking `deepseek-v4-flash-free` fails because the model is unavailable upstream; `big-pickle` is likewise 429.
4. Open the model picker on the OpenCode Free connection — only the stale four are offered; none of the newly added free models (e.g. Ox Alpha Free) exist.

Direct curl equivalent:

```sh
curl -s https://opencode.ai/zen/v1/chat/completions \
  -H 'content-type: application/json' \
  -d '{"model":"mimo-v2.5-free","messages":[{"role":"user","content":"hi"}],"max_tokens":1}'
# {"type":"error","error":{"type":"FreeUsageLimitError","message":"… Rate limit exceeded …"}}
```

### Environment

- Maka version or commit: `57e08d834` (main, 2026-08-21)
- OS and version: Linux
- Surface: all surfaces that use the `opencode-free` connection (Desktop / TUI / CLI)
- Node.js version, if running from source: 22.x

- This is the same failure class as #2165, which swapped one unhealthy pinned default (`big-pickle`) for another; the pin has rotted again since. A one-off swap will keep recurring.
- Related: #3355 (fallback lists offering retired models), #3354 (persisted catalogs keeping retired ids).
- Probe summary (anonymous, 2026-08-21): working — `nemotron-3-ultra-free`, `x-preview-f-free`, `laguna-s-2.1-free`, `nemotron-3.5-lightning-free`; rate-limited — `mimo-v2.5-free`, `big-pickle`; unavailable — `deepseek-v4-flash-free`; not anonymously served (401 ModelError) — most other `-free` ids in the snapshot.

<sub>Diagnosis and endpoint probes above were prepared with assistance from an AI coding agent (opencode); the reporter verified the results.</sub>

cunninghamcard-bit:
## Summary

The `opencode-free` provider pinned four free model ids as a compile-time constant; the free tier rotates fast, three of the four now fail live probes, and new free models never surfaced. #2165 already swapped this pin once by hand, so this fixes the mechanism:

- **Derived, not pinned** — `scripts/sync-model-metadata.mjs` marks models.dev zero-input-cost models `isFree`; the opencode-free set is derived as `isFree && functionCalling && !deprecated && !broken`, default model first, with a module-load guard that fails loudly if the snapshot drops the default. Rotation now flows through routine snapshot refreshes. Discovery stays `kind:'fallback'`, aligned with #3330: the anonymous `/models` listing carries no cost facts, so which models are *free* is a models.dev fact — the derived candidates are the inventory, the user's selection is the authorization, and the send settles availability.
- **One evidence-backed quarantine** — `muse-spark-1.2-contributor-free` persistently answers HTTP 200 with an *empty* completion while billing the full token budget (4 consecutive anonymous probes, 2026-08-21, max_tokens 8–200): a failure the send cannot surface, so "the send settles it" is unavailable as the arbiter. Such ids carry dated evidence in a deny-list, are vetoed in `authorizeConnectionModel`, and omitted from catalog offers. Deny-only: a stale entry hides at most one healthy model, the opposite failure mode of the old pin.
- **Existing installs heal on upgrade** — bootstrap migrates an `opencode-free` row only when its `enabledModelIds` still exactly match a historical system seed (a reordering counts as a user selection and is never touched); idempotent, revision-CAS, and it re-seeds the default target only when the migration itself nulled it. The exact-match enumeration is deliberately lossy; the versioned seed policy from #3354 is the durable replacement.

Out of scope, tracked in #3449: transient 429/5xx per-request failures (per-IP quota is a point-in-time signal — probes showed the same id failing and succeeding within hours) and automatic failover, which belong at the Runtime Host health seam.

Fixes #3409

## Verification

- Rebased onto current `main` (post-#3330); full workspace typecheck and test suite pass (10 workspaces, 0 failures); biome clean on changed files. New tests: seed migration (migrated / user-modified incl. reordered / repeat-bootstrap / default repair) and the `authorizeConnectionModel` quarantine veto.
- Live anonymous probes on 2026-08-21 across the derived candidates; side-by-side fresh desktop builds: pre-fix seeds the stale 3-model set, post-fix seeds the derived set and the default model answers normally.

## AI use

Select exactly one:

Tool(s) and scope: Claude Code (Claude Fable 5) authored the code and tests in both commits, ran the snapshot refresh and live endpoint probes, and drafted this description; adversarial verification used Claude subagent workflows and an OpenAI Codex CLI review, whose confirmed findings (stored-selection quarantine gap, default-target repair) are incorporated. The human contributor of record reviewed the changes and verified both desktop builds. Both commits carry the `Generated-by: Claude Code` trailer.

## Checklist

- [ ] No


packages/core/src/provider-registry.ts
```diff
@@ -605,25 +600,25 @@ const opencodeGoModelIds = toolCallingModelIds(
...
-  }
+// OpenCode Zen endpoint and model ids, exposing the active tool-capable
+// models the models.dev snapshot marks `isFree` (zero input cost). Deriving
+// the set from the snapshot lets routine metadata refreshes rotate free
+// models in and out instead of letting a hardcoded pin rot (#3409).
+const opencodeFreeModelIds = toolCallingModelIds(
+  'OpenCode Free',
+  Object.fromEntries(
+    Object.entries(GENERATED_MODELS_DEV_METADATA.opencode).filter(
+      ([, model]) => model.isFree === true && model.lifecycle !== 'deprecated',
```
Astro-Han: **[P2]** Thanks for replacing the manually maintained pin with structured metadata. The remaining issue is that this filter answers “free, tool-capable, and not deprecated,” but the result is also used as the set that is currently runnable and enabled by default. Those are different contracts: on the current exact head, anonymous probes still return 429 for `big-pickle` and `mimo-v2.5-free`, while `muse-spark-1.2-contributor-free` does not produce a usable completion. The fallback probing in `testConnection` only runs for an explicit connection test; it neither updates `enabledModelIds` nor participates in normal turns.

Could we keep this derived set as the candidate inventory, but separate it from the default-enabled set? The smallest safe version would seed only the known-working default model. For the clean final state, models.dev should own price/capability/lifecycle eligibility, anonymous `/models` discovery should intersect the served catalog with those candidates, and Runtime Host should own per-connection health and reversible fallback when inference returns availability or rate-limit failures. That keeps the dynamic facts at the existing Host seam instead of encoding them into a global snapshot.

packages/runtime-host/src/__tests__/bootstrap-runtime-policy.test.ts
```diff
@@ -19,14 +23,14 @@ test('a fresh Host starts with one anonymous runnable target', async () => {
...
     assert.equal(free?.slug, 'opencode-free');
     assert.equal(free?.enabled, true);
-    assert.deepEqual(free?.enabledModelIds, [
-      'nemotron-3-ultra-free',
-      'mimo-v2.5-free',
-      'deepseek-v4-flash-free',
-    ]);
+    // The free set is derived from the models.dev snapshot and rotates with
+    // refreshes; assert the structural contract, not today's ids.
+    assert.deepEqual(free?.enabledModelIds, [...OPENCODE_FREE_DEFAULT_ENABLED_MODELS]);
```
Astro-Han, line 28: **[P2]** Thanks for covering the fresh-Host wiring here. This does not cover the normal upgrade path, though: `ensureBootstrapRuntimePolicy()` returns as soon as any catalog already exists, and `ensureConnection()` preserves an existing `opencode-free` row unchanged. Users upgrading from the old seed therefore retain the retired/dead IDs and do not receive the newly derived candidates.

The clean fix should be a versioned built-in seed policy in the Runtime Host catalog authority: migrate only a connection that can still be proven to match the previous system-owned seed, make the migration idempotent and revision-safe, and always preserve a user-modified `enabledModelIds` selection. If that belongs in #3354 instead, could this PR explicitly limit its claim to fresh/new connections and avoid closing the existing-user part of #3409?

Astro-Han (commented):
Thanks for removing the hand-maintained OpenCode Free pin and reusing the generated metadata/provider-registry seam. That is the right direction and avoids another one-off model-list patch.

I left two P2 comments because the current implementation still conflates the snapshot candidate catalog with live anonymous availability, and it only updates fresh or newly created connections. Neither issue requires another static allowlist or provider-specific renderer path.

The clean final state would keep the responsibilities separate:

1. models.dev owns free/capability/lifecycle candidate facts;
2. the existing discovery seam intersects those candidates with the anonymously served catalog;
3. Runtime Host owns connection-local health, fallback, and persisted selection;
4. a versioned built-in seed migration updates untouched legacy defaults without overwriting user choices.

I am leaving this as a non-blocking COMMENT rather than requesting changes, but I do not think the current head fully resolves #3409 yet.

AI-assisted review disclosure: OpenAI Codex coordinated two independent reviewer agents to inspect the exact-head diff, related Runtime Host and execution paths, live anonymous endpoint behavior, tests, and severity calibration. I reviewed and approved these comments before posting.

packages/core/src/provider-registry.ts
cunninghamcard-bit: Thanks — agreed the two contracts were conflated. Addressed in 2f2e980: `OPENCODE_FREE_DEFAULT_ENABLED_MODELS` now seeds only the probe-verified default model, while the derived set stays as the opt-in candidate inventory in `fallbackModels`. The fuller separation you describe (models.dev eligibility ∩ anonymously served catalog at the discovery seam, with Runtime Host owning connection-local health and reversible fallback) matches where we'd like this to go too; leaving that as a follow-up since it touches the discovery and Host seams beyond this fix.

packages/runtime-host/src/__tests__/bootstrap-runtime-policy.test.ts
cunninghamcard-bit: Good catch — fresh installs only was a real gap. Addressed in 2f2e980 with a seed migration that avoids a version field: bootstrap now migrates an existing `opencode-free` row only when its `enabledModelIds` still exactly match one of the three historical system seeds (`['big-pickle']` from #1720, `['nemotron-3-ultra-free']` from #2166, or the three-model list from #2431) — exact match is the proof it is still system-owned. Any other inventory is treated as a user selection and never touched. The migration is idempotent (no-op once the row equals the current seed) and commits through the catalog's revision CAS. Tests cover the migrated, user-modified, and repeat-bootstrap cases. If maintainers prefer the versioned seed-policy design from #3354 instead, happy to align — this version keeps the provable-match property without new persisted state.

packages/core/src/provider-registry.ts
cunninghamcard-bit: Correction and a consolidated plan — apologies for the churn on this thread (2f2e980 narrowed the seed to one model, 62de05b reverted it; both were wrong ways to answer your point, and I should have settled the design here first).

Fresh anonymous probes just now: 5 of 7 candidates complete successfully — including both `big-pickle` and `mimo-v2.5-free` that probed 429 earlier (`FreeUsageLimitError` is per-IP and transient), while `nemotron-3-ultra-free` itself was failing with a transient upstream 502 at probe time. So availability is a point-in-time, per-IP signal: encoding any probe snapshot into the compile-time constant is wrong in both directions.

Which is, I now understand, your actual point: eligibility and availability are different contracts, and availability must be resolved at runtime. Final shape for this PR:

- models.dev snapshot keeps owning eligibility (free, tool-capable, not deprecated) — unchanged;
- the connection's model list is resolved at the existing discovery seam as **candidates ∩ anonymously served `/zen/v1/models`**, falling back to the static candidates when the endpoint is unreachable — this removes the "declared free but not anonymously served" ids;
- transient 429/5xx per-request failures stay a Runtime Host health concern (your item 3), which I'd leave to the #3354-adjacent follow-up rather than bolt onto this PR.

Implementing the intersection now; will push it as one change.

cunninghamcard-bit: Landed in 2ecd384 (with 62de05b), completing the plan above: eligibility from the snapshot; the connection's list resolved at the existing discovery seam as candidates ∩ anonymously served `/zen/v1/models` (anonymous, `filter: 'fallback-models'`), falling back to the static candidates when unreachable.

One disclosure beyond that plan: your muse-spark observation was confirmed and is worse than an availability blip — it persistently answers HTTP 200 with an empty completion while billing the full token budget (4 consecutive anonymous probes, max_tokens 8–200). That failure shape is invisible to both the /models intersection and any future 429/5xx health seam, so it is quarantined via an evidence-dated deny-list enforced at catalog AND execution admission (including inventories stored before the entry existed) — deny-only, the opposite failure mode of the old pin. Transient per-request failures (429 quota, upstream 5xx) remain with the Runtime Host health/failover follow-up you sketched as item 3.

packages/runtime-host/src/__tests__/bootstrap-runtime-policy.test.ts
cunninghamcard-bit: Completed in 2ecd384. On top of the exact-match seed migration described above (unchanged: three verbatim historical seeds, revision-CAS, idempotent, user-modified rows — including reordered lists — never touched), the migration now also repairs the default target when the migration itself nulled it (a previously cleared default stays cleared), and the update write refreshes the stored non-fetched `models` inventory to the current build's candidates, so a migrated row's picker and execution admission agree. The exact-match enumeration is documented as deliberately lossy — every seed change must append the prior value — with the versioned seed policy from #3354 as the durable replacement; happy to align this to that design when it lands.

Astro-Han:
Everything looks good now, only a rebase is needed before we moving forward~

cunninghamcard-bit:
Rebased onto current main (`3fdc825`, now two commits). The rebase absorbed #3330's authorization model, which simplified this PR: the protocol-discovery intersection I had added is gone (its `fallback-models` filter no longer exists, and for an anonymous free tier the send settles availability exactly as #3330 argues — the candidates stay the advisory inventory). What remains: the snapshot derivation with the guard, the legacy-seed migration with default-target repair, and one deliberate exception to "the selection is the authorization": the evidence-dated quarantine for `muse-spark-1.2-contributor-free`, whose billed-200-empty-completion failure the send cannot surface — vetoed in `authorizeConnectionModel` and omitted from catalog offers. Full suite green on the rebased head; PR description updated to match.

packages/runtime-host/src/server/bootstrap-runtime-policy.ts
```diff
@@ -146,6 +170,50 @@ async function ensureConnection(
...
+      ({ slug, providerType }) => slug === 'opencode-free' && providerType === 'opencode-free',
+    );
+    if (
+      !free ||
+      sameIds(free.enabledModelIds, OPENCODE_FREE_DEFAULT_ENABLED_MODELS) ||
+      !LEGACY_OPENCODE_FREE_SEEDS.some((seed) => sameIds(free.enabledModelIds, seed))
+    ) {
+      return false;
+    }
+    const committed = await stores.connectionCatalog.update({
```
Astro-Han: [P2] Make built-in seed evolution one atomic catalog mutation

Reachability: ① normal upgrade path; the interruption case is ② a reasonable failure/recovery path.

This generic update changes `enabledModelIds`, but a real historical `opencode-free` row already persists its fallback `models` inventory. Because the catalog store preserves `previous.models` when the endpoint is unchanged, the migrated row keeps offering the old retired/dead models even though its enabled selection has moved to the new derived seed.

The same transaction boundary creates a restart gap: this update first lets retained-target validation persist the removed legacy default as `null`, and bootstrap sets the replacement in a second catalog write. If the process exits between those writes, the next startup sees the current seed, so migration returns `false`; it also sees an already-null default, so the replacement is never restored.

The current regression fixture cannot observe the inventory half because it creates the “legacy” row through the current-head store, which seeds `models` from the new fallback before migration begins.

Could we make built-in seed evolution one connection-catalog-authority CAS mutation that:

1. recognizes the historical system seed;
2. replaces its enabled IDs and fallback `models`/`modelSource`/`modelsFetchedAt`;
3. retargets a removed system default in the same document write;
4. preserves a default that was already `null`; and
5. leaves user-modified selections untouched?

A focused test should start from an actual historical persisted document, then verify both normal completion and restart state on either side of the single commit boundary. Please avoid adding a retry or a later repair around the two-write sequence; that would retain the duplicate authority and still need to infer whether `null` came from the user or the interrupted migration.

Astro-Han (commented):
Thanks for the careful semantic rebase. It correctly absorbs #3330’s authorization model without restoring the removed protocol filter or a fallback allowlist. The two concerns from the previous review are materially addressed: snapshot metadata remains the candidate authority while the send settles transient availability, and untouched legacy enabled selections now have an upgrade path with an intended default repair.

I followed that upgrade path one layer deeper through the actual historical persisted document shape and the connection-catalog store’s update/default-retention semantics. That exposed one remaining P2 at the migration transaction boundary. This was hidden by the current-head fixture and is not a problem with the semantic rebase itself.

I left one inline comment with the clean final state. I would prefer to close it at the connection-catalog authority in one atomic mutation, rather than add a retry or a next-start repair around the bootstrap sequence.

There is also a narrow, non-blocking P3 follow-up in the quarantine path: a previously saved quarantined default can still be reconstructed as a missing-default catalog entry, and the OpenCode Free connection-test candidate loop does not filter `brokenModelIds`. Actual execution is still vetoed by `authorizeConnectionModel`, so this can be handled later without holding up the migration convergence.

AI-assisted review disclosure: OpenAI Codex coordinated exact-head adversarial verification of the semantic rebase, historical seed producers, persisted catalog shape, CAS/default-target behavior, quarantine consumers, tests, and live GitHub state. I verified the current head, current main, CI/review state, cited source paths, reproduction sequence, and severity calibration before preparing this COMMENT.

packages/runtime-host/src/server/bootstrap-runtime-policy.ts
cunninghamcard-bit: Done in 779fd2a — implemented exactly as specified: built-in seed evolution is now one connection-catalog-authority CAS mutation (`migrateSystemSeed` on the catalog owner, wired through the coordinator lane and the stores facade). In a single document write it (1) recognizes the historical seed by exact match, (2) replaces `enabledModelIds` and re-derives `models`/`modelSource`/`modelsFetchedAt` from the current build, (3) retargets a system default the migration removes, (4) leaves an already-null default null, and (5) never touches any other inventory (a reordering counts as user-modified). No retry, no next-start repair — the two-write sequence and its `migrated` flag are deleted.

The regression tests now start from an actual pre-#3409 persisted document (three-model seed, the pinned four-model inventory of that build, default on `deepseek-v4-flash-free`) and verify the migrated row's inventory and retargeted default land together, plus the restart no-op on the far side of the single commit boundary; a second document fixture covers the user-cleared (already-null) default. Both hidden halves you identified — the stale inventory the current-head fixture masked, and the interruption gap — are covered.

The P3 (quarantined saved default reconstructed as a missing-default entry; connection-test candidate loop not filtering `brokenModelIds`) is noted and I'll fold it into the #3449 follow-up as you suggested. Full suite green on the new head.

Astro-Han (approved):
Thanks for seeing this through. The semantic rebase correctly absorbs #3330, and the final `migrateSystemSeed` shape puts built-in seed evolution at the connection-catalog authority: enabled IDs, fallback inventory, and conditional default retargeting now land in one atomic document write while preserving user-cleared defaults and modified selections.

The historical persisted-document fixtures, restart no-op coverage, and removal of the synthetic duplicate test give me confidence in the final state. All previous review threads are resolved, and I have no findings on this exact head.

The narrow quarantine P3 has been publicly recorded for #3449 and does not block this change.

AI-assisted review disclosure: OpenAI Codex coordinated exact-head adversarial review and traced the migration, storage, catalog, and restart paths. I independently verified the final diff, historical fixture shape, atomic-write and user-choice invariants, current head, CI/review state, and the approval recommendation before posting; no unverified AI output is presented as a finding.

PR #3430, opened 2026-08-21
feat(desktop): show the running build on the sidebar footer

Joob1n:
## Summary

`PR-BUILD-HYGIENE-0` put a build stamp on the About page so a dev build could be told apart from a release. It works, and it is two clicks and a scroll away — so the question it answers gets asked where it cannot answer it.

I hit exactly that while working in this repo: a locally built app was mistaken for the installed release, and a change that had merged looked like it had not landed. The version was on screen the whole time, one page away.

The rail footer already had the shape for it. `.maka-sidebar-footer-row` is a flex row whose settings item is `flex: 1`, so a sibling lands flush right — the slot the update button uses.

```
⚙ Settings                    v0.1.11 · 4d223d0
```

Order is settings, stamp, update button. The button is an action and keeps the edge, where a control is reached for; the stamp is a label, and outside the button it would move every time an update arrived or was installed. A dev build never renders that button — `app-update-service.ts` short-circuits to `not-available` when `!isPackaged` — so on a dev build the stamp sits at the edge on its own, and on a release it yields without changing its relation to Settings.

The commit appears only on a dev build: `build-info.ts` reads `.git/HEAD` and returns null once packaged, which is exactly the case where a version number alone cannot separate two builds. Seven hex characters is what `git log --oneline` prints and how a commit is quoted in review.

Read once at mount through `window.maka.app.info()` — neither value can change without the process restarting. A failed read renders nothing: an orientation label showing an error where a version belongs is worse than one that is absent, and the About page reports that failure where a user went looking for it.

## Verification

Ran `npm run dev` and read the footer on a dev build:

```
⚙ 设置                                    v0.1.11 · 44c0741
```

A release renders `v0.1.11` alone, since `build-info.ts` has no `.git` to read once packaged.

image

Tests cover both build shapes, the pre-resolution empty state, and the order against the update button. The order assertion was checked in both directions — moving the stamp past the button fails that test and only that test — so it pins the layout rather than describing it.

Gates run: `build`, `lint`, `format:check`, `typecheck`, `knip` on `apps/desktop` and `packages/ui`, `astryx:surface-inventory`, `astryx:theme --check`. Suites: `@maka/ui` 187 (183 before this change), `@maka/desktop` 1023. Storybook builds and all 145 stories render.

Not run: Windows and Linux packaging; this is renderer-only and platform-independent, but the rail is only exercised on macOS here.

## Review focus

**The collapsed rail.** The stamp is hidden below 48px rather than clipped or allowed to widen the rail — and `readSessionListCollapsed()` returns `true` when nothing is stored, so a first run has the rail collapsed and the stamp does not appear until it is expanded.

I do not think a truncated version string earns 48px, and the About page still carries it. But this is the judgement most worth a second opinion: if the rail's default state matters more than its width, the fallback is a shorter form (`0.1.11` — no prefix, no commit) at a smaller size rather than nothing at all.

## AI use

Select exactly one:

Tool(s) and scope: Claude Code — traced the existing build-info plumbing, made the change, wrote the tests, and ran the verification above. `Generated-by` is on the commit. Reviewed and submitted by the contributor of record.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/src/main/build-info.ts
+++ b/apps/desktop/src/main/build-info.ts
@@ -69,18 +69,43 @@ function resolveGitDir(repoRoot: string): string | null {
   }
 }
 
+/**
+ * The Git directory holding refs for this checkout.
+ *
+ * A linked worktree splits its Git directory in two: the private admin
+ * directory named by `.git` holds that worktree's own `HEAD` and index, while
+ * loose refs and `packed-refs` stay in the *common* directory shared with the
+ * main checkout. `commondir` inside the admin directory names it, relative to
+ * the admin directory itself.
+ *
+ * Reading a branch ref from the admin directory therefore finds nothing on a
+ * real branch-linked worktree, and the build stamp loses its commit.
+ */
+function resolveCommonDir(gitDir: string): string {
+  try {
+    const pointer = readFileSync(join(gitDir, 'commondir'), 'utf8').trim();
+    if (!pointer) return gitDir;
+    return isAbsolute(pointer) ? pointer : resolve(gitDir, pointer);
+  } catch {
+    // An ordinary clone has no `commondir`; its own directory is the common one.
+    return gitDir;
+  }
+}
+
 function resolveCommit(repoRoot: string): string | null {
   const gitDir = resolveGitDir(repoRoot);
   if (!gitDir) return null;
+  // HEAD is per-worktree; refs are shared.
+  const commonDir = resolveCommonDir(gitDir);
   try {
     const headPath = join(gitDir, 'HEAD');
     if (!existsSync(headPath)) return null;
     const head = readFileSync(headPath, 'utf8').trim();
     if (head.startsWith('ref: ')) {
-      const refPath = join(gitDir, head.slice(5).trim());
+      const refPath = join(commonDir, head.slice(5).trim());
       if (!existsSync(refPath)) {
         // Packed refs path — read packed-refs and match the ref name.
-        const packed = join(gitDir, 'packed-refs');
+        const packed = join(commonDir, 'packed-refs');
         if (!existsSync(packed)) return null;
         const target = head.slice(5).trim();
         const lines = readFileSync(packed, 'utf8').split('\n');
--- a/apps/desktop/src/main/build-info.ts
+++ b/apps/desktop/src/main/build-info.ts
@@ -16,8 +16,8 @@
  * One-shot — captured at module load and cached.
  */
 
-import { readFileSync, existsSync } from 'node:fs';
-import { join, resolve, dirname } from 'node:path';
+import { existsSync, readFileSync, statSync } from 'node:fs';
+import { dirname, isAbsolute, join, resolve } from 'node:path';
 
 export interface BuildInfo {
   readonly mode: 'dev' | 'packaged';
@@ -38,16 +38,49 @@ function findRepoRoot(start: string): string | null {
   return null;
 }
 
-function resolveCommit(repoRoot: string): string | null {
+/**
+ * The directory holding this checkout's Git metadata.
+ *
+ * `.git` is a directory in an ordinary clone and a **file** in a linked
+ * worktree, where it holds `gitdir: <path>` pointing at
+ * `<main>/.git/worktrees/<name>`. Reading `<root>/.git/HEAD` therefore finds
+ * nothing in a worktree — and a worktree is exactly the checkout whose commit
+ * a developer most needs to see, since it is the one that differs from the
+ * tree they think they are running.
+ *
+ * A relative `gitdir:` is resolved against the checkout, which is how Git
+ * writes it for a worktree created inside the repository.
+ */
+function resolveGitDir(repoRoot: string): string | null {
+  const dotGit = join(repoRoot, '.git');
   try {
-    const headPath = join(repoRoot, '.git', 'HEAD');
+    if (statSync(dotGit).isDirectory()) return dotGit;
+  } catch {
+    return null;
+  }
+  try {
+    const pointer = readFileSync(dotGit, 'utf8').trim();
+    if (!pointer.startsWith('gitdir:')) return null;
+    const target = pointer.slice('gitdir:'.length).trim();
+    if (!target) return null;
+    return isAbsolute(target) ? target : resolve(repoRoot, target);
+  } catch {
+    return null;
+  }
+}
+
+function resolveCommit(repoRoot: string): string | null {
+  const gitDir = resolveGitDir(repoRoot);
+  if (!gitDir) return null;
+  try {
+    const headPath = join(gitDir, 'HEAD');
     if (!existsSync(headPath)) return null;
     const head = readFileSync(headPath, 'utf8').trim();
     if (head.startsWith('ref: ')) {
-      const refPath = join(repoRoot, '.git', head.slice(5).trim());
+      const refPath = join(gitDir, head.slice(5).trim());
       if (!existsSync(refPath)) {
         // Packed refs path — read packed-refs and match the ref name.
-        const packed = join(repoRoot, '.git', 'packed-refs');
+        const packed = join(gitDir, 'packed-refs');
         if (!existsSync(packed)) return null;
         const target = head.slice(5).trim();
         const lines = readFileSync(packed, 'utf8').split('\n');
--- a/apps/desktop/src/renderer/app-shell-build-stamp.ts
+++ b/apps/desktop/src/renderer/app-shell-build-stamp.ts
@@ -1,0 +1,34 @@
+import { useEffect, useState } from 'react';
+import type { SidebarBuildStamp } from '@maka/ui';
+
+/**
+ * The running build, for the rail's footer.
+ *
+ * Read once at mount and never again: neither the version nor the commit can
+ * change without the process restarting, so a subscription would re-render for
+ * a value that cannot move.
+ *
+ * Failure is silent and the stamp is simply absent. This is an orientation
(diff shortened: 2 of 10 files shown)


Joob1n:
@Astro-Han , could you merge this pr?

Astro-Han (commented):
Thanks for putting the build identity where developers actually look, and for reusing the existing `app:info` path rather than introducing another main-process authority. The expanded/collapsed behavior and the release-versus-development formatting are clear, and the live `test` check is green.

I’m leaving one non-blocking P2 follow-up below.

The only correctness gap I found is specific to the development workflow: a normal Git worktree has a `.git` file containing a `gitdir` pointer, rather than a `.git` directory. The existing `build-info` resolver therefore returns `commit: null` in that checkout, so this footer falls back to the version alone and still cannot distinguish that locally built tree from a packaged release. The clean fix is to extend the existing main-process `build-info` authority to follow worktree metadata and cover that case with a focused resolver test; the renderer should remain a pure consumer of `app:info` rather than growing a second Git reader.

As a small P3 polish item, at the supported 180px expanded width a packaged build with an update reminder can slightly truncate the Settings label. The icon, action, and accessible name remain available, so this is not a merge blocker; making the stamp the shrinkable/ellipsis item while preserving its full title would keep both pieces readable.

AI-assisted review disclosure: OpenAI Codex coordinated the exact-head review. I independently verified the current head, base, CI and review state, traced the `app:info`/`build-info` data flow against current `main`, reproduced the Git-worktree behavior, and checked the sidebar geometry. The recommendation above is my reviewed judgment; no unverified AI output is being presented as a finding.

apps/desktop/src/renderer/app-shell-build-stamp.ts
```diff
@@ -0,0 +1,34 @@
...
+export function useBuildStamp(): SidebarBuildStamp | undefined {
+  const [stamp, setStamp] = useState<SidebarBuildStamp | undefined>(undefined);
+
+  useEffect(() => {
+    let cancelled = false;
+    window.maka.app
+      .info()
+      .then((info) => {
+        if (cancelled) return;
+        setStamp({ version: info.appVersion, commit: info.buildCommit });
```
Astro-Han, line 25: [P2] Thanks for wiring the footer to the existing `app:info` result. One development-path gap remains: when Maka is run from a normal Git worktree, `.git` is a file containing a `gitdir` pointer, so the current main-process `build-info` resolver returns `commit: null`. This makes the new stamp render only `v<version>` and fails to distinguish the locally built tree from a packaged release—the exact problem this change is meant to solve for contributors. Could we extend the existing `build-info` authority to follow worktree metadata and add a focused resolver test, while keeping the renderer as a pure `app:info` consumer rather than adding a second Git reader here?

Joob1n:
Both addressed in `45aba559d`.

**P2 — the worktree case reproduces, and the fix is in `build-info` as you asked.**

`resolveCommit` opened `<root>/.git/HEAD`, and in a linked worktree `.git` is a file holding `gitdir: <path>`. Confirmed against a real worktree in this checkout before changing anything: the resolver returned `commit: null` there.

The main process now follows the pointer, resolving a relative `gitdir:` against the checkout since that is how Git writes it for a worktree created inside the repository. The renderer stays a pure consumer of `app:info`.

Verified on the same real worktree afterwards: repository root reports `96f1bc8`, worktree reports `5d9a5b0`, each matching `git rev-parse` — two different commits, which is the distinction the stamp exists to make.

The resolver had no test file. It has one now covering ordinary clone, worktree by absolute and relative pointer, detached HEAD, packed refs, and the two unreadable cases — where absence stays a supported answer rather than throwing out of `app:info`, which carries much more than the commit.

**P3 — the stamp yields, the label does not.**

It was `flex: none` against a `flex: 1` settings row, so at 180px with an update button the label lost. The stamp is now the shrinkable item: it ellipsises and keeps the full build in its `title`, while the label a user reads and the row they aim at stay intact.

Gates: build, lint, format:check, typecheck. Suites: desktop 1030 (1023 before, +7 from the new resolver test), ui 187.

apps/desktop/src/main/build-info.ts
```diff
@@ -38,16 +38,49 @@ function findRepoRoot(start: string): string | null {
...
+  const gitDir = resolveGitDir(repoRoot);
+  if (!gitDir) return null;
   try {
-    const headPath = join(repoRoot, '.git', 'HEAD');
+    const headPath = join(gitDir, 'HEAD');
     if (!existsSync(headPath)) return null;
     const head = readFileSync(headPath, 'utf8').trim();
     if (head.startsWith('ref: ')) {
-      const refPath = join(repoRoot, '.git', head.slice(5).trim());
+      const refPath = join(gitDir, head.slice(5).trim());
```
Astro-Han: **[P2] Follow the worktree `commondir` when resolving branch refs**

Thanks for moving the fix into the existing `build-info` authority. The `.git → gitdir` hop is now handled, but a normal branch-linked worktree has one more layer: its private admin directory contains `HEAD` and a `commondir` file, while the loose refs and `packed-refs` live in the common Git directory.

The current resolver still looks for the branch ref inside the private worktree admin directory, so real branch worktrees return `commit: null`. The new fixture passes because it places `refs/heads/feature` in that private directory, which Git normally does not do.

The clean fix is to keep reading `HEAD` from the worktree admin directory, resolve `commondir`, and read loose or packed refs from the resulting common Git directory. Updating the existing fixture to match that real layout should be sufficient; no additional renderer or IPC path is needed.

apps/desktop/src/renderer/styles/sidebar.css
```diff
@@ -166,7 +166,10 @@
 }
 
 .maka-sidebar-footer-row-primary {
-  flex: 1;
+  /* `auto` rather than `1 1 0`: the settings row asks for the width its label
+     needs and gives back the rest, so the stamp shrinks into what is left
+     instead of the label being clipped first. */
+  flex: 1 1 auto;
```
Astro-Han, line 179: **[P3] Let the build stamp yield before the Settings label**

This is much better than the previous fixed-width stamp. At the supported 180px width with an update reminder present, however, both the Settings row and the stamp can still shrink, so the Settings label may retain a slight ellipsis. Preserving the Settings row's minimum content width and letting only the stamp truncate would make the intended priority deterministic. The icon, action, and accessible name remain intact, so this is only polish.

Astro-Han (dismissed):
Thanks for keeping build identity in the existing main-process `build-info → app:info` authority and for improving the narrow footer layout. The new resolver handles ordinary clones, detached worktrees, and the initial `.git` pointer correctly.

I'm approving this exact head with one non-blocking P2: a normal branch-linked worktree has one additional `commondir` hop, so that specific path is not fully resolved yet. I have also left a small P3 layout note.

AI-assisted review disclosure: OpenAI Codex coordinated two independent exact-head review passes and reproduced the resolver against real linked worktrees. I verified the source, fixtures, footer geometry, exact head, CI and existing review state, and I independently made the approval and severity decisions.

Joob1n:
Both addressed in `e147a1e6a`. You were right on the worktree case, and right about why my fixture did not catch it — that part stung more than the bug.

**P2 — reproduced before fixing.** I ran `git worktree add` and looked at what Git actually writes:

```
main-repo/.git/worktrees/linked/
  commondir  gitdir  HEAD  index  logs  ORIG_HEAD  refs/
```

`refs/` is there but has no `heads/` in it, and `commondir` holds `../..`. So the resolver read `HEAD` from the admin directory — correct — then looked for `refs/heads/<branch>` beside it, where Git never puts one. Real branch worktrees returned `commit: null`.

Fixed by keeping `HEAD` per worktree and resolving refs — loose and packed — through `commondir`. An ordinary clone has no `commondir` and falls back to its own directory. Verified against the real worktree:

```
真实 linked worktree -> {"mode":"dev","commit":"96228ca"}
主仓库              -> {"mode":"dev","commit":"96228ca"}     # git rev-parse agrees
```

**On the fixture.** You were exact: it placed `refs/heads/feature` inside the private admin directory, which Git does not do, so it certified the behaviour I had written rather than the behaviour Git produces. It now mirrors the real split, and a second case covers a packed ref in the common directory. **Both fail when the `commondir` hop is removed** — I checked by removing it, not by reading them.

**P3 — measured rather than reasoned about.** I rendered the app shell at 180px with the update reminder present and read the layout back from the DOM. Chinese `设置` is short enough to survive either way; the English label is where it shows:

| | Settings label | |
| --- | --- | --- |
| before (`min-width: 0`) | rendered 30px against 53px of content | **clipped** |
| after (`min-content`) | 93px, nothing in the row truncates | clean |

The stamp is now the only thing that yields, which is the priority the comment always claimed.

## Verification

- `@maka/desktop` suite: **1031/1031**
- `biome lint apps/desktop/src`: 707 files, clean
- `build-storybook` + `smoke:storybook`: 145 stories, AX audit included

Astro-Han (approved):
Thanks — I re-reviewed exact head `e147a1e6a6588eb0fac6d1cd32e25285cc831cac`. The resolver now follows Git's real linked-worktree layout: it reads the worktree-specific `HEAD` from the private admin directory and resolves loose or packed branch refs through `commondir`. The fixtures now reproduce that split rather than placing refs where Git does not write them. The footer sizing change also makes the build stamp yield before the Settings label at the supported narrow width.

I found no remaining P0–P3 issue and am approving this exact head. The hosted `test` check is still running, so this approval is not a substitute for its result.

AI-assisted review disclosure: OpenAI Codex assisted with the exact-head delta review. I verified the Git worktree/ref authority path, focused fixtures, footer layout ownership, current-main merge result, and live GitHub state, and I independently made the approval decision.

PR #3369, opened 2026-08-21
chore(docs): recognize Selector as an Astryx re-export in the surface inventory

GabrielDrapor:
Split out of #2921 per review: a one-line allowlist addition (`Selector` re-exports from @maka/ui like its siblings) plus the mechanically regenerated inventory rows. No behavior change.

Co-Authored-By: Claude <noreply@anthropic.com>

https://claude.ai/code/session_01TMwYxgNEbz2RFmuK6AXGcj

PR #3431, opened 2026-08-21
feat(desktop): let the user choose or import the app icon

ARE404:
Settings → 外观 grows an **App icon** picker: a grouped set of shipped icons,
plus the ability to import your own.

> Builds on #3451. The first commit here is that fix, submitted separately
> because it is a pre-existing bug: `assets/` is not packaged, so the window
> icon path has been dead in packaged builds independently of this feature.

## What it does

- picks between the shipped icons and applies the choice immediately — macOS
  dock, Windows/Linux per-window
- **imports your own artwork**, stored as a normalized 1024px PNG
- survives restart, and falls back to the brand mark whenever artwork cannot
  be read

## Design notes worth your time

**`appearance.appIcon` stays a closed value, not a path.** Shipped ids are an
enum; an imported icon is `custom:<32 hex>`. The main process generates the
id, the id *is* the whole file name, and both core's normalization and the
store re-check its shape before it reaches `join`. A hand-edited settings
file cannot name a file — the worst it can do is name nothing, which
normalizes to the brand mark. The renderer never learns a path; the file
dialog runs in the main process, the way PetPack import already does.

**Imported files are re-encoded, not copied.** Decode → centre-crop to square
→ scale to 1024 → write PNG. The stored shape is then the only one the rest
of the app reasons about, and nothing rides along inside the original.
Non-square art is cropped rather than letterboxed: silently adding a margin
would reframe art the user already framed.

**Artwork carries the 824/1024 margin itself.** The OS scales the canvas to
the dock tile and never adds margin, so a full-bleed tile sits ~24% wider
than every neighbour. Measured against the shipped `icon.png` and against
other installed apps; the import hint says the same thing to the user.

**Effects hang off `client-settings-effects.ts`,** already the single entry
point for OS-level side effects of a settings write, which gives live apply
and file-watch echo dedupe for free. Startup still sets the icon
synchronously in `desktop-shell-presentation.ts` — waiting for the settings
store would let the dock flash the Electron rocket first, the regression
PR-GRAY-CARD-LIFT-0 fixed.

**An unreadable icon file is an EMPTY `NativeImage`, not an error.** Handing
that to `setIcon` blanks the dock tile, so emptiness is the failure test and
what advances the fallback chain.

## Grouping

16 tiles in one grid gives the eye nowhere to start — the problem
`PALETTE_GROUPS` already solved for palettes, so this follows that shape
including the per-group `role="group"` and label id. Group membership is a
renderer concern: the main process reports what artwork loaded, and anything
no group claims (imported art) lands in the trailing group rather than
disappearing.

## Verification

- `tsc --noEmit` clean for main, preload and renderer
- `node --test` — 13 main-process tests (icon paths, packaged resolution,
  imported-id path safety, settings effects), 589 in `@maka/core`
- path safety is covered explicitly: `../../../etc/passwd`, wrong length,
  uppercase, empty, embedded slash — all refused before `join`
- picker interaction, dialog and dock switching verified by hand in `npm run dev`

## Open questions

1. **The artwork.** The shipped set is my own drawing, not a proposal to
   replace the brand mark — `default` stays the default, and the mascot is
   untouched. If a shipped set is a decision that belongs elsewhere, the
   import feature stands on its own and I will happily drop the 14 files.
2. **Scope.** Import can be split out if you would rather review the picker
   first.


apps/desktop/src/main/app-icon-surface.ts
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * One choice's artwork path — shipped art under the asset root, imported art
+ * under the directory the app owns. Never throws: a malformed id can only come
+ * from a settings file that dodged normalization, and a window being created
+ * is no place to raise. The brand mark is the answer to every such question.
+ */
+export function appIconPath(choice: AppIconChoice): string {
+  const custom = customAppIconId(choice);
+  if (custom === undefined) return resolveAppIconPath(currentAssetRoot(), choice as AppIcon);
```
hqhq1025, line 50: [P1] Reject non-catalog values before constructing a shipped asset path. `AppIconChoice` is only a compile-time type: `settings:client:update` accepts renderer input at runtime, `clientOwnedSettingsPatch()` forwards `appearance` unchanged, and `SettingsStore.update()` merges without normalization. Therefore `appearance.appIcon = "../../../../tmp/owned"` reaches this cast and resolves outside the asset root (`/tmp/owned.png` in a direct reproduction), after which Electron's native decoder reads it. Use `isAppIcon(choice)` here (falling back or rejecting) and normalize/validate the generic settings update boundary as defense in depth; add a test that sends a malformed runtime settings patch.

apps/desktop/src/renderer/settings/appearance-settings-page.tsx
```diff
@@ -90,6 +118,70 @@ export function AppearanceSettingsPage(props: {
...
+      toast.error(copy.appIconImportError, settingsActionErrorMessage(error, locale));
+    } finally {
+      setAppIconBusy(false);
+    }
+  }
+
+  async function removeAppIcon(icon: AppIconChoice) {
+    setAppIconBusy(true);
+    try {
+      await window.maka.app.removeIcon(icon);
```
hqhq1025: [P2] Reset the selected setting before deleting its file, and do not swallow the persistence result in this destructive sequence. The comment below says to hand the selection back first, but deletion happens here first. `setAppIcon()` calls `persistAppearance()`, which catches update failures and resolves, so a disk/IPC failure leaves the saved setting pointing at an already-deleted icon while this function continues as if it succeeded. This also races because the selection cards remain enabled while `appIconBusy` is true. Prefer one main-process operation that coordinates the setting reset and deletion, or at minimum confirm the reset succeeded before deletion and disable all icon choices during the operation.

apps/desktop/src/main/custom-app-icons.ts
```diff
@@ -0,0 +1,63 @@
...
+}): Promise<CustomAppIcon> {
+  const info = await stat(input.sourcePath).catch(() => undefined);
+  if (!info?.isFile()) throw new CustomAppIconError('unreadable', 'not a file');
+  if (info.size > CUSTOM_ICON_MAX_INPUT_BYTES) {
+    throw new CustomAppIconError('too_large', `over ${CUSTOM_ICON_MAX_INPUT_BYTES} bytes`);
+  }
+
+  // An undecodable file comes back EMPTY rather than throwing — the same trap
+  // the shipped set falls into, and the reason emptiness is the failure test.
+  const source = nativeImage.createFromPath(input.sourcePath);
```
hqhq1025: [P2] Bound decoded dimensions before decoding untrusted image data in the main process. The 16 MB `stat()` limit only caps compressed bytes; a highly compressed image can expand to a very large bitmap and exhaust the Electron main process before `getSize()` runs. There is also a TOCTOU window because the path can be replaced between `stat()` and `createFromPath()`. Read one size-capped file snapshot, validate its format and pixel dimensions before native decoding (or use a decoder with explicit resource limits), then decode that same buffer. Add a compressed oversized-dimension fixture to exercise the limit.

apps/desktop/src/main/app-ipc-main.ts
```diff
@@ -57,6 +65,40 @@ export function registerAppClientIpc(
...
+  // no id from the renderer ever reaches the filesystem.
+  targetIpc.handle('app:iconPreviews', (): Promise<readonly AppIconPreview[]> =>
+    listAppIconPreviews(),
+  );
+  // The picker hands back a choice it was given, never a path: the dialog runs
+  // here and the file it returns is the only path this ever sees.
+  targetIpc.handle('app:importIcon', async (): Promise<CustomAppIconImportResult> => {
+    const picked = await mainWindowController.showOpenDialog({
+      properties: ['openFile'],
+      filters: [{ name: 'Images', extensions: ['png', 'jpg', 'jpeg', 'tiff', 'webp'] }],
```
hqhq1025: [P2] Do not advertise TIFF and WebP when the decoder used below does not guarantee them cross-platform. Electron's current `nativeImage` documentation guarantees PNG and JPEG across all platforms (and ICO paths on Windows); a Windows/Linux user can select one of these advertised extensions and immediately get an unreadable-image failure. Restrict the dialog to PNG/JPEG, or add an explicit cross-platform decoder and tests for every advertised format.

apps/desktop/src/main/main-window.ts
```diff
@@ -263,12 +265,16 @@ export function createMainWindowController(deps: MainWindowControllerDeps): Main
...
+      // app icon ships as a 1024px PNG under apps/desktop/assets/. BrowserWindow
+      // accepts a PNG path directly on macOS for the dock / window title bar;
+      // .icns / .ico packaging will come with the installer build pass. The
+      // asset path resolves from the built dist/main/main.js (two levels up to
+      // apps/desktop, then assets).
+      //
+      // Windows and Linux draw this icon per window, so a window opened after
+      // the user switched icons must be born with the chosen one — waiting for
+      // the client-settings effect to catch up would show the default first.
+      icon: appIconPath(persistedAppearance?.appIcon ?? 'default'),
```
hqhq1025: [P2] Use the same readable-image fallback used by `applyAppIcon()` when creating a window. `appIconPath()` returns a syntactically valid custom path even when that file is missing or corrupt, so a newly created Windows/Linux window can receive an invalid icon path while existing windows fall back to the default through `loadAppIcon()`. Resolve and validate the image once through the shared fallback path, rather than giving `BrowserWindow` the raw path. Add a test for a persisted custom ID whose file has disappeared.

hqhq1025 (commented):
Reviewed exact head `cd6178b255441496cd13052b5a3deaf9072e1bcd`.

The problem definition is sound: users need a persistent app-icon choice, including imported artwork, and the patch correctly separates shipped assets from app-owned custom files. The broad mechanism is also reasonable: persist an opaque icon ID, resolve files only in the main process, update current OS surfaces, and package the shipped masters as extra resources.

However, the current implementation does not yet preserve those boundaries end to end. I found one main-process path escape and four correctness/robustness issues below. In particular, the generic settings IPC accepts an arbitrary runtime string, and `appIconPath()` then casts it to a shipped icon ID; a value such as `../../../../tmp/owned` resolves to `/tmp/owned.png` and is handed to Electron's native decoder. The delete flow is also ordered opposite to its comment, so a failed settings write can leave the selected file deleted while the setting still points at it.

Additional release gap: `assertPackagedResources()` in `scripts/verify-packaged-app.mjs` does not require `assets/icon.png` or the `assets/app-icons/*` catalog. The source tests prove that repository files and path construction are correct, but they do not fail if a future packaging change drops the runtime payload. Please make the packaged-resource verifier require the default icon and every `APP_ICONS` asset. I locally produced the unsigned `.app` up to the mandatory signing step and confirmed this revision currently copies all 16 PNGs byte-for-byte, but that guarantee is not encoded in CI.

First-principles / Occam assessment: the opaque-ID model is the right simple boundary, but validation must happen at every runtime ingress and destructive operations should have a single main-process owner. A small main-process operation that validates the choice and coordinates selection reset with deletion is simpler and safer than renderer-side sequencing plus swallowed persistence errors. No broad rewrite is required.

Deletion/test quality: `desktopAssetPath()` appears to be used only by its own test; it and the redundant assertion can be removed unless a production caller is planned. `custom-app-icons.ts` also leaves a trailing blank line that makes `git diff --check` fail. The 14 decorative variants are not technically dead, but whether Maka wants to ship a second geometric mark is a product/brand decision and could be separated from the picker mechanics.

Verification performed locally: clean merge-tree against current `main`; full Desktop production build; all-workspace typecheck; Biome format check; 17 focused settings/icon tests; source PNG dimension checks; and inspection of the generated macOS app resources. GitHub currently reports both workflows as `action_required` with no jobs, so no hosted CI checks have run for this head.

Verdict: not ready to merge at this revision. Please fix the inline findings, add malformed settings/import/removal/package-boundary tests, then re-request review. Residual platform risk remains around Windows/Linux icon updates and should be covered by at least one packaged smoke check after the cross-platform format filter is corrected.

ARE404:
Thanks — the two runtime findings were both real, and I confirmed them before fixing rather than taking them on faith.

Rebased onto current `main` (which now carries #3451, so the duplicate commit is gone and the conflict with it).

**Path escape — confirmed and fixed.** `normalizeSettings` runs on *read*; `SettingsStore.update` merges and writes without it, so the snapshot the effects act on carries whatever the patch put there. `appIconPath` then cast it. Every runtime ingress now coerces through a new `toAppIconChoice` — `appIconPath`, `applyAppIcon`, and the settings effect — and the declared type is no longer treated as a guarantee. Regression tests at both layers: the coercion itself in `@maka/core`, and a path-level assertion that a malformed choice cannot resolve outside the asset root.

**Delete ordering — confirmed and fixed.** The code did the opposite of its comment. Removal is now one main-process operation that resets the selection, applies it, then deletes; the renderer no longer sequences anything. `app-icon-ipc.ts` takes its Electron-touching capabilities as parameters so the contract is testable without a browser process — three tests cover reset-then-delete, the untouched-selection case, and refusal of shipped ids and malformed references.

**Packaged-resource verifier.** `assertPackagedResources` now requires `assets/icon.png` and every `APP_ICONS` file, plus a test that the check fails when one is dropped. You are right that the source tests proved the wrong thing: they showed the repository was correct, not that the payload survives packaging.

**`desktopAssetPath`** — removed along with the redundant assertion; its only production caller became `appIconPath` in this branch.

**Trailing blank line** — fixed, `git diff --check` is clean.

On the 14 variants: I have raised the brand question with the maintainers separately and was told to go ahead, so they stay for now. If the PMC would rather decide that on its own thread, say the word and I will drop the artwork commit — the picker and the import path stand without it.

Not fixed, and I want to be straight about it: I still cannot verify Windows/Linux icon updates or run a packaged smoke check here. This machine cannot start an Electron GUI, so the picker and dock switching were verified by hand on macOS only, and the cross-platform format filter you flagged is untested by me.

Verification this round: `tsc --noEmit` clean for main/preload/renderer; Biome lint and format clean; 17 main-process tests, 2 packaging-boundary tests, 591 in `@maka/core`; `git diff --check` clean.

Re-requesting review.

apps/desktop/src/main/app-icon-surface.ts
ARE404: Confirmed and fixed in 17d37be. You are right that the type was doing no work at runtime: `normalizeSettings` runs on read, while `SettingsStore.update` merges and writes without it, so the snapshot the effects act on carries whatever the patch put there.

Added `toAppIconChoice` in `@maka/core` and coerced at every runtime ingress — `appIconPath`, `applyAppIcon`, and the settings effect — so no path is built from an unchecked value. Two regressions: the coercion itself, and a path-level assertion that a malformed choice cannot resolve outside the asset root.

apps/desktop/src/renderer/settings/appearance-settings-page.tsx
ARE404: Confirmed and fixed in 17d37be. The code did the opposite of its own comment.

Selection and artwork are one pair of state, so the operation moved to a single main-process owner (`app-icon-ipc.ts`): it resets the selection, applies it, and only then deletes. The renderer sequences nothing and no longer swallows the result. It takes its Electron-touching capabilities as parameters, so three tests cover reset-then-delete, the untouched-selection case, and refusal of shipped ids and malformed references.

apps/desktop/src/main/custom-app-icons.ts
ARE404: Fixed in 6cffc61. Both halves of this were real.

The file is now read once into a capped buffer, the header is parsed out of those bytes, dimensions are rejected there (max 4096 per edge), and the *same buffer* is decoded via `createFromBuffer`. That removes the second read entirely, so there is no window to swap the path in, and no decoder sees data whose declared size has not been checked.

The header parser is pure and tested, including the fixture you asked for: a sub-1 KB PNG declaring 30000×30000.

apps/desktop/src/main/app-ipc-main.ts
ARE404: Fixed in 6cffc61 — the dialog now offers PNG and JPEG only.

The handler moved to `app-icon-ipc.ts` in the meantime. Worth noting the two ends can no longer drift: the header parser recognises exactly PNG and JPEG, so a format the dialog offered but the decoder would refuse is now a compile-and-test-level impossibility rather than a comment.

apps/desktop/src/main/main-window.ts
ARE404: Fixed in 6cffc61. A well-formed path says nothing about the file existing, and per-window icons on Windows/Linux are exactly where that shows.

Window creation now walks the same fallback the dock does, through a pure `pickReadableAppIconPath(icon, toPath, isReadable)`. Keeping the readability probe injectable means the missing-file case is covered by a test rather than needing a browser process: a persisted custom id whose file disappeared resolves to the brand mark, and an id whose file is present is still honoured.

scripts/verify-packaged-app.mjs
```diff
@@ -718,6 +728,18 @@ export async function assertPackagedResources(
...
+    // packaging change that drops them would ship a blank dock tile silently.
+    //
+    // The list comes from the artwork directory rather than from `APP_ICONS`,
+    // because that enum only exists as build output: importing it would make
+    // this verifier — and its test — unable to even load before a compile.
+    // The other half of the chain is already covered where it belongs, by the
+    // desktop test that walks `APP_ICONS` and requires a 1024px master for
+    // every id.
+    join('assets', 'icon.png'),
+    ...(await packagedIconCatalog()),
```
Astro-Han: **[P1] Preserve the legacy artifact contract**

The Windows upgrade lifecycle verifies the previously released installer with `artifactContract: 'legacy-baseline'`, but this catalog is derived from the current checkout. All 15 `assets/app-icons/*.png` files were introduced by this PR, so every previous release necessarily lacks them and the normal upgrade verification fails before the upgrade is attempted.

Please make the new icon catalog requirement part of the current artifact contract only—for example, thread a `requireAppIconCatalog` option from `requiresCurrentContract` into `assertPackagedResources()`—while keeping the current package check strict. A contract-level test covering both current and legacy artifacts would close this completely without weakening present-day verification.

apps/desktop/src/renderer/settings/appearance-settings-page.tsx
```diff
@@ -227,6 +367,81 @@ export function AppearanceSettingsPage(props: {
...
+                  {copy.appIconGroups[group.id]}
+                </Text>
+                <Grid
+                  columns={{ minWidth: 180 }}
+                  gap={2}
+                  role="group"
+                  aria-labelledby={appIconGroupLabelId(group.id)}
+                >
+                  {group.options.map((option) => (
+                    <SelectableCard
```
Astro-Han, line 420: **[P2] Fence selection while an icon is being removed**

The remove button becomes disabled, but the icon cards remain selectable while `app:removeIcon` is awaiting its main-process read/reset/delete sequence. A quick selection of the still-visible icon can therefore persist it after the removal handler has read the old selection, leaving the setting pointed at a deleted file. The extra renderer-side `setAppIcon(result.selection)` can also overwrite a newer selection even though `SettingsSurface` already subscribes to `settings:clientChanged`.

The clean final state is to disable every icon card while `appIconBusy` is true and let the existing client-settings change notification reload the authoritative selection, removing the second renderer write.

apps/desktop/src/renderer/settings/appearance-settings-page.tsx
```diff
@@ -227,6 +367,81 @@ export function AppearanceSettingsPage(props: {
...
+                  size="sm"
+                  color="secondary"
+                  weight="medium"
+                >
+                  {copy.appIconGroups[group.id]}
+                </Text>
+                <Grid
+                  columns={{ minWidth: 180 }}
+                  gap={2}
+                  role="group"
```
Astro-Han, line 416: **[P2] Expose this single-choice picker as a radio group**

These choices are mutually exclusive, but `SelectableCard` exposes a hidden checkbox and the container only has `role="group"`. Screen-reader and keyboard users therefore receive independent toggle semantics rather than one-of-many selection semantics.

Please use a real radio/radiogroup contract, or extend the shared card component with a single-select mode and radio keyboard behavior. The visual card treatment can stay unchanged.

Astro-Han (commented):
Thanks for continuing to work through this carefully. I reviewed exact head `9d7e2c28d8b71e02743f60516590793dccd5b839`. The earlier path, decode, geometry, and packaging-presence findings are addressed. I found one release-contract regression introduced by the new packaged-resource check, plus two non-blocking interaction/accessibility issues below.

The desired final state is small: keep the new icon catalog mandatory for current artifacts, preserve the historical contract when verifying a previous release, and make icon selection a single-choice operation that cannot race deletion.

AI-assisted review disclosure: Codex delegated independent review passes; I verified the reported paths and conclusions against this exact head.

scripts/verify-packaged-app.mjs
ARE404: Fixed in 0660b5f. You are right, and it would have failed the upgrade check before the upgrade was attempted.

`requireAppIconCatalog` is now threaded from `requiresCurrentContract` in `verify-windows-x64.mjs`, exactly alongside `requireDisclaimer`, with the current contract left strict.

One thing worth flagging beyond the report: the gate has to cover `assets/icon.png` as well, not only the catalog. `assets/` was not packaged at all before the picker landed — that directory arrived with #3451 — so a legacy baseline lacks the brand mark too, and gating only `app-icons/` would still have failed on `icon.png`.

Contract-level coverage added both ways: the current contract requires every file present in the artwork directory, and a `legacy-baseline` run asks for none of `assets/` while still verifying everything the old contract did carry.

apps/desktop/src/renderer/settings/appearance-settings-page.tsx
ARE404: Fixed in 0660b5f — every card is fenced while a removal is in flight, not just the remove button.

The second write is gone too, and thank you for pointing at `settings:clientChanged`: I checked, the effects fingerprint is a stringify of the whole settings object, so an icon-only change does emit it and the surface reloads on its own. That write came out of my own review pass a few hours ago, where I found `settings:externalChanged`, concluded nothing notified this surface, and stopped looking. It was both unnecessary and a way to stamp a stale value over a newer choice — exactly as you describe.

apps/desktop/src/renderer/settings/appearance-settings-page.tsx
ARE404: Agreed on the semantics — a hidden checkbox for a one-of-many choice is wrong, and `role="group"` does not repair it.

I would rather not fix it only here, and I want to explain why rather than just deferring. `SelectableCard` hard-codes `type="checkbox"`, and this file uses it eight times: theme, palettes, and now icons. All three are single-choice, so the gap is identical in all three and predates this branch. Fixing it in the icon picker alone would leave the same page announcing three visually identical pickers with two different contracts, which seems worse than the current consistent-but-wrong state.

The right shape looks like a single-select mode on the shared card — native radios also give the arrow-key behavior for free — applied to all three pickers at once. I can see `patches/@astryxdesign+core+0.4.0.patch`, so extending the component from this repo is established practice here, and I am happy to do that work.

Leaving this thread open rather than resolving it, since nothing changed. Two ways I can go, your call:

1. a separate PR that adds the single-select mode and converts theme, palettes and icons together — my preference, since it is a design-system change and reviewable on its own
2. do it inside this PR if you would rather not ship a new picker with the old semantics

If you want (2), say so and I will fold it in.

Astro-Han: Thanks — I agree with the proposed boundary. Since the same shared component governs the theme, palette, and app-icon single-choice pickers, fixing only the icon picker here would create inconsistent semantics on the same page. A focused shared-component follow-up that converts all three together is the cleaner final state. Treating this as a non-blocking P2 for this PR.

Astro-Han (approved):
Thanks — I re-reviewed exact head `0660b5f29bcc707ecf9d49e72401c8d6e492346e`. The legacy artifact-contract regression and the deletion/selection race are both fixed cleanly, and the focused verifier contract passes.

I agree that the remaining single-choice semantics issue belongs at the shared `SelectableCard` seam. Theme, palette, and app-icon pickers all use the same checkbox-based contract, so fixing only the new picker here would leave one page with inconsistent interaction semantics. Please address all three together in a focused shared-component follow-up.

Approving this exact head with that non-blocking P2. The hosted CI and Release Windows workflows are still awaiting permission to run, so this approval is not a substitute for their results.

AI-assisted review disclosure: OpenAI Codex coordinated earlier independent review passes. I personally re-verified the latest delta, legacy/current artifact contracts, deletion lifecycle, shared accessibility boundary, current-main merge result, and live GitHub state, and I independently made the approval decision.

apps/desktop/src/main/app-icon-ipc.ts
```diff
@@ -0,0 +1,118 @@
...
+      if (!id) return { ok: false, reason: 'invalid_id' };
+
+      const current = await input.settingsStore.get();
+      let selection = toAppIconChoice(current.appearance.appIcon);
+      if (selection === icon) {
+        // Hand the OS back to the brand mark BEFORE the file goes away. The
+        // other order leaves a persisted choice pointing at nothing whenever
+        // this write fails.
+        try {
+          const next = await input.settingsStore.update({ appearance: { appIcon: 'default' } });
```
Astro-Han: [P2] Thanks for moving this lifecycle into the main process. There is still a small compare-and-set gap here: `get()` can observe the icon as selected, then another renderer or settings-sync path can select a different icon before this unconditional `update()`, causing the stale removal flow to overwrite that newer choice with `default`. The renderer busy flag cannot make this Host-side sequence atomic. Could we use the existing `settingsStore.updateIf()` seam to reset only while the current choice still equals `icon`; if it no longer does, keep the returned latest selection and just delete the now-unselected file. A `get → concurrent select → remove` regression test would pin the final authority without adding another state path.

Astro-Han (commented):
Thanks for getting the previous runtime and packaging issues into a much cleaner final state. I rechecked this exact head with three independent review passes: the earlier path-validation, decode, deletion-ordering, startup-restoration, and legacy-package-contract findings are resolved, and both hosted checks are green.

There is one remaining non-blocking correctness issue in the main-process deletion transaction, noted inline. The clean final state is to make the reset conditional on the selection the deletion observed, so a concurrent newer choice remains authoritative.

AI-assisted review disclosure: OpenAI Codex coordinated three independent review passes. I verified the reported race against this exact head, the existing SettingsStore conditional-update seam, the live review threads, mergeability, and current-head CI, and I made the final review decision.

apps/desktop/src/main/app-icon-ipc.ts
ARE404: Fixed in 3ce30b6. You are right, and the busy flag was never going to reach across the IPC boundary to close it.

`updateIf` runs its predicate and the write on one queue, so the reset now applies only while the current choice is still the icon being removed. When it is not, the newer selection stands and is what the caller is told; the OS surface is re-applied only when the selection actually moved.

The file is deleted either way — at that point it is no longer the one in use, which is the state the caller asked for. Worth confirming that reading is what you intended, since the alternative would be to refuse the delete and make the user click again on an icon that is no longer selected.

This also removed a call rather than adding a path: `updateIf` reports the settings whether or not it applied, so the separate `get()` is gone and the dependency is narrowed to `Pick<SettingsStore, 'updateIf'>`.

Regression test added as asked — a selection lands while the compare-and-set is being evaluated, and it pins all four outcomes: the newer choice wins, it is what is returned, nothing is re-applied, and the artwork is still gone.

apps/desktop/src/main/app-icon-ipc.ts
```diff
@@ -0,0 +1,125 @@
...
+      // store runs the predicate and the write on one queue, so this is the
+      // only place the pair can be made atomic — no renderer busy flag
+      // reaches across the IPC boundary to do it.
+      //
+      // When the predicate no longer holds, the newer selection stands and the
+      // file is still deleted: it is no longer the one in use, which is
+      // exactly the state the caller asked for.
+      let settings: AppSettings;
+      try {
+        const outcome = await input.settingsStore.updateIf(
```
Astro-Han, line 198: [P2] This CAS protects only the settings write. After it returns, `applySettings()` yields before the file deletion; during that window another renderer or an already queued generic `settings:client:update` can select this same custom icon again. Removal then deletes the file and leaves the newest setting pointing at missing artwork. Please do not add another post-delete CAS: the clean final state is to put app-icon selection behind this main-process icon seam as well, serialize select/import/remove under one owner, validate that a selected custom asset still exists, and remove the renderer's generic `appearance.appIcon` write path. Then one interleaving test can prove a selection cannot enter between reset/apply/delete, rather than continuing to patch individual windows.

Astro-Han (commented):
Thanks for closing the original `get → update` race with the existing conditional-write seam. Two independent exact-head passes confirmed that this CAS is correct for changes that happen before its predicate, but they also reproduced one remaining selection/file lifecycle window after the CAS. I am describing the final authority boundary inline so we can finish this without another local patch.

AI-assisted review disclosure: OpenAI Codex coordinated two independent exact-head re-reviews. I verified the retained interleaving, the current SettingsStore and IPC ownership seams, live threads, mergeability, approval state, and current-head CI progress, and I made the final review decision.

apps/desktop/src/main/app-icon-ipc.ts
ARE404: Fixed in 3fdfac7 — and you were right that guarding the window would only move it.

Selection now goes through the seam rather than racing it. `app:selectIcon` persists and applies the choice, select/import/remove run one at a time under a single owner, and the renderer no longer writes `appearance.appIcon` on the generic settings channel at all. That write path was the thing that could interleave, so it is gone rather than fenced.

Selection can also be refused now: the generic channel would happily persist an id whose file is missing, while this seam checks the artwork is still there first.

The regression test issues a selection mid-removal without awaiting it, and pins the whole sequence rather than one boundary — the selection runs after the removal instead of inside it, is refused as `missing_artwork`, the setting ends on the brand mark, the file is gone, and nothing ran between the reset and the delete.

Also rebased onto current main; the conflict was the generated Astryx surface inventory, which #3369 had regenerated, so I re-ran the generator rather than hand-merging it.

Local verification on this head: full workspace typecheck, Biome lint and format, Knip, the Astryx inventory and theme checks, 1130 desktop main-process tests, 592 in `@maka/core`, the packaging-contract tests, and the Storybook build plus visual smoke — the 外观 story renders.

Astro-Han (approved):
Thanks — accepting a user-supplied image is the kind of feature where the interesting part is everything around the decoder, and that part is done carefully here.

**What this solves / how:** there was no way to change the app icon. This adds a preset catalog plus an import path: the main process reads the candidate with a size cap, parses the header to learn the dimensions *before* decoding, then decodes through Electron's `nativeImage`, crops to square and normalises to 1024.

**On the import path specifically**, since it's a new attack surface — checked each of these against the source rather than taking the structure on faith:

- **No new dependency.** Decoding goes through `electron.nativeImage`; the header parse is hand-rolled and small. Nothing new to keep patched.
- **Bounds are enforced before the decoder ever sees the bytes.** `readCapped` streams through a single handle and aborts past `CUSTOM_ICON_MAX_INPUT_BYTES`, so a 10GB "png" fails on read rather than in memory. `readImageHeader` then rejects on `CUSTOM_ICON_MAX_EDGE` before decode, so a file *declaring* 65535×65535 never reaches an allocation — that ordering is the part that matters, and it's right.
- **The stored filename contains nothing from the user.** The id is `randomUUID()` with dashes stripped, the path is `join(userData/app-icons, id.png)`, and reads validate against a strict hex pattern. The original filename never participates, so there's no traversal to reason about.
- **Preset weight is proportional**: 15 PNGs at ~27KB, ~400KB total, and `verify-packaged-app.mjs` gained a catalog assertion so a missing or extra preset fails packaging rather than shipping quietly.

**Cross-PR note:** this is the third PR today touching `verify-packaged-app.mjs`. The icon-catalog change sits around lines 699/735; #3473's packaged-Git recursion (now merged) is at 705/759+, and #3472's native-probe rewrite is at 334–640. No overlap, and the Git checks are untouched here.

**One thing to be aware of, not a finding:** the `Release Windows check` on this PR is red for reasons unrelated to it — see #3484, a regression from #3001 that breaks the Windows sandbox verifier on every open PR. The `CI` job on this head is green.

LGTM.

---
*AI-assisted review. The bounds and path handling above were each verified against the source at this head.*

ARE404:
Status on head `3fdfac7c1`: **`test` passes** (14m13s — first full green run on this branch). **`package` / Release Windows check fails**, and I believe that failure is not from this PR. Evidence rather than a claim:

```
  code: 'SANDBOX_FILESYSTEM_OPERATION_FAILED'   reason: 'path_changed'
  at verifyWindowsSandboxWorkerE2E (scripts/verify-windows-sandbox-e2e.mjs:97:5)
  at verifyPackagedWindowsApp   (scripts/verify-windows-x64.mjs:192:5)
```

- `git log -S "created while this call waited for the lock" --all` returns exactly one commit: `ec1579bc7` — "fix(runtime): close the path-replacement window for filesystem mutations (#2600) (#3001)". That guard and its message landed on `main` at 10:50; this run started at 11:18.
- This branch touches `verify-windows-x64.mjs` in one place — a `requireAppIconCatalog` flag on the resource assertions around line 110. The failure is at line 192, in the sandbox filesystem-worker E2E, which the run never reaches through anything I changed.
- Release Windows check has not run on `main` since #3001 merged, so this PR looks like the first execution of that workflow on top of it. Its previous `main` run (2026-08-21) also failed, though on a different step — an automatic-update version expectation (`0.1.12` vs `0.1.11`) — so I am not offering that as the same failure.

I have not tried to fix it here: it is upstream runtime behaviour, it belongs with #3001, and folding it into this PR would only make this branch harder to review. Happy to open a separate issue with the log if that is useful.

For completeness, local verification on this exact head: full workspace typecheck, Biome lint and format, Knip on both workspaces, the Astryx inventory and theme checks, 1130 desktop main-process tests, 592 in `@maka/core`, the packaging-contract tests, and a Storybook build plus visual smoke.

Windows and Linux runtime behaviour remains unverified by me — this machine cannot start an Electron GUI, so the picker, dock switching and import were exercised by hand on macOS only.

Astro-Han:
One request before this lands, and it's about the PR rather than the code.

This change is user-visible — it adds fifteen preset icons plus a custom-icon import path — but the PR body has no visuals. Could you add:

1. **Before / after** — the current fixed icon next to the new picker, so the change is legible without checking out the branch;
2. **A short visual walkthrough of the feature** — the picker itself, choosing a preset, and importing a custom file.

Screenshots are enough; a short clip is better for the import flow. The reason is practical: a UI change that only exists as a diff is reviewed as a diff, and reviewers end up guessing at the part that actually matters to users. It also gives whoever reads this PR in six months something to compare against.

The functional review is still in progress on our side and I'll post it separately — this one is independent, so you can start on it now.

apps/desktop/src/main/app-icon-ipc.ts
```diff
@@ -0,0 +1,213 @@
...
+      () => undefined,
+      () => undefined,
+    );
+    return run;
+  }
+
+  /**
+   * Selection lives behind this seam too, so it queues with removal instead of
+   * racing it through the generic settings channel. It is also the only place
+   * that can refuse a choice whose artwork is gone — the generic channel would
```
Astro-Han, line 108: [P3] This comment names a gap that the PR then leaves open. It says the seam "is also the only place that can refuse a choice whose artwork is gone — the generic channel would happily persist an id with nothing behind it." The generic channel is still reachable: `clientOwnedSettingsPatch` in `apps/desktop/src/shared/settings-ownership.ts` forwards `appearance` as one whole object, so a well-formed `custom:<32hex>` can be written through `settings:client:update` without queueing behind this owner — from a second window, an older preload, or a direct `settings.update({ appearance: { appIcon } })`.

The id shape is still validated, so the failure is not a malformed value; it is a valid id landing between `remove`'s settings apply and its file deletion, leaving the setting pointing at artwork that no longer exists. `normalizeSettings:700` documents that case and fails over to the brand mark, so the user-visible result is a graceful fallback rather than a broken tile — which is why this is P3 and not higher.

The reason it is still worth closing: the same function already filters `personalization` field by field rather than forwarding the section wholesale. Dropping `appearance.appIcon` there would make `appearance` consistent with its own neighbour and make this comment true as written, without adding another compare-and-set.

Astro-Han (commented):
Reviewed at head `2c5f96a5` (`test` and `package` both SUCCESS) by two independent lines. No P0–P2 survived verification; two P3s below, plus one request about the PR itself that I posted separately.

The core shape is right, and worth naming because it is the part that will keep this maintainable: `appearance.appIcon` stores a closed id rather than a path, and the main process is the only thing that turns an id into a file. The startup default-then-swap to avoid a first-frame icon flash is a nice touch. The serialized owner comment at `app-icon-ipc.ts:89` is unusually good — it explains why a compare-and-set around the settings write alone was rejected, which is exactly the reasoning that gets lost otherwise.

One thing I want to be explicit about, because a first pass on our side got it wrong: the shape gate is **not** bypassable. `mergeSettings` has exactly one production caller — `normalizeSettings` at `settings.ts:647` — and `normalizeSettings:705` unconditionally applies `isAppIconChoice(...) ?? 'default'`. Any claim that the generic settings channel can persist a malformed id is incorrect. What follows is about ownership and ordering only.

apps/desktop/src/main/app-icon-ipc.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  // The picker asks for the whole set at once; there is no per-id request, so
+  // no id from the renderer ever reaches the filesystem.
+  input.ipcMain.handle(
+    "app:iconPreviews",
+    (): Promise<readonly AppIconPreview[]> => input.listPreviews(),
+  );
+
+  // The dialog runs here and the file it returns is the only path this sees.
+  input.ipcMain.handle("app:importIcon", () =>
```
Astro-Han: [P3] The file dialog runs inside `serialize`, so the queue is held for as long as the dialog is open — which is user time, not work time. While one window has the picker up, `select` and `remove` from any other window are blocked behind it.

The dialog itself changes no state; only the copy-into-place and the subsequent select do. Awaiting `showOpenDialog` before entering `serialize`, and serializing just the write path, keeps the invariant this owner exists for while removing an unbounded hold. Minor, since the practical case is a single window.

Astro-Han:
Correction to my review above, since I stated it more strongly than the code supports.

I wrote that `mergeSettings` has exactly one production caller and that the shape gate is therefore unavoidable. That is wrong. `FileSettingsStore.update` and `updateIf` (`packages/storage/src/settings-store.ts:102` and `:120`) merge and write directly; `normalizeSettings` runs on the **read** side, in `readOrCreate`. So a malformed `appIcon` string can reach `settings.json` on disk — it just cannot survive being read back, and the OS surfaces resolve through `toAppIconChoice` regardless, so nothing downstream sees it.

The practical consequence is unchanged and both P3s stand as written: no path-traversal or malformed-value reaches an icon path, and the ownership point at `app-icon-ipc.ts:108` is still the one worth acting on. But "cannot be persisted" and "cannot be read back" are different claims, and I asserted the stronger one without checking `packages/storage`.

apps/desktop/src/main/app-icon-ipc.ts
ARE404: Fixed in d925da1e. You were right that the comment named a gap the PR then left open — and pointing at `personalization` right beside it was the useful part, because the shape was already there to copy.

`clientOwnedSettingsPatch` now filters `appearance` field by field and drops `appIcon`. A patch that carried nothing else drops the section entirely rather than forwarding an empty object, so it does not register as a write on that channel at all. Three tests pin it: the icon is stripped, an icon-only patch vanishes, and the rest of `appearance` still travels.

The comment on the seam is now true as written rather than aspirational.

apps/desktop/src/main/app-icon-ipc.ts
ARE404: Fixed in d925da1e. The dialog now runs before `serialize`, and only the copy-into-place is queued.

Your framing is the part I had wrong — user time versus work time. The queue exists to order state changes, and a dialog changes nothing until a file comes back, so holding the owner across it bought no invariant and cost every other window.

Covered by a test that leaves a dialog deliberately open and selects a different icon while it hangs: the selection completes rather than waiting.

Astro-Han (approved):
Hi @ARE404, Since it's a new feature and involve some new ui changes, could we have some picture in the PR body to introduce this function? This is a non blocking advice and other than this, LGTM.

ARE404:
Sure, here are some material.

before:
Snipaste_2026-08-23_14-28-06

after:
Snipaste_2026-08-23_14-41-42

a short screen recording showing customize icon:
https://github.com/user-attachments/assets/f198af94-c138-4606-8e95-93a3da48bb54

Astro-Han:
Nice Feature Addition! LGTM.

PR #3433, opened 2026-08-21
fix(desktop): skip idle Host restart prompt

Colafornia:
## Summary

When Maka restarts immediately after the Desktop client exits, an idle Runtime Host can still be within its idle grace period. The new client currently opens the restart prompt even though the Host has no active work.

This change restarts a generation-aware Host automatically when:

- `activeOperations === 0`
- `residencies.length === 0`

Hosts with active operations or residencies still use the existing restart prompt.

## Verification

- [x] `npm --workspace @maka/desktop run build:main`
- [x] `node --test apps/desktop/dist/main/__tests__/runtime-host-desktop-manager.test.js`
  - 24 tests passed
- [x] `git diff --check`
- [x] Manual Electron restart-cycle verification was run in this environment

## AI use

Tool(s) and scope:

Codex assisted with root-cause analysis, implementation, test updates, and verification.

## Checklist

- [x] Yes — idle Runtime Hosts restart without showing a prompt
- [ ] No

### DIFF
--- a/apps/desktop/src/main/runtime-host-desktop-manager.ts
+++ b/apps/desktop/src/main/runtime-host-desktop-manager.ts
@@ -582,7 +582,11 @@ class RuntimeHostDesktopManagerImpl implements RuntimeHostDesktopManager {
         return result.candidate;
       }
       if (result.kind === 'upgrade_required' && result.restartable) {
-        const decision = await this.#resolveRestartable(result);
+        const activity = result.handshake?.activity;
+        const decision =
+          activity && activity.activeOperations === 0 && activity.residencies.length === 0
+            ? 'restart'
+            : await this.#resolveRestartable(result);
         if (decision === 'cancel') {
           throw new RuntimeHostUpgradeCancelledError();
         }
--- a/apps/desktop/src/main/runtime-host-desktop-manager.ts
+++ b/apps/desktop/src/main/runtime-host-desktop-manager.ts
@@ -584,7 +584,10 @@ class RuntimeHostDesktopManagerImpl implements RuntimeHostDesktopManager {
       if (result.kind === 'upgrade_required' && result.restartable) {
         const activity = result.handshake?.activity;
         const decision =
-          activity && activity.activeOperations === 0 && activity.residencies.length === 0
+          activity &&
+          activity.connections === 0 &&
+          activity.activeOperations === 0 &&
+          activity.residencies.length === 0
             ? 'restart'
             : await this.#resolveRestartable(result);
         if (decision === 'cancel') {
--- a/apps/desktop/src/main/__tests__/runtime-host-desktop-manager.test.ts
+++ b/apps/desktop/src/main/__tests__/runtime-host-desktop-manager.test.ts
@@ -603,7 +603,7 @@ test('stops reconnecting when the replacement Host is incompatible', async () =>
   await owner.close();
 });
 
-test('restarts a generation-aware Host through its exact takeover handshake', async () => {
+test('restarts an idle generation-aware Host without prompting', async () => {
   const replacement = candidateHarness();
   const starts: DesktopRuntimeHostCandidateStartInput[] = [];
   const conflict = upgradeRequired(true);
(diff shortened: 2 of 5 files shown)


apps/desktop/src/main/runtime-host-desktop-manager.ts
```diff
@@ -582,7 +582,11 @@ class RuntimeHostDesktopManagerImpl implements RuntimeHostDesktopManager {
         return result.candidate;
       }
       if (result.kind === 'upgrade_required' && result.restartable) {
-        const decision = await this.#resolveRestartable(result);
+        const activity = result.handshake?.activity;
+        const decision =
+          activity && activity.activeOperations === 0 && activity.residencies.length === 0
```
Astro-Han: [P1] The first handshake is only a snapshot: after it reports idle, an operation, residency, or non-ready state can begin before the takeover handshake. This branch then chooses `restart`, while the Host-side takeover currently rechecks only accepted transports, so it can still drain and eventually terminate a Host that is no longer truly idle. Please make the Host the single authority here: during takeover, atomically re-evaluate the existing true-idle predicate (ready, no active operations, no residencies, and the required connection state) and return `upgrade_required` if it changed. A focused regression test where activity starts between discovery and takeover, plus one non-ready-state case, would cover the race without adding another client-side state machine.

Astro-Han (commented):
Thanks for keeping this change focused. The exact-head CI is green, but there is one Host-side lifecycle race that the current client snapshot cannot close. I left the final-state requirement inline so the fix can stay small and authoritative.\n\nAI-assisted review disclosure: Codex performed an independent exact-head review; I verified the lifecycle path, current source, and live PR state before posting this comment.

apps/desktop/src/main/runtime-host-desktop-manager.ts
```diff
@@ -582,7 +582,11 @@ class RuntimeHostDesktopManagerImpl implements RuntimeHostDesktopManager {
         return result.candidate;
       }
       if (result.kind === 'upgrade_required' && result.restartable) {
-        const decision = await this.#resolveRestartable(result);
+        const activity = result.handshake?.activity;
+        const decision =
```
Astro-Han, line 586: **Additional evidence for the existing P1 — the client's idle predicate is narrower than the Host's.**

`activity` carries a `connections` field, and this predicate does not read it. A Host with `activeOperations === 0`, `residencies.length === 0`, and `connections > 0` is treated as idle and restarted with no prompt. If any of those connections is mid-reconnect rather than settled, the restart happens silently underneath it.

This is not a separate finding — it falls inside the "required connection state" part of the P1 we already raised — but it is worth naming explicitly, because it is a *predicate* gap rather than a timing gap and would survive a fix that only made the check atomic.

**[P3] The test helper hard-codes `connections: 0`, which makes that gap invisible.**

`upgradeRequired` sets `connections: 0` unconditionally while `activeOperations` and `residencies` are both parameterized. So the three new cases (idle / `activeOperations > 0` / `residencies > 0`) can never exercise a non-zero connection count, and a fix that starts consulting `connections` would have no test to confirm it.

Two cases worth adding:
- `connections > 0` with zero operations and zero residencies — the direct regression anchor for the predicate gap above.
- `activity` undefined — the fallback path. The `activity &&` guard means this correctly falls back to prompting, which is the safe direction, but nothing pins it, and it is the behaviour you would most want to keep if the activity payload ever changes shape.

Astro-Han (commented):
Follow-up on our earlier review at this same head — we re-verified the TOCTOU finding rather than assuming it still held, and it does. Adding one piece of evidence and one test gap; no new blocking findings.

The re-verification: the idle decision is made from the discovery handshake snapshot, and the `takeoverHostEpoch` takeover handshake happens on the *next* `startCandidate`. We walked the path and there is no second activity read in between, so the window is real rather than theoretical.

Reviewed at exact head `5d583341`. **No new P0–P2. One P3, inline.**

---

Astro-Han (dismissed):
Reviewed on `5d5833416362d8e469bf27a26360e456b2259be5`. Small and self-contained, so a single line plus a gate check rather than the two-line treatment we use above ~100 lines.

The skip condition is one expression in `runtime-host-desktop-manager.ts`:

```ts
const activity = result.handshake?.activity;
const decision =
  activity && activity.activeOperations === 0 && activity.residencies.length === 0
    ? 'restart'
    : await this.#resolveRestartable(result);
```

The question worth asking here is whether a Host that is *stuck* can be mistaken for a Host that is *idle* — because if so, this trades "interrupts the user" for "restarts silently under them". Two things answer it:

- `activity` comes from `result.handshake.activity`, and this branch is only reachable on `upgrade_required && restartable`. Reaching it means the handshake completed, so an unresponsive Host never gets here at all — it fails earlier on a different path.
- The remaining case is a Host that answers the handshake but under-reports its own activity. That is a Host-side reporting bug, and it would be equally wrong on the prompt path; this branch has nothing better to consult.

The leading `activity &&` also points the right way: a missing activity block falls through to `#resolveRestartable` and prompts, rather than defaulting to a silent restart. The unsafe direction requires positive evidence of idleness.

Scope check: no new state, no new field, one conditional on facts the handshake already carried. Of the 51 added lines, roughly 40 are tests. Nothing here needs a second authority on what "idle" means.

CI is terminal green on this head (`test`: completed / success), which is the head this approval is bound to.

Approving.

apps/desktop/src/main/runtime-host-desktop-manager.ts
Colafornia: Fixed at the Host owner. Takeover now rechecks the existing #isTrueIdle() predicate, so readiness, accepted connections, active operations, and residencies are evaluated at admission instead of trusting the discovery snapshot. The client restartable classification now uses the same complete idle shape, including connections, and the desktop path keeps the missing-activity fallback as prompt. Added regressions for late residency, non-ready takeover, non-zero connections, and missing activity. Verified: Host kernel 50/50 and desktop manager 26/26.

apps/desktop/src/main/runtime-host-desktop-manager.ts
Colafornia: Addressed both follow-ups. The desktop restart decision now requires connections === 0 as well as zero active operations and residencies; missing activity still falls through to the prompt. Added focused tests for a non-zero connection count and an absent activity snapshot.

Astro-Han (approved):
Reviewed at `1ee5c1ba`. No P0–P3. Approving.

The question a change like this turns on is which way it errs, and it errs the safe way. `#isTrueIdle()` requires all four of `state === 'ready'`, zero accepted transports, zero active operations and zero residencies, and the desktop and client paths apply the same conjunction rather than a looser local version. The previous client-side test was `activity.connections === 0` alone; this tightens it. A false "busy" costs one extra prompt, while a false "idle" would restart a Host under load — so the asymmetry is pointed in the right direction.

The race that would defeat a check like this is covered too: *does not take over an idle-looking Host with a residency acquired after discovery* exercises a residency taken between discovery and the takeover decision.

`test` is terminal green on this exact head.

<details>
<summary>中文</summary>

在 `1ee5c1ba` 上审，无 P0–P3，approve。

这类改动的关键是"判错时偏哪边"，而它偏在安全的一侧：`#isTrueIdle()` 同时要求 `state === 'ready'`、accepted transports 为 0、active operations 为 0、residencies 为 0，桌面端与客户端走的是同一套合取条件，没有各写一份宽松版本。此前客户端只判 `activity.connections === 0`，这次收紧了。误判为"忙"只多弹一次提示，误判为"空闲"则会在有负载时重启 Host——不对称的方向是对的。

最能击穿这类检查的竞态也有覆盖：*does not take over an idle-looking Host with a residency acquired after discovery* 正是在 discovery 与接管决策之间取得 residency 的场景。

exact head 上 `test` 终态绿。
</details>

Astro-Han:
LGTM — merging. Nice fix: erring toward one extra prompt rather than restarting a busy Host is the right asymmetry, and covering the residency-acquired-after-discovery race is what makes the idle check trustworthy. Thanks!

<details>
<summary>中文</summary>

已确认，合并中。这个修法很好：宁可多弹一次提示也不误重启繁忙的 Host，方向是对的；并且覆盖了 discovery 之后才取得 residency 的竞态，这才让空闲判定可信。感谢贡献。
</details>

Issue #3142, opened 2026-08-17
Claude Code transcripts cannot be imported as Maka Sessions


Joob1n:
## Problem

`ExternalSessionAdapter` is source-agnostic and Codex uses it today, but there is no adapter for Claude Code. A user with history in `~/.claude/projects/<mangled-cwd>/<uuid>.jsonl` has no way to bring it into Maka.

Part of #2499. An earlier attempt is at #2954 (closed); this issue records the problem and what that attempt established, so the next one does not rediscover it.

## What the transcript format requires

- One model response is split across several records sharing `message.id`, and the records of one response can be separated by a `tool_result`. Folding only *consecutive* records imports one answer as several rows — 383 occurrences across 33 of 1236 local transcripts, 378 of them continuations carrying only `tool_use`.
- Tool results arrive as `type: 'user'` records. They are the harness replying, not the human, and must not import as user Turns.
- Sub-agent transcripts are whole `isSidechain` files, not interleaved records (verified: 0 mixed across 1231 files), so exclusion is per-file.
- The directory name mangles path separators, so only a record's own `cwd` can answer a project-scoped query.
- `isMeta` records are context Claude Code injected on the user's behalf and were never typed by a human.

## The load-bearing constraint

This is the part worth knowing before starting.

`isTrustworthyRecoveredTerminal` (`runtime-ledger-repair.ts`) discards a reconstructed terminal RuntimeEvent unless a **recorded** `turn_state` corroborates it. Every imported Run is then repaired to `failed`/`missing_terminal_event`, and the whole conversation renders as errors.

A transcript does not record that a Turn ended — only that another one began. So an importer that wants its history to render has to assert a `recorded` terminal it never observed. That assertion is wrong in a way that matters: 50 of 376 real root transcripts end with a prompt that has no answer after it (a killed process, a crash, or a Session still running), and each would import as a Turn that finished normally.

The honest shape is two changes that only work together:

1. Adapters emit `turn_state` only for facts the transcript carries — the interrupt notice and `isApiErrorMessage` — and let `deriveTurnRecords` label everything else `inferred`.
2. The Ledger accepts an inferred terminal on the transcript-materialization path, where `transcriptRunHeader` derives the header from the same messages, so there is no independent record for the transcript to contradict.

[...]

PR #3435, opened 2026-08-21
feat(storage): import Claude Code transcripts as Maka Sessions

Joob1n:
## Summary

`ExternalSessionAdapter` has been source-agnostic since #2500 with Codex as its only implementation. This adds Claude Code, so history in `~/.claude/projects/<encoded-cwd>/<uuid>.jsonl` can be imported.

**The terminal-state problem, and why it turned out to be answerable without touching the Ledger.**

#3142 records this as the hard part, and it is right about the constraint: the Ledger refuses a reconstructed terminal that no record corroborates — `isTrustworthyRecoveredTerminal` in `runtime-ledger-repair.ts`, and `terminalStatus` in `runtime-event-backfill.ts` independently. An importer either forges a `recorded` status it never observed, or watches every imported Run get repaired to `failed`. The issue proposed opening the Ledger first, as its own change.

That is not necessary, because the issue's premise does not survive the data. It states a transcript *"does not record that a Turn ended — only that another one began."* `stop_reason` is on every assistant record the Anthropic API produced, and `end_turn` is exactly the fact the Ledger asks for.

Measured across 1130 local root transcripts:

| `stop_reason` | count |
| --- | ---: |
| `tool_use` | 21697 |
| **`end_turn`** | **4026** |
| `stop_sequence` | 52 |
| `max_tokens` | 1 |

86.1% of turns carry a terminal stop reason. So this adapter reads recorded evidence the same way the Codex adapter reads `task_complete`, and **`runtime-ledger-repair.ts` and `runtime-event-backfill.ts` are untouched.**

The other 13.9% get no `turn_state` at all: 349 turns with no assistant reply, and 109 stopped at `tool_use` with the result never arriving. Those are genuinely unfinished — a killed process, a crash, a session still open. Reporting them as completed would put a false terminal state on one in seven imported turns, which is the outcome the Ledger's refusal exists to prevent.

## What the transcript shape forces

- **Tool results arrive as `type: 'user'` records** (9800 measured). They are the harness replying, not the human; importing them as user Turns would put the model's own tool output in the user's mouth.
- **Sub-agent transcripts are whole `isSidechain` files**, not interleaved records — 0 mixed files across 1130 — so exclusion is per file. Importing one would present a fragment of a conversation as a conversation.
- **The directory name cannot be reversed.** It encodes the cwd by replacing separators, so `-Users-a-b` is ambiguous between `/Users/a/b` and `/Users/a-b`. Only a record's own `cwd` can answer a project-scoped query.
- **Interrupt notices and `isApiErrorMessage`** are the two other terminal facts a transcript carries; they map to `aborted` and `failed`.

Parsing primitives come from `@maka/core/foreign-session` rather than being reimplemented — the issue asks for this, and the reason is real: the CLI's handoff scanner and this importer disagreeing about what the user actually said would be a defect, not a cosmetic inconsistency.

## Verification

Run against real data, not only fixtures. `listSessions()` returns **1128 sessions from 1185 local transcripts in 816ms** (the difference is 55 sidechain files and a few empty ones). Converting 300 of them:

```
user: 1127   assistant: 2605   turn_state: 1095
tool_call: 3972   tool_result: 3972
terminal state coverage: 97.2%
```

**`tool_call` and `tool_result` are exactly equal** — every call paired with its result. That is the strongest available signal that the record walk is correct, and it is the number that would move first if the tool_result-as-user-record handling were wrong.

11 tests cover the three terminal shapes separately (completed / no-terminal / aborted / failed), tool-result pairing, sidechain exclusion, cwd matching against the mangled directory name, corrupt-line tolerance, and registry wiring. One asserts every emitted message survives `decodeStoredMessage` — the importer round-trips adapter output through it, so a shape this adapter invented would otherwise be rejected at persistence with the adapter unnamed.

Gates: build, lint, format:check, typecheck, `check:release` (49), `check:stale`, knip, and the CI planner tests (53). Suites: storage 860 (849 before), desktop 1023, core 576.

Not run: Windows and Linux. This is filesystem and JSON only, but the paths are exercised on macOS here.

## A consequence this PR creates

The import catalog pages 16 sessions at a time and offers one filter (`includeArchived`). That was proportionate to what the only source produced; this PR makes it not.

| source | sessions on one machine |
| --- | ---: |
| Codex | 19 |
| Claude Code | **1128** |

59×. At 16 per page that is 71 pages of cursor paging, with no way to say which session you want. A Codex-only catalog never had enough rows for paging alone to fail — so this is something this PR introduces, not a gap it inherits.

I have filed and claimed it as #3438 rather than fold it in here. The fix belongs to `ExternalSessionQuery`, which both sources share, and its real decision — the match has to reach the adapter ahead of paging, or it silently searches only the rows already fetched — deserves its own review rather than riding along with transcript parsing.

## Review focus

Whether `stop_reason: 'end_turn'` is the right evidence for a `recorded` terminal state. My reading is that it is the model reporting its own turn ended — not an inference the importer made — which is what `isTrustworthyRecoveredTerminal` is asking for. If a reviewer reads that predicate as requiring a Maka-authored `turn_state` specifically, the conclusion changes and #3142's original two-step plan comes back.

## AI use

Select exactly one:

Tool(s) and scope: Claude Code — measured the transcript corpus, wrote the adapter and tests, and ran the verification above. `Generated-by` is on the commit. Reviewed and submitted by the contributor of record.

## Checklist

- [ ] No

### DIFF
--- a/packages/storage/src/claude-code-session-adapter.ts
+++ b/packages/storage/src/claude-code-session-adapter.ts
@@ -126,7 +126,12 @@ export class ClaudeCodeSessionAdapter implements ExternalSessionAdapter {
     } catch {
       return [];
     }
-    const files: Array<{ path: string; sessionId: string }> = [];
+    // Keyed by session id: the same id can legitimately exist under more than
+    // one project directory after a workspace move or a resumed session. Two
+    // files with one id are two candidates for the same source session, and
+    // list and read must pick the same one or a user selects one summary and
+    // imports the other.
+    const bySessionId = new Map<string, { path: string; sessionId: string; mtimeMs: number }>();
     for (const project of projects) {
       let entries: string[];
       try {
@@ -143,10 +148,26 @@ export class ClaudeCodeSessionAdapter implements ExternalSessionAdapter {
         // The id reaches a path join, so the resolved file must still be under
         // the projects root — a crafted id must not read outside it.
         if (!resolve(path).startsWith(resolve(root))) continue;
-        files.push({ path, sessionId });
+        let mtimeMs: number;
+        try {
+          mtimeMs = (await stat(path)).mtimeMs;
+        } catch {
+          continue;
+        }
+        const existing = bySessionId.get(sessionId);
+        // Newest wins, and the path breaks a tie so the choice does not depend
+        // on directory iteration order. A resumed session's continuation is
+        // the copy a user means when they pick that id.
+        if (
+          !existing ||
+          mtimeMs > existing.mtimeMs ||
+          (mtimeMs === existing.mtimeMs && path < existing.path)
+        ) {
+          bySessionId.set(sessionId, { path, sessionId, mtimeMs });
+        }
       }
     }
-    return files;
+    return [...bySessionId.values()].map(({ path, sessionId }) => ({ path, sessionId }));
   }
 
   async #parse(path: string, sessionId: string): Promise<ParsedTranscript | undefined> {
@@ -275,6 +296,10 @@ export default ClaudeCodeSessionAdapter;
  *  state, the same as any other turn whose end nothing vouches for. */
 const TERMINAL_STOP_REASONS = new Set(['end_turn', 'stop_sequence']);
 
+/** Names the cutoff for a turn the transcript simply stops inside, so a reader
+ *  can tell an imported snapshot's edge from a user's Stop or a provider abort. */
+const EXTERNAL_SNAPSHOT_ABORT_SOURCE = 'external_session_snapshot';
+
 interface TurnAccumulator {
   turnId: string;
   lastTs: number;
@@ -286,10 +311,50 @@ interface TurnAccumulator {
   aborted?: boolean;
 }
 
+/**
+ * One semantic event per source record, before any of it becomes a message.
+ *
+ * A transcript is an append log, not a list of distinct events. Two shapes
+ * break a naive one-line-one-message walk, and both occur in ordinary resume
+ * and recovery histories:
+ *
+ * - **A record can be written twice.** Replaying it emits the prompt, turn, or
+ *   tool identity twice. Measured: 3 repeats across 1130 local transcripts —
+ *   rare, and unrecoverable once persisted as canonical history.
+ * - **One assistant response is split across records sharing `message.id`,
+ *   and the pieces can be separated by the `tool_result` of a call the earlier
+ *   piece made.** Measured: 374 occurrences across 30 transcripts. Emitting
+ *   each piece as its own assistant message turned one reply into four.
+ *
+ * Folding must not flatten the second shape into its first record's position:
+ * a later `tool_use` sharing that id was issued *after* the result in between,
+ * and moving it earlier would invert cause and effect. So the visible prose of
+ * a response is emitted once, at its first appearance, while each tool call
+ * stays where the log put it. Verified against the corpus: `text` and
+ * `thinking` each appear at most once per `message.id`, so emitting them once
+ * loses nothing.
+ */
+function normalizeRecords(records: readonly TranscriptRecord[]): readonly TranscriptRecord[] {
+  const seenUuids = new Set<string>();
+  const normalized: TranscriptRecord[] = [];
+  for (const record of records) {
+    const uuid = typeof record.uuid === 'string' ? record.uuid : undefined;
+    if (uuid) {
+      if (seenUuids.has(uuid)) continue;
+      seenUuids.add(uuid);
+    }
+    normalized.push(record);
+  }
+  return normalized;
+}
+
 export function convertTranscript(
   sessionId: string,
-  records: readonly TranscriptRecord[],
+  rawRecords: readonly TranscriptRecord[],
 ): readonly StoredMessage[] {
+  const records = normalizeRecords(rawRecords);
+  // Prose already emitted for an assistant response, keyed by `message.id`.
+  const emittedProse = new Set<string>();
   const messages: StoredMessage[] = [];
   let turn: TurnAccumulator | undefined;
   let sequence = 0;
@@ -302,12 +367,23 @@ export function convertTranscript(
 
   const closeTurn = (): void => {
     if (!turn) return;
-    // No terminal `turn_state` is emitted for a turn nothing corroborates.
-    // Measured across 1130 local transcripts, 13.9% of turns end this way —
-    // no assistant reply at all, or stopped at `tool_use` with the result
-    // never arriving. Those are genuinely unfinished: a killed process, a
-    // crash, or a session still open. Emitting `completed` for them would put
-    // a false terminal state on 1 in 7 imported turns.
+    // Every turn gets a terminal state, and which one depends on what the
+    // transcript actually says.
+    //
+    // Leaving one out is not the same as preserving "unfinished". Without a
+    // `turn_state`, `deriveTurnRecords` falls back to `inferLegacyTurnStatus`,
+    // which answers `completed` for any turn holding an assistant message
+    // (`session.ts:1250`) and marks it `inferred`. The Ledger then refuses
+    // that uncorroborated terminal and the repair path persists
+    // `failed / missing_terminal_event` — an internal-corruption verdict on a
+    // transcript that was merely cut short. Measured: 13.9% of turns across
+    // 1130 local transcripts end with no assistant reply or at a `tool_use`
+    // whose result never arrived.
+    //
+    // So an unfinished turn is recorded as what it is: a snapshot that ended
+    // mid-turn, with an `abortSource` naming the import rather than a user or
+    // a provider. `end_turn`, interrupt notices and API errors keep their own
+    // evidence and are unaffected.
     if (turn.aborted) {
       messages.push({
         type: 'turn_state',
@@ -336,6 +412,17 @@ export function convertTranscript(
         turnId: turn.turnId,
         ts: turn.lastTs,
         status: 'completed',
+        partialOutputRetained: true,
+      });
+    } else {
+      messages.push({
+        type: 'turn_state',
+        id: id('turn-state'),
+        turnId: turn.turnId,
+        ts: turn.lastTs,
+        status: 'aborted',
+        abortedAt: turn.lastTs,
+        abortSource: EXTERNAL_SNAPSHOT_ABORT_SOURCE,
         partialOutputRetained: true,
       });
     }
@@ -413,7 +500,13 @@ export function convertTranscript(
       const stop = stringOf(message?.stop_reason);
       if (stop && TERMINAL_STOP_REASONS.has(stop)) turn.terminalStop = stop;
 
-      const thinking = thinkingText(message);
+      // One response's prose belongs to the response, not to each record that
+      // carries part of it. A fragment with no id is its own response.
+      const responseId = stringOf(message?.id);
+      const proseAlreadyEmitted = responseId !== undefined && emittedProse.has(responseId);
+      if (responseId !== undefined) emittedProse.add(responseId);
+
+      const thinking = proseAlreadyEmitted ? '' : thinkingText(message);
       if (thinking) {
         messages.push({
           type: 'assistant',
@@ -426,7 +519,7 @@ export function convertTranscript(
           modelId,
         });
       }
-      const text = claudeAssistantText(record);
+      const text = proseAlreadyEmitted ? undefined : claudeAssistantText(record);
       if (text) {
         messages.push({
           type: 'assistant',
--- a/packages/storage/src/claude-code-session-adapter.ts
+++ b/packages/storage/src/claude-code-session-adapter.ts
@@ -28,6 +28,10 @@ import type {
   ExternalSessionSummary,
 } from '@maka/core/external-session';
 import type { StoredMessage } from '@maka/core/session';
+import {
+  resolveTranscriptLineage,
+  type TranscriptRecord,
+} from './claude-code-transcript-lineage.js';
 
 export const CLAUDE_CODE_SESSION_ADAPTER_ID = 'claude-code';
 
@@ -46,8 +50,6 @@ export interface ClaudeCodeSessionAdapterOptions {
   claudeHome?: string;
   maxTranscriptBytes?: number;
 }
-
-type TranscriptRecord = Record<string, unknown>;
 
 interface ParsedTranscript {
   readonly records: readonly TranscriptRecord[];
@@ -312,50 +314,35 @@ interface TurnAccumulator {
 }
 
 /**
- * One semantic event per source record, before any of it becomes a message.
- *
- * A transcript is an append log, not a list of distinct events. Two shapes
- * break a naive one-line-one-message walk, and both occur in ordinary resume
- * and recovery histories:
- *
- * - **A record can be written twice.** Replaying it emits the prompt, turn, or
- *   tool identity twice. Measured: 3 repeats across 1130 local transcripts —
- *   rare, and unrecoverable once persisted as canonical history.
- * - **One assistant response is split across records sharing `message.id`,
- *   and the pieces can be separated by the `tool_result` of a call the earlier
- *   piece made.** Measured: 374 occurrences across 30 transcripts. Emitting
- *   each piece as its own assistant message turned one reply into four.
- *
- * Folding must not flatten the second shape into its first record's position:
- * a later `tool_use` sharing that id was issued *after* the result in between,
- * and moving it earlier would invert cause and effect. So the visible prose of
- * a response is emitted once, at its first appearance, while each tool call
- * stays where the log put it. Verified against the corpus: `text` and
- * `thinking` each appear at most once per `message.id`, so emitting them once
- * loses nothing.
+ * Which records are the conversation, decided before any of them becomes a
+ * message. See `claude-code-transcript-lineage` for what the transcript's own
+ * fields say about rewind branches, compaction boundaries, and fragmented
+ * responses; this file only converts what that returns.
  */
-function normalizeRecords(records: readonly TranscriptRecord[]): readonly TranscriptRecord[] {
-  const seenUuids = new Set<string>();
-  const normalized: TranscriptRecord[] = [];
-  for (const record of records) {
-    const uuid = typeof record.uuid === 'string' ? record.uuid : undefined;
-    if (uuid) {
-      if (seenUuids.has(uuid)) continue;
-      seenUuids.add(uuid);
-    }
-    normalized.push(record);
-  }
-  return normalized;
-}
-
 export function convertTranscript(
   sessionId: string,
   rawRecords: readonly TranscriptRecord[],
 ): readonly StoredMessage[] {
-  const records = normalizeRecords(rawRecords);
-  // Prose already emitted for an assistant response, keyed by `message.id`.
-  const emittedProse = new Set<string>();
+  const records = resolveTranscriptLineage(rawRecords).records;
+  // Every fragment of one assistant response, keyed by `message.id`. A
+  // response is emitted once, from all of its fragments, at the position of
+  // the first — so a later fragment's text is part of the reply rather than
+  // something the first fragment's absence of text can suppress.
+  const responseFragments = new Map<string, TranscriptRecord[]>();
+  for (const record of records) {
+    if (record.type !== 'assistant') continue;
+    const responseId = stringOf(asMessageRecord(record)?.id);
+    if (responseId === undefined) continue;
+    const existing = responseFragments.get(responseId);
+    if (existing) existing.push(record);
+    else responseFragments.set(responseId, [record]);
+  }
+  const emittedResponses = new Set<string>();
   const messages: StoredMessage[] = [];
+  // A boundary can precede the first turn — a transcript that opens straight
+  // after a compaction. A system note needs a turn to hang from, so the fact
+  // waits for one rather than being dropped for arriving early.
+  let pendingCompactBoundaryTs: number | undefined;
   let turn: TurnAccumulator | undefined;
   let sequence = 0;
   const id = (kind: string): string => `claude-code:${sessionId}:${kind}:${sequence++}`;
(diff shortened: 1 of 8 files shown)


packages/storage/src/claude-code-session-adapter.ts
```diff
@@ -0,0 +1,529 @@
...
+        turnId: turn.turnId,
+        ts: turn.lastTs,
+        status: 'completed',
+        partialOutputRetained: true,
+      });
+    }
+    turn = undefined;
+  };
+
+  for (const record of records) {
```
Astro-Han, line 419: [P1] Normalize Claude source identity before emitting canonical messages

This loop currently treats every JSONL line as a distinct semantic event. In normal Claude resume and recovery histories, the same `record.uuid` can be written more than once, while fragments sharing one assistant `message.id` can recur non-consecutively across a `tool_result`. Real transcripts therefore cause this walk to emit duplicate user prompts, turns, and identical `tool_call` ids.

Once those duplicates are persisted as canonical Maka history, the imported Session can be misleading or invalid for continuation. Deleting and importing it again on this head produces the same result, so this is not recoverable without changing the converter.

Could we make source normalization a single precondition of conversion: de-duplicate repeated records by `record.uuid`, fold assistant fragments by `message.id` even when a tool result appears between them, and emit each semantic block exactly once in its original causal order? A focused regression should contain both a repeated UUID and a non-consecutive shared `message.id`. Equal call/result counts are not sufficient here because both sides can be duplicated together.

This is the only issue in my review that should be addressed before approval.

packages/storage/src/claude-code-session-adapter.ts
```diff
@@ -0,0 +1,529 @@
...
+        ...(parsed.updatedAt !== undefined ? { updatedAt: parsed.updatedAt } : {}),
+      });
+    }
+    summaries.sort((left, right) => (right.updatedAt ?? 0) - (left.updatedAt ?? 0));
+    return summaries;
+  }
+
+  async readSession(sessionId: string): Promise<ExternalMakaSession> {
+    assertSafeSessionId(sessionId);
+    const file = (await this.#transcriptFiles()).find(
```
Astro-Han, line 101: [P2] Resolve duplicate transcript stems consistently between list and read

A Claude Session UUID can legitimately appear under more than one project directory after a workspace move, resume, or recovery. `listSessions()` currently exposes each file with the same source id, while this `.find()` reads whichever copy happens to appear first. The user can therefore select a newer summary but import an older or shorter transcript carrying the same UUID.

The clean final state is one resolver shared by both listing and reading. It should either select one canonical valid transcript per UUID deterministically — normally the newest continuation — or, if separate forks are intentionally user-visible, give each one a path-qualified opaque source id. It should also verify that the records' `sessionId` agrees with the filename stem.

A fixture with the same UUID in two project directories and divergent contents would pin the chosen policy. I consider this non-blocking because the source remains intact and the affected import can be deleted, but the current selection is not deterministic from the user's perspective.

packages/storage/src/claude-code-session-adapter.ts
```diff
@@ -0,0 +1,529 @@
...
+  const id = (kind: string): string => `claude-code:${sessionId}:${kind}:${sequence++}`;
+  // Turn ids count separately from message ids. Sharing one counter made turn
+  // ids skip (`turn:0`, `turn:3`) for no reason, and left them one edit away
+  // from colliding with a message id if the emission order ever changed.
+  let turnSequence = 0;
+  const nextTurnId = (): string => `claude-code:${sessionId}:turn:${turnSequence++}`;
+
+  const closeTurn = (): void => {
+    if (!turn) return;
+    // No terminal `turn_state` is emitted for a turn nothing corroborates.
```
Astro-Han: [P2] Give an incomplete imported snapshot a Host-owned terminal outcome

Leaving out `turn_state` does not preserve an unfinished turn in the final system. During transcript materialization, `deriveTurnRecords()` infers a completed turn, the Ledger rejects the recovered terminal because no recorded state corroborates it, and the repair path persists `failed / missing_terminal_event`.

That is reachable for ordinary unanswered prompts, a final `tool_use`, and `max_tokens`. The damage is limited to the imported copy, but the resulting internal-corruption failure is not the source fact this adapter intended to represent.

Could the final state define one explicit imported-snapshot cutoff at the Host transcript-materialization seam — for example a cancelled/aborted terminal with an `external_session_snapshot` source — while keeping `end_turn`, interrupt, and API-error evidence authoritative? An end-to-end test through `HostExternalSessionCoordinator` and Ledger materialization is important here; the adapter-only assertion cannot observe the repair that changes the status.

This is non-blocking in my review, but it should not be described as “no terminal state” until the Host result matches that claim.

Astro-Han (commented):
Thanks for adding Claude Code through the existing `ExternalSessionAdapter` and Host import seams. The overall architecture is the right one: source-specific parsing stays in the adapter, canonical persistence and lifecycle stay with the Host, and the Desktop change is limited to presenting the second source. The recorded `end_turn` evidence, synthetic-user filtering, and tool-result handling are also well motivated.

I found one issue that should be addressed before approval: Claude JSONL records need an identity-normalization pass before canonical messages are emitted. Normal resume/recovery transcripts can repeat the same source record or split one assistant response across non-consecutive records. Persisting those lines independently duplicates prompts, turns, and tool identities, which can make the imported history unreliable for continuation.

The two P2 comments are non-blocking follow-ups: make duplicate transcript-file resolution deterministic, and align unfinished source turns with the Host's actual terminal materialization. I intentionally left timestamp polish, catalog scan cost, and path-normalization improvements out of the blocking scope.

A clean final shape would be:

1. normalize each Claude source identity once;
2. emit one causally ordered canonical representation;
3. resolve one deterministic transcript for list and read;
4. let the Host own the final imported-snapshot terminal fact.

Once the duplicate-source-identity issue is fixed with the focused repeated-record fixture, I would be comfortable approving this revision.

AI-assisted review disclosure: OpenAI Codex assisted with the exact-head analysis of transcript identity, Host Ledger materialization, source resolution, and simplification opportunities. I verified the cited code paths, live PR state, and severity calibration before posting.

Joob1n:
All three addressed in `8316b773c`.

**P1 — you were right, and my first check was not.**

I verified the claim by counting duplicate `tool_call` ids in the output, found zero, and nearly replied that it did not reproduce. That check could not have found anything: the converter mints a fresh id per record, so duplicates are impossible by construction. Your comment had already ruled it out — *"equal call/result counts are not sufficient here"* — and I ran it anyway.

Measuring the right thing showed it plainly. Across the 30 affected transcripts on this machine: **1683 more assistant messages than there were responses.** One reply rendering as four.

Both shapes are now normalized before conversion:

- Records de-duplicate on `record.uuid` — 3 repeats across 1130 local transcripts.
- One response's prose is emitted once, keyed by `message.id` — 374 split occurrences across 30 transcripts.

Folding does **not** relocate later fragments to the first one's position. A `tool_use` arriving after a `tool_result` was issued after it, so moving it earlier would invert cause and effect: prose folds, calls stay where the log put them. I checked the corpus before choosing that rule — `text` and `thinking` each appear at most once per `message.id`, so emitting them once loses nothing.

After the fix, 1683 extra messages become 280, and those 280 are genuinely separate responses: different `message.id`, different text, one per model reply after a tool returned. Regression covers both a repeated UUID and a non-consecutive shared `message.id`, and I verified each fails when its own normalization is reverted.

**P2 — deterministic transcript resolution.** `listSessions` exposed every file while `readSession` took whichever `.find()` reached first. Both now resolve through one map keyed by session id: newest mtime wins, path breaks the tie, so the choice no longer depends on directory iteration order.

**P2 — "no terminal state" was not what the system did with it, and you are right that I should not have described it that way.**

I traced it: without a `turn_state`, `deriveTurnRecords` falls back to `inferLegacyTurnStatus`, which returns `completed` for any turn holding an assistant message (`session.ts:1250`) and marks it `inferred`; the Ledger refuses that uncorroborated terminal and the repair path persists `failed / missing_terminal_event`. A transcript that was merely cut short would have imported as internal corruption — the opposite of the source fact I meant to preserve.

Unfinished turns now carry an explicit cutoff: `aborted` with `abortSource: 'external_session_snapshot'`, distinguishable from a user Stop or a provider abort. `end_turn`, interrupt notices and API errors keep their own evidence and are unchanged. Terminal coverage across 1128 local sessions goes 97.2% → 100%, with the 32 newly covered turns landing in the snapshot bucket rather than in the repair path.

**On the end-to-end test you asked for.** I first wrote that I had not added one and would if you wanted it — which dressed up a decision I had made without checking. `runtime-ledger-repair.test.ts` already builds the stores, calls `materializeTranscriptLedger`, and asserts the resulting run; the harness was there and the cost I declined to pay was imagined. It is in `82287ab56`, as a pair:

- a turn carrying `aborted` / `external_session_snapshot` materializes to a `cancelled` run — the Ledger accepts a recorded terminal;
- the same turn with no terminal state materializes to `failed / missing_terminal_event`.

The second pins what the adapter used to produce, so the reason the first exists cannot quietly stop being true. Neither is observable from the adapter, which is your point about the seam.

Tool calls and results remain exactly paired at 3972 each. Gates: build, lint, format:check, typecheck, `check:release`; storage 865, desktop 1030, ui 187, core 587, runtime 2996/3001 (the five failures are `spawn rg ENOENT` on a machine with no ripgrep binary).

packages/storage/src/claude-code-session-adapter.ts
Joob1n: Fixed in `8316b773c`.

My first check was the wrong one — I counted duplicate `tool_call` ids, found zero, and nearly replied that it did not reproduce. The converter mints a fresh id per record, so that count could never have found anything; your comment had already ruled it out and I ran it anyway. Measuring responses instead: **1683 more assistant messages than responses** across the 30 affected transcripts here.

Records now de-duplicate on `record.uuid`, and a response's prose is emitted once per `message.id`. Folding does not relocate later fragments — a `tool_use` after a `tool_result` was issued after it — so prose folds and calls stay in place. Checked the corpus first: `text` and `thinking` each appear at most once per `message.id`.

1683 extra messages became 280, and those 280 are genuinely separate responses. Regression covers both shapes, each verified to fail when its own normalization is reverted.

packages/storage/src/claude-code-session-adapter.ts
Joob1n: Fixed in `8316b773c`. `listSessions` and `readSession` now resolve through one map keyed by session id — newest mtime wins, path breaks the tie — so the selection no longer depends on directory iteration order.

I did not add the two-directory fixture: no duplicate stems exist in the 1185 local transcripts, so I would have been pinning a policy against data I could not observe. The resolver is deterministic either way; say the word if you want the fixture regardless.

packages/storage/src/claude-code-session-adapter.ts
Joob1n: Fixed in `8316b773c`, and you were right that I should not have described it as "no terminal state".

Traced it: without a `turn_state`, `deriveTurnRecords` falls back to `inferLegacyTurnStatus`, which returns `completed` for any turn holding an assistant message (`session.ts:1250`) and marks it `inferred`; the Ledger refuses it and the repair writes `failed / missing_terminal_event`. A transcript that was merely cut short imported as internal corruption.

Unfinished turns now carry `aborted` with `abortSource: 'external_session_snapshot'`. Terminal coverage across 1128 local sessions goes 97.2% → 100%.

The end-to-end test is in `82287ab56` — I had said I would add it if wanted, which dressed up a decision made without checking; `runtime-ledger-repair.test.ts` already had the harness. Both directions are pinned: the cutoff materializes to `cancelled`, and the same turn without it materializes to `failed / missing_terminal_event`.

Astro-Han (commented):
Thanks for the careful follow-up, the corpus measurements, and the Host-level regression coverage. The duplicate-stem resolver is now deterministic across list/read, and the imported-snapshot terminal now materializes correctly through the Ledger. I consider both previous P2 concerns closed.

I also need to correct my own wording in the original P1. Asking to “fold assistant fragments by `message.id`” and “emit each semantic block once” was not precise enough, and it reasonably pointed toward the `emittedProse` implementation on this head. That framing was mine, and I do not want to send you through another round of response-level guards.

**[P1] Resolve the active Claude lineage before converting it to canonical Maka history**

On the current head, `emittedProse` can drop real output when the first record for a `message.id` contains thinking, a tool call, or no text, and a later fragment carries the text. More importantly, that loss is a symptom of the deeper source-model mismatch: Claude’s JSONL is a UUID/`parentUuid` lineage graph, not one flat causal stream. Resume, rewind, compaction, and parallel tool completion are normal paths. Walking every physical record can therefore import an abandoned branch, place parallel calls/results in append order rather than logical order, or suppress later content from the selected response.

Once persisted, that changes the canonical Session and its continuation replay; the user must delete and re-import after the converter is fixed. That is why this remains the one blocking issue rather than another request for a small `if`.

To make the final state explicit and avoid further patch-by-patch review, I think the converter should have one version-aware source-resolution stage before StoredMessage emission:

1. Resolve UUID updates and select the active `parentUuid` lineage, with an explicit version-gated fallback only for a verified legacy unchained format.
2. Apply one compaction policy: preserve a faithfully representable active checkpoint, or fail closed for an unsupported compacted transcript. Do not both flatten pre-boundary history and discard the generated summary.
3. Recover off-lineage assistant fragments only when they belong to a selected `message.id`, together with their referenced tool-result children. Parallel calls should remain one assistant step and precede their completion-ordered results.
4. Convert that resolved logical sequence into causal assistant segments, preserving every thinking/text delta and linking tool calls to the assistant step. Please do not replace `emittedProse` with `emittedText` or `emittedThinking`; those booleans would encode another unverified source contract.
5. Validate the result through Host Ledger materialization and model replay, checking that no selected text is lost, no abandoned branch is imported, and every tool result remains paired with the correct call.

A compact fixture set should be enough:

- an active leaf beside an abandoned rewind sibling;
- physical `call A → result A → call B fragment → result B`, resolving to one parallel assistant step with calls before results;
- thinking/tool content followed by later text under the same `message.id`;
- a compact boundary with preserved context and current continuation, or an asserted unsupported-version rejection;
- one end-to-end Host replay assertion over the resolved history.

This does not require copying or depending on Claude Code’s private loader implementation. The public evidence in the records is sufficient to express a small resolver; where the format does not expose a trustworthy contract, a version gate or fail-closed result is preferable to a heuristic.

This is the only remaining blocking requirement in my review. The duplicate-stem and snapshot-terminal fixes are accepted.

AI-assisted review disclosure: Codex assisted with the source-model analysis and structural corpus checks. I verified the exact PR head, the current Claude Code session-loading behavior, the relevant Maka authority boundaries, and the conclusions above.

Joob1n:
Addressed in `82e9ee8c3`. You were right on both counts, and measuring them turned up numbers larger than either of us assumed — plus one shape neither of us had.

## The `emittedProse` loss is real, and it is large

**4094 assistant text blocks dropped across 392 of 1130 local transcripts.** In 4093 of 14095 responses the visible text is not in the first fragment, so a guard that emits prose at the first record and suppresses it afterwards throws the reply away.

My earlier verification said "`text` and `thinking` each appear at most once per `message.id`, so emitting them once loses nothing." That was **true and measured the wrong thing**: the guard keys on the first *fragment*, not the first fragment *with text*. Same class of mistake as the P1 tool-call count. You called it before I found it.

A response is now assembled from all of its fragments at the position of the first — thinking and text joined in fragment order, every call before any result. No `emittedText` / `emittedThinking`: nothing is suppressed, so nothing needs a flag.

On the parallel-call ordering you asked for: calls sharing a `message.id` came from one API response, so a call the log wrote after a sibling's result was not issued after it. One assistant step, calls first, results in completion order.

## Where I have to push back: the lineage walk is the wrong primitive

I built the resolver you described first — active leaf, walk `parentUuid` back, recover off-lineage fragments. Then I classified every fork in the corpus, and it does not hold up.

**Of 358 forked parents, 281 are one ordinary shape:** an assistant record carrying a `tool_use` has two children — the *next fragment of that same response*, and the *`tool_result` of the call it just made*. Nothing was abandoned. A response and a completion share a parent.

Selecting the active lineage by walking back from the newest record calls all of that dead. Measured on this corpus, such a walk **strands 6535 tool results and 18815 other records**. That walk is in this branch's history; it was wrong, and the fragment-recovery step you proposed only patches a corner of it.

**The one real ambiguity is 72 parents with two or more *user prompt* children** — a rewind, where the user edited a prompt and resubmitted. The evidence is unmistakable once you look at the pairs:

```
StreamVByte 是什么类型？我忘了            ← withdrawn
StreamVByte 是什么方式？我忘了            ← asked

先加载个skill，~/Downloads/SKILL.md      ← withdrawn
先加载个skill，~/Downloads/SKILL.md，放到我本地的skill目录下   ← asked
```

So resolution is exactly that narrow: among sibling prompts the last written wins and takes the others' subtrees with it — 261 records across 21 transcripts. A tool result is the harness answering the model, not the user asking again, so it is never a prompt sibling. Everything else is kept, because nothing in the records says it was abandoned.

I think this satisfies what you were protecting — no abandoned branch is imported — without the collateral the walk causes. If you still want the leaf-walk, say so and I will implement it, but I would want the 6535 stranded results on the record first.

## Compaction was being lost twice, in both directions

You wrote "do not both flatten pre-boundary history and discard the generated summary." That was exactly the state:

- the `context_compacted` note was keyed on `isCompactSummary`, which belongs to the summary **user** record — consumed by the `user` branch and never reaching that code. The branch was unreachable: **0 notes emitted, ever**;
- meanwhile `claudeUserAuthoredText` drops `isCompactSummary` records, so the summary was not imported either.

Both halves of the event, gone, with the pre-boundary history flattened in beside it. Now keyed on the `compact_boundary` record that states it, and held for the next turn when it arrives before one: **11 of 11 boundaries survive**.

Pre-boundary records stay. Both sides are conversation that happened, and keeping only the newest root would discard 24,695 records here.

**One correction worth its own line:** `logicalParentUuid` on the boundary looks like the backward link that would make a compacted file one chain. It is not — it equals `compactMetadata.preservedSegment.tailUuid`, a record written *after* the boundary. I followed it as a parent, which closed a cycle through the summary and stranded **2319 of 3392 records** on one local transcript. It took tracing the walk to see; the field name invites the mistake. A version gate would not have caught this, because the field is present and well-formed — it just does not mean what it looks like.

## Fixtures

Your list, plus what the classification turned up:

| fixture | what it pins |
| --- | --- |
| withdrawn prompt + its answer vs the edit | the rewind case, in the shape the corpus writes it |
| whole withdrawn subtree, three-way sibling | the drop is a subtree, and survives two edits |
| **ordinary tool fork survives whole** | the 281 case — a walk that picks one child strands the other |
| two tool results under one parent | a result is not a prompt, so it is not a rewind |
| thinking/tool fragments then text last | the 4093 case |
| `call A → result A → call B` | one step, calls before results |
| compact boundary, both sides kept | neither side discarded, boundary marked |
| `logicalParentUuid` not followed | the cycle |
| duplicate uuid, unplaceable record, cycle, empty | the edges |

15 lineage tests, plus the adapter suite.

**End to end**, in `runtime-ledger-repair.test.ts`: raw records → `readSession` through the adapter registry → Ledger materialization → the replay plan a continuation is given. It asserts no selected text is lost, that the withdrawn prompt and its answer are absent, and that every result pairs with its call. It fails if either half of the fix is reverted — I checked by reverting each.

## Verification

- `@maka/storage`: **880/880**
- `@maka/runtime`: 2876/2881 — the 5 failures are `spawn rg ENOENT` on a machine with no ripgrep binary, unrelated
- `biome lint packages/storage/src packages/runtime/src`: 707 files, clean
- corpus, 1130 transcripts, before → after:

| | before | after |
| --- | ---: | ---: |
| assistant text blocks dropped | 4094 | **0** |
| assistant text preserved | 974,915 ch | **2,579,574 ch** |
| `tool_call` / `tool_result` | 11970 / 11970 | **11993 / 11993** |
| calls with no result | — | **0**, matching source exactly |
| compaction boundaries marked | 0 of 11 | **11 of 11** |

**The hosted `test` check has not run on this head.** The run for `82e9ee8c3` is sitting at `action_required` (`32540108801`, 0s) — as is the one on #3443. @Astro-Han, when you next look at this, could you release both workflow runs? Earlier pushes on this branch did get their runs, so it seems to be per-run rather than a standing gate, and I cannot approve my own.

packages/storage/src/claude-code-transcript-lineage.ts
```diff
@@ -0,0 +1,189 @@
...
+ * model — so a response's completions are not prompts, and two of them
+ * sharing a parent is not a rewind. That distinction is the whole of the
+ * difference between the 281 ordinary forks and the 72 real ones.
+ */
+export function isPromptRecord(record: TranscriptRecord): boolean {
+  if (record.type !== 'user') return false;
+  const message = record.message;
+  if (typeof message !== 'object' || message === null) return true;
+  const content = (message as Record<string, unknown>).content;
+  if (!Array.isArray(content)) return true;
```
Astro-Han, line 90: [P2] Use the converter's human-prompt predicate for rewind detection

Reachability: ① normal Claude transcript import. This currently treats every non-`tool_result` user record as a prompt, but conversion deliberately excludes `isMeta`, compact summaries, interrupt notices, and command/local-command/bash wrappers.

I reproduced three mixed sibling groups in the current local corpus. One has a real user prompt followed by three `[Request interrupted by user]` records under the same parent. The resolver keeps the last interrupt as the winning “prompt” and drops the real prompt plus its answer subtree; conversion then skips the interrupt as synthetic, so the imported Session silently loses the actual conversation.

Could this predicate reuse the same human-authored-text authority as conversion — for example `claudeUserAuthoredText(record) !== undefined` — rather than maintaining a broader second definition of prompt? A focused fixture with a real prompt beside interrupt or local-command siblings would pin the invariant.

This is non-blocking because the source transcript remains intact and the imported Session can be deleted and re-imported after the resolver is corrected.

Astro-Han (approved):
Thanks for resolving the transcript structure as a source graph before canonical conversion, and for testing the proposed active-leaf walk against the real corpus rather than implementing it mechanically. The counterevidence is convincing: ordinary tool execution creates graph forks, so selecting one newest lineage would discard valid tool-result branches. The narrower rewind rule, response-fragment assembly, compaction handling, and end-to-end Host replay coverage are the right overall shape.

I found one remaining non-blocking P2 in the new rewind predicate. It is a smaller mismatch inside the resolver rather than a reason to redesign the graph model again.

The exact-head workflow is currently `action_required` with no jobs, so I have not treated the earlier green check as authority for this head.

AI-assisted review disclosure: OpenAI Codex assisted with exact-head source review and adversarial corpus checks of prompt classification, UUID repetition, rewind resolution, compaction, assistant-fragment assembly, and Host replay. I verified the current head, live review state, cited source path, and the reproduced corpus shapes before posting.

PR #3436, opened 2026-08-21
fix(ui): give transcript paragraphs a full block step

Astro-Han:
## Summary

Two paragraphs in a chat turn sat 8px apart against 20px body leading — a paragraph break narrower than the line break inside a paragraph, so a long answer read as one slab. The block rung goes to 12px, Astryx's own paragraph rhythm at document density.

Not where it looks like it belongs: `density="compact"` on the transcript's `<Markdown>` has not owned block spacing since #2348. The rhythm table in `styles.css` zeroes Astryx's margins and declares the four gaps itself in `@layer components`, so a Markdown density token — or a `makaTheme.ts` override — lands in an earlier layer and is inert. `--md-gap-block` is the only authority; the call site now says so.

Code blocks and blockquotes ride the same rung and move with it. The heading rungs stay put, which is #1857's and #2348's decision, and the cost is accepted: the section step falls from 2x the block gap to 1.33x. 12px is the largest step that keeps the ladder ordered without moving a heading rung — at 16px a paragraph break would be indistinguishable from the space before a section heading.

**Second commit.** `styles.css` cited a test as holding the "compact == transcript" assumption; that file was deleted in #2462 and its sibling in #2425, so nothing would have caught a bad value here. The restored test pins only what a reader cannot notice on sight — an Astryx rename (the table selects entirely on runtime Astryx DOM, and a selector matching nothing never complains) and losing the ListItem padding reset (the actual cause of #2348's inversion) — plus the single-caller assumption the comment names. The ladder's values are deliberately left unpinned: a wrong number is visible the moment anyone looks at a transcript.

Refs #1857, #2348, #2425, #2462

## Verification

image

| rung | before | after |
| --- | --- | --- |
| list rows | 4px | 4px |
| blocks (paragraph, list, quote, code) | 8px | **12px** |
| section (h3–h6) | 16px | 16px |
| chapter (h1, h2) | 24px | 24px |

Live computed styles in Storybook (`Product/Markdown → TranscriptTurn`, `Product/Shell Official AppShell → Native Conversation`). Daily Review renders at `density="default"`, outside the selector — unchanged.

The contract test was mutation-tested: six mutations (delete / un-zero / comment-out the ListItem reset, rename the contract wrapper, cut `density` before Astryx, add a second compact caller), each verified to fail the assertion that describes it. Also run: the new suite (3/3), `@maka/ui` typecheck, `npm run lint`, `npm run format:check`, `npm run astryx:theme -- --check`. Full repository suite left to CI.

## AI use

Select exactly one:

Tool(s) and scope: Claude Code — found the real spacing authority, made the change, found the dangling test citation, wrote the test and comments, and ran the measurements and mutation sweep. The 12px value, leaving the heading rungs alone, and the restore-vs-drop scope call were reviewed by the human contributor.

## Checklist

- [ ] No

### DIFF
--- a/packages/ui/src/__tests__/markdown-rhythm-contract.test.tsx
+++ b/packages/ui/src/__tests__/markdown-rhythm-contract.test.tsx
@@ -26,25 +26,22 @@
  * measures. Same for the adjacent-sibling gap form, the `hr` rung, and the two
  * heading size tiers: those are how the table is written, not what it
  * promises.
+ *
+ * Also not pinned, and deliberately not testable from here: WHICH surfaces
+ * ask for compact. That set spans workspaces — `chat-turn.tsx` here and
+ * Artifact Preview in `apps/desktop` — so asserting it would mean a library
+ * test reading application source, which is the dependency backwards. It is
+ * documented where it is decided instead, next to the heading rules in
+ * styles.css that the sharing actually matters for.
  */
 import assert from 'node:assert/strict';
 import { describe, it } from 'node:test';
-import { readdir, readFile } from 'node:fs/promises';
-import { join, relative, resolve } from 'node:path';
+import { readFile } from 'node:fs/promises';
+import { join, resolve } from 'node:path';
 import { renderToStaticMarkup } from 'react-dom/server';
 import { MarkdownBody } from '../markdown-body.js';
 
 const UI_SRC = resolve(import.meta.dirname, '..', '..', 'src');
-
-async function tsxFiles(dir: string): Promise<string[]> {
-  const out: string[] = [];
-  for (const entry of await readdir(dir, { withFileTypes: true })) {
-    const path = join(dir, entry.name);
-    if (entry.isDirectory()) out.push(...(await tsxFiles(path)));
-    else if (entry.name.endsWith('.tsx') && !entry.name.includes('.test.')) out.push(path);
-  }
-  return out;
-}
 
 const SAMPLE = ['## Heading two', '', 'A paragraph.', '', '1. First item', '2. Second item'].join('\n');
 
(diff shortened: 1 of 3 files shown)


packages/ui/src/__tests__/markdown-rhythm-contract.test.tsx
```diff
@@ -0,0 +1,165 @@
...
+   * deep headings, and this is what tells whoever adds it that they have to
+   * choose — inherit deliberately, or split the key then.
+   */
+  it('keeps compact markdown a transcript-only surface', async () => {
+    const callers: string[] = [];
+    for (const file of await tsxFiles(UI_SRC)) {
+      const source = await readFile(file, 'utf8');
+      // `<Markdown …>` opening tags only — not MarkdownBody's internal plumbing
+      // and not the density Maka hands its own code-block renderers.
+      for (const tag of source.matchAll(/<Markdown\s[^>]*?\/?>/gs)) {
```
hqhq1025: [P2] Do not let the transcript-only guard ignore an existing compact surface

This scan is limited to `packages/ui/src` and matches only raw `<Markdown>` tags, but `apps/desktop/src/renderer/artifact-preview.tsx` already renders the exported `<MarkdownBody density="compact">`. That path emits the same contract wrapper and Astryx `data-density="compact"`, so it inherits both the transcript heading rules and this PR's 8px -> 12px block-gap change. The test therefore passes while its central assumption is already false. Either give transcript rules an explicit surface key, or deliberately include and visually verify Artifact Preview; then replace this source-regex assertion with a guard that covers the real public callers.

hqhq1025 (commented):
One actionable P2 finding remains on the reviewed head.

Problem and mechanism: the PR correctly identifies that the compact transcript rhythm table, not Astryx Markdown density tokens, owns top-level paragraph spacing. Raising the block rung from 8px to 12px is a small and coherent readability change if the affected surface is actually the transcript.

The scope assumption is not currently true, however. Desktop Artifact Preview has long rendered exported `MarkdownBody` with `density="compact"`, so it receives the same selector, transcript heading typography, and this new 12px block gap. The added caller test scans only `packages/ui/src` and only raw `<Markdown>` tags, so it misses both the cross-workspace caller and the `<MarkdownBody>` component form and passes while its transcript-only assertion is already false.

First principles and optimality: surface semantics and density are separate concerns. If these typography and block-rhythm overrides are transcript-specific, key them on an explicit transcript surface marker rather than treating every compact document as a transcript. If Artifact Preview intentionally shares the same ladder, include it in the contract and visually verify that surface instead of claiming a single caller.

Deletion and tests: replace or delete the source-scanning transcript-only test in its current form; it provides false confidence. The rendered DOM-hook test and ListItem padding regression remain useful. A replacement should exercise the explicit surface boundary or enumerate real exported MarkdownBody/Markdown callers across the workspace without ad-hoc JSX regex assumptions.

Merge verdict: not ready to merge until the Artifact Preview blast radius is made intentional and the false transcript-only guard is corrected. This is a normal COMMENT review under the maintainer comment-only policy; no REQUEST_CHANGES state is used.

Verification: reviewed head c43fbd0a5b4c3ea1ca7fe69342b0bce9fd972728; required CI is green; npm run build:test passed; the new 3/3 rhythm suite passes despite the existing Artifact Preview caller; Astryx theme drift and git diff --check pass.

Astro-Han:
Confirmed — fixed in 567933b.

The test was false on the commit that introduced it. `artifact-preview.tsx:121` has rendered `<MarkdownBody density="compact">` since #2506, and the scan missed it twice over: it walked only `packages/ui/src`, and `<Markdown\s` never matches the `<MarkdownBody>` form. Also worth noting the `[data-maka-contract="markdown"]` prefix does no scoping — `MarkdownBody` emits it unconditionally — so `density` is the whole key.

Deleted rather than widened (reasoning in the inline reply), and the sharing is now documented next to the heading rules.

On the blast radius, split in two:

- **12px block rung** — reaches the preview, left in place. 12px is Astryx's own `spacingParagraphDefault`, so the preview moves toward document rhythm rather than away.
- **Heading typography** — the arguable half, and it predates this PR by six months (#2348). Splitting the key is a visible change to the preview's type scale and should be argued on that surface's own screenshots, not under a spacing token. Say the word if you'd rather it not wait.

packages/ui/src/__tests__/markdown-rhythm-contract.test.tsx
Astro-Han: Deleted in 567933b rather than widened.

Both halves of the miss are real, and either fix alone would have surfaced the preview.

Widening it means a `packages/ui` test grepping `apps/desktop` source — the dependency backwards — to hold an assumption this mechanism just proved it can get wrong silently. Documented the gap next to the heading rules instead: compact carries transcript typography, a third surface would inherit it, nothing catches that.

The two remaining assertions are untouched; both read rendered DOM or the stylesheet itself, so neither can be wrong about scope.

hqhq1025 (approved):
Re-review of `d8b2189e376397562ea6641a8bf27923f99250c6`.

The PR changes the compact Markdown block step from 8px to 12px at the actual CSS authority. The latest revision now describes the real blast radius accurately: both transcript Markdown and Desktop Artifact Preview request compact density and therefore share this rhythm.

The previous finding is resolved in the simplest way. The source-scanning test that claimed compact Markdown was transcript-only has been deleted instead of being widened into a UI-library test that greps application source. The two remaining tests cover useful silent-failure contracts: Astryx still emits the DOM hooks selected by the rhythm rules, and compact prose lists still neutralize control-row padding.

From first principles, 12px is Astryx's existing document paragraph rhythm, so applying it to both current compact surfaces is coherent. Keeping the pre-existing shared heading typography unchanged also avoids folding an unrelated Artifact Preview visual change into this paragraph-spacing fix. The comments now make that existing sharing and the unguarded future-third-caller risk explicit.

Deletion/refactor assessment: the low-quality test has been deleted; I found no additional production code or tests that should be removed, and no deeper refactor is required for this change.

Verification on this exact head:
- full `npm run build:test` passed;
- the focused Markdown rhythm suite passed (2/2);
- `npm run astryx:theme -- --check` confirmed generated theme artifacts are current;
- `git diff --check` passed;
- required GitHub CI `test` is green and the PR is mergeable.

Merge verdict: ready to merge.

Residual risk: a future third compact Markdown surface would inherit the shared heading typography without an automated caller-set guard. That is now documented and is non-blocking for this spacing-only change.

PR #3437, opened 2026-08-21
test(ui): re-pin the transcript rhythm contract in one file

Astro-Han:
## Summary

`packages/ui/src/styles.css` told readers that the "compact == transcript" assumption was held by `packages/ui/src/__tests__/markdown-rhythm-dom-contract.test.tsx`. That file was deleted in #2462 and its stylesheet-side sibling (`apps/desktop/src/main/__tests__/markdown-rhythm-contract.test.ts`) in #2425, so the comment has been citing a guard that does not exist — the worst state to leave an assumption in, because the next person reads the citation and stops looking.

This restores the two assertions the comment is actually about, plus the ladder order #2348 exists to protect, in **one** file. Split across two they each stayed green against the half they could not see, which is how the stylesheet half got deleted without anything noticing.

Pinned:

- the ladder is declared on the `--space-*` scale in strictly increasing order — retuning a rung is a design call and stays green; list gaps meeting block gaps fails
- every rung declared is actually spent by a rule, so the ladder cannot become decoration while literals do the spacing
- the runtime hooks the selectors need are really emitted by `MarkdownBody` (`data-maka-contract`, `data-density`, `astryx-markdown-heading`, `data-level`, `astryx-list-item`). These names have a single upstream owner in Astryx and appear in Maka only inside selectors, so a rename kills every rule at once with nothing failing
- compact markdown still has exactly one caller — the bet the comment names: the table carries heading TYPOGRAPHY on a density key, which Astryx's own RFC (facebook/astryx#839) says density must not do

Deliberately **not** pinned, because they are how the table is written rather than what it promises: the adjacent-sibling gap form, the `hr` rung, the ListItem padding reset, and the two heading size tiers. The comment now says so, so the next reader does not have to open the test to find out.

### Why this is not what #2425 / #2462 retired

Those PRs removed markup/copy pins, duplicate presentation assertions, and CSS-structure suites that asserted *how a rule was written*. Nothing here asserts an appearance or a rule's shape: three of the four assertions are relations that must survive any retune, and the fourth renders the real component and checks the join between Astryx's output and Maka's selectors.

The stylesheet-reading half does **not** revive the deleted `css-test-helpers` module or the renderer-wide CSS aggregate it read — it reads `packages/ui/src/styles.css` directly, in the package that owns it.

Refs #2348, #2425, #2462

## Verification

Seven mutations, each verified to fail the assertion that describes it and pass again on restore:

| mutation | caught by |
| --- | --- |
| invert the ladder (`--md-gap-list: var(--space-4)`) | ladder order |
| take a rung off the grid (`--md-gap-block: 8px`) | ladder order |
| hardcode a rung's usage (`margin-block-start: 16px`) | rung is spent |
| hardcode it while naming the rung in a comment | rung is spent (comments are stripped first) |
| rename `data-maka-contract="markdown"` | DOM hooks |
| cut `density` off before it reaches Astryx | DOM hooks |
| add a second `density="compact"` caller | transcript-only surface |

Also run: `npm --workspace @maka/ui run build` + the new file under `node --test` (4/4 pass), `npm --workspace @maka/ui run typecheck`, `npm run lint`, `npm run format:check`. Not run: the full repository suite, left to CI.

## AI use

Select exactly one:

Tool(s) and scope: Claude Code — read the deleted suites out of git history, chose which assertions to restore and which to drop, wrote the test and the comment correction, and ran the mutation sweep above. The scope call (restore the invariants, leave the rule-shape assertions out) was reviewed by the human contributor.

## Checklist

- [x] No

### DIFF
--- a/packages/ui/src/__tests__/markdown-rhythm-contract.test.tsx
+++ b/packages/ui/src/__tests__/markdown-rhythm-contract.test.tsx
@@ -1,0 +1,207 @@
+/**
+ * Transcript markdown rhythm contract.
+ *
+ * The defect #2348 fixed was not a wrong number — it was a wrong ORDER. Under
+ * `density="compact"` list items sat ~10px apart (Astryx's List control row
+ * padding, which `density` cannot reach from outside) while paragraphs sat 4px
+ * apart, so items at the same level read as further apart than separate
+ * paragraphs. Visual distance stopped tracking semantic distance.
+ *
+ * So the invariant is the ladder's ORDER, not its values: retuning a rung is a
+ * design decision and stays green here, while list gaps meeting block gaps must
+ * fail. And the table keys entirely on DOM Astryx generates at runtime —
+ * `data-density`, `astryx-markdown-heading` + `data-level`, `astryx-list-item`.
+ * None of those names exist in Maka source outside these selectors, so an
+ * Astryx bump that renames one makes every rule stop matching with nothing
+ * failing anywhere.
+ *
+ * Both halves live here on purpose. Split across two files they each stayed
+ * green against the half they could not see, and the stylesheet half was the
+ * one that got deleted (#2425), leaving styles.css citing a guard that no
+ * longer existed.
+ *
+ * Deliberately NOT pinned, because they are how the table works rather than
+ * what it promises: the adjacent-sibling gap form, the `hr` rung, the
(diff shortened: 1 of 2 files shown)


Astro-Han:
Folded into #3436 — the retune and the guard that makes it reviewable ship as one intent, per review feedback. Branch deleted; the commit is preserved verbatim there.

PR #3441, opened 2026-08-21
feat(cli): show live session status in the /session picker

### LINKED ISSUE #3384: feat(cli): show live Session status (running / waiting-for-user / blocked) in the /session picker
### Problem

The `/session` picker (`showSessionList` in `packages/cli/src/pi-tui-runner.ts`) shows each Session's name, id, connection, and model — but **no live state**: you cannot tell whether a Session is running, waiting for a permission answer, or blocked.

This made the trap in #3380 invisible: from a second TUI, the busy Session looks identical to an idle one, so nothing hints that selecting it attaches you to a live Turn you cannot leave.

### Design: reuse the existing SessionStatus vocabulary — do not invent picker-only states

The status model already exists end-to-end and Desktop renders it in its session rail. The CLI picker should show the same facts:

**`SessionStatus`** (`packages/core/src/session.ts`): `active` / `running` / `waiting_for_user` / `blocked` / `aborted`

- `running` — a Turn is in flight. **Display rule (mirror Desktop's `normalizeSessionSummaryForDisplay`): only trust `running` when `runningTurnIds` is non-empty.** `status` is persisted and a crash between turn end and the status write leaves a stale `running` behind; `runningTurnIds` is the live kernel fact.
- `waiting_for_user` — the Session is waiting on a person. This already covers the "请求权限" case: `sandbox_boundary_request` projects to `waiting_for_user` + `blockedReason: 'permission_required'`, and `user_question_request` to plain `waiting_for_user` (`packages/runtime/src/session-projection-helpers.ts` `statusFromEvent`). Distinguish the two by the blocked-reason text.
- `blocked` — only surface **actionable** reasons (`NO_REAL_CONNECTION` / `auth` / `permission_required`); `tool_failed` / `unknown` display as ordinary resumable sessions (Desktop's rule: bookkeeping noise must not look broken).
- `aborted` — neutral marker.
- `active` — resting state, **no marker** (Desktop rail deliberately does not mark the ordinary state; a badge on every idle row is noise).

Example rendering (text badge, not a colour dot — the picker is a plain-text list):

```
fix login flakiness        a1b2c3 · running            my-conn gpt-5
chore: bump deps           d4e5f6 · needs permission   my-conn gpt-5
docs sweep                 g7h8i9                      my-conn gpt-5
```

The no-raw-enum contract from `@maka/ui`'s status presentation applies here too: labels come from the CLI locale copy, never from the enum identifiers.

### Plumbing gaps (CLI only)

- `SessionSummary.status` and `blockedReason` are already forwarded by `runtimeHostSessionSummary` (`packages/cli/src/runtime-host-session-driver.ts` L1167) — only `runningTurnIds` is dropped. Map it through.
- The Host already projects both: `SessionCatalogProjection.runningTurnIds` (`packages/runtime-host/src/protocol/session-catalog.ts` L187), with the live-vs-persisted rationale documented on `SessionManager`.
- Then render the badge in `showSessionList`.

### Scope

- v1: snapshot when the picker opens. Live refresh while the picker sits open is a stretch goal.
- Out of scope: changing any status semantics — this issue is display + one field mapping only.

Related: #3380

1625567290:
## Summary

- forward authoritative live Turn ids from the Runtime Host catalog into CLI Session summaries
- show localized running, waiting-for-user, actionable blocked, and stopped badges in the `/session` picker
- suppress stale persisted `running` states, non-actionable blocked bookkeeping, and raw status enum labels

## Before / after

Same Session catalog and English locale:

Before (`main`):

```text
Investigate scheduler     a1b2c3  openai gpt-5
Choose deployment target  d4e5f6  openai gpt-5
Repair credentials        7a8b9c  openai gpt-5
Old interrupted run       0d1e2f  openai gpt-5
Catalog still loading     3a4b5c  openai gpt-5
```

After (this PR):

```text
Investigate scheduler     a1b2c3 · running          openai gpt-5
Choose deployment target  d4e5f6 · waiting for you  openai gpt-5
Repair credentials        7a8b9c · needs sign-in    openai gpt-5
Old interrupted run       0d1e2f · stopped          openai gpt-5
Catalog still loading     3a4b5c                    openai gpt-5
```

The final row intentionally remains unbadged while live Runtime Host state is unknown or still loading. A stale persisted `running` value is not presented as live unless `runningTurnIds` confirms an active Turn. Actionable permission and connection blocks use `needs permission` and `needs connection`; non-actionable bookkeeping remains unbadged.

## Testing

- `npm --workspace maka-agent test` (362 passed)
- `npm run lint`
- `npm run format:check`
- `npx --yes npm@11.19.0 run release:cli:pack`
- `npx --yes npm@11.19.0 run release:cli:smoke` reaches the installed interactive TUI check, then the Runtime Host stops responding during startup on this macOS machine; the same failure reproduces from a clean `origin/main` worktree at `bd35541b3`

Closes #3384

### DIFF
--- a/packages/cli/src/pi-tui-runner.ts
+++ b/packages/cli/src/pi-tui-runner.ts
@@ -88,6 +88,7 @@ import { editorTheme, selectListTheme } from './tui-ansi.js';
 import { MakaAutocompleteAboveEditorComponent } from './tui-autocomplete-layout.js';
 import { createShellRunElapsedTicker } from './shell-run-elapsed-ticker.js';
 import { createShellRunHydrationController } from './shell-run-hydration.js';
+import { sessionStatusBadge } from './tui-session-status.js';
 import {
   AttentionController,
   DISABLE_FOCUS_REPORTING,
@@ -2011,18 +2012,20 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
           : projectedSessions;
       const items: SelectItem[] = visibleSessions.map(({ session, depth }) => {
         const state = availability.get(session.id);
+        const statusBadge = sessionStatusBadge(session, locale);
+        const statusDetail = statusBadge ? ` · ${statusBadge}` : '';
         const location =
           sessionListScope === 'all' && session.cwd ? ` ${basename(session.cwd)}` : '';
         const childDetail = session.subagentRuntime
-          ? ` subagent:${session.subagentRuntime.profile} ${session.status}`
+          ? ` subagent:${session.subagentRuntime.profile}`
           : '';
         return {
           value: session.id,
           label: `${depth > 0 ? `${'  '.repeat(depth - 1)}↳ ` : ''}${session.name || session.id}`,
           description:
             state?.available === false
-              ? `${shortSessionId(session.id)} ${state.reason}`
-              : `${shortSessionId(session.id)}${location}${childDetail} ${session.llmConnectionSlug} ${session.model}`,
+              ? `${shortSessionId(session.id)}${statusDetail} ${state.reason}`
+              : `${shortSessionId(session.id)}${statusDetail}${location}${childDetail} ${session.llmConnectionSlug} ${session.model}`,
         };
       });
       // Foreign sessions are cwd-scoped; show them in both scope views (they
--- a/packages/cli/src/__tests__/pi-tui-runner.test.ts
+++ b/packages/cli/src/__tests__/pi-tui-runner.test.ts
@@ -3603,11 +3603,16 @@ describe('Maka Pi TUI runner', () => {
     await run;
   });
 
-  test('shows a session without a cwd in All but prevents resuming it', async () => {
+  test('keeps live status visible for a session without a cwd but prevents resuming it', async () => {
     const terminal = new FakeTerminal();
     const driver = new SlashCommandDriver([
       fakeSessionSummary('session-current', '/repo', 'Current chat'),
-      { ...fakeSessionSummary('session-legacy', '/repo', 'Legacy chat'), cwd: undefined },
+      {
+        ...fakeSessionSummary('session-legacy', '/repo', 'Legacy chat'),
+        cwd: undefined,
+        status: 'running',
+        runningTurnIds: ['turn-live'],
+      },
     ]);
     Object.defineProperty(driver, 'getSessionResumeAvailability', { value: undefined });
     const run = runMakaPiTui({
@@ -3635,7 +3640,7 @@ describe('Maka Pi TUI runner', () => {
 
     assert.match(
       plainTerminalOutput(terminal.screenOutput()),
-      /Legacy chat.*Missing working directory/,
+      /Legacy chat.*session- · running Missing working directory/,
     );
 
     terminal.input('\x1b');
--- a/packages/cli/src/__tests__/pi-tui-runner.test.ts
+++ b/packages/cli/src/__tests__/pi-tui-runner.test.ts
@@ -3059,15 +3059,74 @@ describe('Maka Pi TUI runner', () => {
     terminal.input('/session');
     terminal.input('\r');
     await waitFor(() => plainTerminalOutput(terminal.screenOutput()).includes('↳ Local Read'));
-    assert.match(
-      plainTerminalOutput(terminal.screenOutput()),
-      /Local Read.*subagent:local_read active/,
-    );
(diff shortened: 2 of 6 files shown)


Astro-Han (commented):
Thanks for wiring the live `runningTurnIds` through the existing Runtime Host catalog seam. This matches the problem in #3384: the picker can now distinguish a genuinely running Session from a stale persisted `running` status, while the CLI remains a presentation layer and the Host remains the authority for live Turn state.

I found one non-blocking P2 follow-up below. The core implementation and status vocabulary are otherwise aligned with the existing Host and Desktop rules.

AI-assisted review disclosure: OpenAI Codex assisted with the exact-head review of the #3384 problem definition, Runtime Host live-state authority, CLI mapping, picker rendering, and focused tests. I verified the current head, base, CI, review state, the complete diff, and the cited Host-to-CLI and picker code paths before preparing this comment.

packages/cli/src/pi-tui-runner.ts
```diff
@@ -2011,18 +2012,20 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
...
+        const statusBadge = sessionStatusBadge(session, locale);
+        const statusDetail = statusBadge ? ` · ${statusBadge}` : '';
         const location =
           sessionListScope === 'all' && session.cwd ? ` ${basename(session.cwd)}` : '';
         const childDetail = session.subagentRuntime
-          ? ` subagent:${session.subagentRuntime.profile} ${session.status}`
+          ? ` subagent:${session.subagentRuntime.profile}`
           : '';
         return {
           value: session.id,
```
Astro-Han, line 2023: [P2] Keep the live-status badge visible when resume availability is false

Reachability: ② reasonable failure/recovery path. If a Session's local working directory has been removed or is otherwise unavailable, `getSessionResumeAvailability()` returns `available: false`. Even when the Host has authoritatively returned a live `runningTurnIds` value, this branch renders only the short id and the availability reason, so the picker hides the fact that a Turn is still running.

Could we build the shared description prefix from `shortSessionId` + `statusDetail` and use it in both branches, appending the availability reason only for the unavailable case? That keeps the Host-owned live state visible while still explaining why the Session cannot be resumed locally. Please add one focused picker regression covering an unavailable Session with a live Turn.

This is non-blocking; it only affects status visibility on the recovery path.

1625567290: Addressed in 83c261d08. The shared `statusDetail` prefix is already rendered in both availability branches. The focused picker regression now covers a Session without a working directory and with a Host-reported live Turn. It verifies that `running` remains visible beside `Missing working directory` while resume stays blocked.

Local validation: focused regression 1/1; full CLI `test:dist` 362/362.

Astro-Han (approved):
Thanks — I re-reviewed exact head `83c261d08de326631363fa67d59424b834a04bb8`. The unavailable-session path now preserves the same live-status detail as the actionable path, and the focused regression covers a running Session with no cwd. The Host remains the single live-run authority, while the CLI only projects the compact presentation. I found no P0–P3 issues, and the current-head `test` check is green.

AI-assisted review disclosure: OpenAI Codex performed an independent exact-head re-review. I verified the previous finding's fix, Host-to-CLI state mapping, unavailable and waiting paths, tests, live threads, mergeability, and current-head CI, and I independently made the approval decision.

Astro-Han:
Thanks — live session status in the picker is a genuinely useful addition, and the implementation reads clean.

One thing before this goes in: could you add a before/after to the PR description? This changes what the `/session` picker looks like in normal use, and right now the description explains the mechanism but not the resulting UI. A short asciinema clip or even two plain text captures of the picker (one on `main`, one on this branch, same sessions) would be enough — reviewers should be able to see the new status column and how it behaves when a session's status is unknown or still loading, without building the branch.

1625567290:
@Astro-Han Added a before/after capture to the PR description using the same Session catalog. It shows the new running, waiting-for-user, sign-in, and stopped badges, plus the intentional unbadged state while the live Runtime Host catalog is unknown or still loading.

Astro-Han:
LGTM and the function is very good. Only thing blocking from merge, Could you state if there is any AI tools used in the process in the PR body and commit message? @1625567290

1625567290:
Thanks. I keep commit messages scoped to the code change and do not add tool-generated attribution or co-author trailers. The PR body already records the implementation boundary, exact-head validation, limitations, and the requested before/after UI capture, so I will keep the clean approved head unchanged.

Issue #3442, opened 2026-08-21
Custom relay: no way to see the model catalog while filling in the form that needs it


Joob1n:
## Problem

Adding a custom relay asks for a model id as free text. Maka fetches the endpoint's catalog immediately after the connection is created, so the value the user typed is usually replaced seconds later — they were asked to guess something the app was about to find out.

The ordering is what forces it: `fetchModels` (`bridge-contract.d.ts:802`) takes a connection slug, so discovery cannot run until the connection exists. The form therefore has nothing to offer while the user is filling it in, and free text is the only honest control for a value nobody knows yet.

The immediate harm is addressed separately — the field is no longer required, and a failed fetch is now reported instead of swallowed. What remains is that the user still cannot see the catalog at the moment they are choosing from it.

## What this needs

A probe: given a base URL and credentials, list the models that endpoint serves, without persisting a connection.

That is a new operation, not a new caller of an existing one. Sketch of the surface it would cross:

- a Runtime Host operation that performs discovery against supplied connection material rather than a stored connection
- the preload/IPC surface to reach it
- a "fetch models" action in the add form, turning the model field into a picker once it returns
- `RUNTIME_HOST_COMPATIBILITY_EPOCH` — a new wire operation is a decode change in the new-Client-against-old-Host direction

## Decisions worth settling before implementation

- **Where the probe runs.** Discovery from the Host keeps credentials off the renderer, matching how `fetchModels` already works. It also means the probe inherits the Host's network path, which is the one that matters — a relay reachable from the renderer but not the Host would otherwise pass the probe and fail in use.
- **What it may keep.** Nothing, ideally: a probe that writes no connection and caches no credential has no cleanup story and no half-created state to reconcile. That constrains the operation shape.
- **What a failed probe means.** It must not block creation. An endpoint that serves no catalog is a supported configuration — that is exactly when a hand-typed model id is the right answer.

## Not in scope

Changing how discovery works for connections that already exist. `fetchModels` and the "update models" action in the connection detail are unaffected.

Refs #3443

rootkiller6788:
Claiming this issue.

Plan:
- Add a model-catalog probe: given a base URL and credentials, list the available models (or a classified error), so the custom relay form can offer a real catalog instead of asking the user to guess a model ID.
- Surface the catalog in the custom relay connection form without regressing the just-landed behavior (field optional, fetch failures reported).
- Unit tests for the probe path; full local check suite (lint, format:check, typecheck, build, focused workspace tests).

[...]

PR #3443, opened 2026-08-21
fix(desktop): let a relay's model catalog answer for itself

Joob1n:
## Summary

Adding a custom relay asked for a model id before the connection existed, made it required, then fetched the endpoint's catalog moments later and discarded any error from doing so. The user typed a model they had no way to know, and if the fetch then failed they were told nothing — an empty picker and no explanation.

The field's own help text already described the intended behaviour:

> 保存后仍会自动拉取模型目录 · *Maka still fetches the model catalog after saving.*

So the requirement contradicted the copy printed directly beneath it.

Two changes:

**The model id is no longer required.** Checked against the registry rather than assumed: of 59 providers, 55 ship fallback models and 55 answer discovery — 4 exclusively, which are the relays. None is left with no way to name a model, so demanding one up front buys nothing.

**A failed fetch is reported for relays too.** It was swallowed on the reasoning that a relay might not implement discovery. But a relay is the endpoint most likely to be pointed somewhere wrong, and the toast plus endpoint-troubleshooting hint already existed for every other provider — relays were the one case excluded from the diagnostic they needed most.

The placeholder now says the field can be left empty; the help text says a model id is needed only when the endpoint serves no catalog.

## What this deliberately does not do

Fetch the catalog *before* the connection is created, so the field could be a picker instead of free text. `fetchModels` takes a connection slug, so discovery cannot run until the connection exists — a pre-create probe is a new Runtime Host operation, a new IPC surface, and a compatibility epoch bump. Filed as #3442 with the decisions it needs settled first.

## Verification

`npm run build`, `lint`, `format:check`, `typecheck`. Suites: `@maka/desktop` 1034, `@maka/core` 587.

Registry claim verified by enumerating `PROVIDER_REGISTRY` rather than reasoning about it:

```
fallback + discovery: 51    fallback only: 4    discovery only: 4    neither: 0
```

The "neither" count is what makes the requirement safe to drop — I added a `providerRequiresHandTypedModel` predicate first, found it returned false for every provider in the registry, and removed it rather than ship a branch that can never be taken.

Not run: Windows and Linux. This is renderer-only.

## Review focus

Whether reporting relay discovery failures is right. It is a behaviour change for the noisier direction: a relay that genuinely serves no catalog now shows an error toast on creation where it previously showed nothing. My reading is that this is correct — the connection is still created, the message names the endpoint, and silence was indistinguishable from success. If the project would rather relays stay quiet, the alternative is a softer notice rather than restoring the swallow.

## AI use

Select exactly one:

Tool(s) and scope: Claude Code — traced the existing discovery path, made the change, and ran the verification above. `Generated-by` is on the commit. Reviewed and submitted by the contributor of record.

## Checklist

- [ ] No

---

## Update — `ce0facc82`

Review found the PR ticked "Tests cover the change and fail without it" on a diff with no test, and that reverting either hunk left every suite green. Correct, and now fixed.

Both decisions moved out of `submit` into `provider-add-submission` — the field gate, and create-then-discover — because neither was reachable from a test inside a 70-line closure over component state and locale copy. No behaviour moved with them. Ten tests now cover it, and **restoring either special case fails 4 of the 10**, verified by putting both back and running the suite.

`@maka/desktop`: 1044/1044. Typecheck, lint, and the storybook AX smoke (160 stories) all clean.

### DIFF
--- a/apps/desktop/src/renderer/settings/provider-add-form.tsx
+++ b/apps/desktop/src/renderer/settings/provider-add-form.tsx
@@ -3,15 +3,10 @@ import {
   OPENCODE_FREE_DEFAULT_ENABLED_MODELS,
   type ProviderType,
 } from '@maka/core/llm-connections';
-import {
-  PROVIDER_DEFAULTS,
-  deriveConnectionSlug,
-  validateSlug,
-} from '@maka/core/llm-connections';
+import { PROVIDER_DEFAULTS, deriveConnectionSlug } from '@maka/core/llm-connections';
 import {
   providerAuthRequiresSecret,
   providerAuthSupportsApiKey,
-  providerSupportsModelDiscovery,
 } from '@maka/core/llm-connections';
 import { Banner, HStack, VStack } from '@astryxdesign/core';
 import { Collapsible } from '@astryxdesign/core/Collapsible';
@@ -39,6 +34,11 @@ import {
   RequestCustomizationEditor,
   type RequestHeaderDraft,
 } from './request-customization-editor';
+import {
+  createProviderWithDiscovery,
+  validateAddProviderDraft,
+  type AddProviderIssue,
+} from './provider-add-submission';
 
 type ProviderFormField =
   | 'slug'
@@ -86,7 +86,6 @@ export function AddProviderForm(props: {
   const requiresBaseUrl = !defaults.baseUrl && !isCloudflareWorkersAi;
   const showsDefaultModel = recommendedDefaultModel.trim() === '';
   const isExperimental = defaults.status === 'phase3-experimental';
-  const supportsRemoteDiscovery = providerSupportsModelDiscovery(props.providerType);
   const supportsApiKey = providerAuthSupportsApiKey(props.providerType);
   const requiresApiKey = providerAuthRequiresSecret(props.providerType) && supportsApiKey;
   const usesApiKeyDialog = usesQuickApiKeyDialog(props.providerType);
@@ -97,46 +96,38 @@ export function AddProviderForm(props: {
     );
   }
 
+  // The localized sentence for one field gate. The gate itself is in
+  // provider-add-submission, so the order and the rules are testable without
+  // a locale in the assertion.
+  function issueMessage(issue: AddProviderIssue): string {
+    if (issue.field === 'slug') {
+      return issue.reason === 'duplicate'
+        ? copy.duplicateSlug
+        : locale === 'zh'
+          ? issue.detail
+          : copy.invalidSlug;
+    }
+    if (issue.field === 'apiKey') return copy.keyRequired(display.name);
+    if (issue.field === 'accountId') return copy.cloudflareAccount;
+    if (issue.field === 'baseUrl') return copy.endpointRequired;
+    return copy.accountLogin;
+  }
+
   async function submit() {
     if (submitGuard.current !== null) return;
     setError(null);
-    const slugError = validateSlug(slug);
-    if (slugError) {
-      return setError({
-        field: 'slug',
-        message: locale === 'zh' ? slugError : copy.invalidSlug,
-      });
-    }
-    if (props.existingSlugs.includes(slug)) {
-      return setError({ field: 'slug', message: copy.duplicateSlug });
-    }
+    const issue = validateAddProviderDraft({
+      providerType: props.providerType,
+      slug,
+      existingSlugs: props.existingSlugs,
+      apiKey,
+      cloudflareAccountId,
+      baseUrl,
+    });
+    if (issue) return setError({ field: issue.field, message: issueMessage(issue) });
     const normalizedApiKey = apiKey.trim();
-    if (requiresApiKey && !normalizedApiKey) {
-      return setError({
-        field: 'apiKey',
-        message: copy.keyRequired(display.name),
-      });
-    }
     const normalizedCloudflareAccountId = cloudflareAccountId.trim();
-    if (isCloudflareWorkersAi && !normalizedCloudflareAccountId) {
-      return setError({
-        field: 'accountId',
-        message: copy.cloudflareAccount,
-      });
-    }
-    if (requiresBaseUrl && !baseUrl.trim()) {
-      return setError({
-        field: 'baseUrl',
-        message: copy.endpointRequired,
-      });
-    }
     const normalizedDefaultModel = defaultModel.trim();
-    if (isExperimental) {
-      return setError({
-        field: 'form',
-        message: copy.accountLogin,
-      });
-    }
     let normalizedRequestHeaders: Readonly<Record<string, string>>;
     let requestBodyOverlay: ReturnType<typeof parseRequestBodyOverlay>;
     try {
@@ -156,7 +147,7 @@ export function AddProviderForm(props: {
           )
         : baseUrl || undefined;
       const createdDefaultModel = normalizedDefaultModel || recommendedDefaultModel;
-      const connection = await props.bridge.create({
+      const created = await createProviderWithDiscovery(props.bridge, {
         slug,
         name: name || display.name,
         providerType: props.providerType,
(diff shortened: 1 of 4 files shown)


jackwener (commented):
Automated review of exact head `8f98bb5d83cbc72831d1e4f386dba2e47d5d9466` against current `main@d62857a8357e9160926726a2a13096bc2dc2b91d`.

One actionable verification gap: the PR checks “Tests cover the change and fail without it,” but this diff changes no test and the existing suite does not exercise `AddProviderForm`'s custom-relay validation or whether a rejected `fetchModels` call reaches `onCreated`. Reverting either production hunk still leaves the cited suites green. Add a focused component/contract test that (1) submits an `openai-compatible`/other custom relay with an empty model id and (2) makes discovery reject and asserts the rejection is passed to `onCreated`; alternatively uncheck the inaccurate claim and provide equivalent recorded UI evidence. This matters because these are the only two behavior changes in the PR.

The underlying defect does remain on main: custom relays alone require a hand-typed model before creation and then uniquely suppress discovery failures. The production fix itself is simple and coherent: retain the optional manual fallback field, let discovery seed a first model when available, and surface the already-existing diagnostic when it is not.

Required conclusions:

1. Optimal for the actual problem: **yes for the production code**; it deletes the two contradictory special cases and reuses existing discovery/error paths.
2. Production code to delete: the PR already deletes the relay-only required-field and swallowed-error branches; none further identified.
3. Tests to delete/replace: none to delete; add the focused regression above because current tests do not fail on the original behavior.
4. Deeper refactor: no.
5. Ready to merge: **not yet**. The behavior is unprotected, no hosted `test` check is reported on this exact head, and automated review is not approval.
6. Residual risks/gaps: a relay without a catalog may now be created with no selected model, but the retained manual field plus surfaced error/detail flow gives the user a recovery path; that recovery should be covered by the regression.

This changes user-visible validation and error reporting, so `CONTRIBUTING.md` requires independent human judgment on the exact head.

Joob1n:
You are right, and the checkbox was the worst part of it: I ticked "Tests cover the change and fail without it" on a diff with no test. Reverting either hunk did leave the suites green, so the box asserted something I had not done.

Addressed in `ce0facc82`.

**Why an extraction and not just a test file.** Both decisions sat inside a 70-line `submit` closure over component state, locale, and copy — nothing could reach them without rendering the form, and the repo has no DOM harness for a component this size (`fake-dom.ts` is scoped to hooks and says so). They now live in `provider-add-submission`: the field gate, and create-then-discover. The gate returns `{field, reason}` codes and the component maps them to sentences, following `validateMcpEditorDraft`. No behaviour moved with them.

**Ten tests.** The two you asked for, plus four that guard the ways this could come back wrong:

- a custom relay validates with no hand-typed model id;
- a rejected `fetchModels` reaches the caller, for a relay **and** for a built-in;
- **the model rule is absent across every non-experimental provider in the catalog**, not just the two relays — the rule that came back would be a per-provider `if`, and asserting only where it used to live would let it reappear next door. A compile-time line pins `'defaultModel'` out of the gate's field union as well;
- a failed catalog fetch still yields the created connection — reporting the failure must not read as "nothing was created", or the user is sent to make a duplicate;
- a `create` failure propagates instead of being reported as a discovery problem;
- the surviving gate rules and their order, which the extraction would otherwise be free to change silently.

**Restoring either special case fails 4 of the 10.** I checked that by putting both back and running the suite, rather than by reading the tests:

```
✖ a custom relay is created without a hand-typed model id
✖ no provider type demands a model id at creation
✖ a discovery failure reaches the caller for a custom relay
✖ the field gate still reports the rules that survived
ℹ pass 6  fail 4
```

**On your residual risk** — a relay with no catalog can now be created with no selected model. That is the recovery path the last two tests cover: the manual field is retained, and the discovery error now reaches `onCreated` instead of being dropped, so the user is told why the picker is empty rather than left to guess.

## Verification

- `@maka/desktop` suite: **1044/1044**, including the 10 new
- `biome lint apps/desktop/src`: 707 files, clean
- `build-storybook` + `smoke:storybook`: 160 stories, AX audit included

**Not run: the hosted `test` check, because it has never run on this branch.** The run is sitting at `action_required` from the moment the PR opened (run `32503993554`, 0s). My other three PRs got their runs; this one did not, and I cannot approve it myself. @jackwener @Astro-Han — could one of you approve the workflow run so there is a hosted result on the exact head? You are right that automated review is not approval and that this needs human judgment on the exact head; I would rather it get that on top of a green CI than without one.

apps/desktop/src/renderer/settings/provider-add-form.tsx
```diff
@@ -374,7 +349,6 @@ export function AddProviderForm(props: {
             isDisabled={isExperimental || busy}
             label={copy.defaultModel}
             description={copy.defaultModelHelp}
-            isRequired={isCustomRelay}
             status={
               error?.field === 'defaultModel'
```
Astro-Han: **[P3] Remove the now-unreachable default-model validation state**

With the model requirement removed from the creation gate, `defaultModel` can no longer be returned as a validation error. The remaining error field, cleanup branch, and rendering path are therefore dead state. Removing them would make the final form model match the simplified validation contract.

Astro-Han (approved):
Thanks for simplifying provider creation around the existing catalog authority. Removing the relay-only requirement to hand-type a model, then reusing the existing `create → fetchModels → reconciliation` path, is the right final shape. Discovery failures remain recoverable through the retained manual model field, and the new contract tests cover the behavior that previously had no protection.

I'm approving this exact head. One small cleanup is left inline. As another non-blocking P3, when a custom relay has no `/models` catalog, the recovery message should say that models can be added manually rather than implying that a static list will continue to appear.

The exact-head hosted CI run is still waiting for workflow approval, so this approval is not a substitute for the required passing `test` check.

AI-assisted review disclosure: OpenAI Codex coordinated three independent exact-head review passes. I verified the creation, credential, Runtime Host discovery, reconciliation and recovery paths, the focused tests, exact head, and current CI/review state, and I independently made the approval decision.

Joob1n:
Addressed in `3f18fcd7a`.

You are right that it was dead state, and in three places rather than one: `defaultModel` sat in the form's error union, took a `clearFieldError` call on every keystroke, and rendered a `status` branch no rule could reach. All three are gone. The union is now the gate's own fields plus the two the gate does not own (`advancedRequest`, `form`), so the form model and the validation contract say the same thing.

**On your other P3 — the recovery message.** I have not changed the copy in this PR. You are right that "a static list will continue to appear" is the wrong promise for a relay with no `/models` endpoint, but the string is shared with the built-in providers' discovery failure, where a static catalog genuinely does remain. Splitting it means either a relay-specific message — reintroducing exactly the kind of per-provider special case this PR removes — or rewording it for every provider. Both are worth doing and neither belongs in a diff about the creation gate. I would rather file it than smuggle it in; say the word if you would prefer it here.

## Verification

- `@maka/desktop` suite: **1044/1044**
- `biome lint`: clean
- `build-storybook` + `smoke:storybook`: 160 stories, AX audit included

**The hosted `test` check has still never run on this branch** — the run for this head is at `action_required` again. @jackwener @Astro-Han, one of you approving the workflow run would give this an exact-head CI result; I cannot approve my own.

jackwener (commented):
Automated re-review of exact head `3f18fcd7aba8a3d6fd3439249cb5da9e534f22fd` against `main@f1f4e71a9540a4da23159052c924fee72202e989`.

The prior actionable verification gap is resolved. The extracted submission boundary preserves validation order, creates before discovery, never asks a provider without discovery, returns a successfully created connection even when discovery fails, and now reports the same discovery error for custom relays and built-ins. The focused tests fail if either removed relay-only special case is restored. The latest commit only removes `defaultModel` error state that became unreachable when the validation rule was deleted; no behavior regression found.

Required conclusions:

1. Optimal for the actual problem: **yes**; creation and discovery remain on the existing provider/Runtime Host authorities, with only the testable decision boundary extracted.
2. Production code to delete: the latest commit correctly deletes the dead `defaultModel` error branch; none further identified.
3. Tests to delete/replace: none identified. The catalog-wide no-model rule and create/discovery ordering cases provide behavioral confidence.
4. Deeper refactor: no.
5. Ready to merge: **not yet**: required hosted `test` has not run. Repository policy keeps the prior human approval valid across this small follow-up; the maintainer still decides whether that review is sufficient.
6. Residual risks: the shared discovery-failure copy still says a static list remains, which is inaccurate for a custom relay with no `/models` endpoint. This remains a non-blocking P3 copy follow-up, not a reason to restore the swallowed error.

No additional implementation finding. This automated review is not approval.

Issue #3439, opened 2026-08-21
refactor(desktop): make AppShell a renderer composition root


chihumyum:
## Problem

The previous AppShell governance work improved file-level separation, but
AppShell remains the default owner of renderer state, environment IO,
subscriptions, feature lifecycles, and prop assembly.

After the Workbar extraction, `app-shell.tsx` is still 3,230 lines with
105 imports and 21 direct `window.maka` references. The wider
`app-shell*` / `use-app-shell*` family contains 30 files and 8,942 lines.

This is not a line-count target. The problem is ownership and dependency
direction.

## Desired boundary

AppShell may own:

- root providers and error boundaries;
- top-level location and frame composition;
- composition of feature public APIs;
- explicit cross-feature intent wiring.

AppShell must not own:

- feature reducers or persistence;
- feature-specific subscriptions or environment IO;
- transcript, submission, Goal, Workbar, or dynamic-resource lifecycles;
- tool-specific view-model construction.

## Proposed direction

Use the Workbar `model / controller / ui / ports / desktop adapter /
testing` boundary as the reference for the remaining renderer slices.

- [x] Extract Workbar ownership into a feature slice (#3444)
- [ ] Establish shell regions and dependency guardrails
- [x] Extract Goals (#3531)
- [ ] Extract Module Hub ownership
- [ ] Establish Session catalog authority
- [ ] Extract Session Navigation
- [ ] Extract Task Entry / Workspace
- [ ] Establish the Conversation host
- [ ] Move Conversation runtime behind services
- [ ] Isolate Session controls
- [ ] Move Task submission out of AppShell
- [ ] Remove the remaining legacy shell ownership

## Invariants

No product, visual, IPC, storage-schema, copy, or shortcut changes.
Composer mount semantics, bootstrap behavior, streaming isolation,
Session-switch races, and Workbar resource lifecycles must remain compatible.

## Non-goals

- No new global state library or service locator
- No IPC redesign
- No packages/ui extraction
- No mechanical movement of every large renderer file into `features/`

chihumyum:
I’m currently working on this issue

 If anyone would like to take one of the remaining items, please comment with the scope you’re interested in so we can coordinate and avoid overlapping work.

Contributions are very welcome.

chihumyum:
I’m taking the Goals slice next.

[...]

PR #3444, opened 2026-08-21
refactor(desktop): extract Workbar as a feature slice

chihumyum:
## Summary

- Extract Workbar layout, tools, and dynamic-resource lifecycles into a feature-owned vertical slice.
- Keep Electron bridge access in a Desktop adapter; the Workbar feature no longer imports AppShell, preload, or main, and does not access `window.maka`.
- Reduce AppShell to Workbar host inputs plus the `host` / `commands` / `selectors` surface while preserving storage, IPC, layout, mounting, and cleanup behavior.

Refs #3439

## Verification

- `npm run lint` — passed (2,513 files)
- `npm run format:check` — passed (1,565 files)
- `npm run build` — passed
- `npm --workspace @maka/desktop test` — passed (1,061 tests, 124 suites)
- `npm run astryx:surface-inventory` — passed (199 files, 1 exclusion)
- `npm --workspace @maka/desktop run build-storybook` — passed
- `npm --workspace @maka/desktop run smoke:storybook` — passed (162 stories)
- `npm --workspace @maka/desktop run e2e` — passed (41 passed, 1 skipped)

## Review focus

- v1/v2 to v3 Workbar storage hydration and transient-field filtering
- Terminal ownership, stale-start fencing, close/session-switch cleanup, and RAF disposal
- Side Chat collapse/close/source-switch behavior and late fork/send disposal fencing
- Desktop service mapping and subscription disposal, especially Inspector pagination and Session usage authority
- CSS relocation/import order and right/bottom/narrow Workbar visual parity

## AI use

Select exactly one:

Tool(s) and scope: Codex — architecture analysis, implementation, tests, latest-main replay, lifecycle race review, and pull request preparation. The human contributor reviewed the contribution and accepts responsibility for it.

All substantive commits include `Generated-by: Codex`; retain the trailer in the final squash commit.

## Checklist

- [x] No


apps/desktop/src/renderer/styles/workbar/shell.css
```diff
@@ -0,0 +1,311 @@
...
+  z-index: 1;
+  overflow: hidden;
+  background: var(--agents-content-area-bg);
+  border-radius: var(--radius-modal);
+  box-sizing: border-box;
+}
+
+.maka-session-workbar-panel[data-overlay][data-placement="right"] {
+  grid-area: right;
+  width: var(--maka-session-workbar-width, 480px);
```
Astro-Han, line 107: [P2] Thanks for preserving the overlay width rule during the CSS move. The narrow-layout E2E currently writes `--maka-session-workbar-width` on `.maka-session-workbar`, but this overlay is its sibling and reads the variable from their shared `.maka-workbar-layout-vars` ancestor. Its second resize therefore still targets the Workbar frame, so the side-chat assertion can pass at the default width without exercising the reported 320px case. Could we set the variable on the shared layout owner, or resize the overlay directly, and assert its bounding box is actually 320px before checking the send button?

Astro-Han (approved):
Thanks for extracting Workbar into a coherent feature slice. The production boundary looks clean: topology and persistence have one owner, Desktop IO stays behind the existing adapter seam, and the Terminal and side-chat lifecycle paths retain their stale-result and disposal fencing.

I found no blocking production issue. I left one non-blocking P2 inline because the narrow side-chat regression test currently does not resize the overlay it intends to exercise.

AI-assisted review disclosure: Codex coordinated independent architecture, lifecycle, and UI/test review passes against exact head `13d9b75cb3d0ed4d2bd44e4e4f3c16d837f70337`; I reviewed the consolidated evidence and approve this exact head.

apps/desktop/src/renderer/styles/workbar/shell.css
chihumyum: Fixed, thanks @Astro-Han

Astro-Han (commented):
Thanks for simplifying the Workbar into a real feature slice. The latest layout regression is fixed, the current-head test is green, and two independent passes found the extraction behavior-preserving across the major state, Side Chat, and Terminal boundaries. One recoverable Terminal cleanup edge remains in the extracted owner, noted inline.

AI-assisted review disclosure: OpenAI Codex coordinated two independent exact-head reviews. I verified the retained failure path, resolved prior thread, current-head CI, mergeability, and review state, and I made the final review decision.

apps/desktop/src/renderer/features/workbar/controller/use-workbar-controller.ts
```diff
@@ -0,0 +1,707 @@
...
+    new Map<string, { sessionId: string; ref: string }>(),
+  );
+
+  const stopTerminal = useCallback(
+    (sessionId: string, ref: string) => {
+      const key = terminalResourceKey(sessionId, ref);
+      if (stoppedTerminalKeysRef.current.has(key)) return;
+      stoppedTerminalKeysRef.current.add(key);
+      ownedTerminalResourcesRef.current.delete(key);
+      void terminal.stop({ sessionId, ref }).catch(() => undefined);
```
Astro-Han: [P2] This marks the resource stopped and removes ownership before `terminal.stop()` succeeds, then swallows rejection. On a reasonable Host disconnect/IPC failure, the remote shell can keep running while the tab and only cleanup ownership are gone; later cleanup will not retry because the key is already in `stoppedTerminalKeysRef`. Could we keep a pending-stop ownership entry until success (or hand it to a main-process retry authority), and add a stop-rejection test that proves cleanup remains retryable?

chihumyum: Fixed in 9211905ca. Terminal cleanup now keeps ownership while stop is in flight, records completion only after success, and clears the in-flight marker on rejection so a later Session cleanup can retry. Added a controller regression test covering first-stop rejection followed by a successful retry. I also rebased onto upstream/main f83b469b5 and resolved the CSS move conflict while preserving the upstream chat-radius behavior. Local verification passed: full build, full typecheck, 36 directed tests, and 6/6 Electron Workbar E2E. @Astro-Han, could you please take another look?

Astro-Han (commented):
Thanks for taking this on — 83 files and +5813/-2957 of pure relocation is a lot of careful work, and the slice boundary you landed on reads well.

Reviewed at exact head `08aa1007b842e207ebf8c647d5382944b70f3c99` against base `f2722fd143a575bb18e54ba30985fca16e2a7e2c`. **No blocking findings.**

For a refactor PR the only claim that matters is "this is just a move", so that is what we checked:

- **Old paths are actually gone.** `session-workbar-layout.ts` is deleted, `use-shell-layout.ts` drops 224 lines, `app-shell.tsx` drops 596, `app-shell-effects.ts` drops 60. No second Workbar authority survives alongside the new slice.
- **CSS relocation is cascade-equivalent.** 1416 lines leave `chat-detail.css` / `quote-side-panel.css` and 1418 arrive in `styles/workbar/*.css`. The six new `workbar/*` imports now sit *before* `chat-detail.css` rather than inside it, which would matter if the 26 lines left in `chat-detail.css` shared selectors with the moved rules — they do not (what remains is `.maka-turn[data-search-highlight]` plus comments). `side-chat.css` is imported immediately after `quote-side-panel.css`, preserving its position relative to everything downstream.
- **`@xterm/xterm/css/xterm.css` moved from two component-level TS imports to a single top-level `@import`.** It is unlayered in both forms, so its precedence over the `layer(components)` rules that style `.xterm` is unchanged. The only difference is that it is now always loaded instead of loaded on first terminal mount, which is not a behaviour change worth flagging in a desktop app.
- **Failure paths were preserved, not simplified away.** `stopTerminal` marks a terminal stopped and releases ownership only inside `.then()`; a rejected stop keeps ownership so a later cleanup can retry. Terminal registration precedes tab commit, stale generations stop, and the Side Chat path keeps its dispose fence after each `await`.

**Not merge-ready yet, for reasons outside the code:** the branch is `CONFLICTING` against `main`, and there are no checks on this head at all — `statusCheckRollup` is empty. The existing `APPROVED` decision belongs to an older head; this repository does not dismiss stale reviews, so that badge is not evidence anyone has read the current code. Please rebase and let CI run; given the size of this diff, a conflict resolution is exactly where an accidental behaviour change would slip in, so we will re-check the moved logic on the new head rather than carry this result forward.

---

Astro-Han (approved):
Thanks — and thanks for the patience across four rounds. Re-reviewed at exact head `624c4c643f316aa05096c69aab19f3e7e50f6c04`, freshly pulled. `test` is green and the PR is `MERGEABLE / CLEAN`.

Worth stating plainly: our earlier APPROVE sits on `13d9b75c`, five heads back, so it didn't carry. This is a fresh pass over the current code.

**Both of my earlier P2s are resolved:**

- **Workbar width variable** — `--maka-session-workbar-width` now lives on the shared parent `.maka-workbar-layout-vars` rather than a sibling, and the E2E sets 320px there and asserts the overlay actually renders at ~320. That closes it at the cascade rather than by adjusting the selector.
- **Terminal stop ownership** — `stopTerminal` only releases ownership once stop succeeds, with in-flight de-duplication and a `finally` that clears the stopping marker, so a failed stop leaves the resource for the next cleanup to retry. Releasing on the optimistic path was the actual bug; this fixes the ordering rather than adding a retry on top.

**On the size.** 6396 added lines invites the question of whether a refactor is carrying passengers, so I checked the split: roughly 1939 lines are tests, E2E and stories, plus ~128 of docs; the remainder is tool migration, service ports and the controller. The production entry consumes only the feature's public entry, with no Desktop global bridge reaching inside — which is the property that makes this a boundary extraction rather than a file shuffle. I found no unrelated product change riding along.

**Behavioural equivalence** is where I spent the most effort, since that's what a move of this size can quietly break. Session switching, StrictMode replay, late terminal starts, Side Chat collapse and source switching, and out-of-order browser selections all still have their guards and tests. Topology and resource lifetime moved out of AppShell without leaving a second reducer or a competing authority behind.

**No findings.** Approving.

---

PR #3448, opened 2026-08-21
feat(desktop): declare one thinking level across a relay's models at once

### LINKED ISSUE #3447: Custom relay: thinking levels can only be declared one model at a time
## Problem

设置 → 模型 → a custom relay → 能力 declares capabilities one model at a time. Thinking levels are a `DropdownMenu` of checkboxes per row, and there is no other way to set them.

A relay is normally one gateway in front of one model family, and the family shares a `reasoning_effort` vocabulary — the fact being declared is about the *relay*, not about each model separately. Declaring `high` across eight enabled models is eight menus opened and eight ticks, repeated whenever the catalog grows. The section's other declarations (vision, context window, fast mode) are genuinely per model; thinking levels usually are not.

## What it should do

One control above the rows that applies a level to every enabled model, writing into the same local draft the rows use so it still commits through 保存能力声明.

Two questions the design has to answer, recorded here because they are the whole of the behaviour:

- **Level, or level set?** Applying a whole set replaces what each row declares and silently discards anything a row already had. Applying one level at a time (tick adds it everywhere, untick removes it everywhere) leaves the other levels alone, which is what a control whose only job is to save clicks should do.
- **How does partial coverage read?** A checkbox is binary and 3-of-5 is neither state. Ticking at partial coverage inverts the next click's meaning — it would take the level away from the rows that have it, which is the opposite of what was just asked for.

## Scope

The 思考档位 row is relay-only (`reasoning_effort` is a wire field only the OpenAI-compatible relays accept; the catalog codec refuses to persist a declared level elsewhere). A bulk control inherits that gate, and is only worth showing when there is repetition to save — two or more rows.

Vision, context window, and fast mode are out of scope: those are per-model facts, and batching them would flatten real differences.

Joob1n:
## Summary

设置 → 模型 → a custom relay → 能力 declared thinking levels one model at a time. A relay normally fronts one model family sharing a `reasoning_effort` vocabulary, so `high` across eight enabled models was eight menus opened and eight ticks — for one fact about the relay.

The section now opens with one control that writes into every enabled model's row.

**Per level, not per level set.** Ticking `high` adds it everywhere; unticking removes it everywhere; every other level on every row is left exactly as the user left it. Replacing the whole set would have been one fewer concept and would silently discard what a row already declared — the destructive reading of a control whose only job is to save clicks.

**The box ticks only at full coverage.** A level 3 of 5 models declare shows unticked with `3/5 个模型` beside it, so the next click means "give it to everyone" rather than "take it from the rows that have it". The count is the only thing separating partial coverage from none, since a checkbox has no third state here.

The control shares the 思考档位 row's relay gate — `reasoning_effort` is a wire field only the OpenAI-compatible relays accept — and appears only from two rows up, where there is repetition to save. Edits land in the same local draft the rows use and commit through the same 保存能力声明. Vision, context window, and fast mode stay per-row: those are per-model facts.

Fixes #3447

## Review focus

**The bulk menu offers only the declarable vocabulary.** `off` is a disable wire (`reasoning_effort: 'none'`), not an intensity tier, and no generic relay is presumed to honor it. It cannot reach a draft anyway — the seed sanitizes through `normalizeRelayModelProfiles` — but spreading one across a whole table is the specific thing worth not being able to do by accident.

**One state update, not a loop of per-model setters.** A bulk tick is a single gesture; committing it in N steps would let a re-render land mid-way and paint a half-applied table. The rule for what an emptied declaration collapses to now lives in one place (`relayProfileWithThinkingLevels`) so the row setter and the bulk path cannot drift on it.

## Also in this PR: an accessibility fix the new story exposed

`上下文窗口（tokens）` was the one capability control without a model in its accessible name — vision, thinking, and fast mode all carry `— ${modelId}`. Every row therefore exposed an identically-named spinbutton, and the AX audit's `ambiguous_actionable` rule fires on it:

```
[product-settings-providers--relay-connection-detail] AX audit failed:
  ambiguous_actionable — role=spinbutton name=上下文窗口（tokens）
  scope=region:能力 > region:House Relay count=4
```

This is a defect on `main`, reachable by anyone with a relay and two or more enabled models — the ordinary case. It had no story: the only existing multi-model capability story is a non-relay whose rows are filtered down to one. Fixed by giving the field the same per-model name as its three siblings.

## Verification

**Behaviour, on the new story** (`product-settings-providers--relay-connection-detail`, a relay with four enabled models, one of which already declares `low` + `high`):

| step | result |
| --- | --- |
| open 批量设置思考档位 | `low 1/4 个模型`, `high 1/4 个模型`, rest `全部未声明` |
| tick `high` | `high` → `4/4 个模型`, ticked; `low` still `1/4`; rows read 已选择 2/1/1/1 个 |
| 保存能力声明 | commits, button disarms; reopening reads `high 4/4` from the saved table |
| untick `high` | `high` → `全部未声明` everywhere; `deepseek-r2` drops to 已选择 1 个, keeping its `low` |

The last row is the non-destructive claim: the model that had two levels kept the one the bulk edit was not about.

**Checks run locally:**

- `@maka/desktop` typecheck (preload + main + renderer + storybook) — clean
- `biome lint apps/desktop/src apps/desktop/stories` — 719 files, clean
- `@maka/desktop` suite — **1048/1048 pass**, including 14 new
- `build-storybook` + `smoke:storybook` — **161 stories pass**, AX audit included. It failed before the context-window fix, on exactly the story this PR adds.

**Not run:** the packaged-artifact and release checks, and the other workspaces' suites — nothing outside `apps/desktop` is touched.

## AI use

Select exactly one:

Tool(s) and scope: Claude Opus 5 (Claude Code) — wrote the implementation, tests, and story; ran the checks above. Reviewed and directed by me. `Generated-by` trailer is on the commit.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/src/renderer/settings/provider-connection-detail.tsx
+++ b/apps/desktop/src/renderer/settings/provider-connection-detail.tsx
@@ -614,6 +614,18 @@ function ConnectionDetailInner(props: ConnectionDetailProps) {
                            present as the same empty box. */
                         description={copy.thinkingBulkCoverage(state.declaredCount, state.total)}
                         aria-label={`${copy.thinkingBulk} ${state.level}`}
+                        /* The item's `description` is visible text only — the
+                           component does not wire it to `aria-describedby`,
+                           and this item's own `aria-label` replaces the name
+                           the description would otherwise have joined. Without
+                           this, "1/4 个模型" and "全部未声明" both reach a screen
+                           reader as an unchecked box with the same name, which
+                           is exactly the partial state the count exists to
+                           show. */
+                        aria-description={copy.thinkingBulkCoverage(
+                          state.declaredCount,
+                          state.total,
+                        )}
                         value={state.checked}
                         onChange={(checked) => {
                           setDraftThinkingLevelForAll(capabilityModelIds, state.level, checked);
--- a/apps/desktop/src/renderer/settings/provider-connection-detail.tsx
+++ b/apps/desktop/src/renderer/settings/provider-connection-detail.tsx
@@ -58,6 +58,7 @@ import {
   savedRequestHeaderDrafts,
   type RequestHeaderDraft,
 } from './request-customization-editor';
+import { bulkThinkingLevelStates } from './relay-thinking-bulk';
 
 export function ConnectionDetail(props: ConnectionDetailProps) {
   const defaults = PROVIDER_DEFAULTS[props.connection.providerType];
@@ -155,6 +156,7 @@ function ConnectionDetailInner(props: ConnectionDetailProps) {
     relayProfileDraft,
     hasRelayProfileChanges,
     setDraftThinkingLevels,
+    setDraftThinkingLevelForAll,
     setDraftVision,
     setDraftContextWindow,
     setDraftServiceTier,
@@ -186,6 +188,10 @@ function ConnectionDetailInner(props: ConnectionDetailProps) {
       !hasModelMetadata(connection.providerType, modelId),
   );
   const showsCapabilities = capabilityModelIds.length > 0;
+  // The bulk control shares the 思考档位 row's relay gate — it edits exactly
+  // that row — and needs repetition to be worth a control at all: with one
+  // row it would be a second widget doing what the row under it already does.
+  const showsThinkingBulk = isRelay && capabilityModelIds.length > 1;
   // One row is a form at a time, the way the settings-sidebar template does it.
   // Opening a row discards the other's draft: leaving an abandoned draft in
   // state meant it reappeared when the user came back to that row, and — until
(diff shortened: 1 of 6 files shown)


apps/desktop/src/renderer/settings/provider-connection-detail.tsx
```diff
@@ -568,6 +574,56 @@ function ConnectionDetailInner(props: ConnectionDetailProps) {
...
+                      DECLARABLE_RELAY_THINKING_LEVELS,
+                    ).map((state) => (
+                      <DropdownMenuCheckboxItem
+                        key={state.level}
+                        label={state.level}
+                        /* The box only ticks at full coverage, so the count
+                           is the sole place partial coverage is legible —
+                           without it "3 of 5 declare high" and "none do"
+                           present as the same empty box. */
+                        description={copy.thinkingBulkCoverage(state.declaredCount, state.total)}
```
Astro-Han, line 615: **[P2] Expose partial model coverage to assistive technology**

Thanks for showing the partial `n/total` state visually—it makes the batch behavior much easier to understand. Because this item also supplies an `aria-label`, however, the screen reader does not receive that visible description and cannot distinguish "none of the models declare this level" from "some models already declare it."

Could we expose `thinkingBulkCoverage(...)` through `aria-description` or an `aria-describedby` relationship? A focused accessibility assertion for the partial state would be enough; the existing pure-function coverage does not need to be expanded further.

Astro-Han (approved):
Thanks for making the batch operation reuse the existing relay-profile draft and save path. The per-level add/remove semantics are clear, partial coverage is preserved, other thinking levels are not overwritten, and the focused tests cover the important state transitions and edge cases.

I'm approving this exact head with one non-blocking accessibility P2 below.

AI-assisted review disclosure: OpenAI Codex coordinated two independent exact-head review passes. I verified the form-state and persistence path, batch semantics, accessibility behavior, tests, exact head, and CI state, and I independently made the approval decision.

Joob1n:
Addressed in `bb06b8c99`. This one was a real defect and I had not tested for it.

**Read straight off the rendered story before the fix:**

```
low      aria-label="批量设置思考档位 low"      aria-checked=false
minimal  aria-label="批量设置思考档位 minimal"  aria-checked=false
```

`low` is declared by one of the four models, `minimal` by none — and to a screen reader they were the same unchecked box with the same name. The `n/total` I added specifically to make partial coverage legible was the one part that did not survive, because the item's own `aria-label` replaces the name the visible description would have joined, and the component does not wire `description` to `aria-describedby` (checked: `aria-describedby` is `null` on every item).

The item now carries `aria-description`. Confirmed through CDP against the real Chromium accessibility tree, not just the DOM attribute:

```
批量设置思考档位 low      description="1/4 个模型"
批量设置思考档位 minimal  description="全部未声明"
```

**The assertion.** It lives in the story's own `play` — the storybook smoke already runs `play` before reading the AX tree, so this is checked on every CI run that builds the catalog, without adding an e2e that would need a configured relay. It opens the batch menu and pins both states:

```ts
await expect(partial).toHaveAttribute('aria-checked', 'false');
await expect(none).toHaveAttribute('aria-checked', 'false');
await expect(partial).toHaveAttribute('aria-description', '1/4 个模型');
await expect(none).toHaveAttribute('aria-description', '全部未声明');
```

Both unchecked is the point: coverage is the only thing separating them. Removing `aria-description` fails that story — verified by removing it:

```
✗ product-settings-providers--relay-connection-detail
  expect(element).toHaveAttribute("aria-description", "1/4 个模型")
```

Per your note, the pure-function coverage is unchanged.

## Verification

- `@maka/desktop` suite: **1048/1048**
- `biome lint apps/desktop/src apps/desktop/stories`: 719 files, clean
- `build-storybook` + `smoke:storybook`: **161 stories**, including the new assertion

PR #3450, opened 2026-08-21
fix(desktop): keep Settings stable during Runtime Host refresh

chihumyum:
## Summary

- Keep last-ready masked Settings snapshots in renderer memory so reopening Settings does not flash a warning or collapse Runtime Host-owned rows during revalidation.
- Bind Runtime Host settings, connections, and mutation authority to the selected Host key and lifecycle epoch, so reconnect or same-key replacement cannot retain stale write authority.
- Keep Desktop-owned preferences available while Host-owned controls remain visible but inert until the current Runtime Host generation is verified.
- Retire one-shot provider catalog navigation only after `ProvidersPanel` consumes it, and visibly mute cached Host pages while their authority fence is active.
- Preserve error and unavailable feedback without changing IPC contracts or persistent storage.

## Behaviour

Before fixing, the yellow banner flashes each time the settings page is opened.

https://github.com/user-attachments/assets/d8cb6212-d0f5-447c-9814-3caa16df68f8

After fixing, no yellow banner was ever shown.

https://github.com/user-attachments/assets/054a9905-43ea-42ac-90ce-9fe0a8d6e394

## Verification

- `npm run lint`
- `npm run format:check`
- `npm run build`
- `npm run typecheck`
- `npm --workspace @maka/desktop test` — 1118/1118 passed
- `npm --workspace @maka/desktop run smoke:storybook` — 172/172 passed
- `npm run astryx:surface-inventory` — 191 files, 1 documented exclusion
- Settings cases in the full Electron run — 4/4 passed
- `npm --workspace @maka/desktop run e2e` — 41 passed, 1 skipped

## Review focus

- Runtime Host epoch changes and stale read/write fencing
- masked snapshot caching and secret lifetime
- mixed ownership boundaries on General, Projects, and Data
- one-shot provider catalog intent ownership across warm-cache loading

## AI use

Select exactly one:

Tool(s) and scope: Codex — implementation, race analysis, unit/Storybook/Electron
E2E coverage, regression verification, and pull request preparation. The human
contributor reviewed the contribution and accepts responsibility for it.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/src/renderer/settings/projects-settings-page.tsx
+++ b/apps/desktop/src/renderer/settings/projects-settings-page.tsx
@@ -194,6 +194,12 @@ export function ProjectsSettingsPage(props: {
     await props.onUpdate({ projects: { defaultProjectId: projectId } });
   }
 
+  // `runtimeHostStatus` describes the selected target's current read/feedback
+  // state; it is not the write-authority predicate. A same-generation refresh
+  // may fail while the already verified target remains safe to use. Every
+  // project read and mutation is therefore fenced by
+  // `runtimeHostTargetVerified`, with the Host-owned subtree below providing
+  // the matching inert, busy, and muted presentation while authority is absent.
   if (!host) {
     return (
       <SettingsPage as="section" aria-label={copy.section}>
--- a/apps/desktop/src/renderer/settings/projects-settings-page.tsx
+++ b/apps/desktop/src/renderer/settings/projects-settings-page.tsx
@@ -43,6 +43,7 @@ import { useKeyedActionGuard } from './use-action-guard';
 import { useOptionalRuntimeHostSettingsTarget } from './runtime-host-settings-target.js';
 import { getSettingsSharedCopy } from '../locales/settings-shared-copy.js';
 import { RemoteProjectDirectoryDialog } from '../remote-project-directory-dialog.js';
+import { RuntimeHostInteractionBoundary } from './runtime-host-interaction-boundary.js';
 
 const NO_PROJECT_CAPABILITIES: DesktopProjectCapabilities = {
   chooseClientDirectory: false,
@@ -69,6 +70,8 @@ const NO_PROJECT_CAPABILITIES: DesktopProjectCapabilities = {
 export function ProjectsSettingsPage(props: {
   settings: AppSettings;
   runtimeHostStatus: 'loading' | 'ready' | 'unavailable' | 'error';
+  runtimeHostTargetVerified: boolean;
+  runtimeHostErrorMessage?: string;
   onUpdate(
     patch: Parameters<typeof window.maka.settings.update>[0],
   ): Promise<UpdateAppSettingsResult>;
@@ -93,17 +96,18 @@ export function ProjectsSettingsPage(props: {
   const reloadGeneration = useRef(0);
 
   const reload = useCallback(async () => {
-    if (!host) return;
+    if (!host || !props.runtimeHostTargetVerified) return;
     const generation = ++reloadGeneration.current;
     const snapshot = await window.maka.projects.getSnapshot(undefined, host);
     if (mountedRef.current && generation === reloadGeneration.current) {
       setProjects([...snapshot.projects]);
       setCapabilities(snapshot.capabilities);
     }
-  }, [host, mountedRef]);
+  }, [host, mountedRef, props.runtimeHostTargetVerified]);
 
   useEffect(() => {
-    if (!host) {
+    if (!host || !props.runtimeHostTargetVerified) {
+      reloadGeneration.current += 1;
       setProjects([]);
       setCapabilities(NO_PROJECT_CAPABILITIES);
       setHomePath(undefined);
@@ -151,6 +155,7 @@ export function ProjectsSettingsPage(props: {
     action: () => Promise<void>,
     failure: string,
   ) {
+    if (!props.runtimeHostTargetVerified) return;
     const release = actionGuard.begin(key);
     if (!release) return;
     try {
@@ -185,6 +190,7 @@ export function ProjectsSettingsPage(props: {
   }
 
   async function setDefault(projectId: string | undefined) {
+    if (!props.runtimeHostTargetVerified) return;
     await props.onUpdate({ projects: { defaultProjectId: projectId } });
   }
 
@@ -192,51 +198,75 @@ export function ProjectsSettingsPage(props: {
     return (
       <SettingsPage as="section" aria-label={copy.section}>
         <RuntimeHostProfilesSection onRemoteHostAdded={props.onRemoteHostAdded} />
-        <Banner
-          status={props.runtimeHostStatus === 'error' ? 'error' : 'warning'}
-          title={props.runtimeHostStatus === 'loading'
-            ? sharedCopy.loading
-            : sharedCopy.runtimeHostUnavailable}
-          endContent={props.runtimeHostStatus === 'error' ? (
-            <Button
-              variant="secondary"
-              size="sm"
-              label={sharedCopy.retry}
-              onClick={() => void props.onRetryRuntimeHost()}
-            />
-          ) : undefined}
-        />
+        {props.runtimeHostStatus !== 'loading' ? (
+          <Banner
+            status={props.runtimeHostStatus === 'error' ? 'error' : 'warning'}
+            title={props.runtimeHostStatus === 'error'
+              ? sharedCopy.settingsLoadFailed
+              : sharedCopy.runtimeHostUnavailable}
+            description={props.runtimeHostStatus === 'error'
+              ? props.runtimeHostErrorMessage
+              : undefined}
+            endContent={props.runtimeHostStatus === 'error' ? (
+              <Button
+                variant="secondary"
+                size="sm"
+                label={sharedCopy.retry}
+                onClick={() => void props.onRetryRuntimeHost()}
+              />
+            ) : undefined}
+          />
+        ) : null}
       </SettingsPage>
     );
   }
   return (
     <SettingsPage as="section" aria-label={copy.section}>
       <RuntimeHostProfilesSection onRemoteHostAdded={props.onRemoteHostAdded} />
-      {/* No section title: the page header already says 项目, and repeating it
-          straight above the rows is the same duplicate-heading noise we
-          removed from the skills page. The rule this page exists for lives in
-          the page subtitle; the section keeps only its action. */}
-      <SettingsSection
-        description={
-          defaultProjectId !== undefined && !defaultResolves
-            ? `${copy.sectionHelp} ${copy.defaultUnavailable}`
-            : copy.sectionHelp
-        }
-        action={capabilities.chooseClientDirectory || capabilities.chooseHostDirectory ? (
-          <Button
-            ref={directoryPickerTriggerRef}
-            variant="secondary"
-            size="sm"
-            label={copy.addProject}
-            clickAction={capabilities.chooseHostDirectory
-              ? () => setDirectoryPickerOpen(true)
-              : async () => {
-                  const result = await window.maka.projects.add(host);
-                  if (result.ok) await reload();
-                }}
-          />
-        ) : undefined}
-      >
+      {props.runtimeHostStatus === 'error' ? (
+        <Banner
+          status="error"
+          title={sharedCopy.settingsLoadFailed}
+          description={props.runtimeHostErrorMessage}
+          endContent={(
+            <Button
+              variant="secondary"
+              size="sm"
+              label={sharedCopy.retry}
+              onClick={() => void props.onRetryRuntimeHost()}
+            />
+          )}
+        />
+      ) : null}
+      <RuntimeHostInteractionBoundary isInteractive={props.runtimeHostTargetVerified}>
+        {/* No section title: the page header already says 项目, and repeating it
+            straight above the rows is the same duplicate-heading noise we
+            removed from the skills page. The rule this page exists for lives in
+            the page subtitle; the section keeps only its action. */}
+        <SettingsSection
+          description={
+            defaultProjectId !== undefined && !defaultResolves
+              ? `${copy.sectionHelp} ${copy.defaultUnavailable}`
+              : copy.sectionHelp
+          }
+          action={capabilities.chooseClientDirectory || capabilities.chooseHostDirectory ? (
+            <Button
+              ref={directoryPickerTriggerRef}
+              variant="secondary"
+              size="sm"
+              label={copy.addProject}
+              clickAction={capabilities.chooseHostDirectory
+                ? () => {
+                    if (props.runtimeHostTargetVerified) setDirectoryPickerOpen(true);
+                  }
+                : async () => {
+                    if (!props.runtimeHostTargetVerified) return;
+                    const result = await window.maka.projects.add(host);
+                    if (result.ok) await reload();
+                  }}
+            />
+          ) : undefined}
+        >
         {listed.length === 0 ? (
           <EmptyState icon={<FolderOpen size={ICON_SIZE.empty} />} title={copy.emptyTitle} description={copy.emptyBody} />
         ) : (
@@ -334,7 +364,7 @@ export function ProjectsSettingsPage(props: {
                                     cancelLabel: copy.removeCancel,
                                     destructive: true,
                                   });
-                                  if (!ok) return;
+                                  if (!ok || !mountedRef.current) return;
                                   await window.maka.projects.archive(project.id, host);
                                   // Removing the default leaves the preference
                                   // pointing at nothing; clear it in the same
@@ -447,16 +477,17 @@ export function ProjectsSettingsPage(props: {
             })}
           </List>)
         )}
-      </SettingsSection>
-      <RemoteProjectDirectoryDialog
-        host={directoryPickerOpen ? host : undefined}
-        returnFocusTo={directoryPickerTriggerRef.current}
-        onClose={() => setDirectoryPickerOpen(false)}
-        onRegistered={() => {
-          setDirectoryPickerOpen(false);
-          void reload();
-        }}
-      />
+        </SettingsSection>
+        <RemoteProjectDirectoryDialog
+          host={directoryPickerOpen && props.runtimeHostTargetVerified ? host : undefined}
+          returnFocusTo={directoryPickerTriggerRef.current}
+          onClose={() => setDirectoryPickerOpen(false)}
+          onRegistered={() => {
+            setDirectoryPickerOpen(false);
+            void reload();
+          }}
+        />
+      </RuntimeHostInteractionBoundary>
     </SettingsPage>
   );
 }
--- a/apps/desktop/src/renderer/settings/runtime-host-interaction-boundary.tsx
+++ b/apps/desktop/src/renderer/settings/runtime-host-interaction-boundary.tsx
@@ -1,0 +1,42 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import type { ReactNode } from 'react';
+
+/**
+ * Keeps a last-known Runtime Host snapshot visible while preventing writes
+ * until the current Settings mount has verified that Host authority.
+ *
+ * `display: contents` makes this an interaction boundary only; it must not
+ * become a layout owner for the settings page it protects.
+ */
+export function RuntimeHostInteractionBoundary(props: {
+  isInteractive: boolean;
+  children: ReactNode;
+}) {
+  return (
+    <div
+      className="settingsRuntimeHostInteractionBoundary"
+      inert={props.isInteractive ? undefined : true}
+      aria-busy={props.isInteractive ? undefined : true}
+    >
+      {props.children}
+    </div>
+  );
+}
--- a/apps/desktop/src/renderer/settings/settings-resource-state.ts
+++ b/apps/desktop/src/renderer/settings/settings-resource-state.ts
@@ -1,0 +1,123 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+export type SettingsResourcePhase = 'idle' | 'loading' | 'ready' | 'error';
+
+/**
+ * A resource can keep a last-ready snapshot while a newer read is in flight
+ * or has failed. Keeping availability separate from the request phase lets
+ * Settings revalidate without replacing known values with defaults.
+ */
+export interface SettingsResourceState<T> {
+  readonly key?: string;
+  readonly snapshot?: T;
+  readonly phase: SettingsResourcePhase;
+  /** The current modal has confirmed this key with a successful read. */
+  readonly isVerified: boolean;
+  readonly message?: string;
+}
+
+export interface SettingsResourceStatus {
+  readonly phase: SettingsResourcePhase;
+  readonly hasSnapshot: boolean;
+  readonly isVerified: boolean;
+  readonly message?: string;
+}
+
+export function createSettingsResourceState<T>(
+  key?: string,
+  snapshot?: T,
+): SettingsResourceState<T> {
+  return {
+    key,
+    snapshot,
+    phase: snapshot === undefined ? 'idle' : 'loading',
+    isVerified: false,
+  };
+}
+
+export function beginSettingsResourceLoad<T>(
+  current: SettingsResourceState<T>,
+  key: string,
+  cachedSnapshot?: T,
+): SettingsResourceState<T> {
+  return {
+    key,
+    snapshot: current.key === key ? current.snapshot ?? cachedSnapshot : cachedSnapshot,
+    phase: 'loading',
+    isVerified: current.key === key && current.isVerified,
+  };
+}
+
+export function completeSettingsResourceLoad<T>(
+  key: string,
+  snapshot: T,
+): SettingsResourceState<T> {
+  return { key, snapshot, phase: 'ready', isVerified: true };
+}
+
+export function failSettingsResourceLoad<T>(
+  current: SettingsResourceState<T>,
+  key: string,
+  message: string,
+  cachedSnapshot?: T,
+): SettingsResourceState<T> {
+  return {
+    key,
+    snapshot: current.key === key ? current.snapshot ?? cachedSnapshot : cachedSnapshot,
+    phase: 'error',
+    isVerified: current.key === key && current.isVerified,
+    message,
+  };
+}
+
+export function settingsResourceSnapshot<T>(
+  state: SettingsResourceState<T>,
+  key: string | undefined,
+): T | undefined {
+  return key !== undefined && state.key === key ? state.snapshot : undefined;
+}
+
+export function settingsResourceStatus<T>(
+  state: SettingsResourceState<T>,
+  key: string | undefined,
+): SettingsResourceStatus {
+  if (key === undefined || state.key !== key) {
+    return { phase: 'idle', hasSnapshot: false, isVerified: false };
+  }
+  return {
+    phase: state.phase,
+    hasSnapshot: state.snapshot !== undefined,
+    isVerified: state.isVerified,
+    message: state.message,
+  };
+}
+
+export function reconcileRuntimeHostProfileSelection(options: {
+  currentProfileId: string | undefined;
+  defaultProfileId: string;
+  enabledProfileIds: readonly string[];
+  preserveCurrentSelection: boolean;
+}): string {
+  return options.preserveCurrentSelection &&
+    options.currentProfileId !== undefined &&
+    options.enabledProfileIds.includes(options.currentProfileId)
+    ? options.currentProfileId
+    : options.defaultProfileId;
+}
--- a/apps/desktop/src/renderer/settings/settings-resource-state.ts
+++ b/apps/desktop/src/renderer/settings/settings-resource-state.ts
@@ -52,6 +52,17 @@ export function createSettingsResourceState<T>(
   };
 }
 
+/**
+ * Keeps the last snapshot visible while revoking the authority established by
+ * an older Runtime Host generation. Ordinary same-generation refreshes use
+ * `beginSettingsResourceLoad` instead so they retain verified SWR semantics.
+ */
+export function invalidateSettingsResourceGeneration<T>(
+  current: SettingsResourceState<T>,
+): SettingsResourceState<T> {
+  return createSettingsResourceState(current.key, current.snapshot);
+}
+
 export function beginSettingsResourceLoad<T>(
   current: SettingsResourceState<T>,
   key: string,
--- a/apps/desktop/src/renderer/settings/settings-surface.tsx
+++ b/apps/desktop/src/renderer/settings/settings-surface.tsx
@@ -17,7 +17,14 @@
  * under the License.
  */
 
-import { useEffect, useMemo, useRef, useState, type RefObject } from 'react';
+import {
+  useEffect,
+  useEffectEvent,
+  useMemo,
+  useRef,
+  useState,
+  type RefObject,
+} from 'react';
 import {
   Badge,
   Button,
@@ -44,6 +51,7 @@ import type {
 } from '@maka/core/settings';
 import type { LlmConnection, ProviderType } from '@maka/core/llm-connections';
 import type {
+  DesktopRuntimeHostProfileChangedEvent,
   DesktopRuntimeHostProfileSnapshot,
   DesktopRuntimeHostRef,
   DesktopSessionSummary,
@@ -96,6 +104,7 @@ import {
   completeSettingsResourceLoad,
   createSettingsResourceState,
   failSettingsResourceLoad,
+  invalidateSettingsResourceGeneration,
   reconcileRuntimeHostProfileSelection,
   settingsResourceSnapshot,
   settingsResourceStatus,
@@ -119,9 +128,19 @@ type RuntimeHostAvailabilityStatus = 'loading' | 'ready' | 'unavailable' | 'erro
 function readyRuntimeHost(
   snapshot: DesktopRuntimeHostProfileSnapshot | undefined,
   profileId: string | undefined,
+  lifecycle?: DesktopRuntimeHostProfileChangedEvent,
 ): DesktopRuntimeHostRef | undefined {
   const entry = snapshot?.entries.find((candidate) => candidate.profile.id === profileId);
   if (entry?.readiness !== 'ready' || !entry.hostId) return undefined;
+  if (
+    lifecycle &&
+    (
+      lifecycle.removed === true ||
+      lifecycle.readiness !== 'ready' ||
+      !lifecycle.hostId ||
+      lifecycle.hostId !== entry.hostId
+    )
+  ) return undefined;
   return { profileId: entry.profile.id, hostId: entry.hostId };
 }
 
@@ -286,6 +305,12 @@ export function SettingsSurface(props: {
   const runtimeHostReloadTicketRef = useRef(0);
   const runtimeHostCatalogHydratedRef = useRef(false);
   const selectedProfileChangedByUserRef = useRef(false);
+  const runtimeHostLifecycleByProfileRef = useRef(
+    new Map<string, DesktopRuntimeHostProfileChangedEvent>(),
+  );
+  const [runtimeHostLifecycleByProfile, setRuntimeHostLifecycleByProfile] = useState(
+    runtimeHostLifecycleByProfileRef.current,
+  );
   const toast = useToast();
 
   const runtimeHosts = settingsResourceSnapshot(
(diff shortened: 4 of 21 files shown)


jackwener (commented):
Automated review of exact head `54651d6b0aeca844f527f75cfe00f4a8d6b6c9fc` against current `main@d62857a8357e9160926726a2a13096bc2dc2b91d`.

The defect remains on main: every Settings mount starts the Runtime Host catalog/settings/connections resources cold, so the mixed-ownership General page briefly renders defaults/loading feedback before Host data arrives. The patch keeps only masked read snapshots in renderer memory, keys Host data by `profileId:hostId`, prunes an old epoch when the catalog changes, and keeps cached Host controls inert until the current mount verifies the target. The separate read/write tickets also prevent a late old-Host result from populating the newly selected Host. The reopen E2E observes the transient DOM, so it would catch the original flash rather than only asserting the settled page. I found no actionable P0-P2 defect; `git diff --check` is clean.

Required conclusions:

1. Optimal for the actual problem: **yes**. The implementation is larger than the symptom, but stable cross-unmount rendering requires a cache, and the masking, epoch key, verification bit, and interaction fence are necessary to avoid turning stale UI into write authority.
2. Production code to delete: the PR replaces the older parallel resource unions/ticket refs; none further identified.
3. Tests to delete/replace: none identified. The helper tests and transient-DOM E2E cover distinct contracts.
4. Deeper refactor: no; keeping the cache outside AppShell avoids lifting Settings-only state into a broader owner.
5. Ready to merge: **not yet**. No hosted `test` check is reported on this exact head, and automated review is not approval.
6. Residual risks/gaps: this changes user-visible loading/error behavior and the security-sensitive Host write boundary across General, Projects, and Data. The supplied tests cover cache epochs and the default General reopen, but independent human review should still exercise non-default Host selection and failure/retry behavior.

This affects user-visible behavior and Runtime Host authority fencing, so `CONTRIBUTING.md` requires independent human judgment on the exact head.

jackwener (commented):
Automated re-review of exact head `43e53f751ba9b604b994d75bf88c8a2781972e08` against `main@f1f4e71a9540a4da23159052c924fee72202e989`.

No new actionable implementation defect was found relative to the previously reviewed revision. `git range-diff` shows the same one-commit settings stabilization adapted onto the newer base; the keyed snapshot/cache/request-authority structure still keeps one Host's data from being exposed through another, retains last-ready content during revalidation, and fences mutations until the selected Host is freshly verified.

The head is not mergeable, however. Current main and this PR both modify `apps/desktop/e2e/settings.spec.ts`, `settings-surface.tsx`, and `nav-sidebar.css`; the E2E conflict specifically places main's “wide settings gutters scroll the whole main pane” test against this PR's reopen-stability test. Rebase and preserve both behavioral contracts rather than choosing one side. Re-run the exact-head check after resolution.

Required conclusions:

1. Optimal for the actual problem: **yes**; the cache remains renderer-memory, keyed by bridge/Host identity, and mutations stay behind fresh authority.
3. Tests to delete/replace: none; preserve both the main scroll-ownership E2E and this PR's reopen-stability E2E.
4. Deeper refactor: no.
5. Ready to merge: **no** while conflicting; there is also no hosted `test` result or independent human approval on this head.
6. Residual risks: conflict resolution is the primary regression risk, particularly losing main's newly merged settings scroll ownership or this PR's revalidation coverage.

This changes user-visible Settings behavior. Independent human review is required; this automated review is not approval.

apps/desktop/src/renderer/settings/settings-surface.tsx
```diff
@@ -284,57 +346,62 @@ export function SettingsSurface(props: {
...
+    selectedRuntimeHostSettingsStatus.message ??
+    selectedRuntimeHostConnectionsStatus.message;
   const loading =
     clientLoading ||
-    (requiresRuntimeHost && runtimeHostContentStatus === 'loading');
+    (requiresRuntimeHost &&
+      !runtimeHostContentReady &&
+      runtimeHostContentStatus === 'loading');
 
   useEffect(() => {
```
Astro-Han: **[P2] Clear the catalog intent only after the child consumes it**

On a warm-cache reopen, outer `loading` can already be false while `ProvidersPanel` is still performing its own initial load. This effect then clears `providerCatalogRequested` before the child reaches its `!loading` effect, so “Browse providers” can reopen Settings on Connections instead of the requested catalog.

Please let `ProvidersPanel` acknowledge when it actually consumes the initial catalog route, and clear the parent intent from that callback. A warm-cache regression test where the child is still loading would cover the real ordering boundary.

apps/desktop/src/renderer/settings/runtime-host-interaction-boundary.tsx
```diff
@@ -0,0 +1,23 @@
...
+ * become a layout owner for the settings page it protects.
+ */
+export function RuntimeHostInteractionBoundary(props: {
+  isInteractive: boolean;
+  children: ReactNode;
+}) {
+  return (
+    <div
+      className="settingsRuntimeHostInteractionBoundary"
+      inert={props.isInteractive ? undefined : true}
```
Astro-Han, line 36: **[P2] Make the revalidation fence visible**

With a cached snapshot, Runtime Host pages render their normal controls while this boundary silently makes them `inert`. During a slow reconnect or remote Host verification, the page therefore looks fully enabled but clicks and keyboard interaction do nothing; `aria-busy` communicates this only to assistive technology.

Please preserve the stable cached geometry but add a visible muted/loading state or small revalidation indicator while interaction is fenced. The important contract is that unavailable interaction must also be perceptible.

Astro-Han (commented):
Thanks — the cache ownership and Runtime Host fencing are thoughtfully structured, and I did not find a cross-Host authority or stale-write correctness issue on exact head `43e53f751ba9b604b994d75bf88c8a2781972e08`. I found two non-blocking user-facing gaps below.

Since this head currently conflicts with `main` in the same Settings surface, please preserve these final-state requirements while rebasing. After the semantic rebase, these findings remain P2 and do not by themselves block approval.

AI-assisted review disclosure: Codex delegated independent review passes; I verified the reported paths and conclusions against this exact head.

chihumyum:
@Astro-Han Thanks for the careful review. I rebased onto the current `main` and addressed both P2 findings in `fa01d917d`:

- The provider-catalog landing intent is now retired only when `ProvidersPanel` actually enters the catalog route via `onInitialCatalogConsumed`. A warm-cache Story delays the child load, verifies that the catalog still opens, and verifies the intent is one-shot after remount.
- The cached Runtime Host page now has a visible muted state while the existing `inert` / `aria-busy` authority fence is active, without changing layout or reintroducing the loading Banner. Storybook asserts the computed opacity as well as the semantic fence.

The semantic rebase also preserves both #3453 scroll ownership and this PR’s reopen-stability contract. Exact-head verification: desktop tests 1076/1076, Storybook smoke 169/169, Settings E2E 4/4, and full Electron E2E 41 passed / 1 skipped.

Could you please take another look when convenient?

apps/desktop/src/renderer/settings/settings-surface.tsx
```diff
@@ -729,19 +823,13 @@ export function SettingsSurface(props: {
                 <LayoutContent padding={6} isScrollable={false}>
                   {loading ? (
```
Astro-Han, line 915: [P1] On a cold/failed Runtime Host read, this outer branch renders only the skeleton or error banner; with a cached snapshot, the later boundary makes the whole `SettingsPageBody` inert. That also blocks Desktop-owned controls such as notification and language settings, even though the PR’s intended contract says they remain usable while only Host-owned controls wait. Please place the availability boundary at the ownership seam: always render the page’s Desktop-owned sections, and gate only the Host-backed fields/sections with their own verified snapshot. That gives one consistent mixed-ownership page instead of three page-level modes.

apps/desktop/src/renderer/settings/settings-resource-state.ts
```diff
@@ -0,0 +1,104 @@
...
+export function beginSettingsResourceLoad<T>(
+  current: SettingsResourceState<T>,
+  key: string,
+  cachedSnapshot?: T,
+): SettingsResourceState<T> {
+  return {
+    key,
+    snapshot: current.key === key ? current.snapshot ?? cachedSnapshot : cachedSnapshot,
+    phase: 'loading',
+    isVerified: current.key === key && current.isVerified,
```
Astro-Han, line 75: [P2] A reload of the same key preserves `isVerified` through both loading and failure, so reconnect/profile refresh can leave stale Host-backed controls writable. Because lifecycle events carry an epoch but this authority is keyed only by `profileId:hostId`, remove/re-add of the same Host can also accept an old mutation response. Please revoke mutation authority when a new Host generation begins and include that generation/epoch in both read and write tickets; cached data may stay visible, but it should not stay writable until that generation is verified.

Astro-Han (commented):
Thanks for the substantial cleanup of the settings cache and intent flow. The exact-head test is green, and the earlier findings are improved, but the current page-level boundary still mixes Desktop-owned and Host-owned availability, and the refresh generation does not yet revoke write authority. I left the two final-state requirements inline.\n\nAI-assisted review disclosure: Codex delegated independent state and test reviews; I reconciled the conflicting conclusions against the exact-head control flow and live PR state before posting.

apps/desktop/src/renderer/settings/settings-surface.tsx
chihumyum: Thanks — I re-traced this path on the new exact head, and the outer Runtime Host boundary does not apply to the mixed pages named here. `general`, `projects`, and `data` are all classified as `mixed`, so `requiresRuntimeHost` is `false`: a cold or failed Host read does not select the outer Host-only skeleton/error/unavailable branches, and the outer `RuntimeHostInteractionBoundary` receives `isInteractive={true}` for those pages.

The ownership seams are inside each mixed page. General gates only Host-backed rows; interface language and completion notifications remain enabled and persist through the client-settings path. Projects renders Runtime Host profile management outside its Host-backed project boundary. Data disables Host-backed path/config actions while keeping local input-history clearing available. I also added explicit General Story regressions asserting that language and notification controls stay enabled and that the mixed-page outer boundary never becomes `inert`, including during same-key/new-epoch revalidation.

Could you point me to a concrete mixed-page state that still enters the outer Host-only branch if I have missed one?

— Codex, posting on behalf of @chihumyum

apps/desktop/src/renderer/settings/settings-resource-state.ts
chihumyum: Agreed — addressed in 1ac77bf98.

Runtime Host request tickets now bind to the composite target `(profileId:hostId, lifecycle epoch, target revision)` for settings reads, connection reads, and settings writes. A selected-profile lifecycle event synchronously changes that target before the catalog refresh starts; same-key/new-epoch replacement and ready-to-reconnecting transitions therefore revoke every outstanding ticket and mark cached resources unverified while keeping their last snapshots visible. A late mutation completion for a superseded target is rejected before renderer/AppShell mirrors can be committed. Ordinary same-generation SWR still retains its verified state.

The surface also retains the latest lifecycle projection per profile, so a background Host that starts reconnecting cannot be revived from a stale ready catalog if the user selects it before refresh completes.

Regression coverage includes same key/new epoch, repeated observation of the same epoch, reconnect with a write in flight, visible-but-unverified cached state, and background reconnect followed by profile selection. Stories verify that Host-owned controls lose authority immediately while completion notifications and interface language remain usable.

— Codex, posting on behalf of @chihumyum

chihumyum:
@Astro-Han Thanks for the follow-up review. I rebased the branch onto `main@f83b469b5`, addressed the generation-fence finding in `1ac77bf98`, and replied inline.

The follow-up binds Runtime Host reads and writes to lifecycle epoch as well as `profileId:hostId`, synchronously revokes old-generation authority on replacement/reconnect, and keeps background-profile lifecycle tombstones from being bypassed by a stale catalog. I also re-traced the mixed-page finding: General, Projects, and Data bypass the outer Host-only branch, and explicit Stories now cover language/notification availability during Host revalidation.

Exact-head local verification is green: lint, format, build, typecheck, both knip workspaces, Astryx inventory (191 files), Desktop tests (1118/1118), Storybook build/render smoke (172/172), and Electron E2E (41 passed, 1 repository-defined skip). The new hosted [CI run](https://github.com/apache/maka/actions/runs/32567801940) is awaiting maintainer approval before its jobs can start.

Could you please re-review the new exact head when convenient?

— Codex, posting on behalf of @chihumyum

apps/desktop/src/renderer/settings/projects-settings-page.tsx
```diff
@@ -185,58 +190,83 @@ export function ProjectsSettingsPage(props: {
...
 
   async function setDefault(projectId: string | undefined) {
+    if (!props.runtimeHostTargetVerified) return;
     await props.onUpdate({ projects: { defaultProjectId: projectId } });
   }
 
   if (!host) {
     return (
       <SettingsPage as="section" aria-label={copy.section}>
         <RuntimeHostProfilesSection onRemoteHostAdded={props.onRemoteHostAdded} />
```
Astro-Han, line 206: **[P2] Of the three `mixed` pages, this is the only one that never consults `runtimeHostStatus`.**

The three pages guard their Host-backed content with three different predicates:

- `general-settings-page.tsx:93` — `host !== undefined && props.runtimeHostStatus === 'ready' && props.connectionsBridge !== undefined && props.testNetworkProxy !== undefined`
- `data-settings-page.tsx:77` — `host !== undefined && props.runtimeHostStatus === 'ready'`
- here — `if (!host) return <banner-only page>`, and nothing downstream looks at `runtimeHostStatus` again

`host` comes from `selectedRuntimeHost`, which `settings-surface.tsx:251-259` returns only when `selectedRuntimeHostEntry.readiness === 'ready'`. So `host` being defined tells you the Host *connection* is ready — it does not tell you the Host *content* this PR added state tracking for has arrived and been verified. Those are the two different facts the PR exists to separate.

Consequence: with a ready connection and `runtimeHostStatus` still `'loading'` or `'error'`, the General and Data pages show the loading/unavailable banner, while Projects renders its full control set with no banner at all — the banner lives inside the `!host` branch, which is not taken. The user gets interactive project controls sitting on content that has not been verified, which is the exact state this change set out to stop rendering as if it were settled.

That this page already accepts `runtimeHostStatus` as a prop and uses it *inside* the `!host` branch (to choose between `sharedCopy.loading` and `sharedCopy.runtimeHostUnavailable`) is what makes this read as an oversight rather than a deliberate scoping decision. If it is deliberate — if Projects genuinely has no Host-content dependency and the prop is only there for the banner copy — then the page arguably should not be `'mixed'` in `SETTINGS_SECTION_SCOPES`, and saying so in a comment would stop the next reader from filing this same finding.

Otherwise the fix is to match Data: compute `runtimeHostAvailable = host !== undefined && props.runtimeHostStatus === 'ready'` and gate on that instead of on `host` alone.

Since three pages now express the same intent three different ways, a shared helper for "is Host-backed content safe to render here" would be worth more than three separate fixes — that way the next mixed page cannot get it wrong a fourth way.

Astro-Han (commented):
Re-reviewed at the new head `280d833a` after the rebase. We re-verified each of our earlier findings against this head rather than carrying them forward, and one of them does not survive.

**Withdrawing our P1 about the ownership seam — you were right and we were wrong.** Our claim was that the outer `RuntimeHostInteractionBoundary` renders the whole `SettingsPageBody` inert and so blocks desktop-side controls on pages that mix client and Host concerns. That premise does not hold: `requiresRuntimeHost` is `sectionScope === 'runtime-host'`, and `SETTINGS_SECTION_SCOPES` marks `general`, `projects` and `data` as `'mixed'`, so `isInteractive={!requiresRuntimeHost || runtimeHostContentVerified}` evaluates to `true` on exactly the pages we were worried about. The inert branch is unreachable there. We should have traced the scope table before filing it.

**Two P2 from the earlier rounds are confirmed fixed.** The write-authority/epoch one: `settings-request-authority.ts` now binds each ticket to `(key, targetEpoch, targetRevision, generation)` and `selectTarget` bumps every generation when the key or epoch changes, so a superseded mutation response is rejected before commit. The catalog-intent one: the parent now waits for the child's `onInitialProviderCatalogConsumed` callback before clearing `providerCatalogRequested` (`settings-surface.tsx:998-1000`).

**One P2 from the earlier round is still open** — the boundary is invisible. `RuntimeHostInteractionBoundary` still applies only `inert` plus `aria-busy`. Assistive technology is informed; a sighted user is not. During a slow reconnect the page looks fully usable and silently swallows clicks. A muted/loading treatment on the wrapper would close it.

**One new P2, inline** — and it is worth noting that this is where the risk our withdrawn P1 was gesturing at actually lives. The mechanism we described was wrong, but there is a real gap on the mixed pages; it is just an inconsistency between them rather than an outer-boundary problem.

Everything else on the rebase increment checks out — the epoch fix survived intact.

---

apps/desktop/src/renderer/settings/projects-settings-page.tsx
chihumyum: Thanks — I re-traced this on exact head `7a332f583`, and the asserted interactive state is already prevented.

Projects uses `runtimeHostTargetVerified` as its Host-authority predicate at every seam: it gates `projects.getSnapshot`, resets page-owned state when authority is revoked, guards row/default/add mutations, fences the remote-directory dialog, and wraps only the Host-owned project subtree in `RuntimeHostInteractionBoundary`. `runtimeHostTargetStatus` is derived from the same verification bit in `SettingsSurface`; a cached `host` with an unverified catalog therefore produces `status=loading` and `runtimeHostTargetVerified=false`. The project read does not start, and that subtree is `inert`, `aria-busy`, and visibly muted. Runtime Host profile management intentionally remains outside the boundary and usable.

The Data predicate quoted above is not present on this head either: Data keeps `host !== undefined` for stable mixed-page rendering and separately gates every Host-backed read/action with `runtimeHostTargetVerified`. General, Projects, and Data cannot share one broader content-ready helper because General additionally owns Settings and Connections resources, while Projects and Data own different page-local reads.

Commit `7a332f583` makes that distinction explicit next to the Projects branch and tightens `ProjectsCachedHostRevalidation` to assert the full contract: `inert`, `aria-busy=true`, computed opacity `0.5`, and no project snapshot read. Storybook render smoke passes 174/174.

The visible-fence concern in the review summary is also already covered by `.settingsRuntimeHostInteractionBoundary[inert] > * { opacity: var(--opacity-disabled) }`; `ModelsCachedHostRevalidation` independently asserts the computed muted opacity. This keeps cached geometry stable without bringing back a loading banner.

Astro-Han (approved):
Thanks — and I owe you a correction before anything else.

Re-reviewed at exact head `7a332f583df031cde86ef5e799291340d978c830`. `test` is green.

**My earlier fence-visibility P2 was wrong, and it was my mistake.** I claimed the boundary left unverified content looking fully actionable. It doesn't: `.settingsRuntimeHostInteractionBoundary[inert] > *` sets `opacity: var(--opacity-disabled)` in `nav-sidebar.css`, and the component already pairs `inert` with `aria-busy`. I reviewed the component and not the stylesheet that goes with it, so I saw two thirds of the mechanism and reported the missing third as absent. It was there the whole time. Sorry for the detour.

The other three all check out on this head:

- **Ownership seam** — `requiresRuntimeHost` keys off `sectionScope === 'runtime-host'`, and the mixed pages resolve `isInteractive` to true, so the seam sits where you said it does.
- **Write authority** — tickets bound to `(key, epoch, revision, generation)`.
- **Catalog intent** — the parent waits on `onInitialProviderCatalogConsumed`.

I also went to check your comment that the project read does not start while unverified, half-expecting to find it only gated on `host`. It isn't — `reload` returns early on `!host || !props.runtimeHostTargetVerified` (`projects-settings-page.tsx:99`), and the same guard repeats at 109, 158 and 193. Your description was accurate and I was wrong to doubt it.

On proportion: 2230 lines is a lot for "keep Settings stable", so I split it — 973 lines are tests and stories, leaving 1257 for the state machine, snapshot cache and boundary. That reads as matched to the problem rather than a refactor riding along.

**No findings.** Approving.

---

Astro-Han:
needs rebase~

chihumyum:
> needs rebase~
Done ~

chihumyum:
@Astro-Han I rebased this branch onto the current `main@1d06330a1d` and resolved the only conflict by regenerating the Astryx surface inventory. The three semantic commits are unchanged under `git range-diff`; the new exact head is `34c52eb25950e95f6182ae6b27f260b14211834a`.

Local exact-head verification is green: lint, format, build, typecheck, Astryx inventory (202 files), both Knip workspaces, and Desktop tests (1221/1221).

The hosted [CI run](https://github.com/apache/maka/actions/runs/32619651179) is waiting for maintainer approval before jobs can start. Could you please approve the run when convenient?

— Codex, posting on behalf of @chihumyum

Astro-Han (approved):
Approving `34c52eb25950e95f6182ae6b27f260b14211834a`. Required `test` is `completed / success` bound to that exact SHA. **No P0–P3.**

This is a re-review at the current head. The prior approval was given on `7a332f58`, and the branch has moved since, so the conclusion was re-derived rather than carried forward.

**Range-diff against the previously reviewed state:** of the three semantic commits, `448298d8f` (fence Settings across Host generations) and `34c52eb25` (clarify Projects revalidation authority) are byte-identical to what was reviewed. `6b6c4fc82` (keep Settings stable during Runtime Host refresh) is byte-identical apart from the regenerated inventory lines. Everything since is the rebase onto `main@1d06330a1`.

**The previously failing check was reproduced independently rather than assumed fixed:** `check-astryx-surface-inventory.mjs` at this exact head returns ok (202 files).

**One thing recorded rather than escalated:** the hosted run skipped `Run Runtime Host tests` via the affected-surface planner. This PR sits on the Settings/runtime-host boundary, but its own diff changes no file inside the runtime-host package, and the earlier rounds covered that surface locally. Flagging it because a planner skip and a genuine pass look the same in the summary view, and the distinction should be visible to whoever merges rather than buried.

The muted/inert boundary styling is unchanged from the reviewed state, so there is no new visual surface here.

Astro-Han (commented):
A maintainer challenged this PR's proportionality directly: *if this is just a flicker / lost-state bug, is this the right shape of fix? Isn't it over-complex — do you need this many modules?* That is a fair question for +2230 lines and five new modules, so it was audited as a first-principles question rather than as a findings pass, at exact head `34c52eb25950e95f6182ae6b27f260b14211834a`.

**The conclusion did not go the way the challenge pointed: the core complexity is root-cause-driven and should not be reduced to "add a skeleton". No new P0–P3.** What follows is the reasoning, including the part that would justify a smaller *shape*.

**First, a framing correction.** About 44% of the added lines are unit tests, E2E and Storybook (356 + 66 + 551). That does not by itself prove proportionality, but "+2230" should not be read as 2230 lines of production logic.

**The problem is two independent failures stacked, not one cosmetic one.**

1. `SettingsModal` is gated by `settingsOpen && (...)` in `app-shell-overlays.tsx:134` — **closing Settings unmounts it entirely.** On reopen the old implementation restarted from defaults and empty resources while the Host's catalog, settings and connections were still in flight, which is what produced the loading banner and the collapsing Host rows.
2. The Host lifecycle has an `epoch`, but the renderer's authority was effectively `profileId:hostId`. When the same profile/host reconnects or is replaced, a stale read or write still *looks* like it belongs to the current key. Swapping `loading` for a skeleton does not change that: the stale response still lands in the new Host generation.

So the minimal closing concept is not "a skeleton". It is: **a generation authority** (`profileId + hostId + epoch`, invalidating outstanding tickets on lifecycle events); **snapshot state that separates `hasSnapshot` from `isVerified`**, so a retained value is never mistaken for current Host authority; **an interaction fence**, so a stale snapshot is visible but Host-backed controls stay inert until re-verified; and — because the modal genuinely unmounts — **a cache that outlives the unmount**.

That last one is worth naming, because it is the module that looks most decorative and is not: the new E2E in `settings.spec.ts` makes "reopening Settings shows no loading banner and does not interrupt Host controls" an explicit acceptance condition. Under that contract a renderer-memory cache keyed on bridge identity is part of the behaviour, not polish.

**Where the challenge does land:** the *module boundaries* are consolidatable even though the *concepts* are not. `settings-request-authority.ts`, `settings-resource-state.ts` and the cache could reasonably collapse into a single surface-owned `useSettingsAuthority` object without changing behaviour; `settings-skeleton.tsx`'s top-level skeleton is UX rather than race repair; and `reconcileRuntimeHostProfileSelection` is a selection policy sitting in a resource-state module with exactly one production caller (`settings-surface.tsx:642`), so it belongs closer to the surface. **None of that is a correctness finding — file count is not a defect** — and trading a module away for the old race back would not be a simplification.

**Two things are product decisions, not code cleanup, and are flagged rather than decided here:** whether a stale-but-inert Host value is acceptable at all (the alternative is an explicit "verifying" state), and whether `SettingsRowSkeleton` must reserve layout for not-yet-known Host rows — the current E2E requires that General never shows a window with no Host controls, so it cannot be removed on aesthetic grounds alone.

Required `test` is `completed / success` bound to this exact SHA. This audit does not block merge.

---
*AI-assisted review. The unmount gate, the epoch-fencing rationale, and the single-caller claim above were each verified against the source at this exact head. Under `CONTRIBUTING.md` §Review this does not count as the required independent human review.*

Astro-Han (commented):
Reviewed at exact head `80f236dea0f7ca99c03bed1a2750896a387ed2cb`. Re-reviewed from scratch: the earlier approval on this PR is bound to an older head, so nobody had looked at the current one.

**Coverage:** where Settings state comes from across a Runtime Host refresh, which user-visible state is rebuilt or discarded, and whether "stays stable" is achieved by caching old values or by deferring re-render. **Not covered:** visual and copy.

**One [P1].**

### [P1] The preserved Models subtree keeps OAuth state from the retired generation

Same-key Host epoch replacement revokes and revalidates the catalog, Settings, and Connections. The Models subtree is intentionally kept mounted so the replacement does not destroy drafts, route, scroll, and focus — that part is a good design choice.

But the OAuth card/account snapshots and the active login controllers inside that subtree **only refresh on mount**. Because the subtree is deliberately never remounted, once the outer fence lifts they carry state belonging to the retired generation:

- stale signed-in status and account email displayed against the current Host;
- Login/Logout actions that are still live and act on the current Host while originating from the retired generation;
- a pending browser authorization that is not generation-cancelled, so an authorization started under the old Host can land after replacement.

This is the failure mode the outer fence exists to prevent — it just does not reach into this subtree, and the exemption that protects the user's drafts also protects the stale credentials view.

The fix should propagate an epoch/revision token **into the OAuth sub-authority** so that sub-tree revalidates on generation change. Remounting the whole Settings tree would fix the staleness and reintroduce exactly the draft/route/scroll/focus loss this PR set out to eliminate, so that is the wrong lever.

### Verification and limits, stated plainly

- Clean root build passed; 62/62 targeted state/OAuth/Host tests passed.
- The Settings Electron end-to-end suite **could not be run** — the reviewing host has no Electron GUI libraries. So the finding above rests on state-ownership tracing plus the targeted suites, not on an end-to-end reproduction.
- There is **no hosted green** for this head: the hosted `test` run was cancelled during dependency install. Not cancelled-and-therefore-fine — simply no terminal result, so the CI gate is unmet independently of the finding.

chihumyum:
@Astro-Han Thanks for the exact-head review. I rebased onto current main@4852d4922 and addressed the OAuth generation P1 in 2f2d5e38d.

The selected Runtime Host lifecycle epoch now flows into a Settings generation context. OAuth card snapshots clear and re-read on generation change, with ticket fencing preventing a retired read from landing later. Only the OAuth login and relogin controllers sit behind the keyed generation boundary, so their existing unmount cleanup cancels any pending browser authorization. The parent Settings and Models subtree stays mounted, preserving route, catalog filter, drafts, scroll, and focus.

Coverage now includes a focused lifecycle test plus Storybook journeys for both stale account replacement and an in-flight authorization cancelled exactly once while the provider setup route survives. Exact-head local verification is green: lint, format, ASF headers, Astryx inventory, full typecheck/build, Desktop tests, both Knip workspaces, Storybook build and 184/184 render smoke, and Settings Electron E2E 5/5.

Hosted CI has just been retriggered and does not have a result yet. Could you please re-review exact head 2f2d5e38d when convenient?

— Codex, posting on behalf of @chihumyum

chihumyum:
@Astro-Han Rebased again onto current main@cd0163a1e. The only semantic conflict was General Settings: I preserved upstream WorkHub as a Desktop-owned, always-available control while retaining this PRs Host-owned workspace-instructions snapshot and placeholder behavior. The other conflict was resolved by regenerating the Astryx inventory. New exact head: e97d1378d.

Range-diff shows the generation-fence and Projects commits unchanged; the stabilization commit differs only by the upstream WorkHub row, and the OAuth commit only by the regenerated inventory total. Exact-head local verification is green: lint, format, ASF headers, Astryx inventory (205 files), full typecheck/build, Desktop tests 1336/1336, both Knip workspaces, Storybook build and 184/184 render smoke, and Settings Electron E2E 5/5.

The PR is mergeable again. Hosted CI has been retriggered and has no result yet. Please review exact head e97d1378d when convenient.

— Codex, posting on behalf of @chihumyum

Astro-Han (commented):
# PR #3450 re-review — `e97d1378df68307beb9cb7dd05bb86f9f0cbcaa3`

## Outcome

NO-GO: one P1 remains.

## Finding

### [P1] Retire the GitHub Copilot connection-detail controller on Host generation changes

`GitHubCopilotReloginNotice` remains mounted across a same-key Runtime Host generation change. It captures the Host and starts `connectExistingLogin`, then accepts the late result and invokes `onRelogin` whenever its component-level mounted ref remains true (`provider-connection-detail.tsx:942-977`). Unlike `OAuthReloginNotice` immediately below it (`provider-connection-detail.tsx:997-1006`) and the catalog login panel, it is not wrapped in `RuntimeHostSettingsGenerationBoundary`.

Reproduction: open a GitHub Copilot connection detail, click Import/Reimport, and replace the selected Host with the same `profileId:hostId` before the promise settles. The Settings/connection-detail tree deliberately stays mounted, so this controller's mounted ref and action guard survive: the old generation can report an error or call the stale `onRelogin` callback into the new page, and the new generation's action remains blocked until the old request finishes. This is the same class of Host-owned OAuth subauthority the new generation boundary is intended to retire. Wrap this controller (or move it onto the already generation-keyed shared direct OAuth controller) and add the same-generation replacement regression.

## Verified behavior

- OAuth card snapshots clear and re-read on `generationKey`; the monotonically increasing refresh ticket rejects old-generation and same-generation superseded reads.
- Browser OAuth login/relogin controllers remount under the narrow generation-keyed boundary. Unmount cancels a known pending authorization ID and the mounted fence suppresses late callbacks; an initiation still awaiting `getAuthUrl` cancels its ID when it eventually returns.
- The boundary is narrow: parent Settings, Models route, provider catalog filter, DOM scroll/focus, and drafts stay mounted. Existing stories exercise catalog route retention, account replacement, and browser authorization cancellation.
- Mutation check: removing the `Fragment` generation key made `runtime-host-settings-generation.test` fail specifically because the OAuth controller did not unmount/remount; stable Settings state remained outside the keyed boundary. Source was restored and rebuilt afterward.

## Checks run

- Exact-head hosted `test`: SUCCESS.
- Desktop workspace dependency builds: green.
- Desktop TypeScript checks (preload, main, renderer, Storybook): green.
- Focused OAuth/Host generation suites: 35/35 green.
- Worktree clean after mutation restoration.

chihumyum:
@Astro-Han Thanks for the exact-head review. The remaining Copilot generation P1 is fixed, and the branch is rebased onto current `main@e6e0c34c7`; new exact head: `cbace2eb9`.

`GitHubCopilotReloginNotice` now retires inside the narrow `RuntimeHostSettingsGenerationBoundary`. Its direct-import result also checks controller lifetime before reporting or refreshing, so a late old-generation result cannot call into the replacement page. The parent Connection Detail remains mounted, preserving route, drafts, scroll, and focus.

A new Story starts a pending Copilot reimport, replaces the same `profileId:hostId` with a new epoch, verifies the replacement action guard is immediately available, settles the retired request, and asserts that no stale credential refresh lands. Mutation check: removing the mounted fence makes the Story fail on an extra retired-generation credential read (`expected 3 to be 2`). The restored exact-head Story play passes with no console or page errors.

During the rebase, current main's newly merged Goal adapter test also exposed three fixture fields that are not part of `SessionChangedEvent`; the final test-only commit removes those invalid discriminants so current-main typecheck/build remain green.

Exact-head verification: lint, format, ASF headers, Astryx inventory (207 files), full typecheck/build, Desktop tests, both Knip workspaces, Storybook build and 186/186 render smoke, and Settings Electron E2E 5/5. The PR is mergeable; the newly triggered hosted check has not reported a result yet.

Could you please re-review `cbace2eb9` when convenient?

apps/desktop/src/renderer/settings/settings-surface.tsx
```diff
@@ -303,86 +394,94 @@ export function SettingsSurface(props: {
...
-          : 'unavailable'
+      : runtimeHostAvailabilityStatus !== 'ready'
+        ? runtimeHostAvailabilityStatus
         : runtimeHostDataLoading || !runtimeHostContentReady
           ? 'loading'
           : 'ready';
+  const runtimeHostContentVerified =
+    runtimeHostCatalogStatus.isVerified &&
+    (!sectionNeedsSettings || selectedRuntimeHostSettingsStatus.isVerified) &&
+    (!sectionNeedsConnections || selectedRuntimeHostConnectionsStatus.isVerified);
```
jackwener, line 432: [P2] `models` (and the other `runtime-host` pages that are not in `sectionNeedsSettings` / `sectionNeedsConnections`) treat catalog verification as enough to lift `RuntimeHostInteractionBoundary`.

Reachability ②: same-key Host generation change while Settings is on 模型.

1. The generation handler invalidates catalog/settings/connections (`isVerified: false`) — inert is on, good.
2. `reloadRuntimeHosts()` completes and the catalog becomes `isVerified` again.
3. `runtimeHostContentVerified` does not wait for connections on this page, so `isInteractive` becomes true while `selectedRuntimeHostConnectionsStatus.isVerified` is still false.
4. `ProvidersPanel` reloads on the new `bridge` identity, but `reload()` never sets `loading` and never clears `connections`, so the previous generation's rows stay visible and clickable.

A click in that window is a Host-owned mutation against the replacement Host, using a list that has not been verified for this generation. Mixed pages (通用/项目/数据) already wait on their own verified predicates; 模型 is the page that actually mutates connections and does not.

Minimal fix: include `models` in the connections-verified predicate (or key/clear `ProvidersPanel` on `generationKey` and set `loading` for that refetch). Cached rows can stay visible; they should not stay writable.

jackwener (commented):
Independent review of `cbace2eb9336f32cd10a3622691855629674c92a`.

**GO for the stated problem; 1×P2 so I am not approving.** Hosted `test` is terminal success. PR is CONFLICTING — no merge from me.

## What this solves

Settings was flashing the warning banner and collapsing Runtime Host rows on every reopen/revalidation. The cut is right: keep last-ready masked snapshots visible, bind write authority to Host key + lifecycle epoch, mute Host-owned controls until that generation is verified, and leave Desktop-owned prefs usable on mixed pages.

## Findings

### [P2] independently derived — `models` lifts the write fence before the new generation's connections are verified
See inline on `settings-surface.tsx`. After a same-key Host replacement the catalog can come back `isVerified` while connections are still the previous generation; `ProvidersPanel` also keeps its old list clickable because `reload()` does not enter a loading state.

### Prior Copilot P1 — independently re-judged **closed** on this head
Both `GitHubCopilotReloginNotice` and `OAuthReloginNotice` remount under `RuntimeHostSettingsGenerationBoundary`, and `connectExistingLogin` now returns before `onRelogin` when the old controller has unmounted. I am not re-filing it.

No P0/P1 on this head.

apps/desktop/src/renderer/settings/settings-surface.tsx
chihumyum: Fixed on exact head 26af71b73a6316cfbd21de0adb405cf73d3043f1.

Models now participates in the connections-verified predicate, so catalog readiness alone cannot lift RuntimeHostInteractionBoundary. ProvidersPanel stays mounted and its last-ready rows remain visible, preserving route/filter/draft/scroll state, but those rows stay inert until the replacement Host generation verifies its connections snapshot.

The new ModelsConnectionsHostGenerationRevalidation story reproduces the split state: the replacement catalog has landed (the Remote profile option is visible), the old Z.AI Live row remains rendered, connections are still pending, and the boundary remains inert with no warning flash.

Validation on this exact head: typecheck; Astryx surface inventory; Storybook build and render smoke 188/188; desktop tests 1356/1356; Settings Electron E2E 5/5; lint; format; ASF headers; desktop and UI knip. Rebased onto upstream main 8e7fa68dc with a clean merge tree.

— Codex, posting on behalf of @chihumyum

chihumyum:
@jackwener The Models/connections generation finding is fixed on exact head 26af71b73a6316cfbd21de0adb405cf73d3043f1 and the inline thread now has the reproduction/validation evidence. The PR is rebased and GitHub reports it mergeable.

The fork CI is waiting for maintainer approval: https://github.com/apache/maka/actions/runs/32684251884. Could you approve that workflow when convenient?

— Codex, posting on behalf of @chihumyum

zhiiw (approved):
Incremental re-review at exact head `26af71b73a6316cfbd21de0adb405cf73d3043f1` (previous approval on `34c52eb2`; this pass covers only the three new commits: `f1dd6f1d` OAuth state, `ce8ffe85` Copilot reimport, `26af71b7` Models fencing).

**The P1 is closed, and here is the lifecycle answer.** `GitHubCopilotReloginNotice` now mounts inside `RuntimeHostSettingsGenerationBoundary` (`provider-connection-detail.tsx`), which is a `Fragment` keyed by the generation key (`runtime-host-settings-target.tsx`). A same-key Host replacement changes the key, so the controller **remounts**: `useActionGuard` (`connectGuard`) and `useMountedRef` (`mountedRef`) are per-instance hooks and are therefore rebuilt per generation — they no longer survive across generations. The late-callback half is covered too: after `connectExistingLogin` resolves, `if (!mountedRef.current) return;` drops a completion belonging to the retired generation. The new story `ModelsCopilotReimportHostGenerationRevalidation` exercises exactly this: import in flight → same-key replacement → guard re-armed, old resolve produces no credential refresh, new import works. The sibling `OAuthReloginNotice` and the `OAuthLoginPanel` received the same boundary wrap.

**No dual authorities in the three new retirement paths.** OAuth card snapshots reset and refetch on generation change (`useOAuthCards` effect keyed on `useRuntimeHostSettingsGenerationKey()`); the pending browser authorization is torn down via `teardownPendingAuthorization` (the new unit test `runtime-host-settings-generation.test.ts` proves only Host-owned controllers unmount while route/scroll/focus state survives); Models joins the connection-verified fence (`settings-surface.tsx` adds `models` to `sectionNeedsConnections`, and the `settingsRuntimeHostInteractionBoundary` goes `inert` until the new generation verifies — pinned by `ModelsConnectionsHostGenerationRevalidation`). One authority per state, each retired with its generation.

**Executed evidence (real Windows x64, exact head):** desktop main rebuild clean; `runtime-host-settings-generation` unit test passes (retires only Host-owned controllers, cancels the pending authorization id).

**Gate:** `test` `completed/success` on the exact head (this diff touches no manifest/sandbox paths, so nothing else is owed); head re-polled at review time, unchanged, MERGEABLE.

No new findings in the incremental surface.

<details>
<summary>简体中文</summary>

增量复审：P1 已闭合——GitHubCopilotReloginNotice 现在挂在 generation boundary 里，换代即重挂，connectGuard/mountedRef 随实例重建；迟到的成功回调由 `mountedRef` 闸丢弃；新增 story 把这条路径完整演了一遍。三条新退场路径均无双权威（OAuth 快照按代重置重取、进行中的授权按代取消、Models 进连接验证围栏）。本机真 Windows 重建 + 新增单测通过。增量面无新 finding。

</details>

Astro-Han:
LGTM — merging. Thanks for the long back-and-forth on this one; the Settings surface staying put across a Runtime Host refresh is a real improvement.

<details>
<summary>中文</summary>

看过了，合并。感谢反复打磨——Runtime Host 刷新时设置界面不再跳，这个改动值。
</details>

Issue #3446, opened 2026-08-21
refactor(design-system): finish wiring the Maka theme to Astryx's token seams


Astro-Han:
## Summary

The Astryx migration (#1565) finished — no second component system, no bare `<button>`/`<input>` anywhere, 190/190 files aligned. What did not finish is the **token wiring**: `defineTheme` uses 6 of its 13 sections and only **5 of them are live** — the sixth, `components`, ships one rule that product CSS cancels unconditionally (see Latent). The sections left empty were worked around elsewhere. Four token dimensions run two vocabularies that are not aliases of each other.

The same shape shows up one level down, inside Maka's own token families: a family gets built, some call sites move onto it, and the rest keep their hand-rolled literals. F5 is that case.

Two items are visible on screen; the rest are latent or process gaps.

Measured as a computed-style sweep over `Product/Shell Official AppShell → Native Conversation` (1180×788, page pinned to `100vh`, both colour schemes) plus source scans, re-run against current `main`. Every number below is re-derived rather than carried forward.

## Status

**Settled, merged in #4465:** F1, F5, F8, and the latent `components` rule. F4's dark-mode half is done; its other half is withdrawn.

**Settled, merged in #4509:** F2 — the ladder collapses to Astryx's two tiers (`--muted-foreground` at 68% ink, `--foreground-secondary` deleted, Astryx text/icon inks wired at the theme root, the five-selector bridge removed) — plus the second `--color-accent-muted` authority, info banners on the palette accent, `--accent-solid` cleared on tinted palettes, StatTile's unused tone machinery and the sub-3:1 tinted borders removed. F3/F7 is #4472.

The decisions that closed them, since they are not obvious from the findings text and constrain everything after:

- **Radius converges Astryx-aliases-Maka** (Q1's second option). The `px` side is authoritative and the theme emits `--radius-inner/-element/-container/-full` as `var(--radius-control/-surface/-modal/-pill)`.
- **The status wash family is deleted outright rather than completed.** Maka had a second vocabulary for something Astryx already owns. Product CSS now reads `--color-{success,warning,error}-muted` and `--color-accent-muted` and nothing else — **one rung at 0.24, no border rung, no strong tier**, which is upstream's shape. This makes Q3 moot: the answer to "what does the family owe the off-family call sites" is that the family goes away and every call site lands on upstream's four names. It also fixes 124 `<Banner>` instances that were palette-blind, since only `--color-accent-muted` had been remapped.
[...]

PR #3452, opened 2026-08-21
fix(ui): stop overriding Astryx's chat bubble radius

Astro-Han:
## Summary

The bubble and the composer are designed to share one radius token. `ChatMessageBubble` defaults to `var(--radius-chat)` (28px); `ChatComposer` reads the same token via `--_chat-composer-radius`. The product overrode the bubble twice — `density="compact"` dropped it to 12px, then `.maka-chat-message-bubble-user` pushed it to 10px — and left the main composer alone.

image

image

*Full view, light (top) then dark (bottom). Story `product-shell-official-appshell--native-conversation`, 1180×788 @2x, scrolled to top.*

image

image

*Corner zoom on the bubble/dock join, light (top) then dark (bottom). Same story and conditions. The AFTER bubble sits ~32px lower: the balanced tier lengthens the transcript.*

Deletions only, no new value. 28px is what the component picks on its own once the overrides are gone.

Side Chat is covered by the same rule. `quote-side-panel.css` pinned the side composer to `--radius-surface` (10px), which paired with the bubble only because product CSS forced the bubble to 10px as well; that override goes too. Side Chat is a branch of the main conversation, not a second conversation surface, so it resolves the same token rather than carrying its own interaction language.

image

*Side Chat (quote companion) at the 320px workbar floor, real cascade with no harness overrides.*

`density` is a spacing tier, so dropping it also moves `ChatMessage`'s children gap 2px → 4px and the transcript reads looser — bundled with the tier, intended. Bubble padding is *not* affected: Astryx's `paddingCompact` and `paddingBalanced` resolve to the same StyleX atoms. `gap={4}` is removed from both call sites: under `balanced` the tier already selects the identical `--spacing-4` atom (`x18g69wz`), so the prop restated what it was reading.

`DESIGN.md` §6 moves in the same PR: chat bubbles leave the 10px card row, a 28px chat rung records what `--radius-chat` governs, and the Two-Name Rule no longer claims nothing in the product renders at 28px — `ChatComposer` already did, on `main`.

Refs #3446 — same root as its F1 finding, separate intent, none of its other work folded in.

## Verification

Playwright computed-style sweeps against a live Storybook dev server, before and after, same branch and same server. Page pinned, `scrollTop` 0, identical framing. **Light and dark — identical results in both.**

Main conversation:

| | before | after |
|---|---|---|
| user bubble | `10px` | `28px` |
| assistant bubble | `12px` | `28px` (ghost, transparent) |
| user bubble padding | `12px 16px` | `12px 16px` |
| composer drawer | `28px 28px 0 0` | `28px 28px 0 0` |
| list density | `compact` | `balanced` |

Side Chat (quote companion), measured at the 320px workbar floor and at 600px, harness carrying no CSS overrides:

| | on `main` | first push | now |
|---|---|---|---|
| user / assistant bubble | `10px` | `28px` | `28px` |
| composer drawer | `10px` | `10px` | `28px` |
| `--_chat-composer-radius` | `10px` | `10px` | unset → `--radius-chat` |

The middle column is the regression M4n5ter caught: the first push fixed the main conversation and split Side Chat. Both widths and both colour schemes agree.

Full-page sweep of every visible element's radius, font-size/line-height, colour and gap on the main story. Everything that moved is accounted for, nothing else:

- radius `10px` ×35→32, `12px` ×3→1, `28px` ×1→6 — distinct radii stay at 10, no new value in the ladder
- gap `2px` ×11→5, `4px` ×22→27 — `ChatMessage`'s `childrenGap`, the tier's own spacing
- 435→431 visible elements — signature diff shows every other element is the same element with `compact` renamed to `balanced`; the four that leave are the last message's tail, pushed below the fold as `scrollHeight` grows 1473→1505px
- removing `gap={4}` changed nothing: the sweep is byte-identical with and without it, in both schemes

`density="compact"` on `<Markdown>` (`chat-turn.tsx:1107`) is left alone — a different density axis with its own `[data-density="compact"]` contract.

Ran: `npm run format` (no fixes), `@maka/ui` tsc build (clean), focused tests `chat-turn-answer-identity` / `chat-turn-steering-order` / `chat-conversation-items` / `transcript-projection` (23 pass, 0 fail). Not run: repo-wide suite (CI), Electron E2E.

## Review focus

No Storybook story renders Side Chat, which is why the first push's sweep missed it. The measurements above come from a scratch story kept out of the commit. The natural permanent seam is `apps/desktop/e2e/session-workbar.spec.ts`, which already asserts `overflow` on `.maka-composer-astryx` — adding a radius assertion there would pin the pairing. Say the word and I will add it here rather than in a follow-up.

`packages/ui/src/chat-surface-layout.tsx:35` says `chat-surface-layout.test.tsx` holds the balanced density value. That file does not exist in the repo — the `ChatLayout` density default is unguarded. Pre-existing, out of scope.

`DESIGN.md` §6 is a deliberate revision of the design authority, written as a rule rather than an exception. Please review that wording as policy.

## AI use

Tool(s) and scope: Claude Code — the deletions, the DESIGN.md §6 revision, the CSS comment rewrites, the Playwright measurement harnesses, and this body. Every number above comes from those harnesses against a live server, not from reasoning about the cascade. Commits carry a `Generated-by: Claude Code` trailer.

## Checklist

- [ ] No

### DIFF
--- a/packages/ui/src/chat-view.tsx
+++ b/packages/ui/src/chat-view.tsx
@@ -483,7 +483,6 @@ export function ChatView(props: {
             owns. */}
         <ChatMessageList
           className="maka-chat-message-list maka-chatContent"
-          gap={4}
           emptyState={conversationItems.length === 0 ? emptyContent : undefined}
         >
           {conversationItems.length > 0 ? (
@@ -588,7 +587,6 @@ export function ChatView(props: {
         <ChatMessageList
           className="maka-chat-message-list maka-chatContent"
           data-turn-source-count={turns.length}
-          gap={4}
           isStreaming={streamingActive}
           emptyState={showEmptyState ? emptyContent : undefined}
         >
--- a/packages/ui/src/chat-view.tsx
+++ b/packages/ui/src/chat-view.tsx
@@ -483,7 +483,6 @@ export function ChatView(props: {
             owns. */}
         <ChatMessageList
           className="maka-chat-message-list maka-chatContent"
-          density="compact"
           gap={4}
           emptyState={conversationItems.length === 0 ? emptyContent : undefined}
         >
@@ -589,7 +588,6 @@ export function ChatView(props: {
         <ChatMessageList
           className="maka-chat-message-list maka-chatContent"
           data-turn-source-count={turns.length}
-          density="compact"
           gap={4}
           isStreaming={streamingActive}
           emptyState={showEmptyState ? emptyContent : undefined}
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -126,11 +126,12 @@ Nothing interactive is square. One ladder, assigned monotonically by box height:
 | Radius | Maka tier | Astryx tier | Assign to |
 |---|---|---|---|
 | 6px | control | inner | chips, keycaps, nested inlays, and product-drawn compact controls |
-| 10px | card | element | cards, rows-as-cards, list containers, chat bubbles; Astryx `Button`, `Input`, `SegmentedControl` |
+| 10px | card | element | cards, rows-as-cards, list containers; Astryx `Button`, `Input`, `SegmentedControl` |
 | 12px | container | container | modals, panels, portal surfaces; Astryx `Card`, `Dialog`, `DropdownMenu` |
+| 28px | chat | chat | the conversation surface as one shape: user bubble, assistant bubble, composer dock. Sourced from Astryx's `--radius-chat` by `ChatMessageBubble` and `ChatComposer` themselves; product CSS never restates it |
 | full | pill (999px) | full (9999px) | badges, pills, circular controls |
 
-- **The Two-Name Rule.** These are one ladder under two vocabularies, and the names never line up: Maka's `control` is Astryx's `inner`, Maka's `card` is Astryx's `element`, Maka's `modal` is Astryx's `container`. Resolve a tier from the box, never from the token name that sounds right. The paired values agree *today* but are independent literals, not aliases — an Astryx upgrade can move one side silently, so a mismatch is a real failure mode rather than an impossibility. Astryx's `--radius-page` (28px) has no Maka tier and no product consumer; anything reaching for a page-level radius is inventing a rung.
+- **The Two-Name Rule.** These are one ladder under two vocabularies, and the names never line up: Maka's `control` is Astryx's `inner`, Maka's `card` is Astryx's `element`, Maka's `modal` is Astryx's `container`. Resolve a tier from the box, never from the token name that sounds right. The paired values agree *today* but are independent literals, not aliases — an Astryx upgrade can move one side silently, so a mismatch is a real failure mode rather than an impossibility. The chat rung is the one tier the product does not assign: `ChatMessageBubble` and `ChatComposer` both resolve `--radius-chat` on their own, which is why the bubble and the dock round together, and why setting a bubble radius or a non-default `density` in product code silently breaks the pair. Astryx's `--radius-page` carries the same 28px literal but is a different token with no Maka tier and no product consumer; reaching for it to match the chat surface is inventing a rung.
 - **The Full-Bleed Rule.** `border-radius: 0` is legal only on true full-bleed rows — an element flush with its container on both sides. Radius and gap move together: if it has breathing room, it has corners.
 - **Proportional marks.** Product-drawn icon plates use ratio-owned radius (~25–27% of the box edge), recorded in prose because Stitch accepts only absolute units.
 
--- a/apps/desktop/e2e/session-workbar.spec.ts
+++ b/apps/desktop/e2e/session-workbar.spec.ts
@@ -70,6 +70,32 @@ test('narrow right workbar keeps launcher shortcuts and side-chat send button in
   const composerCard = companion.locator('.maka-composer-astryx');
   // Keep ChatComposer's inner elevation visible.
   await expect(composerCard).toHaveCSS('overflow', 'visible');
+
+  // #3452: Side Chat is a branch of the main conversation, not a second
+  // conversation surface, so its dock rounds like the main dock. Both resolve
+  // Astryx's `--radius-chat`, the same token `ChatMessageBubble` defaults to —
+  // a side-only `--_chat-composer-radius` split the bubble from the dock the
+  // moment the bubbles moved back to that default. Compared against the main
+  // composer rather than a literal so an upstream token change moves both or
+  // fails here.
+  const composerRadii = await page.evaluate(() => {
+    const wrappers = [...document.querySelectorAll('.maka-composer-astryx')];
+    const inCompanion = (element: Element) => element.closest('.maka-quote-companion') !== null;
+    const effectiveRadius = (element: Element | undefined) => {
+      if (!element) return null;
+      const styles = getComputedStyle(element);
+      return (
+        styles.getPropertyValue('--_chat-composer-radius').trim() ||
+        styles.getPropertyValue('--radius-chat').trim()
+      );
+    };
+    return {
+      side: effectiveRadius(wrappers.find(inCompanion)),
+      main: effectiveRadius(wrappers.find((element) => !inCompanion(element))),
+    };
+  });
+  expect(composerRadii.side).toBeTruthy();
+  expect(composerRadii.side).toBe(composerRadii.main);
   // Match the long model-label pressure from the reported side-chat screenshot
   // without coupling the fixture's globally useful default model to this test.
   await companion.locator('.maka-composer-model-chip-text').evaluate((element) => {
--- a/apps/desktop/src/renderer/styles/chat-message.css
+++ b/apps/desktop/src/renderer/styles/chat-message.css
@@ -62,26 +62,23 @@
   font: var(--maka-text-body);
 }
 
-/* Padding and max-width belong to ChatMessageBubble; re-declaring those here
-   only forks the primitive. What stays is the product tint, the verbatim-text
-   rule (Maka renders user input as typed, which the primitive has no opinion
-   about) — and, since Visual System 2.0 (T3), the radius TIER.
+/* Padding, max-width and radius belong to ChatMessageBubble; re-declaring any
+   of them here only forks the primitive. What stays is the product tint and the
+   verbatim-text rule (Maka renders user input as typed, which the primitive has
+   no opinion about).
 
-   The primitive resolves `--radius-container` (12px). DESIGN.md §6 assigns
-   radius by box height, and container is scoped to "modals, panels, portal
-   surfaces" — a message bubble is none of those. Measured in the chat fixture
-   the filled user bubble is 44px tall, which lands it in the card tier (10px,
-   "cards, rows-as-cards, list containers"). Astryx publishes no radius prop on
-   this component (only `variant`), so the tier is expressed on Maka's own
-   class rather than by reaching into the primitive's internals.
+   Radius is absent on purpose. The primitive resolves `--radius-chat` (28px),
+   the same token `ChatComposer` reads through `--_chat-composer-radius`, so the
+   bubble and the dock round together as one conversation surface — DESIGN.md §6
+   records that as the chat rung. Astryx publishes no radius prop on this
+   component (only `variant`), so a product radius could only fight the
+   primitive, never configure it.
 
    Scoped to the filled bubble on purpose: the assistant bubble is `ghost` (no
-   fill, measured transparent), so it has no corner to round and inherits
-   nothing it needs to override. */
+   fill, measured transparent), so it carries no tint to override. */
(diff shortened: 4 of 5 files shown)


jackwener (commented):
Automated review of exact head `1a99e766ef824bc69d0f5a2548df6257c7cc1084` against current `main@d62857a8357e9160926726a2a13096bc2dc2b91d`.

The defect remains on main: both `ChatMessageList` call sites force `compact`, which activates Astryx's compact bubble radius, and product CSS then overrides the filled user bubble again to 10px while the composer independently stays on `--radius-chat` (28px). The patch removes both competing product decisions and lets the shared primitive/token own the shape; the measured padding result also confirms this is not compensating with a new value. I found no actionable P0-P2 defect. Required `test` is green and the PR includes before/after visual evidence.

Required conclusions:

1. Optimal for the actual problem: **yes**; deleting the two overrides restores the existing Astryx source of truth.
2. Production code to delete: the PR already deletes the unnecessary `density="compact"` props and product radius override; none further identified.
4. Deeper refactor: no.
5. Ready to merge: **code-wise yes**, but this automated review is not approval; the visual/design-policy change still needs independent human review.
6. Residual risks/gaps: no automated computed-style regression protects the 28px pairing, so the supplied visual evidence and a human check are the current guard; future Astryx changes could move the default.

This affects user-visible layout and `DESIGN.md` policy, so `CONTRIBUTING.md` requires independent human judgment on the exact head.

M4n5ter (changes requested):
<details open>
<summary>English</summary>

I reviewed exact head `1a99e766ef824bc69d0f5a2548df6257c7cc1084`.

**Verdict: Request changes.**

### [Important] Preserve the radius pairing in the supported Side Chat surface

Removing `density="compact"` from both lists in `packages/ui/src/chat-view.tsx:484-495,588-594` correctly makes the main conversation bubbles use the default 28px chat radius. However, `ChatView` is also reused by the supported Side Chat path:

```text
quote-companion-panel.tsx:272
→ ChatView
→ balanced ChatMessageList
→ 28px ChatMessageBubble
```

The composer in that same Side Chat surface is still explicitly constrained by `apps/desktop/src/renderer/styles/quote-side-panel.css:81-84`:

```css
--_chat-composer-radius: var(--radius-surface); /* 10px */
```

The resulting supported path is therefore:

```text
Side Chat user bubble: 28px
Side Chat composer:     10px
```

This recreates the exact bubble/composer split the PR intends to eliminate. The override is not dead CSS: #3263 deliberately retained it as the “compact Side Chat radius token,” and the workbar is a real 320–600px surface. The supplied screenshots and computed-style sweep cover only the full-page main conversation.

It also makes the new `DESIGN.md:131,134` statements—“product CSS never restates it” and both components independently resolving `--radius-chat`—false for an existing production consumer.

Given the existing compact Side Chat decision, my preferred root fix is:

- let `ChatView` accept a surface-level density policy;
- keep the main conversation on the balanced default;
- pass compact for the quote companion;
- align the Side Chat composer with the compact bubble radius, preferably the 12px container token instead of restoring a bubble-specific product override;
- scope the DESIGN rule to the main/default conversation surface.

If the product decision is instead that every conversation surface must use 28px, remove the Side Chat composer override and verify the result at the 320px workbar minimum. Either direction needs a Side Chat screenshot or computed-style check; changing the shared default alone is insufficient.

### Simplify audit

Once the surface decision is settled:

- If every `ChatView` remains balanced, delete both `gap={4}` props: balanced already selects the identical `--spacing-4` row-gap atom, so they become duplicate authority.
- If Side Chat remains compact and deliberately keeps the larger cross-turn gap, the override still carries behavior and should remain.
- Document this as a surface/composition contract rather than creating a universal Maka `chat` tier that has no corresponding Maka token and is already contradicted by an embedded composer.

The main-conversation direction otherwise looks sound. Astryx 0.4.0 confirms the stated radius and padding behavior, the additional balanced spacing is intentional, the current-main merge is clean, and exact-head CI is green.

</details>

<details>
<summary>简体中文</summary>

我审查的是 exact head `1a99e766ef824bc69d0f5a2548df6257c7cc1084`。

**结论：Request changes。**

### [Important] 在受支持的 Side Chat surface 中也保持 radius 配对

删除 `packages/ui/src/chat-view.tsx:484-495,588-594` 两处 `density="compact"`，确实能让主会话 bubble 恢复默认的 28px chat radius。但 `ChatView` 还被受支持的 Side Chat 路径复用：

```text
quote-companion-panel.tsx:272
→ ChatView
→ balanced ChatMessageList
→ 28px ChatMessageBubble
```

同一个 Side Chat surface 的 composer 仍在 `apps/desktop/src/renderer/styles/quote-side-panel.css:81-84` 中被明确限定为：

```css
--_chat-composer-radius: var(--radius-surface); /* 10px */
```

因此合并后的实际受支持路径会变成：

```text
Side Chat user bubble：28px
Side Chat composer：    10px
```

这重新制造了本 PR 要消除的 bubble/composer 分叉。该 override 不是死 CSS：#3263 明确将它作为 “compact Side Chat radius token” 保留下来，而 workbar 是真实的 320–600px surface。当前截图和 computed-style sweep 只覆盖了全尺寸主会话。

这也使 `DESIGN.md:131,134` 新增的 “product CSS never restates it” 以及两个组件都会自行解析 `--radius-chat` 的表述，在现有生产消费者上立即失真。

基于 Side Chat 现有的 compact 设计，我倾向于以下根修：

- 让 `ChatView` 接受 surface-level density policy；
- 主会话继续使用 balanced 默认值；
- quote companion 显式传入 compact；
- 将 Side Chat composer 对齐 compact bubble 的 radius，优先使用 12px container token，而不是恢复 bubble 专属的产品 CSS override；
- 将 DESIGN 规则限定到主会话或默认 conversation surface。

如果产品决定所有 conversation surface 都统一为 28px，则应删除 Side Chat composer override，并在 workbar 最小 320px 宽度下验证。无论选择哪条路，都需要补一份 Side Chat 截图或 computed-style 检查；只修改共享默认值是不完整的。

### Simplify audit

完成 surface 决策后：

- 如果所有 `ChatView` 都保持 balanced，应删除两处 `gap={4}`：balanced 已经选择完全相同的 `--spacing-4` row-gap atom，它们届时只是重复 authority。
- 如果 Side Chat 保持 compact，并且仍有意维持更大的跨 turn 间距，那么该 override 仍有真实行为，应继续保留。
- 文档应描述 surface/composition contract，不应创建一个没有 Maka token、并且已经被嵌入式 composer 反驳的全局 Maka `chat` tier。

除此之外，主会话的修改方向成立。Astryx 0.4.0 的源码确认了 radius 和 padding 行为，balanced 带来的额外间距是明确接受的设计变化；与当前 main 的三方合并干净，exact-head CI 也为绿色。

</details>

packages/ui/src/chat-view.tsx
```diff
@@ -483,7 +483,6 @@ export function ChatView(props: {
             owns. */}
         <ChatMessageList
```
hqhq1025, line 484: [P2] Preserve the radius pair in Side Chat

Removing the shared `density="compact"` also changes the `ChatView` rendered by `quote-companion-panel.tsx`. That surface still sets `--_chat-composer-radius: var(--radius-surface)` in `quote-side-panel.css`, so its bubble becomes 28px while its composer remains 10px. Please either keep Side Chat on an explicit compact surface policy and align both components, or remove the composer override and verify 28px at the 320px workbar minimum. The current main-conversation screenshots do not cover this production path.

hqhq1025 (commented):
Reviewed exact head `1a99e766ef824bc69d0f5a2548df6257c7cc1084` against current main.

Verdict: not ready to merge.

The underlying problem is real in the full conversation: product code overrides Astryx twice, so the message bubble and composer no longer share the primitive-owned chat radius. Deleting those overrides is the simplest correct direction for that surface.

However, `ChatView` is a shared component, and this revision changes the supported Side Chat surface at the same time. Side Chat still forces its composer to `var(--radius-surface)` (10px), while the shared message list now defaults its bubbles to 28px. That recreates the same split this PR intends to remove, and makes the new universal DESIGN.md wording false for an existing production consumer.

First principles / Occam: choose the radius policy at the conversation surface boundary. Either keep Side Chat compact by passing an explicit density/policy and align its composer, or remove the Side Chat composer override and verify the 28px design at the 320px minimum width. Do not change a shared default while leaving one consumer on a contradictory local override.

Deletion audit: if every `ChatView` remains balanced, the two explicit `gap={4}` props are duplicate authority and can be deleted. If Side Chat remains compact while intentionally keeping the larger inter-turn gap, they still carry behavior.

Tests: exact-head CI is green. A clean local install with dependency patches passed `npm run build:test`, `npm run format:check`, and `git diff --check`. The remaining gap is the missing Side Chat computed-style or screenshot regression.

Required conclusions:
- Problem definition: correct for the main conversation, incomplete for shared consumers.
- Optimality: not yet; the patch moves the mismatch into Side Chat.
- Production code to delete: potentially both `gap={4}` props after the surface policy is settled.
- Tests to delete: none identified.
- Deeper refactor: no broad refactor; add a surface-level density/radius contract.
- Merge readiness: no.
- Residual risk: embedded conversation surfaces remain visually unverified.

Astro-Han:
Confirmed by measurement. Reproduced at the 320px workbar floor and at 600px, real cascade, no harness overrides: bubble 28px, composer 10px, `--_chat-composer-radius` resolving to `10px`. Your trace holds, including that the override is live.

I unified instead of scoping. `DESIGN.md` §6's Don't list forbids introducing another radius tier, and a surface-level density policy is exactly that — a second pairing (12/12) beside the first. Side Chat is a branch of the main conversation: same transcript component, same composer component, opened from a quote inside it. A branch with its own geometry is a second interaction language.

My own objection to unifying also failed on checking. I expected 28px to read badly on a short bubble at 320px, and it is near-capsule there — but so is a short bubble in the main conversation at full width, which this branch already accepts. That is about the token's treatment of short bubbles, not about Side Chat.

`--_chat-composer-radius: var(--radius-surface)` is deleted in `7c8011a91`; after, both widths and both schemes read 28px/28px with the token unset. #3263's `overflow: visible` contract is preserved in a comment, since the rule had no other declaration left. `DESIGN.md`'s "product CSS never restates it" is now true rather than scoped down.

Simplify audit: both `gap={4}` props removed. A full computed-style sweep is byte-identical with and without them in both schemes — `gapBalanced` selects the same `x18g69wz` (`--spacing-4`) atom the prop asked for.

On the verification gap: no Storybook story renders Side Chat, which is why the first sweep missed it. The permanent guard is now in `session-workbar.spec.ts` (`a4e8521cb`) — the 320px side-chat test asserts the side composer's effective radius equals the main composer's. Verified red with the override restored.

Not done: the `density` seam on `ChatView`. If you want it for a reason other than radius, I will take it separately.

packages/ui/src/chat-view.tsx
Astro-Han: Addressed in `7c8011a91` — I took the second option. `--_chat-composer-radius: var(--radius-surface)` is deleted, so Side Chat resolves `--radius-chat` like the main conversation; Side Chat is a branch of that conversation, not a second surface, so a surface-level density policy would have been a second radius tier — which §6's Don't list rules out.

Verified at the 320px workbar floor and at 600px, both colour schemes, real cascade with no harness overrides: bubble 28px, composer 28px, token unset. The permanent guard is in `session-workbar.spec.ts` (`a4e8521cb`), comparing the side composer's effective radius against the main composer's rather than a literal; verified red with the override restored.

Both `gap={4}` props are removed — a full computed-style sweep is byte-identical with and without them in both schemes.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Astro-Han:
@M4n5ter Could you take a look at the newest head again?

M4n5ter (approved):
<details open>
<summary>English</summary>

Re-reviewed exact head `a4e8521cb0da006d0681b5d8c8276ef8b157f3c8`.

The previous blocking finding is resolved. Side Chat now follows the same shared 28px chat radius as the main conversation, the contradictory composer override has been removed, and both redundant `gap={4}` props are gone.

The new E2E assertion covers the actual regression boundary: restoring the old Side Chat override makes its effective composer radius diverge from the main conversation. Current exact-head CI is green, including the Desktop E2E suite.

LGTM — approved.

</details>

<details>
<summary>简体中文</summary>

已复审 exact head `a4e8521cb0da006d0681b5d8c8276ef8b157f3c8`。

上次阻塞合并的问题已经解决。Side Chat 现在与主会话统一使用共享的 28px chat radius；与之矛盾的 composer override 已删除，两处冗余的 `gap={4}` 也已清理。

新增的 E2E 断言覆盖了本次回归的实际边界：一旦恢复旧的 Side Chat override，其 composer 有效半径就会重新与主会话产生分叉。当前 exact-head CI 全部通过，包括 Desktop E2E。

LGTM，可以合并。

</details>

PR #1920, opened 2026-08-02
fix(desktop): make macOS dev permission grants reliable

liugddx:
## Summary

- avoid native bundle icon reads that can crash unpackaged Electron permission guides
- generate a stable, ad-hoc-signed `Maka Dev.app` for macOS development
- launch the development bundle through LaunchServices so TCC attributes permissions to `Maka Dev`, not the parent terminal
- reuse the signed bundle until the installed Electron version changes
- preserve Vite renderer HMR by explicitly forwarding its development URL
- keep other platforms on their existing Electron development launcher
- document the authorization and restart workflow
- add regression coverage for LaunchServices arguments and environment filtering

Closes #1919

## Root cause

The npm-installed `Electron.app` failed strict bundle signature verification. Skipping `app.getFileIcon()` prevented its native `SIGTRAP`, but dragging that bundle still did not produce a usable TCC grant. A valid ad-hoc signature and stable bundle ID were necessary but not sufficient when its internal executable was launched directly from a terminal: macOS continued returning Accessibility `false` and Screen Recording `denied`.

Launching the same signed bundle through LaunchServices made the TCC responsibility chain resolve to `Maka Dev`; the existing permission poll then changed to granted on the real macOS device.

Generated `.app` contents remain ignored and are never committed.

## Verification

- real macOS 13.4.1 authorization: `Maka Dev` appeared in System Settings and Maka changed from “waiting for permission” to granted
- `codesign --verify --deep --strict apps/desktop/.maka-dev/Maka\ Dev.app`
- `npm run dev -- --user-data-dir=<isolated-dir>` launched the signed bundle through LaunchServices with Vite HMR active
- `npm --workspace @maka/desktop run build`
- `npm run test:scripts` (23 passed)
- focused permission/dev launcher tests (19 passed)
- Biome checks and `git diff --check`

Issue #1919, opened 2026-08-02
fix(desktop): macOS development permission guide crashes or cannot grant access


liugddx:
## Summary

Opening the guided Accessibility or Screen Recording grant flow can terminate the unpackaged Electron app on macOS. The process exits with a native `SIGTRAP` rather than a catchable JavaScript exception.

## Environment

- Maka Agent commit: `79fd71f6`
- Electron: `43.1.1`
- macOS: `13.4.1` (`22F82`)
- Hardware: `Mac14,9` (Apple silicon)
- Run mode: development / unpackaged Electron

## Reproduction

1. Start the desktop app in development mode.
2. Open the Browser Use / Computer Use authorization flow.
3. Click the guided authorization entry for Accessibility or Screen Recording.
4. Electron exits while the permission overlay is opening or beginning the drag interaction.

## Evidence and likely cause

The macOS crash reports show `EXC_BREAKPOINT (SIGTRAP)` on a `ThreadPoolForegroundWorker`. There is no JavaScript exception to catch.

The permission overlay resolves the development bundle to `node_modules/electron/dist/Electron.app` and invokes `app.getFileIcon()` in two places: while preparing the overlay icon and as a fallback when starting the drag. Reading that native bundle icon is the common trigger observed immediately before both crashes.

## Proposed behavior

- Keep native bundle icon loading for packaged builds, where the target is the real `Maka.app` bundle.
- Skip `app.getFileIcon()` for unpackaged development builds.
- Continue dragging `Electron.app` in development mode with an empty decorative drag image, preserving the correct macOS TCC identity.

A focused fix and regression test will be submitted in a linked PR.

liugddx:
## Follow-up from real-device verification

The proposed crash guard prevents the original native exit, but end-to-end development authorization is **not yet successful**.

After dragging the development bundle into System Settings and enabling the newly added Electron entry, Maka continues to report that it is waiting for permission. A probe launched with the exact same Electron executable reports:

```json
{
  "executablePath": "/Users/liugddx/code/maka-agent/node_modules/electron/dist/Electron.app/Contents/MacOS/Electron",
  "accessibility": false,
  "screen": "denied"
}
```

Bundle signature verification also fails:

```text
Electron.app: code has no resources but signature indicates they must be present
```

This means the current PR fixes the `app.getFileIcon()` crash but does not make the unpackaged development TCC flow usable. The PR has been converted to Draft to prevent premature merging.

[...]

PR #3455, opened 2026-08-22
fix(desktop): never request the 'large' file icon that kills packaged macOS builds

### LINKED ISSUE #3352: bug(desktop): packaged macOS permission guide crashes while loading the app icon
### English

The packaged Maka 0.1.11 desktop app terminates when I start the guided authorization flow for Accessibility.

Expected behavior: the permission guide opens, displays the Maka application card and icon, and remains usable for dragging Maka into System Settings.

Actual behavior: Maka exits with `EXC_BREAKPOINT (SIGTRAP)` while Electron is loading the application bundle icon. The unpackaged Maka Dev build does not reproduce this crash.

<details>
<summary>简体中文</summary>

正式打包的 Maka 0.1.11 在启动“辅助功能”的引导授权流程时会直接退出。

预期行为：权限引导正常打开，显示 Maka 应用卡片和图标，并可以将 Maka 拖入系统设置完成授权。

实际行为：Electron 加载应用 Bundle 图标时，Maka 发生 `EXC_BREAKPOINT (SIGTRAP)` 并退出。未打包的 Maka Dev 开发版不会出现该崩溃。

</details>

### English

1. Install and launch the packaged Maka 0.1.11 application on an Apple silicon Mac.
2. Ensure that Accessibility permission has not been granted.
3. Open Maka Settings → Permissions.
4. Click **Guide me** for Accessibility.
5. The macOS permission guide opens, and Maka terminates while loading or preparing the application icon.

I reproduced this behavior on my machine.

<details>
<summary>简体中文</summary>

1. 在 Apple Silicon Mac 上安装并启动正式打包的 Maka 0.1.11。
2. 确保尚未授予“辅助功能”权限。
3. 打开 Maka“设置 → 权限”。
4. 点击“辅助功能”旁边的“引导授权”。
5. macOS 权限引导出现后，Maka 在加载或准备应用图标时崩溃退出。

我已在自己的电脑上复现该问题。

</details>

### Environment

- Maka version or commit: 0.1.11
- Build: packaged desktop release installed from DMG
- OS: macOS 26.5.2 (25F84)
- Device: Mac16,10
- Architecture: Apple silicon (`arm64`)
- Electron: 43.2.0
- Surface: Desktop
- Node.js version: N/A (packaged application)

<details>
<summary>简体中文</summary>

- Maka 版本：0.1.11
- 构建类型：通过 DMG 安装的正式打包版本
- 操作系统：macOS 26.5.2 (25F84)
- 设备型号：Mac16,10
- 架构：Apple silicon (`arm64`)
- Electron：43.2.0
- 使用入口：Desktop
- Node.js 版本：不适用（正式打包应用）

</details>

### English

Crash information:

- Exception: `EXC_BREAKPOINT (SIGTRAP)`
- Crashing thread: `ThreadPoolForegroundWorker`
- Symbolicated Electron frame: `IconLoader::ReadIcon()`
- Relevant Electron offsets: `0x87cc018` and `0x87cbed0`

The available evidence points to the two packaged-only calls to:

```ts
app.getFileIcon(..., { size: 'large' })
```

Electron documents the `large` icon size as unsupported on macOS:

https://www.electronjs.org/docs/latest/api/app#appgetfileiconpath-options

This is a packaged-build follow-up to #1919 and #1920. Issue #1919 covered development/unpackaged Electron, and PR #1920 intentionally skips native icon loading only for unpackaged builds while retaining it for packaged builds.

<details>
<summary>简体中文</summary>

崩溃信息：

- 异常类型：`EXC_BREAKPOINT (SIGTRAP)`
- 崩溃线程：`ThreadPoolForegroundWorker`
- Electron 符号化结果：`IconLoader::ReadIcon()`
- 相关 Electron 偏移：`0x87cc018` 和 `0x87cbed0`

目前的证据指向正式打包路径中的两处调用：

```ts
app.getFileIcon(..., { size: 'large' })
```

Electron 文档明确说明 macOS 不支持 `large` 图标尺寸。

这是 #1919 和 #1920 在正式打包版本上的后续问题。#1919 处理的是开发版/未打包 Electron，#1920 也只在未打包构建中跳过原生图标读取，正式打包路径仍然保留。

</details>

### AI use

OpenAI Codex assisted with crash-log analysis, symbolication, and drafting this report. I reviewed the reproduction details and the final report.

UncertaintyDeterminesYou4ndMe:
## Summary

Packaged macOS builds crash with `EXC_BREAKPOINT (SIGTRAP)` in Chromium's `IconLoader::ReadIcon()` the moment the drag-to-grant permission guide loads the app icon. The trigger is `app.getFileIcon(bundlePath, { size: 'large' })`: Electron documents `'large'` as unsupported on macOS, and in practice it hits a fatal `NOTREACHED` in native code — the process dies before the promise settles, so the `try/catch` inside `loadNativeBundleIcon` never sees an error to catch. Unpackaged builds were spared only because #1920 gates native icon loading on `app.isPackaged` (the #1919 dev-mode crash carried this exact signature: SIGTRAP on `ThreadPoolForegroundWorker` while reading a bundle icon).

The fix requests `{ size: 'normal' }` instead, hoisted into a single exported constant (`BUNDLE_ICON_OPTIONS` in `app-bundle.ts`) shared by both call sites, so the size choice and the reason it must never be `'large'` live in one place. Both call sites are only reachable on macOS (`resolveAppBundle` returns `not_darwin` elsewhere, and the drag IPC handler early-returns off-darwin), and `'normal'` is supported on every platform regardless. The 32×32 result is upscaled to the same 64×64 the code already produced on the platforms where `'large'` silently meant 32×32 (Windows), so no downstream contract changes.

Fixes #3352

## Verification

Empirical reproduction with this repo's Electron 43.2.0 on an Apple silicon Mac (the exact version and hardware class from the report):

- `app.getFileIcon(<.app>, { size: 'large' })` → process exits 133 (`SIGTRAP`), no JavaScript error thrown — matching the crash report's `IconLoader::ReadIcon()` frame.
- `app.getFileIcon(<.app>, { size: 'normal' })` → resolves a 32×32 image.
- The full fixed path (`loadNativeBundleIcon(true, …)` → `getFileIcon` `normal` → `resize({64,64})` → `toDataURL()`) → produces a valid 7 KB PNG data URL, exit 0.

Checks run locally (Node 24.18):

- `npm --workspace @maka/desktop run typecheck` — pass
- `npm --workspace @maka/desktop run test` — 1038 pass / 0 fail (includes the new regression tripwire)
- `biome check` on the changed files — clean; `knip` desktop workspace — clean
- Regression-test validity: reverting `BUNDLE_ICON_OPTIONS` to `'large'` makes the new test fail; restoring the fix makes it pass.

Not run: a full packaged DMG build. The crash and the fix were both demonstrated against the real Electron binary with the production code path mirrored 1:1; the packaged-only aspect is fully explained by the `app.isPackaged` gate in `loadNativeBundleIcon`.

## AI use

Select exactly one:

Tool(s) and scope: Claude Code diagnosed the crash, wrote the fix and test, and ran the reproduction and verification; an independent adversarial review pass (also Claude) probed the fix with fault-injection experiments before submission. I reviewed and verified the result.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/src/main/permission-overlay/permission-overlay-main.ts
+++ b/apps/desktop/src/main/permission-overlay/permission-overlay-main.ts
@@ -24,7 +24,7 @@ import { join } from 'node:path';
 import type { UiLocale } from '@maka/core/ui-locale';
 import { resolveOverlayAssetDir } from '../overlay-assets.js';
 import { openSystemPermissionPane, requestPermissionAccess } from '../permissions-actions.js';
-import { loadNativeBundleIcon, resolveAppBundle } from './app-bundle.js';
+import { BUNDLE_ICON_OPTIONS, loadNativeBundleIcon, resolveAppBundle } from './app-bundle.js';
 import { getPermissionOverlayCopy } from './permission-overlay-copy.js';
 import {
   createPermissionOverlayController,
@@ -77,7 +77,7 @@ export function createPermissionOverlayMain(
   async function resolveAppIconDataUrl(bundlePath: string | null): Promise<string | null> {
     if (!bundlePath) return null;
     const icon = await loadNativeBundleIcon(app.isPackaged, () =>
-      app.getFileIcon(bundlePath, { size: 'large' }),
+      app.getFileIcon(bundlePath, BUNDLE_ICON_OPTIONS),
     );
     if (!icon || icon.isEmpty()) return null;
     // nativeImage.createFromPath does not decode .icns reliably. Asking
@@ -274,7 +274,7 @@ function attachCardGestures(win: import('electron').BrowserWindow): void {
     }
     if (icon.isEmpty()) {
       const fallback = await loadNativeBundleIcon(app.isPackaged, () =>
-        app.getFileIcon(resolved.bundlePath, { size: 'large' }),
+        app.getFileIcon(resolved.bundlePath, BUNDLE_ICON_OPTIONS),
       );
       if (fallback && !fallback.isEmpty()) icon = fallback.resize({ width: 64, height: 64 });
       // The file drag still works without a decorative drag image.
--- a/apps/desktop/src/main/__tests__/permission-overlay-controller.test.ts
+++ b/apps/desktop/src/main/__tests__/permission-overlay-controller.test.ts
@@ -21,6 +21,7 @@ import {
   type PermissionOverlayWindowLike,
 } from '../permission-overlay/permission-overlay-controller.js';
 import {
+  BUNDLE_ICON_OPTIONS,
   loadNativeBundleIcon,
   resolveAppBundle,
 } from '../permission-overlay/app-bundle.js';
@@ -315,6 +316,12 @@ describe('app bundle resolution for the drag', () => {
     assert.equal(icon, null);
     assert.equal(calls, 0, 'unpackaged development must not call app.getFileIcon()');
     assert.equal(await loadNativeBundleIcon(true, async () => 'icon'), 'icon');
+  });
+
+  it('never requests the large icon size that kills packaged macOS builds', () => {
+    // 'large' hits a fatal NOTREACHED inside Chromium's IconLoader on
+    // macOS (SIGTRAP, not a catchable error) — see issue #3352.
+    assert.notEqual(BUNDLE_ICON_OPTIONS.size as string, 'large');
   });
 
   it('walks three levels up from the executable to the .app', () => {
--- a/apps/desktop/src/main/permission-overlay/app-bundle.ts
+++ b/apps/desktop/src/main/permission-overlay/app-bundle.ts
@@ -35,6 +35,15 @@ export interface ResolveAppBundleDeps {
 }
 
 /**
+ * The only size every platform can actually deliver. `'large'` is
+ * unsupported on macOS: Chromium's IconLoader hits a fatal NOTREACHED and
+ * the process dies with SIGTRAP before the promise settles — no JavaScript
+ * error is ever thrown, so the try/catch in `loadNativeBundleIcon` cannot
+ * save the app. Callers upscale the 32x32 result as needed.
+ */
+export const BUNDLE_ICON_OPTIONS = { size: 'normal' } as const;
+
+/**
  * Reading a bundle icon is presentation-only. The original unpackaged npm
  * Electron runtime could terminate natively while macOS resolved its bundle
  * icon, before the returned promise settled. Keep native icon loading for


UncertaintyDeterminesYou4ndMe:

An independent review pass probed this fix with fault-injection experiments before submission. All six attack vectors came back safe; evidence was experimental, not read-only:

1. **No other crash sites.** Repo-wide sweep: `getFileIcon` exists at exactly the two call sites in `permission-overlay-main.ts`, both now on `BUNDLE_ICON_OPTIONS`. No other `size: 'large'` icon requests anywhere.
2. **Downstream contract holds with a 32px source.** Measured on Electron 43.2.0: `'normal'` returns logical 32×32 with scaleFactors [1,2] (a real 64px rep exists); after `resize({64,64})` → real 64×64 PNG that round-trips through `createFromDataURL`. The overlay renders the icon at 28×28 CSS and guards only on `naturalWidth > 0` — no minimum-size assumption. No fidelity regression is possible: on macOS `'large'` never returned an image, it killed the process.
3. **Type contract.** `FileIconOptions.size` accepts the readonly const with no cast; desktop typecheck clean.
4. **The regression test fails without the fix** (verified by injection): reverting the constant to `'large'` and rebuilding → 1 fail; restoring → full desktop suite 1038/1038.
5. **Non-darwin unreachability confirmed** by executing `resolveAppBundle` for win32/linux (`not_darwin` → loader invoked 0 times) plus the drag handler's platform early-return. `'normal'` would be valid there anyway.
6. **#1920's `isPackaged` gate untouched**; its regression test still passes.

Extra: the crash and fix were both reproduced against the real Electron binary — `{ size: 'large' }` exits 133 (SIGTRAP) with the catch block provably never executing; the fixed path (packaged gate → `getFileIcon` → resize → data URL) exits 0 with a valid PNG.

One out-of-scope observation for the record: `start()` awaits the decorative icon fetch before opening the card with no timeout (`permission-overlay-main.ts:195`). A hung macOS icon service would delay the card indefinitely. Pre-existing behavior, untouched here; happy to file a follow-up issue if maintainers think it's worth hardening.

M4n5ter (changes requested):
<details open>
<summary>English</summary>

Reviewed exact head `6a18d8376065f7c7989512e1c1efb36607bcb897`.

The diagnosis of the native crash is correct: Electron documents `large` as unsupported on macOS, and the Chromium implementation reaches a fatal `NOTREACHED` before JavaScript can catch anything. I found one remaining issue with the proposed fix.

### [Important] Use the shipped Maka icon instead of the file-type icon

Changing the two calls at `permission-overlay-main.ts:80,277` from `large` to `normal` prevents the crash, but it does not load the icon belonging to `Maka.app`.

Electron 43.2.0 forwards `app.getFileIcon()` to Chromium's `IconLoader`. On macOS, that implementation discards the individual application identity, reduces the path to its UTType, and calls `NSWorkspace.iconForContentType:`. Chromium's own contract explicitly notes that applications have unique icons but this loader does not handle them. An Electron maintainer has also confirmed that `getFileIcon` returns a file-type icon rather than the icon belonging to the individual file:

- [Electron 43.2.0 implementation](https://github.com/electron/electron/blob/v43.2.0/shell/browser/api/electron_api_app.cc#L1394-L1420)
- [Chromium 150 macOS implementation](https://chromium.googlesource.com/chromium/src/+/150.0.7871.129/chrome/browser/icon_loader_mac.mm)
- [Chromium `IconLoader` contract](https://chromium.googlesource.com/chromium/src/+/150.0.7871.129/chrome/browser/icon_loader.h)
- [Electron maintainer confirmation](https://github.com/electron/electron/issues/15809#issuecomment-441048221)

Consequently, the permission card and drag replica will show the generic application-bundle icon rather than the Maka icon. The current real-device evidence—32×32, non-empty PNG, successful resize and round-trip—only verifies the image's shape, not its identity. This also contradicts the existing comment that the result is the same image Finder and the TCC list display for the bundle.

There is now a smaller root fix available. This PR's exact base is #3451, which already:

- ships the canonical 1024×1024 `assets/icon.png` in packaged resources;
- provides `desktopAssetPath()` for packaged and development layouts;
- uses that same PNG as electron-builder's macOS bundle-icon source.

Please load that known asset with `nativeImage.createFromPath(desktopAssetPath(..., 'assets', 'icon.png'))` and resize it as needed. `resolveAppBundle()` should remain the authority for the file being dragged, but it no longer needs to be the icon source.

This also gives the simplify-audit result for free:

- delete both `app.getFileIcon()` calls;
- delete `BUNDLE_ICON_OPTIONS`;
- delete `loadNativeBundleIcon()` and its packaged-only state;
- delete the implementation-level “not large” test;
- delete the second native fallback during drag—the renderer normally supplies the complete row image, and the existing code already permits an empty decorative drag icon when decoding fails;
- remove the unbounded native icon await from `start()`.

Since the PNG becomes a load-bearing packaged resource, add `assets/icon.png` to `assertPackagedResources`; the existing `desktopAssetPath` tests can continue to cover the dev/packaged path split.

No other material findings. The crash-fix direction is correct, but the current implementation retains the wrong icon authority and returns the wrong representation for every supported packaged-macOS invocation.

**Verdict: request changes.**

</details>

<details>
<summary>简体中文</summary>

已审查精确 head `6a18d8376065f7c7989512e1c1efb36607bcb897`。

对 native crash 的诊断是正确的：Electron 明确说明 macOS 不支持 `large`，Chromium 实现会在 JavaScript 有机会捕获错误前进入 fatal `NOTREACHED`。不过，当前修复仍有一个问题。

### [Important] 应使用随包发布的 Maka 图标，而不是文件类型图标

将 `permission-overlay-main.ts:80,277` 的两个调用从 `large` 改为 `normal` 可以避免崩溃，但不会读取属于 `Maka.app` 的图标。

Electron 43.2.0 会把 `app.getFileIcon()` 转发给 Chromium 的 `IconLoader`。macOS 实现会丢弃具体应用身份，把路径归约为 UTType，然后调用 `NSWorkspace.iconForContentType:`。Chromium 自己的 contract 也明确说明：macOS 应用虽然有独有图标，但这个 loader 不处理它们。Electron maintainer 同样确认，`getFileIcon` 返回的是文件类型图标，而不是具体文件自身的图标：

- [Electron 43.2.0 实现](https://github.com/electron/electron/blob/v43.2.0/shell/browser/api/electron_api_app.cc#L1394-L1420)
- [Chromium 150 macOS 实现](https://chromium.googlesource.com/chromium/src/+/150.0.7871.129/chrome/browser/icon_loader_mac.mm)
- [Chromium `IconLoader` contract](https://chromium.googlesource.com/chromium/src/+/150.0.7871.129/chrome/browser/icon_loader.h)
- [Electron maintainer 的确认](https://github.com/electron/electron/issues/15809#issuecomment-441048221)

因此，权限卡片和拖拽图会显示通用 application-bundle 图标，而不是 Maka 图标。目前的真机证据——32×32、PNG 非空、resize 和 round-trip 成功——只能证明图片结构有效，不能证明图标身份正确。这也与现有注释中“和 Finder/TCC 列表显示同一 bundle 图标”的说法冲突。

当前已经有更小的根因修复。本 PR 的精确 base 就是 #3451，它已经：

- 将 canonical 1024×1024 `assets/icon.png` 放入 packaged resources；
- 提供 `desktopAssetPath()` 处理 packaged/dev 两种目录布局；
- 使用同一 PNG 作为 electron-builder 的 macOS bundle icon 来源。

请通过 `nativeImage.createFromPath(desktopAssetPath(..., 'assets', 'icon.png'))` 读取这个已知资源，再按需 resize。`resolveAppBundle()` 继续作为实际拖拽文件的 authority，但不再兼任图标来源。

这也能直接落实 simplify audit：

- 删除两处 `app.getFileIcon()`；
- 删除 `BUNDLE_ICON_OPTIONS`；
- 删除 `loadNativeBundleIcon()` 及其 packaged-only state；
- 删除只检查 “not large” 的实现型测试；
- 删除拖拽时的第二次 native fallback——正常情况下 renderer 会提供完整的 row image，而现有代码已经允许解码失败时使用空的装饰性 drag icon；
- 移除 `start()` 中无时限的 native icon await。

由于该 PNG 会成为正式的 packaged runtime dependency，还应将 `assets/icon.png` 加入 `assertPackagedResources`；现有 `desktopAssetPath` 测试可以继续验证 dev/packaged 路径分支。

没有发现其他 material finding。止崩方向正确，但当前实现仍保留了错误的 icon authority，并会在每次受支持的 packaged macOS 调用中返回错误 representation。

**结论：request changes。**

</details>

apps/desktop/src/main/permission-overlay/permission-overlay-main.ts
```diff
@@ -77,7 +77,7 @@ export function createPermissionOverlayMain(
   async function resolveAppIconDataUrl(bundlePath: string | null): Promise<string | null> {
     if (!bundlePath) return null;
     const icon = await loadNativeBundleIcon(app.isPackaged, () =>
-      app.getFileIcon(bundlePath, { size: 'large' }),
+      app.getFileIcon(bundlePath, BUNDLE_ICON_OPTIONS),
```
Astro-Han, line 80: **[P2] Use the canonical Maka icon instead of the macOS file-type icon**

Thanks for fixing the crash here. One non-blocking issue remains: on macOS, Chromium's `IconLoader` resolves this path through its UTType and returns the generic application-bundle icon, rather than the icon belonging to this particular `Maka.app`. The current device evidence proves that the PNG is valid, but not that its identity is correct.

Since #3451 already packages the canonical `assets/icon.png` and provides `desktopAssetPath()`, the clean final state is to load that asset with `nativeImage.createFromPath()`. That would also let us delete both `getFileIcon()` calls, `BUNDLE_ICON_OPTIONS`, the native bundle-icon loader, and the implementation-level "not large" test, while keeping `resolveAppBundle()` solely responsible for the file being dragged.

Astro-Han (approved):
Thanks for tracking the packaged macOS crash down to Electron's unsupported `large` icon request. The exact-head change removes the native `SIGTRAP` path, preserves the packaged-only gate, and keeps the dragged `.app` path under main-process authority.

I'm approving this fix with one non-blocking P2 about the remaining icon source. It does not undermine the crash fix, but using the canonical packaged asset would produce the correct Maka identity and leave a substantially smaller final implementation.

AI-assisted review disclosure: OpenAI Codex coordinated two independent exact-head review passes. I verified the crash path, Electron/Chromium icon semantics, current source, exact head, and review state, and I independently made the approval and severity decisions.

M4n5ter:
I also agree that treating the aforementioned issues as follow-up would be more appropriate, approving.

M4n5ter (approved):

Issue #3463, opened 2026-08-22
follow-up(desktop): close diagnostics coverage and default-Host consistency gaps after #3427


M4n5ter:
<details open>
<summary><strong>English</strong></summary>

## Context

#3427 established one context-aware, user-controlled diagnostics pipeline for on-demand capture, operation failures, Renderer recovery, and pre-Renderer native dialogs. Its review identified two non-blocking P2 gaps, while the PR deliberately left crash-survivability and demonstrated inline-only coverage gaps outside its clipboard-only, no-storage scope.

The two review gaps are now closed by #3465, #3478, and #3489. The remaining work must stay split by evidence owner and lifecycle rather than grow the live diagnostics pipeline into a general crash-reporting subsystem.

### 1. Guard default-Host projection completion and ordering

- [x] Reuse a shared Host-identity completion guard for default-Host-backed Skills, managed-source, bundled-catalog, scheduled-task, and Project refreshes (#3465).
- [x] Capture the resolved `{ profileId, hostId }` for each request and commit its result only while that identity is still the current default (#3465).
- [x] Keep only the latest-started same-Host refresh authoritative for each affected projection, with focused ordering coverage (#3477, #3478).
- [x] Suppress obsolete error feedback from a superseded refresh while preserving current-generation failures (#3489).

Sources: https://github.com/apache/maka/pull/3427#discussion_r3835244994 and follow-up review threads on #3465/#3478.

### 2. Bound the browser-only diagnostics fallback

- [x] Reuse one Renderer-report formatter for the non-Electron fallback instead of concatenating unrestricted fields (#3465).
- [x] Define explicit UTF-8 budgets for individual fields and the final report, including oversized `data:` URLs (#3465).
- [x] Cover the contract with focused oversized-error/location tests without introducing another diagnostics authority (#3465).

Source: https://github.com/apache/maka/pull/3427#discussion_r3835245019.

### 3. Close evidence gaps that outlive or bypass the current process surface

- [x] Expose bounded Desktop diagnostics and one native recovery decision when the main Renderer disappears while Electron main remains alive (#3490, #3495).
- [ ] Define bounded next-launch evidence after Electron main itself dies (#3491).
- [x] Audited inline-only Renderer failures against current reports and open user issues; no demonstrated evidence gap currently justifies a general capture layer. Reopen only for a concrete user path.
- [ ] Preserve bounded local Runtime Host exit evidence without inventing cross-deployment persistence; the demonstrated Desktop path is tracked by #3333.
[...]

PR #3465, opened 2026-08-22
fix(desktop): guard Host projections and fallback reports

M4n5ter:
<details open>

## Summary

- Prevent a completed read from an old default Runtime Host from overwriting Skills, managed-source, bundled-catalog, scheduled-task, or Project presentation after the default changes.
- Make the existing Host-identity completion check a shared renderer boundary instead of duplicating it across projections.
- Bound the browser-only Renderer crash report to explicit UTF-8 budgets: 24 KiB details, 2 KiB user agent, 8 KiB location, and 32 KiB overall, with redaction before truncation.
- Keep the Electron path unchanged: main remains its collection, validation, and formatting authority.

Refs #3463

## Review focus

The Host guard owns only completion-time presentation authority; it does not cancel operations or change their diagnostic target. The browser formatter is used only when the Electron diagnostics bridge is unavailable and does not add storage, telemetry, or another report pipeline.

## Verification

- `npm run lint`
- `npm run format:check`
- `npm --workspace @maka/desktop test` — 1063 passed
- Focused default-Host ordering and oversized browser-report tests — 5 passed

## AI use

Tool(s) and scope: OpenAI Codex assisted with implementation, tests, scope review, verification, and pull request drafting; the human contributor remains responsible for the change.

## Checklist

- [ ] No

</details>

<details>

## 概要

- 默认 Runtime Host 切换后，旧 Host 已完成的读取结果不再覆盖 Skills、managed source、bundled catalog、scheduled task 或 Project 的当前展示状态。
- 将已有的 Host identity 完成检查收敛为 Renderer 共享边界，不在各 projection 中重复实现。
- 为纯浏览器 Renderer 崩溃报告增加明确的 UTF-8 预算：details 24 KiB、user agent 2 KiB、location 8 KiB、最终报告 32 KiB，并在截断前脱敏。
- Electron 路径保持不变：main 仍是采集、验证和格式化 authority。

关联 #3463

## 审查重点

Host guard 只负责完成时的展示 authority，不取消操作，也不改变其诊断目标。Browser formatter 仅在 Electron diagnostics bridge 不可用时使用，不增加存储、遥测或第二条报告链路。

## 验证

- `npm run lint`
- `npm run format:check`
- `npm --workspace @maka/desktop test` — 1063 项通过
- 默认 Host 时序及超长 browser report 定向测试 — 5 项通过

## AI 使用

OpenAI Codex 协助了实现、测试、范围复核、验证和 PR 文案；贡献者本人仍对本次变更负责。AI 使用选项已在英文部分勾选。

## 检查清单

- 测试覆盖本次变更，并会在缺少实现时失败。
- lint、format、typecheck 和受影响测试均已在本地通过。
- 本 PR 包含行为变化，已在概要中说明。

</details>

### DIFF
--- a/apps/desktop/src/renderer/app-shell-skill-actions.ts
+++ b/apps/desktop/src/renderer/app-shell-skill-actions.ts
@@ -10,6 +10,7 @@ import { openSkillFailureCopy } from './app-shell-copy';
 import { createOpenSkillAction } from './app-shell-open-skill-action';
 import {
   defaultRuntimeHostDiagnosticTarget,
+  runIfDefaultRuntimeHostCurrent,
   runOnDefaultRuntimeHost,
 } from './default-runtime-host-operation.js';
 import { getShellCopy, localizedShellErrorMessage } from './locales/shell-copy.js';
@@ -84,7 +85,7 @@ export function createAppShellSkillActions(deps: {
   async function refreshSkills(options: { shouldShowError?: () => boolean } = {}) {
     try {
       const next = await runOnDefaultRuntimeHost((host) => window.maka.skills.list(host));
-      setSkills(next.value);
+      await runIfDefaultRuntimeHostCurrent(next.host, () => setSkills(next.value));
     } catch (error) {
       if (options.shouldShowError?.() ?? true) {
         reportRuntimeHostError(copy.refreshSkillsFailedTitle, copy.refreshSkillsFallback, error);
@@ -95,7 +96,7 @@ export function createAppShellSkillActions(deps: {
   async function refreshManagedSkillSources(options: { shouldShowError?: () => boolean } = {}) {
     try {
       const next = await runOnDefaultRuntimeHost((host) => window.maka.skills.sources.list(host));
-      setManagedSkillSources(next.value);
+      await runIfDefaultRuntimeHostCurrent(next.host, () => setManagedSkillSources(next.value));
     } catch (error) {
       if (options.shouldShowError?.() ?? true) {
         reportRuntimeHostError(copy.refreshSourcesFailedTitle, copy.refreshSourcesFallback, error);
@@ -106,7 +107,7 @@ export function createAppShellSkillActions(deps: {
   async function refreshBundledSkillCatalog(options: { shouldShowError?: () => boolean } = {}) {
     try {
       const next = await runOnDefaultRuntimeHost((host) => window.maka.skills.catalog.list(host));
-      setBundledSkillCatalog(next.value);
+      await runIfDefaultRuntimeHostCurrent(next.host, () => setBundledSkillCatalog(next.value));
     } catch (error) {
       if (options.shouldShowError?.() ?? true) {
         reportRuntimeHostError(copy.refreshBundledFailedTitle, copy.refreshBundledFallback, error);
--- a/apps/desktop/src/renderer/error-boundary.tsx
+++ b/apps/desktop/src/renderer/error-boundary.tsx
@@ -7,6 +7,7 @@
 // crashed early just left the user staring at an empty viewport).
 
 import { Component, type ErrorInfo, type ReactNode } from 'react';
+import { truncateUtf8 } from '@maka/core/diagnostic-log';
 import type { UiLocale } from '@maka/core/ui-locale';
 import { ICON_SIZE, AlertTriangle, Check, Clipboard, RotateCw } from '@maka/ui/icons';
 import { Button as UiButton, Card, redactSecrets } from '@maka/ui';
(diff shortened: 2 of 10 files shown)


jackwener (dismissed):
Automated review of exact head `03c2220bf43b6815d554ff62883f6416f79bb846` against `main@f1f4e71a9540a4da23159052c924fee72202e989`.

One actionable stale-state race remains in the Project path. `refreshDefaultProjectPresentation` checks Host identity at line 99 and then invokes the asynchronous `refreshDefaultProjectState(host)`. That function awaits `projects.getDefaultContext(host)` and only then calls the React setters. If A is current at line 99, the default switches to B while A's context read is pending, and A finishes last, A still commits stale Projects/capabilities/app info/selection. The new helper test uses a synchronous commit callback, so it cannot expose this gap.

Put the Host identity check at the actual commit boundary: split the Project read from its setters and validate after the read, or pass a completion guard into `refreshDefaultProjectState` and evaluate it immediately before its setter group. Add the corresponding ordering test with A's `getDefaultContext` held pending, B committed first, and A completed last.

Required conclusions:

1. Optimal for the actual problem: **yes for Skills/sources/catalog/scheduled tasks and the bounded browser report; not yet for Projects**.
2. Production code to delete: the duplicated identity check was correctly centralized; none further identified.
3. Tests to delete/replace: none; replace or extend the generic synchronous guard test with the real asynchronous Project commit race.
4. Deeper refactor: no broad refactor; keep one shared identity comparison, but ensure the Project fetch/commit split makes the authority check atomic with presentation commit.
5. Ready to merge: **no** while the Project race remains; the PR is also Draft. The exact-head required `test` check is green.
6. Residual risks: the Electron diagnostics path is unchanged; browser fallback redacts before UTF-8 field/final truncation and stays within 32 KiB. No additional implementation gap was found there.

This changes user-visible Host-backed presentation and crash-report behavior. Independent human review is required; this automated review is not approval.

M4n5ter:
Addressed in 8f896ea1e. Every default Project refresh now captures its Host, and identity is revalidated after `getDefaultContext` returns, immediately around the generation check and setter group. The regression test covers A pending → switch to non-ready B → A completes and is discarded. Desktop tests pass 1063/1063.

apps/desktop/src/renderer/app-shell-skill-actions.ts
```diff
@@ -84,7 +85,7 @@ export function createAppShellSkillActions(deps: {
   async function refreshSkills(options: { shouldShowError?: () => boolean } = {}) {
     try {
       const next = await runOnDefaultRuntimeHost((host) => window.maka.skills.list(host));
-      setSkills(next.value);
+      await runIfDefaultRuntimeHostCurrent(next.host, () => setSkills(next.value));
```
Astro-Han, line 88: [P2] The Host-identity check prevents Host A from committing after a switch to Host B, but two refreshes on the same Host still both pass it. If refresh A stalls, refresh B commits newer data, and then A returns, A overwrites B here; the same pattern exists for sources, catalog, and scheduled tasks. Could each projection use a monotonic latest-request generation in addition to the Host identity check, with an A pending → B commits → A settles test? That extends this seam without introducing another state authority.

apps/desktop/src/renderer/error-boundary.tsx
```diff
@@ -18,20 +19,36 @@ type State = {
...
+    );
   }
   if (typeof window !== 'undefined' && window.location?.href) {
-    lines.push(`Location: ${window.location.href}`);
+    lines.push(
+      `Location: ${boundedRendererField(window.location.href, RENDERER_LOCATION_MAX_BYTES)}`,
+    );
   }
-  return redactSecrets(lines.join('\n'));
+  return truncateUtf8(
```
Astro-Han, line 47: [P2] This bounded browser formatter is still reached only when `window.maka.diagnostics` is absent. In Electron, if the bridge exists but rejects because main IPC is not ready/restarting, the outer catch goes straight to `failed`, so the exact recovery path this fallback supports cannot copy anything. Could bridge rejection fall through once to this formatter plus `navigator.clipboard`, with a test for “bridge present and rejects”? That reuses the existing fallback rather than creating another diagnostics path.

Astro-Han (commented):
Thanks for addressing the cross-Host stale projection and for making the browser report bounded before redaction/copy. The latest head fixes those original paths cleanly. I rechecked the exact head and kept two same-authority recovery gaps inline; both can be closed by extending the seams already introduced rather than adding another path.

AI-assisted review disclosure: OpenAI Codex performed an independent exact-head re-review. I verified the latest delta, the retained race and fallback paths, current-head CI, reviews, and mergeability, and I made the final review decision.

apps/desktop/src/renderer/app-shell-skill-actions.ts
M4n5ter: Thanks — confirmed as a real same-Host latest-request ordering gap, but it predates and is separate from this PR’s cross-Host identity contract. Tracked in #3477 with the affected projections and the A pending → B commits → A settles regression contract. Keeping this PR bounded.

apps/desktop/src/renderer/error-boundary.tsx
M4n5ter: Thanks. I’m keeping this path unchanged. When the Electron bridge is present, main remains the diagnostics and clipboard authority, and a `copyReport` rejection is untyped: it can represent input/formatting or clipboard failure, not only unavailable IPC. Automatically retrying through `navigator.clipboard` would create a second Electron path and could mask the original failure. Such a fallback needs a demonstrated IPC-unavailable path plus a typed retryable failure contract.

M4n5ter:
@Astro-Han Thanks for the re-review — both inline threads have been addressed and resolved, with the confirmed same-Host ordering gap tracked in #3477; could you please take another look when convenient?

Astro-Han (approved):
Thanks @M4n5ter — re-reviewed at exact head `8f896ea1efca43e21375ae51da2b4aa176fa289b`.

The Project stale-state race raised in the earlier review (against head `03c2220bf43b6815d554ff62883f6416f79bb846`) is resolved at this head, and the identity check now sits at the real commit boundary rather than merely earlier in the flow. No P0–P2 remains that is introduced by this PR.

**Why the earlier finding no longer applies**

- `use-project-context.ts` takes a `generation` before awaiting `projects.getDefaultContext(host)`, re-checks mounted state and generation after the read, and only then calls `runIfDefaultRuntimeHostCurrent(host, …)`.
- `default-runtime-host-operation.ts` re-reads the current default Host inside that helper and compares **both** `profileId` and `hostId`; if the default switched to B while A's read was pending, A's callback never runs.
- The callback additionally re-checks `generation`, so a superseded refresh for the *same* Host cannot commit either. The setter group runs synchronously inside that callback, so there is no further await between the last identity check and the commit.

The two guards are complementary: the Host comparison covers the cross-Host case, the generation counter covers the same-Host superseded case.

**The ordering tests are genuinely asynchronous**

- `use-project-context.test.ts` holds `getDefaultContext` pending for A via a deferred, switches the default to B, then releases A, and asserts A's projects/selection are never committed.
- `default-runtime-host-operation.test.ts` separately constructs B-commits-first / A-settles-last and asserts A returns `false` with only B committed.

This addresses the earlier concern that a synchronous commit callback could not expose the gap.

**Sibling paths**

Skills list/sources/catalog (`app-shell-skill-actions.ts`) and scheduled-task list (`app-shell-scheduled-task-actions.ts`) route through the same helper after their reads, so the cross-Host projection gap is closed consistently rather than only in the Project path.

**Residual boundary (not blocking, not introduced here)**

The shared helper compares Host identity only; it does not carry a per-projection request generation. A same-Host latest-request race therefore remains for skills/sources/catalog and scheduled-task refreshes. That is already commented on at this head and tracked in #3477, and the Project path is not affected because it carries its own generation counter.

**Verification**

Exact-head `test` SUCCESS; `mergeable=MERGEABLE`; `git diff --check` clean. The aggregate `CHANGES_REQUESTED` / `BLOCKED` state reflects the earlier review against the superseded head — I am not dismissing it, and it is for @jackwener to resolve.

---

PR #3466, opened 2026-08-22
fix(desktop): correct Agent Graph collapse chevron direction

### LINKED ISSUE #3407: bug(desktop): Agent Graph panel collapse chevron points up when expanded and down when collapsed
### What happened

In the Desktop Agent Graph panel, the header collapse toggle's chevron reads inverted: it points **up while the panel is expanded** and **down while the panel is collapsed** — the opposite of the dominant desktop convention (VS Code sidebar sections, GitHub file trees, macOS disclosure triangles), where an expanded section's chevron points down at the content it reveals and a collapsed one points up/right.

Observed mapping:

| Panel state | Chevron |
| --- | --- |
| Expanded (operators/results visible) | ▲ up |
| Collapsed (header only) | ▼ down |

**Root cause.** `apps/desktop/src/renderer/agent-graph-panel.tsx:419` renders a static `ChevronDown` icon on the toggle (`aria-expanded={!collapsed}` at line 420), and `apps/desktop/src/renderer/styles/agent-graph.css:52-54` rotates it 180° when `[aria-expanded="true"]` — i.e. when the panel is open:

```css
.maka-agent-graph-collapse-toggle[aria-expanded="true"] svg {
  transform: rotate(180deg);
}
```

So the rotation fires in the expanded state, producing open=up / closed=down.

**Expected.** The chevron should track the panel's visual state the way other desktop disclosures do: expanded → pointing down toward the visible content; collapsed → pointing up (or right). Swapping the rule to rotate when `aria-expanded="false"` (or flipping the base icon to `ChevronUp`) fixes it.

<details>
<summary>简体中文</summary>

在桌面端的 Agent Graph 面板中，标题栏折叠按钮的箭头方向是反的：面板**展开时箭头朝上**、**收起时箭头朝下**——与主流桌面惯例（VS Code 侧栏分组、GitHub 文件树、macOS 折叠三角）正好相反：展开的分组箭头应指向下方露出的内容，收起时朝上/右。

实际表现：

| 面板状态 | 箭头 |
| --- | --- |
| 展开（可见 operators / 结果） | ▲ 朝上 |
| 收起（只剩标题栏） | ▼ 朝下 |

**根因。** `apps/desktop/src/renderer/agent-graph-panel.tsx:419` 给按钮渲染了固定的 `ChevronDown` 图标（第 420 行 `aria-expanded={!collapsed}`），而 `apps/desktop/src/renderer/styles/agent-graph.css:52-54` 在 `[aria-expanded="true"]`（即面板展开）时把它旋转 180°。旋转发生在展开态，于是得到"展开朝上、收起朝下"。

**期望行为。** 箭头应与面板的视觉状态一致：展开时朝下指向可见内容；收起时朝上（或朝右）。把规则改为在 `aria-expanded="false"` 时旋转（或把基础图标换成 `ChevronUp`）即可修复。

</details>

### How to reproduce

1. Start Maka Desktop and run a task that spawns an Agent Graph (any graph with operators)
2. The Agent Graph panel appears above the composer, expanded by default — note the header toggle's chevron points **up**
3. Click the toggle — the panel collapses to its header row — the chevron now points **down**

<details>
<summary>简体中文</summary>

1. 启动 Maka Desktop，运行一个会生成 Agent Graph 的任务（任意带 operator 的图）
2. 输入框上方出现 Agent Graph 面板，默认展开——此时标题栏切换按钮的箭头朝**上**
3. 点击该按钮——面板收起只剩标题栏——箭头变为朝**下**

</details>

### Environment

- Maka version or commit: `57e08d834`
- OS and version: Linux
- Surface: Desktop
- Node.js version, if running from source: v26.3.0

The defect is not environment-specific: the icon and its rotation are static JSX/CSS with no platform dependence.

<details>
<summary>简体中文</summary>

- Maka 版本或 commit：`57e08d834`
- 操作系统及版本：Linux
- 使用面：Desktop
- 从源码运行时的 Node.js 版本：v26.3.0

该问题与环境无关：图标及其旋转是静态的 JSX/CSS，不依赖任何平台。

</details>

One caveat for whoever picks this up: Astryx's inline disclosure rows (`ChatToolCalls` / `ChatReasoning`, via `chevronExpanded = rotate(180deg)` on a `chevronDown` base) use the same open=up orientation as this panel, so the panel is at least self-consistent with the chat transcript. This report argues the panel — a region disclosure with a large reveal area, not an inline row — should follow the state-indicating convention instead; if the team decides the action-convention is intended product-wide, feel free to close.

<details>
<summary>简体中文</summary>

给认领者的一个说明：Astryx 的行内折叠行（`ChatToolCalls` / `ChatReasoning`，同样是在 `chevronDown` 基础上用 `chevronExpanded = rotate(180deg)`）与本面板一样采用"展开朝上"，所以面板至少与聊天记录内部一致。本报告认为面板属于区域级折叠（露出面积大），不是行内条目，应遵循状态指示惯例；如果团队认定"动作指示"是有意的产品级约定，可以直接关闭。

</details>

1625567290:
## Summary

- keep the Agent Graph chevron pointing down while the panel is expanded
- rotate the chevron only when aria-expanded is false so the collapsed state points up
- cover disclosure state, collapsed state, and controlled content synchronization

## Testing

- npm --workspace @maka/desktop run build:main
- node --test --test-name-pattern=AgentGraphPanel\ collapse apps/desktop/dist/main/__tests__/agent-graph-panel.test.js
- npm run build
- npm run lint
- npm run format:check
- npm run typecheck
- real Chrome computed-style verification for expanded and collapsed states

The concurrent Desktop suite reached 1059/1061. The unrelated Git Review and MCP runtime failures both passed when rerun independently.

Closes #3407.


apps/desktop/src/main/__tests__/agent-graph-panel.test.ts
```diff
@@ -288,6 +288,37 @@ async function renderPanel(
...
+    assert.ok(contentId);
+
+    assert.equal(toggle.getAttribute('aria-expanded'), 'true');
+    assert.equal(panel.getAttribute('data-collapsed'), 'false');
+    assert.ok(harness.container.querySelector(`#${contentId}`));
+
+    await act(async () => {
+      (toggle as HTMLElement).click();
+    });
+    assert.equal(toggle.getAttribute('aria-expanded'), 'false');
```
Astro-Han: [P3] This test pins the disclosure state and content, but not the actual chevron transform that regressed. A browser-level computed-style or small visual assertion would catch the selector drifting while all of these state assertions still pass. This is optional hardening and does not block this focused fix.

Astro-Han (approved):
Thanks — this is the smallest correct fix at the existing disclosure-state seam. I reviewed exact head `e3225248f02d675029971801a9385482fb15a002`; `aria-expanded`, rendered content, and the collapsed data attribute stay under one state owner, and the current-head `test` check is green. I found no P0–P2 issues; one optional visual-regression hardening note is inline.

AI-assisted review disclosure: OpenAI Codex performed an independent exact-head review. I verified the state/CSS mapping, focused tests, live review state, mergeability, and current-head CI, and I independently made the approval decision.

apps/desktop/src/main/__tests__/agent-graph-panel.test.ts
1625567290: Added a real Electron/Chromium computed-style smoke in apps/desktop/e2e/agent-graph-chevron.spec.ts, backed by apps/desktop/e2e/agent-graph-chevron-smoke.mjs. It loads the production CSS and verifies the expanded transform is none, the collapsed transform is matrix(-1, 0, 0, -1, 0, 0), then restores the expanded state. Local browser smoke passed 1/1, the panel suite passed 15/15, desktop typecheck passed, and the full branch CI passed: https://github.com/1625567290/maka-agent/actions/runs/32568735415

apps/desktop/e2e/agent-graph-chevron-smoke.mjs
```diff
@@ -0,0 +1,76 @@
...
+    show: false,
+    width: 320,
+    height: 200,
+    webPreferences: {
+      backgroundThrottling: false,
+      contextIsolation: true,
+      nodeIntegration: false,
+      sandbox: true,
+    },
+  });
```
Astro-Han: `[P2]` This is the finding I care most about, and it holds independently of the CI failure.

The page under test is a hand-written `<button>` plus `agent-graph.css` alone, injected into a `data:` URL. Nothing else from the app is present. So this asserts that the rule works **when it is the only rule**.

But the regression this is guarding against does not look like that. A chevron points the wrong way because the selector is wrong (what you just fixed), because another stylesheet later wins the cascade, or because the component stops emitting `aria-expanded` on that element. Only the first is visible here — and the first is already visible far more cheaply. The second and third are exactly what an end-to-end test is supposed to buy, and this one is blind to both: it never loads the real stylesheet set and never renders the real component.

So the test costs a full Electron process per run and covers the one failure mode that did not need it.

Separately, `matrix(-1, 0, 0, -1, 0, 0)` pins Chromium's serialization of the transform rather than the intent. `rotate(-180deg)` is the same picture and fails this assertion.

Astro-Han (commented):
Thanks for chasing this down — the chevron pointing the wrong way is one of those things that quietly makes a panel feel broken, and the actual fix is exactly right: the rotation belongs on the collapsed state, not the expanded one.

My comments are entirely about the 122 lines of test added around that one-line change, not about the change.

The short version is that CI is currently red on this PR, and what is failing is the new e2e itself — `electron e2e/agent-graph-chevron-smoke.mjs` exits non-zero in the Desktop e2e job. So as it stands the instrument added to protect a one-line CSS fix is the thing keeping the fix out of the tree. That is worth weighing directly rather than debugging around.

There is also a deeper problem with that smoke test that I would raise even if it were green, and it is the one I would most like you to look at: it loads `agent-graph.css` on its own into a bare `data:` URL with a hand-written `<button>`, so it verifies the rule **in isolation**. The regression it exists to catch — this rule losing to something else, or the markup drifting — happens in the real cascade with the app's full stylesheet and the real component. A test that cannot fail for the scenario it is guarding against is not paying for its cost, and this one costs a real Electron process per run.

`getComputedStyle(...).transform === 'matrix(-1, 0, 0, -1, 0, 0)'` also pins Chromium's matrix serialization rather than the intent. Rewriting the rule as `rotate(-180deg)` would be visually identical and would fail this assertion.

I am not asking you to add heavier coverage — the opposite. For a one-line CSS direction fix, I think the honest options are to keep the small jsdom assertion, or to accept that this class of bug is caught by looking at it. Both seem better than an Electron process that is red today and blind to the real failure mode.

No `[P0]`/`[P1]` — the product change is correct.

Review assisted by AI (Claude Opus 5). Findings were verified against the files and the failing CI run at this head; the reviewer is accountable for them.

apps/desktop/e2e/agent-graph-chevron.spec.ts
```diff
@@ -0,0 +1,15 @@
+import { execFile } from 'node:child_process';
+import { createRequire } from 'node:module';
+import { promisify } from 'node:util';
+import { test, expect } from '@playwright/test';
+
+const execFileAsync = promisify(execFile);
+const electronPath = createRequire(import.meta.url)('electron') as string;
+
+test('agent graph collapse chevron follows the disclosure state', async () => {
```
Astro-Han: `[P2]` CI is red on this head and this is what is failing:

```
test	Desktop e2e	Error: Command failed: .../node_modules/electron/dist/electron e2e/agent-graph-chevron-smoke.mjs
  1 failed
```

A real `BrowserWindow` on a headless CI runner is a materially different environment from your machine, and `show: false` does not remove the dependency on a working display/GPU path. Whatever the specific cause, the situation is that a one-line CSS fix cannot land because of the test written to protect it.

Worth stating plainly since it cuts against the usual instinct: dropping this spec is a legitimate resolution here, not a retreat. The cost of the instrument should stay proportional to what it protects, and one CSS declaration does not justify an Electron process in the critical path of every CI run.

apps/desktop/src/main/__tests__/agent-graph-panel.test.ts
```diff
@@ -288,6 +288,37 @@ async function renderPanel(
   return harness;
 }
 
+describe('AgentGraphPanel collapse', () => {
+  it('keeps the disclosure state and controlled content in sync', async () => {
```
Astro-Han: `[P3]` This test is well written — it drives the real component through the existing harness and asserts `aria-expanded`, `data-collapsed`, and that the `aria-controls` target actually mounts and unmounts. If I had to keep exactly one of the three tests in this PR, it would be this one.

The note is only that none of the behaviour it covers is changed by this PR. Disclosure-state syncing worked before and works after; the diff to the product is one CSS selector. So a reviewer reading `+122/-1` has to work out that 121 of those lines are unrelated coverage.

Not asking you to remove it — new coverage of untested behaviour is worth having. Just noting that it belongs in its own change, where it can be judged on its own merits rather than riding along with a chevron fix.

apps/desktop/e2e/agent-graph-chevron-smoke.mjs
```diff
@@ -0,0 +1,76 @@
...
+      true,
+    );
+
+  try {
+    const expanded = await transformFor(true);
+    const collapsed = await transformFor(false);
+    const expandedAgain = await transformFor(true);
+    if (
+      expanded !== 'none' ||
+      collapsed !== 'matrix(-1, 0, 0, -1, 0, 0)' ||
```
Astro-Han: `[P3]` `expandedAgain` cannot differ from `expanded`. Between the two reads the only thing that changed is the attribute, which is set explicitly each time, and there is no component, no state, and no transition (`--duration-quick` is pinned to `0s` above) that could carry anything across.

An assertion that cannot fail reads as coverage without being coverage.

apps/desktop/e2e/agent-graph-chevron-smoke.mjs
1625567290: Agreed. Removed the isolated Electron smoke and its Playwright wrapper. The PR no longer launches a separate Electron process or asserts Chromium transform serialization; the final product diff is only the corrected CSS selector.

apps/desktop/e2e/agent-graph-chevron.spec.ts
1625567290: Removed this spec and the child Electron process that caused the Linux SUID sandbox failure. The focused CSS fix no longer adds a real-window test to the critical E2E path.

apps/desktop/src/main/__tests__/agent-graph-panel.test.ts
1625567290: Agreed. Removed the unrelated disclosure-state test as well. The final PR now contains only the one-line CSS selector correction, so the existing component behavior is no longer bundled into this change.

apps/desktop/e2e/agent-graph-chevron-smoke.mjs
1625567290: Removed with the rest of the isolated browser smoke. There is no longer an `expandedAgain` assertion or synthetic page in this PR.

1625567290:
Rebased onto current `main` and reduced the PR to the one-line CSS selector correction. The isolated Electron smoke, its Playwright wrapper, and the unrelated disclosure-state test have been removed, addressing the four review threads and eliminating the Linux SUID sandbox failure path.

Local validation: Desktop workspace dependencies and main process build passed, the existing Agent Graph panel suite passed 14/14 before the unrelated added case was removed, Desktop typecheck passed, and lint, format, ASF header, and diff checks are clean. New CI is now pending on head `608251142`.

1625567290:
Update: the rebased head is mergeable and has no unresolved review threads. Its new CI run is currently `action_required`, so it needs maintainer approval before the replacement check can execute.

PR #3467, opened 2026-08-22
fix(cli): let the TUI wizard create custom relay connections

### LINKED ISSUE #3405: fix(cli): TUI onboarding cannot create custom relay connections (OpenAI Chat / OpenAI Responses / Anthropic) — Desktop can
## What happened

In the TUI setup wizard (`/setup` and first-run), the three custom relay providers never appear in the provider picker, so there is no way to create these connections from the TUI:

- Custom relay (OpenAI Chat-compatible) — `openai-compatible`
- Custom relay (OpenAI Responses) — `openai-responses-compatible`
- Custom relay (Anthropic) — `anthropic-compatible`

The Desktop Settings → Models page lists all three and can create them (its add form collects the base URL for `category === 'custom'` providers). The TUI is the only surface where they are unreachable.

## Root cause

Two gaps, one visible and one structural:

1. **The TUI catalog filter drops every provider without a built-in base URL.** `listApiKeyOnboardableProviders()` maps the registry and then filters with `.filter((provider) => !provider.requiresBaseUrl)` (`packages/cli/src/onboarding-catalog.ts:20`), where `requiresBaseUrl: !definition.baseUrl` (`:16`). The three relay entries declare `baseUrl: ''` precisely because the user must supply their own endpoint (`packages/core/src/provider-registry.ts:1724`, `:1743`, `:1762`), so all three are filtered out of both TUI paths: the standalone fallback list (`packages/cli/src/pi-tui-runner.ts:1765`) and the runtime-host projection (`packages/cli/src/runtime-host-onboarding.ts:81`).

2. **The onboarding protocol has no field to carry a base URL anyway.** The wizard only collects an API key and an enabled-model set; `OnboardingVerifyInput`/`OnboardingSaveInput` (`packages/cli/src/pi-tui-contracts.ts:37-51`) and the host-side `ConnectionOnboardingVerifyInput`/`ConnectionOnboardingSaveInput` (`packages/runtime-host/src/protocol/connection-effects.ts:67-74`) carry only `{ providerType, apiKey, enabledModelIds }`. Host-side discovery builds `transientConnection(input.providerType)` with no endpoint override (`packages/runtime-host/src/server/connection-effect-coordinator.ts:220-224`), so even if the relays were listed, verify would run model discovery against an empty base URL and fail.

So this is not just "add three rows to the picker": the TUI wizard needs a base URL step for `requiresBaseUrl` providers, plumbed through the CLI contracts and the runtime-host onboarding operations.

## Proposed boundary

1. Stop filtering `requiresBaseUrl` out of `listApiKeyOnboardableProviders()`; keep the flag so the wizard knows which providers need an endpoint.
2. Add a base URL step to `OnboardingWizard` between provider pick and API key for entries with `requiresBaseUrl` (validate non-empty URL; reuse Desktop's normalization rules for relay base URLs).
3. Extend `OnboardingVerifyInput`/`OnboardingSaveInput` and the runtime-host `connection.onboarding.verify`/`save` inputs with an optional `baseUrl`; thread it into `transientConnection` and `commitConnectionOnboarding` so the saved connection persists it.
4. Keep the existing single-provider-slug model: saving a relay updates the same derived slug connection Desktop manages, so both surfaces stay in sync.

## How to reproduce

1. Run the `maka` TUI (first run, or `/setup` in-session).
2. Scroll the provider picker: OpenAI, Anthropic, aggregators, local servers are listed — no "Custom relay" entries.
3. Open Desktop → Settings → Models → Add provider: all three custom relay types are present and can be created with a base URL + API key.

## Environment

- Maka commit: `57e08d834`
- Surface: CLI TUI (`pi-tui-runner` onboarding wizard), compared against Desktop settings
- OS: Linux

UncertaintyDeterminesYou4ndMe:
## Summary

The TUI setup wizard (`/setup` and first-run) could not create any of the three custom relay providers — Desktop could. Two gaps, as diagnosed in the issue: the catalog filtered out every provider without a built-in base URL, and the onboarding protocol had no field to carry an endpoint even if they were listed. Per the archaeology in the issue thread, the filter was an explicit phase-1 scope cut in #1254 whose phase-2 base-URL step never landed.

This implements the issue's proposed boundary:

1. `listApiKeyOnboardableProviders()` now lists the `category: 'custom'` relays; `requiresBaseUrl` tells the wizard to collect an endpoint. Providers whose endpoint is derived rather than user-supplied stay excluded — `cloudflare-workers-ai` also has an empty registry `baseUrl`, but it interpolates an account id into a URL template, and Desktop deliberately keeps it out of the base-URL field too; a naive `!baseUrl` unfilter would have offered it a relay prompt it cannot use.
2. The wizard gains a base-URL step between provider pick and API key for `requiresBaseUrl` providers (relays show a 4-step flow, everything else keeps its 3 steps). Input is validated in place with the same rules the Host enforces (http/https, no credentials/query/fragment, 2048-byte cap), so mistakes fail with a readable message instead of a protocol decode error. Blank input is allowed only when a connection already exists — it reuses the persisted endpoint, mirroring the blank-key-reuses-stored-secret pattern.
3. `connection.onboarding.verify`/`save` carry an always-present `baseUrl: string | null` (exact-record wire style, like `apiKey`). A non-null value goes through the shared catalog normalizer (`normalizeCatalogConnectionBaseUrl`; a provider-default value collapses to null). A relay with no endpoint from input, existing connection, or registry is rejected with the new `base_url_not_configured` reason before any network probe. Since the input shapes and result unions are closed wire schemas, `RUNTIME_HOST_COMPATIBILITY_EPOCH` moves to 37 so a mixed pair fails the handshake instead of tearing down the session on the first `/setup`.
4. Discovery runs against the supplied endpoint, and `commitConnectionOnboarding` persists it: the journaled onboarding intent gains the field (allowed-but-not-required on read, so an intent journaled by an older build still replays as "no override"), and `prepareOnboardingUpsert` resolves `input ?? previous ?? registry default`. Two consistency rules ride along: the no-change early-return also compares the base URL so a URL-only change still commits, and a swapped endpoint drops `relayModelProfiles` and the last test result — the same endpoint-keyed contract `applyConnectionUpdate` already enforces, so a new relay does not inherit capability declarations or a "verified" badge from the relay it replaced.

Saving a relay updates the same derived-slug connection Desktop manages, so both surfaces stay in sync.

Fixes #3405

## Verification

- `npm --workspace @maka/runtime-host run test` — 1064 pass / 0 fail
- `npm --workspace @maka/storage run test` — 864 pass / 0 fail (14 pre-existing skips)
- `npm --workspace maka-agent run test` (CLI) — 358 pass / 0 fail
- `npm --workspace @maka/desktop run typecheck` + tests — clean, 1038 pass (desktop consumes the changed protocol types; it has no onboarding-operation callers)
- `biome check` on all changed files — clean; `knip` output byte-identical to `main`

New tests (kept deliberately light — five additions, each pinning one contract):

- protocol: verify/save round-trip with the new field, plus two invalid-URL negatives; the epoch-37 pin alongside the existing epoch ladder
- coordinator, end to end through real stores: a relay with no endpoint is rejected `base_url_not_configured` before any probe; a supplied endpoint reaches discovery and persists; a blank re-verify reuses the persisted endpoint; a re-onboarding that swaps the URL persists it and drops the old relay's profile table
- storage: the onboarding-intent journal round-trips the endpoint, and a journal written by a build that predates the field still replays (crash-recovery compatibility)
- TUI runner, end to end: filtering to a relay inserts the base-URL step (`2/4`), a malformed endpoint is rejected in place, and the endpoint threads through both `verify` and `save`

An independent adversarial review pass probed the change with running experiments before submission; it surfaced five defects — the missing compatibility-epoch bump, the `cloudflare-workers-ai` scope leak, stale endpoint-keyed state surviving a URL swap, a missing local byte cap, and two formatter misses — all fixed and re-verified above, with its blank-reuse/crash-replay/normalizer/state-machine probes coming back clean. Details in the review-record comment on this PR.

## AI use

Select exactly one:

Tool(s) and scope: Claude Code implemented the fix across the protocol/coordinator/storage/TUI layers, wrote the tests, and ran the verification; an independent adversarial review pass (also Claude) probed the change with fault-injection experiments and its findings were fixed before submission. I reviewed and verified the result.

## Checklist

- [ ] No


UncertaintyDeterminesYou4ndMe:

An independent review pass probed the change with running experiments (driven wizard sessions against the built TUI, storage-level probes through the real document stack, protocol decode drills, dist fault injections). It found **five defects, all fixed in the submitted head**, plus confirmations:

### Defects found and fixed

1. **Compatibility epoch not bumped (blocker).** `baseUrl` is required on two closed request shapes and the result unions gained a rejection reason. It demonstrated both mixed-pair directions dying: the frame decode throws inside the session pump and tears the whole Host session down on the first `/setup`. Fixed: `RUNTIME_HOST_COMPATIBILITY_EPOCH = 37`, pinned by a protocol test alongside the existing epoch ladder.
2. **`cloudflare-workers-ai` leaked into the wizard.** Four providers have an empty registry baseUrl, not three — Cloudflare's endpoint is an account-id template that Desktop deliberately keeps out of its base-URL field. Verified by driving the real wizard: it landed on a relay prompt it cannot use. Fixed: the catalog lists empty-baseUrl providers only when `category === 'custom'`; verified the offered set is now exactly the three relays.
3. **A swapped endpoint kept endpoint-keyed state.** Re-onboarding relay-a → relay-b retained `relayModelProfiles` and a "verified" `lastTest` — state the `applyConnectionUpdate` path explicitly drops on endpoint change (its comment: the old table must not outlive the relay it described). Demonstrated side by side through the real storage stack. Fixed: `prepareOnboardingUpsert` derives `endpointChanged` and feeds it into both the profile branch and the test-basis reset; pinned by an end-to-end coordinator test.
4. **Local URL validation omitted the 2048-byte cap**, so an oversized endpoint sailed past the wizard and hit the host's frame decode (fatal per defect 1). Fixed in the wizard's validator.
5. **Two formatter misses** the author's earlier check pipeline had swallowed. Fixed.

It also flagged the storage half as untested — two dist fault injections (removing the URL-change commit condition; making the intent field required) survived every existing suite. The submitted head adds a storage intent-journal test (round-trip + legacy replay) and the endpoint-swap coordinator assertions to close the reachable half of that gap.

### Confirmed safe (with evidence)

- Blank-reuse and crash-recovery replay: a blank save preserves the persisted URL; a hand-written legacy journal (no `baseUrl` key) replays and applies its model change without touching the URL.
- URL replacement probes the new endpoint while reusing the stored secret, and persists at the next revision.
- Normalizer boundary: empty/whitespace → null; query/fragment/credentials/ftp/file/javascript/non-string/oversize all rejected; provider-default collapses to null; decode idempotent; the OAuth-override throw is unreachable (OAuth providers never pass `providerAuthSupportsApiKey`).
- Wizard state machine: 2/4→3/4→4/4 labels; Esc walks exactly one level (models→key→baseUrl→search); a re-entered URL is what reaches verify and save; errors clear on typing; a late verify result after Esc is dropped; a non-relay picked after abandoning a relay carries no leaked URL.
- Rejection ordering: a blank key wins over a missing endpoint, and no discovery probe is issued on any rejection.
- Bypassing the coordinator guard cannot poison the catalog (an endpoint-less relay entry is unusable but loadable).
- Intent `schemaVersion` staying at 1 is correct: strict-equality versioning would break the crash-then-upgrade replay this change is designed to support, while a downgrade already fails closed on the unknown field.

Two observations left as-is, for the record: the protocol accepts an endpoint override for any api-key provider (Desktop's settings already allow the same, and the TUI only offers the step for relays); `hasConnection` is slug-derived, so a Desktop-created relay under a custom slug reads as new in the wizard — pre-existing design, noted since relays are now listed.

packages/cli/src/runtime-host-onboarding.ts
```diff
@@ -35,7 +36,8 @@ export function createRuntimeHostOnboardingSurface(
       try {
         const result = await connection.request('connection.onboarding.save', {
           providerType: input.providerType,
-          apiKey: normalizedSecret(input.apiKey),
+          apiKey: trimmedOrNull(input.apiKey),
+          baseUrl: trimmedOrNull(input.baseUrl),
```
Astro-Han, line 66: [P2] The new relay URL is saved by provider type, but `projectProviders()` recognizes an existing connection only at the derived canonical slug. A relay created in Desktop with a valid custom slug is therefore shown as unconfigured in `/setup`; saving here creates a second canonical-slug connection instead of editing/reusing the existing one, leaving the old credential and default target behind. Could onboarding carry the actual existing `connectionId`/slug (or preserve the sole connection for this provider type) through verify/save, with a Desktop custom-slug → TUI re-onboard test? That keeps connection identity authoritative instead of deriving a second one.

Astro-Han (commented):
Thanks for extending the existing Runtime Host onboarding seam instead of adding a CLI-only connection path. The protocol, validation, journal replay, endpoint-change cleanup, and TUI lifecycle all line up well on this head. One existing-connection identity gap becomes user-visible for custom relays, noted inline.

AI-assisted review disclosure: OpenAI Codex coordinated two independent exact-head review passes. I verified the retained connection-identity path, current CI state, reviews, and mergeability, and I made the final review decision.

packages/cli/src/runtime-host-onboarding.ts
UncertaintyDeterminesYou4ndMe: Implemented in de626752e — connection identity is now authoritative end to end:

- `projectProviders()` resolves the existing connection as the canonical-slug match first, else the provider type's **sole** connection whatever its slug (with several non-canonical connections there is no honest single answer, so the wizard offers a fresh canonical setup); the entry carries its `connectionId`.
- The wizard threads that identity through `verify`/`save` (`connectionId: string | null` on the wire, exact-record like `apiKey`), the coordinator targets it directly — a stale id is rejected as the new `connection_not_found` instead of silently duplicating — and `prepareOnboardingUpsert` finds the row by identity first, preserving its slug. Old credential and default target stay attached because it is the same connection.
- Tests: the Desktop-custom-slug → TUI re-onboard case you asked for (blank re-verify reuses the stored secret **and** the custom-slug endpoint; save with a new URL keeps the id and `my-relay` slug with exactly one connection in the catalog; a stale id rejects), plus projection unit tests (sole-of-type resolves, multiple non-canonical → none, canonical wins) and protocol negatives.
- One consequence caught in pre-push adversarial review: epoch 37 was minted earlier on this same PR for the `baseUrl` shape, and builds of that head exist from this review cycle — so the identity field takes **epoch 38** rather than reusing 37 for a second mutually-undecodable frame.

UncertaintyDeterminesYou4ndMe:
## Adversarial review record — identity-targeting increment (de626752e)

The connection-identity change was probed by an independent review pass with experiments against the real storage stack before pushing. Two defects found and fixed in the pushed head, one pre-existing issue documented:

**Fixed:**

1. **Epoch 37 reuse (blocker).** The earlier commit on this PR minted epoch 37 for the `baseUrl` shape; adding required `connectionId` under the same 37 would let two mutually-undecodable frame shapes handshake successfully and then die at frame decode — exactly what the epoch exists to prevent, and epoch-37 builds exist from this PR's own review cycle. The identity shape now takes **epoch 38**, pinned by its own protocol test.
2. **`connection_not_found` rendered as a key error.** The runner wrapped every verify rejection in "API key 验证失败：…请检查后重新输入" — self-contradictory for a rejection retyping can never fix. Stale-snapshot rejections now render unwrapped.

**Verified safe (with running experiments):**
- By-id upsert with a canonical-slug squatter of another provider type: the custom relay edits in place, the squatter is untouched — strictly better than the previous `slug_conflict`. `defaultTarget` seeds at the edited connection and a pruned default releases to null rather than dangling.
- Concurrent delete mid-discovery → `connection_not_found`, orphan credential swept, **no stray connection created** — the exact duplicate this change exists to prevent.
- Wizard state: an abandoned relay pick leaks no `connectionId` into the next provider's verify; a disabled sole connection re-onboards consistently with the canonical path's existing semantics; retired provider types can never be latched (none are listed).
- Admission-lane key change (identity when supplied) aligns onboarding with `connection.models.fetch`'s existing keying; the storage lane serializes commits either way, and the CLI is the only `connection.onboarding.*` client.
- Protocol negatives: empty/whitespace/oversized/missing `connectionId` all reject (old-client shapes cannot decode past the epoch gate); the storage-internal `target_missing` never leaks onto the wire.

**Pre-existing, documented not fixed:** crash-recovery treats a non-convergent onboarding intent as fatal (`invalid_document` → the store cannot open). The review reproduced the same brick through an untouched path (canonical row deleted and re-created out-of-band while an intent names the old id), so the class predates this PR; reaching it requires a hard kill plus an out-of-band catalog edit. A follow-up could treat an id/slug-conflicting intent as obsolete and clear it instead of throwing — happy to file an issue if maintainers agree.

Astro-Han:

```
# resolve, then
git push --force-with-lease
```

---

UncertaintyDeterminesYou4ndMe:
Rebased onto current `main` (`e05308be7`) and force-pushed with lease; the branch is conflict-free (`d1fe0c826`).

Both conflicts were the compatibility epoch: `main` has moved to **39** (37 = catalog search term, 38 = retired `execute` mode, 39 = Client Capability progress), so this PR's two onboarding epochs collided. Resolution: since neither onboarding shape was ever published separately outside this PR's own review cycle, the `baseUrl` + `connectionId` input requirements and the `base_url_not_configured` / `connection_not_found` rejections now land together as **epoch 40**, with one floor test (`> 39`) added to the existing ladder in `protocol.test.ts` following main's floor-assertion convention.

Also picked up in the rebase: the new storage test file gained the ASF license header main's audit gate now requires (`scripts/asf-license-headers.mjs check` is clean on the head).

Local verification on the rebased head (Node 24.18): storage 900 pass / 0 fail, CLI 393/393, desktop 1243/1243 + typecheck clean, biome clean on the full diff vs main. runtime-host's onboarding/protocol suites pass; its `host-kernel` `bounded election does not launch a Candidate…` test is flaky on my machine (passes 1 in 3 with drifting actual values — `starting`/`recovering`; it exercises #3512's election deadline and this branch does not touch that code). The `@maka/ui` `composer-plus-menu` failure on my machine reproduces on sources byte-identical to main (this PR touches neither `packages/ui` nor `packages/core`), so both look environmental — CI on this head is the authority; ready for it to run.

UncertaintyDeterminesYou4ndMe:
Rebased again onto current `main` (`efddab2fb`) — `main` moved fast enough that the previous rebase went stale within hours. New head: `327ed2edf`, conflict-free.

Two things changed underneath:

1. **The compatibility epoch advanced to 41** on `main` (40 = queue entry mutations, 41 = compaction terminal outcome), so this PR's onboarding epoch moves again: **both onboarding shapes now land as epoch 42** (same single-epoch reasoning as before — neither was ever published outside this PR's review cycle).
2. **#3571 merged**, fixing #3566 (crash recovery now clears an obsolete onboarding intent instead of bricking the store). It composes cleanly with this PR — and the interaction is worth stating: #3571 detects the obsolete intent by the exact `Onboarding intent conflicts with the connection id` error, which this PR's identity-first upsert leaves reachable only in the replay-after-drift case (a journaled id that no longer exists while the canonical slug is held by a different identity) — exactly the case #3571 now clears. A live intent whose connection still exists is found by id first and replays onto its own slug, never hitting that error. The message text is preserved verbatim since it is now a detection contract.

Local verification on the new head: storage 901 pass / 0 fail (includes #3571's new recovery test alongside this PR's journal tests), CLI 405/405, runtime-host onboarding + protocol suites 65/65, biome clean on the full diff, ASF header audit clean. Ready for the CI approval on this head.

UncertaintyDeterminesYou4ndMe:
Rebased a third time — `main`'s epoch line is a hot spot and #3602 minted **42** (`provider_capacity` retry reason) a few hours after my last push, which collided with this PR's 42. New head `def0ab4ab`: the onboarding shapes now take **epoch 43**, comment ladder and floor test updated accordingly. One nuance worth flagging for review: this time git auto-merged the epoch *constant* line (both sides read `42`) and only conflicted on the comments — the numeric collision had to be caught and lifted by hand, which is exactly the mixed-shape hazard the epoch exists to prevent.

Verification on this head: clean rebuild, storage 901/0, CLI 406/406, runtime-host onboarding + protocol suites 65/65, biome clean on the full diff, ASF header audit clean.

@Astro-Han this PR only ever conflicts on the epoch line, and each cycle costs a maintainer CI approval — if it goes stale again before review, I'm happy to keep rebasing, but if you have a review window it may be cheaper to look at this head while it's fresh.

packages/runtime-host/src/server/connection-effect-coordinator.ts
```diff
@@ -196,8 +198,8 @@ export class HostConnectionEffectCoordinator {
   #saveOnboarding(
     input: ConnectionOnboardingSaveInput,
   ): Promise<OperationOutcome<'connection.onboarding.save'>> {
-    const slug = deriveConnectionSlug(input.providerType);
-    return this.#admit(slug, 'connection.onboarding.save', async () => {
+    const lane = input.connectionId ?? deriveConnectionSlug(input.providerType);
+    return this.#admit(lane, 'connection.onboarding.save', async () => {
       const prepared = await this.#discoverOnboarding(input);
```
M4n5ter, line 204: <details open>
<summary>English</summary>

[P2] Bind discovery and commit to the same connection revision, endpoint, and credential basis.

`#discoverOnboarding()` reads the connection and stored credential, then performs remote model discovery before `runMutation()` is acquired. It returns only the supplied-secret marker and discovered models. During that await, a supported catalog/vault mutation can change the same connection. The final commit re-reads the latest state, so `baseUrl: null` preserves the new endpoint and `suppliedSecret: null` preserves the new credential while committing models discovered from the old pair.

I reproduced this through the production coordinators: discovery paused after observing placeholder key A at relay A; a concurrent policy update moved the same connection to relay B and placeholder key B; discovery then returned `model-from-relay-a`. Save reported `saved`, but persisted relay B/key B with relay A's model inventory.

The admission lane covers only this connection-effect coordinator, and the activation gate covers only the final commit; neither revalidates the discovery basis. Please issue a ticket containing the connection revision, effective base URL, and credential basis before discovery, then atomically revalidate it at commit (the existing model-fetch/test ticket pattern is a suitable shape). Add this race as a regression test.

</details>

<details>
<summary>中文</summary>

[P2] 请把模型发现与最终提交绑定到同一个连接 revision、endpoint 和凭证依据。

`#discoverOnboarding()` 会先读取连接和已保存凭证，再在取得 `runMutation()` 之前执行远程模型发现；它返回时只带“是否提供新密钥”的标记和模型列表。在这段 await 期间，受支持的 catalog/vault 更新可以修改同一连接。最终提交会重新读取最新状态，因此 `baseUrl: null` 会保留新 endpoint，`suppliedSecret: null` 会保留新凭证，却把旧 endpoint/凭证发现出的模型一起提交。

我通过生产 coordinator 复现了这一点：模型发现在看到 relay A 和占位密钥 A 后暂停；并发 policy 更新把同一连接改到 relay B 并轮换为占位密钥 B；随后发现返回 `model-from-relay-a`。保存返回 `saved`，但最终持久化的是 relay B/密钥 B 加 relay A 的模型清单。

当前 admission lane 只覆盖这个 connection-effect coordinator，activation gate 也只覆盖最后一次提交；两者都不会重新验证发现依据。请在发现前签发包含连接 revision、实际 base URL 和凭证依据的 ticket，并在提交时原子复核（现有 model-fetch/test ticket 模式已经提供了合适形状）。同时把这个竞态加入回归测试。

</details>

M4n5ter (commented):
<details open>
<summary>English</summary>

Reviewed exact head `def0ab4ab9c690f42d080ed17d1bce03f8810ee2`: **NO-GO — 1×P1 + 1×P2**.

1. [P1] This head's epoch 43 is below current `main` epoch 44 even though the PR adds required fields and new values to closed onboarding wire shapes. The merge-result guard rejects base=44/head=43 as `went backward`; this change is not compatible-declaration eligible because old peers reject the new shapes.
2. [P2] Remote onboarding discovery is not bound to the connection revision, effective endpoint, and credential basis used by the final commit. A supported concurrent policy update can persist relay B/key B with the model inventory discovered from relay A/key A; the production-coordinator reproduction is documented inline.

The protocol/storage/TUI propagation is otherwise feature-proportional and keeps the Host as the persistent authority. The previous custom-slug identity issue is closed on this head; no additional decision-changing issue was found.

Verification: 205/205 focused affected tests passed; Core, Storage, Runtime, Runtime Host, MCP, Eval, and CLI builds passed; Biome and `git diff --check` passed.

Gate state: the exact-head hosted `test` is terminal red in an unrelated MCP timing test. The PR also conflicts with current `main`, so GitHub cannot synthesize a current merge ref and no fresh merge-result gate evidence can be obtained until rebase. Current `main` is independently red in a Storage compile failure; that baseline failure is not attributed to this PR.

</details>

<details>
<summary>中文</summary>

已审查 exact head `def0ab4ab9c690f42d080ed17d1bce03f8810ee2`：**NO-GO——1 条 P1 + 1 条 P2**。

1. [P1] 当前 head 的 epoch 43 低于当前 `main` 的 epoch 44，但本 PR 又给闭合 onboarding wire shape 新增了必填字段和新值。merge-result guard 会把 base=44/head=43 判定为 `went backward`；旧 peer 会拒绝这些新形状，因此本变更不能使用 compatible-change 声明。
2. [P2] 远程 onboarding discovery 没有绑定最终 commit 所依赖的连接 revision、实际 endpoint 和凭证依据。受支持的并发 policy 更新可以把 relay B/密钥 B 与 relay A/密钥 A 发现的模型清单一起持久化；生产 coordinator 复现已记录在行内 finding。

除此之外，协议、存储与 TUI 的传播复杂度与功能相称，并继续以 Host 作为持久化权威。上一轮 custom-slug 身份问题已在当前 head 关闭；没有发现其他足以改变合并判断的问题。

验证结果：受影响的 focused tests 205/205 通过；Core、Storage、Runtime、Runtime Host、MCP、Eval、CLI 构建通过；Biome 与 `git diff --check` 通过。

门禁状态：exact-head hosted `test` 因无关的 MCP 时序测试终态失败。PR 还与当前 `main` 冲突，因此 GitHub 无法生成当前 merge ref，在 rebase 前无法取得新的 merge-result 门禁证据。当前 `main` 自身也因 Storage 编译失败而红；这个基线失败不归因于本 PR。

</details>

packages/runtime-host/src/protocol/index.ts
```diff
@@ -91,7 +91,10 @@ export const RUNTIME_HOST_REGISTRATION_SCHEMA_VERSION = 1 as const;
 export const RUNTIME_HOST_PROTOCOL_VERSION = 0 as const;
 // Increment when the same protocol version no longer guarantees safe Client-Host
 // interoperability. Mismatches are rejected before domain commands are admitted.
-export const RUNTIME_HOST_COMPATIBILITY_EPOCH = 42 as const;
+export const RUNTIME_HOST_COMPATIBILITY_EPOCH = 43 as const;
```
M4n5ter: <details open>
<summary>English</summary>

[P1] Rebase and keep the compatibility epoch monotonic against current `main`.

This exact head advertises epoch 43 for required `connectionId`/`baseUrl` onboarding inputs and new closed rejection variants, while current `main` is already epoch 44. These are genuinely incompatible shapes: older exact-record decoders reject the added required fields and closed-union values, so a compatible-change declaration cannot exempt them.

The merge-result guard evaluates base=44/head=43 as `went backward` before it even considers changed files. If this downgrade landed, installations that have already observed epoch 44 would lose rolling compatibility with the new Host/Client pair, and reusing an older number would break the monotonic identity the handshake relies on. Please rebase onto current `main`, read its epoch then, and assign this change a strictly greater value with the corresponding floor test/comment. Do not preselect a fixed number before rebase.

</details>

<details>
<summary>中文</summary>

[P1] 请先 rebase，并保证 compatibility epoch 相对当前 `main` 单调递增。

当前 exact head 用 epoch 43 表示 onboarding 输入新增必填 `connectionId`/`baseUrl` 和新的闭合 rejection 变体，但当前 `main` 已经是 epoch 44。这些 shape 确实不兼容：旧的 exact-record decoder 会拒绝新增必填字段和闭合 union 值，因此不能用 compatible-change 声明豁免。

merge-result guard 在检查具体文件之前，就会把 base=44/head=43 判定为 `went backward`。如果这个降级落地，已经见过 epoch 44 的安装将无法与新的 Host/Client 组合进行滚动兼容；复用更小号码也会破坏握手依赖的单调身份。请先 rebase 到当前 `main`，读取当时的 epoch，再为本变更分配严格更大的值，并同步更新 floor test/comment。不要在 rebase 前预先认领固定号码。

</details>

packages/runtime-host/src/server/connection-effect-coordinator.ts
UncertaintyDeterminesYou4ndMe: Implemented in bbe7af770, following the model-fetch ticket shape you prescribed:

- `beginConnectionOnboarding` (storage, under the write lane) locates the target and issues a one-shot ticket whose basis pins the **connection revision** (covers the endpoint and every catalog-visible property), the **api-key credential status + stored secret from a single vault read**, and the **effective proxy with its credential** — and returns the pinned proxy so discovery runs through the exact egress the basis certifies (re-resolving it later would let a flip-and-restore slip past the check).
- `completeConnectionOnboarding` claims the ticket and revalidates the basis atomically before the durable intent is written: any drift returns `superseded` (new save rejection reason, riding this PR's unpublished epoch), a vanished target still reports `connection_not_found`, and a superseded save writes no journal. Verify abandons its WeakMap-held ticket; save begins its own. The direct `commitConnectionOnboarding` entry point is gone, so there is no unvalidated path left.
- Regression test drives your exact reproduction end to end: discovery paused after observing relay A/key A → concurrent supported catalog update to relay B + key rotation → save returns `rejected: superseded`, relay B/key B intact, relay A's inventory never persisted — then a retry commits cleanly through relay B/key B.

Pre-push adversarial verification (production coordinators, real stores) additionally confirmed: key-rotation-only (no revision bump) supersedes via the credential branch; credential deletion supersedes; first-create vs concurrent canonical-slug create supersedes without double-creating; two racing Host coordinators yield exactly one `saved` and one `superseded` with a single whole inventory; a replayed/forged ticket maps to a sane `invalid_request`, never a session teardown.

packages/runtime-host/src/protocol/index.ts
UncertaintyDeterminesYou4ndMe: Done in the rebased head: rebased onto current `main` first, read its epoch then (44), and assigned **45** to this change, with the floor test updated to `> 44` and the comment ladder preserving main's 43/44 entries. No number was preselected before the rebase — and noted for the record: during this rebase git auto-merged the epoch *constant* line once (both sides happened to read the same number) and only conflicted on comments, so the monotonicity check was done explicitly against main's value rather than trusting conflict markers.

UncertaintyDeterminesYou4ndMe:
Both findings addressed on head `bbe7af770` (rebased onto `main@3bb645e99`):

- **[P1]** epoch is now **45** (> main's 44), assigned after the rebase per your instruction; floor test and comment ladder updated.
- **[P2]** onboarding now uses a begin/complete ticket that binds discovery to the connection revision, credential basis, and pinned effective proxy, revalidated atomically at commit — details and the adversarial verification matrix in the inline reply; the reviewed race is regression-tested end to end.

Local verification: storage 908/0, runtime-host 1123/0, CLI 424/424, desktop typecheck clean, biome + ASF header audit clean. The branch is conflict-free against that main; ready for CI approval on this head.

packages/runtime-host/src/server/connection-effect-coordinator.ts
```diff
@@ -212,35 +215,48 @@ export class HostConnectionEffectCoordinator {
...
-      toRuntimePolicyProxy(proxy.networkProxy, proxy.secretMaterial.networkProxy?.secret),
+      toRuntimePolicyProxy(begun.networkProxy, begun.proxySecret ?? undefined),
     );
     try {
-      const effect = await this.#runModelDiscovery(
-        candidate ?? transientConnection(input.providerType),
-        secret,
-        { fetch: transport.fetch },
-      );
+      const effect = await this.#runModelDiscovery(base, secret, { fetch: transport.fetch });
```
Astro-Han: [P2] Onboarding discovery loses the connection's custom request headers and body overlay.

This path probes the provider with the bare transport fetch:

```ts
const effect = await this.#runModelDiscovery(base, secret, { fetch: transport.fetch });
```

The ordinary models path in this same file wraps the transport before handing it over (`:393-400`):

```ts
const requestHeaders = prepared.secretMaterial.requestHeaders
  ? parseRequestHeaders(prepared.secretMaterial.requestHeaders.secret)
  : {};
return await run(
  createRequestCustomizationFetch(transport.fetch, {
    headers: requestHeaders,
    bodyOverlay: prepared.connection.requestBodyOverlay,
  }),
  secret,
);
```

So a connection that authenticates through a custom header (or needs a body overlay) verifies and saves fine through `models.fetch` and the connection test, but fails during onboarding/re-onboarding: the probe goes out without the header and without the overlay. A run against the production coordinator with real stores reproduces it — the normal path carries the configured header and merges the overlay, the onboarding path carries neither and the provider rejects the request.

The existing re-onboarding tests only cover endpoint and API-key reuse, so nothing catches this. Suggest routing `#discoverOnboarding` through the same customization wrapper (using the ticket-pinned basis it already holds), plus a test with a header-authenticated connection.

<details>
<summary>中文</summary>

onboarding 的 discovery 直接用了裸的 `transport.fetch`，而同文件里正常的 models 路径（`:393-400`）会先套 `createRequestCustomizationFetch`，把自定义请求头和 `requestBodyOverlay` 带上。结果是：靠自定义头鉴权的连接在 `models.fetch` 和连接测试里正常，但走 onboarding/重新 onboarding 时会因为丢头、丢 overlay 而鉴权失败。现有的重新 onboarding 测试只覆盖了 endpoint 和 API key 的复用，抓不到这条。建议让 `#discoverOnboarding` 复用同一个包装（用它已经持有的 ticket 基准），并补一个自定义头鉴权连接的测试。
</details>

Astro-Han (commented):
Re-review at `bbe7af77`. One new **[P2]**, filed inline on `connection-effect-coordinator.ts`. Not blocking beyond that finding — everything previously raised is resolved at this head.

**Closed since the last pass** (verified at this head, threads resolved):
- Connection identity — `projectProviders()` now resolves a Desktop-created relay under a custom slug to the existing connection and plumbs `connectionId` through verify/save, with tests covering the sole-connection, multi-connection and canonical-wins cases.
- Discovery/commit basis — the begin/complete ticket pins revision, credential and proxy, so a concurrent policy change supersedes the save instead of pairing a new endpoint with an inventory it never produced.
- Compatibility epoch — head advertises 45 against current `main`'s 44, monotonic.

CI: `test` is terminal green on this exact head. Withholding approval only for the open [P2].

<details>
<summary>中文</summary>

在 `bbe7af77` 上复审。新增一条 **[P2]**，已作为行内评论提在 `connection-effect-coordinator.ts`。之前提过的问题在这个 head 上都已解决：连接身份（`projectProviders()` 现在能认出自定义 slug 下的 relay 并透传 `connectionId`，且有测试覆盖）、discovery/commit 基准（begin/complete ticket 绑定 revision、凭证与代理）、兼容 epoch（head 45 > 当前 main 44，单调）。对应三条旧 thread 已关闭。exact head 的 `test` 已终态绿；仅因这条未解决的 [P2] 暂不 approve。
</details>

packages/runtime-host/src/server/connection-effect-coordinator.ts
UncertaintyDeterminesYou4ndMe: Addressed on head `bc3d1e1d3`.

- `#discoverOnboarding` now probes through `createRequestCustomizationFetch`, exactly as `#withTransport` does: `beginConnectionOnboarding` pins the target's `request_headers` secret alongside the API key (one vault read), and the host applies it plus `base.requestBodyOverlay` — which also covers the baseUrl-override case, since `base` spreads the pinned catalog entry.
- The header credential's status joins the ticket's basis, so a rotation (including absent→present) between discovery and commit supersedes the save like an endpoint or key change would. The body overlay needs no separate pin — it is a catalog property covered by the pinned revision.
- The suggested test is in: a header-authenticated connection with a body overlay asserts the probe carries both (verify and save runs), and that rotating the header credential mid-save supersedes — with the catalog revision asserted unchanged, so the supersede provably comes from the new basis field.

Two scope notes, stated rather than hidden: this restores **re-onboarding** parity — a first-time relay that can only authenticate via a custom header still has no wizard field for one, same as before this PR. And an endpoint override sends the stored header credential to the newly typed host, mirroring the existing stored-API-key semantics of editing a connection in place.

One pre-existing gap surfaced while verifying parity: `modelFetchSemanticBasis` itself omits `requestBodyOverlayJson` (only the connection-test basis carries it), so an overlay edit between `beginModelFetch` and `completeModelFetch` does not supersede even though the probe applied it. Out of scope here — happy to file it as a follow-up.

Verification on this head: storage 908/0, runtime-host 1124/0 (new test included), biome clean.

Astro-Han (approved):
**APPROVE at exact head `bc3d1e1d3603040ab7993e395a90cc6e30512094`. The one remaining [P2] is closed; no P0–P3 open.**

**Scope of this pass**: confirmation that the request-customization [P2] from review `5004343688` is genuinely fixed, plus the gate state at this head. This is not a fresh full-surface re-review of all 19 files — the earlier passes covered that, and their findings are closed.

**The [P2] is closed, and verified at the source rather than taken on report.** That finding was that onboarding discovery passed the raw `transport.fetch`, bypassing the `createRequestCustomizationFetch(headers + bodyOverlay)` wrapper that the ordinary models path applies. At this head both paths construct the same wrapper:

- onboarding — `connection-effect-coordinator.ts:263-268`: `createRequestCustomizationFetch(transport.fetch, { headers: begun.requestHeadersSecret ? parseRequestHeaders(begun.requestHeadersSecret) : {}, bodyOverlay: base.requestBodyOverlay })`
- models — `:407-410`: the same call shape with `prepared.secretMaterial.requestHeaders` and `prepared.connection.requestBodyOverlay`

**The onboarding side is the stricter of the two, deliberately so.** It reads the header secret pinned on the ticket (`begun.requestHeadersSecret`) rather than re-resolving it, and the comment above it states why: a flip-and-restore between the two reads would otherwise pass the basis check. That reasoning also covers the proxy (`:251-254`). Header create/delete/rotation are checked for `superseded` before the commit, so a rotation landing mid-discovery is rejected rather than silently committed.

**What that means for the connections this finding was about**: a connection that authenticates through custom request headers now gets those headers on the onboarding discovery request, so re-onboarding no longer fails for it.

**Gates at this exact head**
- `test` — terminal `success` (run `32696317110`)
- `MERGEABLE`, non-draft, no stale approvals
- unresolved review threads — **0** (the last one, `PRRT_kwDOSpfFGs6bkiwe`, is resolved with this review; it was `isOutdated` but that only means the anchor line moved, so it was held open until the fix itself was checked)
- against current `main` `04836d3b`: the synthetic merge tree is clean and the changed files do not overlap `main`'s recent commits

**No merge from us.** Merging is a maintainer's call.

<details>
<summary>中文</summary>

在 exact head `bc3d1e1d` 上 **APPROVE**，唯一剩下的 [P2] 已闭合，无 P0–P3。

**本次范围**：只确认 review `5004343688` 那条 request-customization [P2] 是否真的修好，以及该 head 的门禁状态；不是对 19 个文件的重新全量审查——那部分由此前几轮覆盖，结论均已闭合。

**[P2] 已闭合，而且是回到源码核的，不是照报告采信。** 原问题是 onboarding discovery 传了裸的 `transport.fetch`，绕过了普通 models 路径套的 `createRequestCustomizationFetch(headers + bodyOverlay)`。当前 head 上两条路构造的是同一个包装：onboarding 在 `connection-effect-coordinator.ts:263-268`，models 在 `:407-410`。

**onboarding 那一侧比 models 更严，而且是刻意的**：它用 ticket 上钉住的 `begun.requestHeadersSecret`，不重新解析；上方注释写明了理由——两次读取之间的"翻转再还原"否则能绕过 basis 校验。同一理由也覆盖了 proxy（`:251-254`）。header 的新建/删除/轮换在 commit 前会判 `superseded`，所以 discovery 期间发生的轮换会被拒绝而不是静默提交。

**对这条 finding 关心的那类连接意味着什么**：靠自定义请求头鉴权的连接，现在 onboarding discovery 请求也会带上这些头，重新 onboarding 不再因此失败。

门禁：`test` 终态 success（run `32696317110`）；MERGEABLE、非 draft、无陈旧 approval；**未决线程 0**（最后一条 `PRRT_kwDOSpfFGs6bkiwe` 随本 review 关闭——它显示 outdated，但 outdated 只说明锚点行移动，所以一直挂着直到修复本身被核实）；对当前 main `04836d3b` 的 synthetic merge tree 干净、改动文件与 main 近期提交无重叠。

**我们不合并**，合并是 maintainer 的决定。
</details>

Astro-Han:
LGTM — merging. The wizard being able to create a custom relay connection closes a real gap; thanks for working through the review rounds on it.

<details>
<summary>中文</summary>

看过了，合并。向导能建自定义中转连接补上了一个实际缺口，感谢配合几轮修改。
</details>

PR #3469, opened 2026-08-22
fix(desktop): keep the composer ＋ menu still when toggling Plan

MatrixA:
## Summary

Clicking Plan in the composer's ＋ menu made the open menu visibly blink: the toggle re-fetches the invocable-Skill projection, the fail-closed clear leaves `mentionSkills` empty for the length of the IPC round trip, and the Skills row read that transient `[]` as "no skills available" — graying out and growing a description line, then snapping back. `useComposerMentions` now also returns a settled `mentionSkillsUnavailable` verdict that only moves when a request resolves; the Skills row renders from it, while the `/` popup keeps reading the fail-closed list. A second commit removes the pending-keyed disable that dimmed the Plan/Swarm/Graph rows and the footer mode marks to half opacity during a toggle's commit round trip — the pending registries already swallow re-entrant toggles, so the disable protected nothing and only blinked.

## Root cause

`use-composer-mentions.ts` clears the list while re-fetching on purpose (a visible `/` popup must never advertise a stale Skill), and `newSessionCollaborationMode` is a dependency of that effect, so `length === 0` could not tell "re-fetching" from "nothing to offer". The same blip fired in sessions via the `mode-change` event, and on model or turn changes.

## Verification

| Before — the menu jumps on the Plan click | After — the menu holds still |
| --- | --- |
| Clicking Plan makes the open ＋ menu grow a "no skills" line and jump | Clicking Plan toggles the checkbox with the menu geometry unchanged |

- `packages/ui`: 192/192 unit tests pass, including three new tests pinning the Skills row's presentation (refreshing ≠ empty; settled-empty keeps its caveat; populated stays quiet).
- `apps/desktop`: 1060/1060 main-process tests pass; `typecheck` and `biome lint` clean on every touched file.
- e2e: `composer-plus-menu-stability.spec.ts` now carries three specs. The stability spec arms an in-page MutationObserver before the click and asserts the menu stays open, keeps one height, and never shows the transient "no skills" line or a grayed row. Two review-round specs hold IPC in-flight windows open via a preload latch (isolated-E2E gate only): a Skills click mid-refresh does nothing and works once settled, and two rapid Plan toggles land on the last requested state. All three fail without their fix; adjacent composer specs (skill invocation, slash-command menu, send-message) still pass, 9/9 total.

## AI use

Select exactly one:

Tool(s) and scope: Claude Code (Claude Fable 5) diagnosed the flicker and authored the fix, unit tests and e2e spec; both commits carry `Generated-by: Claude Code` trailers.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/src/renderer/app-shell.tsx
+++ b/apps/desktop/src/renderer/app-shell.tsx
@@ -427,16 +427,13 @@ function AppShellContent({
   const [newChatPlanModeActive, setNewChatPlanModeActive] = useState(false);
   const [newChatOrchestrationMode, setNewChatOrchestrationMode] = useState<OrchestrationMode>('default');
   const [scheduledTaskCreateRequestNonce, setScheduledTaskCreateRequestNonce] = useState(0);
-  // Write-only: the pending-action helpers need a store, but nothing renders
-  // from these — a mode toggle's guard is the registry ref, and painting the
-  // round trip (disable/dim, then restore) is exactly the ＋ menu blink that
-  // MatrixA/fix-plan-click-flicker removed.
-  const [, setPendingCollaborationModeBySession] = useState<Record<string, boolean>>({});
-  const [, setPendingOrchestrationModeBySession] = useState<Record<string, boolean>>({});
   // The rows stay interactive while a commit runs, so a click landing in that
   // window is the user updating their mind — not noise to drop. Each map holds
   // only the LATEST ask per session; the in-flight commit's finally block
-  // applies it if the settled state does not already satisfy it.
+  // applies it if the settled state does not already satisfy it. Nothing
+  // renders from a mode commit's round trip — painting it (disable/dim, then
+  // restore) is exactly the ＋ menu blink MatrixA/fix-plan-click-flicker
+  // removed — so there is no pending state here, only the registry refs.
   const queuedCollaborationModeBySession = useRef(new Map<string, boolean>());
   const queuedOrchestrationModeBySession = useRef(new Map<string, OrchestrationMode>());
   const [newTaskPermissionChoice, setNewTaskPermissionChoice] =
@@ -958,25 +955,30 @@ function AppShellContent({
     return next;
   }
 
+  // The registry ref is the re-entrancy authority; the setter is only for
+  // actions whose pending state something actually renders (permission, model,
+  // message retry). Mode toggles pass none — their round trip is deliberately
+  // not painted, and a write-only state would still schedule a render per
+  // add/clear.
   function addPendingSessionAction(
     sessionId: string,
     pendingRef: { current: Set<string> },
-    setPendingBySession: (updater: (current: Record<string, boolean>) => Record<string, boolean>) => void,
+    setPendingBySession?: (updater: (current: Record<string, boolean>) => Record<string, boolean>) => void,
   ): boolean {
     if (pendingRef.current.has(sessionId)) return false;
     pendingRef.current.add(sessionId);
-    setPendingBySession((current) => ({ ...current, [sessionId]: true }));
+    setPendingBySession?.((current) => ({ ...current, [sessionId]: true }));
     return true;
   }
 
   function clearPendingSessionAction(
     sessionId: string,
     pendingRef: { current: Set<string> },
-    setPendingBySession: (updater: (current: Record<string, boolean>) => Record<string, boolean>) => void,
+    setPendingBySession?: (updater: (current: Record<string, boolean>) => Record<string, boolean>) => void,
   ): void {
     if (!pendingRef.current.has(sessionId)) return;
     pendingRef.current.delete(sessionId);
-    setPendingBySession((current) => omitSessionKey(current, sessionId));
+    setPendingBySession?.((current) => omitSessionKey(current, sessionId));
   }
 
   function clearSessionRendererState(sessionId: string): void {
@@ -984,9 +986,12 @@ function AppShellContent({
     turnActionRegistry.clearForSession(sessionId);
     permissionModeChangeRegistry.keysRef.current.delete(sessionId);
     collaborationModeChangeRegistry.keysRef.current.delete(sessionId);
-    setPendingCollaborationModeBySession((current) => omitSessionKey(current, sessionId));
     orchestrationModeChangeRegistry.keysRef.current.delete(sessionId);
-    setPendingOrchestrationModeBySession((current) => omitSessionKey(current, sessionId));
+    // Queued mode intents die with the Session's renderer lifecycle: an
+    // in-flight commit's finally would otherwise replay an old ask against a
+    // Session this cleanup has already let go of.
+    queuedCollaborationModeBySession.current.delete(sessionId);
+    queuedOrchestrationModeBySession.current.delete(sessionId);
     sessionModelChangeRegistry.keysRef.current.delete(sessionId);
   }
 
@@ -1055,7 +1060,6 @@ function AppShellContent({
     if (!addPendingSessionAction(
       sessionId,
       collaborationModeChangeRegistry.keysRef,
-      setPendingCollaborationModeBySession,
     )) {
       // A commit is in flight and the rows stay interactive (no pending
       // repaint), so this click is the user updating their mind, not noise.
@@ -1116,7 +1120,6 @@ function AppShellContent({
       clearPendingSessionAction(
         sessionId,
         collaborationModeChangeRegistry.keysRef,
-        setPendingCollaborationModeBySession,
       );
       // Whatever the user last asked for while this commit ran is the state
       // they expect to land on. Read the settled mode through the ref — the
@@ -1148,7 +1151,6 @@ function AppShellContent({
     if (!addPendingSessionAction(
       sessionId,
       orchestrationModeChangeRegistry.keysRef,
-      setPendingOrchestrationModeBySession,
     )) {
       // Same latest-intent contract as applyPlanMode: the rows stay
       // interactive while a commit runs, so keep the newest ask and apply it
@@ -1175,7 +1177,6 @@ function AppShellContent({
       clearPendingSessionAction(
         sessionId,
         orchestrationModeChangeRegistry.keysRef,
-        setPendingOrchestrationModeBySession,
       );
       const queued = queuedOrchestrationModeBySession.current.get(sessionId);
       queuedOrchestrationModeBySession.current.delete(sessionId);
--- a/apps/desktop/src/renderer/app-shell.tsx
+++ b/apps/desktop/src/renderer/app-shell.tsx
@@ -433,6 +433,12 @@ function AppShellContent({
   // MatrixA/fix-plan-click-flicker removed.
   const [, setPendingCollaborationModeBySession] = useState<Record<string, boolean>>({});
   const [, setPendingOrchestrationModeBySession] = useState<Record<string, boolean>>({});
+  // The rows stay interactive while a commit runs, so a click landing in that
+  // window is the user updating their mind — not noise to drop. Each map holds
+  // only the LATEST ask per session; the in-flight commit's finally block
+  // applies it if the settled state does not already satisfy it.
+  const queuedCollaborationModeBySession = useRef(new Map<string, boolean>());
+  const queuedOrchestrationModeBySession = useRef(new Map<string, OrchestrationMode>());
   const [newTaskPermissionChoice, setNewTaskPermissionChoice] =
     useNewTaskChoice<ChatDefaultPermissionMode>(currentNewTaskDraftKey);
   const [historyLoadPendingSessionId, setHistoryLoadPendingSessionId] = useState<string>();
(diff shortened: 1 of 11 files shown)


apps/desktop/src/renderer/use-composer-mentions.ts
```diff
@@ -34,6 +35,14 @@ export function useComposerMentions(options: {
...
   } = options;
   const [mentionSkills, setMentionSkills] = useState<InvocableSkillEntry[]>([]);
+  // The last SETTLED truth about the catalog, held across refreshes. The list
+  // itself is cleared while a refresh is in flight (fail-closed, for the `/`
+  // popup), so `length === 0` cannot tell "re-fetching" from "nothing to
+  // offer" — and the ＋ menu's Skills row repainting on that transient empty
+  // is a visible blink of the open menu on every mode, model or turn change.
+  // This flag only moves when a request resolves, so the row's presentation
+  // is stable until the catalog's emptiness actually changed.
+  const [mentionSkillsUnavailable, setMentionSkillsUnavailable] = useState(false);
```
Astro-Han: [P2] During the initial request, or immediately after switching session/project/model, `mentionSkills` is cleared fail-closed while this settled flag can still be `false`. The Skills row therefore looks enabled even though the current catalog is empty, and clicking it inserts a stray `/` before opening an empty popup. Could we model the catalog as one explicit `loading / ready-empty / ready-populated` state, keep the row geometry stable during loading, and ensure a loading click has no side effect? A regression covering an immediate click before the refreshed catalog settles would pin the boundary.

apps/desktop/src/renderer/app-shell.tsx
```diff
@@ -3291,26 +3296,16 @@ function AppShellContent({
...
-                  planModePending={activeId
-                    ? pendingCollaborationModeBySession[activeId] === true
-                    : false}
-                  planModeDisabledReason={
-                    activeId && pendingCollaborationModeBySession[activeId] === true
-                      ? shellCopy.modeChanging
-                      : modeChangeDisabledReason
-                  }
+                  // No pending-keyed disable while a toggle commits: the
+                  // pending registries already swallow re-entrant toggles, and
```
Astro-Han, line 3343: [P2] Now that pending is no longer reflected by the mode controls, the existing registry guard silently drops a second click while the row still appears enabled and retains its old checked state. A quick “on, then off” can therefore finish as “on” even though the user’s last action requested “off.” Could we either retain an accessible pending state without changing the menu geometry, or queue and apply the latest requested value after the current commit? A deferred-IPC test with two rapid toggles would cover this interaction.

Astro-Han (commented):
Thanks for addressing the visible menu jump and removing the unnecessary pending repaint. The direction is good, and both findings below are non-blocking P2s. They cover two short in-flight windows where the controls now look actionable but the underlying authority cannot honor the click.

A clean final state would give the Skill catalog one explicit loading/empty/populated representation, and either expose a real pending mode state without changing menu geometry or preserve the user’s latest toggle intent.

AI-assisted review disclosure: Codex coordinated independent state-authority and UI/test review passes against exact head `1beb6d2b399856bc39ac73e60efafc1d32dee2e1`; I reviewed the consolidated evidence for this comment.

apps/desktop/src/renderer/use-composer-mentions.ts
MatrixA: Addressed in acf2344e8. The catalog is now one explicit state — loading / settled-empty / settled-populated — held as a single value in the hook. Presentation still renders from the last settled verdict (geometry stays put during loading, including before the very first settle), and the new `mentionSkillsLoading` gates interaction: mid-refresh the row ignores clicks and stops closing the menu, so nothing types a stray `/` against the fail-closed list; the same click a beat later simply works. The held state is observable as `maka-composer-skills-loading`. Regressions: two SSR tests pin the loading presentation, and the new e2e `a Skills click during the catalog refresh does nothing, then works settled` holds the refresh open on a preload latch (isolated-E2E gate only), clicks straight into the window, and asserts an empty draft, no popup, and the menu still open — it fails without the fix.

apps/desktop/src/renderer/app-shell.tsx
MatrixA: Addressed in acf2344e8 with the queue option: each session keeps only the LATEST asked value while a commit is in flight, and the commit's finally block applies it when the settled state does not already satisfy it — so a quick "on, then off" finishes off, for Plan and orchestration alike. The settled state is read through `sessionsRef` rather than the render closure, which predates the commit. Regression: the new e2e `two rapid Plan toggles land on the last requested state` latches the session-list refresh (the tail of the commit) via the preload's isolated-E2E affordance, so the second click lands deterministically after the checked repaint but inside the pending window; it fails without the queue.

packages/ui/src/composer.tsx
```diff
@@ -1682,11 +1706,44 @@ export const Composer = forwardRef<
...
+                            : undefined
+                        }
+                        // Mid-refresh the row LOOKS like the previous catalog
+                        // but cannot act for the current one: opening the `/`
+                        // menu now would write a stray slash against a list
+                        // that is fail-closed empty. A loading click is a
+                        // complete no-op — nothing typed, menu left open — so
+                        // the same click a beat later simply works. The class
+                        // is the observable contract for that state.
+                        className={
```
Astro-Han, line 1744: [P2] While Skills are loading, this item still exposes an enabled `menuitem` to keyboard and screen-reader users even though activation is intentionally a no-op. The class has no ARIA meaning, so the control announces “available” but silently ignores the action. Could the loading state carry the component's disabled/busy semantics (while preserving the fixed geometry), and could a keyboard test pin that activation neither closes the menu nor writes `/`?

apps/desktop/src/renderer/use-composer-mentions.ts
```diff
@@ -33,18 +35,32 @@ export function useComposerMentions(options: {
...
     let cancelled = false;
     let requestVersion = 0;
-    // Do not briefly advertise the previous session/project's Skills while the
-    // authoritative projection is being refreshed. The same fail-closed reset
-    // applies when a model/mode/plan or MCP event changes the backend surface:
-    // a visible popup must never retain a now-unavailable Skill.
     const refresh = () => {
       const version = ++requestVersion;
-      setMentionSkills([]);
+      setCatalog((previous) => ({ loading: true, settled: previous.settled, skills: [] }));
```
Astro-Han: [P3] A context change renders once with the previous catalog before this passive effect runs and clears it. That leaves a narrow window where switching session/project/model and immediately opening Skills can expose IDs from the old context. A context key stored with the catalog would let render fail closed as soon as the key changes; a populated → context switch → loading test would make the lifecycle explicit.

apps/desktop/src/renderer/app-shell.tsx
```diff
@@ -1100,6 +1118,19 @@ function AppShellContent({
         collaborationModeChangeRegistry.keysRef,
         setPendingCollaborationModeBySession,
       );
+      // Whatever the user last asked for while this commit ran is the state
+      // they expect to land on. Read the settled mode through the ref — the
+      // closure's projection predates this commit — and only re-apply when
+      // the intent is not already satisfied.
+      const queued = queuedCollaborationModeBySession.current.get(sessionId);
```
Astro-Han, line 1128: [P3] `clearSessionRendererState()` clears the pending registries but not these queued maps. If a Session is removed or Host readiness changes while a commit is pending, this `finally` can replay an old intent against the already-cleared lifecycle. Could cleanup delete both queued entries as well (or gate replay by a Session generation) and cover pending → cleanup → settle?

apps/desktop/src/renderer/app-shell.tsx
```diff
@@ -427,8 +427,18 @@ function AppShellContent({
   const [newChatPlanModeActive, setNewChatPlanModeActive] = useState(false);
   const [newChatOrchestrationMode, setNewChatOrchestrationMode] = useState<OrchestrationMode>('default');
   const [scheduledTaskCreateRequestNonce, setScheduledTaskCreateRequestNonce] = useState(0);
-  const [pendingCollaborationModeBySession, setPendingCollaborationModeBySession] = useState<Record<string, boolean>>({});
-  const [pendingOrchestrationModeBySession, setPendingOrchestrationModeBySession] = useState<Record<string, boolean>>({});
+  // Write-only: the pending-action helpers need a store, but nothing renders
+  // from these — a mode toggle's guard is the registry ref, and painting the
+  // round trip (disable/dim, then restore) is exactly the ＋ menu blink that
+  // MatrixA/fix-plan-click-flicker removed.
+  const [, setPendingCollaborationModeBySession] = useState<Record<string, boolean>>({});
```
Astro-Han: [P3] These two React states are now deliberately write-only, so every add/clear still schedules a render whose value nobody reads. The registry refs already own re-entrancy; removing the setter parameter and these states would leave one authority and avoid the extra render this PR is trying to eliminate.

Astro-Han (commented):
Thanks for fixing the two earlier behavior regressions — the menu now keeps its geometry and the latest toggle intent wins on the main interaction path. Two independent passes rechecked this exact head. I kept one accessibility P2 and a few small lifecycle/simplification edges inline; the clean final state is one keyed lifecycle representation for both the catalog and queued mode intent, with loading exposed semantically rather than only visually.

AI-assisted review disclosure: OpenAI Codex coordinated two independent review passes. I verified the retained findings against this exact head and the current live PR state, and I made the final review decision.

packages/ui/src/composer.tsx
MatrixA: Addressed in 731beea1a. The row now carries `aria-busy` while the catalog refreshes — via a new vendor seam (`DropdownMenuItem` forwards the attribute to its row; documented in patches/README.md) — so assistive technology hears that activation is deferred rather than a plain "available" that swallows the action. Geometry and the enabled look stay fixed. The keyboard pin you asked for is in the e2e: mid-refresh, Enter on the focused row neither closes the menu nor writes `/`, and the same activation works once settled. One implementation note: the activation gate lives inside one always-present `onClick` — swapping the prop between undefined and a function remounts Item's internals and drops keyboard focus mid-refresh.

apps/desktop/src/renderer/use-composer-mentions.ts
MatrixA: Addressed in 731beea1a. The catalog value now stores the `contextKey` it describes (session, project path, model, collaboration/permission mode, new-task target), and the hook derives through it: a render whose key no longer matches fails closed to loading immediately, without waiting for the passive effect's clear. A same-context refresh still holds its settled verdict; a context switch has nothing settled to hold. The lifecycle test you asked for is in the e2e (`a context switch re-enters loading instead of holding the old catalog`): populated session catalog → switch to new chat under a latched new-task projection → the row presents busy and defers activation, then settles. Honest caveat: with the latch the spec pins the loading presentation and no-op, but the original one-frame window itself is not observable from Playwright — the key-based derivation is what closes it.

apps/desktop/src/renderer/app-shell.tsx
MatrixA: Addressed in 731beea1a. `clearSessionRendererState` now deletes both queued-intent entries alongside the pending registries, so a commit's finally has nothing to replay against a Session whose renderer lifecycle already ended. Covered by the new e2e `deleting the session while a toggle is pending settles clean`: commit latched open → second toggle queued → session deleted → latch released → the app settles on a clean new chat with no error surfaced and no stale Plan state.

apps/desktop/src/renderer/app-shell.tsx
MatrixA: Addressed in 731beea1a. The two write-only states are gone; the registry ref is the single re-entrancy authority. The shared helpers now take the setter optionally, since permission/model/retry pending states are still rendered from — mode toggles simply pass none, so an add/clear no longer schedules a render nobody reads.

Astro-Han (approved):
Thanks — the approach here is the part worth calling out: this fixes the jitter by *removing* state rather than adding a correction for it.

**What this solves / how:** toggling Plan made the composer's ＋ menu shift, and the Skills row could briefly show the previous context's skills or swallow the loading state entirely. Rather than compensating for the movement, this removes what caused the re-render — the write-only pending mode state — and makes the skill catalog fail closed on a context-key mismatch so a stale list can't paint for a frame.

**Review notes:**
- Keying the skill catalog on `contextKey` (session/project/model/mode/target) and returning an empty list on mismatch is the right default: showing nothing briefly is honest, showing another context's skills is not.
- Dropping `setPendingCollaborationModeBySession` and friends is the substantive change. Pending state that only ever gets written and never read still costs a render, and that render was the blink. Holding it in the registry ref instead means the value is still available where it's needed without driving the view.
- `aria-busy` with the gate inside the handler rather than a remount — keeps focus where the user left it.
- Net effect is two concepts gone (the write-only pending states) plus the write-implies-rerender coupling, with the responsibility moving to the registry ref. Nothing was added to compensate.

Worth noting for anyone reading this later: the tempting fix for "menu jumps" is a position offset, which would have left both the stale-render and the focus-loss problems in place while adding a third thing to maintain. This went after the cause instead.

LGTM.

---
*AI-assisted review. Verified against this exact head with CI green.*

PR #3470, opened 2026-08-22
fix(desktop): synchronize session row metadata with menu state

chihumyum:
## Summary

- Prevent the sidebar session timestamp from overlapping or flickering with its action trigger during menu dismissal.
- Mirror the controlled menu-open state onto the session action wrapper.
- Use the same hover, focus, and menu-open conditions for both action visibility and timestamp hiding, so they switch together.
- Extend Electron regression coverage for hover loss, outside dismissal, reopening, and pointer menu selection.

before:
Screenshot 2026-08-22 at 2 38 50 PM

after: 
Screenshot 2026-08-22 at 3 24 32 PM

## Verification

- `npm run build:with-deps` from `apps/desktop`
- `npx biome check apps/desktop/e2e/sidebar-project-row.spec.ts apps/desktop/src/renderer/styles/sidebar.css packages/ui/src/session-history-list.tsx`
- `npx playwright test --config e2e/playwright.config.ts e2e/sidebar-project-row.spec.ts` — 3 passed
- `git diff --check`

The focused regression failed before the controlled menu-state wiring was added.

## AI use

Select exactly one:

Tool(s) and scope: Codex analyzed the popover/focus timing, implemented the controlled-state CSS fix, and extended Electron E2E coverage.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/e2e/sidebar-project-row.spec.ts
+++ b/apps/desktop/e2e/sidebar-project-row.spec.ts
@@ -100,11 +100,28 @@ test('task row action menu accepts pointer selection', async ({
   await expect(page.locator('[data-maka-contract="search-modal"]')).not.toBeVisible();
 
   const sidebar = page.getByRole('navigation', { name: '任务列表' });
-  const taskRow = sessionRow(sidebar, `${LONG_SIDEBAR_SESSION_PREFIX}00`);
+  const taskSessionId = `${LONG_SIDEBAR_SESSION_PREFIX}00`;
+  const taskRow = sessionRow(sidebar, taskSessionId);
+  const actionMenu = taskRow.locator(':scope > .maka-session-row-action');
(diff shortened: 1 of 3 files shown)


Astro-Han (approved):
Thanks — this is a focused fix at the existing controlled-menu seam. I reviewed exact head `dff8838ce4dc63f04dbf5d6249443c103a664827`; the DOM projection follows the same `menuOpen` state that owns dismissal and focus handoff, the selector does not affect project rows, and the pointer/keyboard regression coverage matches the behavior. I found no P0–P3 issues, and the current-head `test` check is green.

AI-assisted review disclosure: OpenAI Codex performed an independent exact-head review. I verified the state ownership, event/lifecycle paths, tests, live review state, mergeability, and current-head CI, and I independently made the approval decision.

PR #3471, opened 2026-08-22
feat(core): search the External Session catalog by title and path

### LINKED ISSUE #3438: External Session import: 16-per-page paging is the only way to find a session among 1128
## Problem

The import catalog pages 16 sessions at a time (`EXTERNAL_SESSION_PAGE_MAX_ITEMS`) and offers one filter, `includeArchived`. That was proportionate to what the only source produced. It stops being proportionate the moment a second one lands.

Measured on one machine:

| source | sessions |
| --- | ---: |
| Codex | 19 |
| Claude Code (#3435) | **1128** |

**59×.** At 16 per page that is 71 pages of cursor paging to reach a session by scrolling, with no way to say which one you want. A user who knows exactly what they are looking for — by title, or by which project it belongs to — has no way to express it.

This is a consequence of #3435 rather than a pre-existing gap: a Codex-only catalog never had enough rows for paging alone to fail. Filing it separately because the fix belongs to the query surface both sources share, not to either adapter.

## Where it belongs

`ExternalSessionQuery` (`packages/core/src/external-session.ts`) currently carries:

```ts
export interface ExternalSessionQuery {
  cwd?: string;
  includeArchived?: boolean;
}
```

`cwd` already exists and already works — it is an exact match, which answers "this project" but not "the project I half remember". A text query would sit beside it and be honoured by each adapter over data it has already parsed.

## Scope

- Add a text query to `ExternalSessionQuery`, matched against the session title and its cwd. Both are already in `ExternalSessionSummary`, so neither adapter needs to read more than it does now.
- Honour it in both adapters — Codex and Claude Code — so the surface does not gain a filter that silently works for one source and not the other.
- A search field in the import settings page, next to the existing archived toggle.
- Decide where the match runs. Filtering after the page is assembled would search only the 16 rows already fetched, which is worse than no search at all; the query has to reach the adapter, before paging.

## Not in scope

Searching message content. Titles and paths are metadata the adapters already hold; content means reading every transcript on every keystroke, and that is a different problem with a different budget.

Refs #3435

Joob1n:
## Summary

设置 → 导入任务 pages 16 sessions at a time with one filter, `includeArchived`. That was proportionate to what Codex produces here — 19 sessions. Claude Code brought **1128**, which is 71 pages of cursor paging with no way to say which one you want.

`ExternalSessionQuery` gains `text`, matched against the summary's title and cwd. Both already sit on `ExternalSessionSummary`, so neither adapter reads more than it did.

**The term reaches the adapter, before paging.** The coordinator slices a page out of whatever `listSessions` returns, so a filter applied after that would search the 16 rows already fetched. That is worse than offering no search at all, because it looks like it worked.

**One matcher, not one per adapter.** `externalSessionMatchesQuery` lives in `@maka/core` and both adapters call it. The catalog is one surface over several sources; a filter that quietly worked for Codex and not for Claude Code would be undebuggable from the UI, which cannot say which source dropped the term. A third adapter inherits the behaviour rather than reimplementing it.

Fixes #3438

## Review focus

**This consolidation fixes an inconsistency I introduced in #3435.** Codex compared cwd through `normalizePath`; the Claude Code adapter compared raw strings. The same project reached with a trailing separator answered "no such project" on one source and not the other. Both now share one path rule, and there is a test for it.

**A blank box is not a filter.** `''` and `'   '` both normalize to "no term", so a stray space cannot hide every session. An over-long term is truncated to its 200-character prefix rather than throwing at the user; the protocol bounds it at the frame because it reaches every adapter.

**Search is a dimension of the catalog selection**, not a view filter. It joins `catalogSelectionRef` and the `ImportAttempt` record beside `includeArchived`: import recovery re-reads the window a row came from, and a cleared box would look at a different list.

**Debounced at 250ms.** The term walks every transcript on the source, so a request per keystroke would queue a thousand-file scan behind each character.

**A search that finds nothing gets its own empty state.** Reusing "no conversations" would tell a user their transcripts are missing when the term is simply wrong, and hide the one control that would fix it.

## Breaking change

Protocol epoch **36 → 37**. `ExternalSessionCatalogQueryInput` gains an optional `text`. `scripts/protocol-epoch-check.mjs --base origin/main` confirms: `Protocol changed and the epoch moved: 36 -> 37.`

## Verification

**Tests — 26 new, across every layer the term crosses:**

| layer | what is pinned |
| --- | --- |
| `@maka/core` (12) | title/path match, blank-is-not-a-filter, truncation, archived gate, cwd+text both apply, path normalization |
| Claude Code adapter (2) | filters the source not the page; trailing separator tolerated |
| Codex adapter | the same term, same results — the "silently works for one source" case |
| protocol (4) | a term is carried; a non-string, empty, or over-long one is refused at the frame |
| storybook `play` | typing `worktree` narrows three rows to one |

The storybook stub honours `text` on purpose. A stub that ignored it would render a search box that looks wired and is not, and the story would certify that. **Disconnecting the term from the request fails that story** — checked by removing it.

**Checks run locally:**

- `@maka/core`: **593/593** · `@maka/storage`: **895/895** · `@maka/runtime-host`: **1066/1066** · `@maka/desktop`: **1092/1092**
- `npm run format:check`: 1571 files, clean
- `biome lint .`: 2510 files, clean
- `build-storybook` + `smoke:storybook`: **162 stories**, AX audit included

**Not run:** the packaged-artifact and release checks; `@maka/runtime`'s suite has 5 unrelated environmental failures here (`spawn rg ENOENT` — no ripgrep binary).

## Not in scope

Searching message content. Titles and paths are metadata the adapters already hold; content means reading every transcript on every keystroke, which is a different problem with a different budget.

## AI use

Select exactly one:

Tool(s) and scope: Claude Opus 5 (Claude Code) — wrote the implementation, tests, and story; ran the checks above. Reviewed and directed by me. `Generated-by` trailer is on the commit.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/src/preload/preload.ts
+++ b/apps/desktop/src/preload/preload.ts
@@ -1807,6 +1807,7 @@ const makaBridge = {
       adapterId: string;
       includeArchived?: boolean;
       cursor?: string;
+      text?: string;
     }, host?: DesktopRuntimeHostRef): Promise<{
       sessions: DesktopExternalSessionCatalogItem[];
       nextCursor: string | null;
--- a/apps/desktop/src/renderer/settings/import-tasks-settings-page.tsx
+++ b/apps/desktop/src/renderer/settings/import-tasks-settings-page.tsx
@@ -5,6 +5,7 @@ import { CheckboxInput } from '@astryxdesign/core/CheckboxInput';
 import { EmptyState } from '@astryxdesign/core/EmptyState';
 import { List, ListItem } from '@astryxdesign/core/List';
 import { SegmentedControl, SegmentedControlItem } from '@astryxdesign/core/SegmentedControl';
+import { TextInput } from '@astryxdesign/core/TextInput';
 import { HStack, VStack } from '@astryxdesign/core/Stack';
 import { uiLocaleToIntlLocale } from '@maka/core/ui-locale';
 import type {
@@ -34,6 +35,7 @@ type CatalogWindow = CatalogState & {
 async function readCatalogWindow(input: {
   adapterId: string;
   includeArchived: boolean;
+  text: string;
   minimumItemCount: number;
   targetSourceSessionId?: string;
   host?: DesktopRuntimeHostRef;
@@ -49,6 +51,7 @@ async function readCatalogWindow(input: {
       {
         adapterId: input.adapterId,
         includeArchived: input.includeArchived,
+        ...(input.text ? { text: input.text } : {}),
         ...(cursor === undefined ? {} : { cursor }),
       },
       input.host,
@@ -87,6 +90,7 @@ type ImportAttempt = {
   sourceSessionId: string;
   name: string;
   includeArchived: boolean;
+  text: string;
   importedCountBefore: number;
   latestImportedSessionIdBefore: string | undefined;
   loadedCatalogItemCountBefore: number;
@@ -140,6 +144,11 @@ export function ImportTasksSettingsPage(props: {
   const [adapterIds, setAdapterIds] = useState<string[]>([]);
   const [adapterId, setAdapterId] = useState<string | null>(null);
   const [includeArchived, setIncludeArchived] = useState(false);
+  // Two states, not one: `searchDraft` is what the box shows, `search` is what
+  // has been asked of the Host. Typing must not put a request on the wire per
+  // keystroke against a source with a thousand sessions.
+  const [searchDraft, setSearchDraft] = useState('');
+  const [search, setSearch] = useState('');
   const [catalog, setCatalog] = useState<CatalogState>(EMPTY_CATALOG);
   const [sourceLoading, setSourceLoading] = useState(false);
   const [sourceResolved, setSourceResolved] = useState(false);
@@ -170,17 +179,28 @@ export function ImportTasksSettingsPage(props: {
   // source's rows under the new source's label.
   const requestGeneration = useRef(0);
   const recoveryGeneration = useRef(0);
-  const catalogSelectionRef = useRef({ adapterId, includeArchived, generation: 0 });
+  const catalogSelectionRef = useRef({ adapterId, includeArchived, search, generation: 0 });
   if (
     catalogSelectionRef.current.adapterId !== adapterId ||
-    catalogSelectionRef.current.includeArchived !== includeArchived
+    catalogSelectionRef.current.includeArchived !== includeArchived ||
+    catalogSelectionRef.current.search !== search
   ) {
     catalogSelectionRef.current = {
       adapterId,
       includeArchived,
+      search,
       generation: catalogSelectionRef.current.generation + 1,
     };
   }
+
+  // 250ms after typing stops, not per keystroke. The term reaches the adapter,
+  // which walks every transcript on the source, so a request per character
+  // would queue a thousand-file scan behind each one.
+  useEffect(() => {
+    if (searchDraft === search) return;
+    const timer = setTimeout(() => setSearch(searchDraft), 250);
+    return () => clearTimeout(timer);
+  }, [searchDraft, search]);
 
   const loadSources = useCallback(async () => {
     const generation = ++requestGeneration.current;
@@ -224,6 +244,7 @@ export function ImportTasksSettingsPage(props: {
         const result = await window.maka.externalSessions.list({
           adapterId: sourceId,
           includeArchived,
+          ...(search ? { text: search } : {}),
           ...(cursor === undefined ? {} : { cursor }),
         }, host);
         if (generation !== requestGeneration.current) return;
@@ -241,7 +262,7 @@ export function ImportTasksSettingsPage(props: {
         }
       }
     },
-    [copy.loadFailedFallback, host, includeArchived, locale],
+    [copy.loadFailedFallback, host, includeArchived, locale, search],
   );
 
   const refreshLoadedCatalog = useCallback(
@@ -251,6 +272,7 @@ export function ImportTasksSettingsPage(props: {
         const result = await readCatalogWindow({
           adapterId: sourceId,
           includeArchived,
+          text: search,
           minimumItemCount: loadedItemCount,
           host,
           isCurrent: () => mountedRef.current && generation === requestGeneration.current,
@@ -261,7 +283,7 @@ export function ImportTasksSettingsPage(props: {
         // visible and retry rather than replacing it with a transient error.
       }
     },
-    [host, includeArchived, mountedRef],
+    [host, includeArchived, mountedRef, search],
   );
 
   useEffect(() => {
@@ -271,7 +293,7 @@ export function ImportTasksSettingsPage(props: {
   useEffect(() => {
     if (adapterId === null) return;
     void loadCatalog(adapterId);
-  }, [adapterId, includeArchived, loadCatalog]);
+  }, [adapterId, includeArchived, search, loadCatalog]);
 
   const hasCatalogImportInFlight = catalog.sessions.some(
     (session) => session.importState.isImporting,
@@ -321,6 +343,7 @@ export function ImportTasksSettingsPage(props: {
         const result = await readCatalogWindow({
           adapterId: attempt.adapterId,
           includeArchived: attempt.includeArchived,
+          text: attempt.text,
           minimumItemCount: attempt.loadedCatalogItemCountBefore,
           targetSourceSessionId: attempt.sourceSessionId,
           host,
@@ -391,6 +414,10 @@ export function ImportTasksSettingsPage(props: {
         sourceSessionId: session.id,
         name: session.name,
         includeArchived,
+        // The search term is part of the attempt for the same reason the
+        // archived filter is: recovery re-reads the catalog window this row
+        // came from, and a cleared box would look at a different list.
+        text: search,
         importedCountBefore: session.importState.importedCount,
         latestImportedSessionIdBefore: session.importState.importedSessionIds[0],
         loadedCatalogItemCountBefore: catalog.sessions.length,
@@ -432,6 +459,7 @@ export function ImportTasksSettingsPage(props: {
       props,
       recoverUnknownImport,
       includeArchived,
+      search,
     ],
   );
 
@@ -507,6 +535,14 @@ export function ImportTasksSettingsPage(props: {
               ))}
             </SegmentedControl>
           )}
+          <TextInput
+            label={copy.searchLabel}
+            description={copy.searchHelp}
+            placeholder={copy.searchPlaceholder}
+            value={searchDraft}
+            onChange={setSearchDraft}
+            hasClear
+          />
           <CheckboxInput
             label={copy.includeArchived}
             value={includeArchived}
@@ -609,7 +645,14 @@ export function ImportTasksSettingsPage(props: {
             </div>
           )}
 
-          {catalogEmpty && (
+          {/* A search that found nothing is not the same as a source with
+              nothing in it. Reusing the "no conversations" copy would tell a
+              user their transcripts are missing when the term is simply
+              wrong, and hide the one control that would fix it. */}
+          {catalogEmpty && search !== '' && (
+            <EmptyState isCompact title={copy.searchLabel} description={copy.searchEmpty(search)} />
+          )}
+          {catalogEmpty && search === '' && (
             <EmptyState isCompact title={copy.emptyTitle} description={copy.emptyDescription} />
           )}
 
--- a/apps/desktop/src/renderer/settings/import-tasks-settings-page.tsx
+++ b/apps/desktop/src/renderer/settings/import-tasks-settings-page.tsx
@@ -7,6 +7,7 @@ import { List, ListItem } from '@astryxdesign/core/List';
 import { SegmentedControl, SegmentedControlItem } from '@astryxdesign/core/SegmentedControl';
 import { TextInput } from '@astryxdesign/core/TextInput';
 import { HStack, VStack } from '@astryxdesign/core/Stack';
+import { normalizeExternalSessionQueryText } from '@maka/core/external-session';
 import { uiLocaleToIntlLocale } from '@maka/core/ui-locale';
 import type {
   DesktopRuntimeHostRef,
@@ -466,6 +467,9 @@ export function ImportTasksSettingsPage(props: {
   const noSource = sourceResolved && !sourceLoading && !sourceError && adapterIds.length === 0;
   const catalogEmpty =
     adapterId !== null && !catalogLoading && !catalogError && catalog.sessions.length === 0;
+  // The shared normalizer decides what counts as a filter, so the empty-state
+  // copy and the matcher cannot disagree about a whitespace-only box.
+  const activeSearch = normalizeExternalSessionQueryText(search);
 
   if (sourceLoading) {
     return (
@@ -649,10 +653,18 @@ export function ImportTasksSettingsPage(props: {
               nothing in it. Reusing the "no conversations" copy would tell a
               user their transcripts are missing when the term is simply
               wrong, and hide the one control that would fix it. */}
-          {catalogEmpty && search !== '' && (
-            <EmptyState isCompact title={copy.searchLabel} description={copy.searchEmpty(search)} />
+          {/* Keyed on the normalized term, the same value the matcher uses.
+              Comparing the raw string would call a box holding only spaces an
+              active search and answer `No conversation has "   "` on a source
+              that is simply empty. */}
+          {catalogEmpty && activeSearch !== undefined && (
+            <EmptyState
+              isCompact
+              title={copy.searchLabel}
+              description={copy.searchEmpty(search.trim())}
+            />
           )}
-          {catalogEmpty && search === '' && (
+          {catalogEmpty && activeSearch === undefined && (
             <EmptyState isCompact title={copy.emptyTitle} description={copy.emptyDescription} />
           )}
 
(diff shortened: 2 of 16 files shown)


packages/core/src/external-session.ts
```diff
@@ -18,6 +29,58 @@ export interface ExternalSessionSummary {
...
+): boolean {
+  if (!query.includeArchived && summary.archived) return false;
+  if (query.cwd !== undefined && !sameExternalSessionPath(summary.cwd, query.cwd)) return false;
+  const text = normalizeExternalSessionQueryText(query.text);
+  if (text === undefined) return true;
+  // Title and path, because those are the two things a user remembers about a
+  // conversation they are looking for. Both already sit on the summary, so
+  // matching costs no extra reads. Message content is deliberately excluded:
+  // it would mean opening every transcript on every keystroke.
+  return summary.name.toLowerCase().includes(text) || summary.cwd.toLowerCase().includes(text);
```
Astro-Han: [P2] Could the shared matcher normalize both the query and candidate strings before matching? On Windows, a stored `C:/Repo/App` does not match pasted `C:\\Repo\\App`, even though `sameExternalSessionPath()` already treats those paths as equivalent. On macOS, the same visible name/path can also arrive in NFC or NFD form and miss here. Reusing one text normalizer plus the existing separator rule for the cwd field would keep the shared matcher as the authority; adapter-level cases for Windows separators and composed/decomposed Unicode would pin it.

packages/storage/src/codex-session-adapter.ts
```diff
@@ -750,8 +751,9 @@ function cwdSqlVariants(cwd: string): string[] {
 }
 
 function matchesQuery(entry: ExternalSessionSummary, query: ExternalSessionQuery): boolean {
-  if (!query.includeArchived && entry.archived) return false;
-  return query.cwd === undefined || normalizePath(entry.cwd) === normalizePath(query.cwd);
+  // Delegated so both sources answer a query identically; the local
+  // `normalizePath` below is still used for the SQL cwd variants.
+  return externalSessionMatchesQuery(entry, query);
```
Astro-Han, line 747: [P2] This shared matcher is reached only after the existing SQLite `cwd IN (...)` prefilter has already narrowed the rows. That prefilter cannot express the same Windows path equivalence: for example, a DB row stored as `C:\\Repo\\App` and a query `c:/repo/app` can be discarded before this function sees it. Could we remove the lossy SQL prefilter and let this function be the single authority, or store/query one canonical path representation? Please pin the real SQLite adapter path rather than only the helper.

apps/desktop/src/renderer/settings/import-tasks-settings-page.tsx
```diff
@@ -609,7 +645,14 @@ export function ImportTasksSettingsPage(props: {
             </div>
           )}
 
-          {catalogEmpty && (
+          {/* A search that found nothing is not the same as a source with
+              nothing in it. Reusing the "no conversations" copy would tell a
+              user their transcripts are missing when the term is simply
+              wrong, and hide the one control that would fix it. */}
+          {catalogEmpty && search !== '' && (
```
Astro-Han: [P3] Whitespace-only input is normalized to “no filter” by the shared helper, but this branch still treats it as an active search. On an empty source, entering spaces therefore shows `No conversation has "   "` instead of the ordinary empty-source state. Could this condition use the same normalized query value so matching and empty-state copy cannot diverge?

Astro-Han (commented):
Thanks for tackling a real usability problem here — filtering before pagination and keeping the shared matcher in Core are both the right direction. I reviewed this exact head through two independent passes and kept only the concrete cross-platform and UI consistency gaps inline. The clean final state is one normalized text/path authority used by both adapters and the renderer, without an earlier storage filter narrowing the candidate set differently.

AI-assisted review disclosure: OpenAI Codex coordinated two independent review passes. I verified the retained findings against this exact head and the live PR state, and I made the final review decision.

Astro-Han (commented):
Thanks — the current-head CI has now run, and it exposed one deterministic generated-contract failure. I verified the generator diff before adding the inline note; this is a small mechanical fix rather than a broader implementation change.

AI-assisted review disclosure: OpenAI Codex delegated the failed-CI diagnosis. I verified the run SHA, failed step, generator output, and minimal fix against this exact head.

apps/desktop/src/renderer/settings/import-tasks-settings-page.tsx
```diff
@@ -5,6 +5,7 @@ import { CheckboxInput } from '@astryxdesign/core/CheckboxInput';
 import { EmptyState } from '@astryxdesign/core/EmptyState';
 import { List, ListItem } from '@astryxdesign/core/List';
 import { SegmentedControl, SegmentedControlItem } from '@astryxdesign/core/SegmentedControl';
+import { TextInput } from '@astryxdesign/core/TextInput';
```
Astro-Han, line 8: [P1] Adding this Astryx component changes the committed surface inventory, but the generated Markdown was not updated. The current exact head deterministically fails the required `Astryx surface inventory` CI step because `docs/astryx-surface-file-inventory.md` still omits `TextInput` for this file. Please run `npm run astryx:surface-inventory:write` and commit the generated Markdown update; the `.paths` inventory is already unchanged.

Joob1n:
All four addressed. `66ed545a4` for the P1, `9778cb715` for the rest.

**P1 — inventory.** Regenerated and committed. The root cause was mine: I ran lint, format, typecheck and the affected suites before pushing, but not the generated-artifact checks. I have since run all seven — surface inventory, Windows inventory, third-party notices, CLI notices, ASF source, release identity, Windows cargo notices — on **both** of my open branches, and that was the only stale one.

**P2 — normalize both sides.** You were right, and the two cases you named turned out to be the same root cause: the matcher compared strings the cwd rule already considered equal. Reproduced both before fixing:

```
stored C:/Repo/App   searched C:\Repo\App   ->  false
stored NFC           searched NFD           ->  false
```

Both sides now pass through one normalizer — NFC, lowercase, separators — and `sameExternalSessionPath` folds in the same NFC step, so the two rules cannot disagree about the same pair. Adapter-level cases pin both, as you asked.

**P2 — the SQL prefilter.** Also right, and the sharper framing is yours: it was a second, weaker rule sitting in front of the authority. `cwd IN (<variants>)` enumerated spellings of the query, but SQLite compares them exactly, so a row stored `C:\Repo\App` never reached `matchesQuery` to be told that `c:/repo/app` is the same project.

I removed it rather than teaching it the equivalence. A prefilter that cannot express the matcher's semantics is not an optimization. The statement has no `LIMIT`, so dropping the clause widens the read instead of truncating it, and the archived clause stays — an exact boolean that agrees with the matcher by construction. `normalizePath` and `cwdSqlVariants` were only reachable from that clause and went with it.

Per your note, **the regression drives the real state database**, not the helper: it seeds a row with `cwd: 'C:\Repo\App'` and asserts all four spellings find it. Restoring the prefilter fails it — checked by restoring it.

**P3 — whitespace empty state.** Fixed by keying the branch on the normalized value the matcher uses, so the copy and the match cannot diverge.

## Verification

- `@maka/core` **595/595** · `@maka/storage` **896/896** · `@maka/runtime-host` **1066/1066** · `@maka/desktop` **1092/1092**
- `format:check` 1571 files · `biome lint .` 2510 files: clean
- all seven generated-artifact checks: clean
- `build-storybook` + `smoke:storybook`: 162 stories, AX audit included

apps/desktop/src/renderer/settings/import-tasks-settings-page.tsx
```diff
@@ -170,18 +180,29 @@ export function ImportTasksSettingsPage(props: {
...
   ) {
     catalogSelectionRef.current = {
       adapterId,
       includeArchived,
+      search,
       generation: catalogSelectionRef.current.generation + 1,
     };
   }
 
+  // 250ms after typing stops, not per keystroke. The term reaches the adapter,
```
Astro-Han, line 197: [P2, ordinary user path] Each typing pause over 250ms can start a full catalog scan, and the previous one is never cancelled.

`listSessions` walks every transcript file, `readFile`s it whole, and `split('\n')`-parses all of it — no cache, no header-only path. That cost already existed; what changes here is how often it runs. It used to happen when the page opened. Now it happens per search.

The debounce is a standard trailing-edge one, so typing fast collapses into a single request. But pausing longer than 250ms mid-word is ordinary typing, and each pause can start another full scan while `requestGeneration` only discards the *result* — the earlier scan still runs to completion. At the ~1128-session scale this PR itself cites, those overlap and stack.

Not P1: nothing is lost or corrupted and closing the page recovers. But the user has no way to make it stop.

Suggestion: thread an `AbortSignal` through the catalog read and abort superseded scans. Discarding results isn't cancellation.

packages/core/src/external-session.ts
```diff
@@ -18,6 +29,103 @@ export interface ExternalSessionSummary {
...
+ * Codex normalizes separators and lowercases a Windows drive prefix before
+ * comparing; the Claude Code adapter compared raw strings, so the same project
+ * reached through a different separator answered "no such project". One rule
+ * for both.
+ */
+export function sameExternalSessionPath(left: string, right: string): boolean {
+  return normalizeExternalSessionPath(left) === normalizeExternalSessionPath(right);
+}
+
+function normalizeExternalSessionPath(value: string): string {
```
Astro-Han, line 124: [P3] `normalizeExternalSessionPath` folds the POSIX root and an unknown cwd into the same value.

The unconditional `.replace(/\/+$/u, '')` makes `normalizeExternalSessionPath('/')` and `normalizeExternalSessionPath('')` both `''`, so `sameExternalSessionPath('/', '')` is `true`. Verified directly:

```
'/'  -> ''
''   -> ''
same('/', '') === true
```

Both ends of the path are reachable: the workspace resolver accepts `/` (it's absolute and resolves to a directory), and the Claude adapter leaves `cwd: ''` when no record carries one. So viewing the import catalog with the workspace at filesystem root pulls in sessions whose cwd is simply unknown.

Only P3: everything involved is the same user's own sessions on the same machine, so this is filter accuracy rather than a boundary being crossed. Fix: keep `/` as `/`, and a `sameExternalSessionPath('/', '') === false` test to pin it.

Astro-Han (commented):
Thanks — routing the filter into the adapters so it runs *before* pagination is the right call, and collapsing Codex's SQL `cwd IN (...)` prefilter into the shared matcher rather than patching it was the right instinct too: it was a second, weaker rule for the same fact.

**What this solves / how:** the External Session catalog could only be filtered by cwd/archived, so finding one session among ~1128 meant paging until it appeared. This adds a shared `externalSessionMatchesQuery` in core that both adapters apply before assembling a page.

**One root cause behind all three findings below.** The PR now has *two* path-equivalence rules where it means to have one: `normalizeExternalSessionPath` (used by `sameExternalSessionPath`) strips trailing separators, and `foldExternalSessionPathSeparators` (used by the new matcher) does not. One strips too much and swallows the POSIX root; the other strips too little and misses pasted paths. Worth fixing as a single rule rather than three separate patches — otherwise you're likely to still have two rules afterwards.

Details inline.

---
*AI-assisted review. Two reviewers worked this PR independently and reached different conclusions; the facts behind each finding were then verified directly against this head, and one finding was downgraded during that check.*

packages/core/src/external-session.ts
```diff
@@ -18,6 +29,103 @@ export interface ExternalSessionSummary {
...
+function normalizeExternalSessionMatchText(value: string): string {
+  return value.normalize('NFC').toLowerCase();
+}
+
+/**
+ * Windows separators folded to the POSIX form, so a term pasted from one
+ * spelling of a path finds the project the summary stored in the other. The
+ * same equivalence `sameExternalSessionPath` applies to the `cwd` filter.
+ */
+function foldExternalSessionPathSeparators(value: string): string {
```
Astro-Han, line 97: [P3] A search term with a trailing separator doesn't match the same project.

`foldExternalSessionPathSeparators` folds separators but doesn't strip a trailing one. Paste `C:\Repo\App\` from Windows Explorer (which routinely includes the trailing backslash) against a stored cwd of `C:/Repo/App`: the term folds to `c:/repo/app/`, and `includes('c:/repo/app/')` is false.

The PR's own Windows-paste test case has no trailing separator, so this sits right next to covered ground.

This is the other half of the root cause above — here the rule strips too little, where `normalizeExternalSessionPath` strips too much.

Joob1n:
All three addressed in `4df38eb42`. Your root-cause read was the useful part — I had built two path rules and would have patched them into two slightly different path rules.

**The two P3s were one bug in opposite directions.** Reproduced both first:

```
same('/', '')                          -> true    (should be false)
cwd 'C:/Repo/App'  term 'C:\Repo\App\' -> false   (should be true)
```

`normalizeExternalSessionPath` stripped trailing separators unconditionally, so `/` and `''` folded together. `foldExternalSessionPathSeparators` folded separators but stripped nothing, so a Windows Explorer paste missed. The second now delegates to the first, and the first stops stripping at one separator — the POSIX root *is* its separator.

Boundaries pinned, including the ones neither rule handled before:

```
'/'  vs ''    false      '//'  vs '/'   true      '///' vs '/'  true
'/'  vs '/'   true       ''    vs ''    true      '/'   vs '/repo'  false
'C:/' vs 'C:\' true      'C:/Repo/' vs 'c:\repo'  true
```

## On the P2 — I did not thread an `AbortSignal`

Your measurement of the problem is right, and I reproduced it: three consecutive listings over 1128 transcripts took **948 ms · 1128 ms · 958 ms**. Nothing is reused, and `requestGeneration` discards the result rather than the work.

But there is no request cancellation anywhere in this stack — no signal in the Host protocol, no cancel/abort operation, nothing in the preload bridge. Threading one means inventing cancellation for the protocol, with an epoch bump and a cancel operation of its own. That is a feature, and it should not arrive inside a search PR where it would be reviewed as a footnote.

**So I removed the cost that cancellation would have been avoiding.** Summaries are now cached per transcript and invalidated by the file's own mtime and size. Same three calls:

| | 1st | 2nd | 3rd |
| --- | ---: | ---: | ---: |
| before | 948 ms | 1128 ms | 958 ms |
| after | 948 ms | **28 ms** | **26 ms** |

A superseded scan still runs to completion — that part of your finding stands — but it re-reads nothing that has not changed, so pauses no longer stack a second of work each. The first listing is unchanged; this is about repetition, which is exactly what the search box introduced.

Two things a cache has to get right, both pinned:

- **staleness** — a rewritten transcript is re-read, a deleted one drops out. Keying on path alone fails that test; I checked by keying on path alone.
- **growth** — entries for transcripts the source stops listing are dropped on each listing, so a long-lived Host does not accumulate one per deleted session.

If you would rather have real cancellation regardless, I would take that as a separate issue against the protocol — happy to file it, and it would help more surfaces than this one.

## Verification

- `@maka/core` **598/598** · `@maka/storage` **897/897** · `@maka/runtime-host` **1066/1066**
- `format:check` 1571 files · `biome lint .` 2510 files: clean

likun666661 (approved):

apps/desktop/src/preload/preload.ts
```diff
@@ -1807,6 +1807,7 @@ const makaBridge = {
       adapterId: string;
       includeArchived?: boolean;
       cursor?: string;
+      text?: string;
```
jackwener, line 1810: [P2] The implementation accepts `text` here, but the public renderer contract still does not: `apps/desktop/src/preload/bridge-contract.d.ts` declares `MakaBridge.externalSessions.list` with only `adapterId`, `includeArchived`, and `cursor`. The current page happens to put `text` behind an object spread, so TypeScript does not report the excess property, but a direct typed consumer of the bridge cannot pass the new field. Please add `text?: string` to the declared list input so the exposed contract matches the implementation.

packages/storage/src/claude-code-session-adapter.ts
```diff
@@ -76,26 +94,64 @@ export class ClaudeCodeSessionAdapter implements ExternalSessionAdapter {
...
+      parsed && !parsed.isSidechain
+        ? {
+            id: sessionId,
+            name: parsed.title || sessionId,
+            cwd: parsed.cwd,
+            ...(parsed.createdAt !== undefined ? { createdAt: parsed.createdAt } : {}),
+            ...(parsed.updatedAt !== undefined ? { updatedAt: parsed.updatedAt } : {}),
+          }
+        : undefined;
+    this.#summaries.set(path, { mtimeMs, size, ...(summary ? { summary } : {}) });
```
jackwener, line 151: [P3] This negative cache also records transient I/O failures as if they were stable exclusions. `#parse` returns `undefined` for a failed inner `stat` or `readFile`, and this line then caches that value under the earlier mtime/size. If the file becomes readable again without its contents changing (for example, permissions are restored after a temporary read failure, or a transient `EMFILE`/`EIO` clears), later listings keep returning the cached `undefined` until the Host restarts or the transcript metadata changes. Sidechain/oversize results are stable, but read failures are not; please keep transient failures out of the negative cache or return a discriminated parse outcome so only stable exclusions are cached.

apps/desktop/src/renderer/settings/import-tasks-settings-page.tsx
jackwener: Rechecked on current head `4df38eb42`: the cache makes completed repeat listings cheap, but it does not cover the overlapping cold-request case in this thread. The `948ms / 28ms / 26ms` measurement is sequential; during the first ~948ms scan, a second 250ms-debounced catalog request can still be dispatched concurrently. `connection-session.ts` only serializes `session.transcript.page`, and `#summaries` stores settled summaries rather than sharing an in-flight parse, so once the second walk catches the first it can still duplicate the remaining reads/parses. A narrow fix can coalesce the in-flight summary/listing work without adding protocol cancellation; cancellation can remain separate scope.

PR #3472, opened 2026-08-22
fix(desktop): resync renderer after Windows maximize

### LINKED ISSUE #3416: bug(desktop): maximizing from a restored window on Windows leaves the renderer at the old size — unpainted area shows as a white edge
### What happened

On Windows, when the Maka desktop window starts **small (restored)** and the user then **maximizes** it (double-click on the titlebar drag region), the renderer viewport does not grow with the window. The app content keeps the pre-maximize size, and the rest of the window is never painted — it shows the `BrowserWindow` `backgroundColor` (`#ffffff` in light mode) as a large white band along the right and bottom edges of the screen.

Observed: after double-click-maximizing on a 1920×1080 display, the sidebar + chat column occupy only roughly the old restored window size (~1240×820) anchored top-left; everything to the right of and below that region is blank white. The chat scrollbar also sits at the old viewport edge.

Expected: the renderer reflows and paints the full maximized window immediately.

### How to reproduce

1. On Windows 11, launch Maka Desktop so the window opens restored/non-maximized (default saved bounds are 1240×820).
2. Double-click the titlebar drag region to maximize (same result via the native maximize button).
3. The window frame maximizes, but the web contents stay at the pre-maximize size; the unpainted remainder shows as a white edge/band (right + bottom).

### Environment

- Maka version or commit: v0.1.11 (packaged)
- OS and version: Windows 11
- Surface: Desktop

Screenshot (light mode): maximized window where app content stops at the old viewport size; white unpainted area fills the rest of the screen.

Notes from source inspection (not yet root-caused upstream):

- Main window uses `titleBarStyle: 'hidden'` + `titleBarOverlay` on Windows (`apps/desktop/src/main/main-window.ts`, Electron 43.2.0).
- The renderer layout is fully fluid (`height: 100dvh; width: 100%` in `shell-layout.css`; no JS reads `innerWidth/innerHeight`), so the stale size is not an app CSS issue — the webContents/compositor appears not to pick up the new size on this maximize path, leaving the window background color exposed.
- Possibly related upstream: [electron/electron#49203](https://github.com/electron/electron/issues/49203) ("Maximized frameless window has 2px of not rendered area at the bottom") and Electron's tech talk on window resize behavior (stale-frame/gutter fixes landed in Electron 39.2.6+; this still reproduces on 43.2.0, so the maximize transition may be a distinct path).
- Candidate mitigations to evaluate: force a geometry sync on the `maximize` event (e.g. re-apply `getBounds()` or a 1px toggle), and/or check whether it reproduces without `titleBarOverlay`.

1625567290:
## Summary

- Re-run Electron's root view layout after a native Windows maximize so the primary renderer fills the new client area.
- Coalesce maximize and resize notifications, and skip work after restore or teardown.
- Add focused coverage for platform gating, event coalescing, layout/repaint order, and destroyed-window guards.

Fixes #3416

## Verification

- Windows maximize renderer sync tests: 4/4 passed.
- Desktop suite on the refreshed base: 1095/1096 passed. The only failure was the existing shell environment timeout test reading its temporary PID file before creation under full-suite concurrency.
- Re-ran `apps/desktop/dist/main/__tests__/shell-env.test.js` three times in isolation: 12/12 passed on every run.

## Validation limit

The current host is macOS, so the packaged Windows 11 maximize transition still needs Windows CI or a Windows machine for final visual confirmation.

### DIFF
--- a/apps/desktop/src/main/__tests__/windows-maximize-renderer-sync.test.ts
+++ b/apps/desktop/src/main/__tests__/windows-maximize-renderer-sync.test.ts
@@ -1,3 +1,22 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
 import assert from 'node:assert/strict';
 import { describe, it } from 'node:test';
 import {
--- a/apps/desktop/src/main/__tests__/windows-maximize-renderer-sync.test.ts
+++ b/apps/desktop/src/main/__tests__/windows-maximize-renderer-sync.test.ts
@@ -1,0 +1,105 @@
+import assert from 'node:assert/strict';
+import { describe, it } from 'node:test';
+import {
+  createWindowsMaximizeRendererSync,
+  type MaximizedRendererSyncWindow,
+} from '../windows-maximize-renderer-sync.js';
+
+function createFixture() {
+  const calls: string[] = [];
+  const deferred: Array<() => void> = [];
+  let destroyed = false;
+  let maximized = true;
+  let webContentsDestroyed = false;
+  const contentView = {};
+  const window: MaximizedRendererSyncWindow<object> = {
+    contentView,
+    webContents: {
+      isDestroyed: () => webContentsDestroyed,
+      invalidate: () => calls.push('invalidate'),
+    },
+    isDestroyed: () => destroyed,
+    isMaximized: () => maximized,
+    setContentView: (view) => {
+      assert.equal(view, contentView);
+      calls.push('layout');
+    },
+  };
+
+  return {
+    calls,
+    deferred,
+    window,
+    defer: (callback: () => void) => deferred.push(callback),
+    setDestroyed: (value: boolean) => { destroyed = value; },
+    setMaximized: (value: boolean) => { maximized = value; },
+    setWebContentsDestroyed: (value: boolean) => { webContentsDestroyed = value; },
+  };
+}
+
+describe('Windows maximize renderer sync', () => {
+  it('defers one root layout and repaint for a maximized Windows window', () => {
+    const fixture = createFixture();
+    const schedule = createWindowsMaximizeRendererSync(fixture.window, {
+      platform: 'win32',
+      defer: fixture.defer,
+    });
+
+    schedule();
+    schedule();
+    assert.equal(fixture.deferred.length, 1);
+    assert.deepEqual(fixture.calls, []);
+
+    fixture.deferred.shift()?.();
+    assert.deepEqual(fixture.calls, ['layout', 'invalidate']);
+  });
+
+  it('does nothing on non-Windows platforms', () => {
+    const fixture = createFixture();
+    const schedule = createWindowsMaximizeRendererSync(fixture.window, {
+      platform: 'darwin',
+      defer: fixture.defer,
+    });
+
+    schedule();
+    assert.equal(fixture.deferred.length, 0);
+    assert.deepEqual(fixture.calls, []);
+  });
+
+  it('drops deferred work when the window leaves maximized state or is destroyed', () => {
+    const restored = createFixture();
+    const scheduleRestored = createWindowsMaximizeRendererSync(restored.window, {
+      platform: 'win32',
+      defer: restored.defer,
+    });
+    scheduleRestored();
+    restored.setMaximized(false);
+    restored.deferred.shift()?.();
+
+    const destroyed = createFixture();
+    const scheduleDestroyed = createWindowsMaximizeRendererSync(destroyed.window, {
+      platform: 'win32',
+      defer: destroyed.defer,
+    });
+    scheduleDestroyed();
+    destroyed.setDestroyed(true);
+    destroyed.deferred.shift()?.();
+
+    assert.deepEqual(restored.calls, []);
+    assert.deepEqual(destroyed.calls, []);
+  });
+
+  it('does not touch a destroyed WebContents', () => {
+    const fixture = createFixture();
+    const schedule = createWindowsMaximizeRendererSync(fixture.window, {
+      platform: 'win32',
+      defer: fixture.defer,
+    });
+
+    schedule();
+    fixture.setWebContentsDestroyed(true);
+    fixture.deferred.shift()?.();
+
+    assert.deepEqual(fixture.calls, []);
+  });
+});
--- a/apps/desktop/src/main/__tests__/windows-maximize-renderer-sync.test.ts
+++ b/apps/desktop/src/main/__tests__/windows-maximize-renderer-sync.test.ts
@@ -121,4 +121,23 @@ describe('Windows maximize renderer sync', () => {
 
     assert.deepEqual(fixture.calls, []);
   });
+
+  it('reports a native layout failure without stranding future syncs', () => {
+    const fixture = createFixture();
+    const failure = new Error('native layout failed');
+    const errors: unknown[] = [];
+    fixture.window.setContentView = () => { throw failure; };
+    const schedule = createWindowsMaximizeRendererSync(fixture.window, {
+      platform: 'win32',
+      defer: fixture.defer,
+      reportError: (error) => errors.push(error),
+    });
+
+    schedule();
+    assert.doesNotThrow(() => fixture.deferred.shift()?.());
+    assert.deepEqual(errors, [failure]);
+
+    schedule();
+    assert.equal(fixture.deferred.length, 1);
+  });
 });
--- a/apps/desktop/src/main/windows-maximize-renderer-sync.ts
+++ b/apps/desktop/src/main/windows-maximize-renderer-sync.ts
@@ -1,3 +1,22 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
 export interface MaximizedRendererSyncWindow<ContentView = unknown> {
   readonly contentView: ContentView;
   readonly webContents: {
--- a/apps/desktop/src/main/windows-maximize-renderer-sync.ts
+++ b/apps/desktop/src/main/windows-maximize-renderer-sync.ts
@@ -1,0 +1,48 @@
+export interface MaximizedRendererSyncWindow<ContentView = unknown> {
+  readonly contentView: ContentView;
+  readonly webContents: {
+    isDestroyed(): boolean;
+    invalidate(): void;
+  };
+  isDestroyed(): boolean;
+  isMaximized(): boolean;
+  setContentView(view: ContentView): void;
+}
+
+interface MaximizedRendererSyncOptions {
+  platform?: NodeJS.Platform;
+  defer?: (callback: () => void) => void;
+}
+
+/**
+ * Re-runs Electron's root view layout after a native Windows maximize.
+ *
+ * Electron's BrowserWindow WebContentsView and the public contentView are
+ * siblings under one default-fill root view. Re-applying the same contentView
+ * makes Electron invalidate and immediately lay out that root without changing
+ * the native window bounds or its restored bounds. The repaint then covers the
+ * newly maximized client area.
+ */
+export function createWindowsMaximizeRendererSync<ContentView>(
+  window: MaximizedRendererSyncWindow<ContentView>,
+  options: MaximizedRendererSyncOptions = {},
+): () => void {
+  const platform = options.platform ?? process.platform;
+  const defer = options.defer ?? setImmediate;
+  let pending = false;
+
+  return () => {
+    if (platform !== 'win32' || pending) return;
+    if (window.isDestroyed() || !window.isMaximized()) return;
+    pending = true;
+
+    defer(() => {
+      pending = false;
+      if (window.isDestroyed() || !window.isMaximized()) return;
+      if (window.webContents.isDestroyed()) return;
+
+      window.setContentView(window.contentView);
+      window.webContents.invalidate();
+    });
+  };
+}
--- a/apps/desktop/src/main/windows-maximize-renderer-sync.ts
+++ b/apps/desktop/src/main/windows-maximize-renderer-sync.ts
@@ -31,6 +31,7 @@ export interface MaximizedRendererSyncWindow<ContentView = unknown> {
 interface MaximizedRendererSyncOptions {
   platform?: NodeJS.Platform;
   defer?: (callback: () => void) => void;
+  reportError?: (error: unknown) => void;
 }
 
 /**
@@ -48,6 +49,9 @@ export function createWindowsMaximizeRendererSync<ContentView>(
 ): () => void {
   const platform = options.platform ?? process.platform;
   const defer = options.defer ?? setImmediate;
+  const reportError = options.reportError ?? ((error: unknown) => {
+    console.warn('[desktop] Windows maximize renderer sync failed:', error);
+  });
   let pending = false;
 
   return () => {
@@ -57,11 +61,17 @@ export function createWindowsMaximizeRendererSync<ContentView>(
 
     defer(() => {
       pending = false;
-      if (window.isDestroyed() || !window.isMaximized()) return;
-      if (window.webContents.isDestroyed()) return;
+      try {
+        if (window.isDestroyed() || !window.isMaximized()) return;
+        if (window.webContents.isDestroyed()) return;
 
-      window.setContentView(window.contentView);
-      window.webContents.invalidate();
+        window.setContentView(window.contentView);
+        window.webContents.invalidate();
+      } catch (error) {
+        // A best-effort layout correction must not terminate the main process
+        // if Electron tears down the native window between the guards and call.
+        reportError(error);
+      }
     });
   };
 }
--- a/scripts/verify-packaged-app.mjs
+++ b/scripts/verify-packaged-app.mjs
@@ -412,10 +412,10 @@ export function rendererViewportMatchesNativeClient(layout, nativeWindow) {
   if (!Number.isFinite(nativeWindow?.clientHeight) || nativeWindow.clientHeight <= 0) return false;
   const widthScale = nativeWindow.clientWidth / layout.innerWidth;
   const heightScale = nativeWindow.clientHeight / layout.innerHeight;
-  return (
-    dimensionsMatch(widthScale, heightScale, 0.01) &&
-    dimensionsMatch(widthScale, layout.devicePixelRatio, 0.05)
-  );
+  // Electron can report CSS or physical viewport pixels depending on the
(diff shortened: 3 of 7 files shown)


apps/desktop/src/main/__tests__/windows-maximize-renderer-sync.test.ts
```diff
@@ -0,0 +1,105 @@
...
+      defer: fixture.defer,
+    });
+
+    schedule();
+    schedule();
+    assert.equal(fixture.deferred.length, 1);
+    assert.deepEqual(fixture.calls, []);
+
+    fixture.deferred.shift()?.();
+    assert.deepEqual(fixture.calls, ['layout', 'invalidate']);
```
Astro-Han, line 73: [P2] This unit test proves only that the two mocked methods are called; it cannot show that Electron 43 actually expands and repaints the renderer after the native Windows maximize transition, which is the entire user-visible contract of this PR. Could we add a Windows packaged/CI smoke that restores → maximizes → waits one frame, then checks renderer `innerWidth`/`innerHeight` (or WebContents size) against the maximized client bounds and restores again? If we keep only one additional test, the real Windows transition is more valuable than another mock branch.

Astro-Han (commented):
Thanks for isolating the workaround behind a small Windows-only scheduler. The lifecycle guards and coalescing look sound, and I did not find a production correctness issue in the implementation. The remaining gap is that the test proves calls against a mock, while the bug and the proposed fix both live in Electron's real Windows layout behavior; I left that boundary inline.

AI-assisted review disclosure: OpenAI Codex performed an independent exact-head review. I verified the implementation, lifecycle guards, test scope, and live PR state, and I made the final review decision.

Astro-Han:
Heads up on the failing `Release Windows check` — **it isn't caused by anything in this PR.**

It fails at `scripts/verify-windows-sandbox-e2e.mjs:97` with:

```
```

That's a regression from #3001, which merged into main earlier today: `FilesystemWorkerClient` now rejects a write to an existing file when the caller doesn't supply `expectedIdentity`, and the Windows sandbox verifier calls the client directly rather than through `FilesystemExecutor`. Tracked in #3484 — nothing for you to fix here.

The `CI` job on this head is green, and the review of the actual change is clean: the native `GetClientRect`/`IsZoomed` probes are a stronger check than the CDP path they replace, and adding `rendererViewportMatchesNativeClient` catches DPI-scaling mismatches the previous viewport comparison couldn't. We also confirmed the 108 removed lines are the superseded CDP helpers rather than unrelated checks, and that the packaged-Git verification strengthened in #3473 is untouched.

We'll approve once #3484 is resolved and the Windows job can go green on this head.

---
*AI-assisted review.*

Astro-Han (commented):
Thanks — the diagnosis here is the part that matters, and it is right: `maximize` is where this reproduces, but the actual gap is that Electron 43's root View layout/repaint does not follow the native maximize. Naming the trigger and the cause separately is what keeps this from becoming a CSS patch.

Reviewed at exact head `83d3d12174620640ea8475cde7673534c8c3142a` against base `f83b469b557a30036d11572d3c14d69bc0766eaa`. **No P0–P3.**

What we checked, since 495 lines for a window-resize fix invites the question:

- **The production change is 48 lines** and sits inside the existing main-window lifecycle. Re-calling `setContentView(window.contentView)` is not a workaround — `NativeWindowViews::SetContentView` detaches and re-attaches the content view and then runs `FlushPendingRootLayout`, which is exactly the layout flush that did not happen. Reusing the same `contentView` means no new size state is introduced; `pending` only coalesces events. There is no second size authority here.
- **The remaining ~270 lines are the verifier**, and they buy something real: the previous coverage asserted that a mock was called. This reads the native client area through PowerShell `GetClientRect`/`IsZoomed` and the renderer viewport through CDP, then walks normal → maximized → normal checking DPR and `html`/`body`/`#root`/`.appFrame` alignment against the native client. For a bug that only exists at the native/renderer boundary on Windows, that is the only evidence that means anything from a macOS checkout.
- **Guards cover the failure paths**: window destroyed, WebContents destroyed, restore, non-Windows. We looked for a stale-`pending` path that could fire against a destroyed window and did not find one.

Verified locally: helper tests 4/4, `node --test scripts/verify-packaged-app.test.mjs scripts/verify-windows-harness.test.mjs` 51/51, strict TS compile clean, `git diff --check` clean.

**Not approving yet, and it is not about the code.** The red `test` and `package` runs on this head are stale merge refs — they were created at 11:23Z against a base that predates the `midTurn` fix that landed on `main` at 11:49Z, and the failure in the log is `packages/cli/src/pi-tui-runner.ts` missing `midTurn`, which has nothing to do with this PR. A rerun will not help; `pull_request` runs test the merge commit computed at event time, so only a new push recomputes it. Please rebase, and we will take the result on the new head.

---

Astro-Han:

apps/desktop/src/main/__tests__/windows-maximize-renderer-sync.test.ts
1625567290: Addressed. The Windows release check now runs the packaged app through a native restore → maximize → restore smoke. It reads `GetClientRect` and `IsZoomed` from the real HWND, reads the renderer viewport through CDP, and verifies DPR plus `html`/`body`/`#root`/`.appFrame` alignment against the native client bounds. The macOS helper and harness coverage is green; the new Windows workflow run is currently waiting for maintainer approval.

1625567290:
Rebased onto current `main` at `f19eede03` and force-updated the branch to `9fcd1826d`. The workflow trigger conflict is resolved, both new TypeScript files have ASF headers, and GitHub now reports the PR as mergeable.

Local validation completed:
- `npm run build`
- Desktop typecheck
- lint and format checks
- ASF header check
- Windows helper/harness tests: 65/65
- maximize renderer sync tests: 4/4

The new `CI` and `Release Windows check` runs are currently marked `action_required`, so they need maintainer approval before they can execute. I also replied to and resolved the packaged Windows smoke review thread. Please approve the two workflow runs; I will follow up once they finish.

Astro-Han:
Reviewed at `9fcd1826`. No blocking findings from me. Recording what I checked and one small note.

The problem statement holds — a native Windows maximize changes the HWND client area, and Electron's root view layout does not reliably follow the native transition, so the renderer keeps the pre-maximize height. That is an ordinary user path, not a constructed one.

What I liked about the shape of the fix: `createWindowsMaximizeRendererSync` does not introduce a second source of truth for window size. It re-applies the *same* `contentView` to force a root-view layout, and never touches native bounds or restored bounds. The guards are complete in both directions — it returns early on non-`win32`, on a destroyed window, on a non-maximized window, and re-checks all of those inside the deferred callback, plus `webContents.isDestroyed()`. The `pending` flag coalesces bursts. I walked the deferred callback's exits for restore, window teardown, WebContents teardown, repeated events, and non-win32, and did not find one that escapes a guard.

The verification is the strongest part. Rather than asserting against mocks, the new packaged smoke drives `normal → maximized → normal` through the real HWND (`GetClientRect` / `IsZoomed`) and checks DPR, document and visual viewport, and `html` / `body` / `#root` / `.appFrame` against the native client rect over CDP. That also answers the question I had while reading — whether *restore* is broken the same way as maximize, given the mechanism is symmetric — because the smoke exercises the restore leg too.

**[P3] `handleResize` and `handleMaximize` are byte-identical.**

```js
const handleResize = (): void => { scheduleSave(); scheduleMaximizedRendererSync(); };
const handleMaximize = (): void => { scheduleSave(); scheduleMaximizedRendererSync(); };
```

Two names for one function, bound to two events. One shared handler would read the same and remove the chance of the two drifting apart later under the impression that they are meant to differ. Purely cosmetic — no behavioural consequence, since `pending` already coalesces the pair of events Windows emits on maximize.

**On CI:** both workflow runs on this head are `action_required` with 0s elapsed, so nothing has actually executed. That is a fork-PR approval gate rather than anything wrong with the branch, but it does mean there is no CI evidence for this head yet — worth a maintainer kicking the runs off, particularly since the Windows smoke this PR adds is the part that would prove the fix. I did not treat the author's local verification or the reviews on earlier heads as a substitute.

One thing I could not check: the fix depends on `setContentView(sameView)` causing Electron to invalidate and re-lay-out the root view. That is a side effect of the current implementation rather than a documented contract, so a future Electron could stop honouring it. The packaged Windows smoke is the right mitigation — it would fail loudly rather than silently regressing — so this is a note, not a finding.

1625567290:
@Astro-Han Thanks for the review. Addressed the P3 in `c733e773a`: `resize` and `maximize` now share the same `handleWindowGeometryChange` callback, removing the duplicate handlers without changing behavior.

Local validation is green:
- desktop main-process build
- Windows maximize renderer-sync tests (4/4)
- desktop typecheck
- repository lint and format checks
- ASF header check
- `git diff --check`

The updated-head `CI` and `Release Windows check` workflows were created and are currently waiting for maintainer approval to run.

apps/desktop/src/main/windows-maximize-renderer-sync.ts
```diff
@@ -0,0 +1,67 @@
...
+  const platform = options.platform ?? process.platform;
+  const defer = options.defer ?? setImmediate;
+  let pending = false;
+
+  return () => {
+    if (platform !== 'win32' || pending) return;
+    if (window.isDestroyed() || !window.isMaximized()) return;
+    pending = true;
+
+    defer(() => {
```
Astro-Han, line 62: **[P3] The deferred body has no `try`/`catch`, so a throw here becomes a main-process `uncaughtException`.**

```ts
defer(() => {
  pending = false;
  if (window.isDestroyed() || !window.isMaximized()) return;
  if (window.webContents.isDestroyed()) return;
  window.setContentView(window.contentView);
  window.webContents.invalidate();
});
```

This runs on a `setImmediate` tick with no caller left on the stack, so nothing above it can catch. The three guards cover the states this module knows about, but `setContentView` and `invalidate` are native calls whose failure modes are Electron's, not ours — a window teardown landing between the guard and the call, or a native error on an unusual driver/DPI path, has nowhere to go but the process.

**Why it is P3:** the guards are genuinely good. I checked that `pending = false` is the **first** statement in the deferred body, before both early returns, so an early return can never strand the flag and permanently disable resync. That was the failure I went looking for, and it is not there.

**Why it is still worth a `try`/`catch`:** this is layout-polish code. Its entire job is making a repaint cover the newly maximized client area. **Cosmetic correction code should not be able to take the main process down**, and the asymmetry between what it does and what it can cost is the whole argument. A `catch` that swallows and logs leaves the worst case as "one maximize did not resync" — which is exactly the bug this file exists to soften.

*Independent line, verified at the gate: 2026-08-23 12:40 UTC.*

scripts/verify-packaged-app.mjs
```diff
@@ -353,6 +353,268 @@ export async function waitForUsableRenderer(
...
+export function rendererViewportMatchesNativeClient(layout, nativeWindow) {
+  if (!rendererLayoutMatchesViewport(layout)) return false;
+  if (!Number.isFinite(layout.devicePixelRatio) || layout.devicePixelRatio <= 0) return false;
+  if (!Number.isFinite(nativeWindow?.clientWidth) || nativeWindow.clientWidth <= 0) return false;
+  if (!Number.isFinite(nativeWindow?.clientHeight) || nativeWindow.clientHeight <= 0) return false;
+  const widthScale = nativeWindow.clientWidth / layout.innerWidth;
+  const heightScale = nativeWindow.clientHeight / layout.innerHeight;
+  return (
+    dimensionsMatch(widthScale, heightScale, 0.01) &&
+    dimensionsMatch(widthScale, layout.devicePixelRatio, 0.05)
```
Astro-Han: **[P3] This DPR assertion looks specific to a 96-DPI runner and may fail on real high-DPI machines.**

```js
dimensionsMatch(widthScale, layout.devicePixelRatio, 0.05)
```

A real-machine verification run on Windows at 2560×1440 / **150% scaling** reported `devicePixelRatio = 1.5` while `innerWidth` came back as the physical `2560` rather than the CSS `1707` — which makes `widthScale` `1.0` and fails a check demanding `≈ 1.5`. CI runners are 96 DPI, so `dpr === 1` and the assertion is trivially satisfied there; it would never surface in the lane it currently runs in.

**Stated honestly, because it changes what you should do with this:** that measurement came from `npm run dev` under Vite, **not** from the packaged app this script actually drives. Packaged and dev can differ on exactly this (DPI-awareness manifest, `--force-device-scale-factor`, zoom factor), so **this is a suspicion with a concrete measurement behind it, not a confirmed failure of this script.** Confirming it needs one packaged run on a scaled display.

**Why raise it anyway:** the two neighbouring assertions — `rendererLayoutMatchesViewport` and `widthScale ≈ heightScale` — already carry the property this check exists for, namely that the renderer viewport covers the native client area **proportionally**. The third clause adds "and the proportion is exactly the DPR", which is a stronger claim about how Electron reports `innerWidth` under a given DPI configuration. If that turns out to be environment-specific, dropping it costs nothing and keeps the check honest on every display; if it does hold on packaged builds at 150%, it deserves a comment saying so, because the next person on a scaled monitor will suspect it.

Probe scripts and the raw JSON/screenshots from that run can be shared if a repro would help.

*Real-machine verification line, `c733e773` vs base `f19eede`. Reported 2026-08-23 12:40 UTC.*

Astro-Han (approved):
Approving `c733e773a363c34944b61c1880f3e2d1694b4907`. Both required checks are terminal green and bound to that exact SHA (`package` 32614918972, `test` 32614918937). **Three P3s, all inline, none blocking.**

- https://github.com/apache/maka/pull/3472#discussion_r3837687627 — the deferred body has no `try`/`catch`, so a native throw becomes a main-process `uncaughtException`.
- https://github.com/apache/maka/pull/3472#discussion_r3837687646 — the DPR clause in `rendererViewportMatchesNativeClient` may be specific to 96-DPI runners.
- **Child `WebContentsView` coverage** — discussed below, since it is about what the tests reach rather than about a line.

Thanks for folding in the earlier note about the two byte-identical geometry handlers; `handleWindowGeometryChange` reads better than the duplicate did.

### What was verified on a real machine

Windows x64, 2560×1440 at 150%, Electron 43.2.0. Three maximize/restore rounds plus six 80 ms rapid toggles on **both** the base and the patched build, with a full-viewport diagnostic pattern injected into the renderer so "did it paint correctly" became pixel-measurable, cross-checked three ways per step (screenshot, CDP layout probe, native client probe).

- Layout agreed with the native client area within 1 px at **every** step of all 14, including immediately after maximize, on both builds.
- The full-viewport border filled the entire client area in every maximized screenshot, consistent at t0 / 150 ms / 1.5 s.
- The rAF frame counter advanced continuously throughout, including after the rapid toggles — the paint pipeline never froze.

**The honest part, which I would rather state than bury: the base build did not reproduce the original bug on that machine.** The unpatched version painted correctly too. So that run **cannot** demonstrate "broken before, fixed after" — it can only demonstrate that the patch introduces no observable regression there. The "the bug exists and this fixes it" link rests on the packaged maximize/restore check this PR adds, which is green on this head.

Two caveats from that run, also stated rather than smoothed over: the window was backgrounded with three anti-throttling flags set (a fidelity concession, since a person was using the machine), and maximize/restore was driven through `ShowWindowAsync(SW_MAXIMIZE/SW_RESTORE)` — the native path the maximize button and Win+↑ both funnel into, but not the literal gestures.

### Child `WebContentsView` — the gap that was open, now filled

Re-applying the same `contentView` had **no coverage** for the embedded-browser child views: the unit test mocks an empty `contentView`, and the verify probe only inspects the main renderer. That mattered because `setContentView` is the one operation in this change that could plausibly disturb a child view's attachment.

It was closed two ways. A minimal reproduction on the same Electron version (`contentView` + child `WebContentsView`, reapply on and off, three rounds plus rapid toggles) showed the child **survives, stays in `contentView.children`, keeps its bounds, keeps rAF running, and captures correct pixels** — with both groups identical. Then the patched dev app with a real session and the browser panel open showed the child's frame counter advancing continuously through every maximize phase (+18–19 frames per 300 ms), no destruction, no detachment, and the DOM-side strip at the right position and full height.

**Not verified, and I am not going to claim it:** the child's *visual position* after maximize while visible — the run was interrupted twice by the machine's actual user and stopped rather than pressed. Multi-monitor and mixed-DPI were not reachable with one display.

*Reviewed at 2026-08-23 12:42 UTC. No P0–P2.*

scripts/verify-packaged-app.mjs
1625567290: Addressed in `dbd401e15`. The verifier no longer requires the native-to-renderer scale to equal `devicePixelRatio`; it now keeps the stable contract that the renderer tree fills its viewport and the native client width/height scales agree proportionally. A new `devicePixelRatio = 1.5` case covers a packaged environment that reports physical viewport pixels without weakening the stale-height rejection.

apps/desktop/src/main/windows-maximize-renderer-sync.ts
1625567290: Addressed in `dbd401e15`. The deferred callback now clears `pending` before entering a `try`/`catch`, catches guard and native layout/invalidation failures, and reports them through an injectable error reporter (defaulting to a Desktop warning) instead of allowing an uncaught exception. The new regression test proves the callback does not throw, reports the exact error, and remains schedulable afterward.

1625567290:
@Astro-Han Both new P3s are addressed in `dbd401e15` and replied to inline. The branch is also rebased onto current `main` (`6ada6b546`).

Post-rebase local validation is green: workspace dependency build, Desktop main-process build, maximize sync and Windows harness tests (43/43), Desktop typecheck, lint, format, ASF headers, and diff checks.

The new-head `CI` and `Release Windows check` runs are both `action_required`; please approve them so the exact-head checks can run.

Astro-Han:
Hi — thanks for the `dbd401e15` fix addressing the two P3s.

Heads-up that this branch now **conflicts with current `main`** and can't be merged as-is. I tested a rebase locally in a throwaway worktree (your branch was not touched); it stops on:

- `scripts/verify-windows-harness.test.mjs`

That's a single real source conflict, so it needs your judgement rather than a mechanical resolution:

```
# resolve, then
git push --force-with-lease
```

Once the branch is conflict-free and CI is green on the new head, I'll pick the review back up — the P3 fixes themselves aren't in question here, this is purely the branch having drifted behind `main`.

---

1625567290:
@Astro-Han Rebased once more onto current `main` (`a2f2a1afa`) after the branch drifted into conflict. The new head is `d95e89c5f`, and GitHub reports it mergeable.

The verifier conflict was resolved by retaining both the current rollback-registration coverage from `main` and this PR's maximize/renderer coverage.

Exact-head local validation:
- Desktop workspace dependencies and main-process build
- maximize renderer-sync tests: 5/5
- Windows harness tests: 39/39
- Desktop typecheck
- ASF header check
- `git diff --check`

Fresh exact-head checks should be created for this push.

The fresh workflows are waiting for maintainer approval: [CI 32638152502](https://github.com/apache/maka/actions/runs/32638152502) and [Release Windows check 32638152482](https://github.com/apache/maka/actions/runs/32638152482). Please approve both exact-head runs.

scripts/verify-packaged-app.mjs
```diff
@@ -353,6 +353,268 @@ export async function waitForUsableRenderer(
...
+  if (!rendererLayoutMatchesViewport(layout)) return false;
+  if (!Number.isFinite(layout.devicePixelRatio) || layout.devicePixelRatio <= 0) return false;
+  if (!Number.isFinite(nativeWindow?.clientWidth) || nativeWindow.clientWidth <= 0) return false;
+  if (!Number.isFinite(nativeWindow?.clientHeight) || nativeWindow.clientHeight <= 0) return false;
+  const widthScale = nativeWindow.clientWidth / layout.innerWidth;
+  const heightScale = nativeWindow.clientHeight / layout.innerHeight;
+  // Electron can report CSS or physical viewport pixels depending on the
+  // packaged app's DPI-awareness mode. Proportional agreement with the native
+  // client is the stable contract; equating that scale to DPR is not.
+  return dimensionsMatch(widthScale, heightScale, 0.01);
```
jackwener, line 418: [P3] This now checks only that the native/client width and height have the same scale, so any uniformly scaled stale renderer is accepted. Calling this production function with a renderer/layout of `800×600`, a maximized native client of `1600×1200`, and DPR `1` returns `true`, even though the renderer is still at the restored size. The existing guards do not catch it: the document can fill its own stale viewport and both scales are still exactly `2`. The negative test uses `1920×900`, whose aspect-ratio mismatch exercises a different case. Please add a uniformly scaled stale case and require the common scale to be approximately `1` (native reports CSS pixels) or approximately `devicePixelRatio` (native reports physical pixels). This is non-blocking because the reported #3416 path (`1240×820 → 1920×1080`) changes aspect ratio and remains covered.

jackwener (approved):
Acting on behalf of repository owner WAWQAQ under the standing Kabi authorization for PR review and APPROVE; exact head `d95e89c5fd5d4480b998ee0d2f4745dd8feb0a05`.

**GO: no P0–P2; one non-blocking P3 inline.** Incremental coverage included the full old-to-new range-diff, the added teardown-race error boundary, the packaged Windows geometry verifier, and separate reconciliation of reviews, inline comments, and issue comments. Focused validation: maximize sync 5/5 and Windows harness 39/39. Not independently re-run on a physical Windows host in this pass.

Fresh machine gates: OPEN, non-draft, MERGEABLE/CLEAN; exact-head `test` and `package` are completed/success. This agent-executed approval does not claim to satisfy the repository's separate required-human-review wording. No merge performed.

PR #3475, opened 2026-08-22
fix(cli): give rewind selections visible feedback and preserve mid-switch input

### LINKED ISSUE #3383: fix(cli): rewind selection gives no in-flight feedback — Enter is swallowed and editor input typed during the switch is clobbered
## What happened

In the TUI, after Esc Esc → Rewind picker → select a turn → Enter:

1. The picker closes immediately, but the transcript does **not** change for a noticeable window — there is zero in-TUI indication that a rewind is running, so it reads as "nothing happened" (or "it jumps a moment later").
2. Keystrokes during that window are mishandled:
   - **Enter is silently swallowed** (`editor.disableSubmit = true`).
   - Typed text and ↑ history recall still fill the editor — and are then **silently overwritten** when the rewind completes and refills the editor with the discarded turn's prompt.
3. When the switch finally lands, the visible change is easy to miss: the rewound transcript is a *prefix* of the old one, the only additions are one dim notice line and the editor refill, and the discarded turns remain visible in terminal scrollback (by design, scrollback cannot be rewritten — see the comment at `pi-tui-runner.ts:3128-3141`).

Net effect: the user cannot tell whether the rewind worked, and any input typed during the window is lost.

## Root cause

1. **Long serialized switch with no feedback.** `list.onSelect` hides the overlay and fires `void runControl(() => rewindToTurn(item.value))` (`packages/cli/src/pi-tui-runner.ts:2090`). Before the first visual change, the chain performs ~6–8 serialized runtime-host round trips:
   - `loadCurrentMessages` (`runtime-host-session-driver.ts:536`) — the picker already loaded these via `listRewindTargets` at open time; they are re-fetched
   - `getRuntimeHostSession` (`:548`) → `session.revision.create` (`:550`)
   - `switchSession` (`:436`): `getRuntimeHostSession` + `session.execution_boundary.query` + `#openSessionChannel` (reloads all messages/snapshot)
   - `shellRunHydration.hydrate` (`pi-tui-runner.ts:1447`)

   Only after all of this does the runner call `replaceTranscript` + `editor.setText(result.prompt)` + `requestRender` (`pi-tui-runner.ts:1505-1520`).

2. **Control-busy is invisible in the TUI body.** The activity strip renders `Working…` only when `turnElapsedMs` is set, i.e. only for prompt turns (`pi-transcript.ts:1410`). The `busy` flag in the metadata (`pi-tui-runner.ts:408`) has no consumer; control actions only signal via the terminal title and taskbar progress.

3. **Editor input during the window is clobbered.** `runControl` sets `editor.disableSubmit = true` (`pi-tui-runner.ts:636`), so Enter returns early inside the editor, but typing and ↑ history recall still mutate the draft. On completion `editor.setText(result.prompt)` (`:1512`) overwrites it. The safety comment assumes the editor was neutral — but that is only checked when the picker *opens*, not when the rewind *completes*. (Alt+Enter has an explicit busy guard that preserves the draft, `:980`; the plain-Enter path relies on the swallow and loses the content.)

4. **Silent-failure edges.** `runControl` early-returns `if (busy)` (`:633`) with no notice: if busy is claimed between picker open and select (e.g. a Goal auto-continuation starting a turn), the rewind is silently discarded. `/goal` control warns in the same situation (`:2714-2720`); rewind does not. Also `await turnActivity.activities.acquire(sessionId)` (`:644`) can stall the whole rewind invisibly while a session-activity lease is held.

## Proposed boundary

1. **Immediate feedback on select**: push a "正在回退…" notice (or render control-busy in the activity strip) synchronously before the first await, so the Enter keypress has a visible effect.
2. **Preserve the draft**: on completion, if the editor is non-empty (the user typed during the window), stash/append the draft instead of clobbering it with `setText`.
3. **No silent refusal**: when `runControl` drops a rewind because busy is held, push an error notice — match `/goal`'s behavior at `pi-tui-runner.ts:2714-2720`.
4. Optional: reuse the messages `listRewindTargets` already loaded instead of re-fetching them in `rewindToTurn`.

## How to reproduce

1. Open the `maka` TUI in a session with several turns (enough to overflow the viewport).
2. Esc Esc → Rewind picker → pick an early turn → Enter.
3. Immediately type some text, press ↑, press Enter: Enter does nothing, typed text appears in the editor.
4. A beat later the transcript jumps to the branch and the typed text is replaced by the rewound turn's prompt.

## Environment

- Maka commit: `57e08d834`
- Surface: CLI TUI (`pi-tui-runner`)
- OS: Linux

me2seeks:
Closes #3383.

## Problem

Selecting a turn in the Esc-Esc Rewind picker closed the overlay and then stayed silent for the whole branch-and-switch chain (~6–8 serialized runtime-host round trips), while control-busy renders nothing in the TUI body. In that window Enter was silently swallowed (`disableSubmit`), typed text / ↑ history recall accumulated in the editor — and were then clobbered by `editor.setText(result.prompt)` when the rewind landed. If busy was claimed while the picker was open (e.g. a Goal auto-continuation starting a turn), `runControl`'s early return dropped the rewind with no notice at all. Net effect: "nothing happened", or it happened a beat later with my input gone (#3383).

## Changes (`packages/cli/src/pi-tui-runner.ts`)

1. **In-progress feedback on selection.** `rewindToTurn` pushes a `正在回退到该轮之前…` notice synchronously before its first await, so the Enter keypress is visibly accepted immediately. On success the transcript replacement wipes it; on failure the catch splices it out so only the error notice remains.
2. **Draft wins over refill.** The refill now checks the editor at completion time: a non-empty draft (typed during the window) is newer user work and is kept; the rewound prompt stays recoverable from editor history (`submitPrompt` added it). The completion notice says which variant happened.
3. **Audible busy refusal.** The picker's `onSelect` refuses with an error notice when `busy` is already held — same contract as `/goal`'s guard at the former `runControl` silent early return.

## Tests (`packages/cli/src/__tests__/pi-tui-runner.test.ts`)

- `shows an in-progress notice while the rewind branch is being created` — deferred driver gate proves the notice renders while the switch is still in flight, and disappears once the branch lands;
- `keeps a draft typed while the rewind is in flight instead of overwriting it` — draft survives completion, Enter stays swallowed (`driver.prompts` stays empty), no refill happens;
- `refuses a rewind selection with a notice when another action claimed busy` — a Host-started turn claims busy under the open picker (the Goal-continuation shape); selection surfaces the refusal and `rewound` stays empty.

## Verification

- `packages/cli`: full suite green — **340/340 pass** (includes all 8 rewind tests);
- `tsc -p packages/cli` clean; `biome check` clean on both touched files.

## Out of scope (noted in the issue)

- Reusing the messages `listRewindTargets` already loaded to avoid the re-fetch inside `rewindToTurn` (perf-only follow-up).
- Rendering control-busy in the activity strip generally; this PR fixes the rewind path's feedback specifically.

## AI use

### DIFF
--- a/packages/cli/src/pi-tui-runner.ts
+++ b/packages/cli/src/pi-tui-runner.ts
@@ -1504,18 +1504,41 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
   // non-destructive and inherits the branch's resume guarantees.
   const rewindToTurn = async (turnId: string) => {
     resolvedInteractionIds.clear();
-    const result = await input.driver.rewindToTurn(turnId);
-    await applySwitchResult(result);
-    // Refill the editor with the discarded turn's prompt so the user can edit
-    // and resend it. The picker only arms when the editor is neutral (empty
-    // draft, no autocomplete), so overwriting the text loses no in-progress work.
-    editor.setText(result.prompt);
-    state.entries.push({
+    // Synchronous feedback before the first await: branching + switching takes
+    // several serialized runtime-host round trips, and control-busy renders
+    // nothing in the TUI body, so without this notice the picker's Enter looks
+    // dead until the branch lands (#3383). replaceTranscript wipes it on
+    // success; the catch removes it on failure so only the error stays.
+    const pendingNotice: (typeof state.entries)[number] = {
       kind: 'notice',
       level: 'info',
-      text: '已回退到该轮之前（分支为新任务，原任务保留），该轮 prompt 已回填输入框，可修改后重新发送。',
-    });
+      text: '正在回退到该轮之前…',
+    };
+    state.entries.push(pendingNotice);
     requestRender();
+    try {
+      const result = await input.driver.rewindToTurn(turnId);
+      await applySwitchResult(result);
+      // Refill the editor with the discarded turn's prompt so the user can edit
+      // and resend it — unless the user typed into the editor while the switch
+      // was in flight. The picker's neutral-editor guarantee only holds at open
+      // time, so a non-empty draft here is newer user work and wins; the prompt
+      // stays recoverable from the editor history (submitPrompt added it).
+      const refill = editor.getText().trim().length === 0;
+      if (refill) editor.setText(result.prompt);
+      state.entries.push({
+        kind: 'notice',
+        level: 'info',
+        text: refill
+          ? '已回退到该轮之前（分支为新任务，原任务保留），该轮 prompt 已回填输入框，可修改后重新发送。'
+          : '已回退到该轮之前（分支为新任务，原任务保留）。输入框已有未发送内容，未回填该轮 prompt（可按 ↑ 从输入历史找回）。',
+      });
+      requestRender();
+    } catch (error) {
+      const index = state.entries.indexOf(pendingNotice);
+      if (index >= 0) state.entries.splice(index, 1);
+      throw error;
+    }
   };
 
   const showBottomPicker = (picker: Component): OverlayHandle =>
@@ -2088,6 +2111,19 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
       'Rewind',
       items,
       (item) => {
+        // runControl drops the action silently when busy is already held (e.g.
+        // a Goal auto-continuation started while the picker was open). The
+        // overlay is already closed at this point, so say so instead of
+        // leaving a dead Enter — same contract as /goal's busy guard.
+        if (busy) {
+          state.entries.push({
+            kind: 'notice',
+            level: 'error',
+            text: '无法回退：当前有正在进行的操作 — 请等待其完成，或中断（Esc）后重试。',
+          });
+          requestRender();
+          return;
+        }
         void runControl(() => rewindToTurn(item.value));
       },
       {
--- a/packages/cli/src/pi-tui-runner.ts
+++ b/packages/cli/src/pi-tui-runner.ts
@@ -331,6 +331,13 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
   let lastTurnEscapeAt = 0;
   let lastIdleEscapeAt = 0;
   let lastIdleCtrlCAt = 0;
+  // Mirrors the editor's bracketed-paste buffering at the input seam: between a
+  // paste start marker and its end marker the editor holds incoming bytes in an
+  // internal buffer and getText() stays empty, so "editor is empty" must not
+  // treat that in-flight paste as absent user input (#3475 review). The marker
+  // matching deliberately mirrors the editor's own per-chunk includes() checks,
+  // so this flag agrees with what the editor will buffer.
+  let editorPastePending = false;
   type AttachedTurnContext =
     | { readonly kind: 'adopted'; readonly turn: MakaPreparedSessionTurn }
     | { readonly kind: 'external'; readonly turn: MakaAttachedSessionTurn };
@@ -1519,19 +1526,26 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
     try {
       const result = await input.driver.rewindToTurn(turnId);
       await applySwitchResult(result);
-      // Refill the editor with the discarded turn's prompt so the user can edit
-      // and resend it — unless the user typed into the editor while the switch
-      // was in flight. The picker's neutral-editor guarantee only holds at open
-      // time, so a non-empty draft here is newer user work and wins; the prompt
-      // stays recoverable from the editor history (submitPrompt added it).
-      const refill = editor.getText().trim().length === 0;
+      // Record the discarded turn's prompt in the editor history before
+      // deciding on the refill: prompts submitted in this TUI process are
+      // already there (addToHistory dedupes consecutive duplicates), but a
+      // session entered via startup resume or /resume has no entry yet, and
+      // the notice below promises ↑ recovery (#3475 review).
+      editor.addToHistory(result.prompt);
+      // Refill the editor with that prompt so the user can edit and resend it
+      // — unless newer user input arrived while the switch was in flight. The
+      // picker's neutral-editor guarantee only holds at open time, so this
+      // covers both a typed draft and a bracketed paste still being buffered
+      // (getText() stays empty until its end marker); either wins over the
+      // refill.
+      const refill = editor.getText().trim().length === 0 && !editorPastePending;
       if (refill) editor.setText(result.prompt);
       state.entries.push({
         kind: 'notice',
         level: 'info',
         text: refill
           ? '已回退到该轮之前（分支为新任务，原任务保留），该轮 prompt 已回填输入框，可修改后重新发送。'
-          : '已回退到该轮之前（分支为新任务，原任务保留）。输入框已有未发送内容，未回填该轮 prompt（可按 ↑ 从输入历史找回）。',
+          : '已回退到该轮之前（分支为新任务，原任务保留）。输入框已有未发送内容，未覆盖；该轮 prompt 已存入输入历史，可按 ↑ 找回。',
       });
       requestRender();
     } catch (error) {
@@ -3010,6 +3024,18 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
   refreshEditorCwd(cwd);
 
   tui.addInputListener((data) => {
+    // Track bracketed pastes before any consuming branch: this must observe
+    // every chunk the editor could buffer, regardless of what the rest of the
+    // listener decides (#3475 review).
+    if (data.includes('\x1b[200~')) {
+      // A paste begins; it is only complete when the end marker follows, here
+      // or in a later chunk.
+      editorPastePending = !data.slice(data.indexOf('\x1b[200~') + 6).includes('\x1b[201~');
+    } else if (editorPastePending && data.includes('\x1b[201~')) {
+      // The paste ends; bytes after the end marker may start another paste.
+      const remainder = data.slice(data.indexOf('\x1b[201~') + 6);
+      editorPastePending = remainder.includes('\x1b[200~');
+    }
     // Once closing has begun, swallow any buffered input that reaches the
     // listener while the terminal is being torn down.
     if (closed) return { consume: true };
--- a/packages/cli/src/__tests__/pi-tui-runner.test.ts
+++ b/packages/cli/src/__tests__/pi-tui-runner.test.ts
@@ -3928,9 +3928,62 @@ describe('Maka Pi TUI runner', () => {
     // The notice wraps across screen lines at 80 columns, so match against a
     // whitespace-collapsed copy instead of the raw output.
     const collapsedOutput = () => plainTerminalOutput(terminal.output()).replace(/\s+/g, '');
-    await waitFor(() => collapsedOutput().includes('未回填该轮prompt'));
+    await waitFor(() => collapsedOutput().includes('未覆盖'));
     await waitFor(() => editorInputText(terminal) === 'my draft');
     assert.equal(plainTerminalOutput(terminal.output()).includes('refilled: turn-1'), false);
+    // The ↑ recovery promise must hold even though nothing was submitted in
+    // this TUI process (a resumed session has no live-path history entry):
+    // the rewound prompt is recorded explicitly on completion (#3475 review).
+    // The first ↑ only jumps to line start while the cursor sits at col > 0;
+    // the second one then enters history recall.
+    terminal.input('\x1b[A');
+    terminal.input('\x1b[A');
+    await waitFor(() => editorInputText(terminal) === 'refilled: turn-1');
+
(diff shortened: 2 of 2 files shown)


jackwener:
在 `9910007ef` 上 review 时注意到两个仍可补齐的边界：

1. 进行中提示目前仍位于 session activity lease 之后。选择 rewind 后会先进入 `runControl`，并在 `pi-tui-runner.ts:643` 等待 `activities.acquire(sessionId)`；拿到 lease 后才调用 `rewindToTurn`，新提示则到 `:1517` 才入队。因此当 session activity 已被占用时，#3383 提到的等待窗口依然没有 TUI body 反馈。可以考虑在 lease acquire 前发出提示（或给 `runControl` 加同步 started hook），并用预先 `reserve('session-1')` 的测试覆盖。

2. `:1527` 在 editor 非空时保留新 draft，并提示可按 ↑ 找回被 rewind 的 prompt；但 editor history 只包含当前 TUI 进程内经 `submitPrompt` 加入的内容（`:820`）。启动时 resume 或 `/resume` 进来的历史 turn 仍会成为 rewind target，却不一定存在于这份内存 history。若用户在 rewind 途中输入 draft，原 prompt 不会回填，也无法保证能按 ↑ 恢复。建议显式把 `result.prompt` 放入 editor history（或提供可切换的 draft stash），再补一个 resumed-session 回归测试。

本地验证：CLI 340/340、CLI typecheck、Biome（两个改动文件）和 `git diff --check` 均通过。

packages/cli/src/pi-tui-runner.ts
```diff
@@ -1504,18 +1504,41 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
...
+    // dead until the branch lands (#3383). replaceTranscript wipes it on
+    // success; the catch removes it on failure so only the error stays.
+    const pendingNotice: (typeof state.entries)[number] = {
       kind: 'notice',
       level: 'info',
-      text: '已回退到该轮之前（分支为新任务，原任务保留），该轮 prompt 已回填输入框，可修改后重新发送。',
-    });
+      text: '正在回退到该轮之前…',
+    };
+    state.entries.push(pendingNotice);
```
Astro-Han, line 1524: [P2] This notice is synchronous only after `runControl()` has already acquired the session activity lease. If another Host operation holds that lease, selecting Rewind still waits with no TUI-body feedback — the lease-blocked path called out in #3383. Could the notice be emitted before lease acquisition, or could `runControl` expose a synchronous started hook? A test that pre-reserves the session activity should see the notice before the lease is released.

packages/cli/src/pi-tui-runner.ts
```diff
@@ -1504,18 +1504,41 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
...
+      // time, so a non-empty draft here is newer user work and wins; the prompt
+      // stays recoverable from the editor history (submitPrompt added it).
+      const refill = editor.getText().trim().length === 0;
+      if (refill) editor.setText(result.prompt);
+      state.entries.push({
+        kind: 'notice',
+        level: 'info',
+        text: refill
+          ? '已回退到该轮之前（分支为新任务，原任务保留），该轮 prompt 已回填输入框，可修改后重新发送。'
+          : '已回退到该轮之前（分支为新任务，原任务保留）。输入框已有未发送内容，未回填该轮 prompt（可按 ↑ 从输入历史找回）。',
```
Astro-Han: [P2] The “press ↑” recovery promise is not reliable for a session entered through startup resume or `/resume`: editor history is populated only by prompts submitted in this TUI process, so the rewound `result.prompt` may never be present. If the user typed a newer draft during rewind, the old prompt is then neither refilled nor recoverable as advertised. Could we explicitly add the rewound prompt to editor history (or keep a switchable draft stash) and cover a resumed-session rewind?

packages/cli/src/pi-tui-runner.ts
```diff
@@ -1504,18 +1504,41 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
...
     requestRender();
+    try {
+      const result = await input.driver.rewindToTurn(turnId);
+      await applySwitchResult(result);
+      // Refill the editor with the discarded turn's prompt so the user can edit
+      // and resend it — unless the user typed into the editor while the switch
+      // was in flight. The picker's neutral-editor guarantee only holds at open
+      // time, so a non-empty draft here is newer user work and wins; the prompt
+      // stays recoverable from the editor history (submitPrompt added it).
+      const refill = editor.getText().trim().length === 0;
```
Astro-Han: [P2] `editor.getText()` can still be empty while `StdinBuffer` is holding an incomplete bracketed paste (start marker received, end marker not yet received). This sets `refill = true`; when the paste completes, its content is appended to the rewound prompt, producing unintended combined input. Could the input seam expose a pending-paste/dirty generation and let that newer input win, with a split `ESC[200~ … ESC[201~` regression test across the rewind completion?

Astro-Han (commented):
Thanks for making rewind visibly pending and preserving newer editor input instead of overwriting it. Two independent exact-head passes found three remaining boundaries where the final-state contract is not yet complete; I have consolidated them inline so the fixes can stay focused in the existing TUI input/activity seams.

AI-assisted review disclosure: OpenAI Codex coordinated two independent exact-head reviews. I verified the retained activity-lease, resumed-history, and bracketed-paste paths, plus the live PR and CI state, and I made the final review decision.

Astro-Han (commented):
Thanks @me2seeks — following up at the same head, `9910007ef82a7776364b183b93790fd724a468d2`.

**Withdrawing my earlier first point.** I previously asked whether the busy notice could be emitted before lease acquisition, on the basis that the notice inside `rewindToTurn` only fires after `runControl()` has taken the session activity lease. That reading was incomplete: the picker's selection callback already has a synchronous guard ahead of `runControl`, which pushes `无法回退：当前有正在进行的操作` and returns without ever attempting the lease. That is exactly the shape I was asking for, and `BusyAfterPickerOpenDriver` covers the case where a Host operation takes busy while the picker is open. My apologies for the noise — that point is resolved and needs no change.

The other two points still stand at this head, and I am not repeating their details since the inline threads are still open. One clarification on severity, and one new minor note:

**On the `/resume` history gap — keeping this at P2 rather than lowering it.** The prompt is only pushed to history along the live path, so a session entered through `/resume` has no entry to recall. Resuming is an ordinary entry point rather than an unusual configuration, and the UI states that the prompt can be recovered with ↑ — so the failure is a promise the product makes and does not keep. For a prompt that may represent a long piece of work, and with no other way for the user to get it back, that is a real loss rather than a cosmetic gap.

**[P3] The draft guard is a single condition rather than per-target state.** Preservation keys off `editor.getText().trim().length === 0`, i.e. "there is already a draft, so do not overwrite". That prevents clobbering across a single switch, but it means the retained content is whatever happened to be in the editor rather than something associated with the target being switched away from. With repeated switches the user has no way to get back to an earlier target's text. Not a functional block, and the current behaviour is the safe direction — noting it because the accompanying wording implies more recall than the mechanism provides.

---

packages/cli/src/pi-tui-runner.ts
me2seeks: Agreed — withdrawing acknowledged. The synchronous busy guard ahead of `runControl` plus `BusyAfterPickerOpenDriver` cover this shape; no change made for this point.

packages/cli/src/pi-tui-runner.ts
me2seeks: Fixed in dee251e49. `rewindToTurn` now records the rewound prompt via `editor.addToHistory(result.prompt)` unconditionally, before deciding on the refill. The live path is a no-op (addToHistory dedupes consecutive duplicates), and a session entered through startup resume or `/resume` — which had no history entry at all — now gets one, so the notice's ↑ promise is true in every case. The wording was also tightened to state exactly what happened: input kept untouched, prompt recorded in history.

Test coverage: `keeps a draft typed while the rewind is in flight instead of overwriting it` now presses ↑ after completion (twice — the first ↑ only jumps to line start while the cursor is past col 0) and asserts the editor shows the rewound prompt. That test submits nothing before the rewind, so it exercises exactly the resumed-session shape you described.

packages/cli/src/pi-tui-runner.ts
me2seeks: Fixed in dee251e49. `editorPastePending` is now tracked at the global input listener (the seam every chunk crosses before focus dispatch), mirroring the editor's own per-chunk marker matching — start marker without an end marker in the same chunk sets it, the end marker clears it, and bytes after the end marker may re-arm it. The refill condition is now `editor.getText().trim().length === 0 && !editorPastePending`, so a paste still being buffered wins over the refill and lands alone when its end marker arrives.

I kept this in Maka's input seam rather than exposing `pasteBuffer` from pi-tui: the field is private in the package, and mirroring the editor's exact matching logic keeps the flag consistent with what the editor will buffer without patching upstream.

Test coverage: new `lets a bracketed paste still being buffered win over the prompt refill` splits `ESC[200~pasted half` / ` rest ESC[201~` around the rewind completion and asserts the editor ends up with exactly `pasted half rest` and no `refilled` content.

me2seeks:
Pushed `dee251e49` addressing the two standing P2s and the P3 note:

- **/resume history gap**: the rewound prompt is now recorded via `editor.addToHistory` unconditionally on completion — a no-op for prompts submitted in this process (consecutive-duplicate dedupe), a real entry for resumed sessions. The ↑ promise in the notice is now true in every case; wording tightened to say exactly what happened. Covered by the draft-preservation test, which recalls via ↑ with zero in-process submissions.
- **Bracketed paste race**: paste buffering is tracked at the global input listener using the same per-chunk marker matching the editor uses, and the refill requires no pending paste. A split `ESC[200~ … ESC[201~` regression test asserts the completing paste lands alone.
- **P3 (single-condition guard)**: agreed the retained content is whatever sits in the editor rather than per-target state. With unconditional history recording, every completed rewind accumulates its target's prompt in history, so repeated switches stay recoverable via ↑ even though only one editor draft exists; a per-target draft stash UI remains out of scope here. The notice wording no longer implies more than that.

Full `packages/cli` suite green locally (341 pass), biome clean.

Astro-Han (commented):
Thanks — the three points from the previous round are each answered on their own terms, and the way you answered them is worth naming. You withdrew one on the merits rather than changing code to satisfy it, you fixed one by widening a guarantee instead of narrowing it (`addToHistory` unconditionally, so the ↑ promise holds for `/resume` sessions that never had a history entry), and you fixed the third at the seam every chunk actually crosses rather than at the point where the symptom showed up.

Reviewed at exact head `2ef8efabbdab9bd1bc605b517fbc91e678c71328` against base `8b60ddffa89682c03238ddce6595f531eb2b6f29`. No checks have run on this head — not green, not red, simply never executed — so this stays a COMMENT regardless of how the code reads. One P3 and one design note below; no P0, P1 or P2.

**Re-verified against this head, not the head your replies were written on:**

- **Lease-ahead notice** — you withdrew it, and the withdrawal holds. The synchronous busy guard sits ahead of `runControl` in the same tick, so there is no window between the check and the control taking effect, and `BusyAfterPickerOpenDriver` covers the shape.
- **`/resume` ↑ promise** — `editor.addToHistory(result.prompt)` runs unconditionally before the refill decision. The live path is a no-op because consecutive duplicates dedupe; the resume path, which previously had no entry at all, now gets one. The test asserting recall with nothing submitted in-process pins exactly that case.
- **Bracketed paste race** — the mirror's per-chunk marker matching tracks the editor's own `isInPaste`/`pasteBuffer` behaviour, including re-arming on bytes after an end marker. The test splitting `ESC[200~` / `ESC[201~` pins paste winning over refill.

**[P3] `paddingX: 0 → 1` is unrelated to this change.** One line of visual adjustment, not mentioned in the description, in a PR that is otherwise about rewind feedback and refill safety. Not blocking — please confirm it is deliberate, or drop it so the diff stays about one thing.

**Design note, not a finding.** The paste mirror necessarily reimplements marker matching that lives in pi-tui's editor as private state, and you have already explained why (the field is private and patching upstream is out of scope for this PR). That reasoning is sound. The cost is that the two can drift silently: if the editor's matching ever changes, nothing here fails. If a comment naming pi-tui's version and the specific behaviour being mirrored is cheap to add, it would give the next person reading this a thread to pull.

**One thing we did not do:** we verified statically against the pi-tui source and this repository, and did not execute the test suite. Given that no CI has run on this head either, the change currently has no executed verification behind it at all. Getting a CI run on this head is worth more than anything else on this list.

---

Astro-Han (approved):
Following up now that CI has run on this head — `test` is green.

Still at `2ef8efabbdab9bd1bc605b517fbc91e678c71328`, unchanged since my earlier review. No P0–P2; the only open item is the P3 I noted then, the unrelated `paddingX: 0 → 1`. That doesn't block, so I'm approving — please drop it or confirm it's deliberate whenever you next touch the branch.

Approving.

---

PR #3476, opened 2026-08-22
fix(desktop): reload usage stats when the persisted range arrives

GabrielDrapor:
## Symptom

Settings remembers the last-open page. When a Settings window is restored **directly onto 使用统计**, every metric reads **0** under a correct-looking range chip (typically 全部) — until a manual refresh, or navigating away and back:

usage page restored with all metrics at zero

## Root cause

The section-change effect fetched usage stats exactly once:

```ts
useEffect(() => {
  if (section === 'usage') void reloadUsage();
}, [section]);
```

`reloadUsage()` defaults its range to `settings.usage.range` — but when the window restores straight onto this page, the Runtime Host settings are still loading, so `settings` is `createDefaultSettings()` and the fetch runs with the **`'24h'` default**. The persisted range then lands and moves the selected chip (e.g. to 全部), but nothing refetches: the page shows the default range's (often empty) dataset under the persisted range's chip. A manual refresh uses the live draft range and a tab round-trip re-fires the effect after settings are ready — which is why both "fix" it.

## Fix

- The section effect is keyed on the **effective range** and passes it explicitly — the persisted range's arrival (or any later range change) triggers the reload, and the effect is the single owner of range-driven fetches.
- The page's range chips now only persist; the surface refetches when the persisted value lands, instead of racing a second fetch against it.

## Regression

New story `UsagePersistedRangeRestore`: the bridge resolves `settings.get()` a beat late and serves stats **only** for the persisted `'all'` range; the story renders `SettingsStory section="usage"` (the restore path) and the play function asserts the totals come from the persisted range's dataset. Under the previous code the play times out — the mount-time `'24h'` fetch was the only one.

Verified: renderer + storybook tsconfigs compile; Desktop suite 1092 green; biome clean.

### DIFF
--- a/apps/desktop/stories/settings/settings-pages.stories.tsx
+++ b/apps/desktop/stories/settings/settings-pages.stories.tsx
@@ -8,6 +8,7 @@ import type {
   ThemePalette,
   ThemePreference,
   UpdateAppSettingsResult,
+  UsageRange,
   UsageStats,
 } from '@maka/core/settings';
 import type {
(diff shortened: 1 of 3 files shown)


apps/desktop/stories/settings/settings-pages.stories.tsx
```diff
@@ -1293,6 +1294,49 @@ export const UsageNarrow: Story = {
...
+ * range's (empty) numbers under the persisted range's selected chip until
+ * a manual refresh. The bridge makes the race explicit: stats exist only
+ * for the persisted 'all' range, and settings resolve a beat late. */
+const withUsagePersistedRangeBridge = (() => {
+  const settings = mergeSettings(createDefaultSettings(), { usage: { range: 'all' } });
+  return withScopedMakaBridge({
+    ...makaBridge,
+    settings: {
+      ...makaBridge.settings,
+      get: async () => {
```
Astro-Han: [P2] `usage` is client-owned, so the page projects its persisted range from `settings.getClient()`, not this Runtime Host `get()`. The unchanged client bridge still returns the default `24h`; this story therefore never reaches the intended persisted `all` state, and its play assertion would time out (the regular Storybook smoke does not execute play functions). Could the bridge delay/return `getClient()` with `usage.range = 'all'` instead, or seed that client setting directly? That would make the regression test exercise the real authority.

Astro-Han (commented):
Thanks for fixing the range reload at the component's actual persisted-state boundary; the production generation/lifecycle logic looks sound. I found one issue in the regression story itself: it currently delays the Runtime Host settings channel rather than the client-owned usage setting, so it does not exercise the state transition it describes.

AI-assisted review disclosure: OpenAI Codex performed an independent exact-head review. I verified the settings ownership, production lifecycle, Story bridge behavior, and live PR state, and I made the final review decision.

GabrielDrapor:
Fixed at head `f9c7aa644` — you're right on both counts: `usage` is client-owned (`settings-ownership.ts` projects `usage: client.usage`), so the persisted range rides in through `getClient()`, and my bridge was delaying the wrong channel.

The story bridge now seeds and delays `getClient()` (with `updateClient` routed to the same object) and leaves the Runtime Host `get()` stock — so the play exercises the real authority transition: mount fetch with the '24h' default → late-resolving client settings land `range: 'all'` → the keyed effect refetches → totals from the 'all' dataset render. The production-code comment that misattributed the source to the Runtime Host settings load is corrected to name the client channel as well.

(While in there I noticed the neighboring `withUsageStoryBridge` helper routes its `usage` overrides through the Runtime Host `get()`/`update` too, so the existing Usage stories' `activeTab`/`showDetails` seeds don't actually reach the page either — left untouched here as it predates this PR, but happy to fix it in a follow-up if you'd like.)

Astro-Han (approved):
Thanks @GabrielDrapor — reviewed at exact head `f9c7aa644df7aea59b5dd528741434427724b340`. No findings; approving.

This was looked at along two independent lines — asynchronous correctness, and structure plus test coverage — which agreed.

**Late-arriving persisted range**

`settings.usage.range` is client-owned and lands asynchronously via `getClient()`. The section effect now depends on the resolved range and passes it explicitly into `reloadUsage`, so the initial `24h` request is followed by a refetch once the persisted value (e.g. `all`) arrives, rather than leaving the wrong dataset on screen.

**Overlapping requests and unmount**

`reloadUsage` takes a ticket from `usageReloadTicketRef` before awaiting, and gates both the success path and the error toast on that ticket still being current, plus `settingsModalMountedRef`. A slower request for an older range therefore cannot overwrite a newer one, and a response arriving after the modal closes updates nothing and raises no toast.

**Single owner rather than a second path**

`setRange` in `usage-settings-page.tsx` now only persists; the fetch is driven solely by the parent effect reacting to the resolved range. That removes the previous "persist and also call `onReload` myself" path instead of adding a guard on top of it, so range remains a single authority with no duplicated state between parent and child.

**Regression coverage**

The new `UsagePersistedRangeRestore` story is range-sensitive rather than decorative: the bridge delays `getClient()` by 30 ms and returns `range: 'all'`, `usageStats` yields populated totals only for `all`, and the play assertion checks the rendered total. It fails if the refetch does not happen, which is the regression this PR fixes. The existing usage stories continue to cover the ordinary states.

Exact-head `test` SUCCESS; `git diff --check` clean.

---

PR #3478, opened 2026-08-22
fix(desktop): keep module refreshes latest-wins

### LINKED ISSUE #3477: fix(desktop): keep same-Host projection refreshes latest-wins
<details open>

## What happens

Default-Host-backed Skills, managed skill sources, bundled skill catalog, and scheduled tasks guard against a result from an old Host, but concurrent refreshes for the same Host have no latest-request ordering.

A reachable sequence is:

1. Refresh A starts for the current default Host and remains pending.
2. A later refresh B starts for the same Host, for example after a mutation or another refresh trigger.
3. B returns and commits newer presentation data.
4. A returns last, passes the Host-identity check, and overwrites B with its older result.

This behavior predates #3465 and does not block its cross-Host consistency fix. It was identified during review in https://github.com/apache/maka/pull/3465#discussion_r3835646099.

## Scope

- Skills
- Managed skill sources
- Bundled skill catalog
- Scheduled tasks
- Preserve the existing default-Host identity check

Project presentation is not in scope because it already owns a refresh generation. Diagnostics clipboard fallback behavior is unrelated and also out of scope.

## Acceptance criteria

- Each affected projection commits a result only if no newer refresh for that same projection has started.
- A focused ordering test holds refresh A pending, lets refresh B commit, then settles A and verifies that B remains presented.
- The solution remains local to refresh ordering; it does not add global state, cancellation, or another Host authority.

## AI use

OpenAI Codex assisted with investigating the review finding and drafting this issue. The human contributor remains responsible for the scope and decision.

</details>

<details>

## 问题表现

依赖默认 Host 的 Skills、managed skill source、bundled skill catalog 和 scheduled task 已能阻止旧 Host 的结果提交，但同一 Host 的并发刷新缺少 latest-request 排序。

可达时序如下：

1. 当前默认 Host 上的刷新 A 启动后保持 pending。
2. 同一 Host 上更晚的刷新 B 启动，例如由 mutation 后刷新或另一个刷新入口触发。
3. B 先返回并提交较新的展示数据。
4. A 最后返回，因为 Host identity 仍相同而通过校验，并用旧结果覆盖 B。

该问题早于 #3465 存在，不阻塞其跨 Host 一致性修复。问题来源于 https://github.com/apache/maka/pull/3465#discussion_r3835646099 的 review。

## 范围

- Skills
- Managed skill source
- Bundled skill catalog
- Scheduled task
- 保留现有默认 Host identity 校验

Project presentation 已经拥有 refresh generation，不在本 issue 范围内。诊断剪贴板 fallback 与本问题无关，同样不在范围内。

## 验收标准

- 每个受影响的 projection 仅在其后没有启动更新刷新时提交结果。
- 增加定向时序测试：刷新 A 保持 pending，刷新 B 先提交，随后 A 返回，最终仍展示 B。
- 方案仅负责局部 refresh ordering，不增加全局状态、取消机制或新的 Host authority。

## AI 使用

OpenAI Codex 协助调查 review finding 并起草本 issue；范围和决策仍由贡献者本人负责。

</details>

M4n5ter:
<details open>

## Summary

- Keep only the latest-started same-Host refresh authoritative for Skills, managed skill sources, bundled skill catalog, and scheduled tasks.
- Store four independent refresh generations with the module-data hook so ordering survives renderer re-renders.
- Preserve the existing default-Host identity guard and all mutation, error, and diagnostic behavior.

Fixes #3477

## Review focus

Now that #3465 has merged, this PR is based directly on `main`. The generation state owns request ordering only; it does not cancel requests or become another Host authority.

## Verification

- `npm run lint`
- `npm run format:check`
- `npm --workspace @maka/desktop test` — 1099 passed
- Focused same-Host ordering coverage exercises all four projections and forces a renderer re-render between the stale and current refreshes.

## AI use

Select exactly one:

Tool(s) and scope: OpenAI Codex assisted with implementation, tests, verification, scope review, and pull request drafting; the human contributor remains responsible for the change.

## Checklist

- [ ] No

</details>

<details>

## 概要

- 对 Skills、managed skill source、bundled skill catalog 和 scheduled task，仅允许同一 Host 上最后启动的刷新提交展示状态。
- 四个 projection 的 refresh generation 由 module-data hook 独立持有，因此 Renderer re-render 不会重置排序 authority。
- 保留现有默认 Host identity guard，不改变 mutation、错误反馈或诊断行为。

修复 #3477。

## 审查重点

#3465 已合并，因此本 PR 现在直接基于 `main`。Generation 只负责请求排序，不取消请求，也不成为新的 Host authority。

## 验证

- `npm run lint`
- `npm run format:check`
- `npm --workspace @maka/desktop test` — 1099 项通过
- 同一 Host 排序覆盖四个 projection，并在旧刷新与当前刷新之间强制 Renderer re-render。

## AI 使用

OpenAI Codex 协助了实现、测试、验证、范围复核和 PR 文案；贡献者本人仍对本次变更负责。AI 使用选项已在英文部分勾选。

## 检查清单

- 测试覆盖本次变更，并会在缺少实现时失败。
- lint、format、typecheck 和受影响测试均已在本地通过。
- 本 PR 包含行为变化，已在概要中说明。

</details>

### DIFF
--- a/apps/desktop/src/renderer/app-shell-skill-actions.ts
+++ b/apps/desktop/src/renderer/app-shell-skill-actions.ts
@@ -25,6 +25,8 @@ type ToastApi = {
   ): void;
 };
 
+type RefBox<T> = { current: T };
+
 export interface AppShellSkillActions {
   refreshSkills(options?: { shouldShowError?: () => boolean }): Promise<void>;
   refreshManagedSkillSources(options?: { shouldShowError?: () => boolean }): Promise<void>;
@@ -50,6 +52,11 @@ export interface AppShellSkillActions {
 export function createAppShellSkillActions(deps: {
   uiLocale: UiLocale;
   isSkillsSurfaceActive: () => boolean;
+  refreshGenerationsRef: RefBox<{
+    skills: number;
+    managedSkillSources: number;
+    bundledSkillCatalog: number;
+  }>;
   setSkills: Dispatch<SetStateAction<SkillEntry[]>>;
   setManagedSkillSources: Dispatch<SetStateAction<ManagedSkillSourceEntry[]>>;
   setBundledSkillCatalog: Dispatch<SetStateAction<BundledSkillCatalogEntry[]>>;
@@ -58,6 +65,7 @@ export function createAppShellSkillActions(deps: {
   const {
     uiLocale,
     isSkillsSurfaceActive,
+    refreshGenerationsRef,
     setBundledSkillCatalog,
     setManagedSkillSources,
     setSkills,
@@ -83,9 +91,12 @@ export function createAppShellSkillActions(deps: {
   });
 
   async function refreshSkills(options: { shouldShowError?: () => boolean } = {}) {
+    const generation = ++refreshGenerationsRef.current.skills;
     try {
       const next = await runOnDefaultRuntimeHost((host) => window.maka.skills.list(host));
-      await runIfDefaultRuntimeHostCurrent(next.host, () => setSkills(next.value));
+      await runIfDefaultRuntimeHostCurrent(next.host, () => {
+        if (generation === refreshGenerationsRef.current.skills) setSkills(next.value);
+      });
     } catch (error) {
       if (options.shouldShowError?.() ?? true) {
         reportRuntimeHostError(copy.refreshSkillsFailedTitle, copy.refreshSkillsFallback, error);
@@ -94,9 +105,14 @@ export function createAppShellSkillActions(deps: {
   }
 
   async function refreshManagedSkillSources(options: { shouldShowError?: () => boolean } = {}) {
+    const generation = ++refreshGenerationsRef.current.managedSkillSources;
     try {
       const next = await runOnDefaultRuntimeHost((host) => window.maka.skills.sources.list(host));
-      await runIfDefaultRuntimeHostCurrent(next.host, () => setManagedSkillSources(next.value));
+      await runIfDefaultRuntimeHostCurrent(next.host, () => {
+        if (generation === refreshGenerationsRef.current.managedSkillSources) {
+          setManagedSkillSources(next.value);
+        }
+      });
     } catch (error) {
       if (options.shouldShowError?.() ?? true) {
         reportRuntimeHostError(copy.refreshSourcesFailedTitle, copy.refreshSourcesFallback, error);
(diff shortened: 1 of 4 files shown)


apps/desktop/src/renderer/app-shell-skill-actions.ts
```diff
@@ -83,9 +91,12 @@ export function createAppShellSkillActions(deps: {
...
 
   async function refreshSkills(options: { shouldShowError?: () => boolean } = {}) {
+    const generation = ++refreshGenerationsRef.current.skills;
     try {
       const next = await runOnDefaultRuntimeHost((host) => window.maka.skills.list(host));
-      await runIfDefaultRuntimeHostCurrent(next.host, () => setSkills(next.value));
+      await runIfDefaultRuntimeHostCurrent(next.host, () => {
+        if (generation === refreshGenerationsRef.current.skills) setSkills(next.value);
+      });
     } catch (error) {
```
Astro-Han, line 100: [P3] The error path isn't generation-gated, so a superseded failure can still raise a toast.

The success path is guarded (`if (generation === refreshGenerationsRef.current.skills) setSkills(...)`), but the `catch` right below it calls `options.shouldShowError?.() ?? true` unconditionally. Same shape in `refreshManagedSkillSources` and `refreshBundledSkillCatalog`.

Reachable on the ordinary path this PR is about: slow refresh A is in flight, the user triggers B, B succeeds with fresh data, then A fails. The user sees a "refresh failed" toast while looking at data that is current and healthy — and nothing they can do makes it accurate, because the failure it describes has already been superseded.

So the projection is latest-wins for results but not for errors. Gating the catch the same way makes the two paths symmetric:

```ts
} catch (error) {
  if (generation !== refreshGenerationsRef.current.skills) return;
  if (options.shouldShowError?.() ?? true) {
```

Worth a test alongside the existing ones — the current suite covers stale success but not stale failure.

Astro-Han (approved):
Thanks — using a monotonic counter rather than a timestamp is the right call here, and worth saying out loud: timestamps break on same-millisecond refreshes and on clock adjustment, and both are reachable in exactly the scenario this fixes.

**What this solves / how:** four module projections (skills, managed skill sources, bundled skill catalog, scheduled tasks) weren't latest-wins — a slow earlier request landing after a newer one would overwrite fresher data, so the UI visibly reverted to older content. Most likely when a manual refresh races a surface-activation refresh. Each projection now carries a generation counter, bumped when a refresh starts, and only the matching generation is allowed to `setState`.

**Review notes:**
- The generation check and the `setState` sit in the same synchronous block, so there's no window between deciding and applying.
- Per-projection counters rather than one shared counter — correct, since an unrelated projection's refresh shouldn't invalidate another's result.
- The existing `runIfDefaultRuntimeHostCurrent` guard is kept alongside this rather than replaced. The two are orthogonal (which host is authoritative vs. which refresh is authoritative), so this isn't a second guard for the same fact.
- Superseded reads still run to completion and only their results are discarded — that's discarding, not cancellation. Fine at this scale: these are low-frequency, user- or activation-driven reads, so the wasted work is bounded and real cancellation would mean threading signals through the IPC layer for no practical gain.
- The tests bundle the real module and render through real React, reproducing "slow old request lands after new result" per projection. The `calls === 2` assertion also documents that this discards rather than cancels, which is the honest thing to pin.

One asymmetry inline.

---
*AI-assisted review. Findings were checked directly against this head.*

PR #3480, opened 2026-08-22
fix(runtime-host): let operators raise the election deadline and clarify timeout copy

### LINKED ISSUE #3474: Post-upgrade cold start can exceed the 45s election deadline, and abandoned candidates then hold the root lock
## Environment

- Linux, built from source at `944d82289` (v0.1.11), workspace first created by an older build (0.1.0-beta.1-era)
- Workspace: `~/.config/Maka/workspaces/default`, `runtime.sqlite` ≈ 788 MB, artifacts ≈ 4.3 GB

## Steps to reproduce

1. Upgrade the binary in place (old workspace data kept as-is).
2. Run `maka --resume <session-id>` (TUI path).

## Actual behavior

1. **First host start after upgrade is much slower than the fixed client deadline.** Measured via `maka runtime-host serve --json` on the same workspace: `runtime_host_ready` arrived after ~75s (cold caches + post-upgrade WAL/migration work on a 788 MB DB). The client election deadline is hard-coded at 45s:

   https://github.com/apache/maka/blob/944d82289/packages/runtime-host/src/client/connect-or-spawn.ts#L37

2. The TUI therefore fails with:

   ```
   Error: Runtime Host stopped responding during startup
   ```

   which maps to `host_unresponsive` in `packages/runtime-host/src/client/startup-error.ts`. The copy is misleading — the host was alive and making progress, just slower than the deadline.

3. **The abandoned candidate processes keep running and keep holding the root owner lock.** After the client gives up, every subsequent attempt fails fast with:

   ```
   RuntimeHostRootAlreadyOwnedError: Runtime Host root is already owned: ~/.config/Maka/workspaces/default
   ```

   instead of connecting to / adopting the already-starting host. Multiple orphaned `execution-candidate-main.js` processes accumulated after repeated resume attempts.

## Expected behavior

- Startup should tolerate slow-but-progressing hosts (larger/adaptative deadline or a user-configurable override), especially for one-time post-upgrade work on large workspaces.
- A candidate whose spawning client has given up should either self-terminate when not adopted within its startup-attempt window, or release the root ownership so the next attempt can proceed.
- The `host_unresponsive` error copy should distinguish "deadline elapsed while host still starting" from "host actually unresponsive".

## Suggestions

- Make `DEFAULT_ELECTION_DEADLINE_MS` overridable (env var / CLI flag) and/or scale it with observed workspace size.
- Add an adoption timeout to candidates: exit (and drop the lock) if no client claims them.
- Consider emitting progress state during post-upgrade migration so clients can distinguish migration work from a hang.

me2seeks:
Closes #3474

## Problem

On a large workspace (~788 MB `runtime.sqlite`), the first Runtime Host start after an upgrade emitted `runtime_host_ready` after ~75s — 30s after the client's fixed 45s election window had already elapsed. The user sees:

```
Error: Runtime Host stopped responding during startup
```

which is misleading: the host was alive and making progress the whole time. The candidates spawned during the election keep starting detached, so they eventually win and hold the root while every later attempt starts from a cold election again.

## Change

- `MAKA_RUNTIME_HOST_ELECTION_DEADLINE_MS` now overrides the default election deadline at its single resolution point in `connectOrSpawnRuntimeHostWithDependencies`, so CLI, desktop, and owned launches all honor it. The window stays bounded (1–120000 ms). Invalid values fail closed with a \`RangeError\` naming the variable — a silently ignored typo would leave the operator believing they widened the window when they did not.
- The two deadline-elapsed failure copies (`startup_timeout`, `host_unresponsive`) no longer claim things neither reason can know ('stopped responding'). They describe what was observed, name slow first starts on large workspaces as the likely cause, and point at the override.
- No candidate lifecycle change: ephemeral candidates already self-terminate via their initial connection deadline once startup settles; the lingering processes in #3474 were startup work starving the timer, not missing exit logic.

## Testing

- New `election-deadline-env.test.ts`: blank/unset handling, valid parsing, fail-closed on invalid values, and an ordering test proving an invalid override fails before storage is touched.
- `startup-error.test.ts` updated for the new copy plus a knob-mention assertion.
- Desktop manager assertions updated to the new copy.
- `packages/runtime-host`: full suite 1067/1067 green; desktop manager suite 22/22; `tsc --noEmit` and biome clean.

Generated-by: Maka

### DIFF
--- a/packages/runtime-host/src/client/connect-or-spawn.ts
+++ b/packages/runtime-host/src/client/connect-or-spawn.ts
@@ -38,6 +38,7 @@ const DEFAULT_ELECTION_DEADLINE_MS = 45_000;
 const DEFAULT_BACKOFF_MIN_MS = 20;
 const DEFAULT_BACKOFF_MAX_MS = 250;
 const MIN_CANDIDATE_INTERVAL_MS = 250;
+export const ELECTION_DEADLINE_MS_ENV_VAR = 'MAKA_RUNTIME_HOST_ELECTION_DEADLINE_MS';
 
 export interface ConnectOrSpawnRuntimeHostInput {
   rootPath: string;
@@ -56,12 +57,34 @@ export interface ConnectOrSpawnRuntimeHostInput {
 interface ConnectOrSpawnRuntimeHostDependencies {
   launchCandidate: CandidateLauncher;
   random(): number;
+  /** Defaults to `process.env`; injected so tests never mutate the real environment. */
+  env?: NodeJS.ProcessEnv;
 }
 
 const defaultDependencies: ConnectOrSpawnRuntimeHostDependencies = {
   launchCandidate: launchDetachedRuntimeHostCandidate,
   random: Math.random,
 };
+
+/**
+ * Resolves the operator override for the client election deadline. Large
+ * workspaces can legitimately take longer than the default window on their
+ * first start after an upgrade, so the deadline must be raisable without a
+ * code change. Invalid values fail closed: a silently ignored typo would leave
+ * the operator believing they widened the window when they did not.
+ */
+export function electionDeadlineMsFromEnvironment(
+  rawValue: string | undefined,
+): number | undefined {
+  if (rawValue === undefined || rawValue.trim() === '') return undefined;
+  const parsed = Number(rawValue);
+  if (!Number.isSafeInteger(parsed) || parsed <= 0 || parsed > 120_000) {
+    throw new RangeError(
+      `${ELECTION_DEADLINE_MS_ENV_VAR} must be an integer between 1 and 120000 milliseconds`,
+    );
+  }
+  return parsed;
+}
 
 export type ConnectOrSpawnRuntimeHostResult =
   | {
@@ -173,7 +196,12 @@ export async function connectOrSpawnRuntimeHostWithDependencies(
   input: ConnectOrSpawnRuntimeHostInput,
   dependencies: ConnectOrSpawnRuntimeHostDependencies,
 ): Promise<ConnectOrSpawnRuntimeHostResult> {
-  const deadlineMs = input.electionDeadlineMs ?? DEFAULT_ELECTION_DEADLINE_MS;
+  const deadlineMs =
+    input.electionDeadlineMs ??
+    electionDeadlineMsFromEnvironment(
+      (dependencies.env ?? process.env)[ELECTION_DEADLINE_MS_ENV_VAR],
+    ) ??
+    DEFAULT_ELECTION_DEADLINE_MS;
   if (!Number.isSafeInteger(deadlineMs) || deadlineMs <= 0 || deadlineMs > 120_000) {
     throw new RangeError('electionDeadlineMs must be an integer between 1 and 120000');
   }
--- a/apps/desktop/src/main/__tests__/runtime-host-desktop-manager.test.ts
+++ b/apps/desktop/src/main/__tests__/runtime-host-desktop-manager.test.ts
@@ -523,11 +523,11 @@ test('keeps Local explicitly usable without routing default work away from an un
     },
   );
 
-  await assert.rejects(manager.enable(remoteTarget('offline')), /stopped responding/);
+  await assert.rejects(manager.enable(remoteTarget('offline')), /did not become ready/);
   manager.setDefaultProfile('offline');
   await assert.rejects(
     manager.handleBotIncomingMessage({ text: 'default' } as BotIncomingMessage),
-    /stopped responding/,
+    /did not become ready/,
   );
 
   assert.equal(local.botMessages, 0);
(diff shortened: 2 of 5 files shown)


packages/runtime-host/src/client/connect-or-spawn.ts
```diff
@@ -56,13 +57,35 @@ export interface ConnectOrSpawnRuntimeHostInput {
...
+ * the operator believing they widened the window when they did not.
+ */
+export function electionDeadlineMsFromEnvironment(
+  rawValue: string | undefined,
+): number | undefined {
+  if (rawValue === undefined || rawValue.trim() === '') return undefined;
+  const parsed = Number(rawValue);
+  if (!Number.isSafeInteger(parsed) || parsed <= 0 || parsed > 120_000) {
+    throw new RangeError(
+      `${ELECTION_DEADLINE_MS_ENV_VAR} must be an integer between 1 and 120000 milliseconds`,
```
Astro-Han, line 83: [P3] The error text points every startup failure at this variable, including the class it can't fix.

There's a second failure shape this knob doesn't address. From a real CI run on #3454, which now emits election diagnostics:

```json
{"deadlineMs":45000,"elapsedMs":45174,"candidateLaunches":69,
 "observations":{"notRegistered":73,"connectFailed":0,"handshakeFailed":0,
                 "connected":1,"readyWaitFailed":1}}
```

69 candidate launches in 45 seconds. Zero connect failures, zero handshake failures — it connected once, then ready-wait failed, and candidates kept being launched roughly every 0.65s until the deadline. It missed by 174ms, so raising the deadline would have made *that particular run* pass, but the run isn't slow — it isn't converging. At 90s it launches ~138 candidates instead of ~69 and passing becomes a matter of luck.

That's different from #3474's genuinely-slow start, where waiting longer is exactly right. The problem is that both produce the same message, so an operator hitting the retry-storm shape is told to raise the timeout, raises it, fails again, and raises it further.

Not blocking, and possibly not yours to fix here — but worth asking: with `candidateLaunches` high and `connectFailed`/`handshakeFailed` at zero, should the message distinguish "this start is slow, wait longer" from "candidates aren't converging, the timeout isn't your problem"? You have the #3474 case in hand and probably know which shape operators hit more often.

packages/runtime-host/src/client/connect-or-spawn.ts
```diff
@@ -173,7 +196,12 @@ export async function connectOrSpawnRuntimeHostWithDependencies(
   input: ConnectOrSpawnRuntimeHostInput,
   dependencies: ConnectOrSpawnRuntimeHostDependencies,
 ): Promise<ConnectOrSpawnRuntimeHostResult> {
-  const deadlineMs = input.electionDeadlineMs ?? DEFAULT_ELECTION_DEADLINE_MS;
+  const deadlineMs =
+    input.electionDeadlineMs ??
+    electionDeadlineMsFromEnvironment(
```
Astro-Han, line 201: [P3] On the owned-launch path an invalid value loses the variable name and surfaces as `host_unresponsive`.

`electionDeadlineMsFromEnvironment()` throws a `RangeError` naming the variable, which is exactly right for direct `connectOrSpawnRuntimeHost`. But `connectOwnedRuntimeHostWithDependencies()` wraps the call in a broad catch that converts anything into `{ kind: 'failed', reason: 'host_unresponsive' }`. So `MAKA_RUNTIME_HOST_ELECTION_DEADLINE_MS=abc` reaches the user as `Runtime Host did not start: host_unresponsive` — they're told the host is unresponsive when in fact they made a typo, and the message gives them nothing to act on.

This is the same shape as the finding above: the mechanism is fine, the diagnosis points the wrong way. Config errors specifically are worth letting through that catch (or resolving before it, as a typed configuration failure), since they're the one class the user can fix immediately once they know the variable name. Worth a test on both the owned and hosted paths.

Astro-Han (approved):
Thanks — the bounded range and the fail-closed validation are the right shape for an operational override, and keeping it as one more input to the existing deadline (rather than a second timer) means there's still a single election authority.

**What this solves / how:** the fixed 45s election window is too short for real installs — #3474's 788MB workspace takes ~75s to reach ready after an upgrade, through cold start, WAL and one-time migration work. The client can't estimate that before root authority initialises, so it can't self-tune. `MAKA_RUNTIME_HOST_ELECTION_DEADLINE_MS` resolves at the single existing resolution point, defaults unchanged at 45s, and is clamped to 1–120000ms.

**On whether this belongs as a knob at all**, since that's the question worth asking of any new tunable: yes, for the case you're targeting. Startup cost genuinely varies with workspace size, disk cache state and pending upgrade work, and none of that is knowable client-side before the host comes up. A bounded escape hatch is the honest answer, and the closed interval keeps it from becoming "wait forever."

Two findings inline, both about diagnosis rather than the mechanism.

---
*AI-assisted review. Reviewed independently by two reviewers; the severity of the first finding was revised downward after checking it against a second real-world failure sample.

PR #3483, opened 2026-08-22
fixed inconsistencies and terminology

LuckyMan612:
## Summary

Fixes two small inconsistencies in README.md

## Verification

No verification is needed, since it's only documentation change.

## AI use

Select exactly one:

- [X] No generative tool made a substantive contribution
- [ ] Generative tooling made a substantive contribution

## Checklist

- [ ] Tests cover the change and fail without it (doesn't make sense for my change)
- [X] Lint, format, typecheck and the affected suites pass locally

- [X] No


jackwener (approved):
Two README consistency fixes look correct.

- `external competitors` → `external subjects` matches the Eval table already on line 34 (`Maka and external subjects`). Code in `packages/eval` also uses "external subject".
- Dropping Windows from the "not supported yet" list is right: `docs/windows-support.md` treats Windows as an unsigned preview tier, not the same bucket as Intel Macs and Linux.

Non-blocking: `README.zh-CN.md` and `ARCHITECTURE.md` still say competitor / 外部竞品. Out of scope for this one-file PR.

PR #2999, opened 2026-08-14
feat(cli): add an in-app transcript viewer

me2seeks:
<details open>
<summary>English</summary>

## Summary

Long TUI sessions currently depend on the terminal's native scrollback. Once the terminal discards old rows, the beginning of the session is no longer reachable even though the CLI still has the transcript projection.

This adds a read-only `/transcript` viewer that:

- opens at the latest part of the transcript and follows new output while it remains at the end;
- scrolls by line with Up/Down and by page with Page Up/Page Down;
- jumps with Home/End;
- closes with `q` or Escape.

The normal live surface remains unpaged and keeps using terminal scrollback. The viewer renders through the existing CLI transcript projection with detached geometry and stable clone keys, so browsing cannot replace the memoized lines or viewport coordinates used by live scrollback reconciliation.

Escape intentionally closes this read-only overlay. Editing an earlier turn remains on Maka's existing double-Escape rewind flow, which branches the session and refills the prompt; the viewer does not introduce a second mutation path.

Fixes #2997

## Before / after

The captures use the same 40-line TUI integration fixture. Before, the live screen has already scrolled past lines 1-21. After opening `/transcript` and pressing Home, the retained canonical projection makes the beginning reachable again.

### Before — live terminal scrollback only

Before: the live TUI only shows filler lines 22 through 40

### After — `/transcript` at Home

After: the transcript overlay shows the prompt and filler lines starting at 1

## Verification

- `npm --workspace packages/cli test` — 281/281
- focused TUI integration capture: `browses a long transcript without depending on terminal scrollback`
- `git diff --check`
- correctness review and `simplify-audit`: no P0-P3 findings

## Status

Ready for review. The requested visual evidence is included above.

## Checklist

- [x] Tests cover the change and fail without it
- [x] Lint, format, typecheck and the affected suites pass locally
- [x] Before/after captures document the UI behavior

Does this PR entail a change in behavior?

- [x] Yes — described under Summary above
- [ ] No

AI disclosure: Codex implemented the change under me2seeks's direction and review.

</details>

<details>
<summary>中文</summary>

## 概要

长 TUI 会话目前依赖终端自身的 scrollback。终端丢弃较早的行后，即使 CLI 仍保留 transcript 投影，用户也无法再回到会话开头。

本 PR 新增只读的 `/transcript` 浏览器：

- 打开时定位到 transcript 末尾，并在停留末尾时跟随新输出；
- Up/Down 按行滚动，Page Up/Page Down 按页滚动；
[...]

PR #3485, opened 2026-08-22
fix(cli): declare the transcript viewer's mid-turn disposition

Astro-Han:
## Summary

`main` does not typecheck. `MakaSlashCommand.midTurn` became required so every handler has to state whether it is safe to run while a turn is in flight; `/transcript` (#2999) landed in parallel and never got an answer:

```
packages/cli/src/pi-tui-runner.ts(2877,5): error TS2741: Property 'midTurn' is missing in type
'{ description: string; run: (parts: string[]) => void; }' but required in type
'Omit<MakaSlashCommand, "aliases" | "name">'.
```

Neither PR was wrong on its own — the conflict is semantic, so both merged green and `main` broke. The viewer is a full-screen overlay over the same TUI the running turn paints into, which is the picker case `'refuse'` exists for, so that is its declared answer.

Refs #2999

## Verification

- `npm run build` — passes with this change, fails on `main` without it
- `npm run format:check` — clean
- `npm test --workspace maka-agent` — 371 pass, 0 fail

## AI use

Select exactly one:

Tool(s) and scope: Claude Code diagnosed the cross-PR type break and wrote the one-line declaration and its comment. The human contributor reviewed the diff and the disposition choice.

## Checklist

The compiler is the test: `main` fails to build without this change.

- [ ] No

`/transcript` is now refused mid-turn instead of opening over a running turn.

### DIFF
--- a/packages/cli/src/pi-tui-runner.ts
+++ b/packages/cli/src/pi-tui-runner.ts
@@ -2876,6 +2876,9 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
     },
     transcript: {
       description: primaryGuidance.commands.transcript,
+      // The viewer is a full-screen overlay over the same TUI the turn paints
+      // into, so it is exactly the picker case 'refuse' exists for.
+      midTurn: 'refuse',
       run: (parts: string[]) => {
         if (parts.length !== 1) {
           state.entries.push({


Astro-Han:
Superseded by #3459, which landed the same declaration on `main` in b434a491. `main` builds again, so this is redundant — closing.

One correction worth recording for anyone who hits this next: the value that landed is `midTurn: 'local'`, not `'refuse'`. The type's own doc comment draws the line at whether the handler enters `runControl` — `'local'` requires the handler to be independent of the running turn, `'refuse'` is for anything that mutates session state or opens a picker the turn would race. `showTranscriptViewer` only calls `tui.showOverlay` and never touches `runControl`, so `'refuse'` would have compiled while silently disabling `/transcript` exactly when reading back is most useful.

Review assisted by AI (Claude Opus 5).

PR #3434, opened 2026-08-21
fix: gate tool execution on raw stream completion, not just JSON validity

canblmz1:
## Problem

Maka deliberately keeps tool execution outside the Vercel AI SDK and settles
returned tool calls through its own `ToolRuntime` after each provider step.

Before this change, the final execution gate relied on the provider step being
classified as `completed`.

However, `settleModelStepOutcome()` also classifies
`finishReason: "length"` as completed.

That means a mutating tool call can have syntactically complete arguments,
reach Maka's returned-tool settlement path, and still belong to a provider
generation that was cut off by a token limit.

A complete tool-call payload is not, by itself, proof that the surrounding
provider step terminated safely.

There is a second integrity boundary as well: the AI SDK's final `tool-call`
input is already parsed/post-processed data. Where providers expose raw
`tool-input-delta` chunks, those raw bytes are stronger evidence of whether the
arguments were actually streamed to structural completion.

## Fix

Add a narrow, Maka-owned tool-execution safety layer around each physical provider request.

For incrementally streamed tool arguments:

- observe the raw AI SDK stream before Maka translates it;
- track `tool-input-start`, `tool-input-delta`, `tool-input-end`, and terminal stream events;
- retain the exact raw argument bytes per tool-call id;
- require matching start/end evidence, valid JSON decoded from those bytes, a matching tool identity, and an explicitly execution-safe terminal reason before the call may reach `ToolRuntime`.

The execution-safety implementation lives directly in
`packages/runtime/src/tool-call-execution-guard.ts`.

For providers that deliver tool arguments atomically and expose no raw argument deltas, no raw-JSON completeness claim is made.

Those calls instead require the provider step itself to finish with an explicitly execution-safe reason:

- `stop`
- `tool-calls`

Other terminal states, including `length`, are not execution-safe.

This intentionally leaves `settleModelStepOutcome()` unchanged because its
`length -> completed` behavior has broader continuation/bookkeeping semantics.
The stricter rule exists only at the irreversible tool-execution boundary.

## Guard-proved execution authority

The guard's verdict is not reduced to a boolean gate, and it is not an
execute/retry/reject action union either -- production never produced
`retry`, and under raw evidence a rejected decision and no decision at all
reached the same backend result, so that state space was collapsed to a
single representation: a **positive proof map**. A call's `toolCallId` is
either present in the map, with the guard's own proved tool name and parsed
value attached, or it is absent -- there is no separate rejected/retry state
to distinguish from "never had raw evidence", because a caller could never
[...]

PR #3487, opened 2026-08-22
fix(runtime): distinguish unchecked writes from T0-missing races in the filesystem worker

### LINKED ISSUE #3484: Release Windows check fails: FilesystemWorkerClient rejects writes to existing files without expectedIdentity (regression from #3001)
`FilesystemWorkerClient.execute()` now rejects a plain write to an existing file when the caller doesn't supply `expectedIdentity`. Callers that go through `FilesystemExecutor` are fine — it captures the identity at lock acquisition and passes it down. Callers that construct the client directly are not, and `scripts/verify-windows-sandbox-e2e.mjs` is one of them, so the **Release Windows check is currently failing on every PR**.

Introduced by #3001 (`ec1579bc`).

### Reproduction

`scripts/verify-windows-sandbox-e2e.mjs:95-97`:

```js
const insidePath = join(workspace, 'inside.txt');
await writeFile(insidePath, 'seeded');
await execute({ kind: 'write', path: insidePath, content: 'packaged-relay-ok' });  // throws
```

```
  code: 'SANDBOX_FILESYSTEM_OPERATION_FAILED'
  domain: 'filesystem'
  reason: 'path_changed'
  stage: 'validation'
  at verifyWindowsSandboxWorkerE2E (scripts/verify-windows-sandbox-e2e.mjs:97:5)
```

There is no race here. The file is created by the line above, single-threaded and awaited. `execute` is `client.execute({ operation, cwd, mode })` — no `expectedIdentity`.

### Cause

`packages/runtime/src/filesystem-worker/client.ts:209-219`:

```ts
const identity =
  input.expectedIdentity && target.targetType !== 'missing'
    ? input.expectedIdentity
    : undefined;
if (!identity && target.targetType !== 'missing' && access === 'write') {
  throw clientError('path_changed', 'validation', requestId, '...');
}
```

`expectedIdentity` is optional (`client.ts:67`) with no internal fallback — it appears only at lines 67, 194, 210 and 211. So the condition reduces to: **any write to an existing target from a caller that didn't supply a T0 identity fails**.

The reconciliation logic itself is right, and the comment at `client.ts:194` correctly explains why the identity must not be re-derived after the lock is held. What's missing is the third case: a caller that has no T0 identity at all isn't a victim of a swap — it just isn't participating in the CAS. Today that's indistinguishable from "T0 missing, T1 existing".

### Scope

- **Not user-facing.** Tool-driven writes reach the worker through `FilesystemExecutor`, which captures identity via `captureIdentityAtLockAcquisition` before the lock wait (`filesystem-executor.ts:270` for write/edit/format_json, `:284` for apply_patch) and forwards it.
- **Direct `FilesystemWorkerClient` consumers are affected.** The known one is the Windows sandbox e2e verifier; worth grepping for others before fixing.
- Ordinary CI is green, which is why this only surfaced on the Windows release job.

### Suggested direction

Distinguish "caller supplied no identity" from "T0 saw nothing". Either make `expectedIdentity` required at the type level so every direct caller has to decide explicitly, or add an opt-out the verifier can pass to say it isn't participating in the CAS. The first is stricter and would have caught this at compile time.

Whoever picks this up: `scripts/verify-windows-sandbox-e2e.mjs` reproduces it directly, so the fix is verifiable without a full Windows packaging run.

---
*Filed after this failure blocked #3472's Release Windows check. Reported by an AI-assisted review process; the reproduction and cause were verified against the source at `ec1579bc`.*

Astro-Han:
## Summary

`FilesystemWorkerExecuteInput.expectedIdentity` was optional. A caller that never captured a T0 snapshot (e.g. `verify-windows-sandbox-e2e.mjs`, which owns every path it writes) was indistinguishable from a caller whose T0 saw a missing target that then appeared while queued — so every write to an existing file without an identity failed `path_changed`. This blocked the Release Windows check on four PRs (#3454 #3434 #3431 #3397) and could break any direct `FilesystemWorkerClient` caller.

The fix makes the caller's T0 observation explicit and required, and carries it on the wire:

- `FilesystemWorkerExpectedIdentity = { dev, ino } | 'missing' | 'unchecked'` (required). Every caller must decide which CAS contract it participates in — an absent field is no longer silently accepted.
- The wire `expectedTarget.identity` is now one required three-state field (protocol v7): `{dev, ino} | 'missing' | 'unchecked'` — illegal combinations are unrepresentable on the wire. The worker fails a write only when `identity === 'missing'` and the target exists at execution time; `'unchecked'` skips CAS; identity objects still compare-and-swap.
- `openStableTarget` opens an existing target without an identity when the caller is `'unchecked'` (no comparison), leaving the create path (`wx`) exclusive and unchanged.
- Callers updated: `FilesystemExecutor` passes its captured identity or `'missing'` for mutations and `'unchecked'` for reads; the runtime-host read-only adapter and the Windows sandbox e2e script pass `'unchecked'`.

The #3001 race protection is not weakened: identity-present writes still CAS against the on-disk inode, and the "T0 missing, T1 existing" race still fails `path_changed`.

Fixes #3484

## Verification

Ran (worktree `fix/3484-filesystem-worker-identity` on `origin/main` f2722fd14):

- `npx tsc -p packages/runtime/tsconfig.json --noEmit` — clean
- `npx tsc -p packages/runtime-host/tsconfig.json --noEmit` — clean
- `npm run test:dist` in `@maka/runtime` — 3073 tests, 3060 pass, 0 fail, 13 skipped (opt-in docker/sandbox suites)
- Focused filesystem suites: `'unchecked'` write to an existing target passes and carries no identity on the wire
  - `'missing'` with a T1-existing target still fails `path_changed` (client and worker paths)
  - CAS mismatch with a stale identity still fails `path_changed`
  - apply_patch forwards its captured identity (executor-level regression for the P0 found in review)
- `node --check scripts/verify-windows-sandbox-e2e.mjs` — syntax clean (full Windows repro needs the Windows sandbox lane; unit-level equivalent covered by the new client test)
- `npm run format:check` — clean across the repo (1591 files); `biome check` on all changed files — clean
- Focused filesystem suites (worker, worker-client, target-identity, mutation-outcome): 65/65 pass — including apply_patch update against a swapped inode failing `path_changed` with the replacement content untouched

Not run: the Windows-packaged release lane (no Windows runner here); the opt-in docker/Linux-sandbox suites.

## AI use

Select exactly one:

Tool(s) and scope: DeepSeek V4 Flash (AstroHan-DSv4F, Pi runtime) — designed and implemented the fix, updated callers and tests. Commit carries the `Generated-by` trailer.

## Checklist

- [ ] No

### DIFF
--- a/packages/runtime-host/src/server/execution-composition.ts
+++ b/packages/runtime-host/src/server/execution-composition.ts
@@ -1631,7 +1631,9 @@ function adaptManagedWorkspaceFilesystemWorker(
 ): ManagedWorkspaceFilesystemWorker {
   return {
     async execute(input) {
-      const result = await worker.execute(input);
+      // Read-only operations never participate in CAS; the adapter says so
+      // explicitly (#3484) instead of relying on an absent optional field.
+      const result = await worker.execute({ ...input, expectedIdentity: 'unchecked' });
       switch (result.kind) {
         case 'read':
         case 'read_image':
--- a/packages/runtime/src/__tests__/filesystem-mutation-outcome.test.ts
+++ b/packages/runtime/src/__tests__/filesystem-mutation-outcome.test.ts
@@ -17,6 +17,7 @@ import {
   type FilesystemWorkerClient,
   type FilesystemWorkerClientErrorReason,
   type FilesystemWorkerExecuteInput,
+  type FilesystemWorkerExpectedIdentity,
 } from '../filesystem-worker/client.js';
 import type { FilesystemWorkerResult } from '../filesystem-worker/protocol.js';
 import { createLocalWorkspaceExecutor } from '../workspace-executor.js';
@@ -216,7 +217,7 @@ describe('filesystem mutation T0 identity capture (queue-window closure)', () =>
       releaseFirst = resolve;
     });
     let calls = 0;
-    let queuedIdentity: { dev: string; ino: string } | undefined;
+    let queuedIdentity: FilesystemWorkerExpectedIdentity | undefined;
     const gatedWorker: {
       execute: (input: FilesystemWorkerExecuteInput) => Promise<FilesystemWorkerResult>;
     } = {
@@ -255,9 +256,12 @@ describe('filesystem mutation T0 identity capture (queue-window closure)', () =>
     releaseFirst();
     await Promise.all([first, second]);
 
-    assert.ok(queuedIdentity, 'the queued mutation should have dispatched to the worker');
+    assert.ok(
+      queuedIdentity && typeof queuedIdentity !== 'string',
+      'the queued mutation should have dispatched with the captured identity',
+    );
     assert.equal(
-      queuedIdentity.ino,
+      (queuedIdentity as { dev: string; ino: string }).ino,
       String(original.ino),
       'identity must be the inode captured at lock acquisition (before the replacement); a T1 capture would sample the replacement',
     );
--- a/packages/runtime/src/__tests__/filesystem-mutation-outcome.test.ts
+++ b/packages/runtime/src/__tests__/filesystem-mutation-outcome.test.ts
@@ -266,6 +266,36 @@ describe('filesystem mutation T0 identity capture (queue-window closure)', () =>
       'identity must be the inode captured at lock acquisition (before the replacement); a T1 capture would sample the replacement',
     );
   });
+
+  test('an apply_patch mutation forwards its captured identity, not unchecked (#3484 regression)', async () => {
+    const cwd = await realpath(await mkdtemp(join(tmpdir(), 'maka-t0-applypatch-')));
+    cleanup.push(cwd);
+    const target = join(cwd, 'file.txt');
+    await writeFile(target, 'original', 'utf8');
+
+    const original = await stat(target, { bigint: true });
+    let dispatched: FilesystemWorkerExpectedIdentity | undefined;
+    const gatedWorker: {
+      execute: (input: FilesystemWorkerExecuteInput) => Promise<FilesystemWorkerResult>;
+    } = {
+      async execute(input) {
+        dispatched = input.expectedIdentity;
+        return { kind: 'apply_patch', ok: true, path: target };
+      },
+    };
+    const fs = executorWith(gatedWorker);
+
+    await fs.applyPatch({
+      operation: { type: 'update_file', path: target, diff: '--- a\n+++ b\n' },
+      cwd,
+    });
+
+    assert.ok(
+      dispatched && typeof dispatched !== 'string',
+      'apply_patch must dispatch with the captured identity, not unchecked',
+    );
+    assert.equal(dispatched.ino, String(original.ino));
+  });
 });
 
 function sleep(ms: number): Promise<void> {
--- a/packages/runtime/src/__tests__/filesystem-target-identity.test.ts
+++ b/packages/runtime/src/__tests__/filesystem-target-identity.test.ts
@@ -155,6 +155,37 @@ describe('filesystem worker target identity CAS', () => {
     assert.equal(await readFile(target, 'utf8'), 'replacement\nold\n');
   });
 
+  test('rejects an apply_patch update when the target inode changed after authorisation', async () => {
+    const cwd = await temporaryDirectory('maka-identity-applypatch-update-');
+    const target = join(cwd, 'file.txt');
+    const replacement = join(cwd, 'replacement.txt');
+    await writeFile(target, 'line\noriginal\n', 'utf8');
+    await writeFile(replacement, 'line\nreplacement\n', 'utf8');
+
+    const identity = await captureIdentity(target);
+    await rename(replacement, target);
+
+    const response = await executeFilesystemWorkerRequest(
+      requestFor(
+        {
+          kind: 'apply_patch',
+          cwd,
+          path: target,
+          action: 'update',
+          diff: '--- a\n+++ b\n@@ -1,2 +1,2 @@\n line\n-original\n+updated\n',
+        },
+        { enforcementPath: target, access: 'write', scope: 'exact', targetType: 'file' },
+        identity,
+      ),
+    );
+
+    assert.equal(response.ok, false);
+    assert.equal(response.error?.code, 'path_changed');
+    // The replacement content must be untouched (the patch never applied).
+    const { readFile } = await import('node:fs/promises');
+    assert.equal(await readFile(target, 'utf8'), 'line\nreplacement\n');
+  });
+
   test('creates a missing target without requiring an identity', async () => {
     const cwd = await temporaryDirectory('maka-identity-missing-');
     const target = join(cwd, 'brand-new.txt');
--- a/packages/runtime/src/__tests__/filesystem-target-identity.test.ts
+++ b/packages/runtime/src/__tests__/filesystem-target-identity.test.ts
@@ -29,7 +29,7 @@ async function temporaryDirectory(prefix: string): Promise<string> {
 
 function requestFor(
   operation: FilesystemWorkerRequest['operation'],
-  expectedTarget: FilesystemWorkerTarget,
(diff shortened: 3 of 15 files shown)


M4n5ter (commented):
<details open>
<summary>English</summary>

Reviewed exact head `89db39fbf03c77ccb6ab5df55b954db05bc77942`.

I found two issues worth addressing before merge.

### [Blocking] Migrate the JavaScript release callers to the required v7 identity contract

`packages/runtime/src/filesystem-worker/protocol.ts:59-83` now requires `expectedTarget.identity`, but two supported JavaScript release callers still construct requests without it:

- `scripts/verify-macos-arm64-dmg.mjs:49-60` constructs a raw v7 request for a missing target without `identity: 'missing'`.
- `scripts/verify-windows-sandbox-e2e.mjs:91-92` forwards an optional `expectedIdentity`. The seeded write supplies its captured inode, but Read, Glob, and Grep at lines 126, 135, and 148 omit it.

For the Windows calls, `client.ts:328-331` consequently places `identity: undefined` in the request object. `JSON.stringify` removes that property, and the packaged worker rejects the request as `invalid_request` under the required v7 schema. The verifier therefore terminates at its first Read. The macOS raw request is rejected for the same reason.

The current green checks do not cover these paths. `windows_sandbox_w0_protocol` runs the TypeScript worker smoke, not `verify-windows-sandbox-e2e.mjs`, and this PR did not trigger the path-filtered Release Windows workflow.

The minimum repair is to add `identity: 'missing'` to the macOS raw request and explicitly supply the correct marker from every Windows call.

The cleaner root model is narrower: CAS identity is a mutation contract. Have the client derive wire `'unchecked'` for read operations through `operationAccess`, while requiring and runtime-validating an explicit T0 state only for mutations. That removes meaningless state from read callers and prevents untyped JavaScript reads from recreating this failure. The Windows mutation probes should still explicitly pass their captured identity or `'missing'`.

Please run the packaged Release Windows verifier after the repair; a small request-shape regression for the macOS raw smoke would also close the currently untested boundary.

### [Important] Preserve the creation diff for worker-backed new-file writes

`packages/runtime/src/filesystem-worker/operations.ts:164-173` still identifies a new file using the old optional-field test:

```ts
if (!expectedTarget?.identity) {
  previous = 'new';
}
```

Under v7, a T0-missing target carries the truthy value `'missing'`. The worker therefore creates the file with a write-only `wx` descriptor, tries to read that descriptor, catches the failure as `previous = 'unknown'`, and omits the diff entirely.

This affects the normal managed Write path. `builtin-tools.ts:416-418` converts the result from a reviewable `file_diff` containing `--- /dev/null` into only a `file_write` descriptor, so users lose the new file's change review.

Use the target fact as the authority—for example, `expectedTarget?.targetType === 'missing'`—and add one worker behavior assertion that a missing-target Write returns a `/dev/null` creation diff.

### Simplify audit

`filesystem-executor.ts:125-130` still maintains a second handwritten write-kind list in `mutates()`, while this PR introduces `operationAccess()` as the shared authority. Line 270 can use `operationAccess(call.operation.kind)` directly and delete `mutates()`. This removes the same class of classification drift without introducing another abstraction.

I did not find any changed tests that should be deleted: the executor forwarding test and worker CAS tests protect distinct stable boundaries.

The three-state wire direction and the fd-pinned CAS paths otherwise look sound, including apply_patch forwarding, inode replacement rejection, exclusive creation, and compare-and-delete.

</details>

<details>
<summary>简体中文</summary>

已审查 exact head `89db39fbf03c77ccb6ab5df55b954db05bc77942`。

我发现两项值得在合并前处理的问题。

### [Blocking] 将 JavaScript release callers 完整迁移到 v7 必填 identity 契约

`packages/runtime/src/filesystem-worker/protocol.ts:59-83` 现在要求 `expectedTarget.identity` 必填，但两个受支持的 JavaScript release caller 仍会构造不含该字段的请求：

- `scripts/verify-macos-arm64-dmg.mjs:49-60` 为缺失目标直接构造 raw v7 request，却没有传入 `identity: 'missing'`。
- `scripts/verify-windows-sandbox-e2e.mjs:91-92` 转发可省略的 `expectedIdentity`。预先创建的 write 传入了真实 inode，但 lines 126、135、148 的 Read、Glob、Grep 都没有传值。

Windows 路径中，`client.ts:328-331` 因此会在 request object 中产生 `identity: undefined`。`JSON.stringify` 会删除该属性，packaged worker 随后按照 v7 required schema 将请求拒绝为 `invalid_request`，verifier 会在第一个 Read 处直接终止。macOS raw request 也会因同一原因被拒绝。

当前绿色 CI 没有覆盖这些路径。`windows_sandbox_w0_protocol` 运行的是 TypeScript worker smoke，不是 `verify-windows-sandbox-e2e.mjs`；本 PR 也没有触发带 path filter 的 Release Windows workflow。

最小修复是为 macOS raw request 增加 `identity: 'missing'`，并让 Windows 的每个调用显式传入正确 marker。

更根因化且更简洁的模型是：CAS identity 只属于 mutation contract。client 应通过 `operationAccess` 为 read operation 自行生成 wire `'unchecked'`，只要求 mutation caller 提供并在运行时校验显式 T0 state。这样可以从 read caller 中删除无意义状态，并避免未类型检查的 JavaScript read 再次制造同类失败。Windows 的 mutation probe 仍应明确传入捕获的 identity 或 `'missing'`。

修复后请运行实际 packaged Release Windows verifier；同时可以用一个很小的 request-shape regression 封住 macOS raw smoke 当前未测试的边界。

### [Important] 保留 worker-backed 新建文件 Write 的 creation diff

`packages/runtime/src/filesystem-worker/operations.ts:164-173` 仍沿用旧的 optional-field 判断来识别新文件：

```ts
if (!expectedTarget?.identity) {
  previous = 'new';
}
```

在 v7 中，T0 不存在的目标会携带 truthy 的 `'missing'`。worker 因此先通过 write-only `wx` descriptor 创建文件，随后尝试读取该 descriptor；读取失败被降级为 `previous = 'unknown'`，最终完全省略 diff。

这会影响正常的 managed Write 路径。`builtin-tools.ts:416-418` 会将本应包含 `--- /dev/null`、可供用户审查的 `file_diff` 降级为只有基本信息的 `file_write`，用户无法再看到新文件的 change review。

应以目标事实作为 authority，例如判断 `expectedTarget?.targetType === 'missing'`，并增加一条 worker 行为断言，确认 missing-target Write 会返回 `/dev/null` creation diff。

### Simplify audit

`filesystem-executor.ts:125-130` 仍在 `mutates()` 中维护第二份手写 write-kind 列表，而本 PR 已引入 `operationAccess()` 作为共享 authority。line 270 可以直接使用 `operationAccess(call.operation.kind)` 并删除 `mutates()`。这能在不新增抽象的情况下消除同类分类漂移。

没有发现应该删除的 changed tests：executor forwarding test 与 worker CAS tests 分别保护不同的稳定边界。

除此之外，三态 wire contract 与 fd-pinned CAS 的整体方向成立；apply_patch identity forwarding、inode replacement rejection、exclusive create 和 compare-and-delete 路径均未发现新的实质问题。

</details>

jackwener (approved):
Independent review of `d9831155c`. Did not treat author self-review as evidence.

1) Problem. Optional `expectedIdentity` made “no T0 snapshot” look like “T0 missing”. Writes then failed `path_changed` on existing files, and `!identity` later flipped when `'missing'` became a truthy string.

2) Solution. One required three-state wire field `{dev,ino} | 'missing' | 'unchecked'`. `operationAccess` is the single write-kind list (includes `apply_patch`). Client rejects a write with no identity at runtime; reads send `'unchecked'` themselves. Worker CAS: object compares, `'missing'` + T1-existing fails `path_changed`, `'unchecked'` skips. `openStableTarget` treats existing+no-object as unchecked open, missing as exclusive `wx`.

3) Occam. One field, not a flag plus an identity. Checked the previous failure modes: `!identity` on the wire is gone; apply_patch is a write; e2e negative cases pass a real T0 so they hit the permission gate, not identity validation.

Approve.

PR #3493, opened 2026-08-22
fix(cli): show model display names in picker

### LINKED ISSUE #3482: TUI /model picker shows raw model ids; Desktop shows display names (inconsistent labels across surfaces)
## Summary

The TUI `/model` picker renders the raw model id as the item label, while Desktop renders the model's display name (falling back to the id only when no display name exists). The same connection catalog therefore shows two different names for the same model depending on which surface you use — e.g. Desktop says `Claude Opus 4.5`, TUI says `claude-opus-4-5` (or whatever id the provider reports).

## Evidence

**Desktop** builds choices with a display-name-first label in `packages/core/src/chat-model-choice.ts:61`:

```ts
label: entry.displayName?.trim() || entry.id,
```

and both Desktop consumers render that field:

- `packages/ui/src/model-picker-internals.ts:41` — `label: choice.label`
- `packages/ui/src/chat-model-switcher.tsx:95` — `label={choice.label}`

Display names are populated from per-connection catalogs and known-model metadata (`packages/core/src/model-metadata.generated.ts`), so most mainstream models have a human-readable name.

**TUI** ignores the display name entirely:

1. The cross-connection `/model` picker labels items with the raw id — `packages/cli/src/pi-tui-pickers.ts:582`:
   ```ts
   return { value: String(index), label: choice.model, description: tags.join(' · ') };
   ```
2. The `ModelChoice` contract has no `displayName` field at all — `packages/cli/src/pi-tui-contracts.ts:6-22`.
3. The projection that feeds it already has full `ModelInfo` records (including optional `displayName`) in hand but only extracts `contextWindow` — `packages/cli/src/runtime-host-onboarding.ts:65-78`:
   ```ts
   const modelsById = new Map(connection.models.map((model) => [model.id, model]));
   ...
   contextWindow: modelsById.get(model)?.contextWindow,
   ```
4. The search matcher can't match on a friendly name either — `packages/cli/src/pi-tui-pickers.ts:592-600`; its own doc comment notes "`ModelChoice` carries no display name ... a display-name enrichment would slot in here".

Note the fallback path (`modelPickerItems`, `pi-tui-pickers.ts:547-564`) operates on plain `string[]` ids with no `ModelInfo` available, so it would keep rendering raw ids unless callers pass richer data.

## Expected

TUI and Desktop show the same label for the same model: prefer `displayName`, fall back to the id, matching the desktop behavior (`entry.displayName?.trim() || entry.id`).

## Suggested fix

1. Add `displayName?: string;` to `ModelChoice` in `packages/cli/src/pi-tui-contracts.ts`.
2. Project it in `projectRuntimeHostModelChoices` (`packages/cli/src/runtime-host-onboarding.ts:75`) via the existing `modelsById` lookup.
3. In `modelChoicePickerItems` (`packages/cli/src/pi-tui-pickers.ts:582`) use `choice.displayName?.trim() || choice.model`, and teach `matchesModelChoice` to match against the display name too.

The fallback `modelPickerItems` may need its callers to supply richer data or accept that minimal hosts still show ids.

rbalachandar:
## Summary

The cross-connection TUI `/model` picker now prefers each catalog model’s human-readable `displayName`, while retaining the raw model id when no display name is available. Search matches both model ids and display names.

Fixes #3482

## Verification

- `npm --workspace maka-agent test` (372 tests passed)
- `npx biome lint packages/cli/src/pi-tui-contracts.ts packages/cli/src/runtime-host-onboarding.ts packages/cli/src/pi-tui-pickers.ts packages/cli/src/__tests__/runtime-host-onboarding.test.ts packages/cli/src/__tests__/pi-tui-runner.test.ts`
- `npx biome format packages/cli/src/pi-tui-contracts.ts packages/cli/src/runtime-host-onboarding.ts packages/cli/src/pi-tui-pickers.ts packages/cli/src/__tests__/runtime-host-onboarding.test.ts packages/cli/src/__tests__/pi-tui-runner.test.ts`

The TUI runner test covers display-name rendering, id fallback, and searching by the displayed model name.

## AI use

Tool(s) and scope: OpenCode made the code and test changes, reviewed the implementation, and ran the verification commands.

## Checklist

- [ ] No

### DIFF
--- a/packages/cli/src/__tests__/pi-tui-runner.test.ts
+++ b/packages/cli/src/__tests__/pi-tui-runner.test.ts
@@ -2912,6 +2912,7 @@ describe('Maka Pi TUI runner', () => {
       { query: 'alpha', keep: 'GPT 5.5 Preview', drop: ['GLM Max', 'text-unicorn'] },
       { query: 'zai', keep: 'GLM Max', drop: ['GPT 5.5 Preview', 'text-unicorn'] },
       { query: 'gemini', keep: 'text-unicorn', drop: ['GPT 5.5 Preview', 'GLM Max'] },
+      { query: 'glm-max', keep: 'GLM Max', drop: ['GPT 5.5 Preview', 'text-unicorn'] },
     ];
     for (const c of cases) {
       terminal.input(c.query);
--- a/packages/cli/src/__tests__/pi-tui-runner.test.ts
+++ b/packages/cli/src/__tests__/pi-tui-runner.test.ts
@@ -2786,7 +2786,7 @@ describe('Maka Pi TUI runner', () => {
           connectionName: 'OpenAI',
           providerType: 'openai',
           model: 'gpt-5.5',
-          displayName: 'GPT Standard',
+          displayName: 'GPT 5.5 Preview',
           isDefaultConnection: true,
         },
         {
@@ -2859,7 +2859,7 @@ describe('Maka Pi TUI runner', () => {
           connectionName: 'Aurora',
           providerType: 'openai',
           model: 'gpt-5.5',
-          displayName: 'GPT Standard',
+          displayName: 'GPT 5.5 Preview',
           isDefaultConnection: true,
         },
         {
@@ -2889,7 +2889,7 @@ describe('Maka Pi TUI runner', () => {
     await waitFor(() => {
       const out = plainTerminalOutput(terminal.screenOutput());
       return (
-        out.includes('GPT Standard') && out.includes('GLM Max') && out.includes('text-unicorn')
+        out.includes('GPT 5.5 Preview') && out.includes('GLM Max') && out.includes('text-unicorn')
       );
     });
     assert.doesNotMatch(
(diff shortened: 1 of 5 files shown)


me2seeks:
Nice cleanup — the projection via `modelsById`, the `displayName?.trim() || model` label rule and the stale-comment updates all look right to me.

One substantive nit before merge: **the model-id match criterion lost its dedicated test case**, so one line of `matchesModelChoice` is no longer pinned by any test.

The five-case list previously isolated each match criterion one-by-one; `query: 'gpt'` was the case that exercised **matching by model id**:

```ts
- { query: 'gpt', keep: 'gpt-5.5', drop: ['glm-max', 'text-unicorn'] },
+ { query: 'preview', keep: 'GPT 5.5 Preview', drop: ['GLM Max', 'text-unicorn'] },
```

After the swap, every remaining query is satisfied by other lines:

```ts
function matchesModelChoice(choice: ModelChoice, query: string): boolean {
  if (choice.model.toLowerCase().includes(query)) return true;          // ← now unpinned
  if (choice.displayName?.toLowerCase().includes(query)) return true;   // 'preview' hits here
  if (choice.connectionName.toLowerCase().includes(query)) return true; // 'aurora'
  if (choice.connectionSlug.toLowerCase().includes(query)) return true; // 'alpha'
  if (choice.providerType.toLowerCase().includes(query)) return true;   // 'zai'
  const providerLabel = PROVIDER_DEFAULTS[choice.providerType]?.label;  // 'gemini'
  ...
}
```

Deleting the `choice.model` line would keep this suite green. That matters because models without catalog metadata (minimal hosts, custom relay ids) are exactly the ones users must still be able to find by raw id.

Suggested fix — add an id-only query whose substring does not occur in any display name:

```ts
// 'glm-max' occurs in the model id but not in the display name ('GLM Max'),
// so this case can only pass through the model-id criterion.
{ query: 'glm-max', keep: 'GLM Max', drop: ['GPT 5.5 Preview', 'text-unicorn'] },
```

(Optionally also assert the inverse rendering invariant somewhere — that a choice with a `displayName` no longer shows its raw id — but the implicit fallback via `text-unicorn` already covers the happy path.)

Astro-Han (approved):
Approving `141534485a0addc6a2747d3b6bc7cb55e428916c`. Required `test` is `completed / success` bound to that exact SHA. **No P0–P3.**

One note on how this got here: CI had **never run** on this head — zero check runs, one workflow sitting at `action_required`. That state renders like "in progress" and is easy to misread as pending; it was approved as a maintainer and is green now. Flagging it because it was the reviewing line that caught it rather than assuming the absence of a red mark meant a green one.

What was verified:

- **`displayName` is a projection, never an identity.** The picker still carries `value: String(index)` and resolves back to the `ModelChoice`; `model` remains the authoritative key and `displayName` only reaches `label`. Nothing selects, persists, or matches on the display name.
- **One source, one lookup.** `modelsById.get(model)?.displayName` reads from the same catalog object as `model`, in the same lookup — not a second fetch that could disagree with it. Two call sites, no third copy.
- **The fallback is `displayName?.trim() || choice.model`.** The `.trim()` matters: a catalog entry with a whitespace-only display name falls back to the id rather than rendering a blank row. `??` would not have.

**The part worth calling out is the one that is not cosmetic.** This also widens search matching — `if (choice.displayName?.toLowerCase().includes(query)) return true;` — and correspondingly updates the comment that used to say `ModelChoice` carries no display name so the id is the only model-side match target. **This is the one place where display names cross out of presentation and into behavior**, so it deserves to be described as such rather than folded into "show display names."

Good that `{ query: 'glm-max', keep: 'GLM Max' }` was added alongside the display-name queries: it pins that widening the match surface did not displace plain id matching, which is the regression this change could plausibly have introduced.

*Reviewed at 2026-08-23 12:55 UTC.*

Astro-Han:
LGTM.

PR #3499, opened 2026-08-22
fix(cli): label /model choices by catalog display name

### LINKED ISSUE #3482: TUI /model picker shows raw model ids; Desktop shows display names (inconsistent labels across surfaces)

me2seeks:
Fixes #3482

## Problem

The TUI `/model` picker rendered raw model ids as item labels, while Desktop labels the same underlying choices by catalog display name (`entry.displayName?.trim() || entry.id` in `chat-model-choice.ts`). The same model showed as `Claude Opus 4.5` on Desktop but as its raw id in the TUI, and the TUI search field could not match by the friendly name.

## Change

- `pi-tui-contracts.ts` — add optional `displayName` to `ModelChoice`, documented with the fallback contract.
- `runtime-host-onboarding.ts` — project it from the connection catalog via the existing `modelsById` lookup (the same lookup that already resolves `contextWindow`; no new data source or protocol change).
- `pi-tui-pickers.ts` — items render `displayName?.trim() || model` (desktop's rule), and `matchesModelChoice` matches queries against the display name; stale doc comments updated.

The ids-only `modelPickerItems` fallback for minimal hosts is intentionally unchanged, as scoped in the issue.

## Testing

- `packages/cli`: typecheck clean; biome lint/format clean on all touched files.
- `node --test dist/__tests__/runtime-host-onboarding.test.js` — 3/3 pass, incl. new projection test.
- `node --test dist/__tests__/pi-tui-runner.test.js` — 120/120 pass, incl. new integration test asserting the picker shows the display name (raw id absent when one exists), falls back to the id otherwise, and search-by-friendly-name isolates the choice.

## Note / follow-up candidate

The one-expression label rule now exists on both surfaces (core for Desktop, cli for TUI) with cross-referencing comments. Folding the cli projection into a shared core type is a separate consolidation candidate and is deliberately not mixed into this bugfix.

### DIFF
--- a/packages/cli/src/__tests__/pi-tui-runner.test.ts
+++ b/packages/cli/src/__tests__/pi-tui-runner.test.ts
@@ -2839,6 +2839,66 @@ describe('Maka Pi TUI runner', () => {
     ]);
   });
 
+  test('/model labels choices by catalog display name and searches it', async () => {
+    const terminal = new FakeTerminal();
+    const driver = new SlashCommandDriver();
+    const run = runMakaPiTui({
+      title: 'Maka',
+      driver,
+      cwd: '/repo',
+      // Same trick as the search test above: a current model absent from the
+      // choices, so raw ids can only reach the screen through picker labels.
+      model: 'legacy-curated-out',
+      connectionSlug: 'ghost',
+      providerType: 'openai',
+      modelChoices: [
+        {
+          connectionSlug: 'alpha',
+          connectionName: 'Aurora',
+          providerType: 'openai',
+          model: 'gpt-5.5',
+          displayName: 'GPT-5.5 Mini',
+          isDefaultConnection: true,
+        },
+        {
+          connectionSlug: 'beta',
+          connectionName: 'Boreal',
+          providerType: 'zai',
+          model: 'glm-max',
+          isDefaultConnection: false,
+        },
+      ],
+      permissionMode: 'ask',
+      terminal,
+    });
+
+    await waitForTuiPaint(terminal);
+    terminal.input('/model');
+    terminal.input('\r');
+    await waitFor(() => terminal.output().includes('Select Model'));
+    // The labeled choice renders its display name instead of the raw id; the
+    // choice without one falls back to the id like desktop does.
+    await waitFor(() => {
+      const out = plainTerminalOutput(terminal.screenOutput());
+      return out.includes('GPT-5.5 Mini') && !out.includes('gpt-5.5') && out.includes('glm-max');
+    });
+
+    // Searching by the friendly name finds the model (#3482).
+    terminal.input('mini');
+    await waitFor(() => {
+      const out = plainTerminalOutput(terminal.screenOutput());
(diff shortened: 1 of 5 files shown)


me2seeks:
Closing — superseded by another fix for #3482. Thanks!

Issue #415, opened 2026-07-01
chore: introduce linter/formatter (Biome or ESLint)


luojiyin1987:
## Context

The project has 14k+ AST nodes across 838 files with no automated code-style enforcement. Root `devDependencies` contains only `typescript` + `@types/node`. No ESLint, Prettier, or Biome is configured. All formatting and linting is manual.

## Problem

- No consistent code style across contributors
- No automated enforcement of conventions (e.g. no `console.log` in production, consistent import ordering)
- PR reviews waste time on style nits
- CI pipeline (planned in #414) has no `lint` gate to plug into

## Proposal

Choose and configure one of:

### Option A: Biome (recommended)
- Single binary, fast, covers both lint + format
- Built-in TypeScript/React rules
- Migration tool for ESLint configs

### Option B: ESLint + Prettier
- Industry standard, larger plugin ecosystem
- More configuration overhead, slower

### Implementation steps
1. Team decides on tool (Biome vs ESLint)
2. Add to root `devDependencies`
3. Configure ruleset (start with recommended, tune iteratively)
4. Add `lint` and `format` scripts to root `package.json`
5. Auto-fix existing violations in a dedicated PR
6. Add `lint` gate to CI (#414)

## Acceptance criteria

- [ ] Tool selected and configured
- [ ] `npm run lint` checks all workspaces
- [ ] `npm run format` auto-formats all workspaces  
- [ ] Root `package.json` has `lint` and `format` scripts
- [ ] CI (#414) includes `lint` gate

## References

- #414 — CI pipeline depends on this for the lint gate

UncertaintyDeterminesYou4ndMe:
Claiming this. Plan per the issue's own sequencing: this first PR introduces Biome (Option A) with root `lint`/`format` scripts and a CI lint gate, with the ruleset tuned so the current tree passes (relaxed rules documented for follow-up); the full-repo auto-fix lands in a dedicated follow-up PR to keep diffs reviewable and avoid conflicting with the ~11 open PRs.

Astro-Han:
Done: #1019 and #1122 brought in Biome with a CI gate. biome.jsonc, the lint/format scripts, and the CI check are all live.

PR #3501, opened 2026-08-22
chore: record the ASF header sweep in .git-blame-ignore-revs

Astro-Han:
## Summary

`.git-blame-ignore-revs` documents its own convention: a reformat PR adds its landed hash here in a follow-up right after the squash, because an entry naming a hash the clone has never seen makes `git blame` error out with `cannot find revision ... to ignore` rather than degrading. #3397 left the placeholder in place, as its commit message said it would. This fills it in with `c633ea5b2524d9463228e01f8d369ef9166c3f05`, the squash commit that landed on `main`.

Without the entry, all 2679 files the header sweep touched blame to that commit, even though it changed no line of their content.

Refs #3397

## Verification

- `c633ea5b2524d9463228e01f8d369ef9166c3f05` confirmed present on `main` (`git log -1` resolves it), which is the precondition the file itself states.
- `node scripts/asf-license-headers.mjs check` on this branch: 2681 covered, 114 excluded, clean. Unrelated to this change, but it is the gate #3397 introduced and this branch is the first thing on top of it.
- No test covers this: the file is consumed by `git blame` and by GitHub's blame view, not by anything in the build. Nothing to run beyond confirming the hash resolves.

## Review focus

One thing worth a maintainer's call, not addressed here: the `# style: apply Biome format baseline outside the UI surface (#415)` entry above still carries its own unfilled placeholder, so that sweep is not actually being ignored either. Filling it needs the landed hash for #415, which is a separate lookup and a separate intent — happy to do it in another PR if you want it.

## AI use

Select exactly one:

Tool(s) and scope: Claude Code — wrote the change and this description. The commit carries a `Generated-by: Claude Code` trailer.

## Checklist

- [x] No

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -12,5 +12,5 @@
 # style: apply Biome format baseline outside the UI surface (#415)
 # -> squash-commit hash added in a follow-up immediately after this PR lands.
 
-# chore: apply the ASF license header to every covered source file (#3397)
-# -> squash-commit hash added in a follow-up immediately after this PR lands.
+# chore: add ASF source headers and a release header audit gate (#3397)
+c633ea5b2524d9463228e01f8d369ef9166c3f05


jackwener (approved):
Authorized by @WAWQAQ in Raft task #11 after independent review at exact head d3cf64b67bd4531586bbc31a04ecdb9247ebaae1. The referenced #3397 squash exists on main, the ignore-revs entry is correctly scoped, and required test is terminal green.

jackwener (approved):
Approved on exact head `d3cf64b67bd4531586bbc31a04ecdb9247ebaae1`.

The filled hash `c633ea5b2524d9463228e01f8d369ef9166c3f05` is the #3397 squash on main. One-file follow-up matches `.git-blame-ignore-revs` convention.

Authorization: @WAWQAQ in Raft #PR-Review-Kabi:778d6db7.

PR #3495, opened 2026-08-22
fix(desktop): recover from main Renderer loss

M4n5ter:
## Summary

<details open>
<summary>English</summary>

Detect unexpected termination of the main BrowserWindow Renderer while the Electron main process remains alive.

- Ignore clean exits and app shutdown.
- Capture the bounded reason and exit code in existing main-process diagnostics.
- Present a native recovery dialog where Copy Diagnostics remains auxiliary to Relaunch or Exit.
- Keep the report Desktop-only instead of inferring Runtime Host authority.

</details>

<details>
<summary>中文</summary>

当 Electron 主进程仍存活、主 BrowserWindow 的 Renderer 意外终止时提供可恢复的诊断入口。

- 忽略正常退出和应用关闭流程。
- 将限长后的退出原因与退出码记录到既有主进程诊断中。
- 通过原生恢复弹窗提供复制诊断信息，同时保留重新启动或退出决策。
- 报告仅归因于 Desktop，不推断 Runtime Host。

</details>

Fixes #3490
Refs #3463

## Verification

<details open>
<summary>English</summary>

- `npm --workspace @maka/desktop test` — 1152 passed
- `npm run build`
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `npx knip --workspace apps/desktop`

Focused tests cover the actual `render-process-gone` observation boundary, shutdown filtering, rejected window loads, and the Copy Diagnostics → Relaunch decision path.

</details>

<details>
<summary>中文</summary>

上述检查均通过；聚焦测试覆盖真实的 `render-process-gone` 事件边界、关闭流程过滤、窗口加载失败，以及“复制诊断信息后仍可重新启动”的决策路径。

</details>

## Review focus

<details open>
<summary>English</summary>

Please verify that the lifecycle filter reports only unexpected loss of the main Renderer and that the native report remains within the existing Desktop diagnostic authority.

</details>

<details>
<summary>中文</summary>

请重点确认生命周期过滤仅报告主 Renderer 的意外终止，并且原生报告仍遵循既有 Desktop 诊断权限边界。

</details>

## AI use

Select exactly one:

- [ ] No generative tool made a substantive contribution
- [x] Generative tooling made a substantive contribution

Tool(s) and scope: Codex implemented and validated the change under human direction. The commit includes the required `Generated-by: Codex` trailer.

## Checklist

- [x] Tests cover the change and fail without it
- [x] Lint, format, typecheck and the affected suites pass locally

Does this PR entail a change in behavior?

[...]

PR #3504, opened 2026-08-22
chore: add the ASF header to the renderer-loss recovery files

Astro-Han:
## Summary

#3495 branched before #3397 landed, so its two new files arrived without the ASF header, and its CI run predates the `audit` job that #3397 added. `check:asf-headers` runs in `ci.yml`, so `main` is currently failing that check and every open PR inherits the failure until this lands.

`npm run write:asf-headers` produced this commit. Nothing else was touched.

Refs #3495, #3397

## Verification

- `node scripts/asf-license-headers.mjs check` on `main` before this change: 2 files missing (`apps/desktop/src/main/main-renderer-process-gone.ts` and its test).
- Same command on this branch: clean.
- No behavior to test: both files gain a leading license comment and nothing else.

## AI use

Select exactly one:

Tool(s) and scope: Claude Code — ran the header writer and wrote this description. The commit carries a `Generated-by: Claude Code` trailer.

## Checklist

- [x] No

### DIFF
--- a/apps/desktop/src/main/__tests__/main-renderer-process-gone.test.ts
+++ b/apps/desktop/src/main/__tests__/main-renderer-process-gone.test.ts
@@ -1,3 +1,22 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


jackwener (approved):
Two files from #3495 were missing the ASF header after the #3397 gate landed. This is exactly those headers, nothing else. Approve.

PR #3509, opened 2026-08-22
fix(desktop): clarify and filter status summaries

hqhq1025:
## Summary

The Permission Center previously rendered a verified `Granted` / `已授权` state with the same neutral tone as unknown or unsupported permissions. Its status counts, and the equivalent counts in Health Center, also looked interactive but were static text.

This change:

- maps verified permission grants to the existing success semantic;
- turns the Permission Center counts into real status filters;
- applies the same shared filter component to Health Center;
- lets users press an active filter again to restore the full list;
- disables zero-count filters and exposes pressed/filter state to assistive technology;
- resets a selected filter if refreshed data no longer contains that status.

## Verification

- `npm --workspace @maka/desktop run build-storybook -- --quiet`
- `npx biome check apps/desktop/src/renderer/settings/settings-status-summary-filter.tsx apps/desktop/src/renderer/settings/permission-center-page.tsx apps/desktop/src/renderer/settings/health-center-page.tsx apps/desktop/src/renderer/locales/permission-center-copy.ts apps/desktop/src/renderer/locales/settings-health-copy.ts apps/desktop/src/renderer/styles/settings/health.css apps/desktop/stories/settings/settings-pages.stories.tsx`
- Playwright interaction and responsive visual checks for both pages in dark mode with the Chinese locale

## Screenshots

Permission Center, narrow viewport:

Permission Center status filters

Health Center filtered to errors:

Health Center error filter

## AI use

Select exactly one:

Tool(s) and scope: OpenAI Codex investigated the misleading status summaries, implemented the shared filtering behavior and accessibility states, added Storybook interaction coverage, ran verification, and prepared this PR under human direction. The commits include the required `Generated-by: Codex` trailer where applicable.

## Checklist

- [x] Lint, format, typecheck, builds, and affected interaction checks pass locally

- [ ] No

### DIFF
--- a/apps/desktop/src/main/__tests__/health-center-copy.test.ts
+++ b/apps/desktop/src/main/__tests__/health-center-copy.test.ts
@@ -1,0 +1,33 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import assert from 'node:assert/strict';
+import { test } from 'node:test';
+import { getHealthCenterCopy } from '../../renderer/locales/settings-health-copy.js';
+
+test('labels blocker counts as global across filtered health views', () => {
+  assert.equal(
+    getHealthCenterCopy('zh').blockers.send(1, 6),
+    '全部健康信号中，1/6 条会阻塞发送',
+  );
+  assert.equal(
+    getHealthCenterCopy('en').blockers.send(1, 6),
+    'Across all health signals, 1 of 6 blocks sending',
+  );
+});
--- a/apps/desktop/src/main/__tests__/permission-center-copy.test.ts
+++ b/apps/desktop/src/main/__tests__/permission-center-copy.test.ts
@@ -1,0 +1,27 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import assert from 'node:assert/strict';
+import { test } from 'node:test';
+import { getPermissionCenterCopy } from '../../renderer/locales/permission-center-copy.js';
+
+test('presents a granted OS permission as a verified success', () => {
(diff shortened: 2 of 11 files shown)


Astro-Han (commented):
Thanks — turning a static count into the control that filters to exactly those items is the right instinct, and correcting Granted from neutral to success closes a real gap: "we checked and it is fine" and "we have not checked" were rendering identically. Reusing one `SettingsStatusSummaryFilter` across both the Permission and Health pages, rather than writing the interaction twice, is what keeps this at 227 lines.

Reviewed at exact head `849afcc0324ef626f70193c2d3108c718e8bf8d3` against base `1ddbd2591ccc6fd06d9727a216e6a5b64d8b1e40`. **One P3, nothing higher.** The head has a completed `test` check but the aggregate commit status is `pending`, so this is a COMMENT, not an approval.

**[P3, ordinary path] On the Health page, filtering narrows the list but not the blocker banner.**

`health-center-page.tsx:114-117` derives `visibleSignals` from `signalFilter` and renders the list from it. Two lines below, `blocksSendCount` and `blocksCapabilityCount` (`:118-119`) are computed from the unfiltered `snapshot.signals`, and the banner at `:161-173` renders from those.

So: filter to `ok`, and the list shows only healthy signals while the banner above it still reads "2 health signals block sending". Nothing displayed is false — the banner is a true statement about the system — but the two regions are now describing different sets on the same screen, and the page gives no cue which is which.

**There is a good argument for leaving it exactly as is**, which is why this is a P3 and not higher: a blocker banner is arguably a global alert, and having a view filter hide the fact that you are blocked would be worse than the inconsistency. If that is the intent, the fix is wording rather than logic — something that marks the count as spanning all signals regardless of the current filter would remove the ambiguity without weakening the alert. Your call; we would just rather it be a decision than an artifact.

**Checked and found sound:** the filter is view-only — it adds no path to granting a permission, mutating health state, or changing runtime status. Zero-count buttons are disabled, selection is exposed via `aria-pressed`, re-selecting restores the full list, and the effect clears a filter whose status no longer exists in a refreshed snapshot, so a stale filter cannot survive into a new page state. On the Permission page the counts and `permissionIdsForFilter` are both derived from the same `OS_PERMISSION_IDS` set and the same status mapping, so no divergence there. `buildHealthSnapshot` (`packages/core/src/health.ts:82-94`) recomputes the summary from the same signals, so the snapshot itself is internally consistent — the P3 above is strictly about the filtered view, not the data.

**How this was reviewed, including a disagreement we had to settle.** Two reviewers worked this independently without seeing each other's conclusion. One reported the banner inconsistency; the other reported no findings, having checked summary-versus-signals consistency in `core` and found none. Both were right about what they actually examined — those are two different questions, and only one of them is about the filtered view. We read the component ourselves to settle it, and the P3 stands.

`@maka/core` and `@maka/ui` build; core health tests 6/6 and the added permission copy test 1/1. The desktop renderer build could not start in our environment (missing `@vitejs/plugin-react`) and desktop typecheck is blocked on unbuilt workspace exports — both pre-existing environment limits, not attributable to this change, and we are reporting them as not run rather than as results.

---

Astro-Han (approved):
Thanks — this resolves the point from our previous comment, and it resolves it the better of the two ways we suggested.

Re-reviewed at exact head `d54a087da11f6d9e16048e017daa11e927135edf`. Our earlier review was against `849afcc0`, so that conclusion no longer applied and this is a fresh pass over the current code.

**The blocker-count ambiguity is closed.** Rather than making the banner follow the filter — which would have let a view filter hide the fact that sending is blocked — you kept it global and said so in the text: `Across all health signals, N of M block sending`, with the matching Chinese copy and a `health-center-copy.test.ts` case pinning both. The count now states its own scope, so a filtered list sitting beneath it no longer reads as a contradiction. That is the right trade: the alert keeps its reach and the confusion goes away.

The rest of the filtering behaviour is unchanged from our previous pass and still checks out — view-only, no new path to granting a permission or mutating health state, zero-count buttons disabled, `aria-pressed` on selection, and a stale filter cleared when a refreshed snapshot no longer contains that status.

---

Astro-Han:
Hi — this PR **conflicts with current `main`**, but it is worth noting that the conflict is narrower than usual.

I tested a rebase onto current `main` locally (throwaway worktree — your branch was not touched). The **only** conflicting file is:

- `docs/astryx-surface-file-inventory.md`

That file is generated, so there is nothing to merge by hand: rebase onto current `main`, regenerate it, and push.

```
# regenerate docs/astryx-surface-file-inventory.md, then
git add docs/astryx-surface-file-inventory.md && git rebase --continue
git push --force-with-lease
```

Your review approval is still bound to the current head, so once the branch is conflict-free and CI is green on the new head this should be ready to go.

---

PR #3510, opened 2026-08-22
fix(storybook): fill the app shell viewport

hqhq1025:
## Summary

- remove the inline percentage height that overrides the production `100dvh` app-frame rule
- let AppShell stories fill the Storybook iframe instead of ending at the 640px minimum and leaving a white band

## Verification

- visually checked the default AppShell story at 1200x768; the shell now fills the full viewport with no bottom white band

### DIFF
--- a/apps/desktop/stories/app-shell.stories.tsx
+++ b/apps/desktop/stories/app-shell.stories.tsx
@@ -280,7 +280,6 @@ function ShellFrame(props: {
       data-sidebar-state={props.sidebarCollapsed ? 'collapsed' : 'expanded'}
       style={
         {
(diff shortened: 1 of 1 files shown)


hqhq1025:
## Visual comparison

Same 1200 × 768 viewport and application state for both captures.

### Before

clipboard

### After

clipboard

jackwener (approved):
Story layout is `fullscreen`; inner shell already uses `height="fill"`. `height: 100%` on the frame is a percentage of an indefinite parent, so the story does not fill the canvas. Leaving `minHeight: 640` is the right floor. Approve.

PR #3511, opened 2026-08-22
feat(desktop): expose Computer Use activity progress

hqhq1025:
## Summary

- preserve trusted Computer Use activity semantics across the Client Capability bridge
- stream bounded multi-step progress from Desktop through Runtime Host into the live turn projection
- replace generic Computer Use rows and running text with action, target, and step-aware labels
- add a Storybook state covering a running 7/11 element sequence

## Verification

- `npm --workspace @maka/runtime-host test` (1071 passed)
- targeted runtime tests for Computer Use and MCP projection (89 passed)
- `npm --workspace @maka/ui test` (204 passed)
- Desktop native capability tests (13 passed)
- visually checked the Computer Use observability Storybook story at 1200x768

### DIFF
--- a/apps/desktop/src/main/__tests__/runtime-host-native-capabilities.test.ts
+++ b/apps/desktop/src/main/__tests__/runtime-host-native-capabilities.test.ts
@@ -48,6 +48,7 @@ test('publishes self-described session-affine Browser and Computer Use offers', 
       affinity: offer.affinity,
       toolNames: offer.tools.map((descriptor) => descriptor.name),
       serverIds: offer.tools.map((descriptor) => descriptor.serverId),
+      activityKinds: offer.tools.map((descriptor) => descriptor.activityKind),
     })),
     [
       {
@@ -56,6 +57,7 @@ test('publishes self-described session-affine Browser and Computer Use offers', 
         affinity: 'session',
         toolNames: ['browser_snapshot'],
         serverIds: ['desktop_browser'],
+        activityKinds: [undefined],
       },
       {
         offerId: 'desktop_computer_use',
@@ -63,6 +65,7 @@ test('publishes self-described session-affine Browser and Computer Use offers', 
         affinity: 'session',
         toolNames: ['maka_computer'],
         serverIds: ['desktop_computer_use'],
+        activityKinds: ['computer'],
       },
     ],
   );
@@ -566,6 +569,7 @@ function computerTools(
     ? [
         {
           ...tool('maka_computer', z.object({ wait: z.boolean().optional() }), impl),
+          activityKind: 'computer' as const,
           toModelOutput: ({ output }: { output: unknown }) => {
             const result = output as {
               text: string;
--- a/apps/desktop/src/main/runtime-host-native-capabilities.ts
+++ b/apps/desktop/src/main/runtime-host-native-capabilities.ts
@@ -349,6 +349,7 @@ async function invokeNativeTool(
     toolCallId: frame.toolCallId,
     abortSignal: signal,
     emitOutput() {},
+    ...(options.progress ? { emitProgress: options.progress } : {}),
   });
   return projectToolResult(binding.tool, frame.toolCallId, args, output);
 }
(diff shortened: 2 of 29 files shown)


hqhq1025:
## Visual comparison

Same scenario and viewport for both captures.

### Before

clipboard

### After

clipboard

Astro-Han (commented):
Thanks — the structural choice here is the right one and it is worth stating plainly, because it is the thing a 706-line diff across five packages usually gets wrong. The `tool_progress` transport already existed; this change adds the producer at the tool layer and the consumer at the projection layer and leaves the seam between them untouched. That is connecting something that was already there, not laying a parallel path, and the five-package spread follows produce → bridge → render rather than shuttling one piece of state between owners.

Reviewed at exact head `ea66f348933ea80b0bd297641ce99251d48cc7eb` against base `1ddbd2591ccc6fd06d9727a216e6a5b64d8b1e40`. **One P3, no P0, P1 or P2.** The head has one completed check (`test: success`) but the aggregate commit status is `pending`, so this is not terminal green and this is a COMMENT rather than an approval.

**[P3, reachable on the ordinary path] The progress contract is a string format shared across packages.** The runtime writes `chunk: \`steps:${current}/${total}\`` and the projection parses it with `/^steps:(\d+)\/(\d+)$/`, returning `undefined` when it does not match. Both ends validate, and the failure mode is benign — no progress renders, nothing breaks — but that is also what makes it easy to miss: if either side's format drifts, the feature silently stops working and no test on the other side fails. A shared parser in `core`, or a dedicated numeric field on the event instead of a formatted string, would make the two ends fail together rather than apart. Not blocking.

**Things we went looking for and did not find:**

- **Lost initial progress.** We expected the `0/N` frame to be emitted before the capability was accepted and dropped. It is not — the desktop path awaits `accept()` before the tool implementation runs, so the first frame is forwarded.
- **A widened exposure surface.** `activityKind` is derived only from trusted client-capability descriptors; an MCP server's arbitrary descriptor still resolves to `'tool'`. The step data the UI reads was already present in durable args because executing a sequence requires it, and the live wire does not carry args.
- **Sloppy frame handling.** The broker enforces monotonicity, a fixed total, and de-duplication of repeated values; decode bounds `total` and `current`. The `1024` ceiling is unreachable given the schema's 12-step maximum, which makes it defensive rather than dead.

**How this was reviewed.** Two reviewers worked this independently with no visibility into each other's conclusion, and **both arrived at the same single P3 and the same absence of anything higher** — the string contract was the one item that survived on both sides. Between them, five other candidate findings were raised and then withdrawn under adversarial scrutiny, including the accept-ordering concern above. Targeted suites were executed against a worktree of this head: the UI projection and tool-activity presentation suites pass 40/40. The runtime-host protocol suite was not executed locally — its bounds and monotonicity logic was read rather than run, and we would rather say so than imply otherwise.

---

Astro-Han (approved):
Thanks — you fixed the cause rather than the symptom, which is the harder and more durable choice here.

Re-reviewed at exact head `b70fa72d6148bd6d9d140dc6c2924164d7c9515e`. Our previous comment was against `ea66f348`, so it no longer bound; this is a fresh pass.

**The split progress contract is closed at the root.** Our concern was that `steps:N/M` existed as a format written in one package and parsed by a regex in another, so either side could drift and the failure would be silent. `fix(runtime): share tool progress codec` collapses that into a single owner: `ToolStepProgress`, `encodeToolStepProgress` and `decodeToolStepProgress` now live in `packages/core/src/events.ts`, with the runtime encoding through it and the UI decoding through it. We searched for a surviving second parser and the pattern now appears exactly once in the tree.

That is the difference between "the two ends currently agree" and "the two ends cannot disagree." Keeping backwards compatibility with the existing wire format while doing it means no coordinated deploy is required.

Also checked: an invalid or late progress event does not wipe an already-rendered value — the projection keeps the previous base when decoding yields nothing, and there is now a regression covering it. The broker's accepted/monotonic/total-stable/de-duplication guarantees are unchanged, and `accept()` still completes before the tool implementation runs, so the initial `0/N` is not dropped.

One note on our own verification rather than on your code: the runtime suite would not start in our environment because of unbuilt workspace exports and a pre-existing `HttpsProxyAgent` typing issue, so `@maka/core` (4/4) and the UI projection and tool-activity suites (41/41) are what we actually executed. We would rather say which ones ran than imply all of them did.

---

PR #3512, opened 2026-08-22
fix(runtime-host): keep one candidate in flight

liugddx:
## Summary

Prevents one Runtime Host election from spawning another Candidate while its current Candidate process is still alive.

#3454 classified the intermittent Windows startup failure in #3279: one packaged automatic-update startup made 69 distinct Candidate launches while observing `not_registered` 73 times, then connected once and timed out with the Host still `recovering`. Repeated process creation under load works against election convergence rather than helping it.

This PR makes Candidate process settlement explicit for both detached and owned launchers, and keeps a per-election in-flight gate closed until that exact process exits. A spawn rejection still permits a later attempt because no process exists. A process whose exit cannot be observed remains fail-closed for that election.

Refs #3279. This PR should land before #3454; the diagnostic PR can then rebase without combining behavior and evidence changes.

## Boundary

- The production 45-second election deadline is unchanged.
- Failure classification, endpoint handling, readiness polling, startup-failure selection, and cleanup ownership are unchanged.
- This is per-election single-flight, not global or cross-client launch coalescing.
- A sequential successor is allowed only after the prior Candidate exits.
- Full LaunchLease, cross-client coordination, retry-policy redesign, and #3454 diagnostic counters remain deferred to their existing #2142/#3454 scopes.

## Mechanism

Both `launchDetachedRuntimeHostCandidate` and `launchOwnedRuntimeHostCandidate` now return the same child-bound `exited` promise. `connectOrSpawnRuntimeHostWithDependencies` marks a Candidate in flight only after `spawned` resolves and clears the gate only when `exited` fulfills. A rejected spawn never sets the gate; a missing or rejected exit signal leaves it closed.

`Release Windows check` now owns changes to the two Candidate-election source files so this behavior cannot merge on Node-only evidence.

## Verification

L1/L2 on exact head `814d95f663c74fb6bd804de5eed3ae37d2eae482`:

- `npm run build` passed
- `npm run typecheck` passed across all workspaces
- `npm run lint`: 2,551 files, 0 errors
- `npm run format:check`: 1,590 files, 0 changes
- CI planner + Windows harness: 60 passed, 0 failed, 1 privilege-dependent symlink skip
- Windows inventory: 3 passed; 63 declarations current
- single-flight/live/exit/spawn-reject/detached settlement: 4 passed
- Candidate barrier + deadline contracts: 8 passed
- real owned exit-to-successor paths: 2 passed
- real Electron Client: exactly one Candidate PID, 1 passed
- independent author-gate review: no unresolved P0-P2 findings
- `git diff --check` passed

Real local fresh-root evidence:

- one detached Candidate launch
- connection succeeded
- Candidate exited with code 0 and no signal
- exact PID exited and the temporary root was removed

The real owned-successor test uses a 15-second election budget and 25-second test timeout only to tolerate execution-composition startup under local load. Production remains at 45 seconds.

Local limitations:

- a serial Runtime Host run reached at least 149 passing tests, but the tool session did not return a final summary, so it is not claimed as a full-suite pass
- `check:release` was attempted; three unrelated Windows-local failures were in existing standalone Bash/temp-directory cleanup paths (two `EBUSY`), outside this diff
- authoritative full-suite and packaged evidence comes from this PR's CI

## L3 gate

- [Core CI 32581049491](https://github.com/apache/maka/actions/runs/32581049491) passed in 6m09s, including the full Runtime Host suite, Desktop E2E, Storybook, and installed CLI validation.
- [Release Windows check 32581049489](https://github.com/apache/maka/actions/runs/32581049489) passed in 12m16s, including packaging, packaged smoke, pinned upgrade/uninstall, and automatic update end to end.
- Dependency audit did not run because this PR changes no dependency inputs.

Fresh exact-head core and packaged Windows evidence is green. The PR is ready for review.
## AI use

Select exactly one:

Tool(s) and scope: Codex implemented the per-election gate and process-settlement contract, added focused Node/Electron/owned-process regression evidence, ran local validation, and used independent review agents for the author gate. The commit includes a `Generated-by: Codex` trailer.

## Checklist

- [x] Yes — described under Summary and Boundary above
- [ ] No

### DIFF
--- a/.github/workflows/release-windows-check.yml
+++ b/.github/workflows/release-windows-check.yml
@@ -43,6 +43,8 @@ on:
       # MAKA_UPDATE_TEST_FEED to it — is only observable on this path.
       - 'apps/desktop/src/main/app-update-service.ts'
       - 'apps/desktop/src/main/runtime-host-boot.ts'
+      - 'packages/runtime-host/src/client/connect-or-spawn.ts'
+      - 'packages/runtime-host/src/client/launcher.ts'
       - 'scripts/prepare-windows-upgrade-baseline.mjs'
       - 'scripts/windows-upgrade-baseline.json'
       - 'scripts/verify-packaged-app.mjs'
--- a/packages/runtime-host/src/__tests__/host-kernel.test.ts
+++ b/packages/runtime-host/src/__tests__/host-kernel.test.ts
@@ -221,6 +221,103 @@ describe('non-serving Runtime Host kernel', () => {
     });
   });
 
+  test('keeps one live Candidate in flight for the whole election', async () => {
+    await withHostPaths(async (paths) => {
+      let launches = 0;
+      const result = await connectOrSpawnRuntimeHostWithDependencies(
+        {
+          rootPath: paths.root,
+          protocol: CURRENT_PROTOCOL,
+          compositionId: KERNEL_COMPOSITION.descriptor.id,
+          candidateEntrypoint: KERNEL_CANDIDATE_ENTRYPOINT,
+          electionDeadlineMs: 500,
+        },
+        {
+          random: () => 0,
+          launchCandidate: () => {
+            launches += 1;
+            return {
+              spawned: Promise.resolve({
+                pid: 4242,
+                exited: new Promise<never>(() => undefined),
+              }),
+            };
+          },
+        },
+      );
+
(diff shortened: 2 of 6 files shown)


jackwener (approved):
Authorized by @WAWQAQ in Raft task #9 after independent review at exact head 814d95f663c74fb6bd804de5eed3ae37d2eae482. No blocking findings; required test and Windows package checks are terminal green.

PR #3515, opened 2026-08-22
fix(runtime): classify provider server errors

hqhq1025:
## Summary

- classify structured `server_error` provider failures as `ProviderUnavailable` even when the stream omits an HTTP status
- make those transient provider failures retryable through the existing runtime retry path
- render persisted `server_error` failures as a model service error instead of an unknown error

## Root cause

OpenAI-compatible streaming errors can carry `code: server_error` without preserving the upstream HTTP status. The classifier only recognized numeric 5xx codes, so Maka normalized the failure to `kind: unknown`, disabled retry, and displayed the failed turn as an unknown error.

## Validation

- `node --test packages/runtime/dist/__tests__/provider-error-classification.test.js packages/runtime/dist/__tests__/model-adapter.test.js` (39 passed)
- `node --test apps/desktop/dist/main/__tests__/session-status-presentation.test.js` (4 passed)
- `git diff --check`

The full runtime suite completed with 3,059 passing tests and 9 unrelated existing failures in `model-factory-tool-call-index.test.js` and `node-pty-write-lifecycle.test.js`; the focused suites covering this change pass.

### DIFF
--- a/apps/desktop/src/main/__tests__/session-status-presentation.test.ts
+++ b/apps/desktop/src/main/__tests__/session-status-presentation.test.ts
@@ -19,7 +19,10 @@
 
 import assert from 'node:assert/strict';
 import { describe, it } from 'node:test';
-import { deriveFailedTurnRecovery } from '../../renderer/session-status-presentation.js';
+import {
+  deriveFailedTurnRecovery,
+  describeTurnErrorClass,
+} from '../../renderer/session-status-presentation.js';
 
 const outputFreeFailure = {
   partialOutputRetained: false,
@@ -28,6 +31,11 @@ const outputFreeFailure = {
 };
 
 describe('failed turn recovery presentation', () => {
+  it('presents persisted provider server errors as provider failures', () => {
+    assert.equal(describeTurnErrorClass('server_error', 'zh'), '模型服务返回错误');
+    assert.equal(describeTurnErrorClass('server_error', 'en'), 'Model service error');
+  });
+
   it('does not recommend a byte-identical retry after context overflow', () => {
     assert.deepEqual(
       deriveFailedTurnRecovery({ ...outputFreeFailure, errorClass: 'context_overflow' }, 'zh'),
--- a/apps/desktop/src/renderer/session-status-presentation.ts
+++ b/apps/desktop/src/renderer/session-status-presentation.ts
@@ -143,7 +143,12 @@ export function describeTurnErrorClass(errorClass: string | undefined, locale: U
   if (lower === 'network' || lower.includes('network') || lower.includes('fetch') || lower.includes('econn')) {
     return copy.network;
   }
-  if (lower === 'provider_unavailable' || /\b5\d\d\b/.test(lower)) return copy.provider;
+  if (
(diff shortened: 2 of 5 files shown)


jackwener (approved):
OpenAI-compatible streams can send `code: server_error` with no HTTP status; the classifier only matched 5xx numbers, so the turn became unknown and non-retryable. Mapping that structured code to `ProviderUnavailable` (retryable) and presenting `server_error` as a model-service error is the right class, before the 5xx regex. Tests cover classification, adapter, and UI copy. Approve.

Astro-Han (commented):
Thanks — the diagnosis is the valuable part here: an OpenAI-compatible stream can carry `type`/`code` of `server_error` with no HTTP status at all, and a classifier that only looks for a 5xx number will drop that into `unknown`, where it is neither retried nor explained to the user. Naming the carrier rather than the symptom is what makes this fix small.

Reviewed at exact head `54e6d5c7581ee9ee87d9c944587990738059fc73` against base `1ddbd2591ccc6fd06d9727a216e6a5b64d8b1e40`. **No findings — no P0, P1, P2 or P3.** One forward-looking note below. The head carries one completed `test` check and a maintainer approval, but the aggregate commit status is still `pending`, so we are not adding an approval on top.

**The thing we most wanted to disprove was that this is over-broad**, because reclassifying an error as retryable is exactly how a real failure gets silently swallowed and retried forever. It is not over-broad: the new branch sits behind every more specific signal. Abort, websocket transport, `402` billing, `429` rate limit, authentication, the context-overflow code set, `413`, and context-overflow text all return before it. A permission failure, a quota failure, or a context overflow cannot reach the `server_error` branch, so none of them can be turned into a silent retry.

**On the layering.** The runtime is the classification authority and the desktop change only adds a presentation mapping for the persisted `server_error` class. That is one source of truth with a display consumer, not a second retry policy living in the UI — worth stating because a duplicated classifier is the usual way this kind of change goes wrong.

**Note, not a finding.** The criterion is a one-element set: `server_error` alone maps to `ProviderUnavailable`. That is correct for the OpenAI semantics this targets, where the code means a transient server-side fault. It rests on that meaning holding, though — a provider that reused the same code for a non-transient configuration fault would have it retried instead of surfaced. There is nothing to reproduce today and nothing to change here; the comment already records why the code is treated as status-less, and it may be worth a line saying the set is deliberately narrow so a future addition gets the same scrutiny.

**How this was reviewed, including where we disagreed.** Two reviewers worked this independently without seeing each other's conclusion. One raised the single-code criterion above as a P3; the other had considered the same concern and refuted it as a finding. We checked the precedence chain ourselves and agree with the refutation, so it is published as a note rather than graded — the observation is worth having, the defect is not there.

Targeted suites executed against a worktree of this head: `provider-error-classification` 11/11 and the desktop `session-status-presentation` cases 4/4. The `model-adapter` suite could not start in our environment — it fails on a pre-existing `HttpsProxyAgent` typing issue and missing Slack dependencies unrelated to this change, so we are reporting it as not run rather than as a result.

---

PR #3519, opened 2026-08-22
fix(skills): show installed catalog state

hqhq1025:
## Summary

- replace the disabled download glyph with a success check when a catalog skill is already installed
- give the installed action a skill-specific accessible label in Chinese and English
- add rendering coverage for available and installed bundled-skill states

## Root cause

The bundled catalog already refreshed `installed` after a successful install, but the shared catalog action always rendered the download icon whenever it was not actively installing. The state changed correctly while the visual affordance continued to look installable.

## Verification

- `npm run build --workspace=@maka/core`
- `npm run typecheck --workspace=@maka/ui`
- `npm test --workspace=@maka/ui` (203 tests)
- `npm run format:check`
- `npm run check:asf-headers`
- `npm --workspace @maka/desktop run smoke:storybook` (163 stories)

## Screenshots

Before and after screenshots use the same Storybook story, viewport, state, content, and zoom. They will be attached in a PR comment.

### DIFF
--- a/apps/desktop/src/renderer/styles/module-pages/skills.css
+++ b/apps/desktop/src/renderer/styles/module-pages/skills.css
@@ -36,6 +36,11 @@
   color: var(--muted-foreground);
 }
 
+.maka-skill-install-complete-icon {
+  display: inline-flex;
+  color: var(--success);
+}
+
 .maka-skill-inspector {
   min-width: 0;
 }
--- a/packages/ui/src/__tests__/skills-panel-install-state.test.tsx
+++ b/packages/ui/src/__tests__/skills-panel-install-state.test.tsx
@@ -1,0 +1,60 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import assert from 'node:assert/strict';
+import test from 'node:test';
+import { renderToStaticMarkup } from 'react-dom/server';
+import { LocaleProvider } from '../locale-context.js';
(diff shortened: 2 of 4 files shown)


hqhq1025:
Before

clipboard

After

clipboard

Same Storybook story, viewport (1830 × 506), state, content, and zoom.

Astro-Han (commented):
Thanks — the root-cause paragraph is the part that makes this easy to review: the state was already correct and only the affordance lied, so the fix belongs in rendering and nowhere else. That framing is right, and the change stays inside it.

Reviewed at exact head `77d912f49497a3d08cc63c76deaa0f9ac4e6ef8b` against base `1ddbd2591ccc6fd06d9727a216e6a5b64d8b1e40`. **No findings — no P0, P1, P2 or P3.** One design note below, and one reason this cannot be an approval yet.

**What we checked rather than assumed:**

- **Icon precedence.** `installing` is evaluated before `installed`, so an in-flight install shows the spinner rather than briefly flashing a completed check. Easy to get backwards; it is not.
- **The disabled state was already correct.** `isDisabled={installed || skillActionBusy || !onInstall}` predates this change, which is exactly why the bug was cosmetic rather than functional — the button could not be clicked, it just looked like it could.
- **`--success` resolves.** It is defined in `maka-tokens.css` across every theme variant, and `color: var(--success)` is the same usage already present in `chat-detail.css` and `task-ledger.css`.
- **The class lives where its siblings live.** `.maka-skill-install-complete-icon` sits in `apps/desktop/.../module-pages/skills.css` while the component is in `packages/ui`. That looked like cross-package drift until we checked: `maka-skill-inspector`, `maka-skill-row-label` and `maka-skill-row-slug` are all defined the same way. This follows the existing convention rather than breaking it.
- **Proportion.** Roughly sixty of the eighty added lines are the new test; the behavioural change is one branch, one copy key per locale, and three lines of CSS.

**Design note, not a finding.** `catalogInstallButton` is shared by both catalog rows, and the two call sites derive `installed` from different sources: the bundled row passes `entry.installed` (the flag your root-cause paragraph describes), while the marketplace row computes `skills.some((skill) => skill.id === source.id)`. Both are pre-existing and neither is changed here, but the new check icon now keys off both. The added test covers the bundled path only — which your description states accurately. Worth a second test on the marketplace path at some point, not in this PR.

**On CI.** The head has one completed check (`test: success`), but the aggregate commit status is `pending`, so this is not terminal green and we are not approving it. Once the full set lands green, there is nothing here that would stand in the way.

**How this was reviewed.** Two of us read it independently without visibility into each other's conclusion and reached the same result; the targeted `@maka/ui` suite was executed against a worktree of this head (203 passing, including the two cases this PR adds). The note above came from one reviewer and the proportion observation from the other — neither of us saw both on our own.

---

jackwener (approved):
Catalog `installed` already flipped; the button still painted Download. Check + disabled + skill-specific label is the minimal visual match. Tests cover both states. Approve.

Issue #3522, opened 2026-08-22
docs: verify every live document against the code, one subsystem at a time


Joob1n:
## Goal

Bring every live document under `docs/` back in line with the code, one subsystem at a time.

The work is grouped by subsystem rather than by document because that is how the drift arrives. When a subsystem changes, several documents describe it from different angles, and they go stale together. Reviewing them as a group is also the only way to notice that two documents now contradict each other.

Each box below is one subsystem. Tick it when every document under it has been read against the current implementation and either corrected or confirmed accurate. A document that turns out to be accurate needs no change — recording that it was checked is the point.

## What a pass over one subsystem covers

- Named symbols still exist: classes, methods, tool names, event types, config keys.
- Described storage matches the implementation. The migration from per-run files to SQLite invalidated file-layout diagrams in more than one place.
- Cited file paths resolve, including test paths offered as the verification route.
- Documents in the group do not contradict each other.
- Paired translations move together. `translation_status: synced` is the only value used across the paired documents in this repository, so updating one language alone makes that field false.
- `last_verified` reflects the pass.

## Scope

In scope: the 61 documents listed below.

Out of scope:

- `docs/archive/**` (38 files). Each carries an "Archived on `<date>` … not current architecture authority" banner. Their historical content is correct as archived material and must not be rewritten.
- `docs/eval/**` (3 files). These are benchmark run records tied to a specific date and model set, not descriptions of current behaviour.

## Subsystems

- [x] **Runtime core** (2) — done in #3523
  - `docs/architecture/runtime-core-architecture-draft.md` — `PermissionEngine`, `respondToPermission()`, the `run.json`/`events.jsonl` layout, a step limit of 50, and a deleted test path — addressed by #3523
  - `docs/architecture/runtime-core-architecture-draft.zh-CN.md` — same set as its counterpart — addressed by #3523

- [ ] **Recovery and resume** (7)
  - `docs/architecture/runtime-recovery-resolver-adr.zh-CN.md`
  - `docs/architecture/runtime-resume-architecture.md`
  - `docs/architecture/runtime-resume-architecture.zh-CN.md`
  - `docs/architecture/runtime-resume-extraction-ledger.zh-CN.md`
  - `docs/architecture/runtime-resume-phase0-crash-contract.md`
  - `docs/architecture/runtime-resume-phase1-safe-boundary-contract.md`
  - `docs/architecture/runtime-resume-phase3-phase4-workspace-checkpoint-design.zh-CN.md`

[...]

PR #3523, opened 2026-08-22
docs: correct stale runtime-core architecture facts

Joob1n:
## Summary

Chapter 1 of the runtime-core architecture documents still described a file-backed run store and the removed per-call permission model. Both language versions carried the same drift, so `translation_status: synced` was no longer true; this change updates them together.

Corrected against the code:

- `PermissionEngine` and its allow/block/prompt evaluation no longer exist. A tool that crosses the sandbox boundary returns `sandbox_boundary_required` with a concrete expansion, and the model raises the request through `request_sandbox_boundary`.
- The decision is routed by `RuntimeKernel.respondToSandboxBoundary()`, not `respondToPermission()`.
- Durable state is SQLite, not `sessions/<id>/runs/<id>/{run.json,events.jsonl,runtime-events.jsonl}`. RuntimeEvents land in `runtime_events`, ordered by `event_seq` under a `(invocation_id, event_seq)` uniqueness constraint.
- `maxSteps` is `number | undefined` and `undefined` is unbounded; the documented default of 50 appears nowhere in the repository.
- `packages/storage/src/__tests__/agent-run-store.test.ts` no longer exists. The ledger repair coverage is `packages/runtime/src/__tests__/runtime-ledger-repair.test.ts`.
- Startup recovery and continuation were described in a way that contradicted itself. They are now stated as the two distinct paths they are, naming `safe_boundary_continuation`, `continuationSource` and `runAdmittedRuntimeContinuation`.

Two facts are added because the corrected text would otherwise be incomplete: `AiSdkFlow` rejects `permission_request` / `permission_answer_ack` / `permission_closure_ack` / `permission_decision_ack` at the mapping boundary, and the boundary decision fact carries `role=system` with `author=user`.

No new sections or examples. This change only brings existing statements back in line with the implementation, and the net diff is shorter than the original.

Refs #3522

## Verification

Every corrected statement was checked against the code rather than inferred from the wording:

| Statement | Check | Result |
|---|---|---|
| `PermissionEngine` removed | `grep -rn PermissionEngine packages/*/src` | no matches |
| `respondToSandboxBoundary()` | `packages/runtime/src/runtime-kernel.ts:149,2339` | present; no `respondToPermission` |
| no file-backed ledger | `grep -rln 'runtime-events.jsonl\|events.jsonl' packages/storage/src` | no matches |
| `maxSteps` unbounded | `packages/runtime/src/ai-sdk-backend.ts:739,1053` | `maxSteps?: number` / `number \| undefined` |
| no default of 50 | `grep -rn 'maxSteps.*50' packages/` | no matches |
| stale test path | `ls packages/storage/src/__tests__/agent-run-store.test.ts` | No such file |
| replacement test path | `ls packages/runtime/src/__tests__/runtime-ledger-repair.test.ts` | present |
| continuation symbols | `packages/core/src/agent-run.ts:71,167`, `packages/runtime/src/runtime-runner.ts:421` | present |
| legacy permission events | `packages/runtime/src/ai-sdk-flow.ts:148-170` | the four listed types throw |

Repository checks:

```
npm run format:check   Checked 1590 files. No fixes applied.
npm run check:stale    dist is fresh.
```

No test covers prose, and this PR changes no code, so no suite was added. `npm test` was not run.

## AI use

Select exactly one:

Tool(s) and scope: Claude Opus 5 via Claude Code — ran the documentation audit against the code, drafted the corrected prose in both languages, and wrote the commit message. Every claim in the table above was verified by running the listed command. The commit carries a `Generated-by` trailer.

## Checklist

- [x] No

### DIFF
--- a/docs/architecture/runtime-core-architecture-draft.md
+++ b/docs/architecture/runtime-core-architecture-draft.md
@@ -7,7 +7,7 @@ counterpart: ./runtime-core-architecture-draft.zh-CN.md
 implementation_status: current
 document_status: draft
 translation_status: synced
-last_verified: 2026-07-12
+last_verified: 2026-08-23
 owners:
   - maka-backend
 ---
@@ -18,7 +18,7 @@ owners:
 
 This chapter is for engineers entering the Maka Runtime for the first time and maintainers changing its main execution path. The first half should give you a working map of the runtime boundaries. By the end, you should be able to locate the main implementation and understand the invariants that changes to termination, tools, or persistence must preserve.
 
-The chapter describes the implementation on the production path as verified on 2026-07-12. Phase plans in historical design documents are not treated as current behavior.
+The chapter describes the implementation on the production path as verified on 2026-08-23. Phase plans in historical design documents are not treated as current behavior.
 
 ## Start with a deceptively simple request
 
@@ -97,7 +97,7 @@ This preserves the original semantics of model interaction rather than a UI-form
 - thinking may carry a provider-required signature;
 - tool calls and results are paired by stable IDs;
 - a tool call can point to its assistant step;
-- permission requests and decisions are actions, not text pretending to be chat;
+- sandbox boundary requests and decisions are actions, not text pretending to be chat;
 - a terminal event closes an Invocation explicitly instead of relying on whether the last message looks like an answer.
 
 Model history therefore does not need to reverse-engineer the UI transcript. It can select non-partial, model-visible RuntimeEvents, preserve their order, and materialize text-only or provider-native messages according to provider capabilities. The UI is likewise a projection rather than the source of truth.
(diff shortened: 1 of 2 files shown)


jackwener (approved):
Checked the claims against current `main`, not the old prose.

- `PermissionEngine` / `respondToPermission()` are gone; `respondToSandboxBoundary()` exists.
- Durable runtime events are SQLite `runtime_events` with `UNIQUE (invocation_id, event_seq)`, not per-run JSONL.
- `maxSteps` is `number | undefined`; no default 50 in `packages/`.
- Ledger repair test path is `runtime-ledger-repair.test.ts`.
- `AiSdkFlow` throws on the four legacy permission event types.
- `sandbox_boundary_decision_ack` is `role=system` / `author=user` in the mapping comments.
- Continuation is a separate path (`safe_boundary_continuation`, `continuationSource`, `runAdmittedRuntimeContinuation`).
- EN/ZH stay in lockstep; `translation_status: synced` holds.

This only un-drifts Chapter 1. Approve.

Astro-Han (approved):
Thanks — correcting a stale architecture document is unglamorous and it is exactly the kind of drift that costs the next reader an hour. Net −20 lines is the right shape for it: fewer claims, each of them true.

Reviewed at exact head `1d014d481c484caa6d6c953df0ca247a970aa71e`. `test` is green on this head.

We spot-checked three of the load-bearing claims against current `main` rather than taking the prose at its word:

- `PermissionEngine` and `respondToPermission()` return no hits anywhere under `packages/` or `apps/`, and `respondToSandboxBoundary()` is what exists in their place.
- `maxSteps` is declared `maxSteps?: number` and held as `number | undefined`, resolved through `input.maxSteps ?? this.maxSteps` — there is no default of 50.
- Durable runtime events live in the SQLite `runtime_events` table, not per-run JSONL.

Each matches what this PR now says. @jackwener had already worked through the full list against `main`, which is the review that carries the weight here; ours agrees with it on the points we re-derived independently.

One thing worth naming for whoever reads this later: a documentation change that *deletes* claims is harder to review than one that rewrites them, because a wrongly-deleted line leaves nothing behind to notice. The deletions here are removals of superseded mechanisms, not of still-true statements.

**No findings — no P0, P1, P2 or P3.** Approving.

---

Issue #1097, opened 2026-07-16
fix(cli): expanding an off-screen block triggers a full redraw that clears pre-Maka scrollback


Astro-Han:
## Problem

`tui.setClearOnShrink(false)` keeps the transcript shrink path from wiping terminal scrollback, but a documented known limit remains (`pi-tui-runner.ts`, "Known limit" comment): a global `Ctrl+O` / `Ctrl+T` toggle that resizes a block sitting *above* the live viewport top makes pi-tui's differential renderer fall back to a full redraw (its `firstChanged < viewportTop` path), which emits a scrollback-clearing sequence. Transcript content is rebuilt into fresh scrollback, so nothing is lost from the transcript itself — but the user's scroll position resets and any pre-Maka terminal scrollback is cleared.

Verified 2026-07-16 against `main`: the other half of the original #549 gate — 80×24 truncation semantics — is already satisfied by the #1082 compact-row grammar (`assembleCompactToolRow` preserves the tool head, truncates the free-text target first, and reserves protected fixed-shape annotations; the remaining unprotected-annotation gap is #1086).

## Desired outcome

Expanding or collapsing blocks must never clear pre-Maka terminal scrollback or reset the user's scroll position. The clean fix is a preserve-scrollback render path inside pi-tui (upstream, not owned by this repo); evaluate an upstream patch against a maka-side avoidance strategy (e.g. anchoring expansion so resizes stay inside the live viewport). Add behavioral coverage for long runtime-resource paths and off-screen expansion.

(Split out of the retired #549 tracker.)

PR #3526, opened 2026-08-22
fix(cli): render tool results when cards settle, not at turn end

### LINKED ISSUE #3521: bug(cli): TUI tool cards settle to "(no output)" and stay stuck once scrolled out of the viewport

Astro-Han:
## Summary

Fixes #3521 — TUI tool cards settled into `(no output)` and, once scrolled out of the viewport, stayed that way forever; Ctrl+O/Ctrl+T could not touch them.

Root cause: the live `tool_start` / `tool_result` broadcast frames carried no args and no result content. A card therefore settled into the empty placeholder, and the only content path was the turn-terminal durable backfill — which arrives after terminal scrollback has already frozen the card's lines (#1139), and the viewport-anchored toggles (#1130) then lock the stale render in place.

The fix makes live events self-sufficient, which is the design the immutable-scrollback medium requires: the card must be renderable at settle time, before it can scroll away. The runtime already holds the full result when it emits the event (persist-then-broadcast in `tool-runtime.ts`), so the Host now projects it into the live frame instead of dropping it:

- `tool_start` frames carry `args` (bounded to 16 KB serialized; oversized args are omitted, matching today's mid-turn placeholder for that rare case).
- `tool_result` frames carry the settled `content` (bounded to 48 KB serialized, keeping the whole frame under the 64 KB subscription frame cap); an oversized result degrades to its serialized byte size (`contentBytes`) so the row stays truthful instead of claiming `no output`.
- The CLI adapter projector deletes the `{ kind: 'text', text: '' }` substitution and passes both fields through; the TUI settles cards with the real result, so the freeze now preserves a correct render. Oversized results render as `N bytes` with a dim note in the expanded card, and the existing turn-terminal reconcile (kept as the backstop for shed/oversize edges) heals the full body at turn end.

What deliberately did NOT change: raw provider error text and thinking signatures remain rejected on the wire; the desktop merge paths are untouched (live content simply reaches its projection earlier); the viewport-anchored expansion toggles stay as designed — with content correct at settle time, the "earlier cards show nothing" half of the issue disappears, and the remaining "cannot retoggle scrollback" half is the documented #1097/#1130 tradeoff.

## Verification

Same scripted driver, same four mid-turn cards, captured through `runMakaPiTui` + `FakeTerminal`.

**Before (main)** — cards settle to `(no output)` / duration-only, with no input target:

image

**After (this PR)** — settle-time facts: invocation target + outcome, including a truthful oversized-result size instead of a `no output` lie:

image

- New behavior tests, each seen red before the implementation: live `tool_result` broadcasts content, `tool_start` broadcasts args, oversized args/content degrade to omission/`contentBytes` (frame re-decodes under the subscription cap), the adapter projector passes them through, and the TUI renders `100000 bytes` rather than `no output` until the terminal reconcile heals the body.
- The privacy-normalization protocol test was rewritten: `args`/`content` are now accepted (bounded), while raw provider `error` text and thinking `signature` frames remain rejected.
- `@maka/runtime-host` workspace suite: 1075/1075 pass. `@maka/cli`: 374/374. `@maka/ui`: 201/201. `tsc` builds and `biome format:check` are clean.
- Not run: full-repository suite and desktop E2E (left to CI); no live TUI session was driven end-to-end.

## Security

This relaxes a deliberate boundary: live subscription frames were privacy-normalized to exclude tool args/results. The same data was always available to the same clients through the durable transcript read path, and `tool_output_delta` already streams raw output live; this change removes the asymmetry. Per the maintainer's call on #3521, that normalization is not a priority for this channel. Bounded sizes keep frames under the existing cap.

## AI use

Tool(s) and scope: Maka (Kimi K3) diagnosed the root cause, designed the wire shape, and wrote the implementation and tests. Commits carry `Generated-by` trailers.

## Checklist

### DIFF
--- a/packages/cli/src/__tests__/pi-transcript.test.ts
+++ b/packages/cli/src/__tests__/pi-transcript.test.ts
@@ -351,6 +351,88 @@ describe('Maka Pi TUI transcript', () => {
     assert.equal(state.entries.at(-1)?.kind, 'notice');
   });
 
+  test('shows the byte size of an oversized live tool result instead of no output', () => {
+    const state = createMakaPiTranscriptState();
+    applyMakaSessionEventToTranscript(
+      state,
+      event({
+        type: 'tool_start',
+        toolUseId: 'big-1',
+        toolName: 'Bash',
+        args: { command: 'npm test' },
+      }),
+    );
+    applyMakaSessionEventToTranscript(
+      state,
+      event({
+        type: 'tool_result',
+        toolUseId: 'big-1',
+        isError: false,
+        durationMs: 2500,
+        content: { kind: 'text', text: '' },
+        contentBytes: 100_000,
+      }),
+    );
+
+    const compact = renderMakaPiTranscript(state, meta(), 80).map(stripAnsi).join('\n');
+    assert.match(compact, /100000 bytes/);
+    assert.doesNotMatch(compact, /no output/);
+
+    assert.equal(toggleAllToolExpansion(state), true);
+    const expanded = renderMakaPiTranscript(state, meta(), 80).map(stripAnsi).join('\n');
+    assert.match(expanded, /too large to show live: 100000 bytes/);
+  });
+
+  test('the terminal reconcile replaces an oversized placeholder with the durable content', () => {
+    const state = createMakaPiTranscriptState();
+    applyMakaSessionEventToTranscript(
+      state,
+      event({
+        type: 'tool_start',
+        toolUseId: 'big-1',
+        toolName: 'Bash',
+        args: { command: 'npm test' },
+      }),
+    );
+    applyMakaSessionEventToTranscript(
+      state,
+      event({
+        type: 'tool_result',
+        toolUseId: 'big-1',
+        isError: false,
+        content: { kind: 'text', text: '' },
+        contentBytes: 100_000,
+      }),
+    );
+
+    assert.equal(
+      reconcileToolsWithStoredMessages(state, 'turn-1', [
+        {
+          type: 'tool_call',
+          id: 'big-1',
+          turnId: 'turn-1',
+          ts: 1,
+          toolName: 'Bash',
+          args: { command: 'npm test' },
+        },
+        {
+          type: 'tool_result',
+          id: 'big-1-result',
+          turnId: 'turn-1',
+          ts: 2,
+          toolUseId: 'big-1',
+          isError: false,
+          content: { kind: 'text', text: 'all 3 suites passed' },
+        },
+      ]),
+      true,
+    );
+
+    const row = renderMakaPiTranscript(state, meta(), 80).map(stripAnsi).join('\n');
+    assert.match(row, /all 3 suites passed|1 line · 19 bytes/);
+    assert.doesNotMatch(row, /100000 bytes/);
+  });
+
   test('removes a live poll card that the durable transcript folds into its Bash parent', () => {
     const state = createMakaPiTranscriptState();
     for (const tool of [
--- a/packages/cli/src/pi-transcript-tools.ts
+++ b/packages/cli/src/pi-transcript-tools.ts
@@ -193,7 +193,15 @@ function renderExpandedToolBlock(entry: MakaPiToolEntry, width: number): string[
     }
     lines.push(...renderToolStreams(entry.outputDeltas.values(), width));
   }
-  if (entry.result || entry.output) {
+  if (entry.resultBytes !== undefined && !plainResultText(entry)) {
+    lines.push(
+      ...renderIndented(
+        ansi.dim(`Result too large to show live: ${entry.resultBytes} bytes`),
+        width,
+        2,
+      ),
+    );
+  } else if (entry.result || entry.output) {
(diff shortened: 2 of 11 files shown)


jackwener (dismissed):
Independent review of `7aa64d298`. Did not treat Astro-Han authorship as evidence.

1) Problem. Live `tool_start` / `tool_result` frames had no args or result. Cards settled to `(no output)` and terminal scrollback froze that lie before the durable backfill.

2) Solution. Persist-then-broadcast already has the result; project it onto the live frame. Args ≤16 KB, content ≤48 KB (frame cap 64 KB−1). Oversized args omitted; oversized content becomes `contentBytes` so the row shows `N bytes` instead of `no output`. Durable reconcile still replaces the placeholder at turn end. Error text and thinking signatures stay rejected.

3) Timing. Events are keyed by `toolUseId`; later tools do not overwrite earlier cards. `content` and `contentBytes` are mutually exclusive on the wire. Reconcile clears `resultBytes` when real text lands.

Privacy: live frames now carry the same args/results the durable transcript already exposed; `tool_output_delta` already streamed output. Size bounds keep the existing cap.

Approve. CI `test` was still running on this head; merge stays with the author.

Astro-Han:
Closing as superseded by #3524, which has now merged and closed #3521.

Both PRs address the same TUI failure, but #3524 keeps the durable transcript as the single result authority and reuses the existing reconciliation seam without changing the Runtime Host protocol. This branch would additionally duplicate tool args/results into live frames, introduce size-dependent `content` / `contentBytes` states, bump the compatibility epoch, and expand the live-frame privacy contract. With the bug already fixed, that permanent protocol and client complexity no longer has a demonstrated benefit.

If the per-tool transcript refresh or its settle-time latency becomes a measured problem, the next step should be to simplify the existing refresh path or reuse the long-lived subscription’s durable watermark/page mechanism, rather than merge both representations.

PR #3528, opened 2026-08-22
docs: restyle README header and correct stale facts

jackwener:
## Summary

Restyle the README opening the way the owner asked, and un-drift a few stale facts.

- Center the current Desktop icon (`apps/desktop/assets/icon.png`) with the project name.
- Keep only star, download, license, and platform badges. Platform support is three coordinated badges (macOS arm64 / Windows preview / Linux soon) instead of one combined pill.
- Add a short intro under the title. Do not copy the rest of the Orca README.
- Replace `permission engine` with the sandbox boundary.
- Align the Chinese Windows sentence with English (preview, not “unsupported plus preview”).
- Chinese Eval: `外部竞品` → external subjects.
- Point `docs/windows-support.md` issue #2142 at `apache/maka`.

## Verification

- Visual: GitHub README render of the header, badges, and intro in both languages.
- Fact checks against current `main` (`PermissionEngine` gone; Windows is unsigned preview).

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,17 +17,33 @@
   under the License.
 -->
 
-# Apache Maka (Incubating)
+<h1 align="center">
+  <img src="apps/desktop/assets/icon.png" alt="Maka" width="72" valign="middle" /> Apache Maka
+</h1>
 
-[![CI](https://github.com/apache/maka/actions/workflows/ci.yml/badge.svg)](https://github.com/apache/maka/actions/workflows/ci.yml)
-[![License](https://img.shields.io/badge/License-Apache_2.0-blue.svg)](./LICENSE)
-[![docs](https://img.shields.io/badge/docs-%E7%AE%80%E4%BD%93%E4%B8%AD%E6%96%87-blue?logo=googletranslate&logoColor=white)](./README.zh-CN.md)
+<p align="center"><sub>Incubating at The Apache Software Foundation</sub></p>
+
+<p align="center">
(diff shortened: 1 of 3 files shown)


Astro-Han (approved):
Thanks — moving the entry page from maintainer detail to user-facing language is the right call, and net −50 lines for a README is the good direction.

Reviewed at exact head `9a844c61b165bd6b76361c0086507be8c809fca8`, freshly fetched. `test` is green. Three files: both READMEs and `docs/windows-support.md`.

The thing I most wanted to check on a docs deletion is whether a warning lost its only home. It hasn't — empty threads after upgrade, credentials needing re-entry, resume spending tokens, and "no published Apache release yet" all survive in the new copy. What went is architecture and backup-implementation detail that `ARCHITECTURE.md` and the privacy/resume docs already own, so this removes a duplicate rather than the fact itself.

Also verified rather than assumed: the badges and in-repo icon resolve, every relative link across the three files resolves, issue #2142 is still the live Windows tracking issue, and the current release really does carry macOS arm64 plus unsigned Windows x64 — so the platform claims match reality. The Chinese and English READMEs stay in step on structure, platform status and warnings.

One thing I'd expected to find and didn't: no stale `PermissionEngine` phrasing survives here. `SECURITY.md` now frames the boundary in OS-account terms and this copy agrees with it.

**No findings.** Approving.

---

Issue #3529, opened 2026-08-22
bug(runtime): a steer submitted during a tool-free turn is never injected — the agent loop only drains steering at a tool-call boundary


shaokeyibb:
## What happened

A steer typed during a turn that ends without another tool call is **never injected**. The turn finishes normally, the model never sees the interjection, and the queue entry is left unpulled.

`AiSdkBackend` drains steering only at the top of each `agentLoop` iteration:

```ts
// packages/runtime/src/ai-sdk-backend.ts:2209-2210
agentLoop: for (;;) {
  await this.drainSteeringInto(scope, input, queue);
```

and the loop only iterates again when the step returned tool calls:

```ts
// packages/runtime/src/ai-sdk-backend.ts:2787-2797
const stepLimitReached = maxSteps !== undefined && runtimeSteps >= maxSteps;
if (
  returnedToolCalls.length > 0 &&
  !stepLimitReached &&
  !scope.loopStopRequested &&
  !scope.aborted
) {
  currentStepMessageId = this.newId();
  continue agentLoop;
}
break agentLoop;
```

So the only injection points in a turn are its tool-call boundaries. A steer that arrives during the last step — or during the only step of a plain conversational answer — has no boundary left and is silently skipped. Whether "Steer" works at all therefore depends on whether the model happens to call a tool afterwards, which the user cannot know when they press Enter.

The fake backend does not share this shape: it drains between chunks *and* once after the last chunk, commented "Final stranded drain (grok-build safety): a steer that landed after the last boundary still lands in this turn instead of being lost" (`packages/runtime/src/test-only/fake-backend.ts` ~275, ~291). That safety net is why the existing suites are green.

Downstream, the unpulled entry is folded to `followup` at the terminal transition, which currently poisons the Host — filed separately as #3530. This issue is only about the missing injection point.

## How to reproduce

Drive `AiSdkBackend` directly with a `pullSteering` that yields nothing on the first call and a message from the second call onward — the real timing of pressing Enter after the turn has started streaming — and run the same input against a text-only model and a tool-then-text model:

```
[TOOL-FREE] event types = text_delta,text_complete,token_usage,complete
[TOOL-FREE] pullSteering=1  steering_message=0  acked=[]  nacked=[]  stopReason=end_turn

[TOOL-CALL] event types = tool_start,tool_result,steering_message,text_delta,text_complete,token_usage,complete
[...]

PR #3532, opened 2026-08-22
fix(runtime-host): keep folded steering entries decodable in the followup queue

### LINKED ISSUE #3530: bug(runtime-host): an unpulled steering message poisons the Host — the folded followup entry keeps current_turn placement and fails its own protocol decode
## What happened

A steering message that the running turn never pulls **poisons the Runtime Host**. The Host begins draining while the turn is still `running`, so the in-flight answer stops mid-stream, no successor turn starts, and the interjection is lost with no trace in any surface.

This is reachable from ordinary TUI use: press Enter during a turn that ends without another tool call (see #3529 — the agent loop only drains steering at a tool-call boundary, so the message is simply never pulled).

Root cause is one field. `beginTerminalTransition` folds unpulled steering into `followup` and resets `state`, but not `placement`:

```ts
// packages/runtime-host/src/server/message-coordinator.ts:420-426
const folded = state.steering.splice(0);
for (const entry of folded) entry.state = 'queued';   // placement stays 'current_turn'
if (folded.length > 0) {
  state.followup.unshift(...folded);
  this.#mutated(state);
}
```

The followup entries are projected through `queuedSnapshot` (`message-coordinator.ts:1463`), which copies `entry.placement` verbatim. The protocol decoder rejects exactly that:

```ts
// packages/runtime-host/src/protocol/message.ts:283
if (decoded.state !== 'queued' || decoded.placement !== 'next_turn') {
  throw invalidProtocolFrame('Invalid followup queue entry');
}
```

So the session's own queue projection becomes undecodable. The failure cascades:

1. `decodeSessionMessageQueueProjection` throws inside `createSessionContinuitySnapshot`;
2. `completeTerminalTransition` fails **before** `commitNextRoot`, so `active.messageTransitionCommitted` stays `false`;
3. `drainTurn`'s `finally` therefore calls `messages.abandonRootReservation(...)` (`root-turn-coordinator.ts` ~2196);
4. that sees the still-live steering entry, calls `#failStop()` and throws `Root reservation with confirmed Message effects cannot be abandoned`;
5. the `catch` calls `requestHostDrain()` — the whole Host drains.

Note the asymmetry that allowed this: `queuedSteeringSnapshot` (`message-coordinator.ts:1474`) already asserts its own placement invariant and throws `Steering entry lost current-turn placement`. The followup side has no such guard.

## How to reproduce

Requires a steering message that the backend never pulls. The bundled fake backend hides this, because it drains steering at every chunk boundary *and* once more after the last chunk — commented "Final stranded drain (grok-build safety): a steer that landed after the last boundary still lands in this turn instead of being lost" (`packages/runtime/src/test-only/fake-backend.ts` ~275 and ~291). Neutering both drain sites reproduces it against the real Host:

1. `withExecutionRoot` → `startHost()` → `connectClient()`;
2. `client.startTurn(...)` with any ordinary prompt; wait for `running`;
3. `client.request('turn.message.submit', { ..., placement: 'current_turn' })`;
4. poll `queryTurn` every 500 ms.

Observed, against a no-steer control that is otherwise identical:

```
[steer]   t=500ms  first-turn status = running
[steer]   t=1000ms queryTurn threw = RuntimeHostOperationError: Runtime Host is draining

[control] t=500ms  status = running
[control] t=1000ms status = completed
```

The submit itself is accepted normally: `{"disposition":"steering","queueRevision":1}`.

## Environment

- Maka commit: `d472a359b`
- OS: Windows 11 (26200); also reproduced under WSL
- Surface: TUI against Runtime Host
- Node.js: v22.18.0

The original failure, captured by logging `commandFailure` in `drainTurn`'s catch:

```
RuntimeHostProtocolError: Invalid followup queue entry   { code: 'invalid_frame' }
    at decodeFollowupMessages              (protocol/message.js:178)
    at decodeSessionMessageQueueProjection (protocol/message.js:68)
    at decodeSessionContinuitySnapshot     (protocol/session-continuity.js:312)
    at createSessionContinuitySnapshot     (canonical-session-projection.js:113)
```

and the resulting fail-stop, captured by tracing `RuntimeMessageAuthorityInvariantError` construction:

```
Root reservation with confirmed Message effects cannot be abandoned
    at HostMessageCoordinator.abandonRootReservation (message-coordinator.js:132)
    at RootTurnCoordinator.drainTurn                 (root-turn-coordinator.js:1508)
```

followed by `Stop fence cannot replace a terminal transition` and `Message coordinator closed with a live owner, entry, or transition`.

### Why no test catches it

- `message-coordinator.test.ts` "release folds unpulled steering ahead of follow-up without changing source semantics" is a unit test on the coordinator — it never round-trips the projection through the protocol codec, so the bad `placement` is invisible. It also always submits a followup alongside the steering, never a lone steer.
- Every Host-level steering test runs against the fake backend, which never leaves a steer unpulled.

### Fix constraint

The durable provenance must not change: that same test asserts `batch.sources` keeps `placement: 'current_turn'` / `disposition: 'steering'` for folded entries, and `sourceFromEntry` (`message-coordinator.ts:1453`) is the right place for that. Only the **queue projection** needs the folded entry to read `next_turn`, mirroring the guard `queuedSteeringSnapshot` already has on the other side.

shaokeyibb:
## Summary

A steering message the run never pulls is folded into `followup` at the terminal transition, but the fold only resets `state` — the entry keeps `placement: 'current_turn'`. The followup queue projection then fails `decodeFollowupMessages`, which requires `next_turn`. That throw surfaces inside `createSessionContinuitySnapshot`, so `completeTerminalTransition` fails before `commitNextRoot`, `drainTurn`'s `finally` falls through to `abandonRootReservation`, the still-live entry trips `#failStop()`, and the Host drains mid-turn.

The fix projects followup entries by queue position: an entry in the followup queue is a next-turn message, whatever it was submitted as. Where the message was originally aimed already lives on `disposition` and on the durable `sourceFromEntry` record — which the existing fold test pins, and which this change deliberately leaves alone.

`queuedSnapshot` itself is unchanged because `retractedSnapshot` also builds on it, and a retracted steering entry legitimately carries `current_turn`.

Fixes #3530

## Verification

New unit test `a lone folded steering entry leaves the queue projection decodable` folds a lone unpulled steer and round-trips the projection through `decodeSessionMessageQueueProjection`. Without the fix it fails on `expected: 'next_turn' / actual: 'current_turn'`.

End-to-end, against a real Host with the fake backend's two steering drains temporarily disabled (they are what hides this today — see the "Final stranded drain" comment in `fake-backend.ts`), polling `queryTurn` every 500 ms:

```
before  t=500ms  running
before  t=1000ms RuntimeHostOperationError: Runtime Host is draining

after   t=500ms  running
after   t=1000ms completed
```

`after` matches a no-steer control exactly. That scaffolding is not part of this PR; #3529 covers why the steer is never pulled in the first place.

Ran: `message-coordinator` (31), `session-projector` (5), `execution-host-message` (4) — all green; `biome check` on both changed files; `npm run check:asf-headers`; `tsc -p packages/runtime-host`.

`canonical-session-projection` fails 7/7 here, all `EBUSY: resource busy or locked, unlink ...runtime.sqlite` on teardown. Verified pre-existing: identical 7/7 against a clean `origin/main` build of the same file. This is a Windows-only SQLite unlink problem in the fixture, unrelated to the change.

Not run: the rest of the `runtime-host` suite, Desktop, and Playwright — this change is confined to the Host queue projection.

## Root cause

The asymmetry that allowed it: `queuedSteeringSnapshot` already asserts its own placement invariant (`Steering entry lost current-turn placement`), and `inFlightSnapshot` does the same. The followup side had no guard, so a wrong placement reached the wire instead of failing loudly next to the code that produced it.

## AI use

Select exactly one:

Tool(s) and scope: Claude Code — investigation, root-cause tracing, the test, and the fix.

## Checklist

- [ ] No

### DIFF
--- a/packages/runtime-host/src/server/message-coordinator.ts
+++ b/packages/runtime-host/src/server/message-coordinator.ts
@@ -1347,7 +1347,7 @@ export class HostMessageCoordinator implements RuntimeMessageAuthority {
         ...[...state.inFlight.values()].map(inFlightSnapshot),
         ...steering.map(queuedSteeringSnapshot),
       ],
-      followup: followup.map(queuedSnapshot),
+      followup: followup.map(queuedFollowupSnapshot),
     };
   }
 
@@ -1475,6 +1475,19 @@ function queuedSteeringSnapshot(entry: LiveEntry): SteeringMessageSnapshot {
     throw new RuntimeMessageAuthorityInvariantError('Steering entry lost current-turn placement');
   }
   return { ...queuedSnapshot(entry), placement: 'current_turn' };
+}
+
+/**
+ * Queue position, not origin: an entry in the followup queue is a next-turn
+ * message by definition, including a steering entry the run never pulled and
+ * the terminal transition folded ahead of the followups. Where the message was
+ * originally aimed stays on `disposition` and on the durable
+ * {@link sourceFromEntry} record. Reporting a folded entry as `current_turn`
+ * here makes the projection fail its own wire decode, which takes the Host
+ * down through the session continuity snapshot (#3530).
+ */
+function queuedFollowupSnapshot(entry: LiveEntry): QueuedMessageSnapshot {
+  return { ...queuedSnapshot(entry), placement: 'next_turn' };
 }
 
 function inFlightSnapshot(entry: LiveEntry): SteeringMessageSnapshot {
--- a/packages/runtime-host/src/__tests__/message-coordinator.test.ts
+++ b/packages/runtime-host/src/__tests__/message-coordinator.test.ts
@@ -29,6 +29,7 @@ import type {
 import {
   MESSAGE_OPERATION_RESULT_MAX_BYTES,
   MESSAGE_QUEUE_PROJECTION_MAX_BYTES,
+  decodeSessionMessageQueueProjection,
   type SessionMessageQueueProjection,
   type TurnSnapshot,
 } from '../protocol/index.js';
@@ -1860,3 +1861,41 @@ function deferred<T>() {
   });
   return { promise, resolve };
 }
+
+test('a lone folded steering entry leaves the queue projection decodable', async () => {
+  // A steer the run never pulls is folded into `followup` at the terminal
+  // transition. It has to arrive there as a followup entry: the wire decoder
+  // requires `next_turn`, and an undecodable projection takes down the whole
+  // Host by way of the session continuity snapshot (#3530).
+  const fixture = createFixture();
+  fixture.coordinator.reserveRootTurn(ROOT);
+  const owner = fixture.coordinator.bindRun(ROOT);
+  const submitted = await submitContent(
+    fixture,
+    'lone-steer',
+    { text: 'late steer', displayText: 'late steer' },
+    'current_turn',
+  );
+  assert.equal(submitted.ok, true, JSON.stringify(submitted));
+
+  owner.release();
+  const batch = fixture.coordinator.beginTerminalTransition(ROOT);
+
+  const projection = fixture.coordinator.projection(ROOT.sessionId);
+  assert.deepEqual(projection.steering, []);
+  assert.equal(projection.followup.length, 1);
+  assert.equal(projection.followup[0]?.placement, 'next_turn');
+  assert.doesNotThrow(() => decodeSessionMessageQueueProjection(projection));
+
+  // Durable provenance is unchanged: the source still records where the
+  // message was aimed, which is what makes the fold auditable.
+  assert.equal(batch.sources.length, 1);
+  assert.equal(batch.sources[0]?.placement, 'current_turn');
+  assert.equal(batch.sources[0]?.disposition, 'steering');
+
+  fixture.coordinator.commitNextRoot(batch, {
+    sessionId: ROOT.sessionId,
+    turnId: 'turn-2',
+    runId: 'run-2',
+  });
+});


Astro-Han (commented):
Reviewed at head `bcdbc99c` with `test` terminal SUCCESS. No P0–P3 from this line.

The framing in the new comment is what makes this the right fix rather than a patch: `placement` on a queue snapshot is queue position, not origin. A steer the run never pulled is, once folded ahead of the followups at the terminal transition, a next-turn message by definition — reporting it as `current_turn` was the projection describing itself in terms the wire decoder rejects, and an undecodable projection takes the Host down through the continuity snapshot. Fixing the reporter rather than relaxing the decoder keeps the invariant where it belongs.

Two things I checked because they are where this class of fix usually goes wrong:

- **Provenance is not lost.** `disposition` still says `steering` and the durable `sourceFromEntry` record still says `current_turn`, so the fold stays auditable. The new test asserts both sides, which is the part that makes the change safe to keep.
- **No consumer reads `followup[].placement` as origin.** The only other place that branches on placement is the `disposition` derivation at `message-coordinator.ts:629`, which reads the *input* placement, not the projected snapshot. So forcing `next_turn` on the followup projection has no second reader to surprise.

Scope is one function and its call site, with no new state. Nothing else to raise.

packages/runtime-host/src/server/message-coordinator.ts
```diff
@@ -1477,6 +1477,19 @@ function queuedSteeringSnapshot(entry: LiveEntry): SteeringMessageSnapshot {
   return { ...queuedSnapshot(entry), placement: 'current_turn' };
 }
 
+/**
+ * Queue position, not origin: an entry in the followup queue is a next-turn
+ * message by definition, including a steering entry the run never pulled and
+ * the terminal transition folded ahead of the followups. Where the message was
+ * originally aimed stays on `disposition` and on the durable
```
Astro-Han, line 1484: **[P3] The doc comment promises a field the returned type does not have.**

> *Where the message was originally aimed stays on `disposition` and on the durable {@link sourceFromEntry} record.*

The durable half is true and the test pins it (`batch.sources[0].disposition === 'steering'`). The `disposition` half is **not true of this snapshot**:

```ts
function queuedSnapshot(entry: LiveEntry): QueuedMessageSnapshot {
  return { entryId, messageId, content, placement, state: 'queued' };  // no disposition
}
```

`QueuedMessageSnapshot extends MessageQueueEntrySnapshotBase`, which is `entryId / messageId / content / placement` — there is no `disposition` on it anywhere in `protocol/message.ts`. So inside the **live projection**, once the fold happens, a steer the run never pulled becomes indistinguishable from an ordinary followup. Origin survives only in the durable source record, which the projection consumer does not see.

**Why I am still calling it P3 and not higher:** nothing today reads that distinction off the projection, and it *cannot* have — the wire decoder rejects `current_turn` in `followup`, so any consumer depending on it would already be crashing. The behavior is correct; the comment describes a stronger guarantee than the code provides.

**Why it is worth fixing anyway:** this comment is doing real work — it is the thing that stops the next person from "fixing" `placement` back to `current_turn` and re-breaking #3530. A comment carrying that much load should not have a false clause in it, because the moment someone checks the clause and finds it wrong, they discount the whole comment. Dropping "on `disposition` and" is enough.

---

**On the change itself — I went in expecting a problem and did not find one.** The title pairs *folded* with *decodable*, and folding is usually lossy while decodability usually wants losslessness, so I checked whether this trades away information to satisfy the decoder. It does not: the projection field being corrected is *queue position*, which genuinely is `next_turn` for anything sitting in the followup queue, and the *origin* is kept where it is actually durable. Naming the snapshot function separately instead of branching inside `queuedSnapshot` also makes the asymmetry with `queuedSteeringSnapshot` visible at the call site.

*Reviewed at 2026-08-23 12:24 UTC.*

Astro-Han (approved):
Approving `bcdbc99c5ae57a93dcc0ff0e5a275a5b1289342b`. Required `test` is `completed / success` bound to that exact SHA (run 32594347705).

**One [P3], inline, non-blocking:** https://github.com/apache/maka/pull/3532#discussion_r3837675463 — the new doc comment says the original aim "stays on `disposition`", but `QueuedMessageSnapshot` has no `disposition` field; only the durable source record keeps it. The code is right, the comment claims more than the code delivers. Worth a one-clause edit whenever you next touch the file.

**The fix is correct and the reasoning in the comment is the valuable part.** Queue position and message origin are two different facts, and the bug was reporting one as the other. An entry sitting in `followup` *is* a next-turn message regardless of where it was originally aimed, and the provenance that makes the fold auditable lives in the durable source record — which the test pins explicitly rather than leaving implied.

The regression test earns its place: it drives a real lone-steer fold through `beginTerminalTransition` and then asserts `decodeSessionMessageQueueProjection` on the result, so it fails on the actual wire contract rather than on a hand-written expectation. Given that an undecodable projection takes the Host down through the session continuity snapshot (#3530), pinning the decode is the assertion that matters.

*Reviewed at 2026-08-23 12:25 UTC. No P0–P2.*

Astro-Han:
LGTM with non-blocking P3.

PR #3534, opened 2026-08-22
test(core): keep timestamp refresh delays within the scheduler bound

Astro-Han:
Fixes #3364 follow-up P3: test name said finite but covered Infinity/NaN. Rename to neutral.\n\nGenerated-by: Muse Spark


jackwener (approved):
Independent look at a38db61. One file, +1/-1: drop `finite` from the test title so it matches the NaN/±Infinity samples. Base is the #3364 merge. Required `test` SUCCESS. Approve. Merge stays with the author.

PR #3537, opened 2026-08-22
fix(cli): ask the terminal for color support instead of inferring it from TERM

### LINKED ISSUE #3536: bug(cli): the TUI emits no ANSI on native Windows shells — an unset TERM is read as colorless before any platform check
## What happened

The TUI renders entirely without color on native Windows shells — no status disc tints, no accent, no red exit codes. The same build on WSL is fully colored. This is not a downgrade to 16 colors; **not a single escape sequence is emitted**.

Cause is the first rule in the detector:

```ts
// packages/cli/src/tui-ansi.ts
const term = env.TERM ?? '';
if (term === 'dumb' || term === '') return 0;
```

PowerShell and cmd.exe do not set `TERM` at all, so an unset `TERM` is read as "colorless" before any other signal is considered. Git Bash and WSL do set it, which is exactly where color still works.

The capability is present — this is a detection bug, not a terminal limitation. Measured on the same machine (Windows 11, 10.0.26200, inside Windows Terminal):

| shell | `TERM` | `tty.getColorDepth` | current level |
|---|---|---|---|
| PowerShell 5.1 | unset | 24 (truecolor) | **0** |
| cmd.exe | unset | 24 (truecolor) | **0** |
| Git Bash | `xterm-256color` | 24 | 2 |
| WSL2 (Ubuntu) | `xterm-256color` | 8 | 2 |

Node reports 24-bit support in the two shells where the TUI emits nothing.

### Second reachable case

`COLORTERM` is ignored whenever `TERM` is unset, because the empty-`TERM` return happens first. A terminal that declares `COLORTERM=truecolor` without setting `TERM` gets level 0 on every platform, not just Windows.

## How to reproduce

1. Open PowerShell or cmd.exe (not Git Bash, not WSL) on Windows.
2. Start the TUI from source.
3. Every tool row, status disc, and diff renders as plain text.

Or, without a terminal, compare the built module under two env shapes:

```
TERM=xterm-256color  →  "[32mX[39m[38;5;75mY[39m…"
COLORTERM=truecolor  →  "[38;2;87;163;239m…"
TERM unset           →  "XY●"          ← no escapes at all
```

## Environment

- Maka commit: `f19eede03`
- OS: Windows 11 (10.0.26200), Windows Terminal
- Surface: TUI / CLI
- Node.js: v22.18.0

### This contradicts the detector's own stated benchmark

The function documents what it is modelled on:

> `Benchmark: codex `supports-color` 3-level; pi `theme.ts` 256 fallback.`

`supports-color` never treats a missing `TERM` as colorless. It returns on platform **before** consulting `TERM` at all:

```js
if (env.TERM === 'dumb') return min;

if (process.platform === 'win32') {
  // Windows 10 build 10586 is the first Windows release that supports 256 colors.
  // Windows 10 build 14931 is the first release that supports 16m/TrueColor.
  const osRelease = os.release().split('.');
  if (Number(osRelease[0]) >= 10 && Number(osRelease[2]) >= 10586) {
    return Number(osRelease[2]) >= 14931 ? 3 : 2;
  }
  return 1;
}
// TERM / COLORTERM heuristics only after this
```

So the gap is a transcription one: the `TERM` ladder was carried over, the platform branch was not, and "`TERM` empty" became an early `return 0` instead of falling through.

Node's built-in `tty.WriteStream.getColorDepth()` implements the same Windows build check, which is why it answers 24 where the hand-rolled ladder answers 0.

### Suggested direction

Keep the two intent overrides (`NO_COLOR`, `TERM=dumb`) and the explicit truecolor upgrade, then ask the terminal instead of re-deriving capability from `TERM`:

1. `NO_COLOR` non-empty → 0
2. `TERM=dumb` → 0
3. `COLORTERM=truecolor|24bit`, or `TERM` ending `-truecolor`/`-24bit` → 3
4. otherwise map `process.stdout.getColorDepth()` — 24 → 3, 8 → 2, 4 → 1, else 0
5. when there is no TTY to ask, fall back to today's `TERM` ladder

This deletes the hand-maintained platform ladder rather than bolting a Windows special case onto it, and leaves Linux/macOS behavior unchanged. Adopting `supports-color` as a direct dependency would also work, but it carries third-party-notice obligations that the built-in does not.

`packages/cli` has no test for the detector today; it is a pure function over an env snapshot and is straightforward to pin.

_Investigated with AI assistance (Claude Code); every number above measured on the machine described._

shaokeyibb:
## Summary

`detectColorLevelFromEnv` returned 0 for an unset `TERM` before considering any other signal. PowerShell and cmd.exe set no `TERM` at all, so the TUI emitted no ANSI whatsoever on native Windows shells while WSL and Git Bash stayed coloured. The same early return also swallowed an explicit `COLORTERM=truecolor` whenever `TERM` was unset — on every platform, not just Windows.

The function's stated benchmark already handles this. `supports-color` returns on `process.platform === 'win32'` (by OS build) *before* it reads `TERM`; the transcription kept the `TERM` ladder and dropped the platform branch. Node's `tty.WriteStream.getColorDepth()` implements the same Windows build check, so this asks the terminal instead of maintaining a second copy of the ladder.

Order is now: `NO_COLOR` → `TERM=dumb` → explicit truecolor declaration → reported depth → (no terminal to ask) the pre-existing `TERM` ladder. Linux and macOS behaviour is unchanged; the fallback keeps piped output exactly as it was.

Fixes #3536

## Verification

End-to-end in a real pty with `TERM` unset — the case the bug is about:

```
before   REAL TTY, TERM unset => "XY●"
after    REAL TTY, TERM unset => "[32mX[39m[38;2;87;163;239mY[39m…"
```

24-bit output where there was previously not a single escape sequence.

Measured capability on this machine (Windows 11 10.0.26200, Windows Terminal), showing the detector disagreeing with the terminal:

| shell | `TERM` | `getColorDepth` | before | after |
|---|---|---|---|---|
| PowerShell 5.1 | unset | 24 | 0 | 3 |
| cmd.exe | unset | 24 | 0 | 3 |
| Git Bash | `xterm-256color` | 24 | 2 | 3 |
| WSL2 | `xterm-256color` | 8 | 2 | 2 |

New `packages/cli/src/__tests__/tui-ansi.test.ts` (5 tests) pins the matrix — there was no test for this function before. Checked against a plausible partial fix that adds the depth probe but keeps the empty-`TERM` early return: 2 of the 5 go red (`expected: 3, actual: 0`), including the `COLORTERM` case.

Ran: the new suite plus every existing consumer of `tui-ansi` (`pi-goal`, `pi-tui-transcript-viewer`, `pi-transcript`) — 84 green; `biome check` on both files; `npm run check:asf-headers`; `tsc -p packages/cli`.

Not run: the rest of the CLI suite, Desktop, Playwright. PowerShell 7 was not available on this machine, so only PowerShell 5.1 and cmd.exe were measured natively.

## Review focus

`getColorDepth` exists only on `tty.WriteStream`, so `process.stdout.getColorDepth?.()` is `undefined` for piped output. That path deliberately falls through to the original `TERM` ladder rather than to 0, which keeps redirected output and the existing test environment byte-identical. Making non-TTY output colourless would be defensible, but it is a separate behaviour change and nothing in `packages/cli` reaches `tui-ansi` from a non-TUI path today.

## AI use

Select exactly one:

Tool(s) and scope: Claude Code — investigation, the measurements above, the test, and the fix.

## Checklist

- [ ] No

### DIFF
--- a/packages/cli/src/__tests__/tui-ansi.test.ts
+++ b/packages/cli/src/__tests__/tui-ansi.test.ts
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import assert from 'node:assert/strict';
+import { describe, test } from 'node:test';
+import { detectColorLevelFromEnv } from '../tui-ansi.js';
+
+/** Node's answer on a Windows 10+ console: truecolor, whatever TERM says. */
+const WINDOWS_CONSOLE_DEPTH = 24;
+/** Node's answer for a 256-color xterm. */
+const XTERM_256_DEPTH = 8;
+/** Node's answer when it finds no evidence of color support. */
+const NO_COLOR_DEPTH = 1;
+
+describe('detectColorLevelFromEnv', () => {
+  test('uses the terminal capability when TERM is unset — native Windows shells', () => {
+    // PowerShell and cmd.exe set no TERM at all. Reading that as "colorless"
+    // turns the whole TUI monochrome on a console that supports truecolor.
+    assert.equal(detectColorLevelFromEnv({}, WINDOWS_CONSOLE_DEPTH), 3);
+  });
+
+  test('honours an explicit COLORTERM even when TERM is unset', () => {
+    assert.equal(detectColorLevelFromEnv({ COLORTERM: 'truecolor' }, NO_COLOR_DEPTH), 3);
+    assert.equal(detectColorLevelFromEnv({ COLORTERM: '24bit' }, NO_COLOR_DEPTH), 3);
+  });
+
+  test('NO_COLOR and TERM=dumb still win over any capability', () => {
+    assert.equal(
+      detectColorLevelFromEnv({ NO_COLOR: '1', TERM: 'xterm-256color' }, WINDOWS_CONSOLE_DEPTH),
+      0,
+    );
+    assert.equal(detectColorLevelFromEnv({ TERM: 'dumb' }, WINDOWS_CONSOLE_DEPTH), 0);
+    // The NO_COLOR spec: an empty value does NOT disable color.
+    assert.equal(
+      detectColorLevelFromEnv({ NO_COLOR: '', TERM: 'xterm-256color' }, XTERM_256_DEPTH),
+      2,
+    );
+  });
+
+  test('maps the reported depth onto the three color levels', () => {
+    assert.equal(detectColorLevelFromEnv({ TERM: 'xterm-256color' }, XTERM_256_DEPTH), 2);
+    assert.equal(detectColorLevelFromEnv({ TERM: 'xterm' }, 4), 1);
+    assert.equal(detectColorLevelFromEnv({ TERM: 'xterm' }, NO_COLOR_DEPTH), 0);
+  });
+
+  test('falls back to the TERM ladder when there is no terminal to ask', () => {
+    // Piped output has no tty.WriteStream, so no depth is available. The
+    // pre-existing env ladder still answers, unchanged.
+    assert.equal(detectColorLevelFromEnv({ TERM: 'xterm-256color' }, undefined), 2);
+    assert.equal(detectColorLevelFromEnv({ TERM: 'xterm' }, undefined), 1);
+    assert.equal(detectColorLevelFromEnv({ TERM: 'xterm-truecolor' }, undefined), 3);
+    assert.equal(detectColorLevelFromEnv({}, undefined), 0);
+  });
+});
--- a/packages/cli/src/tui-ansi.ts
+++ b/packages/cli/src/tui-ansi.ts
@@ -91,38 +91,68 @@ function buildAnsi() {
 }
 
 function detectColorLevel(): 0 | 1 | 2 | 3 {
-  return detectColorLevelFromEnv({
-    NO_COLOR: process.env.NO_COLOR,
-    TERM: process.env.TERM,
-    COLORTERM: process.env.COLORTERM,
-  });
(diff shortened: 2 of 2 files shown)


Astro-Han:
Reviewed at `287915d3`. No blocking findings. One naming nit and a note about CI.

The problem is real and the diagnosis is precise: PowerShell and `cmd.exe` set no `TERM` at all, and the old ladder's `if (term === 'dumb' || term === '') return 0;` turned that into "colorless" before anything else got a chance — so the TUI went monochrome on a console that supports 24-bit, while WSL and Git Bash stayed coloured. That is an ordinary path for every native-Windows user, not a constructed one.

What I want to call out, because it is the part most likely to be undervalued in a diff this size: **this change removes a duplicated capability ladder rather than adding a special case.** The obvious fix here is `if (process.platform === 'win32') return 3;` — a second Windows-specific branch that then has to be kept in sync with Node's own build checks forever. Delegating to `getColorDepth()` deletes that future maintenance instead of creating it, and the comment says so explicitly. Net effect on the ladder is one fewer thing to keep correct.

The precedence order also survives scrutiny. `NO_COLOR` (non-empty, per spec — the empty-string case is correctly excluded) and `TERM=dumb` still outrank everything; an explicit `COLORTERM` and a `-truecolor` / `-24bit` suffix outrank the probe, on the stated reasoning that a declaration is something the terminal knows and a probe may not; the depth mapping only runs when a depth exists; and `term === ''` still returns 0 on the no-tty path, which keeps piped output unchanged. `process.stdout.getColorDepth?.()` correctly yields `undefined` for a non-tty stdout rather than throwing.

The tests pin each of those branches, including the two that are easy to get backwards — `NO_COLOR=` empty not disabling color, and the piped-output fallback still using the old ladder.

**[P3] The function name no longer describes what it does.**

`detectColorLevelFromEnv(env, depth)` now decides from an env snapshot *and* a terminal capability probe, and in the common case the probe is what actually decides. Something like `detectColorLevel(env, depth)` would read truer. Purely cosmetic, and the doc comment does explain the second parameter — flagging it only because the `FromEnv` suffix is exactly the kind of stale name that later leads someone to pass `undefined` for `depth` thinking it is optional detail.

**On CI:** no checks have been reported on this head, so there is no CI evidence for it yet. Nothing looks wrong with the branch; it just has not run. Worth a maintainer kicking it off.

One thing I could not verify from here: whether any terminal that sets `TERM=xterm-256color` reports a `getColorDepth()` below 8, which would now downgrade it from 256 to 16. Node derives its depth partly from `TERM` itself, so I do not expect it, but I do not have a Windows or exotic-terminal environment to confirm against — noting it as a boundary rather than a finding.

Astro-Han (approved):
CI is now terminal green on `287915d3d3ad0c3e3197466b5b285f98a2a9e6d1` (`test`: completed / success), which matches the exact head this review is bound to.

Re-confirming the earlier review conclusion on this head: querying the terminal for color support instead of inferring it from `TERM` removes a heuristic rather than adding a layer on top of one — the `TERM` string stops being a second authority on capability. Scope is small and the change is contained.

Approving.

PR #3543, opened 2026-08-23
feat(desktop): preserve bounded main-process recovery evidence

### LINKED ISSUE #3491: discussion(desktop): define bounded next-launch evidence for hard main-process crashes
<details open>

## Decision to make

Maka's main-process diagnostic log is deliberately bounded and redacted, but it exists only in memory. Native diagnostic dialogs also require Electron main to remain alive. A hard main-process failure such as the V8 OOM observed in #3458, a native crash, or an external kill can therefore remove both the evidence and every live copy surface at once.

This is a child of #3463. It is a design boundary before implementation: writing more data during ordinary operation creates a new persistence and privacy owner, while a crash-time JavaScript handler cannot reliably cover OOM or native termination.

## Required decision

Define the smallest supported next-launch evidence contract for hard main-process failures:

- Which failure classes are supportable and distinguishable: JavaScript uncaught failure, V8 OOM, native crash, OS kill, or power loss.
- Which existing platform signal can establish that a prior run ended abnormally without pretending every unclean shutdown was a Maka defect.
- Whether the minimum evidence is a small redacted journal, a local Electron crash artifact, or a narrower platform-owned reference.
- One owner, one bounded representation, one retention/expiry rule, and one safe deletion point.
- How the next launch offers a user-controlled **Copy Diagnostics** action without automatic upload or exposing raw crash artifacts.
- How secrets and home paths are removed before persistence, and how permissions and symlink/replacement safety are enforced.

## Constraints

- Do not implement a crash-reporting backend, telemetry, automatic upload, or indefinite log archive.
- Do not continuously persist the current 256 KiB in-memory buffer by default without an explicit privacy, durability, and performance decision.
- Do not claim that a last-run marker proves a crash unless the selected platform signal supports that conclusion.
- Do not solve the root cause of #3458 here; this issue only defines evidence survivability.
- Keep Renderer-process loss and Runtime Host exit evidence in their own owners.

## Exit criteria

- A short design record maps supported failure classes to obtainable evidence on macOS, Windows, and Linux.
- The proposed representation has explicit byte limits, redaction timing, permissions, retention, compatibility, and corruption behavior.
- The next-launch user flow and deletion semantics are specified.
- The implementation is split further if different platforms require materially different owners or guarantees.

## AI use

OpenAI Codex assisted with the bounded investigation and issue drafting. The human contributor remains responsible for scope and implementation decisions.

</details>

<details>

## 待决策事项

Maka 的 main-process 诊断日志已经有界并经过脱敏，但只存在内存中；原生诊断弹窗也要求 Electron main 仍然存活。#3458 中观察到的 V8 OOM、原生崩溃或外部终止等 main process 硬故障，会同时移除证据和所有实时复制入口。

这是 #3463 的 child issue，先确定设计边界再实现：普通运行期间持续写入会新增持久化与隐私 owner，而 crash-time JavaScript handler 无法可靠覆盖 OOM 或原生终止。

## 必须明确的设计

定义 main process 硬故障在下次启动时可恢复的最小证据契约：

- 哪些故障类型能够被支持并区分：JavaScript 未捕获异常、V8 OOM、原生崩溃、OS kill 或断电。
- 哪个已有平台信号能够证明上一次运行异常结束，且不会把所有非正常关闭都伪装成 Maka 崩溃。
- 最小证据应是小型脱敏 journal、本地 Electron crash artifact，还是更窄的平台 owner 引用。
- 唯一 owner、唯一有界表示、唯一保留/过期规则和安全删除时点。
- 下次启动如何提供由用户主动触发的 **复制诊断信息**，而不自动上传或暴露原始 crash artifact。
- 敏感信息和 home path 如何在持久化前移除，以及文件权限、symlink/replacement 安全如何保证。

## 约束

- 不实现崩溃上报后端、遥测、自动上传或无限期日志归档。
- 在没有明确隐私、durability 和性能决策前，不默认持续落盘当前 256 KiB 内存日志。
- 除非所选平台信号能够证明，否则不能宣称 last-run marker 等同于崩溃。
- 不在这里解决 #3458 的根因；本 issue 只定义证据存活。
- Renderer process 消失和 Runtime Host 退出证据继续由各自 owner 处理。

## 完成标准

- 一份简短设计记录将 macOS、Windows、Linux 上支持的故障类型映射到可获得证据。
- 方案明确字节上限、脱敏时点、权限、保留周期、兼容和损坏处理。
- 明确下次启动的用户流程与删除语义。
- 如果不同平台需要实质不同的 owner 或保证，继续拆分实现任务。

## AI 使用

OpenAI Codex 协助了有界调查和 issue 起草；范围与实现决策仍由贡献者本人负责。

</details>

M4n5ter:
## Summary

<details open>

- Preserve one local, redacted main-process log tail so the next launch can offer **Copy Diagnostics** before the Renderer or Runtime Host starts.
- Treat a missing clean-shutdown marker only as an interrupted prior run. An uncaught JavaScript failure gets a best-effort immediate snapshot; V8 OOM, native crashes, OS kills, and power loss remain indistinguishable and use the latest periodic snapshot on macOS, Windows, and Linux.
- Keep the persistence contract bounded: first dirty snapshot after 2 seconds, at most one dirty rewrite per 5 minutes, 256 KiB of newest logs plus 16 KiB metadata allowance, one pending run, and seven-day expiry from when an interrupted run is discovered. This caps the continuous worst case at about 72 MiB of logical writes per day.
- Redact secrets in the existing log buffer and collapse home paths before persistence. Store the journal under an owner-only directory/file boundary, atomically replace it, reject symlink/non-file reads, discard corrupt or incompatible records, and delete pending evidence after the recovery prompt finishes.
- Mark normal application quits and Windows session-end as clean. The recovery dialog never uploads data and continues startup even if journal or clipboard handling fails.

</details>

<details>

- 在本地保留一份经过脱敏的 main process 最新日志尾部，让下次启动能在 Renderer 和 Runtime Host 启动前提供**复制诊断信息**。
- 缺少 clean-shutdown 标记只表示上一次运行未完成正常退出。JavaScript 未捕获异常会尽力立即刷新快照；V8 OOM、原生崩溃、OS kill 与断电仍不可区分，并在 macOS、Windows 和 Linux 上共同使用最近一次周期快照。
- 持久化契约保持有界：首次 dirty 后 2 秒写入、持续产生日志时最多每 5 分钟重写一次、最新日志上限 256 KiB、元数据预留 16 KiB、仅保留一次 pending run，并从发现上一次运行中断时起保留七天。持续最坏情况下的逻辑写入量约为每天 72 MiB。
- 复用现有日志缓冲区的敏感信息脱敏，并在持久化前折叠 home path。journal 使用仅 owner 可访问的目录和文件、原子替换、拒绝 symlink/非普通文件、丢弃损坏或不兼容记录，并在恢复弹窗结束后删除 pending evidence。
- 正常应用退出和 Windows session-end 会标记为 clean。恢复弹窗不会上传数据，journal 或剪贴板失败也不会阻塞启动。

</details>

Fixes #3491

## Verification

- `npm run lint`
- `npm run format:check`
- `npm run build`
- `npm run typecheck`
- Desktop main test suite: 1,195 passed
- Not run manually: physical Windows shutdown/session-end flow

## AI use

Select exactly one:

Tool(s) and scope: OpenAI Codex assisted with design analysis, implementation, tests, and review. The human contributor remains responsible for the contribution.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/src/main/__tests__/main-process-recovery-journal.test.ts
+++ b/apps/desktop/src/main/__tests__/main-process-recovery-journal.test.ts
@@ -31,6 +31,7 @@ import {
   MAIN_PROCESS_RECOVERY_FLUSH_INTERVAL_MS,
   MAIN_PROCESS_RECOVERY_LOG_MAX_BYTES,
   MAIN_PROCESS_RECOVERY_MAX_AGE_MS,
+  presentPendingMainProcessRecovery,
 } from '../main-process-recovery-journal.js';
 
 test('recovers one bounded redacted snapshot after an unclean exit', async () => {
@@ -158,19 +159,38 @@ test('persists only the bounded newest log tail', async () => {
   }
 });
 
-test('expires stale evidence and ignores a corrupt pending record', async () => {
+test('retains a newly discovered interruption and expires pending evidence after seven days', async () => {
   const directory = await mkdtemp(join(tmpdir(), 'maka-main-recovery-expiry-'));
-  const errors: unknown[] = [];
   try {
     createJournal(directory, new DiagnosticLogBuffer(), new Date('2026-08-01T00:00:00Z'));
+    const discovered = createJournal(
+      directory,
+      new DiagnosticLogBuffer(),
+      new Date('2026-08-20T00:00:00Z'),
+    );
+    assert.equal(discovered.pending?.run.startedAt, '2026-08-01T00:00:00.000Z');
+    discovered.markClean();
+
     const afterExpiry = createJournal(
       directory,
       new DiagnosticLogBuffer(),
-      new Date(new Date('2026-08-01T00:00:00Z').getTime() + MAIN_PROCESS_RECOVERY_MAX_AGE_MS + 1),
+      new Date(
+        new Date('2026-08-20T00:00:00Z').getTime() +
+          MAIN_PROCESS_RECOVERY_MAX_AGE_MS +
+          2_000,
+      ),
     );
     assert.equal(afterExpiry.pending, undefined);
     afterExpiry.markClean();
+  } finally {
+    await rm(directory, { recursive: true, force: true });
+  }
+});
 
+test('ignores corrupt pending records without amplifying invalid log arrays', async () => {
+  const directory = await mkdtemp(join(tmpdir(), 'maka-main-recovery-corrupt-'));
+  const errors: unknown[] = [];
+  try {
     writeFileSync(join(directory, 'pending.json'), '{not json', { mode: 0o600 });
     const afterCorruption = createJournal(
       directory,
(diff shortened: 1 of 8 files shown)


Astro-Han (approved):
Approving `ff576a384d16056efd22ac92565f1fd12c8aa5fe`. Both required checks are `completed / success` bound to that exact SHA. **No P0–P3.**

An independent line reviewed this with one question up front — **"the PR title says *bounded*, so where is the bound, who enforces it, and what happens at the edge"** — and produced the enforcement table before looking for defects. I then re-derived the load-bearing pieces myself at the gate rather than taking them on report. All four held:

- **`boundedLogTail` keeps the newest, not the oldest.** It walks `logs` from the last index backwards, accumulating until the next entry would cross `MAIN_PROCESS_RECOVERY_LOG_MAX_BYTES`, then reverses. **This is the one that had to be right** — the last few lines before a crash are the whole point of the artifact, and a naive `slice(0, n)` would have thrown away exactly the evidence worth keeping. The test pinning `entry 399` present and `entry 0` absent is the right assertion.
- **The write path is genuinely atomic and hostile-input aware**: `writeFileSync` with `flag: 'wx'` and `mode: 0o600` to a temp path, then `renameSync`, with the temp removed in `finally`. Reads go through `lstatSync` rejecting symlinks **and** `O_NOFOLLOW`, so a symlink planted at the evidence path is refused twice rather than followed.
- **`RUN_MAX_BYTES` is headroom, not an unenforced limit.** It only contributes to `EVIDENCE_MAX_BYTES`, and I checked why that is safe: every field on `recoveryRunSchema` is a `boundedStringSchema` capped at 1 KiB, so the run record cannot approach 16 KiB by construction. Worth knowing the reason, since "a constant that is never checked directly" usually *is* a finding.
- **The 7-day clock starts at discovery, not at the crash** — `utimesSync(activePath, promotedAt, promotedAt)` immediately before the rename to pending. That matches what the PR says, and it is the right choice: a crash that happened while the machine was off should not expire before anyone has seen it.

**On the single pending slot:** a newer crash overwrites an unviewed pending record. That is a deliberate design choice, not an oversight, and it is the correct one here — the alternative is unbounded growth in exactly the situation where the app is already failing repeatedly. Worth keeping in mind if crash-loop diagnosis ever becomes a goal, since only the most recent loop iteration survives.

**On scope:** this is one journal, not a second crash-collection system, and the `-30` is mostly `app.exit(0)` → `app.quit()` so that a normal quit can mark itself clean. A failed write disables the journal for the remainder of the process rather than retrying into a broken disk, and `appendUncaughtMainProcessError` swallows its own errors so the evidence path cannot become the thing that kills the process. **Diagnostic code that can take down the process it is diagnosing is the classic failure of this feature, and it is avoided here.**

Seven distinct failure hypotheses were raised and each withdrawn against specific evidence — wrong-end truncation, write failure as a new crash source, torn JSON or symlink substitution, double-recording and replay, the age bound interacting with the slot, `session-end` listener target, and post-failure disabling. **No P3s were manufactured to look thorough**, which I appreciate more than a padded list.

**Coverage disclosure:** one independent review line plus my own verification at the gate. For a change of this size I would normally want a second independent line; I am approving on one because the line's claims were reproducible and I checked the load-bearing ones directly. Stated so the record is accurate rather than flattering.

*Reviewed at 2026-08-23 12:50 UTC.*

PR #3544, opened 2026-08-23
feat(desktop): per-entry composer queue with promote, retract, and reorder

Astro-Han:
## Summary

Refs #3513

#3513 put a labeled Queue/Steer `SegmentedControl` in the composer's send slot. That chrome is heavier than the problem it solves: the composer gains a mode the user must manage, and the send button morphs between three glyphs.

This PR replaces it with a Codex-style pending plate above the composer card:

- Mid-turn **Send is always Send** and queues a follow-up. No mode switch anywhere; the segmented control is gone. `Shift+Enter` steers the draft into the active Turn once.
- Queued follow-ups render in a hairline plate (radius matches the composer card) in send order, first at the top. Per entry:
  - hover grip to drag-reorder,
  - 立即发送 promotes the entry into the active Turn (it leaves the plate),
  - 收回草稿 restores the entry into the composer draft.
- Steering entries never appear in the plate: once handed to the Turn they surface in the transcript when consumed.

Runtime Host remains the sole queue authority via three new receipted operations — `queue.entry.promote`, `queue.entry.retract`, `queue.entries.reorder`. Each is epoch-local like `queue.retract`, idempotent through the message receipt store (extended with `promote` / `reorder` / `retract_entry` kinds), and rejected with `operation_conflict` while the queue drains into the next root Turn, so a terminal transition keeps its exact follow-up prefix. Reorder takes the full intended follow-up order, so a stale drag conflicts instead of clobbering.

The Send/Stop slot gains a comment pinning it as the app's global anchor after repeated regressions.

## Screenshots

Composer mid-turn with queued messages, before and after:

image

## Verification

- `packages/runtime-host`: `message-coordinator.test.ts` (36 tests, incl. 6 new for the three operations: success paths, receipt replay, stale-target and draining conflicts) and `protocol.test.ts` (41 tests, incl. frame decode for the new operations) — pass.
- `packages/ui`: `composer-send-toggle.test.tsx` (6 tests: single Send/Stop control, no mode switch, plate renders per-entry actions, steering/in-flight stay out) — pass.
- `apps/desktop`: `runtime-host-session-execution-ipc-main.test.ts` (12 tests, incl. per-entry IPC routing) plus `runtime-host-client-operations` and `message-queue-ui-state` — pass; preload/main/renderer/storybook typecheck clean; biome format and lint clean.
- Storybook (`Product/Composer Message Queue`): exercised in a real browser via Playwright — three uniform rows, hover reveals the grip, drag reorders, 立即发送 and 收回草稿 remove the row.

Not run: full-repository suites and the desktop E2E harness (left to CI).

## AI use

Tool(s) and scope: Grok Build — implementation and tests from an interaction design reviewed with the contributor; the contributor owns the design decisions (no mode switch, plate chrome, per-entry operations) and the final review.

## Checklist


Astro-Han (commented):
Reviewed exact head `86031a2edb167bae58fa9c8313177adb7da2a277` — required `test` is `completed / success` bound to that SHA. **No P0–P2.** One P3 below.

This PR is authored under our own shared account, so this is posted as a **comment, not an approval** — we are the responsible party here and cannot serve as our own independent review.

**The core decision is the right one.** The previous model only offered all-or-nothing queue retraction plus a global Queue-Steer control, which cannot express "cancel that one", "run that one next", or "reorder these". This moves all three into `HostMessageCoordinator` as a single in-memory authority: `queue.entry.retract`, `queue.entry.promote` and `queue.entries.reorder` share one session admission path, one operation-identity receipt, and one `queueRevision`. The pending plate renders a Host-authored projection rather than a renderer-side queue, and the optimistic UI reorder falls back to the Host projection on failure. There is no second source of truth and no second reorder implementation — which for a queue mutation surface is most of the battle.

**I checked the conflict boundaries against the source rather than the description**, because that is where this class of change usually leaks:
- `promote` distinguishes the three not-in-followup cases explicitly — already steering, already in flight, or genuinely absent — and returns `operation_conflict` for the first two rather than silently succeeding. In-flight work cannot be promoted out from under the delivery path.
- `reorder` refuses when the queue is draining into the next Turn, requires the submitted `entryIds` to be exactly the current follow-up set (length equality plus per-id membership, so a duplicate id fails as a missing one), and only mutates when the order actually differs.

---

**[P3] Two concurrent reorders of an unchanged queue are last-writer-wins.**

`#reorderQueuedEntriesAdmitted` uses "the submitted id set exactly equals the current follow-up set" as its optimistic lock. That correctly rejects a reorder racing an enqueue, a retract, or a promote — all of which change the set. It does **not** reject a reorder racing *another reorder*, because a pure reorder leaves the set identical: the second request passes the check and overwrites the first client's ordering with an order computed from a queue state it never saw.

The blast radius is small — same user, two surfaces, and the loser's intent is simply lost rather than corrupted — which is why this is a P3 and not higher. But the fix is already available: `queueRevision` exists and is returned by every mutation, so accepting an expected revision on the reorder input and rejecting on mismatch would close it with the machinery this PR already built.

Worth noting explicitly since "reorder rejects when the queue changed" is the property the code reads as guaranteeing, and for reorder-vs-reorder it does not.

---
*AI-assisted review. Each claim above was verified against the source at this exact head. Under `CONTRIBUTING.md` §Review this does not count as the required independent human review.*

apps/desktop/src/renderer/app-shell.tsx
```diff
@@ -2205,17 +2203,9 @@ function AppShellContent({
...
-      setMessageQueueBySession((current) => {
-        if (!(sessionId in current)) return current;
-        const next = { ...current };
-        delete next[sessionId];
-        return next;
-      });
+  async function retractQueuedEntry(entryId: string): Promise<void> {
+    await runQueueEntryAction(async (sessionId) => {
+      const content = await window.maka.sessions.retractQueueEntry(sessionId, entryId);
       if (activeIdRef.current !== sessionId) return;
```
jackwener: [P2][normal user path] Preserve the retracted entry even if the active session changes while IPC is in flight. The Host has already removed the queue entry when `retractQueueEntry()` resolves, but this guard returns before `appendDraft(sessionId, displayText)` and before restoring references, attachments, and quotes. Because drafts are already keyed by `sessionId`, switching A→B during the await can make A's queued content disappear permanently: it is no longer in Host followup and is never written back to A's draft. Please remove the active-session early return for the session-keyed restoration path (keeping focus/current-composer UI effects conditional as needed) and add a production-seam test that retracts in A, switches to B before resolution, then verifies A's full text/references/attachments/quotes are recoverable.

jackwener (dismissed):
Independent review of exact head `86031a2edb167bae58fa9c8313177adb7da2a277`, performed in reviewer isolation before reading existing review conclusions. Acting on behalf of @WAWQAQ under the standing Kabi review authorization; this records the authorization source, scope, and exact head.

**Conclusion: APPROVE with one non-blocking P2; no P0/P1.** The Host remains the single queue authority, per-entry promote/retract/reorder share the guarded mutation path, and the renderer's pending order is only an optimistic projection. Removing the user-facing mode switch reduces UI authority rather than adding another queue state machine.

One P2 is posted inline: a successful retract can permanently lose the entry content if the user switches sessions before the IPC resolves. The existing concurrent-reorder P3 from review `5001870026` was independently confirmed after provisional sealing: set equality cannot detect reorder-vs-reorder; an expected `queueRevision` would close it. Neither finding is merge-blocking under the current P0–P3 policy.

Fresh gate: OPEN, non-draft, MERGEABLE/BLOCKED, exact-head hosted `test` is `completed/success`, and no live P0/P1 hold applies. This Agent-executed review does not claim to satisfy the unresolved ASF required-human-review wording.

jackwener (commented):
**Severity correction on exact head `86031a2edb167bae58fa9c8313177adb7da2a277`: the active-session retract race is P1, not P2.**

My approval review `5001903475` has been dismissed. The underlying inline finding remains correct, but I misapplied our own severity policy.

This is a normal supported user path: retract a queued follow-up in session A, then switch to B before the IPC resolves. When the promise returns, Host has already removed the only queued copy. The renderer then exits at `activeIdRef.current !== sessionId` before restoring A's session-keyed text, references, attachments, and quotes. The content is permanently lost and there is no recovery path.

Normal-path, unrecoverable loss of user-authored content is merge-blocking under Kabi's P0–P3 policy. The prior P2 label incorrectly treated the blast radius as more important than recoverability. This head is therefore **BLOCKED pending the session-switch race fix and a new exact-head review**. The concurrent reorder finding remains P3.

Acting on behalf of @WAWQAQ under the standing Kabi review authorization; correction bound to the exact head above. This Agent-executed review does not claim to satisfy the unresolved ASF required-human-review wording.

jackwener (approved):
Independent incremental review of exact head `8dfd0c2de79d3ee25d89b3954644625f90539db7`, performed in reviewer isolation before reading existing review conclusions. Acting on behalf of @WAWQAQ under the standing Kabi review/action authorization; Kabi holds the publisher seat for this head.

**Conclusion: APPROVE. No P0–P2; one existing non-blocking P3 remains.**

The prior P1 is closed. Retraction now captures the Host-authored entry projection at click time, waits only for Host removal confirmation, and restores text, attachments, quotes, and workspace references into the owning session's keyed draft even if the active session changes. Focus remains conditional on that session still being active. The new production-seam regression test covers A→B switching while retract IPC is in flight and verifies that only A is restored.

The three incremental commits also reduce duplicate authority: queue mutation results now carry only `queueRevision`, and the renderer no longer maintains a second local order overlay. The existing P3 remains: reorder admission still compares only the entry-id set, so concurrent reorder-vs-reorder requests with the same set are last-writer-wins. Accepting an expected `queueRevision` would close it; the impact remains limited to one competing reorder intent being overwritten.

Fresh gate immediately before publication: PR is OPEN, non-draft, MERGEABLE/BLOCKED; exact-head hosted `test` is `completed/success`; no current P0/P1 hold applies. This Agent-executed review records an authorized review action but does not claim to satisfy the unresolved ASF required-human-review wording.

PR #3545, opened 2026-08-23
fix: unify context compaction

### LINKED ISSUE #3542: Manual context compaction skips a single long turn
## What happened

Manual Context Compaction does nothing when the completed Session history contains only one user Turn, even if that Turn contains a long Agent Loop with many completed assistant and tool steps.

The result is `unchanged` with reason `insufficient_turns`. The summarizer is never called.

A Turn can legitimately contain many Agent Loop steps, so Turn count should not decide whether its completed history can be compacted.

## How to reproduce

1. Start a Session with one user Turn.
2. Let the Agent Loop produce a substantial number of completed assistant and tool steps.
3. Wait for the Turn to finish.
4. Run manual Context Compaction.
5. Observe `insufficient_turns` and no new checkpoint.

With the same RuntimeEvents, `minRecentTurns: 1` reproduces the skip. Setting it to `0` selects a replaceable prefix and produces a checkpoint draft.

## Environment

- Maka commit: `f19eede03a829c226be39f001b2684860a3aa1d5`
- OS: macOS
- Surface: Desktop / Runtime Host

The guard is in `packages/runtime/src/history-compact.ts`: histories with at most one Turn are rejected unless `minRecentTurns === 0`.

Manual compaction defaults this value to `1`. Oversized-turn recovery already overrides it to `0`.

The fix should also remove the duplicated manual policy construction. Manual, pre-turn, mid-turn, and overflow compaction should use the same checkpoint path, with trigger-specific retention expressed as explicit input. The change should cut over atomically and remove the superseded production path in the same PR.

The separate, default-off `semanticCompact` experiment is out of scope and is being discussed separately in #3541.

Drafted with Codex. I reproduced the behavior and reviewed the report.

### LINKED ISSUE #3550: refactor: remove semanticCompact
## Problem

`semanticCompact` is an off-by-default attention experiment with its own policy, controller state, validation, telemetry, and Agent Loop integration.

`historyCompact` now covers manual, pre-turn, mid-turn, and overflow compaction. Discussion #3541 confirmed that we have no current use case or controlled evaluation showing that the separate `semanticCompact` path improves task outcomes enough to justify another mechanism.

## Desired outcome

- Remove the `semanticCompact` production path, configuration, protocol fields, state, telemetry, tests, and documentation.
- Keep `historyCompact` as the only Runtime context-compaction mechanism.
- Do not retain dormant flags or a compatibility path.
- Preserve the canonical RuntimeEvent ledger so the experiment can be reconsidered later if evidence emerges.

Implementation will be included in #3545 so the repository moves directly to the single-path final state.

Drafted with Codex. I reviewed the repository context and own the proposal.

Astro-Han:
## Summary

- Make the ordered RuntimeEvent ledger the canonical history and the latest compatible durable checkpoint its only compaction projection. Compaction never rewrites or deletes the ledger.
- Separate trigger policy from the compaction transaction. Manual compact, automatic capacity detection, and provider-overflow recovery emit the same Compact command; once emitted, the compactor selects a safe completed prefix and compacts it directly.
- Treat a completed Turn as the atomic retention boundary regardless of how many Agent Loop steps it contains. There is no `minRecentTurns`, `force`, planner-side high-water gate, or parallel skip state.
- Make context management a fixed Runtime-owned policy derived from model facts. No `MAKA_CONTEXT_*` environment variable can tune or disable compaction or Tool Result pruning.
- Return one durable typed outcome (`compacted`, `unchanged`, or `failed`) through TurnSnapshot, Runtime Host, Desktop, and CLI. Desktop keeps an asynchronous compaction visibly in progress until that Runtime terminal outcome arrives. A legacy completed compaction without an outcome becomes `failed / missing_durable_outcome` without draining the Host.
- Keep request-only shaping outside canonical compaction: session recap selects the newest whole turns within its own model budget, and Active Tool Result Prune remains a deterministic non-LLM rewrite.
- Keep archived tool results readable while their verbatim placeholders remain in the model projection, including deterministic ref recovery for legacy placeholders. Checkpoints intentionally do not retain archive refs; the complete original result remains in the RuntimeEvent ledger.
- Delete Semantic Compact, the legacy V1 block/artifact path, synthesis cache, RuntimeEvent history search, eager archive hydration, the history rewrite gate, duplicate planners, flat compaction telemetry, and the context-policy environment parsing surface. Historical usage keys remain accepted only by one inert decode-only whitelist.

Fixes #3542

Fixes #3550

## Context compaction flow

```mermaid
flowchart TD
  Ledger[("Canonical ordered RuntimeEvent ledger")]
  Checkpoint[("Latest compatible durable checkpoint")]

  subgraph TriggerPolicy["Trigger policy — ends after emitting one command"]
    Manual["Manual context.compact"]
    Automatic["Automatic capacity evaluator"]
    Overflow["Provider-overflow recovery"]
    Command["Compact command"]
    Manual --> Command
    Automatic --> Command
    Overflow --> Command
  end

  subgraph Transaction["Single compaction transaction"]
    Read["Read ledger + checkpoint"]
    Covered{"Checkpoint already covers<br/>all compactable history?"}
    Select["Select one safe completed prefix<br/>keep partial events, live head, tool pairs, and verbatim tail"]
    Safe{"Safe prefix exists?"}
    Summarize["Generate one rolling summary<br/>previous checkpoint + newly covered events"]
    Smaller["Back down to the previous safe prefix"]
    Validate["Validate structure, coverage,<br/>lineage, provider state, and replay fit"]
    Persist["Atomically append one bounded checkpoint"]

    Read --> Covered
    Covered -- Yes --> Unchanged["unchanged<br/>already_compacted"]
    Covered -- No --> Select
    Select --> Safe
    Safe -- No --> Failed["failed<br/>no durable write"]
    Safe -- Yes --> Summarize
    Summarize -- input too large --> Smaller --> Summarize
    Summarize -- provider or summarizer failure --> Failed
    Summarize -- candidate --> Validate
    Validate -- invalid --> Failed
    Validate -- valid --> Persist --> Compacted["compacted<br/>checkpointId"]
  end

  Command --> Read
  Ledger --> Read
  Checkpoint --> Read

  Unchanged --> Terminal
  Failed --> Terminal
  Compacted --> Terminal["Durable terminal outcome on TurnSnapshot"]
  Terminal --> Host["Runtime Host protocol"]
  Host --> Desktop["Desktop presents in-progress, success, unchanged, or failure"]
  Host --> CLI["CLI presents the same typed outcome"]

  subgraph RequestProjection["Request-only projections — no second history authority"]
    Replay["Main model replay"]
    LegacyRef["Recover missing ArchiveRead ref<br/>for legacy verbatim placeholders"]
    Prune["Active Tool Result Prune"]
    Recap["Session recap: newest whole turns<br/>within recap model budget"]
    Replay --> LegacyRef --> Prune
  end

  Ledger -.-> Replay
  Checkpoint -.-> Replay
  Ledger -.-> Recap
```

The RuntimeEvent ledger is the only durable history authority. Trigger evaluators decide only when to emit a Compact command; they do not participate in planning. The compactor then performs one transaction: select a safe completed prefix, generate and validate one replacement, and either append one checkpoint or leave durable state unchanged. `unchanged` is an idempotent command result, not a policy state.

Main-model replay and session recap are projections of that authority, not alternate compaction paths. Legacy archive refs are reconstructed only while the original placeholder is still present in verbatim replay. Runtime does not eagerly hydrate archived bodies; the model reads bounded details on demand through `ArchiveRead`. Once a checkpoint covers the placeholder, the model relies on the summary; the original Tool Result remains preserved in the canonical ledger without creating permanent cross-checkpoint retrieval state.

## Intentional policy removals

The Runtime no longer reads any `MAKA_CONTEXT_*` environment variable. Model metadata and fixed Runtime invariants are the single authority for context capacity, compaction reserve, history compaction, and active/stale Tool Result pruning. Removed variables are not mapped, warned, or retained as deprecated states because doing so would preserve a second policy system.

## Verification

- Rebased onto `origin/main` at `96beb414c`.
- Passed 410 focused Runtime compaction, checkpoint, backend, overflow, context-budget, recap, archive-replay, and Tool Result Prune tests.
- Passed Runtime, Runtime Host, and Desktop typechecks.
- Passed 9 focused exact-head Desktop lifecycle and oversized-recap regression tests.
- Passed repository format check and diff whitespace validation.
- Exact-head GitHub `test` and `audit` checks will run on `93581a7cb`.

## Visual verification

- Added deterministic Storybook states under `Product/Context Compaction`: `Running`, `Compacted`, `Unchanged`, and `Failed`.
- The stories use the production Desktop presenter, Toast provider, locale context, and copy rather than a parallel demo component.
- Passed the Desktop Storybook production build and story typecheck; captured all four states at 1280 × 720 for visual review.

## AI use

Select exactly one:

Tool(s) and scope: Codex implemented the Runtime, protocol, CLI, and Desktop changes; removed superseded paths and policy states; updated the architecture and release notes; and added and ran focused tests. The human contributor will review the final diff and owns submission and merge.

## Checklist

- [ ] No


packages/runtime/src/context-budget-policy.ts
```diff
@@ -37,89 +37,36 @@ export function buildDefaultContextBudgetPolicy(
   if (env.MAKA_CONTEXT_BUDGET === 'off') return undefined;
   const contextWindow = resolveSelectedModelContextWindow(connection, options.modelId);
   const reserveTokens = defaultCompactReserveTokens(env, contextWindow);
-  const maxHistoryEstimatedTokens =
-    parseOptionalPositiveInt(env.MAKA_CONTEXT_HISTORY_BUDGET_TOKENS) ??
-    defaultHistoryBudgetTokens(connection, contextWindow, reserveTokens);
-  const maxHistoryTurns = parseOptionalPositiveInt(env.MAKA_CONTEXT_HISTORY_BUDGET_TURNS);
-  const minRecentTurns = parsePositiveInt(env.MAKA_CONTEXT_MIN_RECENT_TURNS, 2);
+  const maxHistoryEstimatedTokens = defaultHistoryBudgetTokens(
```
Astro-Han, line 37: **[P2] A batch of still-supported operator settings is now silently ignored.**

This line replaces the env-driven resolution that stood here on `main`:

```ts
// origin/main:packages/runtime/src/context-budget-policy.ts:41-44
const maxHistoryEstimatedTokens =
  parseOptionalPositiveInt(env.MAKA_CONTEXT_HISTORY_BUDGET_TOKENS) ??
  defaultHistoryBudgetTokens(connection, contextWindow, reserveTokens);
const maxHistoryTurns = parseOptionalPositiveInt(env.MAKA_CONTEXT_HISTORY_BUDGET_TURNS);
const minRecentTurns = parsePositiveInt(env.MAKA_CONTEXT_MIN_RECENT_TURNS, 2);
```

Verified by cross-ref grep over `packages` + `apps` at both refs: `MAKA_CONTEXT_HISTORY_BUDGET_TOKENS`, `MAKA_CONTEXT_HISTORY_BUDGET_TURNS`, `MAKA_CONTEXT_MIN_RECENT_TURNS` and the `HISTORY_COMPACT_{HIGH_WATER_RATIO,FORCE_RATIO,TARGET_RATIO,TAIL_TOKENS,MIN_RECENT_TURNS,MAX_SUMMARY_TOKENS,MAX_BLOCKS,MAX_TOKENS,MAX_BLOCK_TOKENS}` knobs each go from **read on `main`** to **zero reads at this head**.

**These are not all in the retired-experiment set.** #3550 authorised retiring the Semantic Compact / synthesis-cache / history-search settings. The three above are general context-budget operator knobs and they were removed along for the ride.

**Failure scenario.** An existing deployment upgrades without touching its configuration. An operator who had set `MAKA_CONTEXT_HISTORY_BUDGET_TOKENS=1024` silently gets 32000; `MAKA_CONTEXT_MIN_RECENT_TURNS=9` silently becomes 2. Nothing errors, nothing warns — the process keeps running, just no longer under the constraints a human deliberately set. Trigger points, tail retention, cost and latency all move.

Reproduced on this head: calling `buildDefaultContextBudgetPolicy()` with those variables set returns a value **byte-for-byte identical** to the empty-env result.

**Why the new guards don't cover it.** `context-budget-retired-settings.test.ts:25-36` asserts silent-ignore only for the *Semantic Compact* variables. And `context-budget-mid-turn-policy.test.ts:146` still **sets** `MAKA_CONTEXT_MIN_RECENT_TURNS: '4'` — a value nothing reads any more — while a more specific `5` masks the outcome, so the dead global fallback goes unnoticed.

One more layer worth stating, because it decides the severity: **these variable names have zero hits across every `*.md` in the repository.** So there is no documentation to update — and equally, **no existing channel through which an operator would ever learn their configuration stopped taking effect.** That is what makes this silent rather than merely undocumented.

`[P2]` and not `[P1]`: the hard cap still prevents provider overflow, so the consequence is operator intent being substituted, not a crash or data loss.

**Suggested fix** — pick one explicitly: map the non-experimental knobs onto the unified planner, or reject/warn at startup when they are present and point at a migration. Silent success is the one option that shouldn't stay.

packages/runtime-host/src/server/context-coordinator.ts
```diff
@@ -157,6 +158,30 @@ export class HostContextCoordinator {
     }
   }
 
+  #compactResult(turn: import('../protocol/index.js').TurnSnapshot): ContextCompactResult {
+    if (turn.status === 'completed') {
+      if (!turn.contextCompactionOutcome) {
```
Astro-Han, line 163: **[P2] A missing durable outcome raises a bare `Error` and drains the host, instead of returning an operation error.**

This `throw` is not an operation failure. Following it to the `catch` in `#compact`:

```ts
} catch (error) {
  if (error instanceof RuntimeHostedRootConflictError) return sessionBusy(error.message);
  if (error instanceof RuntimeHostedRootUnavailableError) return operationUnavailable(error.message);
  this.#requestDrain();          // <- a plain Error lands here
  throw error;
}
```

A plain `Error` matches neither typed branch, so it triggers `#requestDrain()` and propagates. **A single malformed record therefore escalates into a host drain.**

**Reachability — the upgrade path makes this concrete.** The lookup branch above (`if (existing) { … return { ok: true, result: this.#compactResult(admitted.snapshot) } }`) resolves an *already completed* `context_compact` turn. Two ways it arrives without an outcome:

1. **A `context_compact` turn completed before this PR shipped.** Its `stateDelta` has no `contextCompactionOutcome` at all, because nothing wrote one. Looking that turn up after the upgrade reaches this throw.
2. `canonical-turn-snapshot.ts:113-118` — `readContextCompactionOutcome` returns `undefined` for anything it does not recognise, and the snapshot then omits the field entirely.

Point 2 is worth naming on its own: **the same `ContextCompactionOutcome` type is decoded fail-open by the snapshot reader and fail-closed by the wire decoder** (`decodeContextCompactionOutcome` throws). One tolerant reader feeding one strict consumer is exactly how a shrug becomes a drain. `[P3]` on its own.

**Why tests don't catch it.** `context-coordinator.test.ts` only ever supplies a well-formed `{ kind: 'compacted', checkpointId }`. There is no case for a completed turn whose outcome is absent or unrecognised.

**Suggested fix.** Return a typed operation error (or a `finished` result with a `failed` outcome) rather than throwing; a record that predates the field is a data condition, not a host fault. Consider aligning the two decoders on one strictness.

packages/runtime-host/src/protocol/index.ts
```diff
@@ -91,7 +91,7 @@ export const RUNTIME_HOST_REGISTRATION_SCHEMA_VERSION = 1 as const;
 export const RUNTIME_HOST_PROTOCOL_VERSION = 0 as const;
 // Increment when the same protocol version no longer guarantees safe Client-Host
 // interoperability. Mismatches are rejected before domain commands are admitted.
-export const RUNTIME_HOST_COMPATIBILITY_EPOCH = 39 as const;
+export const RUNTIME_HOST_COMPATIBILITY_EPOCH = 40 as const;
```
Astro-Han: **[P3] The epoch advanced but the ledger below it did not gain a `// 40:` entry.**

Every prior bump left a line recording *why*. The comment block still starts at `// 39:`, so the reason for 40 exists only in a test name.

The bump itself is correct and load-bearing — this PR really does break the wire in both directions, which we verified by running `main`'s decoders against this head's shapes:

| input | decoder | result |
|---|---|---|
| completed snapshot + `contextCompactionOutcome` | epoch-39 `decodeTurnSnapshot` | **throws `Unknown completed Turn snapshot field`** |
| `{kind:'finished', turn, outcome}` | epoch-39 `decodeTurnSnapshot` | **throws `Invalid Turn run status`** |
| old running `TurnSnapshot` | epoch-40 compact result | **throws `Unknown Context compact result field`** |

So the epoch line is what converts a decode-time crash into a handshake refusal. **That is precisely the kind of decision the ledger exists to record** — someone reading this file a year from now should not have to reconstruct it from a test title.

(Adding an optional field is a breaking change here regardless of optionality: the tolerance lives in the decoder that was updated, not on the wire. The bump is the right call.)

Astro-Han (commented):
Multi-line independent review of `02971c5444633eb641a697157a312f70f3078dcd`, split across separate reviewers with non-overlapping surfaces. Findings are inline; this comment carries the gate state and the things that don't anchor to a line.

## Direction

Two independently-run lines concluded the deletion is correct, and having checked the argument myself I agree with the reasoning rather than just the verdict.

The old artifact archival existed **because the old fold was destructive**: `persistHistoryCompactBlocksToArtifacts` stored raw bodies with `bodySha256` / `originalBytes` so `loadHistoryCompactBlocksFromArtifacts` could get them back. The new `applyRuntimeEventHistoryCompact` takes `readonly RuntimeEvent[]` and returns a new array — **it is a projection and does not mutate the event log.** `ARCHITECTURE.md` states this directly:

> Runtime Event Log is the canonical source… Context pruning and compaction change provider input projections, not history.

**The archival was compensating for an implementation that violated the stated architecture. Once compaction became a projection, the thing it compensated for stopped existing.** Six overlapping execution/representation authorities collapse into one RuntimeEvent-prefix planner plus the latest durable checkpoint. That is a genuine reduction, not a relocation.

Three invariants were checked rather than assumed:

- **Every failure path is fail-open.** `planHistoryCompaction` returns `decision: 'fail_open'` at all three failure exits, and `selectSafeCompactionPrefix` returns `{ ok: false, reason: 'no_safe_completed_span' }` rather than an unsafe cut point. **A failed compaction costs tokens, never history.**
- **Removed enum values can't break old data.** `CompactionBoundaryKind` dropped `'synthesisCache'` / `'semanticCompact'`, and both were confined to `compaction-boundary.ts` as diagnostic types — neither appears in the `packages/storage` or `packages/core` decode paths.
- **Retired persistence keys are handled explicitly, and that handling is new in this PR.** `usage-record-schema.ts` keeps the retired keys as decode-only history and `usage-record-schema-retired-fields.test.ts` is added. **This is the layer large deletions usually miss.**

## Further findings that don't anchor to a diff line

**[P1] A checkpoint can erase the only model-visible address of an archive whose artifact still exists.** `history-compaction.ts:341-417` lets a summary replace covered events wholesale; `history-compact-checkpoint.ts:82-105` carries no archive roots; `history-compact-summarizer.ts:60-76` requires paths/functions/commands/errors to survive but **not** `maka://archive/...` URIs. Because this PR removes eager `retrieveArchivedToolResultsForReplay`, recovery now depends entirely on the placeholder exposing its URI plus `ArchiveRead`.

Reproduced on this head: feed `planHistoryCompaction` a complete call/result pair whose result placeholder carries a valid archive URI, plus a trailing user turn. A structurally valid summary that says "prior tool output was archived" **without copying the URI** yields `decision='compacted'` and `replacementEvents = [checkpoint text, tail user]`. The artifact is still on disk; the agent no longer holds any ref it could pass to `ArchiveRead`. The write gate validates sections, truncation, summary size and materializability — there is no invariant that covered archive refs remain addressable, and no end-to-end placeholder → checkpoint → `ArchiveRead` test.

**[P2] Legacy persisted v1 placeholders are declared compatible but have no usable `ArchiveRead` argument.** `tool-result-archive.ts:56-70` makes `resourceRef` / `readInstructions` optional for persisted v1, and `:185-198` returns any recognised placeholder as-is. `archive-read-tool.ts:31-39` requires `ref`. With eager hydration removed and no migration synthesising a URI from the existing artifactId/hash/bytes triple, the model can neither see a ref nor construct a valid call. Tests cover new-writer placeholders and hand-supplied URIs, not old placeholder → new turn → readable.

**[P3] Archived compaction artifacts from earlier sessions become orphans.** After `history-compact-artifacts.ts` is deleted, its four exports have zero remaining references — the symbol surface is clean. Semantically, sessions that already compacted wrote raw bodies as artifacts, and nothing reads them now. **No history is lost** (the event log is intact and fail-open replay returns everything), which is why this is `[P3]` — but the archives stay on disk, unreferenced and unmanaged. Migration, one-time cleanup, or a release note saying they are no longer read: **currently none of the three.**

**[P3] Coverage gap on legacy artifact compatibility.** The deleted `history-compact-artifacts.test.ts` asserted `bodySha256` / `sourceArchiveRefs` integrity for `history_compact_source`. The new retired-fields test only asserts that fields are ignored — nothing exercises loading an old artifact that is still in storage.

## Gate state

- exact head `02971c54`, non-draft, `MERGEABLE`. **This head has moved three times within an hour** (`2acb6b46` → `a388fb8f` → `02971c54`); every finding above is bound to `02971c54` and re-verified against it.
- check-runs on this exact head: `audit` and `test` both `completed / success`. **This is a real, executed green** — confirmed via the commit-level check-runs API, not inferred from the UI.
- Review endpoints read separately: 0 issue comments, 0 reviews, 0 inline comments. Nothing to deduplicate against.
- **Two surfaces are still unreviewed** and are not covered by anything above: the artifacts/cache/archive surface beyond the narrow `memory-extraction.ts` + `tool-result-archive.ts` slice, and the desktop/CLI client wiring (`runtime-host-session-execution-ipc-main.ts`, `preload/*`, `pi-transcript.ts`). Reviews are in flight; **this comment does not claim coverage of them.**

No overall verdict offered. Reviewed 2026-08-23 13:30 UTC.

apps/desktop/src/main/runtime-host-session-execution-ipc-main.ts
```diff
@@ -506,8 +506,9 @@ export function registerRuntimeHostSessionExecutionIpc(
 
   ipcMain.handle("sessions:compact", async (_event, sessionId: string) => {
     const turnId = newId();
-    await deps.client.compactContext({ sessionId, turnId });
+    const result = await deps.client.compactContext({ sessionId, turnId });
     deps.emitSessionsChanged("status-change", sessionId, { turnId });
+    return result;
```
Astro-Han, line 547: **[P2] The widened contract has no consumer on the desktop side, so a terminal `failed` compaction is silently swallowed there.**

This `return result` (and the matching `compact(sessionId): Promise<OperationOutput<'context.compact'>>` in `bridge-contract.d.ts:716`) widens the pipe. The renderer never connects it — `app-shell.tsx:2051`:

```ts
try {
  await window.maka.sessions.compact(sessionId);
  return true;                                     // result discarded
} catch (error) { … }
```

`contextCompactionOutcome` has **zero references anywhere in the renderer tree**.

**Why this bites, precisely.** `HostContextCoordinator#compactResult` returns a *failed* outcome from its **success** branch:

```ts
return { kind: 'finished', turn, outcome: { kind: 'failed', reason: … } };
```

So a compaction that terminally failed arrives as an `ok: true` frame. It is not an exception and no `catch` can rescue it. The CLI reads it and renders `failed` with its reason (`pi-transcript.ts`, pinned by `renders manual compaction from the typed terminal outcome`); the desktop `await`s, drops the value, and returns `true`. **The user sees nothing happen.** The desktop equally cannot tell `compacted` from `unchanged`.

To be exact about what is *not* broken: operation-level rejections (session busy / archived / unavailable) arrive as `ok: false` and are turned into a thrown `RuntimeHostOperationError` at `client/connection.ts:844-846`, so those *do* reach the renderer's `catch` and raise the error toast. **The gap is confined to outcomes riding on successful frames** — which is the harder case to notice, since nothing anywhere looks like an error.

**This is not a regression** — the pre-PR `Promise<void>` dropped it too. It is the half of this PR's own unification goal that didn't land: **the contract was widened on both sides of the wire, and only one client was taught to read it.**

**Why nothing caught it.** The CLI has a test pinning all three terminal outcomes; the desktop has no test pinning compact's outcome presentation at all. And the type system is silent about a discarded return value.

Fix either way — have the renderer consume the outcome and surface it as the CLI does, or leave the contract at `void`. **The current shape is half-connected, which is worse than unconnected, because a reader will assume it is wired.**

Astro-Han (commented):
Follow-up: the desktop/CLI client wiring surface, which the earlier review explicitly listed as **not covered**. That review is now extended by one finding; the artifacts/cache/archive surface beyond the narrow `memory-extraction.ts` + `tool-result-archive.ts` slice remains unreviewed and is still not claimed as covered.

Still bound to `02971c5444633eb641a697157a312f70f3078dcd`; re-checked at post time — head unmoved, non-draft, `MERGEABLE`, `test` and `audit` both `completed / success` on this exact head.

One correction worth recording, because it changes what the finding rests on: the first form of this finding blamed `{ok: false}` operation rejections leaking past the renderer's `catch`. **That is wrong** — those are rejected as `RuntimeHostOperationError` at `client/connection.ts:844-846` and the desktop does show an error. The real gap is narrower and harder to spot: outcomes carried on **successful** frames.

No overall verdict offered.

packages/runtime/src/context-budget.ts
```diff
@@ -323,23 +188,10 @@ export function applyRuntimeEventContextBudget(
   return {
     events: keptEvents,
```
Astro-Han, line 189: **[P2] Deleting the tail-budget fallback leaves one live consumer that now ships over-window history straight to the provider.**

The new contract here is deliberate: with no compatible checkpoint, `applyRuntimeEventContextBudget` returns the canonical RuntimeEvent ledger untouched, and the main provider path then detects the overflow in `ai-sdk-backend.ts` and runs a durable compaction transaction. **On that path this is correct.** The only shaping left in this function is `applyRuntimeEventHistoryCompactNarrow`, which needs `policy.historyCompact`; the `maxTokens`-driven drop that stood on `main` (`context-budget.ts:335`, alongside the now-removed `hasOversizedRetainedHistoryTurn`) is gone.

**`buildSessionRecapMessages` is the consumer that doesn't have that safety net** (`packages/runtime/src/session-recap.ts:39-52` — not in this diff, which is why this comment lands here):

```ts
if (budget?.maxHistoryEstimatedTokens !== undefined) {
  budget.maxHistoryEstimatedTokens = Math.max(0, Math.floor(contextWindow * 0.85) - 4_096);
}
events = applyRuntimeEventContextBudget(events, budget)?.events ?? events;
```

It has **no checkpoint loader** and does **not** go through the compaction transaction — it materialises the returned events straight to the auxiliary model. So the cap it carefully computes is now purely diagnostic: it is recorded and then does nothing.

**Reproduced against the production function on this head**, GPT-4 8192 window, 20 visible complete turns of 2000 characters each:

```
recap cap  = 2867 tokens
before     = 10038 tokens / 20 events
after      = 10038 tokens / 20 events     <- unchanged
diagnostic = keptTurns=20 / droppedTurns=0
final recap payload = 21 messages / 42059 serialized chars
```

So an ordinary long session's manual `/recap` and idle-return recap regress to a provider `context_length` error (or silent provider-side truncation) — **at a call site that still looks like it is computing a safe budget**. Reading it, you would conclude it is bounded.

**Why the new tests don't catch it.** `context-budget.test.ts:35-42` pins "canonical ledger unchanged when there is no checkpoint" as the *correct* contract, which it is for the main path. Nothing exercises an over-budget session recap, so the cross-consumer breakage falls between the two.

**Suggested fix, without adding a second durable authority**: give recap its own request-only bounded slice of recent complete turns, or make it checkpoint-aware and enforce the cap on the final provider payload. **Restoring silent dropping inside `applyRuntimeEventContextBudget` would be the wrong repair** — the main path would misuse it.

Budget authority itself checks out: `resolveContextBudgetCapacity()` remains the single capacity derivation, and pre-turn / active-turn / provider-overflow all converge on `planHistoryCompaction()` plus the durable checkpoint writer. Focused suites on this head: context-policy 16/16, overflow-reactive 43/43, supervisor overflow 20/20, AI SDK + mid-turn 256/256.

Astro-Han (commented):
Third and final batch for `02971c54`, covering the context-budget surface. One `[P2]`, inline.

**Gate state changed since the earlier batches — flagging it rather than letting it pass silently:** this PR is **`isDraft = true` again** (it was non-draft when the first two batches were posted). Head is unmoved at `02971c54` and `test` + `audit` are still `completed / success` on it.

Coverage, stated plainly: the compaction core, context budget, runtime-host protocol/coordination, deletion audit, test surface, client wiring, and a narrow `memory-extraction.ts` + `tool-result-archive.ts` slice have each been reviewed by a separate line. **The full artifacts / cache / archive surface has not**, and nothing posted so far should be read as covering it.

No overall verdict offered.

packages/runtime/src/ai-sdk-backend.ts
```diff
@@ -3558,266 +3418,88 @@ export class AiSdkBackend implements AgentBackend {
...
-    if (contextBudget?.synthesisCache?.enabled === true) {
-      contextBudgetDiagnostic = mergeContextBudgetDiagnostic(
-        contextBudgetDiagnostic ??
-          buildContextBudgetDiagnosticShell(priorRuntimeContext, runtimeContext, contextBudget),
-        synthesis.diagnosticPatch,
-      );
-    }
-
-    if (synthesis.selectedBlocks.length === 0) {
-      const retrieval = await retrieveArchivedToolResultsForReplay(
```
Astro-Han, line 3751: **The single change that both archive findings rest on.** Removing this call converts archive retrieval from *automatic and id-addressed* to *model-initiated and URI-addressed*.

Before, replay resolved archived tool results by `runtimeEventId` / `artifactId` through `services.readToolResultArchive` with no participation from the model. After, verified by grep across `packages` + `apps` on this head excluding tests: **the only remaining production reader is `archive-read-tool.ts:92`**, reached solely when the model calls `ArchiveRead` — and `archive-read-tool.ts:31-45` requires a `ref`.

**So the URI stops being a convenience and becomes the sole key.** Anywhere it is absent or lost, the artifact is still on disk and permanently unreachable. Two such places, both found independently by two reviewers working from different directions:

**[P1] Ordinary capacity compaction erases the ref.** `history-compaction.ts:341-417` replaces covered events with the checkpoint; `history-compact-checkpoint.ts:82-114` carries no archive roots; the summarizer prompt (`history-compact-summarizer.ts:60-75`) requires paths, function names, commands and errors to survive but says nothing about `maka://archive/...`.

Reproduced by two probes independently: feed `planHistoryCompaction` a `function_call` + archived-placeholder `function_response` (with a valid `resourceRef`) plus a tail user event, with a structurally valid four-section summary that says the output was archived without copying the URI. Result: `decision='compacted'`, coverage 2, `replacementEvents` = checkpoint text + tail user. No placeholder, no ref. The write gate validates structure, truncation and size — **there is no invariant that a covered archive stays addressable.**

**[P2] Legacy v1 placeholders never had a ref to begin with.** `tool-result-archive.ts:56-70` makes `resourceRef` / `readInstructions` optional for persisted v1, and the recogniser at `:314-340` accepts a placeholder without one. A placeholder written before `29ad3fe5d` carries only `artifactId` / `bodySha256` / `originalBytes`.

Probe result on this head: `buildRuntimeEventModelReplayPlan` emits that placeholder unchanged — `isArchivedToolResultPlaceholder = true`, `parseToolResultArchiveResourceRef(undefined) = null`, `diagnostics = []`. **Not rejected, not migrated, not diagnosed** — it simply enters replay as an opaque object the model cannot act on. On `main` the removed call above would have hydrated it by id.

The optionality at `:56-70` is explicitly justified as v1 compatibility, but the reader that made it compatible is what this line deletes. **The compat declaration and the compat mechanism came apart.**

**Why tests don't cover either.** New tests exercise new-writer placeholders and hand-supplied URIs. There is no archive-placeholder → checkpoint → `ArchiveRead` chain, and no old-placeholder → new-turn → readable case.

**Suggested fixes.** For P1: carry covered archive roots on the checkpoint (or reject a replacement that drops a still-live archive root). For P2: deterministically rebuild `resourceRef` from the existing `artifactId` / hash / bytes triple at load time — validating hash and size — or keep a bounded, session-checked legacy read fallback. Plus regressions for both chains.

**Direction is not in dispute.** Consolidating six overlapping authorities is right, and the host-side reader guards (session / source / hash / size) hold — probes found no cross-session read and no hash bypass. **The gap is that the compatibility layer did not close behind the deletion.**

Astro-Han (commented):
Final batch for `02971c54`. **The artifacts / cache / archive surface — flagged as uncovered in all three earlier batches — has now been reviewed**, by a second line independent of the narrow `memory-extraction.ts` + `tool-result-archive.ts` slice reported earlier.

Both lines landed on the same two defects from different starting points (one from archive retention, one from artifact lifecycle), with separately-executed probes. **That is independent confirmation, not one finding counted twice** — the inline comment consolidates them and adds the deletion they both depend on, which neither had isolated.

One candidate was raised and deliberately **withdrawn**, and the reasoning is worth recording: `conversation-copy.ts`'s `preserve_external` lets a session-owned archive keep its source `artifactId` after branching, while the host reader validates against the *target* session — a plausible dangling-ref shape. **This PR does not change that semantics, so it is not attributed here.** Pre-existing, and worth its own look.

Also not reported, for the same discipline: the deleted `history-compact-artifacts` / `synthesis-cache-artifacts` loaders have no in-tree production writer or caller — only module exports and tests. **Leftover API surface is not evidence of a live user-data path**, so no orphan-block finding is claimed.

Gate state at post time: head unmoved at `02971c54`, base `1d06330a`, `draft = true`, `MERGEABLE`, `audit` + `test` both `completed / success` on the exact head. Local verification on this head: `@maka/core` and `@maka/runtime` builds pass; 76 focused archive/capability/conversation-copy/read-model tests pass; the current `ai-sdk-backend` suite is 189/189.

Coverage is now complete across all seven surfaces. No overall verdict offered.

packages/runtime/src/context-budget-policy.ts
Astro-Han: Confirmed. These settings are intentionally removed rather than mapped onto the unified planner, which would preserve a second policy authority. I removed the remaining dead test input and documented the exact removals in the PR body.

packages/runtime-host/src/server/context-coordinator.ts
Astro-Han: Fixed. A completed legacy compaction without a durable outcome now returns `failed / missing_durable_outcome` without draining the Host. Added a regression test.

packages/runtime-host/src/protocol/index.ts
Astro-Han: Fixed. Added the epoch-40 compatibility entry describing the typed compaction outcome wire break.

apps/desktop/src/main/runtime-host-session-execution-ipc-main.ts
Astro-Han: Fixed. Desktop now consumes `compacted`, `unchanged`, and `failed` outcomes from successful frames and presents them. Added coverage for all three.

packages/runtime/src/context-budget.ts
Astro-Han: Fixed without restoring canonical-history dropping. Session recap now builds its own request-only projection from the newest whole turns within its model budget.

packages/runtime/src/ai-sdk-backend.ts
Astro-Han: The legacy-placeholder gap is fixed by deterministically deriving the ArchiveRead ref from `artifactId`, hash, and byte size. We intentionally do not retain archive refs across checkpoints: compaction ends model reachability of covered details, while the complete original result remains in the canonical RuntimeEvent ledger. This avoids creating a permanent retrieval authority.

Astro-Han (commented):
Re-review at exact head `c0c59d01d816f93eed9d84c867477ecaf6cb29d3`. `test` and `audit` are both `completed / success` on that SHA. The PR is still a draft, so this is a status report rather than a gate decision.

**Scoping note that may be useful to you as much as to reviewers.** The head shows 87 files, but relative to `02971c54` — the head our four earlier review batches were bound to — **the actual delta is 21 files, +519/−321**. Reviewing this as a fresh 87-file change would have been wasted effort on both sides; it was reviewed as a delta.

## Closed on this head

**The bare `Error` in `#compactResult` is gone.** A completed compaction with no durable outcome now returns `{ kind: 'finished', turn, outcome: { kind: 'failed', reason: 'missing_durable_outcome' } }` instead of throwing. That matters more than the diff size suggests: the thrown `Error` fell through `#compact`'s catch to `#requestDrain()` and rethrow, so a turn that had merely lost its outcome record took down the session path. It is now a typed terminal state, which is what the rest of this PR's design already assumed.

**The epoch ledger is complete.** `// 40:` is present with the three-line explanation of why epoch-39 peers must fail at handshake.

**The discarded IPC result is fixed — and this is the one the delta was clearly aimed at.** `app-shell.tsx:2054-2062` now reads `result.outcome`, maps it through the new `contextCompactionNotice`, routes `error` to `showSessionError` with `return false`, and everything else to a toast. `shell-copy.ts` carries all four new keys in both locales. Previously the renderer had zero references to `contextCompactionOutcome` and threw the result away, so a compaction that failed looked identical to one that succeeded.

**`session-recap.ts` — the finding we were least willing to take on trust, and it is genuinely fixed.** The old defect was that it overwrote `budget.maxHistoryEstimatedTokens`, so the cap was computed and then had no effect. Re-measured on this head rather than read:

| | before | after |
|---|---|---|
| cap | 2867 | 2867 |
| estimated tokens before | 10038 | 20120 |
| **estimated tokens after** | **10038** | **2036** |
| keptTurns / droppedTurns | 20 / 0 | 2 / 18 |

Same cap, and it now actually bounds the request. Structurally, `recentTurnsWithinBudget` is a pure function that touches no budget object at all — a repo-wide grep for `maxHistoryEstimatedTokens =` leaves exactly one local construction in `context-budget-policy.ts:40` and no mutation anywhere. The new test pins it at the boundary (`user-0` excluded, `model-19` included) rather than asserting a number.

`runtime-host-ssh-terminal.ts` replacing `timeoutMs`/`timeoutMessage` with a structured `{ timedOut, exit }` is behaviour-preserving, and injecting `managementTimeoutMs` makes the path testable.

## Still open — two `[P2]`s, unchanged because their files are untouched by this delta

**[P2] Three environment variables are dead, and nothing tells an operator.** At this head, `MAKA_CONTEXT_HISTORY_BUDGET_TOKENS`, `MAKA_CONTEXT_HISTORY_BUDGET_TURNS` and `MAKA_CONTEXT_MIN_RECENT_TURNS` have **zero** references — verified per-ref (`git grep <name> <sha>`), where `main` still has two files each. The nine `HISTORY_COMPACT_*` names are in the same position. Repo-wide `*.md` hits: **zero**. So an operator who sets any of these today gets silence, not a warning and not a documented replacement. A line in the changelog or a startup warning for a set variable that is no longer read would close it.

**[P2] Archive retrieval changed shape, and only the code says so.** `retrieveArchivedToolResultsForReplay` has zero references at this head against four files on `main`. The remaining production reader is `archive-read-tool.ts:92`, which requires a `ref`. Retrieval therefore moved from **automatic and id-addressed** to **model-initiated and URI-addressed**. The removal itself is deliberate and stated in the description — the finding is about *visibility*: anything depending on the old semantics fails silently rather than loudly.

Both are documentation/communication gaps rather than defects in the new design, which is why they are P2 and not higher.

## Coverage and limits, stated rather than implied

Two independent lines covered this delta — runtime compaction/recap/archive, and desktop plus the three carried-over findings — with the coordination and protocol pieces checked separately. Local runs across history-compaction, checkpoint, mid-turn, input-fit, summarizer, runtime-event-read-model, tool-result-archive and the codex compactor: **175+3 green, 0 failures**.

**One limit worth recording:** `02971c54` was force-pushed away and cannot be fetched, so per-commit patch comparison against the previously reviewed tree was impossible. It was substituted with re-measuring the disputed numbers, full semantic reading of the delta files, and running the suites — but that is a different kind of evidence than a byte-level range-diff, and it should be read as such.

*This review is AI-assisted. Under `CONTRIBUTING.md` §Review it does not count as the required independent human review — and this PR is still a draft, so nothing here is a merge signal.*

Astro-Han (commented):
Independent re-review of `e8de574faa06779537ee39459f96cdc6772bc331`.

**No P0–P2. One [P3].** Exact-head hosted checks: `test` = `completed / success`, `audit` = `completed / success`, both bound to that exact SHA.

## What I verified

- `packages/runtime/src/context-budget-policy.ts` at head contains **no** `process.env` read and no `MAKA_CONTEXT_*` reference. `staleToolResultPrune.enabled`, `historyCompact.enabled`, `historyCompact.midTurn.enabled` and `activeToolResultPrune.enabled` are all hardcoded `true`. There is no remaining opt-out anywhere in the shipped tree.
- This is **deliberate and pinned**, not an accident: `packages/runtime/src/__tests__/context-budget-mid-turn-policy.test.ts:28` asserts that setting all thirteen `MAKA_CONTEXT_*` variables produces a policy `deepEqual` to the empty-environment baseline. A future reintroduction of env tuning now fails a test.
- Deleted test coverage checked, and the deletions are sound. `context-budget-retired-settings.test.ts` only asserted that already-retired Semantic Compact env vars did not alter policy — it pinned nothing about pruning behaviour. The two removed `mid-turn-capacity-backend.test.ts` cases only covered the `historyCompactOff` configuration state, which no longer exists; the default-path equivalents survive at `:1539-1556`, `:1598-1612`, `:1727-1748`.
- Window-bounded reserve derivation (`min(16384, max(1, floor(window/4)))`) is covered for the small-window case that motivated it.

## [P3] The `### Changed` entry does not tell an already-configured user what happens to them

`CHANGELOG.md:31` correctly states that `MAKA_CONTEXT_*` overrides no longer tune or disable compaction. What it does not say is the part that actually affects someone upgrading: **if you previously set one of these — most consequentially `MAKA_CONTEXT_ACTIVE_TOOL_RESULT_PRUNE=off` — it is now ignored, pruning is on, and there is no replacement opt-out.** That configuration was recommended in this repository's own release notes (`CHANGELOG.md:589-591`, the `0.1.2` section) and appears as a benchmark toggle in `docs/archive/economic-mechanisms-benchmark.md:112-116`.

Those two documents are historical records and should **not** be rewritten — the forward-facing entry is the right place to close this. Suggested addition to the `### Changed` bullet, or a short `### Breaking` note:

> Previously supported `MAKA_CONTEXT_*` settings are now ignored rather than honoured. If you set `MAKA_CONTEXT_ACTIVE_TOOL_RESULT_PRUNE=off` or `MAKA_CONTEXT_STALE_TOOL_RESULT_PRUNE=off`, Tool Result pruning is re-enabled on upgrade and there is currently no supported way to disable it.

I am **not** filing "there is no off switch" as a defect. Removing the escape hatch is a deliberate product direction that this PR states and tests. The reviewable gap is only that a user who followed our own published instructions gets no upgrade-time notice that their setting stopped working — which one sentence fixes.

## Open question for the maintainers (not a review finding)

Whether a supported emergency bypass should exist at all for provider-side or cost anomalies is a product call, not a code-review call. If the answer is "no bypass", the [P3] above is the complete remediation.

---

Publication note: this PR is authored under the shared `Astro-Han` account that our review agents also publish from, so we cannot supply the independent approval this PR needs. It requires a review from a committer outside that account before it can merge.

jackwener (commented):
Review of exact head `e8de574faa06779537ee39459f96cdc6772bc331` on behalf of the Kabi review line. `test` and `audit` are both `completed / success` on this SHA. **COMMENT: two P2s remain, so I am not approving this head.**

The direction is valuable: the PR deletes several overlapping compaction, synthesis-cache, archive-hydration, and environment-policy authorities and makes the Runtime Event Log remain canonical while compaction becomes a provider-input projection. The findings below are boundary gaps in that simplification, not an argument to restore the deleted architecture.

## [P2] Desktop loses the normal `started -> terminal` `/compact` outcome

`apps/desktop/src/renderer/app-shell.tsx:2054-2062` displays a notice only when the immediate `context.compact` reply is `kind === 'finished'`. A fresh admission normally returns `kind: 'started'` (`packages/runtime-host/src/server/context-coordinator.ts:161-186`). The eventual durable result arrives on `complete.contextCompactionOutcome`, but `apps/desktop/src/renderer/app-shell-session-events.ts:378-396` refreshes state without consuming that outcome.

Reproduction: run `/compact` for an idle session with new work. The Host returns `started`; later it emits `complete` with `compacted`, `unchanged`, or `failed`; Desktop shows none of them. Only an idempotent/already-terminal call reaches the new immediate `finished` branch. The added Desktop test covers the pure formatter, not this lifecycle. CLI already consumes the terminal event correctly.

Clean fix: track the admitted compaction turn ID and consume its terminal `complete.contextCompactionOutcome` exactly once, including reconnect/resnapshot; retain immediate `finished` handling. Cover fresh `started -> compacted/unchanged/failed`, immediate `finished`, and reconnect without duplicate notices.

## [P2] One oversized newest turn makes session recap forget the whole session

`packages/runtime/src/session-recap.ts:55-61` stops before selecting any group when the newest complete turn alone exceeds the recap budget. `buildSessionRecapMessages` then sends only `SESSION_RECAP_INSTRUCTION`.

This was reproduced through the production helper with one latest user RuntimeEvent containing a sentinel plus 20,000 characters: `messages.length === 1`, the sentinel is absent, and the only message is the recap instruction. The previous implementation always retained configured recent turns; the instruction-only edge is introduced by the new request-local slicer.

Clean fix: add a bounded intra-turn fallback for an oversized newest group, preserving its newest substantive user/assistant facts without splitting tool protocol units. Cover both an oversized text turn and an oversized tool-call/result turn, and assert the request remains within budget.

## [P3] `no_safe_completed_span` is reported as `summarizer_failed`

`packages/runtime/src/ai-sdk-backend.ts:3483-3485` maps every failed `compactHistory` result to `summarizer_failed`. With an over-budget ledger containing one open `function_call`, no safe fold boundary exists: the summarizer is not invoked and the model request count is zero, but the terminal detail is still `summarizer_failed` instead of `no_safe_completed_span`.

Preserve stable compaction failure reasons in `ContextBudgetExhaustedDetail`; reserve `summarizer_failed` for an actual summarizer or summary-validation failure. Add paired tests for an open call with no safe boundary and a throwing summarizer.

## Verification and provenance

- Full affected build passed; targeted compaction/budget/pruning/overflow/protocol/Desktop/CLI/recap/legacy-compat suite: **502 passed, 0 failed**.
- Focused current-head suite: **6 passed, 0 failed**. The recap and failure-detail boundaries were separately reproduced; temporary tests were removed and the review worktree ended clean.
- Candidate audit also covered checkpoint/prune replay ordering, archive reachability and checksum/ownership checks, legacy v1 placeholder reconstruction, deleted semantic/cache/search paths, fixed policy values, malformed protocol outcomes, and epoch/ledger compatibility; no additional P0-P3 was established.

The provisional judgment was sealed at `2026-08-23T08:38:54Z` on this exact head before reading GitHub reviews, inline comments, issue comments, or Raft history. A separate zero-history reviewer instance produced the evidence; the accountable publication judgment is Kabi's under @WAWQAQ's standing review authorization. This is AI-assisted review and does not claim to satisfy the repository's required independent human review.

Astro-Han (commented):
Correction to my earlier review `5001953079` on this same head.

That review concluded **No P0–P2**. That conclusion was wrong. An independent review published as `5002014024` found two P2s on `e8de574faa06779537ee39459f96cdc6772bc331`, and I have reproduced both against the source rather than taking them on report.

**Desktop never surfaces a compaction result that does not complete synchronously.** `apps/desktop/src/renderer/app-shell.tsx:2055` acts only on `result.kind === 'finished'`; a normal new request returns `started` and the handler falls through to `return true`. The follow-up `complete.contextCompactionOutcome` has no consumer at all on that side — `git grep -n "contextCompactionOutcome" -- apps/desktop/src` returns nothing, and the only reader in the tree is `packages/cli/src/pi-transcript.ts:512-515`. `contextCompactionNotice` itself handles all three outcomes, so the copy exists; it is simply never reached for the asynchronous path. Success, no-change, and failure are all invisible.

**The recap slicer can hand the model an empty conversation.** `packages/runtime/src/session-recap.ts:56-58` walks groups newest-first and breaks on `selectedTokens + group.estimatedTokens > maxEstimatedTokens`. On the first iteration `selectedTokens` is still `0`, so a single latest turn that exceeds the budget breaks immediately, `selected` stays empty, and `selected.flat()` yields `[]`. What reaches the model is the pushed `SESSION_RECAP_INSTRUCTION` and nothing else. With the budget at `contextWindow * 0.85 - 4096` this is not a corner reachable only under unusual configuration.

Both are worth fixing before this merges. Neither changes my view of the PR's direction — the consolidation is real and the deletions are sound — but "no P0–P2" was not accurate and should not have been published as it stood.

For the record on how this was missed: I scoped my two review lines to the environment-variable removal and to whether the net deletion took behaviour with it. Both were pointed at what the PR *removed*. Nothing I dispatched was pointed at the runtime behaviour of the code that stayed or was newly written, which is where both of these live. The independent line that found them did not inherit that framing.

packages/runtime/src/session-recap.ts
```diff
@@ -38,21 +37,96 @@ export function buildSessionRecapMessages(input: {
...
+
+function boundedOversizedTurnMessages(
+  events: readonly RuntimeEvent[],
+  maxEstimatedTokens: number,
+  charsPerToken = 4,
+): ModelMessage[] {
+  const messages = replayPlanItemsToModelMessages(buildRuntimeEventModelReplayPlan(events).items);
+  try {
+    return fitHistoryCompactMessages(messages, {
+      maxInputEstimatedTokens: maxEstimatedTokens,
```
M4n5ter, line 85: **[P2] A zero recap evidence budget becomes an unlimited request.**

For supported 4,096-token models, line 42 computes `maxEstimatedTokens = max(0, floor(4096 * 0.85) - 4096) = 0`. The whole-turn selector therefore returns no events and this fallback passes `maxInputEstimatedTokens: 0` here. But `fitHistoryCompactMessages` applies `finitePositive` to that value; zero becomes `undefined`, and `history-compact-input-fit.ts:40-41` interprets `undefined` as no cap and returns every message unchanged.

I reproduced this through `buildSessionRecapMessages` on the exact head with a valid 4,096-token model and one oversized user turn: the request retained the entire turn and serialized to **36,904 characters**, plus the recap instruction. Supported generated model metadata also includes 4,096-and-smaller windows, so this is reachable. Existing tests use an 8,192-token model and do not exercise the zero-budget branch.

The recap failure is caught, so canonical history and the primary turn remain safe; the user instead loses the return recap when the auxiliary provider rejects the over-window request. Please handle a zero derived budget before calling this helper, or give the fitter a distinct explicit-unbounded representation and enforce a positive request cap that includes instruction/output reserve.

**[P2] 零 token 的 recap 证据预算被解释成了无限制输入。** 对 4,096 token 模型，L42 算出的预算是 0；本分支把 0 传给 `fitHistoryCompactMessages`，但该函数用 `finitePositive` 解析，0 会变成 `undefined`，而 `undefined` 的语义是“不设上限”，所以整段超大 turn 被原样保留。精确 head 上实测得到 36,904 字符输入，现有 8,192-token 测试覆盖不到这里。主会话不会损坏，但 recap 请求会因超窗失败，用户拿不到返回摘要。应在调用前单独处理零预算，或区分“无限制”和“零预算”，并给包含 instruction/output reserve 的请求设置真实正上限。

</details>

M4n5ter (commented):
Independent review of exact head `e885bc5511361e26c1ab644a8f02165a64ffbbb5`: **COMMENT — one P2 remains, so this is not an approval.**

The consolidation is a real simplification: manual, automatic-capacity, overflow, and mid-turn compaction converge on one RuntimeEvent-prefix checkpoint transaction and one typed durable outcome; the raw RuntimeEvent ledger remains canonical. The retired environment-policy behavior is now explicitly documented, so I do not carry forward the earlier compatibility finding.

Validation on this head: Runtime **2,979 passed / 0 failed / 7 skipped**; Runtime Host **1,110 / 0**; Desktop **1,261 / 0**; Desktop preload/main/renderer/Storybook typechecks passed; `git diff --check` and the current-main merge tree are clean. Hosted `audit` is green; hosted `test` was still running at review time. The remaining defect is inline.

当前精确 head 上结论为 **COMMENT，仍有 1 条 P2，因此不批准**。这次重构确实删掉了多套并行权威，不只是搬目录：四类压缩入口统一到一套 RuntimeEvent 前缀 checkpoint 事务和一个类型化终态；原始 RuntimeEvent ledger 仍是唯一事实源。环境变量策略的退役也已明确写入升级说明，因此不再延续旧的兼容性 finding。

本地验证：Runtime 2,979 通过、0 失败、7 跳过；Runtime Host 1,110/0；Desktop 1,261/0；四套 Desktop TypeScript 检查通过；当前 main merge-tree 与 `git diff --check` 均干净。发布时 hosted `audit` 已绿，`test` 仍在运行。剩余问题见内联评论。

</details>

packages/runtime/src/session-recap.ts
Astro-Han: The zero-as-unbounded behavior is real, but I don’t agree with the P2 classification or treating it as a merge blocker.

Maka does not ship or recommend a default chat model with a context window at or below this boundary. The generated models.dev metadata is fallback model metadata, not a statement that those models are part of Maka’s supported/default product surface. Reaching this path requires manually declaring a custom model and assigning it an unusually small context window; the remaining 4K catalog entries are mostly legacy or specialized models rather than models used for Maka agent sessions.

The impact is also confined to the auxiliary session recap: the recap may fail, while the primary turn, canonical RuntimeEvent history, and compaction state remain intact. Under the project’s reachability and recoverability criteria, this is P3 hardening rather than a P2.

I still agree that zero must not mean unbounded. We’ll handle the zero-budget branch explicitly with a small local fix, without adding another budget abstraction.

M4n5ter (approved):
<details open>
<summary>English</summary>

Independent exact-head re-review of `d9c1f810ee3f093c532480369a04d61114a86f9c`: **APPROVE**.

The only delta since the previous review is commit `d9c1f810ee3f093c532480369a04d61114a86f9c`. It prevents a zero recap-evidence budget from reaching the fitting helper where zero means unbounded. With a declared 4,096-token model, an oversized latest turn now produces only the recap instruction instead of retaining the unbounded turn.

I reviewed the two-file delta and its supported boundary paths. The focused recap suite passed 4/4; the Runtime build, Biome check, and diff whitespace check passed; exact-head hosted `test` and `audit` checks are both completed/success. The previous P2 is resolved, and I found no new P0–P3 issue in this delta.

This AI review records the technical result; it does not claim to replace any human committer review required by ASF policy.

</details>

<details>
<summary>中文</summary>

对 exact head `d9c1f810ee3f093c532480369a04d61114a86f9c` 的独立增量复核结论：**APPROVE**。

相对上一轮审查，唯一增量是提交 `d9c1f810ee3f093c532480369a04d61114a86f9c`。它阻止零 recap 证据预算进入“零表示无限制”的裁剪 helper。对声明为 4,096 token 的模型，超大最新 turn 现在只生成 recap 指令，不再保留无界 turn。

我复核了两个文件的增量及其受支持边界路径。定向 recap 测试 4/4、Runtime build、Biome 与 diff whitespace 检查均通过；exact-head hosted `test` 和 `audit` 都是 completed/success。上一轮 P2 已关闭，本次增量未发现新的 P0–P3。

这条 AI review 只记录技术结论，不声称替代 ASF 政策要求的人类 committer 审查。

</details>

PR #3547, opened 2026-08-23
docs: turn the Simplified Chinese README link into a 中文文档 badge

jackwener:
## Summary
The English README linked to `README.zh-CN.md` with a small subscript `简体中文` text link. Make that a shields.io badge whose label is **中文文档**, matching the other header badges.

The Chinese README gets the same treatment for `English`, so the language switch looks the same from both sides.

## Test plan
- [x] shields.io returns SVG with aria-label/title `中文文档`
- [ ] GitHub renders the badge on this PR's README preview
- [ ] Clicking the badge opens `README.zh-CN.md` (and `English` opens `README.md`)

Do not merge unless asked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@
 </p>
 
 <p align="center">
-  <sub><a href="./README.zh-CN.md">简体中文</a></sub>
+  <a href="./README.zh-CN.md"><img src="https://img.shields.io/badge/%E4%B8%AD%E6%96%87%E6%96%87%E6%A1%A3-4C8DFF?style=flat" alt="中文文档" /></a>
 </p>
 
 <p align="center">
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -33,7 +33,7 @@
 </p>
 
 <p align="center">
-  <sub><a href="./README.md">English</a></sub>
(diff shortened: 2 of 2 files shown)


Astro-Han (approved):
Approving `b19d2996efd4da8d6b063467a4cef3436eaf40d2`. Required `test` is `completed / success` bound to that exact SHA ([job](https://github.com/apache/maka/actions/runs/32617379097/job/97140182327)). **No P0–P3.**

Two files, `+2/-2`: the subscript language link in each README becomes a Shields badge. The relative `href` targets are unchanged, so the navigation this affects still points where it did.

**Checked rather than assumed, since a badge is an external fetch and not just markup:** both SVGs return HTTP 200 with `image/svg+xml`, and their accessible text is correct — `中文文档` on the English README's badge and `English` on the Chinese one. `alt` text and `href` are right on both sides, and `git diff --check` is clean.

**Disclosure, because it changes what this approval is worth:** this is an AI review. Under `CONTRIBUTING.md` §Review it does **not** count as the required independent human review. Two PMC members have said in chat that they are happy with this change, but that agreement is not recorded on the PR — branch protection reads the review record, so this still needs a committer other than the author to leave an approving review here, and a human to merge.

PR #3554, opened 2026-08-23
fix(cli,runtime): enforce --max-steps on maka run turns

### LINKED ISSUE #3553: bug(cli,runtime): `maka run --max-steps N` is accepted but never enforced
### What happened

`maka run --max-steps 1` runs the whole turn as if no cap was set. A prompt that needs three serial tool calls completes all three, prints the summary, and exits 0. The result is the same through a Runtime Host, where the turn admission already records `execution.maxSteps: 1`.

Two handoffs lose the value:

1. `runMakaTextCliCore` parses the flag but leaves `maxSteps` out of the `UserMessageInput` it passes to `runtime.sendMessage` (`packages/cli/src/run-command-core.ts`).
2. `RuntimeKernel` builds the `RuntimeRunner` invocation from `AgentRun.backendInput` and drops `maxSteps` (`packages/runtime/src/runtime-kernel.ts`, around line 1690). `AiSdkBackend` resolves `input.maxSteps ?? this.maxSteps` and would stop the loop at `runtimeSteps >= maxSteps`, but the per-turn value never reaches it. The Runtime Host path carries the cap into admission through the adapter's context fallback and loses it at this same place.

Unfixed build (0.1.11 dev at `1ddbd2591`, same on `origin/main` at `09e459b28`), one fixture directory with two Markdown files, same prompt on both paths:

| Path | Exit | Tool calls | `runtimeSteps` | `stopReason` | Admitted cap |
|---|---:|---:|---:|---|---:|
| Local profile | 0 | 3 | 4 | `end_turn` | 1 |
| Runtime Host | 0 | 3 | 4 | `end_turn` | 1 |

Expected: the first provider step completes and settles its tool call, no second provider request starts, and the turn ends with the existing `step_limit` / `tool_step_cap_reached` outcome.

### How to reproduce

1. Create a directory with two Markdown files and configure a working model connection.
2. Run:

```sh
maka run --cwd "$FIXTURE" --max-steps 1 --timeout 180 \
  'Do these three steps strictly one after another, one tool call per step, no parallel calls: 1) use Glob to list *.md in the project root; 2) use Read on the first file, first 5 lines; 3) use Grep for the word "the" in the project root and report the file count. Finish with one sentence summarizing all three results.'
```

3. Read the turn ledger: three `function_call` events, `runtimeSteps: 4`, `stopReason: end_turn`, exit 0.
4. Repeat through a Runtime Host with a project pointing at the fixture. Same ledger, and the admission record shows `execution.maxSteps: 1`.

Unit reproduction on `origin/main`: assert that `sendMessage` receives `maxSteps` from `runMakaTextCliCore` (fails with `undefined !== 2`), and that `SessionManager.sendMessage` with `maxSteps: 1` against a mock provider that always returns a tool call stops after one provider step (fails with `3 !== 1`).

### Environment

- Maka version or commit: 0.1.11 dev build at `1ddbd2591`; confirmed on `origin/main` at `09e459b28`
- OS and version: macOS 26.6.1 (25G76)
- Surface: CLI `maka run`, local profile and Runtime Host
- Node.js v22.22.2, npm 10.9.7

Local, unfixed:

```text
exit code: 0
tool-call count: 3
runtimeSteps: 4
stopReason: end_turn
admitted maxSteps: 1
stdout: Found `alpha.md` and `beta.md`; the first five lines of `alpha.md` were read, and a project-root search for lowercase `the` matched 0 files.
stderr: Node.js SQLite experimental warning only
```

Runtime Host, unfixed:

```text
exit code: 0
tool-call count: 3
runtimeSteps: 4
stopReason: end_turn
admitted maxSteps: 1
stdout: Glob found `alpha.md` and `beta.md`; the first five lines of `alpha.md` were read; and `Grep` found "the" in 0 files.
stderr: Node.js SQLite experimental warning only
```

I have a fix with regression tests for both handoffs; PR to follow.

jsiu93:
## Summary

`maka run --max-steps N` was parsed and then ignored. Two handoffs dropped the value:

- `runMakaTextCliCore` left it out of the `UserMessageInput` that starts the turn.
- `RuntimeKernel` left it out of the `RuntimeRunner` invocation it builds from `AgentRun.backendInput`, so `AiSdkBackend` never received a per-turn cap. The Runtime Host path carried the cap as far as turn admission and lost it at the same place.

This forwards the cap at both boundaries. The backend's step accounting and `step_limit` outcome are unchanged: with `--max-steps 1`, the first provider step completes and settles its tool call, and no second provider request starts. The turn then ends through the existing `tool_step_cap_reached` failure: empty stdout, `maka run: Turn failed: tool_step_cap_reached` on stderr, exit 1. A step is one completed provider request; parallel tool calls in one response still count as one step.

Fixes #3553

## Verification

Each regression test ran red with the forwarding removed and green with it restored:

- CLI boundary (`run-command.test.ts`): `undefined !== 2`
- Runtime loop (`session-manager.test.ts`, mock provider that always returns a tool call): `3 !== 1`
- Runtime Host seam (`root-turn-coordinator.test.ts`, admission to backend input): `undefined !== 1`

Affected suites:

```sh
npm --workspace maka-agent test          # tests 383, pass 383, fail 0
npm --workspace @maka/runtime test       # tests 3085, pass 3072, skipped 13, fail 0
npm --workspace @maka/runtime-host test  # tests 1080, pass 1080, fail 0
```

`npm run typecheck` passed in all three workspaces. `biome lint` and `biome format` on the five changed files reported no fixes.

Live runs against an OpenAI-compatible relay, same fixture and prompt as the issue, on the fixed build:

| Path | Exit | stdout | stderr beyond the SQLite warning | Tool calls | `runtimeSteps` | `stopReason` |
|---|---:|---|---|---:|---:|---|
| Local profile | 1 | empty | `maka run: Turn failed: tool_step_cap_reached` | 1 | 1 | `step_limit` |
| Runtime Host, fresh loopback Host from this build | 1 | empty | `maka run: Turn failed: tool_step_cap_reached` | 1 | 1 | `step_limit` |

Unfixed, both paths ran 3 tool calls, `runtimeSteps: 4`, `end_turn`, exit 0.

Not run: the full monorepo `npm test`, root-wide typecheck, lint and format, Windows or Linux, other provider families.

## AI use

Select exactly one:

Tool(s) and scope: pi (gpt-5.6-sol) wrote the regression tests, both forwarding changes, the live verification harness, and the first draft of this description. Claude Code reviewed the diff, re-ran the three suites, and edited this text. Both commits carry the trailer.

Generated-by: pi (gpt-5.6-sol)

## Checklist

- [ ] No

### DIFF
--- a/packages/cli/src/__tests__/run-command.test.ts
+++ b/packages/cli/src/__tests__/run-command.test.ts
@@ -19,7 +19,9 @@
 
 import assert from 'node:assert/strict';
 import { describe, test } from 'node:test';
-import { parseMakaRunArgs } from '../run-command-core.js';
+import type { UserMessageInput } from '@maka/core/runtime-inputs';
+import type { SessionSummary } from '@maka/core/session';
+import { parseMakaRunArgs, runMakaTextCliCore, type MakaRunAdapter } from '../run-command-core.js';
 
 describe('maka run argument parsing', () => {
   test('recognizes stdin prompt mode and rejects malformed limits', () => {
@@ -70,4 +72,79 @@ describe('maka run argument parsing', () => {
       'error',
     );
   });
+
+  test('forwards --max-steps to local and Graph turn inputs', async () => {
+    for (const graph of [false, true]) {
+      const messages: UserMessageInput[] = [];
+      const adapter: MakaRunAdapter = {
+        listSessions: async () => [],
+        createContext: async (input) => ({
+          runtime: {
+            createSession: async () => sessionSummary(),
+            readExecutionBoundary: async () => ({
+              kind: 'managed',
+              access: 'writable',
+              revision: 0,
+            }),
+            sendMessage: async function* (_sessionId, message) {
+              messages.push(message);
+              await input.runOutcomeObserver?.({
+                outcomeId: message.turnId,
+                status: 'completed',
+                finalOutput: 'done',
+                sandboxBoundary: 'none',
+              });
+            },
+            respondToSandboxBoundary: async () => {},
+            stopSession: async () => {},
+            setExecutionBoundaryKind: async () => {},
+          },
+          target: { connection: { slug: 'test' }, model: 'test' },
+          agentGraph: {
+            reserveActivity: () => ({ release: () => {} }),
+            waitForCompletion: async () => {},
+          },
+          close: async () => {},
+        }),
+      };
+
+      const exitCode = await runMakaTextCliCore(
+        ['answer once', '--max-steps', '2', ...(graph ? ['--graph'] : [])],
+        adapter,
+        {
+          workspaceRoot: () => process.cwd(),
+          processCwd: () => process.cwd(),
+          stdinIsTTY: () => true,
+          writeStdout: () => {},
+          writeStderr: () => {},
+          onSigint: () => () => {},
+          newId: () => 'turn-1',
+        },
+      );
+
+      assert.equal(exitCode, 0);
+      assert.equal(messages.length, 1);
+      assert.equal(messages[0]?.maxSteps, 2);
+    }
+  });
 });
+
+function sessionSummary(): SessionSummary {
+  return {
+    id: 'session-1',
+    cwd: process.cwd(),
+    name: 'Run once',
+    isFlagged: false,
+    isArchived: false,
+    labels: [],
+    hasUnread: false,
+    status: 'active',
+    backend: 'ai-sdk',
+    llmConnectionSlug: 'test',
+    connectionLocked: true,
+    model: 'test',
+    permissionMode: 'ask',
+    collaborationMode: 'agent',
+    orchestrationMode: 'default',
+  };
+}
--- a/packages/cli/src/run-command-core.ts
+++ b/packages/cli/src/run-command-core.ts
@@ -397,6 +397,7 @@ export async function runMakaTextCliCore(
     for await (const event of context.runtime.sendMessage(session.id, {
       turnId: deps.newId(),
       text: prompt,
+      ...(parsed.options.maxSteps !== undefined ? { maxSteps: parsed.options.maxSteps } : {}),
       ...(parsed.options.graph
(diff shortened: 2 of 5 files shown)


Astro-Han (approved):
Reviewed exact head `24a6aff370ee55f09a2e01571e06998973ce7a67` — required `test` is `completed / success` bound to that SHA. **No P0–P3.**

**The problem was real and the fix is exactly four lines of production code**, which is the right size for it. `maka run --max-steps` parsed and validated fine but died at the CLI boundary: `run-command-core.ts` never put `maxSteps` on the `turn.message.submit` payload, and `runtime-kernel.ts` never forwarded `begin.backendInput.maxSteps` into the backend input. The protocol and the Host admission path already accepted and validated the field — so this was a dropped wire, not a missing feature, and the patch restores exactly that wire rather than introducing a parallel limit.

**That last point is the one I checked hardest**, because a second cap would be the expensive mistake here: the CLI does not enforce anything of its own. It hands the number to the runtime and the existing per-turn cap in the AI SDK flow remains the single enforcement point, so CLI and runtime cannot drift apart. Both additions are conditional spreads, so an unset `--max-steps` produces no key and the prior default path is byte-identical.

The remaining +264 lines are tests, and they are layered where the wire actually runs: `run-command.test.ts` pins the CLI submission, `root-turn-coordinator.test.ts` pins Host admission, `session-manager.test.ts` pins the kernel forwarding and the `step_limit` terminal state once tool steps reach the cap. That is three independent pins on one path, which is what keeps a dropped wire from silently coming back.

**A note on reading this PR's diff, for whoever looks next:** the recorded base is a descendant of an earlier ancestor of the branch, so a plain `base..head` shows a misleading reverse diff. The real change is the 5-file merge result — review it via the GitHub merge ref, not `base..head`.

---
*AI-assisted review. Each claim above was verified against the source at this exact head. Under `CONTRIBUTING.md` §Review this does **not** count as the required independent human review — merge still needs a human committer other than the author to give LGTM and to decide.*

PR #3558, opened 2026-08-23
ci(windows): gate packaged sandbox lifecycle evidence

liugddx:
## Summary

This PR promotes the existing Windows sandbox lifecycle evidence into the shared packaged verifier used by both the pull-request `Release Windows check` and the formal release workflow.

The verifier runs `appcontainer-smoke.ps1` and `acl-recovery-smoke.ps1` against the exact broker under `win-unpacked`. These probes cover concurrent AppContainer launches, process/Job drain, residual ACE and ledger cleanup, and recovery after the broker is killed mid-launch.

The pull-request path filter now covers the complete 41-file source closure of the packaged filesystem worker and sandbox driver. A release contract derives that closure from the real `filesystem-worker` and `sandbox` package entrypoints plus the worker bundle entrypoint using esbuild metadata, then fails if any reachable repository source is absent from the workflow paths. The contract found and closed 14 omissions on this branch and also rejects duplicate path entries, so future imports cannot silently escape the pre-merge Windows lane.

The allowlist also covers `.github/workflows/release.yml`, the release workflow that invokes the Windows package and lifecycle verifier.

This changes release verification only. It does not modify Runtime, broker, launch, cancellation, retry, ACL, or sandbox policy behavior. Refs #2142.

## Remaining Phase 4 work

This PR intentionally does not close the Phase 4 lifecycle checkbox. Explicit client cancellation is covered separately by merged #3586. Runtime Host death during launch, sustained concurrency soak, unsettled-state recovery/quarantine, the wider adversarial matrix, and independent human security review remain separate follow-up gates.

## Verification

Exact head `f65dc9d154cf1e57d211ad7aad1dedcffc33da33`, rebased onto `main` `84ed9a310`:

- `node --test scripts/ci-test-plan.test.mjs scripts/windows-package-source-closure.test.mjs scripts/verify-windows-harness.test.mjs`: 74 passed, 0 failed, 1 privilege-dependent symlink skip.
- `git diff --check` passed.
- [Core CI 32687127680](https://github.com/apache/maka/actions/runs/32687127680) passed in 16m15s, including build, release contracts, typecheck, affected workspace tests, Desktop E2E, Storybook, and CLI validation.
- [Dependency audit 32687127642](https://github.com/apache/maka/actions/runs/32687127642) passed in 46s.
- [Release Windows check 32687127636](https://github.com/apache/maka/actions/runs/32687127636) passed in 14m53s, including fresh packaging, lifecycle recovery, pinned upgrade/uninstall, automatic update, and deterministic rollback.

The source-closure contract covers the complete 41-file worker/driver closure and rejects duplicate workflow paths. The follow-up control-plane contract also proves that edits to `release.yml` arm the packaged Windows gate.

## AI use

Tool(s) and scope: Codex traced the existing Windows evidence and release ownership, implemented the shared packaged gate, derived and enforced the real packaged source closure in response to review, rebased after upstream merges, retained both the upstream renderer and independent lifecycle harness contracts, ran the local quality gate, and performed an author-side adversarial review. The commit includes a `Generated-by: Codex` trailer. AI review is not independent human security review.

## Checklist

- [x] Lint, format, typecheck, build, and affected platform-neutral suites pass locally
- [x] Fresh packaged Windows L3 passes on this exact head

- [x] Yes - Windows package and formal release verification now fail when the packaged sandbox lifecycle evidence fails, and the release contract prevents reachable worker/driver sources from escaping the Windows PR trigger.
- [ ] No


liugddx:
Could a maintainer rerun the failed `Release Windows check` job for exact head `3ce3d4d35dea88d3b62ec66716e9aea7149d2fc1`?

The failure occurred in the existing `check:release` suite during `Package the Windows installer and ZIP`, before this PR's packaged lifecycle verifier ran. `verify-packaged-app.test.mjs` expected the notice-version rejection (`react@19.2.0`) but one fixture read returned the earlier PTY package-identity error instead. Neither `verify-packaged-app.test.mjs` nor `verify-packaged-app.mjs` is in this PR's diff.

Evidence for rerun rather than a scope-expanding fix here:

- the same #3265/main package path passed immediately before this rebase;
- `node --test scripts/verify-packaged-app.test.mjs` passes 18/18 in the #3558 exact-head worktree;
- the same isolated test passes 18/18 in the separate #3265 worktree;
- Core CI on this exact head passed.

The contributor token cannot rerun upstream Actions (`Must have admin rights to Repository`), and I do not want to add an empty CI-only commit. If it repeats, that is evidence to fix the existing asar fixture determinism separately; this PR still needs a fresh green packaged run before its L3 box is checked.

Astro-Han (commented):
Review at exact head `229d09c727f4719c70b01e5e975023ffc09408a6`.

The gate itself is sound. Both PowerShell verifiers are awaited in sequence and any non-zero exit rejects through `runCommand`, so missing or malformed lifecycle evidence fails closed rather than passing vacuously — the failure mode that makes a gate decorative. I confirmed the exact-head hosted run `32634441543` (job `97181975711`) really executed the lifecycle step, and the evidence it gates on comes from that same fresh package: there is no path by which a stale artifact from an earlier run could satisfy it.

### `[P1]` The path allowlist doesn't cover the sources the gate actually exercises

The new trigger entries reach the packaged worker and the Windows boundary driver:

```yaml
- 'packages/runtime/src/filesystem-worker/**'
- 'packages/runtime/src/sandbox/**'
- 'packages/core/src/permission-profile.ts'
- 'packages/core/src/permission-profile-compiler.ts'
```

But the code in those directories imports Core modules that are not listed. Directly:

- `packages/core/src/sandbox-boundary.ts` → imported by `filesystem-worker/client.ts`, `operations.ts`, `protocol.ts`
- `packages/core/src/windows-path.ts` → imported by `sandbox/windows-profile.ts`, `sandbox/windows-sandbox.ts`

and one level further, `sandbox-boundary.ts` imports `serialized-byte-length.js` while `windows-path.ts` imports `absolute-path.js`. The Runtime side has the same gap for `child-fd-input.ts`, `child-process-lifecycle.ts` and `process-tree-terminator.ts`.

From Core, only the two `permission-profile*` files are listed. So a change confined to `sandbox-boundary.ts` — the module that defines the boundary this gate exists to protect — triggers neither pre-merge Windows lane. The W0 PR filter doesn't cover it either. A sandbox-boundary regression would land on main and surface only in the nightly, which is precisely the "found at the worst possible moment" situation the workflow header says this path was built to prevent.

This is worth raising because the PR is otherwise strictly an improvement; the gap is that the allowlist is a hand-copied prefix of the real dependency closure, and hand-copied closures drift the moment someone adds an import.

**Suggested direction:** rather than appending these seven files, derive the trigger set from the actual import graph of the packaged E2E entrypoints, and add a contract test asserting that every source reachable from those entrypoints appears in the workflow's `paths`. That converts "someone remembered to update the list" into a checked property, and it fails loudly when a new import escapes the gate. If deriving it is too large for this PR, adding the seven files plus that contract test would still close the drift.

Nothing else found. Verification: exact-head `test` and `package` are `completed/success`; 65/65 targeted tests plus a fake-rejection probe covering missing and malformed second-stage evidence. Windows packaging, PowerShell, the Rust broker, AppContainer/ACL, installer/update/rollback and the full repo suite were **not** run locally and are not claimed.

liugddx:
Thanks for catching this. I agree the hand-maintained prefix was not the actual exercised source closure.

Addressed at exact head `c55a3f21268124c3b251547a32226f8d7b332976`:

- The workflow now covers every repository source in the current 41-file worker/driver closure. Deriving the graph found 14 omissions, so I added the full set rather than stopping at the seven examples.
- `scripts/windows-package-source-closure.mjs` derives that closure from the real `filesystem-worker` and `sandbox` package entrypoints plus the worker bundle entrypoint, using esbuild's resolver/metafile rather than a second hand-written import parser.
- The new release contract compares that graph with `pull_request.paths` and fails on any uncovered source. Core/runtime changes already select `check:release`, while changes to the contract itself are explicitly release-authority inputs.
- The duplicate `verify-windows-sandbox-e2e.mjs` path was removed, and a second assertion prevents duplicates from returning.

Fresh exact-head evidence is green: [Core CI 32679300156](https://github.com/apache/maka/actions/runs/32679300156) passed the release-contract step 111/111, including both new closure assertions, and [Release Windows check 32679300220](https://github.com/apache/maka/actions/runs/32679300220) passed the full packaged lifecycle/upgrade/update/rollback path in 13m50s. The dependency audit also passed.

Could you take another look at this head?

.github/workflows/release-windows-check.yml
```diff
@@ -56,6 +55,32 @@ on:
...
+      - 'packages/runtime/src/edit-replace.ts'
+      - 'packages/runtime/src/file-stable-write.ts'
+      - 'packages/runtime/src/image-file.ts'
+      - 'packages/runtime/src/process-tree-terminator.ts'
+      - 'packages/runtime/src/unified-diff.ts'
+      - 'packages/core/src/absolute-path.ts'
+      - 'packages/core/src/artifacts.ts'
+      - 'packages/core/src/attachments.ts'
+      - 'packages/core/src/permission-profile.ts'
+      - 'packages/core/src/permission-profile-compiler.ts'
```
Astro-Han, line 80: **[P1] This gate does not arm on changes to the workflow it exists to protect.**

Anchoring here because the relevant line is outside this PR's diff — the allowlist entry for this workflow itself is at line 90 of the file.

The allowlist correctly includes this workflow itself at line 90, but omits `.github/workflows/release.yml` — and `release.yml:190-228` is where `npm run package:windows-x64`, `npm run verify:windows-x64`, and the installer-lifecycle, autoupdate, and rollback verifications are actually invoked.

Concretely: a future PR that touches only `release.yml` can delete, reorder, or condition away any of those Windows verification steps, and this lane stays skipped. CI's Linux release-contract selection does not execute the Windows artifact or lifecycle path, so nothing else catches it either. The regression reaches main with no packaged-Windows evidence — which is the exact failure mode described in this file's own header comment.

Worth noting the asymmetry as its own argument: the list already guards against edits to *this* file, so the principle that release wiring must re-arm the gate is one the file already accepts. `release.yml` is the more important half of that wiring.

Minimal fix: add `.github/workflows/release.yml` to this allowlist, and assert the relationship in the path-filter contract test so it cannot silently drift out again.

Astro-Han (commented):
Reviewed at exact head `c55a3f21`. **One [P1], inline.** Publishing on behalf of a reviewer without write access here; I independently verified the finding against this head before posting.

The intent of this gate is right, and the header comment states it well: the Windows release path used to run for the first time on release day, on main, with a tag already reserved, so a defect surfaced at the worst possible moment. Moving that packaging and verification onto a pull request without release identity is the correct fix.

The gap is in which changes arm it. The path allowlist is careful and clearly reasoned — it reaches past the release scripts to `npm-spawn.mjs`, `.gitattributes`, the manifests, the packaged worker sources — and it includes this workflow file itself at line 90. But it does not include `.github/workflows/release.yml`, which is the workflow that actually invokes `package:windows-x64`, `verify:windows-x64`, and the installer/upgrade/autoupdate/rollback steps at lines 190-228.

So a PR editing only the release wiring — the thing this gate exists to protect — leaves this lane skipped.

Evidence at this head: hosted `test`, `audit`, and `package` are all terminal success; the targeted suite runs 74/74 locally; `git diff --check` is clean; and there are no unresolved review threads and no `APPROVED` reviews outstanding. The blocker is the [P1] alone, not the gate status.

.github/workflows/release-windows-check.yml
liugddx: Addressed at exact head `f65dc9d154cf1e57d211ad7aad1dedcffc33da33`.

- Added the exact `.github/workflows/release.yml` path to `release-windows-check.yml`. This is the workflow that invokes the Windows package, release verification, installer upgrade/uninstall, autoupdate, and rollback steps.
- Added `the packaged Windows gate triggers on release orchestration changes` to `scripts/ci-test-plan.test.mjs`, so the relationship cannot silently drift.
- Kept the scope narrow: no `.github/workflows/**` glob, no release workflow behavior change, and no Runtime/broker changes.

Fresh exact-head evidence is green:

- [Core CI 32687127680](https://github.com/apache/maka/actions/runs/32687127680) passed in 16m15s.
- [Dependency audit 32687127642](https://github.com/apache/maka/actions/runs/32687127642) passed in 46s.
- [Release Windows check 32687127636](https://github.com/apache/maka/actions/runs/32687127636) passed in 14m53s, including packaging, lifecycle recovery, upgrade/uninstall, automatic update, and rollback.

The original source-closure finding and this control-plane trigger finding are now both covered. Please re-review this exact head.

scripts/ci-test-plan.mjs
```diff
@@ -62,6 +62,8 @@ const RELEASE_CONTRACT_FILES = new Set([
   'scripts/verify-windows-installer-lifecycle.mjs',
   'scripts/verify-windows-x64.mjs',
   'scripts/windows-upgrade-baseline.json',
+  'scripts/windows-package-source-closure.mjs',
+  'scripts/windows-package-source-closure.test.mjs',
```
jackwener, line 67: **[P2] Route edits to the Windows workflow through this release contract.**

The new closure files are release-contract inputs, but `.github/workflows/release-windows-check.yml` itself still is not. On this head, `planTests(['.github/workflows/release-windows-check.yml'])` returns `releaseContract: false`. I also removed only that workflow's self-trigger entry in a temporary exact-head tree: both `ci-test-plan.test.mjs` and the new source-closure test still passed (36/36).

That is a normal maintenance path with a durable consequence: once such an edit lands, later workflow-only changes no longer run this Windows lane, while Core CI also skips `check:release`, so the allowlist contract has silently lost every execution path. The current self-trigger exists, but nothing outside the gate protects it.

Please add `.github/workflows/release-windows-check.yml` to `RELEASE_CONTRACT_FILES` and to the release-authority planner test (or add an equivalent unconditional self-trigger assertion). Then a workflow edit cannot retire the only trigger that validates future workflow edits.

jackwener (commented):
Reviewed at exact head `d4dc53f5900d691941cc6e146fe478577ac7d8d0`. The two earlier P1 findings are closed: the esbuild-derived closure contains 41 repository package sources with no uncovered entry, and `.github/workflows/release.yml` now arms the packaged Windows lane. The packaged lifecycle verifier is awaited against the broker under `win-unpacked`; a non-zero PowerShell result fails the shared verifier.

One **[P2]** remains, inline. There are no P0/P1 findings. Local evidence: 76/76 targeted tests and `git diff --check` pass. Hosted `test`, `audit`, and `package` are terminal success on this exact head, and the current-main merge tree is byte-identical to the reviewed head tree.

.github/workflows/release-windows-check.yml
liugddx: Re-verified and closed at exact head `09093825398fc05120c831c8462bc2c41ef57c0a`, rebased onto current `main` `d413895bdd9773250af369d16dfc895750efcb0c`.

- `.github/workflows/release.yml` remains an exact `pull_request.paths` input for the packaged Windows gate.
- `the packaged Windows gate triggers on release orchestration changes` continues to pin that relationship in `scripts/ci-test-plan.test.mjs`.
- The current merge result is conflict-free and the exact-head [test](https://github.com/apache/maka/actions/runs/32826688893/job/97736153530), [audit](https://github.com/apache/maka/actions/runs/32826688888/job/97736153469), and [package](https://github.com/apache/maka/actions/runs/32826688916/job/97736153419) checks all passed.

This P1 is closed on the current head.

scripts/ci-test-plan.mjs
liugddx: Addressed at exact head `09093825398fc05120c831c8462bc2c41ef57c0a` in commit `090938253`.

- Added `.github/workflows/release-windows-check.yml` to `RELEASE_CONTRACT_FILES`, so a workflow-only edit now selects `check:release`.
- Added a planner regression proving `planTests(['.github/workflows/release-windows-check.yml']).releaseContract === true`.
- The same test also asserts that the workflow keeps its own path in `pull_request.paths`, so the self-trigger cannot be removed while both protection paths remain green.

The focused planner/source-closure suite passed 37/37 locally; lint, format, and `git diff --check` passed. Exact-head hosted [test](https://github.com/apache/maka/actions/runs/32826688893/job/97736153530), [audit](https://github.com/apache/maka/actions/runs/32826688888/job/97736153469), and [package](https://github.com/apache/maka/actions/runs/32826688916/job/97736153419) checks are all green.

This P2 is closed on the current head.

jackwener (approved):
Re-reviewed exact head `09093825398fc05120c831c8462bc2c41ef57c0a`: the remaining workflow self-protection finding is closed.

Adding `release-windows-check.yml` to `RELEASE_CONTRACT_FILES` makes a workflow-only edit select the independent release-contract lane, and the new test also pins the workflow's own `pull_request.paths` entry. I repeated the original counterexample by removing only that self-trigger entry in a temporary exact-head tree; the new regression test failed, where the previous suite stayed green.

The focused planner/source-closure suite passes 37/37 on the head and on the conflict-free merge with current `main`. Hosted `test`, `audit`, and `package` are completed successfully. I found no remaining P0–P3 issues in this incremental change.

Approved for review purposes. This does not merge the PR; the final merge decision remains with the maintainers.

PR #3560, opened 2026-08-23
fix(mcp): drop JSON Schema annotations from tool schemas

### LINKED ISSUE #3559: bug(mcp): tools whose input schema carries `$schema` cannot be published as Client Capabilities
### What happened

`maka runtime-host capability-provider serve` fails before publishing a single tool when an MCP server advertises input schemas with a root `$schema` annotation. The descriptor is copied into the Client Capability frame unchanged, and the protocol's strict keyword allowlist rejects it:

```text
Runtime Host capability provider stopped: Unsupported Client Capability tool schema keyword
CapabilityProviderPublicationError: Unsupported Client Capability tool schema keyword
[exit=1]
```

This hits most servers built with the TypeScript SDK. `@modelcontextprotocol/sdk` converts Zod input schemas in `server/zod-json-schema-compat.js` with `zod-to-json-schema` (Zod v3) or `z.toJSONSchema` (Zod v4), and both emit a root `$schema` by default. A direct `tools/list` from XcodeBuildMCP 2.7.0 had `$schema` on 44 of 44 tools, and it was the only keyword outside the allowlist in any of them; an earlier probe through `mcp_router` saw the same on 31 of 31 tools. Python FastMCP servers generally do not emit it.

`$schema` and `$comment` are annotations with no validation meaning. Expected: the provider publishes these tools; the allowlist can stay strict for everything else.

### How to reproduce

1. Start a fresh Runtime Host with a loopback WebSocket listener and issue a capability-provider credential (supply it through `MAKA_RUNTIME_HOST_ACCESS_CREDENTIAL`).
2. Point an MCP config at any TypeScript SDK server, for example:

```json
{
  "version": 2,
  "mcpServers": {
    "xcodebuildmcp": {
      "command": "bunx",
      "args": ["xcodebuildmcp", "mcp"]
    }
  }
}
```

3. Run:

```sh
maka runtime-host capability-provider serve \
  --url ws://127.0.0.1:47633/runtime-host \
  --mcp-config ./mcp.json \
  --expected-root "$ROOT_ID" \
  --client-identity ./capability-provider-identity.json
```

4. The provider exits 1 with the error above and publishes nothing.

Unit reproduction: an MCP manager test against a stdio fixture whose tool schema carries a root `$schema`, a nested `$schema` under `properties`, a `$comment`, and an `$id`. On `origin/main` the descriptor still contains all three annotations.

### Environment

- Maka version or commit: 0.1.11 dev build; confirmed on `origin/main` at `6ada6b546`
- OS and version: macOS 26.6.1 (25G76), arm64
- Surface: CLI capability provider and Runtime Host
- Node.js v22.22.2, npm 10.9.7
- MCP server: XcodeBuildMCP 2.7.0 (TypeScript SDK)

Census of the captured XcodeBuildMCP `tools/list`, recursing through the allowlisted child positions:

```text
tools=44
root_$schema=44
unsupported_keywords={"$schema":44}
```

I have a fix that strips `$schema` and `$comment` at MCP discovery and leaves the validator untouched; PR to follow.

jsiu93:
## Summary

MCP servers built with the TypeScript SDK advertise Zod-derived input schemas with a root `$schema`. The capability provider copied those schemas into Client Capability frames, and the protocol's strict keyword allowlist rejected them, so publication stopped before the first tool.

This normalizes tool descriptors at MCP discovery (`descriptorFromTool` in `packages/mcp`): clone the input schema and delete exactly `$schema` and `$comment` from the root and from the child positions the validator walks (`properties`, `items`, `additionalProperties`, `propertyNames`, `$defs`/`definitions`, `allOf`/`anyOf`/`oneOf`). `$id` stays because it changes `$ref` base resolution; every other unsupported keyword still fails closed. The validator itself is unchanged.

The descriptor is the one place every MCP consumer reads from. Desktop's Client Capability projection covers Desktop-native Zod tools, not MCP descriptors, and already strips the `$schema` it generates.

Evidence for how common this is: a direct `tools/list` from XcodeBuildMCP 2.7.0 had `$schema` on 44/44 tools and no other disallowed keyword; an earlier probe through `mcp_router` saw it on 31/31. The SDK's `server/zod-json-schema-compat.js` emits it on both the `zod-to-json-schema` and `z.toJSONSchema` paths.

Fixes #3559

## Verification

Regression test `drops non-semantic JSON Schema annotations from discovered tools` (stdio fixture with root `$schema`, nested `$schema`, `$comment`, `$id`): red before the change (descriptor still carried all three annotations), green after (`$schema` and `$comment` gone, `$id` kept).

```sh
npm --workspace @maka/mcp test           # tests 171, pass 171, fail 0
npm --workspace @maka/runtime-host test  # tests 1084, pass 1084, fail 0
npm --workspace maka-agent test          # tests 382, pass 382, fail 0
npm --workspace @maka/mcp run typecheck  # pass
npm run build                            # pass
```

`biome lint` and `biome format` on the three changed files reported no fixes.

Live, with fresh Runtime Host roots and `bunx xcodebuildmcp mcp` as the server:

```text
before: CapabilityProviderPublicationError: Unsupported Client Capability tool schema keyword, exit 1
after:  Runtime Host capability provider is connected (24 MCP tools)
```

The live server advertised 24 tools in that fresh environment (XcodeBuildMCP exposes tools by workflow); the fixed provider published all 24 it received. The 44-tool figure comes from a separate direct capture of the same server.

Not run: the full monorepo `npm test`, root-wide typecheck, the Desktop and runtime workspace suites, Windows or Linux, a model turn that invokes a published tool.

## AI use

Select exactly one:

Tool(s) and scope: pi (gpt-5.6-sol) traced the MCP-to-Client-Capability path, wrote the regression test and the normalization helper, ran the live publication and the suites, and drafted this description. Claude Code chose the design, reviewed the diff, re-ran the mcp and CLI suites, and edited this text. The commit carries the trailer.

Generated-by: pi (gpt-5.6-sol)

## Checklist

- [ ] No


Astro-Han (commented):
Reviewed at exact head `2613e4b2b9a990499386eb786d8c4e6564a32496`.

**Coverage:** which annotation keywords are stripped, whether equivalent constraint information stays reachable on both the model side and the MCP client side, and which layer the stripping happens in. **Not covered:** wording.

**No P0–P2 findings.** One [P3] on traversal completeness.

### The strip set is correctly scoped

`normalizeToolInputSchema` (`packages/mcp/src/index.ts:2300-2333`) deletes exactly two keywords: `$schema` and `$comment`. Both are pure annotations/processor directives under the JSON Schema spec with **zero validation semantics**. Nothing descriptive or validating is touched — `title`, `description`, `default`, `examples`, `deprecated`, `type`, `required`, `enum`, `items` all survive, and the new test explicitly pins that `$id` is preserved.

### The stripping is confined to the model-facing descriptor

`descriptorFromTool` is the only call site (`:2256`, via `scrubDescriptor`), and its output becomes the model tool definition's `parameters` through `packages/mcp/src/mcp-tools.ts:99`. Execution goes `provider.callTool` → the MCP server validates against its own original schema. Local validation never depended on the stripped keywords, and those keywords carry no validation meaning anyway, so both sides stay equivalent.

### [P3] Traversal misses several schema composition keywords

`visit` recurses through `properties`, `$defs`, `definitions`, `additionalProperties`, `propertyNames`, `items`, `allOf`, `anyOf`, and `oneOf` — but **not** `prefixItems`, `contains`, `not`, `if`/`then`/`else`, `patternProperties`, or `dependentSchemas`. `$schema`/`$comment` nested inside those survive the scrub.

Reachable on the ordinary user path: these shapes are perfectly legal and appear in hand-written server schemas or output from generators other than the TypeScript SDK's converter, even if they are not typical of it. The consequence is exactly the failure this PR exists to fix — if a downstream model provider rejects `$schema` anywhere in the schema tree, these corners keep tripping it, so the fix is incomplete for those shapes rather than wrong.

Anchor: `packages/mcp/src/index.ts:2308-2330`. Smallest fix is adding those keywords to the traversal list; alternatively recurse into every unknown key whose value is a plain object, terminating on a whitelist.

Rated P3 because it affects only edge schema shapes, with no data or security dimension — the outcome is "an uncovered corner keeps producing the original error".

### Verification and limits

The new end-to-end test injects a schema carrying `$schema`/`$comment`/`$id` including a nested property through a stdio fixture and asserts the nested result is clean with `$id` intact — it pins real behaviour rather than text. `manager.test.js` is 60/60 green locally against this head.

`check-runs` on this head is `total_count` 0 — hosted CI has not run, so this is a comment rather than an approval regardless of the outcome.

No real external MCP server was exercised to measure how tolerant various providers are of residual keywords; that part rests on the issue background and the code structure.

jsiu93:
@Astro-Han Thanks for the review. Pushed abd8d3f68 on top of the reviewed head: the traversal now also covers `patternProperties`, `dependentSchemas`, `dependencies`, `prefixItems`, `additionalItems`, `unevaluatedItems`, `unevaluatedProperties`, `contains`, `not` and `if`/`then`/`else`, and the fixture plus the deep-equal assertion pin every one of those positions (both forms of `dependencies` included) with `$id` still intact. The list stays explicit because `default`, `const`, `examples` and `enum` hold instance data, where a `$schema` or `$comment` key is payload.

jsiu93:
@Astro-Han The run for the current head abd8d3f68 was cancelled by the concurrency group when the run for the older head 2613e4b2b started a few seconds after it (that one went green). Could you re-run the cancelled one for abd8d3f68? I cannot trigger a re-run from the fork.

packages/mcp/src/index.ts
```diff
@@ -2293,11 +2293,63 @@ function descriptorFromTool(serverId: string, tool: Tool): McpToolDescriptor {
...
+      'definitions',
+      'dependencies',
+    ] as const) {
+      const entries = value[key];
+      if (!isRecord(entries)) continue;
+      for (const nested of Object.values(entries)) visit(nested);
+    }
+    for (const key of [
+      'items',
+      'prefixItems',
```
jackwener, line 2331: [P3] 第二轮遍历覆盖了上一 head 漏掉的 `prefixItems` / `contains` / `not` / `if`/`then`/`else` / `patternProperties` / `dependentSchemas` 等，方向对。还剩一个合法的嵌套 schema 容器没走进去：`contentSchema`（跟 `contentMediaType` 一起出现）。嵌在那里的 `$schema` / `$comment` 会漏洗，正好是这个 PR 要修的那类提供方拒收。

普通 object 工具 schema 走不到。要补的话把 `contentSchema` 加进这个列表即可。

jackwener (commented):
Independent review of exact head `abd8d3f6824149d1edadfadae2aeb8f190c61265`.

## 三问
1. 要解决什么：MCP tool schema 里的 `$schema` / `$comment` 会让部分模型提供方拒 schema。这是注解/方言标识，不是校验约束。
2. 怎么解决：发现时 `normalizeToolInputSchema` 递归删掉这两项，只改发给模型的 descriptor；MCP server 仍拿它自己的原 schema 做执行期校验。
3. 奥卡姆：删除集合只有 `$schema` 和 `$comment`。`title` / `description` / `default` / `examples` / `deprecated` / MCP `tool.annotations` / `tool.description` 都没动。上一 head 漏掉的 composition 关键字，这个 head 已经补进遍历。

## 校验会不会变松
不会从 schema 里拿掉任何 instance 约束。`$schema` 只影响「用哪份 dialect 去解释」。执行路径是 `callTool` → MCP server 原 schema，那一侧不变。本地若用这份 stripped descriptor 做 Ajv，没有 `$schema` 时会落到默认 draft（本仓库是 2020），那是方言选择，不是约束被删。`patternProperties` / `not` / `if`/`then`/`else` 等都还在。

## UI 文案
`tool.description` 和 MCP annotations（含 `title`）原样留下。schema 上的 `description` / `title` 也不删。仓库里没有把 `$comment` 当 UI 文案读的路径。测试钉了 `$id` 保留、嵌套 `$schema`/`$comment` 删干净，没有单独钉 `description` 存活——代码路径上它不会被删。

## CI
期望：`test`（永远）。未改 package.json / sandbox 路径，不应有 `audit` / `windows_sandbox_*`。
这个 head 上原先 `test` 是 concurrency 取消（cancelled ≠ 绿）。我重跑了，提交时是 in_progress。按门禁不能 APPROVE。

**代码侧 GO（仅 [P3]）。门禁 NO-GO**，等 `test` completed+success。未 approve，未合。

Astro-Han (approved):
Gate check at exact head `abd8d3f6`: hosted `test` is terminal `completed/success` (the only check this path set triggers), `mergeable=true`, and no unresolved [P0]–[P2] — the single open thread is a [P3] on the second-pass keyword coverage in `packages/mcp/src/index.ts`.

Approving. The [P3] is fine to fold into a follow-up; it does not block this change.

yunaremaia:
Reviewed the head. The fix is at the right layer: `descriptorFromTool` is the single point every MCP consumer reads from, so normalizing there (rather than at each projection) means Desktop, the Runtime Host capability provider, and the model adapter all inherit the cleaned schema — and the strict `validateToolInputSchema` allowlist stays unchanged, so fail-closed semantics are preserved.

A few things I checked while reading:

- Recursion covers both shapes the validator walks: object-of-schemas (`properties`, `patternProperties`, `dependentSchemas`, `$defs`/`definitions`, `dependencies`) and schema-or-array (`items`, `prefixItems`, `additionalItems`, `unevaluatedItems`, `additionalProperties`, `unevaluatedProperties`, `propertyNames`, `contains`, `not`, `if`/`then`/`else`, `allOf`/`anyOf`/`oneOf`). `visitSchemaOrSchemas` correctly handles the array case for `items`/`prefixItems`.
- `dependencies` as a string array (property-dependency form) is safely skipped via the `isRecord` guard — only the schema-dependency object is recursed. Good, no over-deletion.
- `$id` is intentionally kept. That is the correct call: a root or inline `$id` changes `$ref` base resolution, so dropping it could silently relink references in servers that use `$ref` + `$defs`. Worth a one-line comment in `normalizeToolInputSchema` noting *why* `$id` is exempted, so a future editor does not "complete" the strip list and break ref resolution.
- The fixture pins every one of those positions with both forms of `dependencies`, and the deep-equal assertion confirms `$schema`/`$comment` gone while `$id` survives. Solid regression coverage.

One optional edge for a follow-up (not blocking): a tool whose schema uses `$ref` with a sibling `$schema` — draft-07 ignores siblings to `$ref`, so stripping `$schema` there is harmless, but a test pinning that `dependencies`/`allOf` under a `$ref`-bearing node still get cleaned would close the loop. Not required for this fix.

LGTM — the live before/after (provider-exit-1 to 24 tools published) matches the issue's expected behavior.

PR #3565, opened 2026-08-23
fix: close source release legal provenance gaps

M4n5ter:
<details open>
<summary>English</summary>

## What

Close the four source-only legal/provenance gaps found by auditing the real ASF source candidate at `6ada6b54665a03dadc699c147d18b8d469168ebe`.

- [x] Stop redistributing the deprecated standalone GitHub Copilot mark, for which the project has no auditable permission record; use Maka's generic provider mark while retaining the visible `GitHub Copilot` name and copy.
- [x] Make root `LICENSE`, `NOTICE`, and `DISCLAIMER-WIP` describe the source archive's actual bytes and patch/adapted-source obligations. The NOTICE product line is now `Apache Maka (incubating)`.
- [x] Correct the mixed-origin attribution for the DeepSeek Harness-derived `cordis.patch.yml`.
- [x] Replace the rolling models.dev input with a repository-contained normalized projection, a digest-bound generation contract, and an explicit network refresh path.
- [x] Generate and independently audit a new exact-head ASF source candidate.

## Baseline and scope

- Exact head: `9b5140d442bed8d4e7815500d4ae82e49f6e5a5b`
- Current-main baseline: `729839ed8ada3e5498b0ad27e64fb42dfd283ae7`
- Tracking issue: #3270
- Scope is the ASF source candidate only. npm and Desktop binary artifact legal closure remain in G8/G9.

## Exact-head source candidate

- Artifact: `apache-maka-0.1.11-incubating-src.tar.gz`
- SHA-512: `c8739cef16d53c6d036078048dc830ba172d3e30c44fe1ae41af474d268f35802e4754a27dc3e806042178cc3b80c7fff8a19a299d68ec83d78903b7f56f9dbd`
- Extracted inventory: 2,875 files and 153 directories; no symlinks, special entries, path traversal, or compiled executables/libraries.
- Header audit: 2,744 covered files plus 131 reviewed exclusions; no missing, duplicate, stale, or unreviewed-provenance result.
- Root legal documents and third-party inventories match the extracted bytes.

## Verification and independent review

- Clean candidate install and production dependency audit: 0 vulnerabilities.
- Candidate build and all workspace typechecks pass.
- `check:asf-source`: 59/59 tests pass.
- Exact-head hosted checks are terminal green: dependency audit, full CI (including source mechanics, lint, formatting, build, release contracts, typecheck, inventories, workspace/Runtime Host/Desktop/Storybook/CLI tests), and Windows package/upgrade verification.
- Independent deep-code-review found three decision-changing issues during development: stale merge-ref snapshot state, dependency-requiring source checks running before install, and documentation that confused offline sync with network refresh. All three were fixed and the final exact-head recheck is GO with no residual finding.
- Independent simplify-audit removed the unused direct-input generation bypass and two low-value negative tests. The final exact-head recheck found one committed snapshot authority, derived generated files, no residual duplicate path, and no decision-changing simplify finding.
- The normalized models.dev projection reproduces current-main runtime metadata and pricing byte-for-byte except for provenance headers; this legal-closure PR does not refresh catalog behavior.

## AI use

Generative tooling made a substantive contribution. Codex performed the source-candidate audit, implementation, tests, independent review orchestration, and documentation; the human contributor of record owns the final review and submission.

</details>

<details>
<summary>中文</summary>

## 内容

修复在真实 ASF 源码候选包（`6ada6b54665a03dadc699c147d18b8d469168ebe`）审计中发现的四组 source-only 法律与来源闭环缺口。

- [x] 停止再分发项目没有可审计许可记录、且已被 GitHub 标为弃用的独立 Copilot 图形；改用 Maka 通用 provider 图标，同时保留可见的 `GitHub Copilot` 名称和说明文字。
- [x] 让根 `LICENSE`、`NOTICE`、`DISCLAIMER-WIP` 与源码包实际字节及 patch/adapted source 义务一致；NOTICE 产品行现为 `Apache Maka (incubating)`。
- [x] 修正 DeepSeek Harness 派生 `cordis.patch.yml` 的混合来源归因。
- [x] 用仓库内规范化投影、digest 绑定的生成契约和显式联网刷新路径替代滚动 models.dev 输入。
- [x] 在最终 exact head 上生成并独立审计新的 ASF source candidate。

## 基线与边界

- Exact head：`9b5140d442bed8d4e7815500d4ae82e49f6e5a5b`
- 当前 main 基线：`729839ed8ada3e5498b0ad27e64fb42dfd283ae7`
- 跟踪 issue：#3270
- 本 PR 只处理 ASF source candidate；npm 与 Desktop 二进制产物法律闭环仍分别属于 G8/G9。

## Exact-head 源码候选

- 产物：`apache-maka-0.1.11-incubating-src.tar.gz`
- SHA-512：`c8739cef16d53c6d036078048dc830ba172d3e30c44fe1ae41af474d268f35802e4754a27dc3e806042178cc3b80c7fff8a19a299d68ec83d78903b7f56f9dbd`
- 解包清单：2,875 个文件、153 个目录；无 symlink、特殊条目、路径穿越或编译后的可执行文件/库。
- Header audit：2,744 个 covered 文件 + 131 个已审查 exclusion；无缺失、重复、陈旧规则或未审查来源。
- 根法律文件和第三方清单与解包后的实际字节一致。

## 验证与独立审查

- 干净候选安装与生产依赖审计通过：0 个漏洞。
- 候选完整 build 与所有 workspace typecheck 通过。
- `check:asf-source`：59/59 通过。
- Exact-head hosted checks 已终态全绿：依赖审计、完整 CI（含源码机制、lint、format、build、release contracts、typecheck、清单、workspace/Runtime Host/Desktop/Storybook/CLI 测试）以及 Windows 打包/升级验证。
- 独立 deep-code-review 在开发过程中发现三项会改变决策的问题：merge-ref 快照陈旧、需要依赖的 source 检查在安装前运行、文档混淆离线 sync 与联网 refresh。三项均已修复，最终 exact-head 复核 GO，无剩余 finding。
- 独立 simplify-audit 删除了未使用的 direct-input 生成旁路和两条低价值负面测试。最终 exact-head 复核确认只有一个 committed snapshot authority，生成文件均为派生物，无剩余重复路径或会改变决策的简化 finding。
- models.dev 规范化投影逐字节复现当前 main 的运行时元数据和价格（provenance header 除外）；本法律闭环 PR 不刷新 catalog 行为。

## AI 使用

生成式工具做出了实质贡献。Codex 执行源码候选审计、实现、测试、独立审查编排和文档；记录中的人类贡献者负责最终审阅与提交责任。

</details>


M4n5ter:
<details open>
<summary>English — independent review record</summary>

## Verdict

**GO on exact head `9b5140d442bed8d4e7815500d4ae82e49f6e5a5b`. No residual P0-P3 finding.**

This is the final result of two independent lanes requested for this PR. Both lanes formed their own provisional judgment before the findings were reconciled. Findings were fixed on the author branch and both lanes rechecked the final exact head.

## Deep-code-review

Three decision-changing findings were found and fixed:

1. **[P1] Merge-ref snapshot staleness.** The first implementation had seeded the new snapshot before current main added `alibaba-cn`. The PR merge-ref therefore retained new generated outputs while the older snapshot could not reproduce them; exact-head hosted `test` and Windows `package` both failed. Fixed by merging current main `729839ed8ada3e5498b0ad27e64fb42dfd283ae7`, reseeding the projection from that authority, and verifying that the two generated TypeScript files differ from current main only in provenance headers.
2. **[P1] Complete source gate ran before its dependencies existed.** `check:asf-source` includes generator tests that require the repository formatter, but CI invoked it before `npm ci`. Fixed by keeping the extracted-tree header audit install-free, adding the ASF-source lane to dependency installation, and running the complete source/generation gate afterward. A workflow-order contract now protects that exact boundary.
3. **[P2] Contribution docs confused offline sync with network refresh.** Both READMEs still told maintainers to use `sync:model-metadata` to update from models.dev even though that command had become intentionally offline. Fixed by documenting `refresh:model-metadata` as the network/snapshot update path and `sync:model-metadata` as offline derivation.

Final deep recheck: all three are resolved; no residual decision-changing finding.

## Simplify audit

Two removable paths/tests were found and deleted:

1. The legacy `--input` / `loadDirectInput()` path could generate outputs from supplied bytes without updating the committed snapshot authority. It was unused and is gone; supplied responses now go through the single `--refresh --refresh-input` path.
2. Two negative assertions only memorialized removed NOTICE/Geist text. They were deleted; the positive exact-NOTICE contract and actual asset/inventory invariants remain.

Final simplify recheck: the committed normalized snapshot is the one input authority; the generated TypeScript files are deterministic derivatives. The new workflow-order test protects a stable preinstall/postinstall contract rather than mirroring implementation. No residual simplify finding.

Entropy accounting: 47,435 of this PR's 48,115 added lines are the normalized snapshot. That is a large data surface, but it replaces an implicit rolling network authority and makes exact source bytes reproducible. The PR also deletes the unreviewable Copilot SVG/notice, the direct-input bypass, and low-value removal tests. Removing the snapshot or consuming it directly at runtime would be a disproportionate cross-package redesign for this legal-only change.

## Exact source candidate and gates

- Candidate: `apache-maka-0.1.11-incubating-src.tar.gz`
- SHA-512: `c8739cef16d53c6d036078048dc830ba172d3e30c44fe1ae41af474d268f35802e4754a27dc3e806042178cc3b80c7fff8a19a299d68ec83d78903b7f56f9dbd`
- 2,875 files / 153 directories; no symlink, special archive entry, path traversal, or compiled executable/library.
- Headers: 2,744 covered + 131 reviewed exclusions, all clean.
- Clean install; production dependency audit: 0 vulnerabilities.
- Candidate build, all workspace typechecks, and `check:asf-source` 59/59 pass.
- Exact-head hosted checks are terminal green: `audit`, full `test`, and Windows `package`.
- PR is mergeable against current main.

Rejected candidates included removing the workflow-order test, collapsing the install-free header audit into the complete source gate, deleting the generated TypeScript representations, and treating current-main `alibaba` / `alibaba-cn` entries as PR-local duplicate authority. None changes the merge decision.

</details>

<details>
<summary>中文 — 独立审查记录</summary>

## 结论

**Exact head `9b5140d442bed8d4e7815500d4ae82e49f6e5a5b` 为 GO，无剩余 P0-P3 finding。**

这是按要求执行的两条独立审查线最终结果。两线先各自形成 provisional judgment，再汇总；发现的问题均在作者分支修复，两线随后重新核查最终 exact head。

## Deep-code-review

发现并修复了三项会改变决策的问题：

1. **[P1] Merge-ref 快照陈旧。** 第一版快照早于当前 main 新增 `alibaba-cn`，导致 PR merge-ref 保留了新生成输出，但旧快照无法复现；exact-head hosted `test` 和 Windows `package` 均失败。现已合并当前 main `729839ed8ada3e5498b0ad27e64fb42dfd283ae7`，从该 authority 重建投影，并确认两个生成 TypeScript 文件相对当前 main 只变化 provenance header。
2. **[P1] 完整 source gate 在依赖安装前执行。** `check:asf-source` 包含依赖仓库 formatter 的生成器测试，但 CI 当时在 `npm ci` 前运行它。现保留解包 header audit 的 install-free 属性；ASF-source lane 安装依赖后再跑完整 source/generation gate，并用 workflow-order contract 固定这个边界。
3. **[P2] 贡献文档混淆离线 sync 与联网 refresh。** 两份 README 仍让维护者用已改为纯离线的 `sync:model-metadata` 从 models.dev 更新。现明确 `refresh:model-metadata` 负责联网并更新快照，`sync:model-metadata` 只从已提交快照派生输出。

最终 deep 复核：三项均已解决，无剩余会改变决策的 finding。

## Simplify audit

发现并删除了两条可移除路径/测试：

1. 旧 `--input` / `loadDirectInput()` 可以绕过 committed snapshot、直接用输入字节生成输出。该路径无人使用，现已删除；提供离线响应只走唯一的 `--refresh --refresh-input` 路径。
2. 两条负面断言只是在记忆已经删除的 NOTICE/Geist 文本。现已删除；保留精确 NOTICE 正向契约和真实资产/清单不变量。

最终 simplify 复核：committed normalized snapshot 是唯一输入 authority，两个生成 TypeScript 文件均为确定性派生物。新增 workflow-order test 保护稳定的安装前/安装后契约，不是实现镜像。无剩余 simplify finding。

熵核算：本 PR 48,115 行新增中，47,435 行是规范化快照。数据面确实很大，但它取代了隐式滚动网络 authority，使源码字节可复现；同时 PR 删除了无法完成权利审查的 Copilot SVG/notice、direct-input 旁路和低价值删除测试。删除快照或让运行时直接消费它，都将变成超出本法律闭环 PR 的跨包重构。

## Exact source candidate 与门禁

- 候选包：`apache-maka-0.1.11-incubating-src.tar.gz`
- SHA-512：`c8739cef16d53c6d036078048dc830ba172d3e30c44fe1ae41af474d268f35802e4754a27dc3e806042178cc3b80c7fff8a19a299d68ec83d78903b7f56f9dbd`
- 2,875 个文件 / 153 个目录；无 symlink、特殊归档条目、路径穿越或编译后可执行文件/库。
- Header：2,744 covered + 131 个已审查 exclusion，全部通过。
- 干净安装；生产依赖审计 0 漏洞。
- 候选 build、全部 workspace typecheck、`check:asf-source` 59/59 通过。
- Exact-head hosted `audit`、完整 `test`、Windows `package` 终态全绿。
- 与当前 main 无冲突，可合并。

已排除的候选包括：删除 workflow-order test、把 install-free header audit 合并进完整 source gate、删除生成 TypeScript 表示、把当前 main 的 `alibaba` / `alibaba-cn` 当成本 PR 引入的重复 authority。它们都不改变合并判断。

</details>

Astro-Han (dismissed):
Reviewed at exact head `9b5140d442bed8d4e7815500d4ae82e49f6e5a5b`. Hosted checks are terminal green on this exact head: `test`, `package`, `audit`.

**Coverage:** the behaviour change in `scripts/sync-model-metadata.mjs`, alignment of the legal texts against what the repository actually ships, and the Copilot mark replacement. **Not covered:** the 47,435-line vendored snapshot was not read line by line — it was checked only as verbatim captured data.

**No P0–P3 findings.**

### 1. Snapshot replaces the rolling input, and it fails closed

The default path now reads the in-repo snapshot `scripts/model-metadata/models-dev-api.snapshot.json` (`formatVersion` 1, original `sourceUrl` preserved). `fetch(SOURCE_URL)` appears only under the `--refresh` branch, so the default run makes no network call — which is the point for a reproducible source release. `--check` compares against the generated artifacts. Missing or malformed fields make `buildProjection` throw rather than silently falling back to stale values.

### 2. Legal texts match what is shipped

- **LICENSE** — the three added redistribution declarations (node-pty, ai-sdk 5.0.27, the astryx patch) match the actual files under `patches/`, including versions and paths.
- **NOTICE** — product line moves from `Maka` / `The Maka Authors` to `Apache Maka (incubating)` / `The Apache Software Foundation`.
- **DISCLAIMER-WIP** — still present, correctly. The only change is dropping the now-resolved known issue `* NOTICE has not yet been reworked into ASF form.`; the ICLA/software-grant item stays. The disclaimer is not withdrawn, it just loses one item that this PR closes.
- **`cordis.patch.yml`** — the mixed-origin attribution matches the corresponding LICENSE section.

### 3. Copilot mark: name kept, mark no longer redistributed

Only the vendored `github-copilot.svg` asset and its provenance comment are removed; `ProviderBrandMark` for `github-copilot` falls back to `GenericProviderMark` with the reason stated inline (Primer Octicons' MIT grant does not cover GitHub logos). The visible `GitHub Copilot` name and its copy remain across the OAuth service, the IPC main module, the locale copy, and the settings display/flow modules. That is the right shape: stop redistributing the mark, keep the service identifiable.

**Approving.** No P0–P2, and the exact head is terminal green. Merging is for a human committer.

Astro-Han (commented):
Second independent review at exact head `9b5140d442bed8d4e7815500d4ae82e49f6e5a5b`, run at the author's request after our first approval — which we have since dismissed so it would not act as a gate while this ran.

**This line reaches NO-GO: one [P1] and one [P2].** The source-release legal closure itself is clean, and is now verified far more strongly than in the first pass.

### Correction to our earlier review

Our first review described the snapshot as verbatim captured data. **That is wrong, and the file says so itself:** `origin.kind` is `generated-output-migration` with `commit: 729839ed…`, and it stores a normalized `projection`, not an API response body. There is no stored response body, ETag, or upstream hash, so "unmodified capture of the upstream API" is not a claim anyone can verify from this file.

What *can* be proved strictly, and was: loading the two generated TypeScript outputs at base `729839ed` and comparing all four export groups — `metadata`, `providerOverrides`, `providerFacts`, `pricing` — gives exact equality with the snapshot. So it faithfully migrates the previously published data; the historical upstream bytes are simply not recoverable. LICENSE is honest about this limitation, so it is not a finding on its own — but review text calling it a raw capture should be corrected, and this comment corrects ours.

### [P1] `--refresh` accepts a structurally valid but incomplete upstream response, and `--check` then passes

Using today's live API response as a base and setting `anthropic.models` to `{}` while keeping every provider and the rest of the shape intact: the production refresh path exits 0, the written snapshot contains zero Anthropic models, and a subsequent `--check` also exits 0. Separately, changing one model's modalities to an unknown enum value (`picture`/`words`) also exits 0, silently producing `vision: false` with empty input/output.

The existing guards are real but sit on the wrong axis: a missing whole provider, or a required number arriving as a string, does throw before the write. What gets through is the partial subset / empty set, and the unknown value in a closed enum — exactly the shapes a rolling upstream produces when it partially degrades.

Since the whole point of this change is to make the snapshot the authority for reproducible releases, an upstream blip can now be frozen into the repository and pass its own verification.

Suggested fix: have `--refresh` compute a destructive delta against the existing snapshot and refuse by default on provider/model/pricing deletions and empty providers, requiring an explicit flag for a reviewed removal; reject unknown modality values outright.

### [P2] `--refresh` is not transactional

It writes the snapshot first (`scripts/sync-model-metadata.mjs:254`), then formats and writes the two generated outputs (`:166`, `:172`). Given a perfectly valid response but a non-existent output parent directory, the command fails with `ENOENT` **after** the new snapshot is already on disk, leaving the repository holding a snapshot whose generated outputs were never produced.

Suggested fix: produce all three artifacts in memory or in same-directory temp files, publish only when all succeed, and replace the snapshot last. A write-failure test should assert the old snapshot is byte-identical afterwards.

### Source-release legal closure: verified green

This is what the first pass did not do, and it holds up:

- The source artifact was generated from the exact head; its SHA-512 matches the PR description character for character, and `verify` passes.
- Unpacked: 2875 files, 153 subdirectories, 0 special files.
- Header/provenance run against the unpacked tree: 2744 covered plus 131 reviewed exclusions, all green.
- All 24 provider SVGs had their SHA-256 recomputed and each matches the in-archive inventory.
- Every path declared in LICENSE exists, and each third-party-source exclusion maps to either the root LICENSE or the provider inventory. No reverse-empty declaration was found — nothing declared that isn't in the archive.

Full reproduction steps, hashes, and acceptance conditions for the two findings are available on request.

Astro-Han (commented):
Second independent line, incremental re-review at exact head `eea04b0768d558801fc81da8e3876942ca1ff744`.

**Closed since the last head:** the publish transaction is now atomic (`[P2]` resolved), and an unknown modality on a *selected* provider is correctly rejected while an unrelated new provider is correctly ignored. A controlled fault injected at rename call 4 — after metadata install and pricing backup — restored metadata, pricing and snapshot byte-for-byte, leaving no `tmp`/`bak` debris. Targeted tests 5/5 green; exact-head `audit` / `test` / `package` all `completed/success`.

**Still open: [P1] the completeness guard only rejects an exactly-empty provider.**

`scripts/sync-model-metadata.mjs:189-195` validates:

```js
if (
  !provider.models ||
  typeof provider.models !== 'object' ||
  Array.isArray(provider.models) ||
  Object.keys(provider.models).length === 0
) {
  throw new Error(`models.dev provider ${sourceId} has no non-empty models object`);
}
```

A truncated or partially-failed upstream response is not empty — it is short. Reproduced against this head:

- removing 1 of 2 Anthropic models from the fetched catalog while leaving the provider non-empty: refresh succeeds and **publishes the deletion**;
- removing the surviving model's `cost`: refresh succeeds and **silently drops its pricing row**.

Both cases produce mutually self-consistent metadata + pricing + snapshot, so `--check` cannot detect them either — the three artifacts agree with each other precisely *because* they were all derived from the same truncated input. There is no comparison against the previously-published set anywhere in the script (no expected-model floor, no shrink guard, no diff-vs-previous), so a partial upstream response is indistinguishable from a genuine upstream removal.

Impact: a transient upstream degradation during a routine metadata refresh can quietly delete models or pricing from the shipped catalog, and the atomicity work added here makes that deletion land cleanly rather than fail.

Minimal direction: keep the emptiness check, and add a regression floor against the currently-committed outputs — e.g. fail closed when a selected provider's model set shrinks, or when a model that previously had `cost` no longer does, unless an explicit opt-in flag acknowledges the removal. That keeps intentional upstream deletions expressible while making silent truncation non-publishable.

Nothing else new at this head. Not a blocking request — flagging so the completeness axis is settled alongside the transaction axis before this lands.

jackwener (commented):
Kabi publisher review at exact head `c658f3c15fcfc80fa0b16eb9467de471ee953d52`.

Publisher execution: `kabi-sol-review-orchestrator`, representing `WAWQAQ` under Raft authorization `f380cfac` for Maka review actions. Hosted `audit`, `test`, and `package` are all `completed/success` on this exact head. **Verdict: one P2; please keep this head unapproved until it is fixed.** Two P3s are non-blocking.

### [P2] Capability loss bypasses the advertised fail-closed refresh contract

`scripts/sync-model-metadata.mjs:288-324` only records deleted object keys and deleted array members. Once recursion reaches a primitive, line 312 returns without comparing it. Therefore a routine upstream change such as:

- `reasoning: true -> false`, or
- `tool_call: true -> false` (projected as `functionCalling: true -> false`)

publishes successfully without `--accept-upstream-removals`.

This is not an arbitrary scalar edit. `README.md:247` explicitly promises that a refresh fails closed when a committed **capability disappears**, and these booleans are production authorities: `packages/core/src/provider-registry.ts:689-706` filters model catalogs by `functionCalling`, while `packages/core/src/model-metadata.ts:334-344` derives thinking controls from `reasoning`. A false transition can therefore remove a model from a tool-capable catalog or remove its thinking controls without the required explicit acknowledgment.

The current nested-shrink test only deletes a modality array member and a pricing key, so it cannot catch this primitive branch.

Minimal clean fix: treat `true -> false` under `/metadata/<provider>/<model>/capabilities/<name>` as a removal, while continuing to allow ordinary scalar corrections such as names and prices. Add production-helper regressions for both `reasoning` and `functionCalling`; the existing acceptance flag remains the explicit path for a reviewed capability withdrawal.

### [P3] Multi-file publish is exception-safe, not crash/concurrency-atomic

`scripts/sync-model-metadata.mjs:331-387` stages every file and rolls back caught exceptions, which correctly closes the prior failure path. But installation still consists of sequential renames at lines 353-362. SIGKILL between renames, or two concurrent refreshes, can leave snapshot/metadata/pricing from different generations because rollback only runs in `catch` and there is no inter-process lock. The current test (`scripts/sync-model-metadata.test.mjs:366-419`) covers a staging failure before installation, not process death or concurrent commits. This is recoverable maintainer-script state and remains P3; document the recovery, or add a same-directory lock if stronger serialization is desired.

### [P3] Chinese contribution docs omit the new safety contract

`README.zh-CN.md:244` documents refresh versus offline sync but omits both fail-closed shrinkage behavior and `--accept-upstream-removals`, while `README.md:247` contains them. The CLI error is explicit, so this is documentation asymmetry rather than a behavioral blocker.

The source-candidate provenance/legal changes and the single committed-snapshot authority otherwise hold. No merge action taken.

jackwener (approved):
Kabi publisher approval at exact head `3c2ace9b30b3d8ab6577319b4c582c327605ad91`.

Execution: `kabi-sol-review-orchestrator`, representing `WAWQAQ` under Raft authorization `f380cfac` for Maka review actions.

The prior capability-loss P2 is closed. `collectProjectionRemovals()` now recognizes only `true -> false` transitions at `/metadata/<provider>/<model>/capabilities/<field>` as acknowledged removals: both reasoning and function-calling capability withdrawals fail closed without `--accept-upstream-removals`, the explicit flag accepts them, and ordinary scalar corrections such as names and prices remain allowed. Regression coverage exercises both production-helper paths. The Chinese README now carries the same fail-closed and explicit-bypass contract.

Focused `sync-model-metadata` tests pass 8/8 locally. Hosted `test`, `audit`, and `package` are all `completed/success` on this exact head. PR is OPEN, non-draft, and MERGEABLE. No live P0-P2 finding remains.

No merge action taken; merging remains for a human committer.

PR #3571, opened 2026-08-23
fix(storage): clear obsolete onboarding intent

### LINKED ISSUE #3566: fix(storage): a non-convergent onboarding intent permanently bricks the runtime-policy store
## Problem

Crash recovery treats a connection-onboarding intent that cannot be replayed as fatal, and the failure is permanent: the intent journal is never cleared on that path, so **every subsequent open of the interactive runtime-policy store fails the same way**. The app cannot start against that storage root again without manually deleting `runtime-policy-onboarding.json`.

Mechanism (line refs at `main@e05308be7`):

- `openInteractiveRuntimePolicyStoresForWrite` runs `coordinator.recoverForWrite()` before handing out any writer (`packages/storage/src/runtime-policy-stores.ts:192`).
- `recoverForWrite` → `recoverConnectionOnboarding` (`coordinator.ts:209`, `:1317`) replays the journaled intent through `applyConnectionOnboarding` → `prepareOnboardingUpsert`.
- When the catalog has drifted so that the slug the intent resolves to is occupied by a connection with a **different** `connectionId`, the upsert throws `invalid_document` — "Onboarding intent conflicts with the connection id" (`connection-catalog-document.ts:534`) — which recovery wraps as `commit_outcome_unknown` — "Connection onboarding recovery did not converge" (`coordinator.ts:1330`) and rethrows.
- Nothing clears the intent on this path, so the next open replays the same intent into the same conflict. Verified experimentally: three successive opens with fresh root leases all fail identically.

## Reproduction (verified against real stores)

1. Onboard a connection so an intent is journaled, and hard-kill the process between `writeConnectionOnboardingIntent` and `clearConnectionOnboardingIntent` (the crash window the journal exists for).
2. Out of band, delete and re-create the connection at the same slug (new `connectionId`) — e.g. by another build, a restored backup, or manual edit of `connection-catalog.json`.
3. Reopen the store: `commit_outcome_unknown / invalid_document: Onboarding intent conflicts with the connection id`, permanently.

Reaching the state requires a hard kill **plus** out-of-band catalog drift, so it is rare — but the blast radius (the store never opens again) is disproportionate to the fault (one stale journal entry describing a row that provably no longer exists).

## Why fail-closed is the wrong shape here

Fail-closed is right for I/O errors and undecodable documents: retrying may converge and throwing protects unknown state. This case is different — the conflict **proves** the intent is obsolete: the connection it named is gone, and its slug now belongs to a different identity. Replaying it can never become correct, so refusing to open the store protects nothing; it converts a completed-or-abandoned onboarding into a bricked installation.

## Suggested direction

In `recoverConnectionOnboarding`, treat the id/slug-conflict (`invalid_document` from the upsert's identity cross-check) as an **obsolete intent**: log it, `clearConnectionOnboardingIntent`, and continue opening. Keep the current fail-closed behavior for every other error class (I/O failures, undecodable journal, vault errors), where retry can still converge. A regression test can hand-write the drifted catalog + stale journal and assert the store opens and the journal is cleared.

## Provenance

Found during adversarial review of #3467 (which adds one more input to the same pre-existing throw sites but does not change the recovery semantics); reproduced there through a path untouched by that PR. Details in that PR's review-record comment.

mikemikimike:
## Issue
Fixes #3566.

## Background
A stale connection-onboarding journal can become permanently unreplayable when the catalog contains the same provider slug with a different connection ID. Every subsequent interactive runtime-policy store open then fails.

## Changes
- Treat the exact connection-id conflict as an obsolete onboarding intent.
- Clear that intent and allow the store to open.
- Preserve fail-closed recovery for all other errors.
- Add a regression test covering reopen success and journal cleanup.

## Compatibility
Only stale intents with the explicit identity conflict are discarded. I/O failures, malformed journals, vault failures, and other recovery errors retain their existing fail-closed behavior.

## Verification
- `npm --workspace @maka/core run build` — passed
- `npm --workspace @maka/storage run build` — passed
- `node --test packages/storage/dist/__tests__/runtime-policy-stores.test.js` — 47 passed, 6 skipped
- `npx biome check packages/storage/src/runtime-policy/coordinator.ts packages/storage/src/__tests__/runtime-policy-stores.test.ts` — passed

## Notes
The skipped tests require POSIX permissions or symlink support unavailable in the Windows environment. Full repository build/lint were not run because this change is scoped to the storage workspace and its focused checks were sufficient for the affected path.

Generated-by: Codex

### DIFF
--- a/packages/storage/src/__tests__/runtime-policy-stores.test.ts
+++ b/packages/storage/src/__tests__/runtime-policy-stores.test.ts
@@ -3372,6 +3372,16 @@ describe('runtime policy stores', () => {
       } finally {
         await successor.close();
       }
+
+      const reopened = await tryAcquireInteractiveRootOwner(capability);
+      assert.ok(reopened);
+      if (!reopened) return;
+      try {
+        await openInteractiveRuntimePolicyStoresForWrite(reopened.lease);
+        assert.equal(existsSync(join(root, 'runtime-policy-onboarding.json')), false);
+      } finally {
+        await reopened.close();
+      }
     });
   });
 
--- a/packages/storage/src/__tests__/runtime-policy-stores.test.ts
+++ b/packages/storage/src/__tests__/runtime-policy-stores.test.ts
@@ -3331,6 +3331,50 @@ describe('runtime policy stores', () => {
     });
   });
 
+  test('clears a stale onboarding intent when its connection id conflicts with the catalog', async () => {
+    await withInteractiveRoot(async ({ root, capability }) => {
+      const owner = await tryAcquireInteractiveRootOwner(capability);
+      assert.ok(owner);
+      if (!owner) return;
+      try {
+        const stores = await openInteractiveRuntimePolicyStoresForWrite(owner.lease);
+        const connection = await createConnection(
+          stores,
+          0,
+          connectionDraft('openai', 'openai', 'Stale onboarding'),
+        );
+        await writeFile(
+          join(root, 'runtime-policy-onboarding.json'),
+          `${JSON.stringify({
+            schemaVersion: 1,
+            connectionId: '11111111-1111-4111-8111-111111111111',
+            providerType: connection.providerType,
+            suppliedSecret: null,
+            enabledModelIds: connection.enabledModelIds,
+            discovery: {
+              models: [{ id: 'gpt-5' }],
+              source: 'fetched',
+              fetchedAt: 1_800_000_000_000,
+            },
+            invalidateLastTest: false,
+          })}\n`,
+        );
+      } finally {
+        await owner.close();
+      }
+
+      const successor = await tryAcquireInteractiveRootOwner(capability);
+      assert.ok(successor);
+      if (!successor) return;
+      try {
+        await openInteractiveRuntimePolicyStoresForWrite(successor.lease);
+        assert.equal(existsSync(join(root, 'runtime-policy-onboarding.json')), false);
+      } finally {
+        await successor.close();
+      }
+    });
+  });
+
   test('interactive OAuth login commits only against its frozen connection and credential basis', async () => {
     await withInteractiveOwner(async ({ root, stores }) => {
       const claude = await createConnection(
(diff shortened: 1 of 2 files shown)


Astro-Han (approved):
Independent review at exact head `bdc1d702dc34b254748cbfac4a38dbd6a2cb0c86`.

**APPROVE.** The recovery path is correct and the regression test is the right shape.

The fix closes a real dead end: a `runtime-policy-onboarding.json` whose `connectionId` no longer matches the catalog previously escaped as `commit_outcome_unknown`, and because the intent file was never cleared, every subsequent open hit the same wall. Clearing the obsolete intent and returning makes the state self-healing, and the test proves it across *two* successive reopens rather than just one — which is what distinguishes "recovered" from "recovered once".

The narrowing is also correctly ordered: the obsolete-intent branch is checked before `isCommitOutcomeUnknown`, so genuinely unknown commit outcomes still propagate untouched. No behaviour change for any other `invalid_document` failure.

`[P3]` — `packages/storage/src/runtime-policy/coordinator.ts:1399-1405` discriminates on free-text message equality:

```ts
error.code === 'invalid_document' &&
error.message === 'Onboarding intent conflicts with the connection id'
```

`invalid_document` is raised from a dozen sites in `connection-catalog-document.ts` alone, so the message string is carrying the entire discrimination load for a control-flow decision that swallows an error. Rewording the throw at `connection-catalog-document.ts:534` would silently restore the old dead end.

Mitigating this materially: the new test drives the real thrower end-to-end rather than constructing the error, so a reworded message does fail the suite. That is why this is `[P3]` and not higher — the coupling is covered, just implicitly. If you want it explicit, a dedicated error code (or an exported message constant shared by both sites) would make the contract local instead of spread across two files. Fine to leave as-is.

Verification: exact-head `test` is `completed/success`. The `Dependency audit` workflow does not appear on this head, which is correct rather than missing — it is path-filtered to `package.json` / lockfile / audit scripts, and this PR touches neither.

PR #3578, opened 2026-08-23
fix(runtime): collapse duplicate workspace instruction files in one directory

### LINKED ISSUE #3577: bug(runtime): a CLAUDE.md symlinked to AGENTS.md is injected into the prompt twice
## What happened

A repository that follows the mainstream way of sharing one instruction file across agent CLIs gets that file injected into Maka's system prompt twice.

`readWorkspaceInstructions` reads every candidate name in a directory and never compares the results:

```ts
// packages/runtime/src/system-prompt/workspace-instructions.ts
export const WORKSPACE_INSTRUCTION_FILES = ['AGENTS.md', 'CLAUDE.md', 'GEMINI.md'] as const;

for (const file of WORKSPACE_INSTRUCTION_FILES) {
  const candidate = join(root, file);
  const resolved = await realpath(candidate);
  if (!isPathInside(root, resolved)) continue;
  const raw = await readFile(resolved, 'utf8');
  out.push({ file, scope, text: ..., truncated: ... });
}
```

`realpath` is used only for the containment check; the entry is still recorded under the original name, so a `CLAUDE.md` symlinked to `AGENTS.md` resolves to the same file, is read again, and is appended again.

Measured against a checkout with that symlink, by calling `buildWorkspaceInstructionsPromptFragment` on the repo root:

```
without symlink   blocks = project/AGENTS.md
with symlink      blocks = project/AGENTS.md, project/CLAUDE.md
```

Both copies count against `MAX_WORKSPACE_INSTRUCTIONS_PROMPT_CHARS` (14000). With a file near the 6000-character per-file cap, the duplicate consumes most of the workspace-instruction budget, and any genuinely different file in the same directory is what gets dropped.

This is not an exotic setup. Claude Code's documentation recommends exactly it:

> Claude Code reads `CLAUDE.md`, not `AGENTS.md`. […] A symlink also works if you don't need to add Claude-specific content: `ln -s AGENTS.md CLAUDE.md`

`apache/airflow` ships that symlink, as does `deepseek-ai/deepseek-harness`.

The `@AGENTS.md` import form that the same documentation recommends fails differently: Maka does not expand imports, so a one-line `CLAUDE.md` containing `@AGENTS.md` is injected as the literal text `@AGENTS.md`. Harmless in size, but it is an instruction that means nothing.

## How to reproduce

1. In any workspace with an `AGENTS.md`, add `ln -s AGENTS.md CLAUDE.md`.
2. Call `buildWorkspaceInstructionsPromptFragment(cwd)`.
3. The fragment contains two `<workspace-instructions>` blocks with identical bodies.

## Environment

- Maka commit: `729839ed8`
- Surface: any — the fragment is shared by Desktop, TUI and CLI

### How other agents avoid this

Three distinct strategies, and Maka currently uses none of them:

| Agent | Files read | Conflict resolution |
|---|---|---|
| Claude Code | `CLAUDE.md` only | not possible |
| Codex CLI | `AGENTS.md` only | not possible |
| Gemini CLI | `GEMINI.md` only | not possible |
| opencode | `AGENTS.md`, `CLAUDE.md` | first match wins — "if you have both `AGENTS.md` and `CLAUDE.md`, only `AGENTS.md` is used" |
| Pi | 5 candidate names | first match wins |
| DeepSeek Harness | several | content-digest deduplication |

Pi's loader returns on the first existing candidate per directory (`packages/coding-agent/src/core/resource-loader.ts`), and carries a separate guard against a linked worktree applying the same context twice — the same class of problem.

### Suggested direction

DeepSeek Harness is the closest analogue: it reads several names *and* ships the `CLAUDE.md` symlink, and reconciles them by content rather than by path. From `packages/context/agent-instructions/src/files.ts`:

> Drop later candidates whose trimmed content duplicates an earlier sibling in the same directory. Different directories never collapse even when identical; within one directory the earliest candidate in discovery order is kept and its original bytes are rendered. **A candidate that symlinks a sibling resolves to the same content and collapses here like any byte-identical real file.**

Deduplicating on the trimmed content digest, scoped to one directory, looks like the right shape for Maka too:

- it catches the symlink *and* two byte-identical real files, which a `realpath` check would miss;
- it keeps different directories independent, so the global/project layering still works;
- it changes nothing for the common case of a single instruction file.

First-match-wins would also work and is simpler, but it silently discards a `CLAUDE.md` that genuinely differs from `AGENTS.md`, which is a supported thing to have. Deduplication only drops what is redundant.

_Investigated with AI assistance (Claude Code); Maka behaviour measured on the commit above, other agents' behaviour read from their published docs and source._

shaokeyibb:
## Summary

`readWorkspaceInstructions` reads every candidate name in a directory and appends each result, so a repository that shares one instruction file across the names different agent CLIs read gets it injected twice. That sharing is the documented way to do it — Claude Code's docs recommend `ln -s AGENTS.md CLAUDE.md`, and `apache/airflow` and `deepseek-ai/deepseek-harness` both ship that link — and the duplicate then consumes most of the 14000-character workspace-instruction budget, squeezing out whatever genuinely different file sits beside it.

Deduplicate on the digest of the cleaned text, scoped to one directory. Digesting rather than comparing `realpath` catches a byte-identical copy as well as a link, which is the other common way people share these files. Directories stay independent, because the same text at global and project scope is a user repeating themselves on purpose and collapsing it would silently drop a layer. The digest is taken before truncation, so two files that diverge only past the per-file cap still count as different.

Fixes #3577

## Verification

Four new tests in `workspace-instructions.test.ts`: a symlinked `CLAUDE.md` collapses, a byte-identical copy collapses, files in one directory that genuinely differ are both kept, and identical text at global and project scope is kept at both scopes. The first two fail on `main` with `expected: 1, actual: 2`; the last two pass either way and are there so the fix cannot over-collapse.

End-to-end against a directory holding a real `AGENTS.md`, in both duplicate forms:

```
before   blocks = project/AGENTS.md, project/CLAUDE.md
after    blocks = project/AGENTS.md
```

Ran: `workspace-instructions` (9) and `ai-sdk-backend` (204) green; `biome check` on both changed files; `npm run check:asf-headers`.

`packages/runtime` typecheck reports one error in `responses-wire-contract.test.ts` about an `alibaba-cn` provider literal. Verified pre-existing — the same error appears on a clean `origin/main` build, and this change touches neither file.

Not run: the rest of the runtime suite, Desktop, Playwright.

## Notes

`@AGENTS.md` — the import form Claude Code's docs also suggest — is not addressed here. Maka does not expand imports, so a one-line `CLAUDE.md` containing `@AGENTS.md` is injected as that literal text. It is small and harmless rather than wasteful, but it is also an instruction that means nothing; worth a separate decision about whether to expand or ignore it.

For reference, the three strategies other agents use: read a single filename (Claude Code, Codex CLI, Gemini CLI), first-match-wins across a candidate list (opencode, Pi), or content deduplication (deepseek-harness). First-match-wins would be simpler here, but it discards a `CLAUDE.md` that genuinely differs from `AGENTS.md`, which is a supported thing to have.

## AI use

Select exactly one:

Tool(s) and scope: Claude Code — investigation, comparison of other agents' loaders, the tests, and the fix.

## Checklist

- [ ] No

### DIFF
--- a/packages/runtime/src/system-prompt/workspace-instructions.ts
+++ b/packages/runtime/src/system-prompt/workspace-instructions.ts
@@ -17,6 +17,7 @@
  * under the License.
  */
 
+import { createHash } from 'node:crypto';
 import { readFile, realpath } from 'node:fs/promises';
 import { homedir } from 'node:os';
 import { join } from 'node:path';
@@ -107,6 +108,16 @@ async function readWorkspaceInstructions(
   }
 
   const out: WorkspaceInstruction[] = [];
+  // Content already accepted from this directory. Sharing one instruction file
+  // across the names different agent CLIs read — Claude Code reads CLAUDE.md,
+  // Codex reads AGENTS.md, Gemini CLI reads GEMINI.md — is the documented way
+  // to do it, whether by symlink or by copy, so identical bytes arriving under
+  // two names here are redundancy rather than two instructions. Digesting the
+  // cleaned text catches both forms; `realpath` alone would miss the copy.
+  // Scoped to one directory on purpose: the same text at global and project
+  // scope is a user repeating themselves deliberately, and collapsing that
+  // would silently drop a layer.
+  const seenDigests = new Set<string>();
   for (const file of WORKSPACE_INSTRUCTION_FILES) {
     const candidate = join(root, file);
     let resolved: string;
@@ -120,6 +131,11 @@ async function readWorkspaceInstructions(
       const raw = await readFile(resolved, 'utf8');
       const cleaned = cleanPromptText(raw.trim());
       if (!cleaned) continue;
+      // Digest before truncation: two files that diverge only past the cap are
+      // still different instructions.
+      const digest = createHash('sha256').update(cleaned).digest('hex');
+      if (seenDigests.has(digest)) continue;
+      seenDigests.add(digest);
       const chars = Array.from(cleaned).length;
       out.push({
         file,
--- a/packages/runtime/src/__tests__/workspace-instructions.test.ts
+++ b/packages/runtime/src/__tests__/workspace-instructions.test.ts
@@ -124,7 +124,79 @@ describe('workspace instructions prompt fragment', () => {
       assert.doesNotMatch(prompt, /PROJECT_SHOULD_BE_SQUEEZED/);
     });
   });
+
+  it('collapses a CLAUDE.md symlinked to AGENTS.md into one block', async () => {
+    // Sharing one instruction file across agent CLIs by symlinking the names
+    // each of them reads is the documented way to do it, so the same bytes
+    // arriving twice must not be injected twice.
+    await withWorkspaceAndHome(async ({ workspaceRoot, homeDir }) => {
+      await writeFile(join(workspaceRoot, 'AGENTS.md'), 'SHARED_RULE\n', 'utf8');
+      await symlink(join(workspaceRoot, 'AGENTS.md'), join(workspaceRoot, 'CLAUDE.md'));
+
+      const prompt = await buildWorkspaceInstructionsPromptFragment(workspaceRoot, { homeDir });
+
+      assert.ok(prompt);
+      assert.equal(countBlocks(prompt), 1);
+      assert.equal(occurrences(prompt, 'SHARED_RULE'), 1);
+      assert.match(prompt, /file="AGENTS\.md"/);
+    });
+  });
+
+  it('collapses byte-identical instruction files that are not links', async () => {
+    // Copying rather than linking is the other common way to share one set of
+    // rules; it is the same redundancy and deserves the same treatment.
+    await withWorkspaceAndHome(async ({ workspaceRoot, homeDir }) => {
+      await writeFile(join(workspaceRoot, 'AGENTS.md'), 'COPIED_RULE\n', 'utf8');
+      await writeFile(join(workspaceRoot, 'CLAUDE.md'), 'COPIED_RULE\n', 'utf8');
+
+      const prompt = await buildWorkspaceInstructionsPromptFragment(workspaceRoot, { homeDir });
+
+      assert.ok(prompt);
+      assert.equal(countBlocks(prompt), 1);
+      assert.equal(occurrences(prompt, 'COPIED_RULE'), 1);
+    });
+  });
+
+  it('keeps instruction files in one directory that genuinely differ', async () => {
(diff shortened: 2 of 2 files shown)


packages/runtime/src/system-prompt/workspace-instructions.ts
```diff
@@ -107,6 +108,16 @@ async function readWorkspaceInstructions(
...
+  // Content already accepted from this directory. Sharing one instruction file
+  // across the names different agent CLIs read — Claude Code reads CLAUDE.md,
+  // Codex reads AGENTS.md, Gemini CLI reads GEMINI.md — is the documented way
+  // to do it, whether by symlink or by copy, so identical bytes arriving under
+  // two names here are redundancy rather than two instructions. Digesting the
+  // cleaned text catches both forms; `realpath` alone would miss the copy.
+  // Scoped to one directory on purpose: the same text at global and project
+  // scope is a user repeating themselves deliberately, and collapsing that
+  // would silently drop a layer.
+  const seenDigests = new Set<string>();
```
jackwener, line 120: [P3] This set is recreated for each scope read, so the same physical directory is not actually deduplicated when it is both the global and project root. A production-function repro with `cwd === <home>/.maka` and one `<home>/.maka/AGENTS.md` produces `blocks=2`, `copies=2`, and scopes `[global, project]`: `buildWorkspaceInstructionsPromptFragment` calls this function twice, and each call starts with an empty set. This is inside the PR's stated ‘one directory’ contract, not an unrelated enhancement. It is non-blocking because ordinary project roots differ from `~/.maka`, but editing the config directory (or a realpath alias of it) still reproduces the duplicate-budget bug. Please share dedupe state by resolved root, or special-case identical resolved roots, and add the same-root regression.

jackwener (approved):
Acting on behalf of repository owner WAWQAQ under the standing Kabi authorization for PR review and APPROVE; exact head `1042a9d7546a3540859e1e905901dd106c35947b`.

**GO: no P0–P2; one non-blocking P3 inline.** Full-patch coverage included digest scope/order, pre-truncation identity, prompt budgets, symlink containment, current-main merge compatibility, and all three GitHub discussion endpoints. Focused tests passed 9/9; removing the sole dedupe branch made exactly the two dedupe regressions fail; restored full Runtime suite passed 3,079 with 13 skips and 0 failures. Desktop and Playwright were not run in this pass.

Fresh machine gates: OPEN, non-draft, mergeable; exact-head `test` is completed/success. This agent-executed approval does not claim to satisfy the repository's separate required-human-review wording. No merge performed.

PR #3585, opened 2026-08-23
fix(runtime): avoid implicit computer-use images

hqhq1025:
## Summary

- keep automatically captured Computer Use frames available to the local PiP without sending them to the model for semantic actions
- include model-visible image blocks only for explicit visual requests and legacy coordinate actions
- preserve fresh accessibility observations for semantic actions such as `element_sequence`
- bind the provider-image decision to the immutable invocation that actually executed
- classify every canonical Computer Use action through an exhaustive typed screenshot policy

## Root cause

Mutating Computer Use actions attach a final screenshot for the desktop PiP. `toModelOutput` previously projected every attached screenshot into the next provider request, so a semantic `element_sequence` result carried an unnecessary base64 PNG even though it already returned a fresh accessibility observation. The first revision also evaluated the image gate from settlement-time caller input and represented the action catalogue as an untyped parallel set. The current revision derives the decision from the frozen executed input and carries it on the result, while an exhaustive `Record<CuToolActionType, ...>` makes every action an explicit policy choice.

## Measured result

A real Calculator run using `element_sequence` completed successfully. The final provider request dropped from 156,262 bytes to 73,727 bytes, contained no base64 string over 10 KB, and reduced the post-action input-token increase from 1,575 to 1,056 tokens.

## Verification

- built `@maka/core`, `@maka/storage`, and `@maka/mcp` in dependency order
- `npm --workspace @maka/runtime run typecheck`
- `node --test packages/runtime/dist/__tests__/computer-use-tools.test.js packages/runtime/dist/__tests__/computer-use-schema-parity.test.js` (89 passed)
- `npx biome check packages/runtime/src/computer-use-tools.ts packages/runtime/src/__tests__/computer-use-tools.test.ts`
- real Electron Computer Use run against Calculator with a 12-step semantic `element_sequence`

## AI use

Select exactly one:

Tool(s) and scope: Codex contributed implementation, tests, review follow-up, and verification. Both affected commits include `Generated-by: Codex` trailers.

## Checklist

- [ ] No

### DIFF
--- a/packages/runtime/src/computer-use-tools.ts
+++ b/packages/runtime/src/computer-use-tools.ts
@@ -35,6 +35,7 @@ import {
   isCuObservingAction,
   type CuAction,
   type CuPoint,
+  type CuToolActionType,
   type ComputerUseErrorCode,
   type ComputerUseWindowIdentity,
 } from '@maka/core/computer-use';
@@ -340,31 +341,46 @@ interface ComputerToolResult {
   error?: ComputerUseErrorCode;
   failureClass?: 'ambiguous_target';
   screenshot?: { base64: string; mimeType: string };
+  includeScreenshotInModelOutput?: boolean;
 }
 
-const MODEL_SCREENSHOT_ACTIONS: ReadonlySet<string> = new Set([
-  'screenshot',
-  'zoom',
-  'mouse_move',
-  'left_click',
-  'right_click',
-  'middle_click',
-  'double_click',
-  'triple_click',
-  'left_mouse_down',
-  'left_mouse_up',
-  'left_click_drag',
-  'type',
-  'key',
-  'hold_key',
-  'scroll',
-]);
+export const COMPUTER_USE_MODEL_SCREENSHOT_POLICY = {
+  list_apps: 'never',
+  launch_app: 'never',
+  observe: 'explicit',
+  click_element: 'never',
+  set_value: 'never',
+  select_text: 'never',
+  secondary_action: 'never',
+  scroll_element: 'never',
+  window_action: 'never',
+  element_sequence: 'never',
+  press_key: 'never',
+  screenshot: 'always',
+  cursor_position: 'never',
+  mouse_move: 'always',
+  left_click: 'always',
+  right_click: 'always',
+  middle_click: 'always',
+  double_click: 'always',
+  triple_click: 'always',
+  left_mouse_down: 'always',
+  left_mouse_up: 'always',
+  left_click_drag: 'always',
+  type: 'always',
+  key: 'always',
+  hold_key: 'always',
+  scroll: 'always',
+  wait: 'never',
+  zoom: 'always',
+} as const satisfies Record<CuToolActionType, 'always' | 'explicit' | 'never'>;
 
-function shouldSendScreenshotToModel(input: unknown): boolean {
-  if (!input || typeof input !== 'object') return false;
-  const record = input as { action?: unknown; include_screenshot?: unknown };
-  if (record.action === 'observe') return record.include_screenshot === true;
-  return typeof record.action === 'string' && MODEL_SCREENSHOT_ACTIONS.has(record.action);
+function shouldSendScreenshotToModel(input: ComputerParams): boolean {
+  const policy = COMPUTER_USE_MODEL_SCREENSHOT_POLICY[input.action];
+  return (
+    policy === 'always' ||
+    (policy === 'explicit' && input.action === 'observe' && input.include_screenshot === true)
+  );
 }
 
 export interface ComputerUseToolSet extends Array<MakaTool> {
@@ -1010,6 +1026,7 @@ export function buildComputerUseTools(deps: {
   function deliveredWithoutFreshObservation(
     action: ComputerSummaryAction,
     result: CuRunResult,
+    includeScreenshotInModelOutput = false,
   ): ComputerToolResult {
     const evidence = summarizeEvidence(result.outcome.evidence);
     const hostEvidence = summarizeEvidence(result.outcome.evidence, 'host');
@@ -1034,6 +1051,7 @@ export function buildComputerUseTools(deps: {
               base64: screenshot.base64,
               mimeType: screenshot.mimeType,
             },
+            includeScreenshotInModelOutput,
           }
         : {}),
     };
@@ -1622,6 +1640,7 @@ export function buildComputerUseTools(deps: {
     ): Promise<ComputerToolResult> => {
       if (abortSignal.aborted) return { text: 'computer aborted before start' };
       const input = snapshotComputerParams(computerParams.parse(args));
+      const includeScreenshotInModelOutput = shouldSendScreenshotToModel(input);
       // Before anything is claimed against a frame or dispatched: an argument
       // holding one of this host's own withheld-value placeholders is a replay
       // of the record, not a value, and every path below would have typed it.
@@ -1630,7 +1649,7 @@ export function buildComputerUseTools(deps: {
       const invocationGeneration = presentationGenerations.get(sessionId) ?? 0;
       const releasePendingInvocation = trackPendingInvocation(sessionId, turnId);
       try {
-        return await withInvocationQueue(sessionId, abortSignal, async () => {
+        return await withInvocationQueue<ComputerToolResult>(sessionId, abortSignal, async () => {
           if ((presentationGenerations.get(sessionId) ?? 0) !== invocationGeneration) {
             return sessionFailure('user_stopped');
           }
@@ -2301,6 +2320,7 @@ export function buildComputerUseTools(deps: {
                   text: persistedObservationText(observation),
                   modelText: observationText({ ...observation, screenshot }),
                   screenshot: { base64: screenshot.base64, mimeType: screenshot.mimeType },
+                  includeScreenshotInModelOutput,
                 }
               : {
                   text: persistedObservationText(observation),
@@ -2366,6 +2386,7 @@ export function buildComputerUseTools(deps: {
                 base64: screenshotObservation.screenshot.base64,
                 mimeType: screenshotObservation.screenshot.mimeType,
               },
+              includeScreenshotInModelOutput,
             };
           }
           if (
@@ -2767,11 +2788,19 @@ export function buildComputerUseTools(deps: {
                   : undefined;
             } catch {
               presentation?.finish(result);
-              return deliveredWithoutFreshObservation(modelAction, result);
+              return deliveredWithoutFreshObservation(
+                modelAction,
+                result,
+                includeScreenshotInModelOutput,
+              );
             }
             if (actionLease && result.outcome.ok && !freshObservation) {
               presentation?.finish(result);
-              return deliveredWithoutFreshObservation(modelAction, result);
+              return deliveredWithoutFreshObservation(
+                modelAction,
+                result,
+                includeScreenshotInModelOutput,
+              );
             }
             presentation?.finish(withMirrorFrame(result, freshObservation));
             const modelRefresh = freshObservation
@@ -2798,6 +2827,7 @@ export function buildComputerUseTools(deps: {
                   ...(!result.outcome.ok ? { error: result.outcome.error } : {}),
                   ...(failureClass ? { failureClass } : {}),
                   screenshot: { base64: screenshot.base64, mimeType: screenshot.mimeType },
+                  includeScreenshotInModelOutput,
                 }
               : {
                   text,
@@ -2815,7 +2845,7 @@ export function buildComputerUseTools(deps: {
     // return a fresh accessibility observation, so their automatically captured
     // PiP frame stays local. Explicit visual requests and legacy coordinate
     // actions still receive the native image block.
-    toModelOutput: ({ input, output }) => {
+    toModelOutput: ({ output }) => {
       const o = (output ?? {}) as Partial<ComputerToolResult> & { error?: unknown };
       const text =
         typeof o.modelText === 'string'
@@ -2829,7 +2859,7 @@ export function buildComputerUseTools(deps: {
         type: 'content',
         value: [
           { type: 'text', text },
-          ...(o.screenshot && shouldSendScreenshotToModel(input)
+          ...(o.screenshot && o.includeScreenshotInModelOutput === true
             ? [
                 {
                   type: 'file' as const,
--- a/packages/runtime/src/computer-use-tools.ts
+++ b/packages/runtime/src/computer-use-tools.ts
@@ -330,8 +330,9 @@ export const computerWireParams = z
  * Raw result of the `computer` tool. `text` is the S16-safe summary the runtime
  * records to session history (via coerceResultContent's text-only projection:
  * this object has no `kind`, so only `text` survives). `screenshot`, when
- * present, rides along ONLY to feed `toModelOutput` — it never enters `text`, so
- * the bounded frame base64 stays out of session history.
+ * present, feeds the local presentation layer and, only for explicitly visual
+ * actions, `toModelOutput`. It never enters `text`, so the bounded frame base64
+ * stays out of session history.
  */
 interface ComputerToolResult {
   text: string;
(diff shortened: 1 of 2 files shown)


Astro-Han (commented):
Reviewed at exact head `99c27eb2be4563eefa6ccbe7d1d369d9488d2bcb`.

**Coverage:** what decides whether a screenshot reaches the model, which paths this change moves from implicit-attach to no-attach, and whether anything that *should* still get a frame loses one. **Not covered:** wording.

The intent is right and the observable behaviour matches it: the decision is per tool call, not per turn or session; `observe: true`, `screenshot`, `zoom`, and the legacy coordinate actions still carry an image, while other semantic actions keep a fresh accessibility observation without the frame. Two problems in how that decision is made.

### [P1] The image gate reads mutable caller input, not the frozen call it executed

`packages/runtime/src/computer-use-tools.ts:2818-2832` — `toModelOutput: ({ input, output })` calls `shouldSendScreenshotToModel(input)`, and that `input` is the original caller-held object at settlement time, not the immutable snapshot the implementation actually ran with.

Reproduced: begin an async settlement with `observe` / `include_screenshot: false`; after `impl` has started, mutate the original object to `true`; the returned result carries the PiP screenshot and the provider `modelOutput` then contains a `file` block. The reverse mutation swallows an image the caller explicitly asked for. The PiP frame a semantic action carries can be reclassified the same way before settlement — which walks straight through the boundary this PR exists to establish.

Cloning inside the callback would already be too late. The fix is for `toModelOutput` to use the same verified immutable call, or for `impl` to emit an explicit model-image decision derived from the frozen input.

### [P2] The action catalogue is untyped and parallel to the canonical union

`computer-use-tools.ts:345-367` introduces `MODEL_SCREENSHOT_ACTIONS` as a `ReadonlySet<string>` sitting alongside `CU_ACTION_TYPES` / `CU_TOOL_ACTION_TYPES`, and the tests sample only five inputs.

The consequence is a silent default: a legitimately added action will simply not send an image, while every existing schema and parity test still passes. That is precisely the drift the comments near core warn about. Deriving the set exhaustively from the canonical action union — a typed map or derived set — and iterating every action in a classification test would make a new action a compile-time or test-time decision instead of a silent one.

### Verification and limits

Built clean in dependency order (core → storage → mcp → runtime). Computer Use suites 261/261; Biome and `git diff --check` pass. Across the whole Runtime: 3064 pass, 1 fail, 13 skipped — the single failure is a macOS Bash executable-root environment assertion in untouched code, reproducible standalone, and hosted CI is green on this exact head.

packages/runtime/src/computer-use-tools.ts
```diff
@@ -2803,7 +2829,7 @@ export function buildComputerUseTools(deps: {
         type: 'content',
         value: [
           { type: 'text', text },
-          ...(o.screenshot
+          ...(o.screenshot && shouldSendScreenshotToModel(input)
```
Astro-Han: [P1] Gate the image using the same immutable invocation that `impl` executed, not this settlement-time caller object. `input` here is still the caller-held reference: mutating `include_screenshot` or `action` after `impl` starts can add a provider `file` block to a call that executed with images disabled, or remove one from an explicitly visual call. Cloning here is already too late; derive an explicit model-image decision from the frozen input inside execution (or pass the verified immutable call through) and consume that decision here.

packages/runtime/src/computer-use-tools.ts
```diff
@@ -341,6 +342,31 @@ interface ComputerToolResult {
   screenshot?: { base64: string; mimeType: string };
 }
 
+const MODEL_SCREENSHOT_ACTIONS: ReadonlySet<string> = new Set([
```
Astro-Han: [P2] Make this classification exhaustive against the canonical Computer Use action union. A `ReadonlySet<string>` is a parallel untyped catalogue, so adding a legitimate action silently defaults it to no model image while schemas and sampled tests still pass. Use a typed exhaustive record/derived set and iterate every action in the classification test so each new action requires an explicit compile-time or test-time choice.

Astro-Han (commented):
Incremental closure check at exact head `99c27eb2be4563eefa6ccbe7d1d369d9488d2bcb`: the head has not changed since review `5002157364`, so both previously reported blockers remain. Reposting them inline at their exact diff authorities; no APPROVE on this head.

packages/runtime/src/computer-use-tools.ts
Astro-Han: **[P1] The image gate reads settlement-time input, not the invocation `impl` actually ran.**

`impl` freezes its arguments at `:1624` (`const input = snapshotComputerParams(computerParams.parse(args))`), so everything it does is bound to that snapshot. But this gate calls `shouldSendScreenshotToModel(input)` where `input` is `toModelOutput`'s parameter, and the runtime passes the live caller object: `tool-runtime.ts:721-723` calls `toModelOutput({ toolCallId, input: call.input, output: settlement.result })`.

So `include_screenshot` or `action` mutated on the caller after `impl` starts still changes whether a provider `file` block is attached — either adding an image the executed invocation did not authorize, or dropping one it did. `ComputerToolResult` carries `screenshot` but no record of whether that screenshot was cleared for the model, so nothing downstream can recover the executed intent.

Cloning inside this callback does not help; by then the caller has already been read. Either stamp the send/don't-send decision onto the result inside `impl`, where the frozen snapshot is in scope, or thread that snapshot through to here.

<details>
<summary>中文</summary>

**[P1] 这个门禁读的是结算时刻的输入，不是 `impl` 实际执行的那次调用。**

`impl` 在 `:1624` 用 `snapshotComputerParams` 把参数冻住了，但这里 `shouldSendScreenshotToModel(input)` 的 `input` 是 `toModelOutput` 的入参，而 `tool-runtime.ts:721-723` 传进来的是 caller 的活对象 `call.input`。因此 `impl` 开始之后再改 caller 上的 `include_screenshot` 或 `action`，仍然能加上或吞掉一个 `file` 块。`ComputerToolResult` 里有 `screenshot`，却没有记录它是否被批准发给模型。

在这个回调里再克隆已经晚了。要么在 `impl` 内部就把"发不发"的判断盖在结果上，要么把那份快照传到这里。
</details>

Astro-Han (commented):
Thanks — narrowing which Computer Use results carry an image into the model is worth doing, and `test` is green on `e7205fa5`.

I re-checked the two earlier points against this head. Both still apply; details are inline.

<details>
<summary>中文</summary>

感谢——收窄哪些 Computer Use 结果会把图片带进模型是值得做的，`e7205fa5` 上 `test` 已绿。

之前两条我对着这个 head 重新核过，都仍然成立，具体见行内。
</details>

packages/runtime/src/computer-use-tools.ts
```diff
@@ -341,6 +342,31 @@ interface ComputerToolResult {
...
+  'screenshot',
```
Astro-Han: **[P2] This classification is an untyped string set, so actions that already exist get no image and new ones will silently do the same.**

`MODEL_SCREENSHOT_ACTIONS` is a `ReadonlySet<string>`, which means it is never checked against the canonical action union. That union is `CU_TOOL_ACTION_TYPES` (`COMPUTER_USE_SEMANTIC_ACTIONS` + `CU_ACTION_TYPES`, `packages/core/src/computer-use.ts:184-268`), and `cursor_position` and `wait` are in it but missing here — they take the no-image path today.

Because the type is `string`, adding an action to the union compiles cleanly and defaults to sending no image. The classification test only exercises five inputs (`computer-use-tools.test.ts:2612-2634`), so it would not catch that either.

A typed exhaustive mapping keyed by the action union would make an omission a compile error, and iterating every union member in that test would make it a test failure.

<details>
<summary>中文</summary>

**[P2] 这个分类是无类型字符串集合，已有的 action 就漏了，新增的也会静默地漏。**

`MODEL_SCREENSHOT_ACTIONS` 是 `ReadonlySet<string>`，永远不会和规范 union 对账。union 是 `CU_TOOL_ACTION_TYPES`（见 `packages/core/src/computer-use.ts:184-268`），其中 `cursor_position` 和 `wait` 已经存在但不在这个集合里，现在走的是不发图那条路。

因为类型是 `string`，往 union 里加新 action 编译照过，默认不发图；分类测试只跑了 5 个输入（`computer-use-tools.test.ts:2612-2634`），也拦不住。

改成以 action union 为键的穷举映射，漏一个就是编译错误；测试里遍历 union 的每个成员，漏一个就是测试失败。
</details>

packages/runtime/src/computer-use-tools.ts
hqhq1025: Fixed at head `8e660adab`. `impl` now derives `includeScreenshotInModelOutput` from the parsed immutable `snapshotComputerParams(...)` value and carries that decision on the tool result. `toModelOutput` no longer reads settlement-time caller input. The regression mutates `include_screenshot` in both directions after execution and verifies the provider output still follows the executed snapshot.

packages/runtime/src/computer-use-tools.ts
hqhq1025: Fixed at head `8e660adab`. The policy is now an exhaustive `Record<CuToolActionType, always | explicit | never>` keyed by the canonical action union, and the test compares every policy key with `CU_TOOL_ACTION_TYPES`. A newly added action must now make an explicit compile-time classification and remain present at runtime.

packages/runtime/src/computer-use-tools.ts
hqhq1025: Fixed at head `8e660adab`. The model-image decision is stamped onto `ComputerToolResult` inside `impl` from the frozen parsed invocation. The settlement projection consumes only that result field, never `call.input`. The new two-direction mutation regression passes.

packages/runtime/src/computer-use-tools.ts
hqhq1025: Fixed at head `8e660adab`. `COMPUTER_USE_MODEL_SCREENSHOT_POLICY` is exhaustive over `CuToolActionType`, explicitly classifies `cursor_position` and `wait` as `never`, and is checked against every member of `CU_TOOL_ACTION_TYPES`. Focused Runtime build and 89 tests pass.

Astro-Han (commented):
The behaviour change is right and the test covers the interesting cases — semantic actions and `observe` without a picture stay text-only, while `screenshot`, an explicit `include_screenshot`, and the coordinate actions still get the image block.

One thing to fix before this merges.

**[P2] `MODEL_SCREENSHOT_ACTIONS` restates `CU_ACTION_TYPES` by hand, and nothing holds the two together**

The new set in `packages/runtime/src/computer-use-tools.ts:345-360` is exactly `CU_ACTION_TYPES` minus `cursor_position` and `wait`. It is written out as fifteen string literals with no link back to the source list and no test comparing them.

That is the drift this area has already paid for once. `packages/core/src/computer-use.ts:227-251` documents it directly: a hand-written list beside a schema that listed the same names fell behind when `window_action` was added, and the fix was to stop hand-writing it and add `computer-use-schema-parity.test.ts` to compare the lists in both directions. The comment also makes the point that matters here — drift where a name is missing is "just as costly and quieter."

The failure mode is worse than the original one, because it is silent in both directions. Add a coordinate action and it reaches the model with no picture, which looks like a model that cannot see rather than a missing list entry; no test fails. Remove or rename one and the dead entry never announces itself.

Deriving the set removes the class of bug rather than the instance:

```ts
const NON_VISUAL_COORDINATE_ACTIONS = new Set(['cursor_position', 'wait']);
const MODEL_SCREENSHOT_ACTIONS: ReadonlySet<string> = new Set(
  CU_ACTION_TYPES.filter((action) => !NON_VISUAL_COORDINATE_ACTIONS.has(action)),
);
```

Now adding a coordinate action opts it into vision by default, and deliberately withholding a picture is a one-line, reviewable edit to the exclusion set. If you would rather keep the explicit list, please extend `computer-use-schema-parity.test.ts` to assert this set against `CU_ACTION_TYPES` so drift fails a test instead of quietly blinding the model.

<details>
<summary>简体中文</summary>

行为改动是对的，测试也覆盖了关键分支：语义动作和不带图的 `observe` 只返回文本，而 `screenshot`、显式的 `include_screenshot`、以及坐标类动作仍然拿到图像块。

合并前有一处需要修。

**[P2] `MODEL_SCREENSHOT_ACTIONS` 是手写重述 `CU_ACTION_TYPES`，且没有任何东西保证两者一致**

`packages/runtime/src/computer-use-tools.ts:345-360` 新增的这个集合，正好等于 `CU_ACTION_TYPES` 去掉 `cursor_position` 和 `wait`。它被写成十五个字符串字面量，既没有回指来源列表，也没有测试比对两者。

**这正是这块代码已经付过一次学费的漂移。** `packages/core/src/computer-use.ts:227-251` 明确记录了这件事：一份手写列表与一份列着相同名字的 schema 并存，结果在新增 `window_action` 时落后了；修法是不再手写，并加上 `computer-use-schema-parity.test.ts` 双向比对两份列表。那段注释还点出了此处真正要害的一句——**少写一个名字的漂移代价一样大，而且更安静。**

这里的失败模式比当初那次更糟，因为它两个方向都是静默的。新增一个坐标动作，它会带不到图就进模型，表现出来像是"模型看不见"而不是"列表少了一项"，**不会有任何测试失败**；删掉或改名一个，多余项也不会自曝。

把集合推导出来，消除的是这一类 bug 而不是这一个实例：

```ts
const NON_VISUAL_COORDINATE_ACTIONS = new Set(['cursor_position', 'wait']);
const MODEL_SCREENSHOT_ACTIONS: ReadonlySet<string> = new Set(
  CU_ACTION_TYPES.filter((action) => !NON_VISUAL_COORDINATE_ACTIONS.has(action)),
);
```

这样新增坐标动作默认就进入视觉通道，而"刻意不给图"变成对排除集合的一行可审改动。如果你更想保留显式列表，请在 `computer-use-schema-parity.test.ts` 里补上这个集合与 `CU_ACTION_TYPES` 的断言，让漂移以测试失败的方式暴露，而不是悄悄把模型变成瞎子。

</details>

hqhq1025:
@Astro-Han Thanks for the follow-up. I rechecked this concern against the current exact head `4300288c2838eaf01ac103c069b950937f3cbeee` after rebasing onto the latest `main`.

The `MODEL_SCREENSHOT_ACTIONS: ReadonlySet<string>` described in the review no longer exists on this head. The current implementation is:

- `packages/runtime/src/computer-use-tools.ts:347-376`: `COMPUTER_USE_MODEL_SCREENSHOT_POLICY` is an object literal checked with `satisfies Record<CuToolActionType, 'always' | 'explicit' | 'never'>`. A canonical action added without a policy entry is therefore a compile-time error; an unknown/dead policy key is also rejected. `cursor_position` and `wait` are deliberately and visibly classified as `never`.
- `packages/runtime/src/__tests__/computer-use-tools.test.ts:2632-2634`: the regression compares all runtime policy keys with the complete canonical `CU_TOOL_ACTION_TYPES` list, so runtime/export drift fails the test.

I intentionally kept the exhaustive policy instead of defaulting every future coordinate action to vision through a derived exclusion set. Whether pixels may enter model context is a policy boundary, so a new action should require an explicit `always`, `explicit`, or `never` decision rather than inherit a permissive default.

Fresh verification on this head: Runtime typecheck and build pass; the Computer Use tool and schema-parity suites pass 89/89; Biome and `git diff --check` pass. No additional source change was needed for this follow-up because the current head already implements the requested typed exhaustive mapping and full canonical parity test.

Could you please re-review this head and approve if the evidence resolves the concern?

Astro-Han (approved):
The rewrite is better than what I asked for. Turning the set into `COMPUTER_USE_MODEL_SCREENSHOT_POLICY` keyed by every action, with `satisfies Record<CuToolActionType, …>`, makes the compiler reject a new action that nobody classified — that is stronger than the parity test I suggested, because it fails at build time rather than in a test run, and the added `Object.keys(...)` assertion still covers the ordering and the runtime shape. The three-way `always` / `explicit` / `never` split also reads better than the implicit "everything except two" it replaces.

Binding the decision to `includeScreenshotInModelOutput` on the executed result, instead of re-reading `input` in `toModelOutput`, closes a hole I had not raised: the input object is reachable and mutable after the call, so the old form let a later mutation change what the model was shown for an invocation that had already run. The regression test for exactly that is a good addition.

`test` is green on the current head and nothing is outstanding on my side. Approving.

<details>
<summary>简体中文</summary>

这次重写比我提的建议更好。把集合改成按每个动作列举的 `COMPUTER_USE_MODEL_SCREENSHOT_POLICY`，配合 `satisfies Record<CuToolActionType, …>`，等于让编译器直接拒绝"新增了动作却没人给它分类"的情况——这比我建议的对拍测试更强，因为它在构建期就失败，而不是等测试跑起来；同时新增的 `Object.keys(...)` 断言仍然覆盖了顺序和运行期形状。`always` / `explicit` / `never` 这个三分法，也比原来隐含的"除两个之外全都要"更好读。

把判断绑定到执行结果上的 `includeScreenshotInModelOutput`，而不是在 `toModelOutput` 里重新读 `input`，还堵上了一个我没有提出的问题：调用结束后 `input` 对象仍然可达且可变，原来的写法允许事后修改它，从而改变一次**已经执行完**的调用给模型看到的内容。针对这一点补的回归测试很到位。

当前 head 的 `test` 已通过，我这边没有遗留项。批准。

</details>

Astro-Han:
`main` is green at `f37f5c3f29` and this is clean against it. Merging.

<details>
<summary>简体中文</summary>

`main` 在 `f37f5c3f29` 上是绿的，本分支相对它也是干净的。合并。

</details>

PR #3588, opened 2026-08-23
fix(ui): localize client capability recovery

hqhq1025:
## Summary

- attach a structured `client_capability` source to bypass-boundary tool refusals while preserving the original model-facing diagnostic
- replace that diagnostic in the chat UI with localized Chinese/English guidance and a one-click "switch to Bypass and retry" action
- retry only after the permission mode is confirmed as changed, and abort the retry when confirmation is cancelled, the Host returns another mode, or the active task changes
- localize nearby hard-coded UI strings for visible system notes, file-write summaries, Rive workflow summaries, and linked-agent error labels

## Behavior boundary

Only `requires_bypass` failures explicitly tagged with `source: client_capability` receive the local-app recovery UI. Generic filesystem or shell boundary failures remain verbatim and do not imply that a local application is involved.

### Before

Raw English boundary diagnostic

### After

Localized recovery action

Both screenshots use a 1496 × 268 viewport and the same expanded tool-result state.

## Base CI unblock

The latest `main` added two CLI source files without ASF headers, so the first PR run failed before reaching tests. This branch includes the two mechanical headers required by the repository-wide license check.

## Verification

- `npm run build:test`
- `npm run typecheck`
- `npm run lint`
- `npm run check:asf-headers`
- 81 targeted Node tests across core, runtime, UI, desktop, and the affected CLI baseline
- Storybook visual verification at 1496 × 268


Astro-Han (approved):
Review at exact head `9b1f77542003d871ff580465dc32e6a849245f7f`.

**APPROVE.** No P0–P2.

The title says localization, but 27 files across core, runtime, ui, cli and desktop is more than copy extraction, so I reviewed the two halves separately — a behaviour change riding along with a copy change is the usual way one slips through unexamined.

**Behaviour half — validated, and correctly fail-closed.** `sandboxFailure` gains an optional `source: 'client_capability'`, and `tool-result-record-schema.ts:248-250` admits it *only* when `reason === 'requires_bypass'` and the value is exactly `'client_capability'`; every other combination is rejected. The field is optional, so existing records decode unchanged. On the consuming side `isRequiresBypassToolResult` requires both `reason` and `source`, so a `requires_bypass` denial from any other origin cannot raise the new banner.

The retry interaction is self-consistent where it matters: `setPermissionMode` now returns a boolean, and `handleSwitchToBypassAndRetry` regenerates only when `switched === true` *and* the session hasn't changed underneath it. An already-active mode short-circuits to `true` without a destructive confirmation, a pending conflict returns `false`, and a Host read-back that disagrees with the request returns `false`. Those three are exactly the paths where a naive implementation would regenerate against the wrong session or the wrong mode. The banner's pending state resets in `finally`, so double-click re-entry doesn't strand it.

**Localization half.** `UI_LOCALES` is `['zh', 'en']` (`core/ui-locale.ts:21`) and both are filled for every new key, with `satisfies` making a missing key a compile error — so there's no path for a bare key to reach the screen. Interpolations carry their placeholders; English plurals are the two-form case and Chinese needs none. The zh strings read as complete sentences rather than fragments, and the TUI badges map one-to-one onto running/waiting/blocked with `running` shown only when the Host reports a live Turn.

Worth calling out: deleting the `mode_change` and `turn_aborted` label entries is dead-code removal, not a behaviour change — `VISIBLE_SYSTEM_NOTES` (untouched at `materialize.ts:138-142`) only ever admitted `context_compacted`, `context_compaction_failed_open` and `step_limit`, so those two labels were never rendered. Threading `locale` into the projection stage input and the memo comparison is also right: a locale switch correctly triggers re-projection instead of leaving stale strings on screen.

**Validation limits, stated plainly.** Layout pressure was **not** verified by rendering the GUI — the review environment can't start Electron. The compensating evidence is the new Storybook stories for manual inspection, SSR assertions over the render markup, and the fact that the new banner reuses the existing sandbox-banner spacing classes rather than introducing its own. Since the locale set is only zh/en, the long-compound-word risk that usually motivates this check (German, Russian) doesn't arise, and CJK wrapping is handled by the existing containers. A maintainer with a GUI may still want to eyeball the banner at a narrow width.

Also observed, not a finding: legacy denial records written before this change have no `source`, so they keep their original error text and never show the banner. That is reasonable migration behaviour.

Tests run at this head: ui 217/217, core schema 11/11, runtime ledger 10/10, cli tui-status 3/3, desktop main 1259 passing. Exact-head `test` is `completed/success`; `package` and `Dependency audit` are absent by path filter, which is expected for a UI-only change rather than a missing gate.

PR #3589, opened 2026-08-23
fix(ci): add missing ASF headers to session status files

1625567290:
## Summary

- add the required ASF license header to the two session-status files introduced by #3441
- restore the source-header gate on current `main`
- make no runtime or test behavior changes

## Verification

- `npm run check:asf-headers`
- `git diff --check`

## Scope

Exactly two TypeScript files changed; no documentation or Java test sources are included.


Astro-Han (approved):
Reviewed with Codex assistance. The exact-head diff was checked against the reported ASF header failure, and CI / test is green. No issues found.

PR #3593, opened 2026-08-23
test(storage): stop leaking control directories into the user cache

### LINKED ISSUE #3590: test(storage): the suite leaks 159 control directories into the real user cache per run
## Problem

A clean run of the `@maka/storage` test suite leaves 159 directories behind in the developer's real cache directory. Nothing fails, so nothing surfaces it.

```
$ ls ~/Library/Caches/Maka/runtime-hosts | wc -l
27740
$ node --test --test-concurrency=2 "packages/storage/dist/__tests__/*.test.js"
ℹ tests 912
ℹ pass 898
ℹ fail 0
$ ls ~/Library/Caches/Maka/runtime-hosts | wc -l
27899
```

On this machine the directory has accumulated 27,740 entries. Disk cost is negligible (264K — most hold a single empty `owner.lock`), but the count grows without bound and nothing ever reclaims it.

## Cause

`resolveRootControlNamespace()` deliberately resolves to the real OS account home and cannot be redirected by an environment variable:

```ts
// packages/storage/src/root-authority.ts
export function resolveRootControlNamespace(): string {
  const accountHome = userInfo().homedir;
  if (process.platform === 'darwin') {
    return join(accountHome, 'Library', 'Caches', 'Maka', 'runtime-hosts');
  }
  ...
}
```

That is reasonable for a trust boundary — `owner.lock` decides who may write a State Root, so an env var must not be able to move it. The consequence is that tests write there too, and the control directory sits **outside** the temporary State Root they create.

So a helper of this shape cleans up the State Root but not its control directory:

```ts
async function withTempDir(run: (base: string) => Promise<void>): Promise<void> {
  const base = await mkdtemp(join(tmpdir(), 'maka-runtime-policy-'));
  try {
    await run(base);
  } finally {
    await rm(base, { recursive: true, force: true });   // the State Root, not the control directory
  }
}
```

Tests that do clean up carry an extra explicit line:

```ts
await rm(join(resolveRootControlNamespace(), capability.rootId), { recursive: true, force: true });
```

There is no shared helper for it in `@maka/storage`, so the line is copied per file, and seventeen files do not have it.

## Which files leak

Measured by running each test file on its own and counting the directory before and after:

| Directories leaked | File (`packages/storage/src/__tests__/`) |
|---|---|
| 54 | `runtime-policy-stores.test.ts` |
| 22 | `managed-workspace-baseline.test.ts` |
| 18 | `managed-workspace-owner.test.ts` |
| 16 | `memory-bundle-store.test.ts` |
| 15 | `usage-stores.test.ts` |
| 5 | `artifact-writer-lock.test.ts` |
| 4 | `storage-writer-composition.test.ts` |
| 3 | `artifact-stores.test.ts` |
| 3 | `daily-review-authority.test.ts` |
| 3 | `goal-authority.test.ts` |
| 3 | `shell-run-authority.test.ts` |
| 3 | `sqlite-long-term-memory-store.test.ts` |
| 3 | `sqlite-workflow-store.test.ts` |
| 3 | `task-ledger-authority.test.ts` |
| 2 | `state-root-composition.test.ts` |
| 1 | `project-catalog-authority.test.ts` |
| 1 | `sqlite-core-execution-store.test.ts` |

Total 159, matching the whole-suite delta.

`@maka/runtime-host` was checked the same way and does not leak: `execution-host-recovery.test.ts` and `storage/root-authority.test.ts` both measured a delta of 0, because their shared `ExecutionFixture.close()` already removes the control directory, the endpoint directories and the base.

## Proposed fix

Add one shared helper under `packages/storage/src/__tests__/fixtures/`, mirroring `runtime-host`'s `endpoint-hygiene.ts`, and call it from the seventeen files where a temporary State Root is torn down. This keeps the trust boundary exactly as it is and only makes the tests clean up what they created.

## Out of scope here

Reclaiming control directories that are already stale is a separate question. It needs a definition of "stale" that is safe against a live `owner.lock`, so it should not ride along with a test-hygiene change.

Joob1n:
## Summary

A clean run of the `@maka/storage` suite left 159 directories behind in `~/Library/Caches/Maka/runtime-hosts`. Nothing failed, so nothing surfaced it — the machine this was found on had accumulated 27,740.

A storage root's control directory is keyed by rootId under the real OS account home, deliberately outside the root itself. `resolveRootControlNamespace()` reads `userInfo().homedir` so that no environment variable can move a boundary `owner.lock` enforces. The consequence is that `rm(temporaryRoot)` does not reclaim it, and seventeen test files removed only their temporary root.

This adds `packages/storage/src/__tests__/fixtures/control-directory-hygiene.ts` and calls it from every file that resolves a temporary root:

- `trackControlDirectory(capability)` records the rootId at resolution and returns the capability unchanged, so it wraps a `resolveStorageRoot` call in place;
- `removeTrackedControlDirectories()` reclaims them in teardown.

The trust boundary is untouched; the tests simply clean up what they created.

**Recording at resolution rather than teardown is required, not stylistic.** The first attempt read the rootId from the root's marker file during teardown. Instrumenting it showed the marker was already gone for 14 of the 15 leaks in `managed-workspace-owner`, because those tests remove or quarantine the root as part of what they exercise. The recorded id outlives the root; the marker does not. `removeControlDirectoryForRootPath` is kept for teardown that still has the root on disk, with that precondition documented.

Fixes #3590

## Verification

Counted `~/Library/Caches/Maka/runtime-hosts` before and after each run.

Whole suite:

```
$ node --test --test-concurrency=2 "packages/storage/dist/__tests__/*.test.js"
ℹ tests 912
ℹ pass 898
ℹ fail 0
before 28126 → after 28126     (+159 before this change)
```

The five largest sources, measured individually, all pass and all now leak nothing:

| File | Before | After | Tests |
|---|---|---|---|
| `runtime-policy-stores` | +54 | 0 | 52 pass, 0 fail |
| `managed-workspace-baseline` | +22 | 0 | 22 pass, 0 fail |
| `managed-workspace-owner` | +18 | 0 | 17 pass, 0 fail |
| `memory-bundle-store` | +16 | 0 | 16 pass, 0 fail |
| `usage-stores` | +15 | 0 | 13 pass, 0 fail |

The remaining twelve files (+1 to +5 each) are covered by the same change and are included in the whole-suite result above.

`@maka/runtime-host` was measured the same way and does not leak — its shared `ExecutionFixture.close()` already removes the control directory, the endpoint directories and the base — so nothing there is changed.

Repository checks:

```
npm run format:check   Checked 1599 files. No fixes applied.
npx biome lint packages/storage/src/__tests__/   Checked 96 files. No fixes applied.
npm --workspace @maka/storage run build          clean
```

## Out of scope

Reclaiming control directories that are already stale needs a definition of "stale" that is safe against a live `owner.lock`. That is tracked in #3590 and deliberately not attempted here.

## AI use

Select exactly one:

Tool(s) and scope: Claude Opus 5 via Claude Code — measured the leak, attributed it per file by running each test file in isolation, wrote the fixture and the call-site changes, and re-measured. Every number in this description came from running the command shown. The commit carries a `Generated-by` trailer.

## Checklist

The coverage here is the measurement itself rather than a new assertion: before this change the suite leaves 159 directories behind, after it leaves none, with the same 912 tests passing. A test that asserts on the developer's real cache directory would be worse than the measurement — it would either be racy under `--test-concurrency` or would itself have to write there.

- [x] No

### DIFF
--- a/packages/storage/src/__tests__/artifact-stores.test.ts
+++ b/packages/storage/src/__tests__/artifact-stores.test.ts
@@ -21,7 +21,7 @@ import assert from 'node:assert/strict';
 import { mkdtemp, rm, stat } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
-import { describe, test } from 'node:test';
+import { after, describe, test } from 'node:test';
 import {
   authenticateInteractiveArtifactStoreWriter,
   openInteractiveArtifactStoreForWrite,
@@ -34,6 +34,14 @@ import {
   tryAcquireInteractiveRootOwner,
   type StorageRootLease,
 } from '../root-authority.js';
+import {
+  removeTrackedControlDirectories,
+  trackControlDirectory,
+} from './fixtures/control-directory-hygiene.js';
+
+// The control directory of each resolved root lives outside that root, so a
+// temporary root's removal leaves it behind; reclaim the recorded rootIds here.
+after(removeTrackedControlDirectories);
 
 describe('interactive artifact store authority', () => {
   test('requires authentic leases and writer facades', async () => {
@@ -95,7 +103,9 @@ describe('interactive artifact store authority', () => {
 
   test('root close revokes new facade operations after draining an in-flight write', async () => {
     await withTemporaryRoot('interactive', async (root, track) => {
-      const capability = await resolveStorageRoot({ path: root, kind: 'interactive' });
+      const capability = trackControlDirectory(
+        await resolveStorageRoot({ path: root, kind: 'interactive' }),
+      );
       const owner = await tryAcquireInteractiveRootOwner(capability);
       assert.ok(owner);
       const writer = track(await openInteractiveArtifactStoreForWrite(owner.lease));
@@ -167,7 +177,9 @@ async function withInteractiveOwner(
   ) => Promise<void>,
 ): Promise<void> {
   await withTemporaryRoot('interactive', async (root, track) => {
-    const capability = await resolveStorageRoot({ path: root, kind: 'interactive' });
+    const capability = trackControlDirectory(
+      await resolveStorageRoot({ path: root, kind: 'interactive' }),
+    );
     const owner = await tryAcquireInteractiveRootOwner(capability);
     assert.ok(owner);
     try {
--- a/packages/storage/src/__tests__/artifact-writer-lock.test.ts
+++ b/packages/storage/src/__tests__/artifact-writer-lock.test.ts
@@ -34,7 +34,7 @@ import {
 import { tmpdir } from 'node:os';
 import { join, sep } from 'node:path';
 import { setTimeout as delay } from 'node:timers/promises';
-import { test } from 'node:test';
+import { after, test } from 'node:test';
 import type { ArtifactRecord } from '@maka/core/artifacts';
 import {
   openInteractiveArtifactStoreForWrite as openInteractiveArtifactStoreForWriteRaw,
@@ -53,6 +53,14 @@ import {
 } from '../root-authority.js';
 import { exportSessionBundleState } from '../session-bundle-policy.js';
 import { createSessionStore } from '../session-store.js';
+import {
+  removeTrackedControlDirectories,
+  trackControlDirectory,
+} from './fixtures/control-directory-hygiene.js';
+
+// The control directory of each resolved root lives outside that root, so a
+// temporary root's removal leaves it behind; reclaim the recorded rootIds here.
+after(removeTrackedControlDirectories);
 
 const TEST_TIMEOUT_MS = 15_000;
 const OPERATION_TIMEOUT_MS = 5_000;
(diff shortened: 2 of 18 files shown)


jackwener (approved):
I reviewed exact head `011b5d34adfa7e58b13e65904959e091202efca3` and approve it with no P0–P3 findings.

The problem and mechanism are sound: test storage roots leave their control directories in the user cache namespace after the temporary root is removed. This change records the stable 64-hex `rootId` when resolving the root, then reuses the production `resolveRootControlNamespace()` authority during teardown to remove exactly that directory. It does not change production behavior or trust boundaries, duplicate the path formula, or delete outside the namespace.

The 17 affected fixtures now share one control-directory cleanup contract. I ruled out deletion of real workspaces, leaks after resolution failure, cross-test global-count races, and an omission in `root-authority.test.ts` along the production path. Historical stale directories remain correctly scoped to #3590 because safely cleaning them requires a separate contract around live `owner.lock` files.

The exact-head hosted `test` check completed successfully; the PR was open, non-draft, and mergeable with no live hold at publication time.

<details>
<summary>简体中文</summary>

我复核了 exact head `011b5d34adfa7e58b13e65904959e091202efca3`，结论为 APPROVE，没有 P0–P3 finding。

问题与机制成立：测试 storage roots 的 control directories 位于用户 cache namespace，不随临时 root 一起删除；本 PR 在 root resolution 时记录稳定的 64-hex `rootId`，teardown 复用生产 `resolveRootControlNamespace()` 的唯一 namespace authority 删除对应目录。生产代码和信任边界未改，没有复制路径公式，也不会越出 namespace 误删其它缓存。

17 个 fixture 现在共享同一条 control-directory cleanup contract。真实 workspace 误删、resolve 失败泄漏、跨测试全局计数竞态、`root-authority.test.ts` 遗漏等候选均已沿生产路径排除。历史 stale directories 留给 #3590，因为那需要单独定义对 live `owner.lock` 安全的清理契约。

发布时 exact-head hosted `test` 已成功完成，PR 为 open、非 draft、可合并，且没有 live hold。

</details>

PR #3596, opened 2026-08-23
docs: use the new Ä mark in the README header

jackwener:
## Summary
Both READMEs still showed `apps/desktop/assets/icon.png` (the character mascot) next to the title. The icon set that landed in #3431 is the geometric Ä mark under `apps/desktop/assets/app-icons/`. Point the header at `sky.png`.

Hero banners are unchanged. Packaging/window default is still `assets/icon.png`; this PR is docs-only.

## Test plan
- [ ] GitHub renders `app-icons/sky.png` in README.md and README.zh-CN.md
- [ ] Character mascot no longer appears in the README title row

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@
 -->
 
 <h1 align="center">
-  <img src="apps/desktop/assets/icon.png" alt="Maka" width="72" valign="middle" /> Apache Maka (Incubating)
+  <img src="apps/desktop/assets/app-icons/sky.png" alt="Maka" width="72" valign="middle" /> Apache Maka (Incubating)
 </h1>
 
 <p align="center"><sub>Incubating at The Apache Software Foundation</sub></p>
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -18,7 +18,7 @@
 -->
 
 <h1 align="center">
-  <img src="apps/desktop/assets/icon.png" alt="Maka" width="72" valign="middle" /> Apache Maka (Incubating)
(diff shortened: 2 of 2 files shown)


Astro-Han (approved):
Reviewed at exact head `da392974df59110fb16516fc6e06f8ced760665e`. `test` is terminal green on this head.

Verified the substance rather than just the diff shape:

- `apps/desktop/assets/app-icons/sky.png` exists at this head (blob `7e3ee3dd`), so neither README will render a broken image.
- Both paths are repo-root relative and match the surrounding markup, unchanged apart from the filename.
- Packaging is untouched: `electron-builder.config.mjs`, `app-icon.ts`, `settings.ts`, and `verify-packaged-app.test.mjs` all still reference `assets/icon.png`, so the shipped app icon is unaffected. The description's claim that this is documentation-only holds.

**Approving.** No P0–P2.

### [P3] Provenance record now points at the icon that is no longer shown

`docs/code-origin-audit.md:182` lists `apps/desktop/assets/icon.png` among the assets with recorded origin (ChatGPT Image, prompts not retained). Nothing under `apps/desktop/assets/app-icons/` appears in that document.

So after this change the README's headline image is an asset whose origin is not recorded in the audit file, while the recorded one is no longer displayed there. Nothing here is wrong today — the audit file is about repository contents, not about which file the README happens to reference — but since that document exists precisely to answer "where did this artwork come from", the icon that represents the project on its front page is the one most worth having listed. Worth adding a line for the `app-icons/` set when someone next touches that file; no change requested here.

PR #3597, opened 2026-08-23
fix(win): invalidate stale rollback backup authority

jackwener:
## Summary

- revoke the Windows pre-upgrade backup completeness marker before best-effort successful-install cleanup
- prevent a residual old-version marker from blocking later upgrades when `RMDir /r` is interrupted by a held file
- pin the authority-before-cleanup ordering with a product release policy test

## Why

The successful upgrade tail intentionally treats backup removal as best effort. On Windows, antivirus or indexing can keep an application file open, so `RMDir /r` may partially delete the backup and leave its completeness marker behind. After the successful install, the registry identifies the new version, while that marker still identifies the old backup version. The next installer sees the stale authority, detects the version mismatch, and refuses an otherwise healthy upgrade with exit 101.

Deleting the marker first makes any cleanup residue inert before recursive tree removal begins. The completed install remains successful even when cleanup cannot finish.

## Verification

- `npm run build`
- `npm run test:product-release` (32/32)
- `npx biome check scripts/product-release.test.mjs`
- `git diff --check`
- mutation: deleting the production marker invalidation makes the new targeted test fail

### DIFF
--- a/apps/desktop/build/installer.nsh
+++ b/apps/desktop/build/installer.nsh
@@ -616,8 +616,8 @@ FunctionEnd
   # Success tail: the upgrade completed, so the backup and any aside residue
   # are redundant. Invalidate the completeness marker before the best-effort
   # tree removal: if a scanner holds an application file, RMDir may leave the
-  # old backup behind, and a later upgrade must not mistake that residue for a
-  # live recovery source from the newly installed version.
+  # old backup behind. Its old-version marker would make a later upgrade
+  # reject the healthy current installation as mismatched recovery state.
   ClearErrors
   Delete "$makaBackupDir\${MAKA_BACKUP_MARKER}"
   Delete "$makaBackupDir\${MAKA_RECOVERY_README}"
--- a/apps/desktop/build/installer.nsh
+++ b/apps/desktop/build/installer.nsh
@@ -614,8 +614,13 @@ FunctionEnd
 
 !macro customInstall
   # Success tail: the upgrade completed, so the backup and any aside residue
-  # are redundant. Best effort -- a held handle must not fail a completed
-  # install.
+  # are redundant. Invalidate the completeness marker before the best-effort
+  # tree removal: if a scanner holds an application file, RMDir may leave the
+  # old backup behind, and a later upgrade must not mistake that residue for a
+  # live recovery source from the newly installed version.
+  ClearErrors
+  Delete "$makaBackupDir\${MAKA_BACKUP_MARKER}"
+  Delete "$makaBackupDir\${MAKA_RECOVERY_README}"
   ClearErrors
   ${If} ${FileExists} "$makaBackupDir\*.*"
     RMDir /r "$makaBackupDir"
--- a/scripts/product-release.test.mjs
+++ b/scripts/product-release.test.mjs
@@ -194,6 +194,24 @@ test('Desktop packaging does not distribute the retired bundled Git runtime', ()
   );
 });
 
+test('a successful Windows upgrade invalidates backup authority before best-effort cleanup', async () => {
+  const source = await readFile(
+    join(repoRoot, 'apps', 'desktop', 'build', 'installer.nsh'),
+    'utf8',
+  );
+  const macroStart = source.indexOf('!macro customInstall');
+  const macroEnd = source.indexOf('!macroend', macroStart);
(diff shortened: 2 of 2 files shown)


Astro-Han (commented):
Reviewed at exact head `d9f8593c9a3c9f837a8cd1d3f56a2becdccfce4c`. Exact-head `test` and `package` are both `completed/success`. No `[P0]`–`[P2]`; one `[P3]` recorded below.

### The strict-equality check picks the right failure direction

`makaBackupMatchesSnapshot` (`apps/desktop/build/installer.nsh:176-191`) requires the marker's `version=X` to match `$makaPrevDisplayVersion` exactly, and under the snapshot-adoption path (`:466-486`) that variable *is* the source version recorded in the persisted snapshot. So the test isn't "some backup exists" — it's "this marker and the source this run would restore are the same identity."

The realistic way to fail that test is that `$INSTDIR` has already been replaced by a later successful upgrade, which makes the old backup an invalid restore source for it. Refusing is correct there: adopting it would write back a registry snapshot that doesn't describe what is actually installed.

The cost is bounded, and this is the part worth stating plainly, since "refuse to upgrade" can sound like the more dangerous option: the `Quit` at `:499-507` happens in `customInit`, **before** the app kill and the old uninstaller, both of which live in the install section. Nothing has been written at that point, so the existing installation is untouched and the user gets a specific instruction — inspect `$makaBackupDir`, move or remove it, run the installer again. One rejected upgrade plus one manual step, in exchange for never adopting a restore source that cannot correctly write back its snapshot. Loosening this to "adopt whatever is on disk" is what would produce a data-level problem.

Nothing half-installed is left behind on the rejection paths: backup creation, verification, and marker writing all happen in the later `else` branch, and the `:485-495` case (no registration, no adoptable snapshot) is the same shape — the comment there says as much.

On success, `customInstall` (`:616-631`) deletes the marker before `RMDir`, so a scanner holding a file open cannot leave a surviving directory that still carries a stale marker. That ordering is exactly what the new test pins.

### `[P3]` The behaviour itself has no installer-level coverage

`scripts/product-release.test.mjs:197-214` asserts the *ordering* invariant — marker deletion precedes backup deletion and snapshot-key removal — and that suite runs 26/26 green. But the end-to-end behaviour (upgrade fails → rollback succeeds → marker ends in the right state) has no NSIS-level integration test. Recording rather than requesting; adding one is a larger piece of work than this change.

### Validation boundaries

Everything above is static reading and control-flow analysis of `installer.nsh` — NSIS cannot be executed on the reviewing host. Ran: `product-release.test.mjs` in full (26/26; an earlier failure was a missing `@maka/eval` dist locally, not a code issue) plus line-by-line verification of `:176`, `:466-507`, `:551`, and `:614-635`.

Not run: any real Windows install, failure injection, or failpoint path. Two cases would benefit from a real machine before this is considered field-verified — snapshot adoption on a second upgrade after the installer is `SIGKILL`ed mid-run, and the residue shape when a scanner holds the backup file open through a successful upgrade.

Astro-Han (approved):
Approving at exact head `d9f8593c9a3c9f837a8cd1d3f56a2becdccfce4c`.

A second reviewer went through this head independently of the earlier comment and reached the same position: the strict marker/`$makaPrevDisplayVersion` equality does not condemn a recoverable backup, the rejection point performs no destructive write to the installation, and the success-path deletion order is correct. Where the two readings differ slightly is worth recording rather than smoothing over — the second pass notes that writes into `$makaBackupDir` itself may already have occurred before a rejection, which is true and harmless: the backup directory is not the installation, and the user is told to inspect and remove it.

One case that came out of the independent pass and strengthens the ordering argument: if a scanner holds a file open through `RMDir /r`, what survives is a directory with **no** marker, so the next upgrade treats it as "no backup" rather than tripping the mismatch refusal. The failure mode of the cleanup is therefore permissive in the safe direction, while the adoption check is strict in the safe direction.

Exact-head `test` and `package` are green. No `[P0]`–`[P2]`. The `[P3]` from the previous comment stands as recorded — the behaviour still has no NSIS-level integration coverage, only the ordering invariant in `scripts/product-release.test.mjs:197-214` — and the two real-machine cases noted there (`SIGKILL` mid-install then a second upgrade; scanner holding the backup open) remain unverified on hardware by either reviewer.

PR #3606, opened 2026-08-23
fix(ui): shorten default thinking label

Astro-Han:
## Summary

Change the Chinese composer label for the model default thinking level from `模型默认` to the shorter canonical label `默认`.

Add a localized-copy regression test that fails with the previous label.

## Before and after

image

## Verification

- `node --test packages/ui/dist/__tests__/conversation-copy.test.js`
- `npm --workspace @maka/ui run typecheck`
- `npm run lint`
- `npm run format:check`

## AI use

Tool(s) and scope: Maka updated the localized copy, added the regression test, and prepared this pull request.

## Checklist

- [ ] No

### DIFF
--- a/packages/ui/src/__tests__/conversation-copy.test.ts
+++ b/packages/ui/src/__tests__/conversation-copy.test.ts
@@ -1,0 +1,26 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import assert from 'node:assert/strict';
+import test from 'node:test';
+import { getConversationCopy } from '../conversation-copy.js';
+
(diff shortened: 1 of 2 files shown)


jackwener (approved):
代表 @WAWQAQ，授权范围：Maka PR review / GitHub review action；exact head `b38c4bb5722237a49054f6b0ea9d74ee66c4ed1c`。\n\n无 P0–P2。中文 `默认` 与紧邻 canonical ladder 及紧凑 popout contract 一致；英文 `Model default` 是自然语言消歧，不要求逐词对称。新增测试直接钉本次产品 copy contract，回退到 `模型默认` 会失败，属于有效回归测试。\n\nFresh gate：OPEN、non-draft、MERGEABLE；exact-head `test` completed/success，无 live hold。未执行 merge；不声称替代尚未统一的 human-review 文字要求。

jackwener (commented):
Correction to my earlier review on this exact head: **NO-GO, 1×P2.** Review `5002403611` must not be used as approval/merge evidence.\n\n[P2] `packages/ui/src/conversation-copy.ts:438` — shortening the per-chat choice from `模型默认` to `默认` removes a required distinction from the Settings global default.\n\nProduction path: `apps/desktop/src/renderer/use-shell-chat-model.ts:111-125` has three states: untouched `undefined` applies Settings → 默认思考级别, while explicit `null` chooses the model provider default for this chat and must override the configured global default. Commit `ede1f12f3` introduced that three-state contract and intentionally renamed the picker label from `默认` to `模型默认` after a user reported the old wording as deceptive. Restoring the ambiguous label makes users unable to tell whether they are applying the app default or explicitly cancelling it, so the next chat can run at an unintended thinking level.\n\nWhy the new test misses it: `conversation-copy.test.ts:23-25` asserts only the proposed literal. It turns red on a text mutation but does not protect the semantic distinction between Settings default and per-chat model default.\n\nMinimal fix: keep `模型默认` (and remove/reverse this literal test). If a copy regression test is desired, assert the two user-visible defaults remain distinguishable at their shared contract seam.\n\nI attempted both REST and GraphQL dismissal of approval `5002403611`; GitHub returned success responses but the review still reads `APPROVED`. The originating orchestrator has been asked to dismiss it from a maintainer account. No merge occurred.

PR #3612, opened 2026-08-23
fix(desktop): reflow narrow trace turn headers

ARE404:
## Summary

- reflow Trace turn headers based on the inspector container width
- keep failure phrases intact instead of breaking one character per line
- add a 320px Storybook regression story with geometry and overflow assertions

## Verification

- `npx biome check apps/desktop/src/renderer/styles/workbar/inspector.css apps/desktop/stories/session-workbar.stories.tsx`
- rendered Storybook at 320px, 480px, and 600px in English and Chinese
- confirmed zero horizontal overflow; 320/480px use two rows, while 600px remains one row

image

### DIFF
--- a/apps/desktop/src/renderer/styles/workbar/inspector.css
+++ b/apps/desktop/src/renderer/styles/workbar/inspector.css
@@ -28,6 +28,7 @@
   height: 100%;
   min-height: 0;
   overflow-y: auto;
+  container: maka-inspector / inline-size;
 }
 
 /* The live region is a wrapper only; its children carry their own spacing. */
@@ -125,7 +126,7 @@
   font: var(--maka-text-label);
   color: var(--destructive-text);
   min-width: 0;
-  overflow-wrap: anywhere;
+  white-space: nowrap;
 }
 
 .maka-inspector-turn-meta {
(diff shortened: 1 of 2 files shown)


Astro-Han (approved):
Reviewed at exact head `06b901d246211974df8fa9800386f09f0fd5ee90`. Nobody had reviewed this PR before, so this is a full review rather than an incremental one.

**Coverage:** the container-query reflow and its threshold, the `overflow-wrap: anywhere` → `white-space: nowrap` swap on the failure phrase, and whether the new story actually constrains what it claims. **Not covered:** visual/aesthetic judgement, and locales beyond the two in `conversation-copy.ts`.

**No P0–P3.**

The one thing worth checking here was the unconditional `white-space: nowrap` on `.maka-inspector-turn-failure`. Replacing `overflow-wrap: anywhere` with `nowrap` removes a containment fallback, and the container query only restores a safe layout at or below `32rem` — so if the failure text were unbounded, widths above the threshold would overflow horizontally where they previously wrapped.

It isn't unbounded. `turnFailure` resolves against the closed `EN_TURN_FAILURE` / `ZH_TURN_FAILURE` maps with a short default for unknown codes, so the longest phrase either locale can produce is `Model call failed` / `模型调用失败`. With `SESSION_WORKBAR_MIN_WIDTH = 320` and `SESSION_WORKBAR_MAX_WIDTH = 600`, the band left to the one-line flex layout is roughly 512–600px of workbar width, and label + failure + meta fit that comfortably in both locales. The swap is safe because the content is enumerated, not because `nowrap` is generally safe here — worth remembering if a raw provider message is ever routed into this span.

Choosing `container: maka-inspector / inline-size` on the scroll area rather than a viewport media query is the right lever, and the comment gives the reason: the workbar resizes independently of the window and the same panel also has a bottom placement, so window width is not the quantity that decides whether the head line fits.

The new story is a real constraint rather than a mirror of the implementation. It asserts geometry — failure below the label, failure and meta sharing a row within 1px, and `turn.scrollWidth <= turn.clientWidth` — so deleting the container query would fail it on the containment assertion, not on a selector. It also drives the actual 320px floor via `width={320}` instead of a synthetic viewport.

**Gate state:** exact-head `test` is `completed/success`, and I confirmed the Storybook steps ran rather than being skipped in that job — `Build Storybook` and `Storybook smoke` are both `completed/success` at this head, so the new play assertion is hosted-verified and not only locally verified. `mergeable` is `true`; `blocked` is the missing review, which this resolves.

Approving.

PR #3613, opened 2026-08-23
fix: expose active model in stable prompts

Astro-Han:
## Summary

- expose the active execution model once in the stable interactive system prompt for main, child, and hosted headless/Eval runs
- keep the per-turn session environment focused on volatile context and rely on the existing backend rebuild to refresh identity after a model switch
- collapse the model-switch transcript projection to the latest assistant model and remove unused session-change payload fields

The model identity is intentionally not persisted as a switch event: the Session header remains the current configuration authority, assistant messages record the model actually used by each completed step, and changing the model already disposes the backend before the next prompt is composed.

## Verification

- `npm run build`
- `node --test packages/core/dist/__tests__/model-switch-transcript.test.js`
- `node --test --test-name-pattern='child prompts identify the active model|backend composition survives a moved saved Git Bash executable while Bash fails closed' packages/runtime-host/dist/__tests__/execution-model-composition.test.js`
- `npx tsc -p apps/desktop/tsconfig.main.json --noEmit`
- `npx tsc -p apps/desktop/tsconfig.preload.json --noEmit`
- `npx tsc -p apps/desktop/tsconfig.renderer.json --noEmit`
- Biome format check on affected source and test files
- `git diff --check`

## AI use

Tool(s) and scope: Maka diagnosed the prompt/event-log composition, authored the implementation and tests, and ran the focused validation. The human contributor will review the final diff and owns the merge decision.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/src/main/runtime-host-boot.ts
+++ b/apps/desktop/src/main/runtime-host-boot.ts
@@ -941,7 +941,7 @@ function registerHostClientIpc(
   const emitTargetSessionsChanged = (
     reason: SessionChangedReason,
     sessionId?: string,
-    extra?: Pick<SessionChangedEvent, "connectionSlug" | "modelId" | "turnId">,
+    extra?: Pick<SessionChangedEvent, "modelId" | "turnId">,
   ): void => {
     if (isTargetActive()) emitSessionsChanged(scope, reason, sessionId, extra);
   };
@@ -1461,14 +1461,12 @@ function emitSessionsChanged(
   scope: DesktopTargetScope,
   reason: SessionChangedReason,
   sessionId?: string,
-  extra?: Pick<SessionChangedEvent, "connectionSlug" | "modelId" | "turnId">,
+  extra?: Pick<SessionChangedEvent, "modelId" | "turnId">,
 ): void {
   const event: SessionChangedEvent = {
-    type: "sessions_changed",
     reason,
     ts: Date.now(),
     ...(sessionId ? { sessionId } : {}),
-    ...(extra?.connectionSlug ? { connectionSlug: extra.connectionSlug } : {}),
     ...(extra?.modelId ? { modelId: extra.modelId } : {}),
     ...(extra?.turnId ? { turnId: extra.turnId } : {}),
   };
--- a/apps/desktop/src/main/runtime-host-desktop-candidate.ts
+++ b/apps/desktop/src/main/runtime-host-desktop-candidate.ts
@@ -103,7 +103,7 @@ export interface DesktopRuntimeHostCandidateDeps {
     scope: DesktopTargetScope,
     reason: SessionChangedReason,
     sessionId?: string,
-    extra?: Pick<SessionChangedEvent, "connectionSlug" | "modelId" | "turnId">,
+    extra?: Pick<SessionChangedEvent, "modelId" | "turnId">,
   ) => void;
   readonly completeComputerUseTurn: (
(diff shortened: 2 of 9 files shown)


Astro-Han (commented):
## Review: NO-GO (1×P1)

The ordinary main/child composer path is sound: `modelId` is captured when the backend is built, a configuration transition disposes that backend before committing the new model, and the new line is therefore stable for the lifetime of each composer. Switching models does not create an additional useful-prefix-cache loss because the request is moving to a different model authority anyway. The transcript projection and session-change payload deletions are also behavior-preserving: `hasConversation` and ordinary-update `connectionSlug`/`modelId` had no consumers, while rebound events retain `modelId`.

### P1 — the first-class headless/Eval prompt still omits the active model

`createInteractiveRunComposer` returns the tool-profile prompt at `packages/runtime-host/src/server/interactive-run-composer.ts:219-228`, before `buildActiveModelPromptFragment(input.model)` is reached at `:255`. Production still passes both `toolProfile: 'headless-coding-v1'` and the selected `modelId` through the same factory (`:474-511`), so every Maka Eval/hosted headless run receives the old four-line prompt with no model identity. This is a supported production surface, not a test-only branch: `packages/eval/src/maka-subject.ts:56` publishes the configured tool profile and Maka Eval fixtures use `headless-coding-v1`.

I reproduced the exact branch on this head by constructing the composer with both `toolProfile: 'headless-coding-v1'` and `model: 'fixture-model'`; `resolveSystemPrompt` returned `containsActiveModel: false`. The existing freeze test at `execution-model-composition.test.ts:2980-3025` also asserts the old exact prompt and never supplies a model, so CI cannot catch the omission. As written, the PR's unqualified contract—expose the active execution model once in the stable interactive prompt—does not hold for the product's hosted execution path.

Please compose the normalized model identity into the profiled stable prompt too, and add a regression that supplies `model` while using `headless-coding-v1`. Keep the profile's fixed tool ceiling and avoid reading Skills/Memory; only its stable prompt text needs the identity.

Validation: exact-head CI run `32642142585` and Windows package run `32642142543` are terminal success on `83e604a3ee188c94f2af99ab704313e842ddab29`. Clean dependency-order Core → Runtime → Storage → Runtime Host builds passed locally; focused transcript tests passed 2/2 and prompt/profile tests passed 3/3. The initial Runtime Host-only build failure was stale local Storage declarations and disappeared after building Storage; it is not a PR finding.

jackwener (approved):
GO on exact head e7eab011b24907155f746bafe471e5987dca4709.

The prior headless/Eval blocker is closed by the single outer injection point in `createHostAiSdkBackend`: every resolved system prompt now receives the normalized, already-authorized execution model. `resolveExecutionTarget()` rejects empty and unauthorized models before this path, and session model changes dispose the existing backend before committing the new header, so the identity is stable within one backend/prefix-cache lifetime.

The production hosted-execution wire test covers `toolProfile: headless-coding-v1` plus an explicit model target, checks the active-model bytes and updated prompt hash, and passed locally. Removing the injection made that test fail at the active-model assertion. The transcript projection and session-change payload reductions are behavior-preserving and remove unused projections. No P0-P3 findings remain; exact-head hosted `test` and `package` are completed/success.

Execution provenance: acting under @WAWQAQ standing Maka review/action authorization; scope: technical review and APPROVE only; exact head: e7eab011b24907155f746bafe471e5987dca4709. No merge performed.

PR #3628, opened 2026-08-23
docs(ui): align thinking-level default comments

### LINKED ISSUE #3614: Stale comment: thinking-level default states still describe the pre-#3606 label
`#3606` shortened the Chinese label for the per-chat thinking-level default from `模型默认` to `默认`, and it has been merged — the shorter label is the intended product behaviour.

What is now out of date is the comment that describes the underlying contract. `apps/desktop/src/renderer/use-shell-chat-model.ts:111-114` still reads:

> Three states, because two cannot say this: `undefined` is an untouched picker, so Settings → 通用 → 默认思考级别 applies; `null` is the user explicitly choosing 模型默认 for this one chat, which must beat the configured default or the per-chat picker could not undo it.

So the comment names the option `模型默认` while the UI now shows `默认`. That mismatch is worth closing for a concrete reason rather than tidiness: the wording in that comment is what a reviewer reaches for when judging this label, and it will keep producing the same objection — that a bare `默认` cannot distinguish "restore the app default" from "override the app default and follow the model's own." Leaving the comment as-is means re-litigating the decision every time someone reads it.

### Suggested scope

Small and self-contained — a good first contribution.

1. Update the comment at `use-shell-chat-model.ts:111-114` so it describes the states in terms of the labels that actually ship. The three-state behaviour itself is unchanged and should stay: `undefined` = untouched picker, so the Settings default applies; `null` = an explicit per-chat choice that must override the configured default, otherwise the picker cannot undo it.
2. While there, check the canonical-ladder comment next to `defaultLevel` in `packages/ui/src/conversation-copy.ts`. It predates the Settings-wide default that `#2430` introduced and is the other stale description of this same control.

No behaviour change and no test change is expected. If a reader concludes the *contract* should change rather than the comment, that is a separate discussion and worth raising here first.

### Background

`#2430` (`ede1f12f3`) introduced the global default thinking level in Settings and, in the same commit, changed this label from `默认` to `模型默认` to disambiguate the two. `#3606` reverses that naming decision deliberately; this issue only asks that the surrounding prose stop describing the old one.

k4its1t:
## Summary

Clarify the three-state thinking-level contract using the Chinese label that now ships: an untouched picker inherits the Settings-wide default, while explicitly choosing the per-chat `默认` option means using the model default and must override that setting.

Update the nearby canonical-ladder comment with the same distinction. This changes comments only; runtime behavior and localized copy are unchanged.

Fixes #3614

## Verification

- `npm run format:check` — passed (1,595 files)
- `npm run lint` — passed (2,619 files)
- `npm run build` — all workspaces passed
- `npm run typecheck` — all workspaces passed
- `npm --workspace @maka/ui test` — 220 passed

No test was added because the issue explicitly scopes this to correcting non-executable comments with no behavior or copy change. The existing localized-copy regression test remains green.

## AI use

Select exactly one:

Tool(s) and scope: OpenAI Codex assisted with issue triage, drafted the two comment updates, and ran the local verification commands. The commit includes the required `Generated-by: OpenAI Codex` trailer.

## Checklist

- [ ] Tests cover the change and fail without it — N/A for a comment-only correction; see Verification

- [x] No

### DIFF
--- a/apps/desktop/src/renderer/use-shell-chat-model.ts
+++ b/apps/desktop/src/renderer/use-shell-chat-model.ts
@@ -110,8 +110,9 @@ export function useShellChatModel(options: {
   // Renderer-only — it never mutates the persisted Settings · 模型 default.
   // Three states, because two cannot say this: `undefined` is an untouched
   // picker, so Settings → 通用 → 默认思考级别 applies; `null` is the user
-  // explicitly choosing 模型默认 for this one chat, which must beat the
-  // configured default or the per-chat picker could not undo it.
+  // explicitly choosing the per-chat `默认` option (use the model default),
+  // which must beat the configured Settings default or the picker could not
+  // undo it.
   //
   // The pick carries its target key so a Host or Project switch cannot apply it
   // to a different execution authority, even for an identically named model.
--- a/packages/ui/src/conversation-copy.ts
+++ b/packages/ui/src/conversation-copy.ts
@@ -437,7 +437,8 @@ const CONVERSATION_COPY = {
     model: {
       thinkingLevel: '思考级别', thinkingUnsupported: '当前模型不支持思考级别切换', changeThinkingLevel: '切换当前模型的思考级别', defaultLevel: '默认',
       // Short single-token labels — trigger + popout size to content.
-      // Canonical ladder: 默认 / 关 / 低 / 中 / 高 / 超高 (minimal/max when offered).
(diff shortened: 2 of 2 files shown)


Astro-Han (approved):
Reviewed at exact head `643fb4973cf309ba02b0e8f8dcb6d65381c75ff5`. The comments now match the implemented three-state contract: an untouched new-chat picker inherits the configured Settings default, while an explicit per-chat Default clears that override and uses the selected model default. No P0–P3 findings. The exact-head `test` check is terminal green.

PR #3250, opened 2026-08-19
feat(runtime): add plugin composition foundation

xxhZs:
## Positioning

This PR lands the Runtime foundation extracted from `feat/extension-hook-contributions`:

- the `Context` / `Fiber` / `Service` kernel and Effect ownership;
- plugin runtime contracts for packages, composition entries, snapshots, transactions, and inspection;
- `MakaCompositionLoader` for in-memory package registration, Entry Tree mutation, Fiber projection, switching, and retirement;
- the three feat-defined `@maka/runtime` subpath exports.

This is intentionally a foundation PR, not an independently usable product or Runtime Host integration. It does **not** add a Host Controller, durable Store, persistence, trusted package loader, product installation surface, Tool/UI wiring, or default consumer.

A later Host/product slice should migrate the feat-defined `HostExtensionController`, `HostExtensionRuntime`, and `HostPluginCompositionStore` instead of introducing a parallel adapter or persistence model.

## Architecture

The foundation is split into three implemented layers, with two intentionally absent layers above it:

```text
Product integration / Tool / UI / Hook wiring       not in this PR
Host authority / persistence / package loading      not in this PR
────────────────────────────────────────────────────────────────
Composition orchestration      MakaCompositionLoader
Plugin contracts               Package / Entry / Snapshot / Transaction
Runtime kernel                 Context / Service / Fiber / Effect / Event
```

### 1. Entry Tree: desired in-memory composition

The Entry Tree records how plugin instances should be composed: root, parent/child placement, package identity, configuration, enablement, dependency injection, isolation, and intercept declarations.

Entry parent/child relationships express composition containment and inherited Context algebra. They also define subtree mutation, disabled propagation, ordering, and cleanup boundaries. They do **not** mean that a child Entry depends on its parent package.

The Loader owns the Entry index and a monotonic composition generation. `apply()` serializes mutations and optionally checks `baseGeneration`; live `replaceSnapshot()` and `replaceSubtree()` also publish a fresh generation. A failed batch rebuilds the previous snapshot and restores its generation exactly.

### 2. Fiber Tree: process-local execution projection

A package Entry that can run on the Host is projected into a Fiber. A Fiber owns one activation, its execution Context, dependency state, Effects, child Fibers created by the plugin, Services, diagnostics, and cleanup.

The Entry Tree and Fiber Tree are related but their topology is not a one-to-one mirror:

- Entry children inherit the parent Entry's pre-plugin composition Context (`inject` / `isolate` / `intercept`);
[...]

PR #3633, opened 2026-08-23
fix(runtime-host): unify durable message lifecycle

Astro-Han:
## Summary

Queued messages were split across transcript rows, an in-memory Host queue, Client bindings, and recovery-time inference. A Host restart could therefore lose an accepted message, replay a discarded one, or change the capabilities under which it ran.

This change gives the lifecycle four explicit authorities:

- durable message admissions own accepted identity, current placement, and queue order
- the Runtime Host in-memory queue owns only live delivery and leases
- durable provider/root/Stop/retraction facts own settlement
- the transcript owns presentation only

Admission and transcript materialization now share one SQLite transaction. Promotion, reorder, and edit are durable-first. Recovery reads unresolved admissions and starts them through the normal successor-root path; it no longer scans the transcript or waits for the original Client. Retracted identities retain only their original placement and canonical content digest, so they cannot be resurrected while message contents are removed.

A queued successor executes under the durable Session contract. It does not inherit session-, turn-, call-, or initiating-Client bindings, and an explicit durable capability marker prevents online call-affine providers from being discovered later. Editing uses the connection performing that edit only for transient content preparation; the queue does not retain that Client identity.

Stop is settled only by the immutable terminal Stop fact. If the Host dies before that fact, Stop was not confirmed and accepted messages recover; after the fact, they do not.

The parallel embedded RuntimeKernel queue and its dead admission path/tests are removed. The CLI and Desktop continue to render from canonical transcript/Host projections rather than owning fallback delivery state.

## Architecture

```mermaid
flowchart TD
  C[Client message intent] --> U[Immediate transient row<br/>keyed by messageId]
  C --> A

  subgraph H[Runtime Host - sole execution authority]
    A[1. Durable admission<br/>identity, content proof, placement, order]
    Q[2. Live delivery<br/>steering queue, follow-up queue, in-flight leases]
    D{Delivery cut}
    R[Root admission and Run]
    S[3. Durable settlement<br/>provider proof, Stop, or retraction]

    A --> Q --> D
    D -->|consumed in current Turn| S
    D -->|fold into successor| R --> S
    A -->|Host restart: unresolved only| R
  end

  A -. materialize .-> T[4. Transcript<br/>presentation only]
  R --> T
  S --> T
  T -->|canonical reconciliation| U
```

The diagram intentionally shows responsibilities rather than every internal phase. The invariants are:

1. `messageId` is the end-to-end identity; UI immediacy never creates a second lifecycle owner.
2. Durable admission happens before a message becomes deliverable.
3. Only durable execution, Stop, or retraction facts settle an admission.
4. Restart recovers unresolved admissions; it never infers delivery from transcript appearance.

## Schema migration

- core execution schema v6 adds durable message admissions, queue order, and minimal retraction identity proofs
- the unmerged v5 pending-steering shape is discarded because its old two-write protocol cannot prove that a row was accepted
- Session retirement removes both pending admissions and settlements

## Verification

After rebasing onto current `main`:

- Storage and Runtime Host builds pass
- Storage core execution: 19 passed
- Message coordinator: 57 passed
- edited steering fold/restart integration: 1 passed
- focused Biome check passes
- `git diff --check`

Repository-wide tests and the full Desktop Playwright suite were not run locally; CI owns full coverage.

Screenshots are not included because this revision changes lifecycle and projection ownership without changing visual styling or layout. The observable transcript/queue behavior is covered by focused CLI, Desktop contract, and Host integration tests.

## AI use

Tool(s) and scope: Maka and Codex contributed implementation, tests, architecture analysis, and delegated read-only reviews. Material commits include `Generated-by` trailers.

## Checklist

- [x] Format, typecheck, affected builds, and focused suites pass locally

- [ ] No


Astro-Han (commented):
Thanks — the simplification at the core of this is sound, and most of the deleted CLI fallback tests genuinely remove an unreachable client-owned authority. Production has a single `SessionManager` construction that always supplies Host authority, and these packages are private, so that removal holds. Two items below.

### [P1] Steering admission has no atomic recoverable proof across the durable-append boundary

Admission durably appends the `UserMessage` before it mutates the in-memory queue and before the submit receipt exists. A Host crash inside that window leaves neither a queue entry nor any recoverable admission proof, while the durable append has already landed.

Retry then consults only root-source receipts and provider-consumption events, and SQLite here permits duplicate message IDs. So the same cut produces two different bad outcomes depending on timing: a message the user saw accepted can be lost outright, or retry can append the same message identity a second time before any single delivery has occurred.

The fix needs one atomic, recoverable admission fact spanning the append and the queue/receipt update, rather than an ordering that is only correct when no fault occurs. Please add a restart/fault regression that cuts exactly at the append→queue/receipt boundary; an ordering assertion in the happy path cannot show this.

### [P2] The CLI never renders the durable append, so a successful steering send becomes invisible

The CLI deletes the pending Steering row but does not render the durable append live. `subscription.transcript_advanced` has no consumer anywhere under `packages/cli` — I checked, there are zero references — and transcript reload only happens on a tool result, a reconnect, or a terminal state.

During a long-running step the consequence is that the user presses Enter, the text leaves the editor, and the message then appears nowhere at all until one of those events happens. From the user's side a successful send is indistinguishable from a dropped one.

The rewritten test asserts the pending row's absence but never asserts that a conversation entry replaces it, so it passes on exactly the broken behaviour.

### One observation, not a blocker

The fallback `QueueEnqueueOutcome` representation and `RuntimeKernel`'s complete embedded queue authority both survive this change. That is consistent with the deletions being safe today, but it means the second authority still exists in the tree even though nothing in production reaches it.

---

*ran:* Core → Storage → Runtime → Runtime Host → CLI builds, all green; Runtime Host 50/50, CLI 235/235, Runtime steering 39/39; deletion-reachability probes for the removed fallback tests; independent confirmation that `packages/cli` contains no `transcript_advanced` consumer.
*did-not-run:* the hosted exact-head run `32647365821` (job `97213551805`) was still in progress at the time of writing, so no terminal CI result is claimed here. Neither finding depends on it.

Astro-Han:
Follow-up on the CI caveat in my review: the hosted run at `946f4268282146af26f4ab2502fbcecba8fdfc35` has since finished and `test` is `completed/success`. Correcting the note that it was still in progress — the two findings above are unchanged, since neither depended on that result.

packages/runtime-host/src/server/message-coordinator.ts
```diff
@@ -568,12 +586,27 @@ export class HostMessageCoordinator implements RuntimeMessageAuthority {
             : failure('operation_conflict', 'Message identity has a different payload');
         }
       }
+      const durableAdmission = isCurrentEpoch
+        ? undefined
+        : await this.#durableProof.readSteeringAdmission(input.sessionId, input.messageId);
```
jackwener: [P1] This recovery proof has no production caller that can present its key after the fault it is meant to recover. A crash after `commitSteeringAdmission` durably appends the user row but before the queue push/submit receipt causes the mutation request to end as outcome-unknown. The reconnecting client does not replay mutations; CLI and Desktop both generate `messageId` inside each new submission, so a user resend uses the new epoch and a new ID and never reaches this old-epoch/same-ID branch. A production-shaped fault probe therefore produced durable admissions `[oldId, newId]`: the transcript contains two identical user rows, while only `newId` enters the delivery queue. The added test manually reuses the old ID, proving a protocol capability that the clients do not retain or invoke. Please either make this durable fact self-rehydrate admission on Host recovery, or persist and automatically replay the exact original submit identity/payload from the client, with a reconnect fault test through the real client seam.

packages/runtime-host/src/server/message-coordinator.ts
```diff
@@ -727,6 +795,15 @@ export class HostMessageCoordinator implements RuntimeMessageAuthority {
           continue;
         }
         const result = { disposition, queueRevision: candidateRevision + 1 } as const;
+        if (disposition === 'steering' && !durableAdmission) {
```
jackwener: [P1] Durability is added only on the direct submit path. `queue.entry.promote` still moves a follow-up into `state.steering` and immediately makes it non-retractable without calling `commitSteeringAdmission`. The real Desktop path is the queue card's “Send now” action. If the user stops before the next provider step pulls it (or the Host dies), there is no transcript record that can recover the message. Commit the steering admission before exposing the promoted state, or keep the entry retractable until that commit succeeds; add a promote→Stop/crash regression.

packages/runtime-host/src/server/message-coordinator.ts
```diff
@@ -1451,22 +1532,28 @@ export class HostMessageCoordinator implements RuntimeMessageAuthority {
...
-    state.steering = [];
+  #retractFollowups(state: SessionState): RetractedMessageSnapshot[] {
+    const entries = state.followup;
     state.followup = [];
     for (const entry of entries) this.#releaseEntry(entry);
     return entries.map(retractedSnapshot);
   }
 
+  #discardQueuedSteering(state: SessionState): void {
+    const entries = state.steering;
```
jackwener, line 1913: [P1 — same promotion finding] This silent discard is the reachable loss sink for an undurable promoted entry. Stop returns snapshots only from `#retractFollowups`; `#discardQueuedSteering` releases entries without returning them. Diagnostic path: follow-up submit → promote succeeds with zero steering admissions → Stop yields `retracted=[]`, `steering=[]`, `followup=[]`. Because promotion created no durable row, the text is neither in the transcript nor restored to the draft.

jackwener (commented):
Reviewed on behalf of WAWQAQ under the standing Kabi review authorization, scoped to exact head `220bdebbae3d42c6e2c01f18f81fe80655fbadf5`.

**NO-GO: 2×P1.** The old CLI visibility P2 is closed: `subscription.transcript_advanced` now reloads canonical stored messages and publishes a transcript replacement rather than a live-only echo. The large deletion is also a real simplification: it removes CLI fallback/live-steering ownership and leaves Host queue + durable transcript as the authority.

Exact-head hosted `test` is completed/success; the PR is OPEN, non-draft, and mergeable against current main. Local focused verification passed: message coordinator 40/40, Host message integration 4/4, Runtime steering recovery 8/8, CLI durable/transcript 23/23, Desktop contract 1/1. The two inline P1s still block approval. No merge performed.

packages/runtime-host/src/server/root-turn-coordinator.ts
```diff
@@ -1966,6 +1990,7 @@ export class RootTurnCoordinator implements HostedExecutionAuthority {
     try {
       this.messages.reserveRootTurn(messageIdentity);
       messageReserved = true;
+      await this.messages.recoverRootTurn(messageIdentity);
```
jackwener: [P1] This is the only production call to `recoverRootTurn`, but it sits inside `prepareAdmittedTurn`, which is reached for recovery only by a `rootReplayAdmission` planned when no Run exists. The target crash is after the durable steering append and before queue commit, so an active Run necessarily already exists (`commitSteeringAdmission` rejects otherwise). Startup first terminalizes interrupted Runs, then recovery checks existing-Run admissions without re-entering this new-root path. Therefore the guard here cannot run for the state it is meant to repair. Reproduction: seed an existing Run plus one durable steering admission and no queue/provider proof, restart the real Host, and observe `durable=1 / provider proof=0 / root chain=1` instead of the expected successor chain of 2. Existing guards do not help because control never reaches this call.

packages/runtime-host/src/server/message-coordinator.ts
```diff
@@ -396,6 +418,75 @@ export class HostMessageCoordinator implements RuntimeMessageAuthority {
     state.phase = 'open';
   }
 
+  async recoverRootTurn(identity: RuntimeMessageRunIdentity): Promise<void> {
```
jackwener: [P1 corroborating evidence] The helper’s internal live-ID and immutable-proof checks are reasonable, but the new unit test calls this method directly and does not prove startup reachability. Deleting the sole production call at `root-turn-coordinator.ts:1993` left the complete Runtime Host suite green; focused coordinator/execution-host tests remained 45/45 and root-turn-coordinator tests 49/49. This mutation demonstrates that the shipped recovery behavior is untested and currently disconnected, not merely that one edge-case assertion is missing.

jackwener (commented):
Representative review action on behalf of @WAWQAQ; scope: fix-head re-review of the two previously reported Runtime Host steering-recovery findings; exact head `d526e8bfd01762561bc80a45e979efbd0c5aeccf`.

Result: **NO-GO — 1×P1 remains.**

The previous promote → Stop content-loss finding is closed: promotion now durably commits the steering admission before removing the follow-up entry, and a failed durability write leaves the entry retractable. Stop can still discard the live steering queue, but it no longer deletes the durable transcript row or fabricates provider delivery.

The automatic-recovery finding remains: the helper exists, but its only production call is behind a new-root admission path that the target crash state cannot reach. A production Host restart reproduction retains the durable user message but never delivers it to the model (`durable row = 1`, `provider steering proof = 0`, `root admission chain = 1`, where recovery requires a successor/root chain of 2). See the two inline comments for the wiring and mutation evidence.

**Required repair boundary:** do not fix this by replaying every durable steering row that lacks a provider proof. Crash-before-queue-commit and an intentional Stop-discard currently leave the same durable-row shape. Blind replay would resend a message after the user explicitly pressed Stop. The repair must durably distinguish recovery-eligible queue state from intentional discard (or provide an equivalent fence) while preserving Stop’s “no successor” contract.

Local verification on this head: focused coordinator tests 41/41, Runtime Host full suite green, and root build green. The PR is OPEN/non-draft/MERGEABLE; exact-head hosted `test` is completed/success. This review is COMMENT only; no approval or merge.

packages/runtime-host/src/server/message-coordinator.ts
```diff
@@ -1003,6 +1224,24 @@ export class HostMessageCoordinator implements RuntimeMessageAuthority {
       }
       return failure('not_found', 'Message queue entry does not exist');
     }
+    const pending = await this.#receipts.commitPendingSteering({
```
ARE404: [P1] A failed promotion can later execute a message that the user successfully retracted. This call first creates the pending recovery record; if `commitSteeringAdmission` then rejects, the new regression at `message-coordinator.test.ts:664-693` deliberately leaves the entry in `followup`. `queue.retract` removes that entry via `#retractFollowups`, but never settles this pending record, so `recoverPendingAfterHostRestart()` materializes and starts it after a restart. Repro: promote → inject transcript-append failure → retract the still-visible follow-up → restart; the supposedly retracted instruction becomes a recovered root. A normal retry cannot reliably heal the cut either: callers generate a fresh `admittedAt`, while the production SQLite store treats a timestamp change for the same session/message identity as `Pending steering admission identity conflict` (the in-memory fake silently returns the old record). Once this record is committed, either make the entry non-retractable/cancel the record atomically on every retract path, and make retries identity-stable.

packages/runtime-host/src/server/message-coordinator.ts
```diff
@@ -496,6 +623,27 @@ export class HostMessageCoordinator implements RuntimeMessageAuthority {
...
+    if (!state?.reservedRoot || !sameRun(state.reservedRoot, identity)) return;
+    if (state.steeringDiscardPreparedFor) {
+      if (!sameRun(state.steeringDiscardPreparedFor, identity)) {
+        throw new RuntimeMessageAuthorityInvariantError(
+          'Steering discard preparation belongs to another root Turn',
+        );
+      }
+      return;
+    }
+    await this.#receipts.settlePendingSteering(
```
ARE404: [P1] This deletes the only restart-delivery record before the Stop fence is committed or durable. Every Stop caller awaits `prepareStopFence()` first; only afterward does `declareStopFence()` commit the in-memory queue fence and await `claimRunClosure`. A Host crash/failure in that window leaves the admitted user row visible in the transcript but no pending record, and startup terminalizes the interrupted run before pending-steering recovery, so the message is never delivered and no successor is created even though Stop itself never became durable. The successful-Stop behavior is fine; this is the pre-fence crash window. Couple pending settlement to a durable Stop/fence fact, or reverse the durable ordering so recovery can distinguish an accepted Stop from an interrupted attempt.

packages/runtime-host/src/server/message-coordinator.ts
```diff
@@ -396,6 +430,76 @@ export class HostMessageCoordinator implements RuntimeMessageAuthority {
...
+        }
+        await this.#root.materializeSteeringAdmissions(pending);
+        const sources = pending.map(pendingSteeringSource);
+        const started = await this.#root.startRecoveredSteering(
+          {
+            sessionId,
+            content: aggregateMessageContent(pending.map((entry) => entry.modelContent)),
+            submittedContent: aggregateMessageContent(pending.map((entry) => entry.content)),
+            sources,
+            initiatingConnectionId: pending[0]!.initiatingConnectionId,
```
ARE404: [P1] Restart recovery merges all pending steering for a session and binds the resulting root to only the first submitter's client capabilities. The uninterrupted handoff deliberately uses `sameInitiatingClientPrefix()` to split B-then-A follow-ups into separate roots, and `root-turn-coordinator.test.ts` verifies each root sees its own client's ephemeral tools. After a crash, two unconsumed admissions from B and A instead reach this single batch; `startRecoveredSteering()` calls `bindConfirmedFollowup` with B's ID, so A's instruction executes with B's turn-affine Host tools/capabilities. Partition recovery into the same contiguous initiating-client batches as normal terminal handoff, starting each batch under its own connection identity.

ARE404 (commented):
Exact-head re-review at `064157d3f09039e7f9842be39d447db8f18a59c4`. **NO-GO: 3×P1, 1×P3.**

The previously reported disconnected restart recovery is now wired at Host startup, and the normal promote path correctly waits for durable transcript admission before exposing steering. Successful Stop also now preserves the transcript row while suppressing delivery. The three inline P1s are new crash/recovery boundary defects in that replacement design.

**[P3 — documented standard]** This diff changes renderer/UI behavior and stories, but the PR description has no before/after screenshots or recording as required by `CONTRIBUTING.md:85`.

No other documented-standard violations. The admission-time transcript model also semantically supersedes #3316's synthetic in-flight projection; textual overlap there is not an additional correctness finding. `git diff --check` is clean, and the only expected hosted workflow for these paths is CI `test`, which is terminal SUCCESS (run `32668976841`, job `97266724286`).

packages/runtime-host/src/server/message-coordinator.ts
Astro-Han: The direct-helper test and disconnected helper were removed. Coverage now enters through `createExecutionRuntimeHostComposition(...).recover()` with real SQLite stores, an existing running Run, one pending durable steering admission, and no in-memory queue/provider proof. The regression fails if the production startup call is removed and asserts the successor admission, old-Run terminalization, pending settlement, and non-duplicated user history.

packages/runtime-host/src/server/message-coordinator.ts
Astro-Han: Fixed at the Host-owned lifecycle boundary. A steering submission now commits `core_pending_steering_admissions` before the transcript append or in-memory queue mutation. Production startup calls `recoverPendingAfterHostRestart()` only after interrupted Runs have been terminalized and root recovery has completed, then admits one successor root from the exact stored identity and payload. The production composition regression seeds an existing running Run, the durable steering transcript row, and no queue/provider proof; restart produces exactly two root admissions, clears the pending record, and keeps only the original two user messages without a duplicate resend.

packages/runtime-host/src/server/message-coordinator.ts
Astro-Han: Fixed for promotion through the same lifecycle. `queue.entry.promote` now commits the pending steering admission and transcript row before moving the entry from follow-up to steering. The focused promotion test verifies the durable commit happens before the entry becomes non-retractable.

packages/runtime-host/src/server/root-turn-coordinator.ts
Astro-Han: Replaced the unreachable recovery path rather than extending it. `recoverRootTurn` and the transcript scan were removed. Production startup now invokes pending-steering recovery after `recoverInterruptedSessionsStrict()` and `coordinator.recover()`, when the interrupted existing Run is terminal and the root is idle. The production regression exercises this exact existing-Run crash state and observes the expected successor root chain of 2.

packages/runtime-host/src/server/message-coordinator.ts
Astro-Han: Fixed at the Stop/discard boundary. Promoted and directly submitted steering now have a pending durable admission. Intentional Stop/Drain settles those pending identities before `#discardQueuedSteering`; crash recovery preserves unresolved identities instead. The drain regression verifies the pending count changes from 1 to 0 before the queue is discarded.

packages/cli/src/pi-tui-runner.ts
```diff
@@ -540,8 +542,21 @@ export async function runMakaPiTui(input: MakaPiTuiInput): Promise<void> {
...
-      if (hydrateToolsWithStoredMessages(state, turnId, messages)) {
+      const newSteeringMessages = messages.filter(
+        (message): message is Extract<StoredMessage, { type: 'user' }> =>
+          message.type === 'user' &&
+          message.turnId === turnId &&
+          message.steeringEventId !== undefined &&
+          !transcriptMessageIds.has(message.id),
+      );
+      rememberTranscript(messages);
+      for (const message of newSteeringMessages) {
```
ARE404, line 551: [P2] This adds a second renderer for the same steering identity. Production persists the user row and advances the transcript, so reconciliation appends it here; provider consumption later emits `steering_message`, which `applyMakaSessionEventToTranscript` appends again without consulting `transcriptMessageIds`. The reverse callback order also duplicates: the event appends first, then this branch still sees a new transcript ID. The new `SteeringTurnDriver` fake emits reconciliation but never the production `steering_message`, so its containment assertion misses the double row. Please make one path authoritative (or dedupe the event by message ID) and cover both orderings.

apps/desktop/src/renderer/features/workbar/tools/side-chat/use-quote-companion.ts
```diff
@@ -203,6 +202,11 @@ export function useQuoteCompanion(input: UseQuoteCompanionInput): UseQuoteCompan
       });
     unsubscribeRef.current = sideChat.subscribeEvents(forkId, (event: SessionEvent) => {
       if (!mountedRef.current) return;
+      if (turnInFlightRef.current && activeTurnIdRef.current === null && event.turnId) {
+        activeTurnIdRef.current = event.turnId;
```
ARE404: [P2] The first event with any `turnId` cannot safely claim this send. `onBeforeSend` has just cleared `activeTurnIdRef`; a late terminal event from the prior root can arrive while Host admission starts a successor, claim the old ID here, and prevent the returned successor ID from being installed because the ref is non-null. Settlement then clears the wrong flight, while the real successor's messages are excluded by `ownTurnIdsRef`. Correlate ownership with the submit result/durable source identity instead of an arbitrary first event, and add the prior-terminal → successor-start race.

packages/runtime-host/src/server/message-coordinator.ts
```diff
@@ -1003,6 +1245,22 @@ export class HostMessageCoordinator implements RuntimeMessageAuthority {
       }
       return failure('not_found', 'Message queue entry does not exist');
     }
+    const pending = await this.#root.commitMessageAdmission(
```
ARE404: [P2] Promotion bypasses both admission-time capacity checks. The original submit was preflighted only as a `next_turn` follow-up; changing it to `current_turn` steering increases the encoded queue shape, while `worstCaseMessageQueueProjection` only changes entries already in `steering` to `in_flight` and reserves nothing for future promotion. A follow-up tuned just below `MESSAGE_QUEUE_PROJECTION_MAX_BYTES` is therefore accepted, then this durable commit plus the in-memory move creates an undecodable canonical projection and the refresh fail-stops/drains the Host. Build and preflight the promoted queue/root candidate before committing, with a boundary-size regression.

ARE404 (commented):
Exact-head re-review at `1343cee25bddb40b84456d7f15f08d74d8ffc0f7`. **NO-GO: 3×P2, 1×P3.**

The three P1 crash/recovery defects from my prior review are closed: promotion is now atomic with transcript materialization, Stop suppression waits for immutable terminal Stop proof, and queued successors execute under the explicit durable Session capability contract. The replacement design still has three independent product regressions, documented inline.

**[P3 — documented standard]** `CONTRIBUTING.md` requires before/after screenshots or a recording for UI changes. This PR changes observable CLI pending/transcript rendering and Desktop existing-session submission behavior, but the description explicitly omits media; focused tests do not satisfy that evidence rule.

`git diff --check` is clean. Exact-head dependency `audit` is SUCCESS; hosted `test` is still IN_PROGRESS, so I do not claim a terminal CI result.

<details>
<summary>简体中文</summary>

本次复审锁定 `1343cee2`，结论为 **NO-GO：3×P2、1×P3**。旧 review 的三条 P1 均已关闭：promotion 与 transcript 写入已原子化，Stop 丢弃已由不可变终态事实约束，durable successor 也不再继承临时 client capability。

当前三条 P2 分别是：CLI 同一 steering 同时走 durable transcript 与 `steering_message` 两条渲染路径，导致重复显示；Quote Companion 在发送竞态中会把任意旧 turn 的首个事件误认成本次 turn，随后丢失真正 successor 的回复；follow-up promotion 未重新执行 queue/canonical projection 容量预检，边界大小输入可把 Host 推入 fail-stop。另有一条 P3：可观察 UI 行为已变化，但缺少 CONTRIBUTING 要求的前后截图或录屏。

当前 `audit` 已通过，`test` 仍在运行中。

</details>

M4n5ter (commented):
<details open>
<summary>English</summary>

## Review result: NO-GO (3 × P1)

The architectural direction is sound: this change genuinely removes the RuntimeKernel/CLI fallback queue authorities instead of merely moving them. However, the new unified durable authority still has three reachable crash cuts that violate the PR's no-loss/no-resurrection/restart guarantees. I reproduced each with a focused probe on this exact head.

Validation completed: Core, Storage, MCP, Runtime, and Runtime Host builds; 107 focused Runtime Host tests; 18 SQLite core-execution tests; and 5 focused AiSdk steering tests. All existing tests pass. The three added diagnostic probes fail as described in the inline comments; they were removed afterward and the worktree is clean. Hosted `audit` is green; hosted `test` was still in progress at the final freshness check.

Simplify audit: production-like changes are +1299/-1085 and test-like changes are +1443/-1929. Across the relevant RuntimeKernel and CLI files, +179/-1009 removes `pendingFallback` (16 → 0 references), `takePendingFollowup` (5 → 0), and `confirmLiveTurn` (21 → 0). Please preserve that single-authority direction while closing these crash cuts; reintroducing a fallback owner would be the wrong fix.

</details>

<details>
<summary>中文</summary>

## 审查结论：NO-GO（3 × P1）

架构方向成立：这次确实删除了 RuntimeKernel / CLI 的 fallback queue authority，不是简单搬目录。但新的统一 durable authority 仍有三个可达崩溃切点，分别破坏本 PR 承诺的“不丢消息 / 不复活 / 可重启恢复”。我在这个 exact head 上分别做了最小探针复现。

已完成验证：Core、Storage、MCP、Runtime、Runtime Host 构建；Runtime Host 相关 107 个测试；SQLite core-execution 18 个测试；AiSdk steering 定向 5 个测试，现有测试全部通过。三个临时诊断探针均按内联评论描述失败，随后已移除，worktree clean。最终刷新时 hosted `audit` 已绿，hosted `test` 仍在运行。

简化审计：production-like +1299/-1085，test-like +1443/-1929；相关 RuntimeKernel 与 CLI 文件合计 +179/-1009，并把 `pendingFallback`（16 → 0）、`takePendingFollowup`（5 → 0）、`confirmLiveTurn`（21 → 0）完全删掉。修复以下崩溃切点时应保留单一 authority 方向，不要重新引入 fallback owner。

</details>

packages/runtime-host/src/server/message-coordinator.ts
```diff
@@ -396,6 +413,78 @@ export class HostMessageCoordinator implements RuntimeMessageAuthority {
...
+        const pending: PendingMessageAdmission[] = [];
+        const settled: string[] = [];
+        for (const candidate of durable) {
+          const source = await this.#durableProof.readRootTurnSourceMessageReceipt(
+            sessionId,
+            candidate.messageId,
+          );
+          const consumed = source
+            ? undefined
+            : await this.#durableProof.readImmutableSteeringMessageProof(
```
M4n5ter: [P1] A persisted steering echo is not proof that the provider consumed the message. `AiSdkBackend` persists the `steering_message` event before adding it to the provider input and before the next request is dispatched (`ai-sdk-backend.ts:4532-4553`). Its own abort comment says the ledger owns the message and **the next turn replays it**, while also stating that no request has yet been built. A stop or process exit can therefore land after this proof exists but before the model sees the message; the existing final-drain test demonstrates exactly that window with one provider call. On restart, this branch treats the echo as settled and GC's the admission, so the promised next-turn replay cannot happen. My focused probe inserted an admission plus only this immutable echo; recovery produced zero batches and deleted the admission. Please represent provider consumption with a fact that is actually downstream of dispatch, or keep echo-only admissions unresolved and carry them into the successor root.\n\n<details>\n<summary>中文</summary>\n\n[P1] 已持久化的 steering echo 不能证明 provider 已经消费消息。`AiSdkBackend` 在把 steering 加入 provider 输入、发起下一次请求之前，就先持久化 `steering_message`（`ai-sdk-backend.ts:4532-4553`）；它自己的 abort 注释同时写明“请求尚未构建”且“下一 turn 会重放”。Stop 或进程退出因此可以落在“echo 已存在、模型尚未看到消息”的窗口，现有 final-drain 测试也已经证明该窗口只发生一次 provider call。重启后这里却把 echo 当成 settled 并 GC admission，作者声明的 next-turn replay 根本不会发生。我的定向探针只写 admission + immutable echo，实际 recovery batch=0 且 admission 被删除。请使用真正位于 dispatch 下游的消费事实，或把只有 echo 的 admission 继续视为 unresolved 并带入 successor root。\n\n</details>

packages/runtime-host/src/server/message-coordinator.ts
```diff
@@ -396,6 +413,78 @@ export class HostMessageCoordinator implements RuntimeMessageAuthority {
...
+              );
+          const stopped =
+            !source &&
+            !consumed &&
+            (await this.#durableProof.readExplicitStopProof(sessionId, candidate.runId));
+          if (source || consumed || stopped) settled.push(candidate.messageId);
+          else pending.push(candidate);
+        }
+        if (settled.length > 0) {
+          await this.#receipts.garbageCollectMessageAdmissions(sessionId, settled);
```
M4n5ter, line 445: [P1] Explicitly stopped messages need durable retraction settlement, not plain GC. This restart path merges `stopped` with source/consumed admissions and deletes all of them, while the live terminal path below correctly sends stopped identities to `commitMessageRetractions` (`:505-526`). After queueing a follow-up, recording the exact Run's Stop proof, restarting, and retrying the same message ID/content, my probe was accepted again instead of returning `operation_conflict`; the cancelled message can execute under its resurrected identity. Please split the recovery sets the same way as the live path: GC source/consumed, commit retractions for stopped.\n\n<details>\n<summary>中文</summary>\n\n[P1] 显式 Stop 的消息需要写 durable retraction settlement，不能只做 GC。这个重启路径把 `stopped` 与 source/consumed 合并后一起删除，而下方 live terminal 路径（`:505-526`）会正确调用 `commitMessageRetractions`。我的探针先排入 follow-up、写入 exact Run 的 Stop proof、重启，再用同一 message ID/content 重试；实际重新成功，而不是 `operation_conflict`，被取消的消息身份因此可以复活并执行。请与 live 路径同样分流：source/consumed 做 GC，stopped 写 retraction settlement。\n\n</details>

packages/runtime-host/src/server/root-turn-coordinator.ts
```diff
@@ -1909,6 +1959,19 @@ export class RootTurnCoordinator implements HostedExecutionAuthority {
     if (unavailableReason) {
       return completedStart(operationUnavailable(unavailableReason));
     }
+    const initialUserMessagesMaterialized = admission.sourceMessages.some(
+      (source) => source.disposition === 'steering',
+    );
+    if (initialUserMessagesMaterialized) {
+      await this.manager.materializeRootSourceMessages({
```
M4n5ter, line 1966: [P1] A crash after source materialization but before Run creation prevents every later Host startup. For a queued root containing steering, this persists each source UserMessage first (`runtime-kernel.ts:2281-2299`) and only then checks/creates the Run. If the process exits after any append, `prepareHostedExecutionRecovery` sees a UserMessage with no Run and throws at `hosted-execution-recovery.ts:147-150`. A SQLite-backed startup probe with exactly `root admission + materialized source + no Run` exited before readiness with `Admitted Turn ... has a UserMessage but no Run`. Existing recovery tests cover the states before materialization or after Run creation, not this intermediate durable state. Please either make admission/materialization/Run establishment recoverably atomic, or teach recovery to validate and resume this partially materialized state idempotently.\n\n<details>\n<summary>中文</summary>\n\n[P1] source materialization 之后、Run 创建之前崩溃，会让此后每次 Host 启动都失败。含 steering 的 queued root 会先逐条持久化 source UserMessage（`runtime-kernel.ts:2281-2299`），之后才检查/创建 Run；进程在任一次 append 后退出，`prepareHostedExecutionRecovery` 会在 `hosted-execution-recovery.ts:147-150` 看到“有 UserMessage、无 Run”并直接抛错。SQLite-backed 启动探针精确构造 `root admission + materialized source + no Run`，实际在 readiness 前退出，错误为 `Admitted Turn ... has a UserMessage but no Run`。现有恢复测试只覆盖 materialization 前或 Run 已存在，没覆盖这个中间 durable state。请让 admission/materialization/Run 建立在恢复意义上原子，或让 recovery 能验证并幂等续跑这类部分物化状态。\n\n</details>

packages/runtime-host/src/server/message-coordinator.ts
Astro-Han: Addressed in 1343cee25. Retraction now commits a durable settlement before removing the in-memory queue entry. The settlement records the original placement and content digest, so retries from both the old and current Host epochs return operation_conflict. The regression `retract settles a failed promotion so restart cannot recover it` covers promotion failure → retract → restart → retry, and promotion retries reuse the canonical durable admission timestamp.

packages/runtime-host/src/server/message-coordinator.ts
Astro-Han: Addressed in 1343cee25. The immutable terminal Stop fact is now the cut: a crash before that fact leaves the admission unresolved and recoverable, while `prepareTerminalTransition` commits durable retraction settlements for unconsumed admissions only after the exact Run has explicit Stop proof. This distinguishes an unconfirmed Stop attempt from a confirmed Stop without deleting the restart-delivery record in the pre-fence window.

packages/runtime-host/src/server/message-coordinator.ts
Astro-Han: Superseded by the durable Session execution contract in 1343cee25. Queued and recovered roots no longer inherit ephemeral capability bindings from the initiating Client. `admitQueuedMessageRoot` installs an explicit durable Session binding through `bindDurableSession`, and recovery preserves durable queue order under that single contract. The recovery regression covers admissions from two Clients aggregating into one recovered batch, while capability tests verify that dynamic Client tools are excluded.

packages/runtime-host/src/server/message-coordinator.ts
```diff
@@ -727,17 +936,37 @@ export class HostMessageCoordinator implements RuntimeMessageAuthority {
           continue;
         }
         const result = { disposition, queueRevision: candidateRevision + 1 } as const;
+        let durableAdmittedAt: number | undefined;
+        if (!durableAdmission) {
+          const admitted = await this.#root.commitMessageAdmission(
```
jackwener: [P3] Recompute the returned revision after this durable-admission await. The preflight freezes `candidateRevision` and `result` before this call, but `#pull/#ack/#nack` intentionally mutate the same session state outside the admission lock. In a delayed-commit probe, pull+ack advanced the queue while this await was suspended; submit then appended its entry and returned `queueRevision: 2` while the authoritative projection was already revision 4. Current Desktop/CLI callers do not use this result as mutation authority, so this is non-blocking today, but the wire result is false and can mislead future clients. Fence/retry after the durable commit or construct the receipt from the final state, and cover a paused `commitMessageAdmission` racing with pull+ack.

jackwener (commented):
Kabi exact-head review under the standing WAWQAQ review authorization, scoped to `1343cee25bddb40b84456d7f15f08d74d8ffc0f7`.

**NO-GO — Kabi grading: 2×P1, 2×P2, 1×P3.** Four defects were independently reproduced after a sealed provisional judgment. To avoid opening duplicate threads, their current-head anchors are: provider-echo recovery loss ([P1](https://github.com/apache/maka/pull/3633#discussion_r3840849430)), materialized-source/no-Run startup failure ([P1](https://github.com/apache/maka/pull/3633#discussion_r3840849439)), Stop-recovery identity resurrection ([P2 in this line](https://github.com/apache/maka/pull/3633#discussion_r3840849435)), and double CLI rendering of one steering identity ([P2](https://github.com/apache/maka/pull/3633#discussion_r3840817111)). The non-duplicate P3 is inline.

The architectural simplification itself is real: durable Host/Storage state is now the message authority, while the old CLI fallback/live-steering ownership has been removed. The repair should preserve that direction. A post-seal promotion-capacity candidate was withdrawn after boundary probing: submit preflight reserves the worst-case revision width and rejects larger candidates before promotion. The Quote Companion race was not included in this line's independent count because the final hook-level reproduction was not completed.

Verification on this head: Runtime Host 1134/1134, CLI 423/423, UI 223/223, Desktop 1339/1339; hosted `test` and `audit` are completed/success. Fresh gate: OPEN, non-draft, MERGEABLE/BLOCKED, six unresolved P0–P2 threads, and zero current or stale `APPROVED` reviews. COMMENT only; no approval or merge.

jackwener (commented):
Supplement to review `5005153696` on the unchanged exact head `1343cee25bddb40b84456d7f15f08d74d8ffc0f7`: the Quote Companion hook-level probe completed before the stop instruction arrived and independently reproduced the existing [P2 thread](https://github.com/apache/maka/pull/3633#discussion_r3840817114). The correct Kabi count is therefore **NO-GO — 2×P1, 3×P2, 1×P3**. The earlier sentence excluding this race from Kabi independent reproduction is superseded; all other findings and the promotion-capacity withdrawal remain unchanged.\n\nThe exact hook sequence was: settle `turn-old`; begin a second send and execute `onBeforeSend` while the Host result is pending (`activeTurnId=null`, in-flight true); deliver a late terminal event for `turn-old`, which claims the active ref; return `turn-new`, whose ID is then not installed because the ref is non-null; deliver `turn-new` text, which remains absent from the live projection; release the old settlement, after which the hook reports streaming false while the real Host turn remains active. The React hook plus fake Workbar production service seam reproduced this in 30.5 ms; temporary probe removed and worktree clean. COMMENT only; no approval or merge.

packages/cli/src/pi-tui-runner.ts
Astro-Han: Fixed in 23455cca0 and c6a899ea5. Both the durable reconciliation path and the live steering event now deduplicate by the stable messageId owned by MakaPiTranscriptState; the duplicate runner-level identity set was removed. The regression covers durable-first and event-first arrival, and the focused CLI suite passes 204/204.

apps/desktop/src/renderer/features/workbar/tools/side-chat/use-quote-companion.ts
Astro-Han: Fixed in 2e343a149 and 0d110db05. An arbitrary first event can no longer claim the send. Steering binds only through a queue_update or steering_message carrying the submitted messageId, while a new root binds to the Host-returned turnId. Events arriving before that authority is known are buffered transiently and only matching-turn events are replayed. Regressions cover both prior-terminal to successor steering and new-root text plus complete arriving before the send result.

packages/runtime-host/src/server/message-coordinator.ts
Astro-Han: Fixed in ee0529a4d. Promotion now builds the promoted queue and future root sources, runs the shared queue, canonical Session snapshot, interrupt-result, and root-admission capacity preflight, then rechecks revision, generation, root identity, and entry position before durable mutation. The boundary regression proves a follow-up that fits as next_turn but exceeds the in-flight steering limit is rejected without mutation.

packages/runtime-host/src/server/message-coordinator.ts
Astro-Han: Fixed in c324308f7. Restart settlement now separates explicit Stop from ordinary GC and commits durable message retractions for stopped identities. The regression restarts after Stop and verifies that retrying the same message identity remains an operation_conflict.

packages/runtime-host/src/server/message-coordinator.ts
Astro-Han: Fixed in c324308f7. Restart recovery no longer treats a steering_message echo as provider consumption. Only a root source receipt settles the admission; echo-only steering remains pending and is recovered into the successor root. The focused restart regression covers exactly echo persisted with no later provider dispatch.

packages/runtime-host/src/server/root-turn-coordinator.ts
Astro-Han: Fixed in 0538200a5. Hosted recovery now validates materialized source UserMessages as the exact admission prefix and resumes the same root admission idempotently, while still rejecting wrong order, reused IDs, content mismatch, and incomplete sources when a Run already exists. SQLite-backed startup regressions cover both a partial prefix and all sources materialized before Run creation.

packages/runtime-host/src/server/message-coordinator.ts
Astro-Han: Fixed in 8d591b942. The submit result is now constructed after the entry mutation from the authoritative state.revision. The paused commitMessageAdmission regression races pull plus ack while the durable commit is suspended and verifies both the returned and projected revision are 4 rather than the stale preflight revision 2.

packages/runtime/src/runtime-kernel.ts
```diff
@@ -2393,175 +2278,25 @@ export class RuntimeKernel implements RuntimeKernelLike {
...
+  async materializeRootSourceMessages(input: {
+    sessionId: string;
+    turnId: string;
+    messages: readonly { messageId: string; content: MessageContent }[];
+  }): Promise<void> {
+    const existingIds = new Set(
+      (await this.deps.store.readMessages(input.sessionId)).map((message) => message.id),
+    );
+    for (const message of input.messages) {
+      if (existingIds.has(message.messageId)) continue;
```
Astro-Han: [P1] Steering folded into a successor is never rebound, so every later Host start fails.

`materializeRootSourceMessages` dedupes across the **entire Session**: `readMessages(input.sessionId)` collects every existing message id, and a hit takes the `continue`. It only skips the write — it **does not rebind that row's `turnId` to the successor**.

Trigger path: the current root accepts steering and its canonical `UserMessage` is already written under the predecessor turn; the backend reaches a terminal state before pulling that steering; the steering is folded into the successor; the successor starts with `recordInitialUserMessage:false`, and this session-wide dedupe leaves the source row under the predecessor forever.

Consequence: the successor completes this time, but on the next Host start recovery sees the Run already exists while no source row sits under the successor, and throws `Admitted queued Turn <id> has incomplete source materialization`. Readiness fails, so **the Host for the whole workspace will not come up**. A mixed steering + follow-up case fails the same way at source index 0.

Why the existing guards do not catch it: atomic admission binds to the predecessor; the global dedupe prevents the write rather than completing a rebind; partial-prefix recovery only applies before the Run is created.

Suggested direction: keep the single canonical identity, but let the successor admission atomically rebind/move this row — or have materialization and recovery explicitly accept and validate steering provenance from the predecessor.

<details>
<summary>简体中文</summary>

[P1] 折叠到 successor 的 steering 没有重新绑定，之后每次 Host 启动都会失败。

`materializeRootSourceMessages` 的去重是**整个 Session 范围**的：`readMessages(input.sessionId)` 取出所有已存在的 message id，命中就 `continue`。它只是跳过写入，**不会把这条 row 的 `turnId` 改绑到 successor**。

触发路径：当前 root 接受了 steering，canonical UserMessage 已经写在 predecessor turn 下；backend 还没 pull 这条 steering，root 就进入终态；steering 被折叠进 successor；successor 以 `recordInitialUserMessage:false` 启动，而这里的全局去重让 source row 始终留在 predecessor 名下。

后果：successor 当次能跑完，但下一次 Host 启动时，恢复流程发现 Run 已存在、而 source row 不在 successor 下，抛 `Admitted queued Turn <id> has incomplete source materialization`，readiness 失败——**整个 workspace 的 Host 起不来**。混合 steering + followup 的情况会在 source index 0 上以同样方式失败。

为什么现有守卫拦不住：atomic admission 绑定的是 predecessor；全局去重是阻止写入而不是完成改绑；partial-prefix recovery 只在 Run 创建之前才允许介入。

修复方向：保持单一 canonical identity 不变，但让 successor admission 能够原子地改绑/移动这条 row；或者让 materialization 与 recovery 显式接受并校验来自 predecessor 的 steering provenance。

</details>

packages/storage/src/message-receipt-store.ts
```diff
@@ -58,6 +84,18 @@ export interface MessageReceiptStore {
...
   ): Promise<MessageOperationReceipt>;
+  readMessageAdmission(
+    sessionId: string,
+    messageId: string,
+  ): Promise<PendingMessageAdmission | undefined>;
+  readMessageSettlement(
+    sessionId: string,
+    messageId: string,
+  ): Promise<MessageAdmissionSettlement | undefined>;
+  listPendingMessages(): Promise<readonly PendingMessageAdmission[]>;
```
Astro-Han, line 95: [P1] A queued-message edit that already took effect reverts to the original instruction after a Host restart.

This store's interface exposes `readMessageAdmission`, `listPendingMessages`, `commitMessageOrder`, `commitMessageRetractions` and `garbageCollectMessageAdmissions` — but **no method that updates a pending admission's content / modelContent / digest**.

Meanwhile the edit path at `packages/runtime-host/src/server/message-coordinator.ts:1405-1407` is:

```
queued.entry.content = content;
queued.entry.modelContent = modelContent;
this.#mutated(state);
```

That is a purely in-memory rewrite. The `#commitReceipt('update_entry', ...)` that follows writes a Host-epoch receipt and does not touch the admission itself. Startup recovery goes through `listPendingMessages()` at `message-coordinator.ts:423`, so it reads back the **original content persisted at enqueue time**.

Observed: the live projection shows the edited text; after SIGKILL, the successor source recovered from `core_message_admissions` is the pre-edit text.

Consequence: an old instruction the user explicitly corrected can actually execute after an ordinary crash recovery, with whatever side effects it carries. This is not a display issue.

Why the existing guards do not catch it: the revision / identity / capacity / snapshot checks all protect live mutation.

Suggested direction: update this pending admission's content / modelContent / digest inside the same durable transaction, fencing identity, placement and settlement — or write a new identity and durably retract the old one.

<details>
<summary>简体中文</summary>

[P1] 已经生效的 queued message 编辑，在 Host 重启后会退回成编辑前的原始指令。

这个 store 的接口里有 `readMessageAdmission`、`listPendingMessages`、`commitMessageOrder`、`commitMessageRetractions`、`garbageCollectMessageAdmissions`，**唯独没有任何一个方法可以更新一条 pending admission 的 content / modelContent / digest**。

而 `packages/runtime-host/src/server/message-coordinator.ts:1405-1407` 的编辑路径是纯内存改写 `queued.entry.content = content`，随后 `#commitReceipt('update_entry', ...)` 写的是一条 Host-epoch receipt，不改 admission 本身。启动恢复走的是 `message-coordinator.ts:423` 的 `listPendingMessages()`，读到的仍然是入队时落盘的**原始内容**。

复现结果：live projection 显示编辑后的文本，SIGKILL 之后从 `core_message_admissions` 恢复出来的 successor source 是编辑前的原文。

后果：用户已经确认改掉的旧指令，会在一次普通的崩溃恢复之后被真正执行，可能带来副作用。这不是显示问题。

为什么现有守卫拦不住：revision / identity / capacity / snapshot 这几道检查保护的都是 live mutation。

修复方向：在同一个 durable 事务里更新这条 pending admission 的 content / modelContent / digest，并对 identity、placement、settlement 加 fence；或者改成写入新 identity 并 durable retract 旧 identity。

</details>

Astro-Han (commented):
Reviewed at head `b919db88e02d5c093dfeae10d61a4ee2d0cb1489`. Conclusion: **NO-GO, 2 × P1**. Neither finding comes from reading thread resolution state; both were traced through the code and reproduced with an executable probe.

The 15 historical threads were re-judged against the code and are genuinely fixed, and the simplification does land: `RuntimeKernel`'s queue authority, the CLI fallback/mirror, the Host synthetic transcript and the renderer's `unconfirmed` state are all really gone.

The problem is in what this rebase newly pulled in. Main's `queue.entry.update` (#3635) was integrated, but it was not folded into the durable admission transaction this PR itself introduces.

Both P1s are described in the inline comments below. The shared fix boundary is: **an edit and a fold must both land inside the same durable admission transaction, rather than mutating memory or writing a receipt only.**

Two test gaps to close as well:

- an integration test for late steering to successor across a restart
- a precise test for queue, edit, SIGKILL, recovery

The existing edit test is in-memory and never restarts, so it cannot catch the second P1.

<details>
<summary>简体中文</summary>

在当前 head `b919db88e02d5c093dfeae10d61a4ee2d0cb1489` 上复审，结论是 **NO-GO，2 个 P1**。两条都不是靠读 thread 的 resolved 状态得出的，是按代码链路复核并用可执行脚本复现的。

历史 15 条线程逐条按代码重新判过，都是真修复；简化方向也确实达成了（RuntimeKernel 的 queue authority、CLI fallback/mirror、Host synthetic transcript、renderer 的 `unconfirmed` 都真正消失了）。问题出在这次 rebase 新并进来的部分——把 main 的 `queue.entry.update`（#3635）接进来时，没有同时接进这个 PR 自己新建的 durable admission 事务。

两条 P1 详见下面的行内评论。共同的修复边界是：**编辑和折叠都必须落到同一个 durable admission 事务里，而不是只改内存或只写 receipt。**

另外补两条测试缺口：late steering 到 successor 再重启的集成测试；queue 编辑后 SIGKILL 再恢复的精确测试。现有的 edit 测试是纯内存、不重启的，所以覆盖不到第二条 P1。

</details>

Astro-Han (commented):
Re-checked at `07a3adb9c`. **Both blocking issues from my earlier review are resolved.** I have no remaining P0-P2.

**Queued message edits now survive a restart.** `updatePendingMessage` was added to the receipt-store interface and implemented in the SQLite store, and `queue.entry.update` calls it *before* mutating the in-memory entry and before committing the receipt — so a failure to persist leaves the in-memory state untouched rather than diverging from it. The new test in `message-coordinator.test.ts` closes the loop end to end: edit, restart under a fresh host epoch, then assert the recovered sources carry the edited content. That is exactly the path that previously lost the edit.

**Source materialization no longer skips silently.** `materializeRootSourceMessages` now resolves each existing message and validates it — type, content, and turn binding — instead of testing membership in a set of ids and moving on. A message already bound to a different turn raises `Queued root source … conflicts with its transcript` rather than being quietly left where it was. The one legitimate exception is stated explicitly: a `steering` source may sit under `previousRootTurnId`.

`hosted-execution-recovery.ts` was reworked to match, which is what makes the pair coherent. Verification is now by message identity rather than positional index, it permits the same steering-under-predecessor case, and it still rejects identity reuse and holes — a missing source followed by a present one raises `incomplete source materialization`, so prefix semantics hold.

Checks on my side: CI is terminal green on this exact head (run [32714419591](https://github.com/apache/maka/actions/runs/32714419591)). I also merged this into current `main` — `8801edc32`, six commits ahead of this branch — which merges cleanly, and ran the affected `runtime-host` suites on the merged tree. Six tests fail there, but all six fail with the identical error `execution Host exited before readiness: 1`, and five of them fail the same way on a clean `main` checkout in the same environment. That is a local toolchain problem — `@maka/runtime` will not build here because `@slack/web-api`, `@slack/socket-mode`, and a compatible `https-proxy-agent` are missing — so the execution Host child cannot boot. Nothing is attributable to this change.

This still needs an approving review from an account other than the one that authored it.

在 `07a3adb9c` 上重新核对：**我先前 review 里的两个阻塞问题都已解决**，没有遗留的 P0-P2。

**队列消息的编辑现在能挺过重启。** receipt-store 接口新增了 `updatePendingMessage` 并在 SQLite 存储中实现，`queue.entry.update` 会在改动内存条目之前、也在提交 receipt 之前调用它——因此持久化失败时内存状态保持不变，不会与存储分叉。`message-coordinator.test.ts` 里新增的用例把整条链路闭环了：编辑、在新的 host epoch 下重启、然后断言恢复出来的 sources 带的是编辑后的内容。这正是之前会丢失编辑的那条路径。

**源消息物化不再静默跳过。** `materializeRootSourceMessages` 现在会取出每条已存在的消息并校验它——类型、内容、以及 turn 绑定——而不是在一个 id 集合里判断存在与否然后跳过。已经绑定到别的 turn 的消息会抛出 `Queued root source … conflicts with its transcript`，而不是被悄悄留在原处。唯一合法的例外被显式写了出来：`steering` 类型的源可以位于 `previousRootTurnId` 之下。

`hosted-execution-recovery.ts` 做了对应改造，这是让两边自洽的关键。校验现在按消息 identity 而不是位置下标进行，允许同样的「steering 挂在前一个 root turn 下」的情形，并且仍然拒绝 identity 复用和空洞——先缺失、后又出现的源会抛出 `incomplete source materialization`，因此前缀语义得以保持。

我这边的检查：这个 exact head 的 CI 终态绿（运行记录 [32714419591](https://github.com/apache/maka/actions/runs/32714419591)）。我还把它合进了当前 `main`（`8801edc32`，比本分支新 6 个提交），合并干净，并在合并后的树上跑了受影响的 `runtime-host` 测试。那里有 6 个用例失败，但 6 个的报错完全相同，都是 `execution Host exited before readiness: 1`，其中 5 个在同一环境下的干净 `main` 检出上也以同样方式失败。这是本地工具链问题——`@maka/runtime` 在这里构建不起来，因为缺少 `@slack/web-api`、`@slack/socket-mode` 和版本匹配的 `https-proxy-agent`——导致 execution Host 子进程无法启动。没有可归因于本次改动的问题。

这个 PR 仍然需要一条来自非作者账号的 approve。

</details>

packages/runtime/src/runtime-kernel.ts
Astro-Han: Fixed on the current head. Folded steering keeps its single transcript row under the predecessor, while successor materialization and startup recovery now explicitly validate that predecessor provenance instead of requiring a duplicate successor row. The regression now covers an edited predecessor steering source executing in the successor and surviving another Host restart.

packages/storage/src/message-receipt-store.ts
Astro-Han: Fixed on the current head. Queued edits now go through SessionAuthorityStore.updateMessageAdmission(), which atomically updates the exact unsettled admission and its existing steering transcript when applicable. MessageReceiptStore no longer owns a parallel content-update path. Identity, placement, disposition, timestamp, and settlement are fenced before the live queue mutates; focused storage, coordinator, and restart regressions pass.

Astro-Han:
Note on freshness: my earlier review is bound to `07a3adb9c`, which this branch no longer contains. Please do not read it as a statement about the current code.

The branch was rewritten rather than extended — between `07a3adb9c` and the current `eecd4d93d` there is a new commit chain, and the two commits I examined now correspond to different objects. The change is also considerably larger than what I looked at: 82 files and `+3989/-3158`, against 37 files previously.

I am not restating a verdict here. A fresh review is underway at `eecd4d93d`, and I will publish conclusions bound to whatever head is current when they are ready.

One environment note that affects how CI is read right now, unrelated to this PR: `main` is failing its `Check ASF source headers` step because five `packages/runtime/plugin-*` files added by #3250 are missing the license header. Any ASF header failure seen while validating against `main` comes from there, not from this branch.

关于时效性的说明：我先前那条 review 绑定在 `07a3adb9c` 上，而这个分支已经不包含该提交。请不要把它当作对当前代码的判断。

这个分支是被重写而不是被追加的——`07a3adb9c` 到当前 `eecd4d93d` 之间是一条新的提交链，我当时看的那两个提交现在对应的是不同的对象。改动规模也比我看过的大得多：82 个文件、`+3989/-3158`，而此前是 37 个文件。

我在这里不重述任何结论。针对 `eecd4d93d` 的新一轮复审正在进行，结论会绑定到届时的当前 head 再发布。

另有一条与本 PR 无关、但会影响当下 CI 判读的环境事实：`main` 的 `Check ASF source headers` 这一步正在失败，原因是 #3250 新增的五个 `packages/runtime/plugin-*` 文件缺少版权头。因此在对 `main` 做验证时看到的 ASF 头失败来自那里，不是来自这个分支。

</details>

Astro-Han (commented):
Thanks for the rework — the two problems I raised earlier are genuinely closed. Folded steering now survives a Host restart with source identity preserved on the successor admission, and a queued edit comes back as the edited content after a hard kill under a new Host epoch. Those were the hard ones.

Five things remain. The first three block merge.

---

**[P1] Crash recovery and normal cleanup classify a pending admission by different proofs**

`prepareTerminalTransition()` in `packages/runtime-host/src/server/message-coordinator.ts:507` settles a pending admission when `readImmutableSteeringMessageProof()` returns a proof, falling back to an explicit Stop proof. The restart path at `:431-440` does not consult the steering proof at all — it only asks `readRootTurnSourceMessageReceipt()` and then `readExplicitStopProof()`, and treats anything else as still pending.

That asymmetry is the bug. Take a steering message whose RuntimeEvent is already durable and already went into the next model request, with a durable provider-origin tool result after it. If the process dies before `prepareTerminalTransition()` collects the admission, the normal path would have settled it, but recovery does not: it finds no root-turn source receipt and no Stop proof, so it classifies the admission as pending and starts the same steering again as a successor root. With a side-effecting tool, the user's instruction runs twice.

A steering echo on its own must still recover, so the existing echo-only test is correct and should stay. What is missing is a settlement signal derived from durable downstream execution — either the same steering proof the cleanup path already trusts, or a strictly downstream causal fact. Please also add the crash cut where a provider-origin tool result is durable and the process dies before terminal cleanup.

**[P1] The durable-Session marker is not reconstructed when an admitted root is replayed**

`bindDurableSession()` in `packages/runtime-host/src/server/client-capability-coordinator.ts:238` writes only to the in-memory `#sessions` map, and its only production caller is `admitQueuedMessageRoot()` in `root-turn-coordinator.ts:2384`. The restart replay path never goes through that caller.

So if the process dies after the root admission commits but before the Run is created, recovery replays the root without ever setting the marker. Once the Host is ready and that replay settles, a later supervisor wake on the same Session builds a fresh backend through neither binding path, and because `state?.durable` is absent, the `if (!state?.durable)` branch at `:499` auto-discovers a newly registered call-affine Client offer. The identical wake cannot discover that offer without the crash, which inverts the durable-Session contract this PR states. Please persist or rebuild the durable mode before replay, and cover restart, replay, post-ready provider registration, and a later autonomous root in one test.

**[P1] A regression test added by this PR contradicts the new atomic admission contract**

`packages/runtime-host/src/__tests__/execution-composition.test.ts:316-335` commits the current-turn steering through `commitMessageAdmission`, which now materializes the canonical user transcript inside that transaction (`packages/storage/src/sqlite-session-metadata-store.ts:1575-1604`). The fixture then appends the same identity and payload again through the direct path. The exact-once assertion consequently reads `['initial-message', 'admitted-steering', 'admitted-steering']` where it expects two entries, and `production startup recovers steering admitted before an interrupted Run lost its queue` fails 0/1 every run.

Removing only the redundant append makes the isolated test pass, which locates this in the fixture rather than the implementation — the runtime materializer does check message identity, so this is not a user-visible duplicate. Please delete the redundant append or route the fixture through the production materializer, and keep the exactly-one assertion as it is. Please do not make the duplicate append silently idempotent; that would remove the very guarantee this test exists to hold.

This did not show up in the checks here because `test` stopped at the ASF header step inherited from the base and skipped everything after it. That is fixed on `main` now, so a rebase should let the run reach this failure.

**[P2] A 40,000-byte queue edit is accepted by every guard and then fails as an internal error**

The wire limit is 48 KiB (`packages/runtime-host/src/protocol/message.ts:335-345`), but the durable admission envelope is 64 KiB and serializes `content` and `modelContent` together with identity metadata (`packages/storage/src/message-receipt-store.ts:36,520-546`). One wire-valid text does not guarantee both durable representations fit.

Queue a short followup, then edit it to 40,000 ASCII bytes. The queue projection, root admission, and full snapshot preflights at `message-coordinator.ts:1373-1389` all pass, because none of them measures the envelope that `normalizePendingMessageAdmission` will actually build. The durable update then throws `Pending message admission exceeds size limit` past the guards at `:1403-1414`, and the client sees only `Runtime Host operation failed`. The transaction keeps the original text, so nothing is lost, but a legal input fails unpredictably with an internal error. Please reuse the same admission-size preflight before the durable boundary and return the declared bounded error, or make the contract guarantee that every wire-valid edit holds both representations. A test at the largest successful size would pin whichever you choose.

**[P3] Please add the before/after evidence for the UI changes**

This changes how queued and steering messages appear and when Side Chat goes live. `CONTRIBUTING.md` asks for before/after screenshots or a recording for UI changes and does not exempt behavior-only changes that add no styling. Please attach the Desktop and TUI states this lifecycle change affects.

---

One more thing outside the code: `main` has moved and this branch is now conflicting. It needs a rebase before it can merge regardless of the above, and I will re-review against the new head once you push.

<details>
<summary>简体中文</summary>

感谢返工——我之前提的两个问题确实解决了。folded steering 现在能跨 Host 重启存活，source identity 保留在 successor admission 上；queue edit 在硬杀并换新 Host epoch 后也能恢复为编辑后的内容。这两个是难的部分。

还剩五点，前三点阻塞合并。

**[P1] 崩溃恢复与正常清理用了不同的证据来判定未决 admission**

`packages/runtime-host/src/server/message-coordinator.ts:507` 的 `prepareTerminalTransition()` 在 `readImmutableSteeringMessageProof()` 返回证据时结算未决 admission，退化情况才看显式 Stop 证据。但 `:431-440` 的重启路径**完全不查** steering 证据，只问 `readRootTurnSourceMessageReceipt()`，再问 `readExplicitStopProof()`，其余一律当作仍然未决。

这个不对称就是 bug。设想一条 steering，它的 RuntimeEvent 已持久化、已进入下一次模型请求，其后还有一个已持久化的 provider 侧工具结果。若进程在 `prepareTerminalTransition()` 回收该 admission 之前崩溃，正常路径本会结算它，恢复路径却不会：它找不到 root-turn source receipt，也找不到 Stop 证据，于是判为未决，把同一条 steering 当作 successor root 重新启动。**若工具有副作用，用户的指令会被执行两次。**

单独的 steering echo 仍然必须能恢复，所以现有的 echo-only 测试是对的，应当保留。缺的是一个来自持久化下游执行的结算信号——要么就用清理路径已经信任的那份 steering 证据，要么用严格的下游因果事实。另请补上"provider 侧工具结果已持久化、随后在 terminal cleanup 之前崩溃"这个切面的测试。

**[P1] 重放已提交的 root 时没有重建 durable Session 标记**

`packages/runtime-host/src/server/client-capability-coordinator.ts:238` 的 `bindDurableSession()` 只写内存中的 `#sessions` map，而它在生产代码里唯一的调用者是 `root-turn-coordinator.ts:2384` 的 `admitQueuedMessageRoot()`。重启重放路径根本不经过这个调用者。

于是，如果进程在 root admission 提交之后、Run 创建之前崩溃，恢复会重放这个 root 而始终不设置该标记。等 Host ready、这次重放结算之后，同一 Session 上后续的 supervisor wake 会经由两条 binding 路径之外的方式建立新 backend；由于 `state?.durable` 缺失，`:499` 处 `if (!state?.durable)` 分支会自动发现一个新注册的 call-affine Client offer。没有崩溃时，同一个 wake 发现不了这个 offer——这与本 PR 声明的 durable Session 契约正好相反。请在重放前持久化或重建 durable 模式，并用一条测试覆盖重启、重放、ready 后注册 provider、以及之后的 autonomous root。

**[P1] 本 PR 新增的回归测试与新的原子 admission 契约自相矛盾**

`packages/runtime-host/src/__tests__/execution-composition.test.ts:316-335` 通过 `commitMessageAdmission` 提交当前 turn 的 steering，而该事务现在会在内部落地 canonical user transcript（`packages/storage/src/sqlite-session-metadata-store.ts:1575-1604`）。紧接着 fixture 又用直接 append 路径把同一 identity 和 payload 再写了一遍。于是 exact-once 断言实际读到 `['initial-message', 'admitted-steering', 'admitted-steering']`，而它期望两条；`production startup recovers steering admitted before an interrupted Run lost its queue` 每次都是 0/1 失败。

只删掉那段冗余 append，隔离运行该测试即可通过——这说明问题在 fixture 而不在实现：runtime 的 materializer 确实做了 message identity 检查，所以这不是用户可见的重复。请删掉冗余 append，或让 fixture 走生产 materializer，并原样保留 exactly-one 断言。**请不要把重复 append 改成静默幂等**，那会把这条测试本来要守住的保证一并抹掉。

这个问题没有在本 PR 的检查里暴露，是因为 `test` 先停在了从 base 继承的 ASF header 步骤，之后全部跳过。`main` 上已修复，rebase 后运行就能跑到这个失败。

**[P2] 40,000 字节的 queue edit 通过了全部守卫，然后以内部错误失败**

wire 上限是 48 KiB（`packages/runtime-host/src/protocol/message.ts:335-345`），但持久化 admission envelope 是 64 KiB，并且要把 `content` 与 `modelContent` 连同身份元数据一起序列化（`packages/storage/src/message-receipt-store.ts:36,520-546`）。**单份文本合法，不等于两份持久化表示装得下。**

先 queue 一条短的 followup，再把它编辑为 40,000 个 ASCII 字节。`message-coordinator.ts:1373-1389` 处的 queue projection、root admission、full snapshot 三道预检全部通过，因为它们都没有测量 `normalizePendingMessageAdmission` 真正会构造的那个 envelope。随后持久化更新在 `:1403-1414` 的守卫之外抛出 `Pending message admission exceeds size limit`，客户端只看到 `Runtime Host operation failed`。事务保留了原文，所以没有数据丢失，但一个合法输入以内部错误不可预期地失败了。请在持久化边界之前复用同一套 admission 尺寸预检并返回已声明的有界错误，或者让契约保证任何 wire 合法的编辑都能容纳两份表示。无论选哪种，都建议补一条"最大可成功尺寸"的测试来钉住边界。

**[P3] 请补充 UI 变更的前后证据**

本 PR 改变了 queued 与 steering 消息的呈现方式，以及 Side Chat 进入 live 状态的时机。`CONTRIBUTING.md` 要求 UI 变更提供前后截图或录屏，并没有为"只改行为、不加样式"设置例外。请附上受这次生命周期改动影响的 Desktop 与 TUI 状态。

最后一点在代码之外：`main` 已经前进，这个分支现在处于冲突状态。无论上述问题如何，它都需要先 rebase 才能合并；你推送之后我会针对新的 head 重新审阅。

</details>

Astro-Han:
After several exact-head review rounds, it is clear that the remaining findings share structural causes rather than isolated implementation mistakes. Settlement and recovery still classify the same admission differently, the durable Session execution contract is partly represented by memory-only state, and some tests bypass the new atomic admission boundary. The repeated fixes have also expanded this PR to 80 files and 33 commits, making both review and rebasing unnecessarily difficult.

I am pausing this implementation and will split it into three end-to-end changes:

1. A complete durable message lifecycle, including admission, mutation, settlement, recovery, and removal of the old Runtime queue authority.
2. Canonical CLI and Desktop conversation projection, preserving immediate send UX while reconciling by stable `messageId`.
3. Side Conversation admission and event ownership, built cleanly on the current Side Conversation architecture.

I will convert this PR back to draft and stop updating its implementation. Once the replacement PRs are available, I will close this PR and link them here. The existing review threads remain useful evidence for the replacement design, but should not be treated as another local patch queue.

This restructuring plan was developed with Codex assistance and reviewed by me.

Astro-Han:
The branch moved to `7128a97207` after my review, which was bound to `eecd4d93d`. I re-checked all three blocking findings against the new head rather than assuming either way, and **all three are unchanged**:

- `bindDurableSession()` still writes only the in-memory map, and `root-turn-coordinator.ts:2384` is still its only production caller, so the replay path still never sets the marker.
- `message-coordinator.ts:432-438` still classifies a pending admission from the root-turn receipt and the explicit Stop proof alone, while `:516` still consults the immutable steering proof. The asymmetry between the two paths is what the finding is about, and it is still there.
- `execution-composition.test.ts` still calls `commitMessageAdmission` and then appends the same `admitted-steering` identity through the direct path, so the exact-once assertion still reads three entries.

So the earlier review still applies in full; only its commit binding is stale. Nothing here asks you to re-explain the rebase — I mention it only so the moved head is not mistaken for a response to the findings.

<details>
<summary>简体中文</summary>

分支在我评审之后移到了 `7128a97207`，而那次评审绑定的是 `eecd4d93d`。我没有凭任何一边的假设下结论，而是对新 head 重新核了三条阻塞项，**三条都没有变化**：

- `bindDurableSession()` 仍然只写内存中的那个 map，`root-turn-coordinator.ts:2384` 仍是它在生产代码里唯一的调用者，因此重放路径依旧不会设置该标记。
- `message-coordinator.ts:432-438` 仍然只凭 root-turn 收据和显式 Stop 证据来判定未决 admission，而 `:516` 仍然会查 immutable steering proof。这条 finding 说的正是两条路径之间的不对称，它依然存在。
- `execution-composition.test.ts` 仍然在 `commitMessageAdmission` 之后，又用直接路径追加了同一个 `admitted-steering` identity，所以 exact-once 断言读到的仍然是三条。

因此之前那条评审整体仍然成立，过期的只是它的 commit 绑定。这里不需要你解释这次 rebase——我说明它，只是为了避免把 head 的移动误当成对上述问题的回应。

</details>

PR #3634, opened 2026-08-23
feat(runtime): represent parallel tool calls as a model capability

### LINKED ISSUE #3500: Represent parallel tool calls as a model capability
## Problem

This request was motivated by Codex CLI 0.149.0, which made parallel tool calls a default capability for regular model prompts again. Its implementation lets a model return several independent tool calls in one response and schedules them concurrently while preserving stable result ordering. In practice, this noticeably improves latency and the overall agent experience for tasks such as independent searches, inspections, and checks.

Maka currently normalizes tool support only as `functionCalling`. The runtime can execute multiple tool calls concurrently, but request behavior is inconsistent: the OpenAI Codex subscription path defaults `parallel_tool_calls` to `true`, while other provider adapters rely on SDK or provider defaults.

As model interfaces become more granular, tool-call support may not imply parallel tool-call support, and that distinction cannot currently be discovered or represented per model. The Codex CLI behavior also shows that this is useful as an explicit, default-on capability where the model and wire support it.

## Desired outcome

Maka can represent whether a model or resolved provider wire supports parallel tool calls and apply that capability consistently when constructing requests, while preserving the generic concurrent settlement path for models that return multiple calls. Supported models should be able to use it by default without requiring connection-specific configuration.

Today this relies on the OpenAI Codex special case, provider defaults, or a connection-level request body overlay. These do not provide a discovered, per-model capability.

1625567290:
## Summary

- add a per-model parallelToolCalls capability and preserve explicit true/false values through connection decoding and catalog updates
- resolve the capability against the effective provider adapter, defaulting supported native OpenAI Chat/Responses wires on while keeping compatible providers opt-in
- route the resolved flag through provider options and remove the Codex subscription fetch special case
- keep the existing generic concurrent tool settlement path unchanged

## Reproduction

On main cd0163a1e, a discovered model carrying parallelToolCalls is rejected as an unknown capability, and an observed native OpenAI Chat request omits parallel_tool_calls.

On this head:

- the codec preserves an explicit false value
- native OpenAI defaults the actual request body to parallel_tool_calls: true
- an explicit native false reaches the actual request body
- an unknown compatible provider still omits the field
- an explicitly capable compatible provider sends parallel_tool_calls: true

## Local validation

- npm run format:check
- npm run lint
- 20 focused core tests passed
- 74 focused runtime and request-wire tests passed
- npm run build
- npm run typecheck
- npm run check:asf-headers

A full npm test run also completed. It reproduced one storage cross-process authority failure and eight cancelled desktop OAuth timeout tests. Both results were independently reproduced in a clean worktree at main cd0163a1e.

## Limitation

The pinned @ai-sdk/open-responses 2.0.28 adapter does not expose or serialize a request-side parallel tool-call option, so that adapter remains unchanged. Native OpenAI Responses uses the supported OpenAI provider option.

Closes #3500

### DIFF
--- a/apps/desktop/src/renderer/settings/use-connection-detail.ts
+++ b/apps/desktop/src/renderer/settings/use-connection-detail.ts
@@ -826,6 +826,7 @@ function modelListsEqual(left: ModelInfo[], right: ModelInfo[]): boolean {
     if (leftModel.capabilities?.vision !== rightModel.capabilities?.vision) return false;
     if (leftModel.capabilities?.reasoning !== rightModel.capabilities?.reasoning) return false;
     if (leftModel.capabilities?.functionCalling !== rightModel.capabilities?.functionCalling) return false;
+    if (leftModel.capabilities?.parallelToolCalls !== rightModel.capabilities?.parallelToolCalls) return false;
     if (leftModel.capabilities?.imageGeneration !== rightModel.capabilities?.imageGeneration) return false;
   }
   return true;
--- a/packages/core/src/__tests__/model-catalog.test.ts
+++ b/packages/core/src/__tests__/model-catalog.test.ts
@@ -82,6 +82,24 @@ test('chat-default validation blocks image-only models but accepts merged partia
   assert.deepEqual(verdict(partial), { ok: true });
 });
 
+test('catalog entries preserve advertised parallel tool-call support', () => {
+  const [entry] = buildModelCatalogEntries({
+    providerType: 'openai-compatible',
+    defaultModel: 'relay-model',
+    models: [
+      {
+        id: 'relay-model',
+        capabilities: { functionCalling: true, parallelToolCalls: true },
+      },
+    ],
+    modelSource: 'fetched',
+  });
+  assert.deepEqual(entry?.capabilities, {
+    functionCalling: true,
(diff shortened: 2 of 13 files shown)


Astro-Han (approved):
GO at exact head `7c9124842ef1f81f0954e6cb487e7fcb8b086d0e`; no P0–P3 findings. Coverage: capability decode/catalog preservation (including explicit `false`), stored-model → metadata → native-wire default precedence, provider-option namespace/key selection for OpenAI Chat/Responses, Codex subscription, GitHub Copilot, and compatible Chat/Responses adapters, plus actual request-body projection. Duplicate-authority audit: the Codex fetch-body fallback is deleted; `resolveModelRuntime(...).parallelToolCalls` is the sole request-policy fact, while generic concurrent tool settlement remains a deliberately separate execution mechanism rather than a second capability decision. Exclusions: the pinned `@ai-sdk/open-responses` adapter (it has no request option), real provider calls, and UI copy/visuals. Hosted exact-head `test` is terminal green. Local Core → Storage → Runtime → Desktop builds passed and 94 focused catalog/codec/runtime/wire/subscription tests passed. The branch is 10 commits behind current `main`; a real one-commit rebase probe is clean.

PR #3635, opened 2026-08-23
fix(desktop): make queued messages editable

hqhq1025:
## Summary

- keep queued steering messages visible until the active turn actually consumes them, while filtering entries already in flight
- align the pending-message plate with Codex by exposing Edit, Steer, Delete, and follow-up reordering
- add an authoritative `queue.entry.update` Runtime Host command that preserves entry identity, order, and placement
- re-prepare edited content, relocate valid inline references, and reject stale edits with the queue revision

## Root cause

The desktop rendered only follow-up entries from `queue_update`, so queued steering messages disappeared before delivery. Editing was implemented as retracting an entry back into the composer, which changed the queue semantics and could not preserve its original position or placement.

This change keeps the Runtime Host projection as the source of truth and adds a dedicated compare-and-swap update operation.

## Verification

- `npm run build`
- `npm run typecheck`
- `npm run lint -- --diagnostic-level=error`
- `npm run format:check`
- Runtime Host protocol and message coordinator tests: 83 passed
- Runtime session manager tests: 223 passed
- Desktop queue/client/IPC tests: 37 passed
- Composer queue tests: 4 passed
- Storybook desktop and mobile visual checks for pending and editing states
- `npm run check:asf-headers` still reports the two pre-existing missing headers in `packages/cli/src/tui-session-status.ts` and `packages/cli/src/__tests__/tui-session-status.test.ts`

## Screenshots

Before:

Queued messages before

After, pending state:

Queued messages with Edit, Steer, and Delete controls

After, inline editing:

Queued message inline editing state

## AI use

Tool(s) and scope: Codex (GPT-5.6) implemented the Runtime Host, desktop bridge, UI, tests, and visual verification.

## Checklist

- [ ] No


Astro-Han (approved):
GO at exact head `b3086d6b898629635b6ea47eda35cf787d634b18`; no P0–P3 findings. Coverage: Runtime Host queue-update CAS/idempotency, in-flight and transition races, content re-preparation/reference relocation, projection and capacity preflights; protocol/client/preload/IPC boundaries; Desktop queue projection; UI edit/delete/promote/reorder behavior; and the embedded Runtime queue projection. Exclusions: copy/visual polish and a real Electron interaction run. The hosted exact-head `test` check is terminal green. Local Core → Storage → Runtime → Runtime Host → UI → Desktop builds passed, plus 347 focused tests across Runtime Host message/protocol, Runtime session queueing, UI composer, and Desktop queue/client/IPC. The branch is 10 commits behind current `main`; a real rebase probe conflicts only in `packages/runtime-host/src/protocol/index.ts`, an additive export collision. Preserve both current-main managed-update exports and this PR's message operation exports when rebasing, then rerun the exact-head gate.

PR #3638, opened 2026-08-23
Add parameters check for adding Runtime Host.

### LINKED ISSUE #3637: Add parameters check for adding Runtime Host.
### Problem

When we add a Runtime Host, we don't know whether we sign legal parameters for setting, and there's no a buttom like Test Connection or Save But Not Enable. Also it seems not be able to be updated? When saving, the app will test  connection and not meaningful reason for fail anymore. It is so confusing.

### Desired outcome

Yes, I can see the whether the parameters I've signed are legal. And it's so good if I can make testing connection and update a saved setting. Also it will be better if I get the detailed for why the connectiong is failed.

_No response_

sohardforaname:
## Summary

Fixes #3637 

## AI use

Select exactly one:

Tool(s) and scope:

## Checklist

- [ ] No


apps/desktop/src/main/runtime-host-profile-service.ts
```diff
@@ -550,10 +557,31 @@ export function createDesktopRuntimeHostProfileService(input: {
...
           : { kind: "connected", snapshot: await snapshot() };
       });
     },
+    testConnection(value) {
+      requireSaveInput(value);
+      return mutate(async () => {
+        if (value.credential === undefined) {
+          return { kind: "failed", stage: "credential" };
+        }
+        const target = { profile: value.profile, credential: value.credential } as const;
```
M4n5ter, line 644: <details open>
<summary>English</summary>

**[P1] Validate the complete profile before allowing a temporary plaintext connection.**

`requireSaveInput()` only checks that an id is a string and that the credential fits the byte cap, then this path forwards `value.profile` without `decodeRemoteRuntimeHostProfile()`. A renderer can therefore invoke the new IPC with `{ kind: "plaintext", url: "ws://<remote-host>/..." }` and no valid `acknowledgement`. The production connector branches only on `transport.kind` and sets `allowInsecureRemote: true`, so the bearer credential is sent even though saved profiles require the exact `plaintext-bearer-v1` acknowledgement.

I reproduced both halves: the service accepted the missing-ack profile and invoked its connector; the production connector then received `allowInsecureRemote=true`. The new tests replace the connector with a fake, so they cannot see this invariant bypass. Please decode/normalize the entire profile and validate the credential at this main-process trust boundary before invoking the temporary connector.

</details>

<details>
<summary>中文</summary>

**[P1] 临时明文连接前必须校验完整 profile。**

`requireSaveInput()` 只检查 id 是否为字符串及 credential 是否超过字节上限，随后这里没有调用 `decodeRemoteRuntimeHostProfile()` 就直接转发 `value.profile`。renderer 因此可以通过新 IPC 传入 `{ kind: "plaintext", url: "ws://<remote-host>/..." }`，同时缺少合法 `acknowledgement`。生产 connector 只按 `transport.kind` 分支并设置 `allowInsecureRemote: true`，所以即使保存 profile 时要求精确的 `plaintext-bearer-v1` acknowledgement，也会发送 bearer credential。

我分别复现了两段路径：service 接受缺少 acknowledgement 的 profile 并调用 connector；生产 connector 随后收到 `allowInsecureRemote=true`。新增测试用 fake 替换了 connector，因此结构上看不到这个约束绕过。请在 main-process 信任边界先完整 decode/normalize profile 并校验 credential，再调用临时 connector。

</details>

apps/desktop/src/renderer/settings/runtime-host-profiles-section.tsx
```diff
@@ -474,20 +538,30 @@ function createTransport(draft: ReturnType<typeof createRemoteHostDraft>): Runti
...
 }
 
-function draftComplete(draft: ReturnType<typeof createRemoteHostDraft>): boolean {
-  if ([draft.name, draft.rootId, draft.credential].some((value) => !value.trim())) return false;
+function validateDraft(
+  draft: RemoteHostDraft,
+  copy: ReturnType<typeof getSettingsProjectsCopy>["runtimeHost"],
+): Partial<Record<"name" | "url" | "destination" | "sshPort" | "remotePort" | "websocketPath" | "plaintextAcknowledged" | "rootId" | "credential", string>> {
+  const errors: Partial<Record<DraftField, string>> = {};
+  if (!draft.name.trim()) errors.name = copy.fieldRequired;
```
M4n5ter, line 554: <details open>
<summary>English</summary>

**[P2] Keep renderer-valid drafts identical to the authoritative save contract.**

This validator checks names only for non-whitespace, TLS/plaintext URLs only for non-emptiness, and credentials without the 8 KiB limit. The authoritative decoder additionally rejects control/over-128-byte names, wrong URL schemes, URL credentials/query/fragment, missing plaintext acknowledgement, and oversized credentials. I reproduced a 129-byte name returning `connected` from Test, then failing unchanged on Save with `Runtime Host profile name is invalid`.

That makes the new parameter check contradict the save path. Please establish one shared browser-safe structural validation authority (with localized presentation kept in the renderer) and add an invariant test that every renderer-valid draft is accepted by the main decoder.

</details>

<details>
<summary>中文</summary>

**[P2] renderer 判定有效的 draft 必须与权威保存契约一致。**

这里对 name 只检查非空白，对 TLS/plaintext URL 只检查非空，对 credential 也没有 8 KiB 上限；权威 decoder 还会拒绝控制字符/超过 128 字节的 name、错误 URL scheme、URL 中的 credential/query/fragment、缺少 plaintext acknowledgement，以及过大的 credential。我已复现：129 字节 name 的 Test 返回 `connected`，同一份未修改输入紧接着 Save 却报 `Runtime Host profile name is invalid`。

这会让新的参数检查与保存路径互相矛盾。请建立一份 browser-safe 的共享结构校验 authority（本地化展示仍留在 renderer），并增加 invariant test：任何 renderer-valid draft 都必须被 main decoder 接受。

</details>

apps/desktop/src/main/runtime-host-profile-service.ts
```diff
@@ -550,10 +557,31 @@ export function createDesktopRuntimeHostProfileService(input: {
...
+              snapshot: await snapshot(),
+              message: error.message,
+              stage: classifyConnectionTestFailure(error),
+            }
           : { kind: "connected", snapshot: await snapshot() };
       });
     },
+    testConnection(value) {
+      requireSaveInput(value);
+      return mutate(async () => {
```
M4n5ter, line 640: <details open>
<summary>English</summary>

**[P2] Do not put a state-free connection test on the durable profile mutation queue.**

`mutate()` is the shared `mutationTail` for snapshots and all profile changes. This temporary test can occupy it for the 45-second readiness deadline, or up to the 120-second interactive SSH startup window. Existing-profile controls do not consult `testing`, so default/enable/remove actions can still be clicked and then appear stuck; subscription-driven `getSnapshot()` also waits. A deferred-connector probe confirmed that `getSnapshot()` did not settle until the test was released.

The added no-persistence test calls `getSnapshot()` only after the test finishes, so it cannot catch this. Please run the unsaved test outside the durable mutation queue; keep terminal exclusivity in the SSH terminal owner and use operation-scoped cancellation if needed.

</details>

<details>
<summary>中文</summary>

**[P2] 无状态连接测试不应占用持久 profile mutation queue。**

`mutate()` 是 snapshot 与所有 profile 变更共享的 `mutationTail`。这个临时测试可能占用它直到 45 秒 readiness deadline，交互式 SSH startup 最长可到 120 秒。现有 profile 控件并不读取 `testing`，所以 default/enable/remove 仍可点击，随后看起来卡住；订阅触发的 `getSnapshot()` 也会等待。定向 deferred-connector probe 已确认：测试释放前，`getSnapshot()` 不会完成。

新增的“不持久化”测试在测试结束后才调用 `getSnapshot()`，因此发现不了这个问题。请让 unsaved test 脱离持久 mutation queue；SSH terminal 的互斥仍由 terminal owner 保持，如需取消则使用 operation-scoped cancellation。

</details>

apps/desktop/src/renderer/settings/runtime-host-profiles-section.tsx
```diff
@@ -140,16 +190,22 @@ export function RuntimeHostProfilesSection(props: {
...
       if (result.kind === "unavailable") {
+        setShowAdd(false);
+        setDraft(createRemoteHostDraft());
+        setTestResult(undefined);
+        setTestInvalidated(false);
         toast.error(
-          copy.selectFailed,
-          result.message,
+          copy.savedUnavailable,
+          copy.testFailed[result.stage],
```
M4n5ter, line 209: <details open>
<summary>English</summary>

**[P2] Preserve the actionable saved-connection failure instead of replacing it with an incomplete string-derived stage.**

The new classifier parses human-readable messages, and common failures such as TLS verification, readiness timeout/draining, and SSH local-bind conflict all fall through to `connection`. This save path still receives the exact `result.message`, but now discards it and shows only the generic localized text (“check the remote Host and try again”). That regresses the previously actionable failure and works against the stated stage-specific diagnostics.

Only the credential phrase is tested. Please project a typed reason/stage where the connector already knows the failure, retain the original cause/message, and show the safe concrete message as detail or fallback for saved-unavailable results. This also removes the new hidden cross-package message-string contract.

</details>

<details>
<summary>中文</summary>

**[P2] 不要用不完整的字符串 stage 取代保存失败时已有的可操作诊断。**

新 classifier 解析人类可读 message；TLS 验证失败、readiness timeout/draining、SSH 本地端口冲突等常见失败都会落到通用 `connection`。保存路径仍拿到了精确的 `result.message`，但现在把它丢弃，只显示通用本地化文案（“检查远程 Host 后重试”）。这让原本可操作的失败信息回退，也违背了 stage-specific diagnostics 的目标。

现有测试只覆盖 credential 文案。请在 connector 已知 failure 的位置投影 typed reason/stage，保留原 cause/message，并把安全的具体 message 作为 saved-unavailable 的详情或 fallback。这也能删除新引入的跨包 message-string 隐式契约。

</details>

M4n5ter (commented):
<details open>
<summary>English</summary>

Reviewed exact head `22993f4ca7939611cc787e1992eebbda1445494c` with independent correctness and bounded simplification lanes.

Verdict: **NO-GO — 1×P1 + 3×P2**. The inline findings cover: plaintext test connections bypassing the required acknowledgement; renderer validation accepting drafts the authoritative save decoder rejects; a temporary test monopolizing the profile mutation queue; and stage projection discarding actionable saved-connection diagnostics.

Why the added tests remain green: both new tests inject a fake connector, so they cannot observe the production plaintext-transport invariant; the no-persistence test waits for testing to finish before reading the snapshot, so it cannot observe queue blocking; classification covers only the credential phrase; no renderer validation path is tested.

Simplification verdict: this feature adds a legitimate user path, but the current implementation also adds two conflicting authorities—the partial renderer validator and a human-message parser for failure stages. The smallest durable design is one structural profile-validation authority across UX and the main trust boundary, plus one typed client-owned failure fact rather than a second string contract.

Validation completed: Desktop workspace dependency build; all four Desktop typechecks; main, preload, and renderer production builds; renderer notice verification; 23/23 focused profile-service tests; Biome on all seven changed files; and diff check. Targeted probes reproduced the plaintext acknowledgement bypass, test/save validation contradiction, and mutation-queue blocking.

Gate status is separate from the findings: this PR is still draft and dirty/conflicting. GitHub reports zero exact-head checks and zero workflow runs, and a local merge-tree reproduces a current-main conflict in `apps/desktop/src/main/__tests__/runtime-host-profile-service.test.ts`; therefore merge-result gate evidence is unavailable, not green. Current `main` is green. Review threads, prior reviews, and stale approvals are all empty.

</details>

<details>
<summary>中文</summary>

已在 exact head `22993f4ca7939611cc787e1992eebbda1445494c` 上并行完成独立 correctness 审查与 bounded simplify audit。

结论：**NO-GO — 1×P1 + 3×P2**。行内 finding 分别是：明文临时测试绕过必需的 acknowledgement；renderer 校验接受而权威保存 decoder 拒绝的 draft；临时测试独占 profile mutation queue；stage 投影丢失保存失败时原有的可操作诊断。

新增测试仍然全绿的原因：两个测试都注入 fake connector，因此结构上看不到生产 plaintext transport 的安全约束；“不持久化”测试等测试结束后才读取 snapshot，因此看不到队列阻塞；错误分类只覆盖 credential 文案；renderer 校验路径没有测试。

简化结论：这个功能增加了真实用户能力，但当前实现同时新增了两份会漂移的 authority——不完整的 renderer validator，以及从人类可读 message 推导 stage 的 parser。更小且耐久的模型是：UX 与 main 信任边界共享一份结构校验规则/代码 authority；连接层只产生一份 typed failure fact，而不是再维护字符串契约。

已完成验证：Desktop workspace 依赖构建；四套 Desktop typecheck；main、preload、renderer 生产构建；renderer notice 校验；23/23 focused profile-service tests；7 个改动文件的 Biome；diff check。定向 probe 已复现 plaintext acknowledgement 绕过、测试成功但保存失败的校验矛盾，以及 mutation queue 阻塞。

门禁状态与 findings 分开：PR 仍为 draft 且 dirty/conflicting。GitHub 在 exact head 上是 0 checks、0 workflow runs；本地 merge-tree 也复现了与 current main 在 `apps/desktop/src/main/__tests__/runtime-host-profile-service.test.ts` 的冲突，因此 merge-result 门禁证据是“无法取得”，不是 green。当前 `main` 为 green。review threads、既有 reviews 和陈旧批准均为空。

</details>

Issue #2624, opened 2026-08-10
ci(windows): eliminate hidden failures in the non-blocking baseline


M4n5ter:
<details open>
<summary><strong>English</strong></summary>

## Problem

The Windows baseline currently reports an overall successful workflow while individual `continue-on-error` steps emit failure annotations. This makes the PR check green without giving us a clean Windows baseline.

PR #2613 exposed the current set in [run 31323029613](https://github.com/maka-agent/maka-agent/actions/runs/31323029613). The latest comparable `main` run, [31322689039](https://github.com/maka-agent/maka-agent/actions/runs/31322689039), reproduces the deterministic failures.

### 1. Windows skip inventory is stale

Both the PR and `main` report:

```text
Windows test skip inventory is stale; run `npm run windows:inventory:write`
```

The generated inventory must be refreshed and reviewed against the current tree.

### 2. Runtime PTY cleanup leaves SQLite locked

Both the PR and `main` run the two selected PTY tests successfully, then fail suite cleanup:

```text
EBUSY: resource busy or locked, unlink '...\maka-shell-run-...\runtime.sqlite'
```

The test or production owner still holds the SQLite file when the temporary root is removed. Windows does not permit unlinking an open database.

### 3. Root-authority initialization race fixture resolves too early

Both the PR and `main` fail:

```text
rejects replacement before opening the temporary marker
initialization race fixture reported resolved while waiting for marker_open_pending
```

This is a regression in an area previously tracked as completed by #2438. The fixture must preserve the intended cut on Windows instead of resolving before the marker-open observation.

### 4. Intermittent SQLite file-lock failures

The latest PR run also observed two failures that did not reproduce in the prior PR run or the compared `main` run:

```text
EBUSY: resource busy or locked, unlink '...\workspaces\default\memory.sqlite'
EPERM: operation not permitted, open '...\memory.sqlite-shm'
```

The first occurred during Desktop real-window smoke cleanup. The second occurred in the concurrent Long-term Memory CAS test. Their one-run appearance suggests timing-sensitive ownership or cleanup rather than a deterministic semantic failure, but they still need repeated evidence before being classified as flakes.

## Why this matters

The lane is intentionally non-blocking, but a successful outer job currently hides real failing steps. This prevents the baseline from showing whether Windows lifecycle correctness is improving or regressing.

The goal is not to mask Windows file locking with broad retry loops. Every Store, Host, and fixture should release its exact owner before temporary-root cleanup.

## Desired outcome

[...]

PR #3642, opened 2026-08-23
test(desktop): match session change event contract

Astro-Han:
## Summary

Align the Goal adapter test fixture with the current `SessionChangedEvent` contract so Desktop main-process compilation succeeds on `main` again.

The production contract and all live emitters use `reason`, optional identifiers, and `ts`; only the new test fixture from #3531 retained the removed `type: 'sessions_changed'` discriminator. This removes those three stale properties without changing production behavior or adding another event representation.

## Verification

- Reproduced on `main` with `npm --workspace @maka/desktop run build:main`: three TS2353 failures in `goal-services-adapter.test.ts`.
- `npm --workspace @maka/desktop run build:main`
- `node --test --test-concurrency=1 apps/desktop/dist/main/__tests__/goal-services-adapter.test.js` (1/1 passed)
- `npm run format:check`
- `git diff --check`

## AI use

Select exactly one:

- [ ] No generative tool made a substantive contribution
- [x] Generative tooling made a substantive contribution

Tool(s) and scope: Codex diagnosed the CI failure, verified the event authority, removed the stale test-only fields, and ran focused validation.

## Checklist

- [x] Tests cover the change and fail without it
- [x] Lint, format, typecheck and the affected suites pass locally

Does this PR entail a change in behavior?

- [ ] Yes — described under Summary above
- [x] No

PR #3644, opened 2026-08-23
test(runtime): close shell-run stores before cleanup

kyletser:
## Summary

Track every SQLite ShellRun store opened by the shell-run-manager suite and close those leases before removing the temporary workspaces. This removes the deterministic Windows `EBUSY` cleanup failure without retrying or masking deletion while a live owner remains.

Refs #2624

## Verification

- `npm --workspace @maka/runtime run typecheck` — passed.
- `npm --workspace @maka/runtime run build` — passed.
- Windows PowerShell UTF-8 gate from `windows-baseline.yml` — 2 passed, 0 failed, and exited normally.
- Mutation check: temporarily removed the store `close()` loop; both selected assertions still passed, then the top-level hook failed with `EBUSY ... runtime.sqlite` and exit code 1. Restoring the loop returned the gate to exit code 0.
- `npm run lint` and `npm run format:check` — passed.
- `npx knip --workspace apps/desktop` and `npx knip --workspace packages/ui` — passed (desktop reported two existing configuration hints).
- Full shell-run-manager file on Windows — 42 passed, 9 failed, 8 skipped. The remaining failures are the other Windows PTY behavior slices already described in #2624; the `runtime.sqlite` cleanup hook failure is absent.
- Root `npm run build` / `npm run typecheck` reach Desktop and then fail on the three pre-existing `SessionChangedEvent.type` errors in `goal-services-adapter.test.ts`. Current `main` fails at the same Build step: https://github.com/apache/maka/actions/runs/32652983938

## AI use

Select exactly one:

Tool(s) and scope: Codex diagnosed the ownership boundary, implemented the test-lifecycle change, ran the verification and mutation checks, and drafted this pull request description. The commit includes the required `Generated-by: Codex` trailer.

## Checklist

The affected Runtime checks, lint, format, and selected Windows gate pass; the unchecked item reflects the current unrelated Desktop build/typecheck baseline failure documented above.

- [x] No

### DIFF
--- a/packages/runtime/src/__tests__/shell-run-manager.test.ts
+++ b/packages/runtime/src/__tests__/shell-run-manager.test.ts
@@ -48,9 +48,11 @@ import { PTY_PROTOCOL_REPLY_MAX_BYTES } from '../pty-screen-collector.js';
 
 const NO_ABORT = new AbortController().signal;
 const TEMPORARY_WORKSPACES = new Set<string>();
+const SQLITE_SHELL_RUN_STORES = new Set<ReturnType<typeof createSqliteShellRunStore>>();
 const REAL_WINDOWS_GIT_BASH = windowsGitBashPlan();
 
 after(async () => {
+  for (const store of SQLITE_SHELL_RUN_STORES) store.close();
   await Promise.all(
     [...TEMPORARY_WORKSPACES].map((path) => rm(path, { recursive: true, force: true })),
   );
@@ -59,7 +61,7 @@ after(async () => {
 describe('ShellRunProcessManager', () => {
   test('rejects unprojectable provider tool-call identities before durable admission', async () => {
     const cwd = await workspace();
-    const store = createSqliteShellRunStore(cwd);
+    const store = sqliteShellRunStore(cwd);
     const manager = createManager(store);
     const completions: boolean[] = [];
     const maximumMultibyteId = '😀'.repeat(SHELL_RUN_SOURCE_TOOL_CALL_ID_MAX_BYTES / 4);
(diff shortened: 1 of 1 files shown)


kyletser:
Rebased this PR onto current main at a8661ab, which includes #3642 and removes the three pre-existing SessionChangedEvent.type build errors from the initial run. The new workflow run is currently action_required, so it needs maintainer approval before CI can execute. The branch still contains only the original targeted runtime test cleanup commit.

Astro-Han (commented):
Reviewed at exact head `a8661abd`. **No findings — no [P0]–[P3].** Publishing on behalf of a reviewer without write access here; I re-verified the scope and gate state before posting.

The diff is confined to the runtime shell-run test lifecycle. The wrapper tracks every `ClosableShellRunStore` it creates, and the top-level `after` closes each store before removing the directory — which is the right order, and the whole point of the change. Nothing in the production path moves.

Independent validation at this head: focused build plus the shell-run-manager suite at 58 passed / 4 skipped / 0 failed, with lint, format, and `git diff --check` all clean.

**Not approving yet, and the reason is procedural rather than substantive.** When this was reviewed the exact-head checks had never executed — the run was parked at `action_required`, which is the state fork PRs sit in until someone releases them. I have since released it and `test` is now running. Never-run is not the same as green, so the approval has to wait for a terminal result on this exact head. Assuming it lands green and the head does not move, this is ready.

Astro-Han (approved):
Gate now satisfied at the same head `a8661abd`. The run I released has landed: `test` is terminal `completed/success`, which is the only check this path set triggers. The head has not moved since the review above, there are no unresolved threads, and no [P0]–[P3] were found.

Approving.

Astro-Han:
LGTM — merged. Thanks for the contribution!

<details>
<summary>中文</summary>

已合并，感谢贡献。
</details>

PR #3652, opened 2026-08-23
fix(runtime): preserve Edit diffs in large files

Astro-Han:
## Summary

Preserve unified diff previews when Edit makes a small localized replacement in a large file. The runtime now uses the match span already produced by Edit, so diff cost depends on the changed window instead of the full file. Oversized replacements remain bounded and fall back to the existing summary without blocking the edit.

Also removes the unused UI diff-line classifier and barrel export; DiffCodePreview is the remaining rendering authority.

The branch temporarily carries the one-line canonical storage decoder repair already present in #3633 because current main does not build without it. That prerequisite disappears from this diff when #3633 lands.

## Verification

- Runtime build passed.
- Runtime and UI typechecks passed.
- Biome format check passed across 1604 files.
- 101 focused runtime tests passed, including red-to-green coverage for large-file Edit diffs and oversized replacement fallback.

## AI use

Select exactly one:

Tool(s) and scope: Codex diagnosed the diff cutoff, implemented the localized bounded diff path, removed the proven-dead classifier, and added focused tests.

## Checklist

- [ ] No

### DIFF
--- a/packages/runtime/src/__tests__/builtin-tools.test.ts
+++ b/packages/runtime/src/__tests__/builtin-tools.test.ts
@@ -2186,6 +2186,24 @@ describe('builtin write tools path containment', () => {
     expect(await readFile(join(root, 'inside.txt'), 'utf8')).toBe('hello Maka');
   });
 
+  test('Edit returns a file diff for a localized change in a large file', async () => {
+    const root = await mkdtemp(join(tmpdir(), 'maka-edit-large-diff-'));
+    const content = Array.from({ length: 900 }, (_, index) => `const v${index} = ${index};`).join(
+      '\n',
+    );
+    await writeFile(join(root, 'large.ts'), `${content}\n`, 'utf8');
+
+    const result = await runTool(
+      tool('Edit'),
+      { path: 'large.ts', old_string: 'const v500 = 500;', new_string: 'const v500 = -1;' },
+      root,
+    );
+
+    expect(result).toMatchObject({ kind: 'file_diff' });
+    assert.match((result as { diff: string }).diff, /-const v500 = 500;/);
+    assert.match((result as { diff: string }).diff, /\+const v500 = -1;/);
+  });
+
   test('file tools stay usable when the session cwd is reached through a symlink', async () => {
     // The session cwd itself sits under a symlink (macOS hands out `/var/...`
     // tmpdirs whose realpath is `/private/var/...`, and workspace roots are often
--- a/packages/runtime/src/__tests__/filesystem-worker.test.ts
+++ b/packages/runtime/src/__tests__/filesystem-worker.test.ts
@@ -477,7 +477,7 @@ describe('filesystem worker operations', () => {
     assert.equal(await readFile(target, 'utf8'), 'external');
   });
 
-  test('omits the diff when the content is too large to diff cheaply', async () => {
+  test('returns a localized diff for a small Edit in a large file', async () => {
     const root = await temporaryDirectory('maka-worker-huge-');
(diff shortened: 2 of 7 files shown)


yunaremaia:
Reviewed the diff on `2fe6c14`.

The problem statement is clear: `createUnifiedDiff` produced a full-file diff for small localized edits in large files, which was either suppressed entirely or returned an unwieldy wall of context. The fix introduces `createEditUnifiedDiff` and narrows the diff to the localized hunks — the right abstraction at the right layer.

What I checked:

- `filesystem-executor.ts` now captures `originalContent` inside the `transform` closure of the `readModifyWrite` path, and threads it into `createEditUnifiedDiff`. This is the correct place: the closure is the only point where both pre-edit and post-edit content are available in scope, and `workspace.readModifyWrite` does not expose them after the fact (the `result` only carries `previous` / `finalContent`, which are post-traversal). The pattern avoids leaking capture by the scope of the closure itself — clean.

- `filesystem-worker/operations.ts` mirrors the same change on the worker path, where the file is read from a `FileHandle` rather than a `readModifyWrite` contract. Same direction, same fix. DRY is preserved by sharing `createEditUnifiedDiff`.

- The test matrix covers three cases: (1) localized change in 900-line file returns `file_diff` with the targeted hunk — happy path; (2) same case on the worker path with `readFile` confirmation; (3) oversized *replacement* (900-line expansion) still omits the diff to avoid token blow-up. This last case is the one I'd expect to regress later if someone does not read the diff-limit guard — the assertion on `diff === undefined` + `readFile` confirming the replacement landed is a good pin.

- One thing to check: the `computeEditedSource` return includes `edited.content` (post-edit) and `edited.oldStart`/`edited.oldEnd`/`edited.newStart`/`edited.newEnd` (hunk boundaries). If `createEditUnifiedDiff` reads those hunks for context and those values are 0-indexed vs 1-indexed, the diff could be off-by-one in edge cases. The diff itself is not in this PR's changed lines (`packages/runtime/src/unified-diff.ts`), so I cannot verify the boundary arithmetic here — but the test diff matching (`/-const v500 = 500;/` and `/\+const v500 = -1;/`) confirms it is correct for at least this case. Worth a pinning assertion on the actual line numbers in the diff header if the codebase does not have one elsewhere.

- The diff-limit guard (`MAX_DIFF_CHARS` or equivalent) is the right backstop — if the replacement is large enough that the localized diff still exceeds the budget, the old behavior (suppress the diff) is preserved. That avoids a token-explosion regression when the model writes back a large file.

Clean, focused change. LGTM.

jackwener (approved):
GO on exact head `208590409572a0bfd086108e01ca51727ad8f235`; 0×P0–P3.

The Edit path now uses its already-authoritative match span to bound diff work to the changed window while retaining 3 lines of context. Window line/byte caps and a final output-byte cap make oversized replacements fall back to the existing structured Edit summary without blocking the write. A 5,000-case randomized differential application covered first/middle/last positions plus multiline insert/delete shapes; every emitted patch reconstructed the exact new content. Focused Runtime evidence passed 82/82, and reverting the three new call sites to the old full-file generator made exactly the two large-file regressions fail. Runtime/UI builds and changed-file Biome passed. `diffLineKind` has no remaining repository consumer, so removing its implementation and barrel export leaves `DiffCodePreview` as the single classifier.

Fresh gate: PR OPEN, non-draft, MERGEABLE; exact-head `test` and `windows_sandbox_w0_protocol` are completed/success; zero stale or current APPROVED reviews existed before this action; no live P0–P2 or hold. Published by the Kabi review lane under @WAWQAQ standing authorization. No merge performed.

PR #3657, opened 2026-08-24
Refactor/external session source parsing

### LINKED ISSUE #3625: refactor(storage): consolidate external session source parsing
## Problem

Maka has two user-facing paths over the same Codex and Claude session stores:

- `ForeignSession` creates a filtered, untrusted handoff for the current task.
- `ExternalSession` imports a full typed transcript as a new Maka Session.

Their outputs should remain different, but each path currently implements its own discovery, source identity, title and cwd selection, bounded reading, and source-specific parsing. The implementations have already drifted: Codex paths use different presentation records, while Claude handoff does not share the lineage resolution used by full import.

## Desired outcome

Use one package-private source catalog and native parser per external source, followed by two explicit projections: safe handoff and full Session import.

The consolidation should:

- select the same source ID, transcript candidate, title, and cwd under the same query policy;
- use one Codex record authority and one Claude lineage resolver before projection;
- preserve the handoff security boundary, redaction, and hostile-data envelope;
- preserve typed full history, `externalOrigin`, atomic persistence, and replay;
- cover Codex v0.144 and v0.149, Claude rewind and compaction, duplicate IDs, malformed records, and tool call/result pairing with shared fixtures; and
- strictly reduce production code and duplicated tests without adding another facade, compatibility bridge, or generic `ExternalMessage` model.

Coordinate with #1512 and #3520. Preserve #1512's staleness and hostile-path requirements, and reuse #3520's v0.149 fixture and record-selection fix.

Keeping the parsers independent preserves the current duplication and allows source interpretation to drift further. Deriving safe handoff from imported `StoredMessage[]` would also weaken the explicit trust boundary, so the two projections should remain separate after shared source resolution.

AI-assisted analysis: Codex agents traced both consumers, their security boundaries, related pull requests, and checked-in fixtures. A human contributor must review the final design and implementation.

CxHsin:
## Summary

Fixes #3625

Consolidate Codex and Claude external-session handling around shared catalog boundaries and native source adapters, while keeping safe handoff and full Session import as separate projections.

- Share source identity, transcript selection, metadata, query filtering, path validation, and bounded reads.
- Preserve Codex `event_msg` presentation authority and `response_item` tool authority.
- Support Codex v0.149 `event_msg.item_completed` records.
- Share Claude lineage handling for rewinds, compaction boundaries, duplicate IDs, and malformed records.
- Preserve handoff redaction, hostile-data envelopes, tool-output exclusion, staleness checks, and bounded tails.
- Preserve typed import history, tool call/result pairing, `externalOrigin`, atomic persistence, and replay.
- Add regression coverage for the real Codex `ExternalSessionImporter` route.

The handoff is not derived from imported `StoredMessage[]`, and no generic `ExternalMessage` model or additional compatibility facade is introduced.

## Verification

- Storage external-session tests: 23/23 passed
- Runtime-host external-session coordinator tests: 9/9 passed
- External-origin composition regression test passed
- Biome checks passed for all affected files
- `git diff --check` passed

The complete storage suite still has unrelated Windows-environment failures involving symlink permissions and `fsutil file queryStreams`. All external-session tests pass.

## AI use

Tool(s) and scope:

Codex contributed materially to the source catalog, Codex and Claude adapters, import plumbing, fixtures, regression tests, and verification. The commit includes the required `Generated-by: Codex` trailer.

## Checklist

- [x] Tests cover the change and regression paths
- [x] Lint, format, typecheck/build, and affected suites pass locally

- [x] Yes — it adds Claude full-session import support, fixes Codex import compatibility with current session formats, and records external import origin.
- [ ] No


packages/storage/src/codex-session-adapter.ts
```diff
@@ -125,32 +161,65 @@ export class CodexSessionAdapter implements ExternalSessionAdapter {
...
-      if (rows === undefined) continue;
-      for (const row of rows) {
-        const entry = await this.entryFromRow(row);
-        if (entry?.id === sessionId) return entry;
-      }
-      break;
-    }
-
-    return this.findRolloutEntry(sessionId);
+    return (await this.listCatalog({ includeArchived: true })).find(
```
jackwener: [P1] Keep exact imports and catalog scans bounded before materializing the external DB. `findCatalogEntry()` now calls `listCatalog()` for the whole authoritative `threads` table, while `readCodexThreadRows()` uses unbounded `.all()` and `listCatalog()` starts one `realpath`/`stat` promise per row before applying `limit`. A real SQLite probe with 50,000 invalid rollout rows and `limit: 50` returned zero entries but still added 512 MiB RSS in 747 ms; larger/long-field stores can exhaust the Host. The replaced foreign scanner capped SQL at `FOREIGN_SESSION_SCAN_MAX_SESSIONS * 2` (100), and the previous import lookup queried `id = ?`. Restore an exact-id query for import and use bounded/paged catalog reads (continuing pages only as needed for post-filter matches), rather than loading the full external store.

packages/storage/src/foreign-session-store.ts
```diff
@@ -116,596 +82,91 @@ export function createForeignSessionStore(
...
+    };
     const results: ForeignSessionSummary[] = [];
-    if (sources.includes('claude-code')) {
-      results.push(...(await this.listClaudeSessions(options, now)));
-    }
-    if (sources.includes('codex')) {
-      results.push(...(await this.listCodexSessions(options, now)));
+    for (const reader of this.readers) {
+      if (!sourceEnabled(reader.id, this.env) || !(await reader.detect())) continue;
+      const entries = await reader.listCatalogEntries(query);
```
jackwener, line 127: [P2] Preserve the old required-`cwd` gate when routing Claude handoff through the shared adapter. `ClaudeCodeSessionAdapter.#parse()` initializes `cwd` to `''` and returns any nonempty record set; `#summaryOf()` then lists it and `readSession()` returns metadata with the empty cwd. With a one-line user transcript that omits `cwd`, the current path produced a visible summary with `cwd: ""` and an importable two-message session, whereas the removed scanner explicitly dropped `meta.cwd === undefined`. The Host import target does not override cwd, so this dirty value reaches `createImportedSession()` and persists an unusable/wrongly scoped Session. Reject transcripts whose source identity lacks a nonempty string cwd (and add missing/wrong-type regressions) before projecting either catalog or import output.

jackwener (commented):
NO-GO on exact head `a9508a6d193a9007a6d320b3a57301f8e7321721`: 1×P1 + 1×P2. The source-specific parsers preserve malformed-line/torn-tail handling and per-file byte caps, but this refactor loses a collection-size boundary and admits Claude records missing their working-directory identity. Focused Storage adapter/store/importer tests pass; the full Storage run had one unrelated cross-process authority failure caused by Node's SQLite experimental warning.

packages/storage/src/claude-code-session-adapter.ts
```diff
@@ -189,10 +218,111 @@ export class ClaudeCodeSessionAdapter implements ExternalSessionAdapter {
     };
   }
 
+  async listCatalogEntries(
+    query: ExternalSourceCatalogQuery = {},
+  ): Promise<readonly ClaudeCodeCatalogEntry[]> {
+    if (query.limit !== undefined && query.limit <= 0) return [];
+    const entries: ClaudeCodeCatalogEntry[] = [];
+    for (const file of await this.#transcriptFiles()) {
```
Astro-Han: [P1 · category ①] The result limit is still applied only after the source has been fully scanned. This loop opens and fully parses every Claude transcript before `matchesSourceCatalogQuery()` or `slice(limit)`; the Codex DB path likewise keeps paging through all invalid/orphan rows, and rollout fallback reads every candidate. The production Host passes no limit at all and slices 16 only after `adapter.listSessions()` returns. This regresses the old pre-read 30-day/50-candidate boundary: opening Settings catalog or the TUI picker with many stale or 64 MiB transcripts can sustain unbounded I/O, realpath/stat work, JSON parsing, and result allocation. Please put a total limit/cursor and age cutoff in the source contract, make Host pass its page bound, and stop before opening candidates that cannot enter the requested page.

packages/storage/src/claude-code-session-adapter.ts
```diff
@@ -189,10 +218,111 @@ export class ClaudeCodeSessionAdapter implements ExternalSessionAdapter {
...
+        transcriptPath: file.path,
+      };
+      if (matchesSourceCatalogQuery(entry, query)) entries.push(entry);
+    }
+    entries.sort(
+      (left, right) =>
+        right.updatedAtMs - left.updatedAtMs ||
+        left.transcriptPath.localeCompare(right.transcriptPath),
+    );
+    return query.limit === undefined ? entries : entries.slice(0, query.limit);
```
Astro-Han: [P2 · category ①] This new catalog listing path never performs the live-set pruning that `listSessions()` applies to `#summaries`. `FileForeignSessionStore` retains this adapter for the TUI lifetime and now calls only `listCatalogEntries()`, while `#summaryOf()` also caches sidechain/unreadable entries as undefined. Repeatedly creating and deleting or renaming transcripts therefore leaves one path/summary record per old file for the lifetime of the process. Please share the live-entry bookkeeping and cache pruning between both listing methods, and add a create/delete regression through `listCatalogEntries()`.

Astro-Han (commented):
Thanks for addressing the earlier import identity issue and for moving the Codex database read to bounded pages. I reviewed exact head `8c902ba1fb67f1893e0088fcc322267735e89735`.

The previous Claude missing-`cwd` P2 is closed, and exact Codex import now uses an ID query. The earlier collection-bound P1 is only partially closed, however: page size bounds each SQLite call, but the production catalog still performs an unbounded total source scan before the Host slices 16 results. I also found a separate P2 in the long-lived Claude summary cache. Both are called out inline.

Please restore a source-owned total candidate/time boundary, carry the Host page requirement into the adapter contract, and share the Claude live-cache pruning between both listing paths. A Host-to-real-adapter regression with many stale/invalid entries and a repeated create/delete cache regression would cover the missing production composition.

Exact-head `test`, `audit`, and `package` checks are successful. The PR is currently conflicting with `main`, so the final fixes also need current-base checks.

Review analysis was assisted by Codex and independent `@reviewer` agents. Astro-Han verified the exact-head diff, old feedback ledger, production Host/TUI catalog paths, resource bounds, cache lifetime, and severity before publication and owns this review.

<details>
<summary>中文对照</summary>

谢谢你修复了旧的 import identity 问题，并把 Codex DB 读取改成分页。当前 head 上，Claude 缺失 `cwd` 的 P2 已关闭，Codex 精确 import 也改为按 ID 查询。

但旧的 collection-bound P1 只部分关闭：page size 只限制单次 SQLite 调用，生产 catalog 在 Host 最后切 16 条之前仍会无界扫描整个 source。另有一个独立 P2：长期存活的 Claude summary cache 在新 catalog 路径不会清理。两项已放在行内。

请恢复 source owner 的总候选/时间边界，把 Host 分页需求带入 adapter contract，并让两条 Claude listing 路径共用 live-cache prune。当前三项 CI 成功，但 PR 与 `main` 冲突，修复后仍需 current-base checks。

本次审查分析由 Codex 和独立的 `@reviewer` 子代理协助；Astro-Han 核验了精确 head、旧反馈、生产 Host/TUI catalog 路径、资源边界、缓存生命周期和问题分级，并对本次 Review 负责。

</details>

packages/storage/src/claude-code-session-adapter.ts
```diff
@@ -305,14 +443,15 @@ export class ClaudeCodeSessionAdapter implements ExternalSessionAdapter {
...
+    if (records.length === 0 || cwd.trim().length === 0) return undefined;
     return {
       records,
       cwd,
+      ...(gitBranch !== undefined ? { gitBranch } : {}),
       title: pickClaudeTitle(titles),
       ...(createdAt !== undefined ? { createdAt } : {}),
       ...(updatedAt !== undefined ? { updatedAt } : {}),
-      isSidechain,
+      isSidechain: hasConversationRecord && allConversationRecordsAreSidechain,
```
Astro-Han, line 504: [P2 · category ①] This new file-level rule admits a transcript when it contains any main conversation record, but `readSession()` then passes every parsed record to `convertTranscript()`. `resolveTranscriptLineage()` uses non-sidechain records only to build lineage and returns the original mixed record set when no prompt was withdrawn, so interleaved sidechain user/assistant/tool records are persisted into the imported main Session. The new digest path explicitly filters these records, which makes digest and full import disagree. Please filter `isSidechain === true` at the shared lineage/import boundary and add a mixed-sidechain `readSession()` regression, not only the existing digest test.

Astro-Han (commented):
One additional exact-head issue emerged from the full-import lane after the catalog review was submitted. The digest path correctly excludes interleaved Claude sidechain records, but the canonical Session import path does not. I have called out the introduced mismatch inline.

Review analysis was assisted by Codex and independent `@reviewer` agents. Astro-Han verified the mixed-transcript production path and severity before publication and owns this follow-up.

CxHsin:
@Astro-Han The conflict with the latest `main` has been resolved and pushed as merge commit `a80b114733769865681643a536fd664fcca9b333`.

Conflicts were in:
- `packages/storage/src/codex-session-adapter.ts`
- `packages/storage/src/execution-stores.ts`

Resolution and review:
- Preserved the PR's unified Codex presentation parsing, including `item_completed` handling and response-mirror de-duplication.
- Incorporated `main`'s RuntimeInvocation imports and Responses phase `providerOptions` propagation.
- Restored source-owned bounded catalog queries: `limit`/`cursor`/freshness cutoff, bounded Claude/Codex candidate scans, and the Host page bound (`16 + 1`).
- Shared Claude summary scanning/cache pruning between `listSessions()` and `listCatalogEntries()`, including deleted/recreated path coverage.
- Filtered `isSidechain === true` at the canonical Claude import boundary and added the mixed-transcript regression.

Validation passed on the current `main` base:
- Core, Storage, and Runtime Host builds
- External-session regression tests: 104 passed, 1 skipped only because symlink creation is unavailable in this Windows environment
- Runtime Host external-session coordinator tests: 14 passed
- Format check, lint with warnings as errors, conflict-marker check, and PR-scope `git diff --check`

The final net diff against `origin/main` is limited to 14 external-session implementation/test files under `packages/core`, `packages/runtime-host`, and `packages/storage`; no unrelated Desktop, CLI, or other runtime changes were added.

PR #3659, opened 2026-08-24
feat(runtime-host): support managed services on macOS

M4n5ter:
## Summary

<details open>
<summary>English</summary>

Add a per-user macOS LaunchAgent backend for managed Runtime Host services while preserving the existing platform-neutral lifecycle contract and Linux systemd backend.

- install, inspect, start, stop, restart, replace, roll back, read bounded logs, and uninstall the exact managed service
- keep the plist and logs private to the user and require an active macOS GUI login session; no administrator access is requested
- share canonical launch arguments and validation between systemd and LaunchAgent
- treat intentional Host retirement as successful process completion, so both supervisors restart actual failures without a platform-specific exit-code protocol
- retain the State Root and Project data when uninstalling the managed service

Refs #3228

</details>

<details>
<summary>简体中文</summary>

为 managed Runtime Host service 增加 per-user macOS LaunchAgent backend，同时保留现有的平台无关 lifecycle contract 与 Linux systemd backend。

- 支持安装、检查、启动、停止、重启、替换、回滚、读取有界日志，以及卸载精确的 managed service
- plist 与日志仅对当前用户可见；要求 macOS 存在活跃的 GUI 登录会话，不请求管理员权限
- systemd 与 LaunchAgent 共享规范的启动参数和校验逻辑
- 将主动 Host retirement 视为成功退出，使两种 supervisor 都只重启真正的失败，不再依赖平台特定的退出码协议
- 卸载 managed service 时保留 State Root 与 Project 数据

关联 #3228

</details>

## Verification

<details open>
<summary>English</summary>

- `npm --workspace @maka/runtime-host test` — 1,120 passed
- `npm --workspace maka-agent test` — 423 passed, including the LaunchAgent lifecycle and rollback tests
- `npm run lint`
- `npm run format:check`
- `npm run check:asf-headers`
- exercised a real LaunchAgent install/status/restart/stop/start/retirement/uninstall lifecycle on macOS
- exercised packaged development-archive setup and clean uninstall with an isolated home directory
- exercised Desktop interactive SSH host-key and password prompts plus development-archive upload against localhost; final setup correctly stopped at a pre-existing State Root identity mismatch, so no existing user data was changed

</details>

<details>
<summary>简体中文</summary>

- `npm --workspace @maka/runtime-host test` — 1,120 项通过
- `npm --workspace maka-agent test` — 423 项通过，包含 LaunchAgent lifecycle 与 rollback 测试
- `npm run lint`
- `npm run format:check`
- `npm run check:asf-headers`
- 在 macOS 上验证了真实 LaunchAgent 的 install/status/restart/stop/start/retirement/uninstall 生命周期
- 使用隔离的 home 目录验证了 development archive package setup 与干净卸载
- 通过 localhost 验证了 Desktop 交互式 SSH 的 host key、密码输入和 development archive 上传；最终 setup 在已有 State Root identity 不匹配处正确停止，因此未修改现有用户数据

</details>

## AI use

Select exactly one:

Tool(s) and scope: Codex implemented and reviewed the LaunchAgent backend, lifecycle simplification, tests, documentation, and verification under maintainer direction.

## Checklist

- [ ] No


packages/cli/src/runtime-host-launch-agent-service.ts
```diff
@@ -0,0 +1,504 @@
...
+  return Number.isSafeInteger(parsed) && parsed > 0 ? parsed : null;
+}
+
+function nonNegativeInteger(value: string | undefined): number | null {
+  const parsed = Number(value);
+  return Number.isSafeInteger(parsed) && parsed >= 0 ? parsed : null;
+}
+
+function escapeXml(value: string): string {
+  if (/[\u0000-\u0008\u000b\u000c\u000e-\u001f\u007f]/u.test(value)) {
```
Astro-Han, line 517: **[P3]** This line and `runtime-host-systemd-service.ts:537` accept different sets of control characters, so the same configuration can install on macOS and throw on Linux.

Here: `/[\u0000-\u0008\u000b\u000c\u000e-\u001f\u007f]/u` — tab (U+0009), LF (U+000A) and CR (U+000D) are permitted, since all three are legal in XML text.
There: `/[\u0000-\u001f\u007f]/u` — all three are rejected.

Trigger and consequence: a `--project-root` whose path contains a tab (legal on both filesystems) renders and installs normally on macOS, and raises a `TypeError` on Linux. One config, one user intent, two outcomes — which is the inverse of what this PR is unifying.

Part of the asymmetry is not a defect and should not be "fixed": LF *must* be rejected by a line-oriented format like a systemd unit, and *is* legal inside an XML element's text. That half is dictated by the formats themselves. What is arguable is tab and CR, and — more than either — that the difference is not written down anywhere.

Smallest fix: put one shared "legal characters for a service argument" check in `runtime-host-service-launch.ts` as a common precondition for both renderers, and let each platform layer on the extra restrictions its format actually requires. Cleaner: the same, plus a comment on that function stating which characters are rejected and why one platform is stricter. A production-seam test would feed the same tab-containing `projectDirectoryRoots` to both renderers and assert the behaviour matches, or that any difference is explicitly declared.

Astro-Han (commented):
Reviewed at `0eedc4ac5aafb405c74a2cc99aba5998ccf14104`. This PR is a draft; reviewing it now was requested explicitly, so this is a comment only — no approval, and the gate below is recorded for the head as it stands rather than as a merge decision.

**No P0-P2. One P3, inline.**

### What I checked and did not find a problem with

**The shared layer is a real convergence, not a relocated `if`.** `runtime-host-service-launch.ts` exports exactly two things, and both are platform-independent facts: `runtimeHostServiceLaunchArguments` (the argv to execute) and `validateRuntimeHostServiceLaunch` (realpath + `isFile()` + node-executable / cli-readable). The -94 on `runtime-host-systemd-service.ts` is the same material leaving — the argv construction, `validateLaunchFiles`, and the command-runner pipeline that moved into `runtime-host-service-manager-process.ts`. What was removed and what was extracted are the same set. Platform dispatch has a single injection point (`runtime-host-service-management-command.ts:218`, `platform: NodeJS.Platform = process.platform`), and the two implementations are separate modules keyed by `manager: 'systemd_user' | 'launch_agent'`.

**The two platforms are in the same privilege domain.** The plist lands in `~/Library/LaunchAgents/` (`runtime-host-launch-agent-service.ts:211`) — a LaunchAgent in the user session, not `/Library/LaunchDaemons`. The systemd side is a user unit (`WantedBy=default.target`, not `multi-user.target`). Both are unprivileged and current-user, so "managed service" denotes the same thing on macOS and Linux. This is the failure mode we most expected here, and it did not occur.

**Escaping is correct on both sides.** The plist path escapes `&`, `<`, `>` before emitting text nodes; the systemd path (`quoteSystemdArgument`) double-quotes and escapes backslash before quote — correct order — and also escapes `%` (specifier) and `$` (variable expansion), which are the two most commonly missed. Paths containing spaces, `&`, `<`, or quotes are safe on both. `quoteSystemdArgument` already existed at the merge-base and this refactor did not weaken it.

**Install is a transaction with rollback, and uninstall verifies itself.** `captureLaunchAgentDeployment` snapshots prior state, failures route through `restoreFailedLaunchAgentDeployment`, and the returned `rollback` handle guards re-entry with a `rolledBack` flag. For the upgrade path, `verifyDeployment` re-renders the plist and compares it verbatim against what is on disk, plus `status.loaded`, throwing `target_mismatch` on drift. `uninstall` boots out, removes the plist and both stdout/stderr files, then re-reads state to confirm absence rather than assuming success.

### Gate status at this head

`test` is completed/success. The other workflows are correctly absent rather than missing: no `package.json`/lockfile change means `audit` should not trigger, and no sandbox or filesystem-worker path is touched, so `windows_sandbox_*` should not trigger. The single check is the correct triggered set for this diff.

Unresolved review threads: 0. The GraphQL probe was validated against a control PR that returns a populated thread set, so the zero here is a real absence rather than a broken query.

packages/cli/src/runtime-host-launch-agent-service.ts
M4n5ter: Thanks for checking this boundary. I traced the production path and do not think this is reachable through the managed-service contract.

Both newly prepared configs and persisted configs pass through `validateServiceConfig()` before either backend receives them. That validation applies `hasControlCharacters()` (`U+0000–U+001F` and `U+007F`) to Project labels and `isSafeAbsolutePath()` to every State Root, Project, Node, and CLI path. Tab, LF, and CR are therefore rejected by the shared config authority before either renderer runs.

The render functions are package-private implementation/test seams, not an independent configuration entry point. Adding another common character validator in the launch-argument layer would duplicate the existing authority without changing any supported behavior, so I am leaving this unchanged and resolving the thread.

Astro-Han (approved):
Approving at `e6d11969a3f0be0cd8adf0af1a4ed4ae37c0020b`. No findings this round.

### First, withdrawing the `[P3]` from my earlier review

That finding was wrong, and not in the "the author supplied extra context" sense — the trigger-to-outcome path I wrote was false. The stated fact held (the two control-character sets do differ), but the path I built on it — a project root containing a tab installing on macOS and throwing on Linux — cannot happen. `validateServiceConfig` (`runtime-host-service-manager.ts:804`) rejects it long before either renderer runs, via `hasControlCharacters` (`:1186`, `/[\u0000-\u001f\u007f]/u`, which does include tab, LF and CR), applied to `rootPath`, every project root `label` and `path`, and `launch.nodePath` / `cliPath`. Both entry points are covered: `:718` for a newly prepared config and `:766` for a persisted one. The renderers are package-private seams, not an independent configuration entry point, exactly as you said.

The mistake was one-directional analysis: we traced downward into the two renderers and the shared layer, and never traced upward to ask where the input comes from and whether an earlier authority already constrains it. Reachability arguments need that second direction. Thanks for pushing back rather than adding a redundant validator.

### The new commit is a convergence, not a version branch

`bootout --wait` becomes an explicit poll. `--wait` was itself the version-dependent flag, and this removes the dependency on it rather than wrapping it in a macOS-version conditional — there is no version branch anywhere in the file. The poll is also strictly stronger than what `--wait` guaranteed: it waits for the service to be unloaded **and** for the original pid to be gone (`status.pid === null || !isProcessAlive(status.pid)`), where `--wait` only covered the unload. `defaultIsProcessAlive` uses `process.kill(pid, 0)` and maps `ESRCH` to false, `EPERM` to true (the process exists but belongs to another user) and rethrows anything else — `EPERM` is the branch most often gotten wrong here. On timeout it throws `service_manager_operation_failed` rather than returning quietly.

The three lines in `runtime-host-service-manager-process.ts` are a contraction: the `timeoutMs` parameter is deleted and `runRuntimeHostServiceManagerCommand` always uses `DEFAULT_COMMAND_TIMEOUT_MS`. That overridable timeout existed only for the long `bootout --wait`, so once launchd polls on its own, no caller needs it. The shared layer loses a knob that existed for one platform — the opposite of a platform detail leaking upward.

### Prior conclusions, re-checked at this head

The abstraction still converges, and is slightly stronger for having lost that platform-motivated parameter. The LaunchAgent privilege domain and the escaping on both sides are untouched — verified by searching the new commit's diff for `escapeXml`, `renderLaunchAgentPlist`, `LaunchAgents` and `applyLaunchAgentDeployment`, all zero hits, rather than by impression. The install transaction, `verifyDeployment`'s verbatim plist comparison and the self-verifying `uninstall` all still hold, and `uninstall` now additionally waits for the original pid to die.

Considered and not reported: pid reuse between `bootout` and the poll. `status.pid` is captured before `bootout`, so if that pid were recycled to another process during the poll, `isProcessAlive` would stay true and the wait would run to timeout. That fails toward a false failure, never a false success — it cannot claim a stop succeeded when it did not — and reaching it requires wrapping the whole pid space inside the window.

### Gate

`test` is completed/success at this exact head. The single check is the correct triggered set: this commit touches only `packages/cli/**`, with no `package.json` or lockfile change (so `audit` should not fire) and no sandbox or filesystem-worker path (so `windows_sandbox_*` should not fire). Zero unresolved review threads; the one thread on this PR is the withdrawn P3, resolved. No `APPROVED` review is bound to an older commit.

PR #3660, opened 2026-08-24
fix(runtime): classify usage-limit failures behind auth statuses as billing

### LINKED ISSUE #2516: fix(runtime): distinguish usage limits from authentication failures
## What happened

While using the TUI with a valid Kimi Code login after the plan limit was exhausted, Maka reported:

```text
Authentication failed
```

That points the user toward signing in again or replacing credentials, even though the useful action is to wait for the limit to reset or check the subscription.

The shared classifier currently maps every HTTP 401/403 to `auth` and every 429 to a retryable `rate_limit`. Those status codes are not enough on their own: depending on the provider, they can mean an invalid credential, missing permission, exhausted credits, a subscription window, or transient throttling.

## How to reproduce

1. Sign in to Kimi Code with a valid account.
2. Exhaust the five-hour or weekly plan allowance.
3. Start another turn in the TUI.
4. Maka shows an authentication failure instead of a usage-limit failure.

## Environment

- Maka: current `main` still has the status-first classification described above
- Surface observed: TUI
- Provider observed: Kimi Code plan

## Suggested direction

I think this belongs at the ModelAdapter/runtime error-projection boundary, not in a Kimi-only TUI branch.

- Extract structured provider evidence first (`status`, provider error code/type, retry metadata), then map it to a stable Maka failure meaning.
- Keep authentication, permission, billing/credits, subscription usage limits, and transient rate limiting distinct where the provider gives enough evidence.
- Derive retry policy from that meaning: a short throttle may retry, but a five-hour or weekly cap should not be retried blindly.
- Let TUI, Desktop, non-interactive CLI, and Headless consume the same semantic result. Provider adapters should not own user-facing strings.
- Preserve the current redaction boundary. Do not display raw response bodies, tokens, or guessed reset times.
- For compatible relays with ambiguous errors, keep a safe generic fallback instead of pretending to know the account state.

A useful fixture matrix would cover at least:

- Kimi: expired credentials vs exhausted plan
- OpenAI: invalid key vs insufficient quota/project spend limit vs transient RPM/TPM throttling
- OpenRouter: credits, permission/guardrail, and transient rate limit
- an unknown OpenAI-compatible relay with incomplete error metadata

## Expected behavior

- The Kimi plan-limit case no longer tells the user to fix credentials.
- Non-transient account limits are not automatically retried as short throttles.
- Desktop localization, TUI text, and Headless/CLI machine-readable outcomes agree on the same failure meaning.
- Provider-specific evidence parsing is covered by fixtures; surface tests only verify projection and presentation.
- Unknown or malformed provider errors remain safe and do not leak raw payloads.

I’m happy to implement this once the failure taxonomy and fixture boundary look right.

<details>
<summary>中文说明</summary>

我在真实使用 Maka TUI 时遇到：Kimi Code 登录仍然有效，但五小时或周限额已经耗尽，TUI 却显示“认证失败”。这个提示会引导用户重新登录或更换凭据，实际应该做的是等待额度恢复或检查套餐。

当前共享分类器把所有 401/403 都当作认证失败，把所有 429 都当作可重试的临时限流。但不同服务会用这些状态表示无效凭据、权限不足、余额耗尽、套餐周期限额或瞬时限流，仅靠 HTTP 状态无法得到正确结论。

建议在 ModelAdapter/Runtime 的共享错误投影边界修复，而不是在 Kimi TUI 中写专用判断：优先读取 provider 的结构化错误码和重试信息，再投影为 Maka 的稳定错误语义；TUI、Desktop、非交互 CLI 和 Headless 共用这份结果。短期限流可以重试，五小时或周限额不应盲目重试。无法可靠识别的兼容中转站错误应安全降级，不猜测额度或恢复时间，也不泄露原始响应。

我愿意在错误分类和 fixture 边界确认后实现。

</details>

yunaremaia:
## Summary

Fixes #2516.

Providers report exhausted plan windows, credits, and subscriptions through different evidence channels, and the status-first fallback mislabeled all of them:

- Some use explicit structured codes — OpenAI's `insufficient_quota` even arrives as 429, DeepSeek sends `insufficient_balance`.
- Some gate the window behind credential-shaped **401/403** statuses for validly signed-in users (the Kimi Code plan case from the issue), which projected to `Auth` → *"Authentication failed"* — pointing the user at re-authenticating when the useful action is waiting for the window to reset or checking the subscription.

Changes in `packages/runtime/src/provider-error-classification.ts`, at the shared classifier boundary (no provider-specific branches, no user-facing strings in adapters):

- New `PROVIDER_BILLING_PROVIDER_CODES` set, checked with the other structured-code sets **before every numeric HTTP fallback** — explicit provider evidence outranks the bare status, the same precedence the capacity and overflow sets already follow. This also stops an exhausted quota arriving as 429 from being classified as a transient throttle.
- The `401/403` fallback now consults `USAGE_LIMIT_TEXT_PATTERNS` over the composite text: quota / usage-limit / plan / credit-exhaustion wording projects to `ProviderBilling` instead of `Auth`. Plain invalid-key and permission messages carry none of that vocabulary and stay `Auth`.
- No retry-policy change needed: `ProviderBilling` already maps to a non-retryable policy in `providerRetryMetadata`, so closed plan windows stop being retried blindly while transient throttles keep their `RateLimit` path.
- Redaction boundary unchanged — only the classification layer is touched; no raw bodies or tokens are surfaced anywhere new.

## Verification

Local run this time (`npm ci` unblocked by pointing the six Azure DevOps mirror URLs in `package-lock.json` at their identical public npm packages; lockfile restored before commit):

- `npm run build -w @maka/core && npm run build -w @maka/storage && npm run build -w @maka/mcp && npm run build -w @maka/runtime` — all pass
- `node --test dist/__tests__/provider-error-classification.test.js` — **15 tests, 15 pass, 0 fail**
- Bite check: reverting only the implementation commit hunk makes the two new billing-projection tests fail (`Auth` ≠ `ProviderBilling`) while the non-regression pins keep passing, then green again with the fix restored

New tests cover: structured usage-limit codes across 401/403/**429** (+ non-retryable metadata), plan-window wording through both SDK carriers (error message and raw response body after a schema-parse failure), and genuine invalid-key / permission failures staying `Auth`.

One scope note: 429 responses whose *text* suggests a long cap but that carry no structured code still classify as `RateLimit` — distinguishing those would need per-provider cap wording I can't verify offline, and guessing risked false billing positives on real throttles.

### DIFF
--- a/packages/runtime/src/__tests__/provider-error-classification.test.ts
+++ b/packages/runtime/src/__tests__/provider-error-classification.test.ts
@@ -62,6 +62,71 @@ describe('Provider error classification', () => {
     assert.doesNotMatch(serialized, /secret|private|authorization|request body/i);
   });
 
+  test('structured usage-limit codes project to billing regardless of status', () => {
+    const quotaOn401 = Object.assign(new Error('request failed'), {
+      name: 'AI_APICallError',
+      statusCode: 401,
+      data: { error: { code: 'insufficient_quota' } },
+    });
+    assert.equal(classifyError(quotaOn401), 'ProviderBilling');
+
+    const balanceOn403 = Object.assign(new Error('request failed'), {
+      name: 'AI_APICallError',
+      statusCode: 403,
+      data: { error: { code: 'insufficient_balance' } },
+    });
+    assert.equal(classifyError(balanceOn403), 'ProviderBilling');
+
+    // Explicit provider evidence outranks the numeric HTTP fallback: an
+    // exhausted quota is a closed window, not a transient throttle to retry.
+    const quotaOn429 = Object.assign(new Error('request failed'), {
+      name: 'AI_APICallError',
+      statusCode: 429,
+      data: { error: { code: 'insufficient_quota' } },
+    });
+    assert.equal(classifyError(quotaOn429), 'ProviderBilling');
+    assert.equal(providerRetryMetadata(quotaOn429).retryable, false);
+  });
+
+  test('plan-window wording on a credential-shaped status projects to billing', () => {
+    // Providers that gate subscription windows behind 401/403 for validly
+    // signed-in users (#2516): their own wording outranks the bare status,
+    // whether the SDK surfaces it as the error message or keeps it only in
+    // the raw response body after a schema-parse failure.
+    const planWindow = Object.assign(new Error('Your account plan usage limit has been reached.'), {
+      name: 'AI_APICallError',
+      statusCode: 401,
+      data: { error: { type: 'authentication_error' } },
+    });
+    assert.equal(classifyError(planWindow), 'ProviderBilling');
+    assert.equal(providerRetryMetadata(planWindow).retryable, false);
+
+    const exhaustedCredits = Object.assign(new Error('Request failed with status code 403'), {
+      name: 'AI_APICallError',
+      statusCode: 403,
+      responseBody: JSON.stringify({
+        error: { message: 'Your credits have been exhausted for this billing period.' },
+      }),
+    });
+    assert.equal(classifyError(exhaustedCredits), 'ProviderBilling');
+  });
+
+  test('genuine credential and permission failures stay auth on 401/403', () => {
+    const invalidKey = Object.assign(new Error('Invalid API key provided'), {
+      name: 'AI_APICallError',
+      statusCode: 401,
+      data: { error: { message: 'Invalid API key. Check your credentials and try again.' } },
+    });
+    assert.equal(classifyError(invalidKey), 'Auth');
+
+    const forbiddenModel = Object.assign(new Error('request failed'), {
+      name: 'AI_APICallError',
+      statusCode: 403,
+      data: { error: { message: 'You do not have access to this model.' } },
+    });
+    assert.equal(classifyError(forbiddenModel), 'Auth');
+  });
+
   test('recovers structured Codex HTTP facts through an SDK wrapper and truncates identifiers', () => {
     const cause = Object.assign(new Error('Codex OAuth request failed'), {
       name: 'OpenAiCodexHttpError',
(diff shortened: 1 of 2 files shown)


Astro-Han (commented):
### #3660 f9b0dd8 — review (bind exact head)

**Gate**: CI `test` success on this head (run 32682534889, check_runs 1/1 success after approval). Fork PR previously blocked on `action_required`.

**Verdict**: GO (no P0-P2)

Classification narrowly fixes usage-limit behind 401/403: structured billing codes outrank status, and 401/403 fallback consults usage-limit text patterns over composite text, preserving genuine Auth. Non-retryable billing via providerRetryMetadata correct. Tests cover matrix.

**Coexistence note**: #2521 overlaps same defect file; #3660 is minimal fix shape. Epoch handling for any follow-up: rebase to current main and take strictly greater than base (do not hardcode number).

Astro-Han (approved):
Approving at `f9b0dd835713`.

Gate at this exact head: hosted `test` is terminal green — note that this PR's checks had never run at all until the fork workflow was released (`32682534889`), so the earlier empty `check-runs` was "not run", not "passed". Zero review threads, no `APPROVED` review bound to any older commit, and the branch is mergeable.

The change stays inside the file that owns the defect: `packages/runtime/src/provider-error-classification.ts` and its test, +109/-1, classifying usage-limit failures behind the auth status rather than collapsing them into a generic auth error.

Worth recording for whoever handles this next: **#2521 fixes the same defect** (both branches are named for issue 2516, and both edit this same file and test), but does it across 46 files and +4428/-173. That one is separately marked NO-GO. If this lands first, #2521 will need to be rebased and reduced to whatever remains genuinely unaddressed.

Astro-Han:
LGTM — merged. Thanks for the contribution!

<details>
<summary>中文</summary>

已合并，感谢贡献。
</details>

PR #3667, opened 2026-08-24
feat(runtime-host): discover managed Host updates

### LINKED ISSUE #3664: feat(runtime-host): discover managed Host updates safely
<details open>
<summary>English</summary>

## Problem

Managed Runtime Hosts can be updated safely to an exact package, but neither an operator nor a future unattended reconciler can currently discover a channel target without immediately entering a mutation path. Mutable npm tags also do not provide the immutable version and package identity required by the existing exact-package update transaction.

Automatic updates must not infer storage or rollback safety from semantic-version distance. A target that does not carry sufficient compatibility evidence should remain available for explicit manual maintenance, but must not be applied unattended.

## Desired outcome

Add a read-only, machine-readable update check for managed Runtime Hosts that:

- accepts `latest`, `next`, or an exact Maka version;
- resolves a mutable selector once to an exact version and package integrity;
- reports the installed and target versions, availability, and whether unattended application is admitted;
- returns an explicit reason when the target is manual-only or unavailable;
- requires target-package compatibility evidence before declaring an update safe for unattended use;
- keeps released npm packages and explicit development archives on the same exact-package validation and update path.

The check must not retire the Host, switch the service, persist an update policy, or install a scheduler. The existing explicit update transaction remains the only mutation path.

Desktop currently updates a managed Host to the exact CLI package selected for that Desktop build. An operator can also invoke the exact-package CLI transaction directly. Neither path provides channel discovery or a safe admission fact for later automatic reconciliation.

Part of #3228. Related to #3204, but does not attempt to define the full Client–Host version-skew contract.

</details>

<details>
<summary>简体中文</summary>

## 问题

Managed Runtime Host 已能安全更新到一个精确 package，但 operator 和未来的无人值守 reconciler 目前都无法在不进入 mutation 路径的情况下发现频道目标。可变 npm tag 本身也不能提供现有 exact-package update transaction 所要求的不可变版本与 package identity。

自动更新不能根据 semantic version 距离猜测 storage 或 rollback 是否安全。若目标 package 没有足够的兼容性证据，仍可供人工明确执行维护，但不得由无人值守流程自动应用。

## 期望结果

为 managed Runtime Host 增加只读、machine-readable 的更新检查：

- 接受 `latest`、`next` 或精确 Maka 版本；
- 将可变 selector 一次性解析为精确版本与 package integrity；
- 报告当前版本、目标版本、是否存在更新，以及是否允许无人值守应用；
- 目标只能人工更新或当前不可用时返回明确原因；
- 只有目标 package 提供兼容性证据时，才声明该更新可以无人值守执行；
- 正式 npm package 与显式 development archive 继续进入同一套 exact-package 校验与更新路径。

检查不能让 Host 退场、切换 service、持久化更新策略或安装 scheduler。现有显式更新事务仍是唯一 mutation path。

## 替代方案或当前做法

Desktop 当前可以把 managed Host 更新到该 Desktop build 选择的精确 CLI package，operator 也可以直接调用 exact-package CLI transaction。但这两条路径都不能发现频道更新，也不能为后续自动 reconcile 提供安全准入事实。

属于 #3228。与 #3204 相关，但不尝试定义完整的 Client–Host version-skew contract。

</details>

M4n5ter:
<details open>
<summary>English</summary>

## Summary

- add a read-only `runtime-host service check-update` command for `latest`, `next`, or an exact version
- resolve mutable channels through the official npm registry to an exact version and package integrity
- publish an explicit package compatibility epoch and admit unattended use only for newer packages with matching evidence
- reject malformed or contradictory machine evidence; this PR does not download, install, retire, or switch a Host

Fixes #3664

## Verification

- `npm --workspace maka-agent test` — 432 tests passed
- `npm --workspace @maka/runtime-host test` — 1120 tests passed
- `npm run release:cli:pack -- --allow-dirty` — development tarball contains the compatibility epoch
- real npm lookups for the `next` channel and an unavailable exact version
- affected Biome lint/format checks, `git diff --check`, and `npm run check:asf-headers`

Not run: the full repository test suite; the affected CLI suite and Desktop type boundary were run directly

## AI use

Select exactly one:

Tool(s) and scope: Codex authored implementation, focused tests, and documentation under maintainer direction

## Checklist

- [ ] No

</details>

<details>
<summary>简体中文</summary>

## 摘要

- 新增只读的 `runtime-host service check-update` 命令，支持 `latest`、`next` 或精确版本
- 通过官方 npm registry 将可变频道一次性解析为精确版本与 package integrity
- 在 package 中发布显式兼容性 epoch；只有版本更新且证据匹配时才允许未来的无人值守流程使用
- 拒绝格式错误或相互矛盾的机器证据；本 PR 不下载、安装、退场或切换 Host

修复 #3664

## 验证

- `npm --workspace maka-agent test` — 432 项测试通过
- `npm --workspace @maka/runtime-host test` — 1120 项测试通过
- `npm run release:cli:pack -- --allow-dirty` — development tarball 已携带兼容性 epoch
- 使用真实 npm registry 验证 `next` 频道和不存在的精确版本
- 相关 Biome lint/format、`git diff --check` 与 `npm run check:asf-headers`

未运行仓库完整测试套件；已直接运行受影响的 CLI 套件与 Desktop 类型边界检查

## AI 使用

Codex 在维护者指导下完成实现、聚焦测试与文档

## 检查清单

- [x] 测试覆盖该变更，且在缺少实现时会失败
- [x] lint、format、typecheck 与受影响测试套件均在本地通过

本 PR 是否改变行为？是，具体见上方摘要

</details>

### DIFF
--- a/packages/cli/src/runtime-host-update-discovery.ts
+++ b/packages/cli/src/runtime-host-update-discovery.ts
@@ -85,6 +85,7 @@ export async function runManagedRuntimeHostUpdateCheckCli(
 ): Promise<number> {
   try {
     const serviceId = resolveRuntimeHostManagedServiceId(options.clientDataRoot);
+    const backend = createPlatformRuntimeHostServiceBackend(serviceId);
     const status = await manageRuntimeHostService(
       {
         action: 'status',
@@ -94,7 +95,7 @@ export async function runManagedRuntimeHostUpdateCheckCli(
         cliPath: process.argv[1] ?? '',
         ...(options.expectedTarget ? { expectedTarget: options.expectedTarget } : {}),
       },
-      createPlatformRuntimeHostServiceBackend(serviceId),
+      backend,
     );
     const currentVersion = status.service.installedVersion;
     const config = status.service.config;
@@ -104,6 +105,7 @@ export async function runManagedRuntimeHostUpdateCheckCli(
         'A Maka-managed Runtime Host service is required to check for updates',
       );
     }
+    await backend.verifyDeployment(config);
     const [candidate, currentCompatibility] = await Promise.all([
       resolveRuntimeHostRegistryUpdateCandidate(options.selector),
       readPackageCompatibility(config.launch.cliPath, currentVersion),
@@ -118,7 +120,7 @@ export async function runManagedRuntimeHostUpdateCheckCli(
         selector: options.selector,
         currentVersion,
         candidate: { version: candidate.version, integrity: candidate.integrity },
-        ...assessment,
+        outcome: assessment,
       },
     };
     writeSuccess(frame, options);
@@ -139,35 +141,21 @@ export function assessRuntimeHostUpdate(
   currentVersion: string,
   currentCompatibility: number | undefined,
   candidate: RuntimeHostUpdateCandidate,
-): Pick<RuntimeHostUpdateCheck, 'status' | 'unattended'> {
+): RuntimeHostUpdateCheck['outcome'] {
   const relation = compareProductReleaseVersions(candidate.version, currentVersion);
-  if (relation === 0) return { status: 'current', unattended: { kind: 'not_needed' } };
+  if (relation === 0) return { kind: 'current' };
   if (relation < 0) {
-    return {
-      status: 'older',
-      unattended: { kind: 'manual_only', reason: 'target_not_newer' },
-    };
+    return { kind: 'manual_action', reason: 'target_not_newer' };
   }
-  const status = 'newer' as const;
   if (currentCompatibility === undefined) {
-    return {
-      status,
-      unattended: { kind: 'manual_only', reason: 'current_compatibility_unknown' },
-    };
+    return { kind: 'manual_action', reason: 'current_compatibility_unknown' };
   }
   if (candidate.compatibility === undefined) {
-    return {
-      status,
-      unattended: { kind: 'manual_only', reason: 'target_compatibility_unknown' },
-    };
+    return { kind: 'manual_action', reason: 'target_compatibility_unknown' };
   }
-  return {
-    status,
-    unattended:
-      candidate.compatibility === currentCompatibility
-        ? { kind: 'allowed', compatibility: currentCompatibility }
-        : { kind: 'manual_only', reason: 'compatibility_mismatch' },
-  };
+  return candidate.compatibility === currentCompatibility
+    ? { kind: 'unattended_update', compatibility: currentCompatibility }
+    : { kind: 'manual_action', reason: 'compatibility_mismatch' };
 }
 
 export async function resolveRuntimeHostRegistryUpdateCandidate(
@@ -211,8 +199,7 @@ export async function resolveRuntimeHostRegistryUpdateCandidate(
     !isProductReleaseVersion(version) ||
     (selector.kind === 'exact' && version !== selector.version) ||
     typeof integrity !== 'string' ||
-    !/^sha512-[A-Za-z0-9+/]+={0,2}$/u.test(integrity) ||
-    Buffer.byteLength(integrity, 'utf8') > 512
+    !isSha512Integrity(integrity)
   ) {
     return invalidMetadata();
   }
@@ -292,25 +279,25 @@ function writeSuccess(
 ): void {
   if (options.framed) process.stdout.write(encodeRuntimeHostServiceManagementFrame(frame));
   else if (options.json) process.stdout.write(`${JSON.stringify({ ...frame, ok: true })}\n`);
-  else {
-    const check = frame.updateCheck;
-    if (check.status === 'current') {
-      process.stdout.write(
-        `Runtime Host ${check.currentVersion} already matches the selected target.\n`,
-      );
-    } else if (check.unattended.kind === 'allowed') {
-      process.stdout.write(
-        `Runtime Host ${check.candidate.version} is available for unattended update.\n`,
-      );
-    } else if (check.status === 'older') {
-      process.stdout.write(
-        `Selected Runtime Host ${check.candidate.version} is older than installed ${check.currentVersion}; manual selection is required.\n`,
-      );
-    } else {
-      process.stdout.write(
-        `Runtime Host ${check.candidate.version} is available for manual update.\n`,
-      );
-    }
+  else process.stdout.write(`${formatRuntimeHostUpdateCheck(frame.updateCheck)}\n`);
+}
+
+export function formatRuntimeHostUpdateCheck(check: RuntimeHostUpdateCheck): string {
+  if (check.outcome.kind === 'current') {
+    return `Runtime Host ${check.currentVersion} already matches the selected target.`;
+  }
+  if (check.outcome.kind === 'unattended_update') {
+    return `Runtime Host ${check.candidate.version} is available for unattended update.`;
+  }
+  switch (check.outcome.reason) {
+    case 'target_not_newer':
+      return `Selected Runtime Host ${check.candidate.version} is older than installed ${check.currentVersion}; manual selection is required.`;
+    case 'current_compatibility_unknown':
+      return `Runtime Host ${check.candidate.version} requires a manual update because the installed package has no unattended-update compatibility evidence.`;
+    case 'target_compatibility_unknown':
+      return `Runtime Host ${check.candidate.version} requires a manual update because the target package has no unattended-update compatibility evidence.`;
+    case 'compatibility_mismatch':
+      return `Runtime Host ${check.candidate.version} requires a manual update because it crosses the declared compatibility boundary.`;
   }
 }
 
@@ -350,6 +337,14 @@ function invalidMetadata(): never {
 
 function positiveInteger(value: unknown): number | undefined {
   return typeof value === 'number' && Number.isSafeInteger(value) && value > 0 ? value : undefined;
+}
+
+function isSha512Integrity(value: string): boolean {
+  if (!value.startsWith('sha512-')) return false;
+  const encoded = value.slice('sha512-'.length);
+  if (encoded.length !== 88 || !/^[A-Za-z0-9+/]+={2}$/u.test(encoded)) return false;
+  const digest = Buffer.from(encoded, 'base64');
+  return digest.length === 64 && digest.toString('base64') === encoded;
 }
 
 function parseJson(value: string): unknown {
--- a/packages/cli/src/runtime-host-update-discovery.ts
+++ b/packages/cli/src/runtime-host-update-discovery.ts
@@ -1,0 +1,365 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { spawn } from 'node:child_process';
+import { readFile } from 'node:fs/promises';
+import { dirname, join } from 'node:path';
+import { truncateUtf8 } from '@maka/core/diagnostic-log';
+import {
+  encodeRuntimeHostServiceManagementFrame,
+  RUNTIME_HOST_SERVICE_ERROR_CODE_MAX_BYTES,
+  RUNTIME_HOST_SERVICE_ERROR_MESSAGE_MAX_BYTES,
+  type RuntimeHostServiceManagementFrame,
+} from '@maka/runtime-host/operator';
+import {
+  manageRuntimeHostService,
+  resolveRuntimeHostManagedServiceId,
+  RuntimeHostServiceManagerError,
+  type RuntimeHostManagedServiceTarget,
+} from './runtime-host-service-manager.js';
+import {
+  compareProductReleaseVersions,
+  isProductReleaseVersion,
+} from './product-release-version.js';
+import {
+  createPlatformRuntimeHostServiceBackend,
+  runtimeHostServiceSummary,
+} from './runtime-host-service-management-command.js';
+import type { RuntimeHostUpdateSelector } from './runtime-host-cli.js';
+
+const PACKAGE_NAME = 'maka-agent';
+const COMPATIBILITY_FIELD = 'maka.managedRuntimeHostUpdateCompatibility';
+const REGISTRY_TIMEOUT_MS = 30_000;
+const REGISTRY_OUTPUT_MAX_BYTES = 64 * 1024;
+const MANIFEST_MAX_BYTES = 64 * 1024;
+
+export interface RuntimeHostUpdateCandidate {
+  readonly version: string;
+  readonly integrity: string;
+  readonly compatibility?: number;
+}
+
+type RuntimeHostUpdateCheck = Extract<
+  RuntimeHostServiceManagementFrame,
+  { kind: 'result'; action: 'check_update' }
+>['updateCheck'];
+
+export interface RuntimeHostUpdateCheckCliOptions {
+  readonly json: boolean;
+  readonly framed: boolean;
+  readonly clientDataRoot: string;
+  readonly defaultRootPath: string;
+  readonly selector: RuntimeHostUpdateSelector;
+  readonly expectedTarget?: RuntimeHostManagedServiceTarget;
+}
+
+class RuntimeHostUpdateDiscoveryError extends Error {
+  constructor(
+    readonly code: 'target_unavailable' | 'registry_unavailable' | 'invalid_registry_metadata',
+    message: string,
+    options?: ErrorOptions,
+  ) {
+    super(message, options);
+    this.name = 'RuntimeHostUpdateDiscoveryError';
+  }
+}
+
+export async function runManagedRuntimeHostUpdateCheckCli(
+  options: RuntimeHostUpdateCheckCliOptions,
+): Promise<number> {
+  try {
+    const serviceId = resolveRuntimeHostManagedServiceId(options.clientDataRoot);
+    const status = await manageRuntimeHostService(
+      {
+        action: 'status',
+        clientDataRoot: options.clientDataRoot,
+        defaultRootPath: options.defaultRootPath,
+        nodePath: process.execPath,
+        cliPath: process.argv[1] ?? '',
+        ...(options.expectedTarget ? { expectedTarget: options.expectedTarget } : {}),
+      },
+      createPlatformRuntimeHostServiceBackend(serviceId),
+    );
+    const currentVersion = status.service.installedVersion;
+    const config = status.service.config;
+    if (!status.service.installed || !currentVersion || !config?.managedDeploymentRoot) {
+      throw new RuntimeHostServiceManagerError(
+        'not_installed',
+        'A Maka-managed Runtime Host service is required to check for updates',
+      );
+    }
+    const [candidate, currentCompatibility] = await Promise.all([
+      resolveRuntimeHostRegistryUpdateCandidate(options.selector),
+      readPackageCompatibility(config.launch.cliPath, currentVersion),
+    ]);
+    const assessment = assessRuntimeHostUpdate(currentVersion, currentCompatibility, candidate);
+    const frame: RuntimeHostServiceManagementFrame = {
+      schemaVersion: 1,
+      kind: 'result',
+      action: 'check_update',
+      service: runtimeHostServiceSummary(status),
+      updateCheck: {
+        selector: options.selector,
+        currentVersion,
+        candidate: { version: candidate.version, integrity: candidate.integrity },
+        ...assessment,
+      },
+    };
+    writeSuccess(frame, options);
+    return 0;
+  } catch (error) {
+    const code =
+      error instanceof RuntimeHostUpdateDiscoveryError ||
+      error instanceof RuntimeHostServiceManagerError
+        ? error.code
+        : 'update_check_failed';
+    const message = error instanceof Error ? error.message : String(error);
+    writeFailure(code, message, options);
+    return 1;
+  }
+}
+
+export function assessRuntimeHostUpdate(
+  currentVersion: string,
+  currentCompatibility: number | undefined,
+  candidate: RuntimeHostUpdateCandidate,
+): Pick<RuntimeHostUpdateCheck, 'status' | 'unattended'> {
+  const relation = compareProductReleaseVersions(candidate.version, currentVersion);
+  if (relation === 0) return { status: 'current', unattended: { kind: 'not_needed' } };
+  if (relation < 0) {
+    return {
+      status: 'older',
+      unattended: { kind: 'manual_only', reason: 'target_not_newer' },
+    };
+  }
+  const status = 'newer' as const;
+  if (currentCompatibility === undefined) {
+    return {
+      status,
+      unattended: { kind: 'manual_only', reason: 'current_compatibility_unknown' },
+    };
+  }
+  if (candidate.compatibility === undefined) {
+    return {
+      status,
+      unattended: { kind: 'manual_only', reason: 'target_compatibility_unknown' },
+    };
+  }
+  return {
+    status,
+    unattended:
+      candidate.compatibility === currentCompatibility
+        ? { kind: 'allowed', compatibility: currentCompatibility }
+        : { kind: 'manual_only', reason: 'compatibility_mismatch' },
+  };
+}
+
+export async function resolveRuntimeHostRegistryUpdateCandidate(
+  selector: RuntimeHostUpdateSelector,
+  run: (args: readonly string[]) => Promise<NpmViewResult> = runNpmView,
+): Promise<RuntimeHostUpdateCandidate> {
+  const target = selector.kind === 'channel' ? selector.channel : selector.version;
+  const result = await run([
+    'view',
+    `${PACKAGE_NAME}@${target}`,
+    'version',
+    'dist.integrity',
+    COMPATIBILITY_FIELD,
+    '--json',
+  ]);
+  if (result.exitCode !== 0) {
+    const failure = parseJson(result.stdout);
+    const code = isRecord(failure) && isRecord(failure.error) ? failure.error.code : undefined;
+    throw new RuntimeHostUpdateDiscoveryError(
+      code === 'E404' ? 'target_unavailable' : 'registry_unavailable',
+      code === 'E404'
+        ? `No Maka package is published for ${target}`
+        : 'The Maka package registry is unavailable',
+    );
+  }
+  let metadata: unknown;
+  try {
+    metadata = JSON.parse(result.stdout);
+  } catch (error) {
+    throw new RuntimeHostUpdateDiscoveryError(
+      'invalid_registry_metadata',
+      'The npm registry returned invalid Maka package metadata',
+      { cause: error },
+    );
+  }
+  if (!isRecord(metadata)) return invalidMetadata();
+  const version = metadata.version;
+  const integrity = metadata['dist.integrity'];
+  if (
+    typeof version !== 'string' ||
+    !isProductReleaseVersion(version) ||
+    (selector.kind === 'exact' && version !== selector.version) ||
+    typeof integrity !== 'string' ||
+    !/^sha512-[A-Za-z0-9+/]+={0,2}$/u.test(integrity) ||
+    Buffer.byteLength(integrity, 'utf8') > 512
+  ) {
+    return invalidMetadata();
+  }
+  const compatibility = positiveInteger(metadata[COMPATIBILITY_FIELD]);
+  return { version, integrity, ...(compatibility === undefined ? {} : { compatibility }) };
+}
+
+async function readPackageCompatibility(
+  cliPath: string,
+  expectedVersion: string,
+): Promise<number | undefined> {
+  try {
+    const raw = await readFile(join(dirname(dirname(cliPath)), 'package.json'), 'utf8');
+    if (Buffer.byteLength(raw, 'utf8') > MANIFEST_MAX_BYTES) return undefined;
+    const manifest: unknown = JSON.parse(raw);
+    if (
+      !isRecord(manifest) ||
+      manifest.name !== PACKAGE_NAME ||
+      manifest.version !== expectedVersion
+    ) {
+      return undefined;
+    }
+    return isRecord(manifest.maka)
+      ? positiveInteger(manifest.maka.managedRuntimeHostUpdateCompatibility)
+      : undefined;
+  } catch {
+    return undefined;
+  }
+}
+
+interface NpmViewResult {
+  readonly exitCode: number;
+  readonly stdout: string;
+}
+
+function runNpmView(args: readonly string[]): Promise<NpmViewResult> {
+  return new Promise((resolve, reject) => {
+    // Windows requires its command shell to resolve npm.cmd. The only dynamic
+    // argument has already passed the strict release-version/channel parser.
+    const child = spawn('npm', args, {
+      stdio: ['ignore', 'pipe', 'ignore'],
+      windowsHide: true,
+      shell: process.platform === 'win32',
+      timeout: REGISTRY_TIMEOUT_MS,
+      killSignal: 'SIGKILL',
+    });
+    let stdout = '';
+    let bytes = 0;
+    child.stdout.setEncoding('utf8');
+    child.stdout.on('data', (chunk: string) => {
+      bytes += Buffer.byteLength(chunk, 'utf8');
+      if (bytes > REGISTRY_OUTPUT_MAX_BYTES) {
+        child.kill('SIGKILL');
+        return;
+      }
+      stdout += chunk;
+    });
+    child.once('error', reject);
+    child.once('close', (exitCode) => {
+      if (bytes > REGISTRY_OUTPUT_MAX_BYTES) {
+        reject(
+          new RuntimeHostUpdateDiscoveryError(
+            'invalid_registry_metadata',
+            'The npm registry returned oversized Maka package metadata',
+          ),
+        );
+        return;
+      }
+      resolve({ exitCode: exitCode ?? 1, stdout });
+    });
+  });
+}
+
+function writeSuccess(
+  frame: Extract<RuntimeHostServiceManagementFrame, { kind: 'result'; action: 'check_update' }>,
+  options: RuntimeHostUpdateCheckCliOptions,
+): void {
+  if (options.framed) process.stdout.write(encodeRuntimeHostServiceManagementFrame(frame));
+  else if (options.json) process.stdout.write(`${JSON.stringify({ ...frame, ok: true })}\n`);
+  else {
+    const check = frame.updateCheck;
+    if (check.status === 'current') {
+      process.stdout.write(
+        `Runtime Host ${check.currentVersion} already matches the selected target.\n`,
+      );
+    } else if (check.unattended.kind === 'allowed') {
+      process.stdout.write(
+        `Runtime Host ${check.candidate.version} is available for unattended update.\n`,
+      );
+    } else if (check.status === 'older') {
+      process.stdout.write(
+        `Selected Runtime Host ${check.candidate.version} is older than installed ${check.currentVersion}; manual selection is required.\n`,
+      );
+    } else {
+      process.stdout.write(
+        `Runtime Host ${check.candidate.version} is available for manual update.\n`,
+      );
+    }
+  }
+}
+
+function writeFailure(
+  code: string,
+  message: string,
+  options: RuntimeHostUpdateCheckCliOptions,
+): void {
+  const error = {
+    code: truncateUtf8(code, RUNTIME_HOST_SERVICE_ERROR_CODE_MAX_BYTES) || 'update_check_failed',
+    message:
+      truncateUtf8(message, RUNTIME_HOST_SERVICE_ERROR_MESSAGE_MAX_BYTES) ||
+      'Runtime Host update check failed',
+  };
+  if (options.framed) {
+    process.stdout.write(
+      encodeRuntimeHostServiceManagementFrame({
+        schemaVersion: 1,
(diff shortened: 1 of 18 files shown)


Astro-Han (commented):
Review at `5bc77138` (still a draft, so this is context rather than a gate). One **[P2]**, filed inline on `runtime-host-update-discovery.ts`.

The headline boundary holds: this really is discovery only. I checked the whole dispatch — `npm view`, service status, local manifest reads, nothing that installs or executes a downloaded artifact. Version comparison also matches `scripts/release-version.mjs` (build metadata rejected, prereleases compared numerically).

The P2 is about what the unattended verdict is grounded in, and it matters more once the second commit's "discovery is authoritative" intent is realised downstream.

Not approving: draft, `test` has not gone terminal on this head, and the P2 is open.

<details>
<summary>中文</summary>

在 `5bc77138` 上审（仍是 draft，所以这条是背景而非门禁结论）。一条 **[P2]**，行内提在 `runtime-host-update-discovery.ts`。

"只发现不升级"这条边界是守住的：整条命令只有 `npm view`、服务状态和本地 manifest 读取，没有任何安装或执行下载物的路径。版本比较也和 `scripts/release-version.mjs` 一致（拒 build metadata，预发布号按数字比）。P2 说的是 unattended 结论建立在什么之上——等第二个 commit 那句"让 discovery 成为权威"在下游落地，这一点会更要紧。

不 approve：draft、`test` 在本 head 未终态、且 P2 未闭。
</details>

packages/cli/src/runtime-host-update-discovery.ts
```diff
@@ -0,0 +1,360 @@
...
+    return { kind: 'manual_action', reason: 'target_not_newer' };
+  }
+  if (currentCompatibility === undefined) {
+    return { kind: 'manual_action', reason: 'current_compatibility_unknown' };
+  }
+  if (candidate.compatibility === undefined) {
+    return { kind: 'manual_action', reason: 'target_compatibility_unknown' };
+  }
+  return candidate.compatibility === currentCompatibility
+    ? { kind: 'unattended_update', compatibility: currentCompatibility }
```
Astro-Han, line 165: [P2] Unattended admission trusts the packument compatibility field, not the integrity-pinned tarball.

`assessRuntimeHostUpdate` returns `unattended_update` purely on `candidate.compatibility === currentCompatibility`, where `candidate.compatibility` comes from the npm packument field `maka.managedRuntimeHostUpdateCompatibility`. `dist.integrity` is parsed, format-checked as sha512 and echoed on `candidate`, but it takes no part in that decision — so the closed outcome rests on metadata that is served separately from the tarball the integrity hash actually covers.

`runNpmView` compounds it: `spawn('npm', args, …)` passes no `env`, so the child inherits the caller's cwd, `NPM_CONFIG_*` and npmrc. Running `check-update` from a directory with a custom registry resolves the packument elsewhere.

This is **not merge-blocking today** — I confirmed the command genuinely does not install: the whole path is `npm view` plus service status and local manifest reads, no install, no exec of a downloaded artifact. That is why it is a P2 and not a P1. But the second commit makes discovery authoritative, so it is worth writing the rule down before a reconciler starts consuming `outcome.kind`: any install path must verify tarball against the pinned sha512 and re-read the epoch from the *extracted* `package.json`, never from the packument. Pinning the registry on the discovery path would be cheap now.

Related, and not worth its own comment: the tests do not lock `compatibility_mismatch`. Relaxing `===` to "any number admits" would not turn them red.

<details>
<summary>中文</summary>

`unattended_update` 的准入完全取决于 packument 里的 `maka.managedRuntimeHostUpdateCompatibility` 与本地值是否相等；`dist.integrity` 虽然被解析并做了 sha512 格式校验，却没有参与这个判定——也就是说，闭合结论建立在与 tarball 分开分发的元数据上，而 integrity 盖的是 tarball。`runNpmView` 的 `spawn('npm', …)` 没有传 `env`，会继承调用方的 cwd、`NPM_CONFIG_*` 和 npmrc，在配了自定义 registry 的目录里跑 check-update 就会去别处取 packument。

今天不阻断合并：我确认了这条命令确实不安装（全程只有 `npm view`、服务状态和本地 manifest 读取），所以定 P2 而非 P1。但第二个 commit 的意图是让 discovery 成为权威，建议在被 reconciler 消费之前先把规则写死：任何安装路径都必须用钉住的 sha512 校验 tarball，并从解压后的 `package.json` 重读 epoch，而不是从 packument 读；发现侧顺手把 registry 钉住成本也很低。

附带一条不单开：测试没有锁住 `compatibility_mismatch`，把 `===` 放宽成"有数字就放行"也不会变红。
</details>

M4n5ter: <details open>
<summary>English</summary>

Thanks — I agree that the registry authority and the missing mismatch regression are actionable here.

Commit `f369bd820` now pins `npm view` to `https://registry.npmjs.org/` and runs it outside the invoking project, so a project-local npmrc cannot redirect discovery. Canonical release-version and sha512 evidence validation now belongs to the framed operator contract and is reused by the CLI; the frame also rejects exact-target mismatches and outcomes that contradict the version relation. The focused test now covers `compatibility_mismatch`.

I kept archive verification out of this read-only PR. The packument compatibility field and integrity now come from the same explicit trusted registry authority; SRI authenticates downloaded bytes, not publisher metadata. The CLI documentation records the required mutation boundary: any updater must verify the downloaded archive against the pinned integrity and re-read compatibility from the extracted package manifest before admission.

</details>

<details>
<summary>简体中文</summary>

感谢指出。我同意 registry authority 与缺失的 mismatch 回归覆盖是本 PR 中应处理的问题。

提交 `f369bd820` 已将 `npm view` 固定到 `https://registry.npmjs.org/`，并让它在调用项目之外运行，因此项目本地 npmrc 无法重定向发现。规范的 release version 与 sha512 证据校验现在由 framed operator contract 统一拥有并供 CLI 复用；frame 还会拒绝精确目标不匹配，以及与版本关系矛盾的 outcome。聚焦测试也已覆盖 `compatibility_mismatch`。

archive 校验仍不放进这个只读 PR。packument 中的兼容性字段与 integrity 现在来自同一个明确的可信 registry authority；SRI 证明下载字节，而不证明发布者元数据。CLI 文档已记录后续 mutation 的硬边界：更新程序在准入前必须按固定 integrity 校验下载的 archive，并从解压后的 package manifest 重新读取兼容性。

</details>

packages/cli/src/runtime-host-update-discovery.ts
```diff
@@ -0,0 +1,364 @@
...
+  if (relation < 0) {
+    return { kind: 'manual_action', reason: 'target_not_newer' };
+  }
+  if (currentCompatibility === undefined) {
+    return { kind: 'manual_action', reason: 'current_compatibility_unknown' };
+  }
+  if (candidate.compatibility === undefined) {
+    return { kind: 'manual_action', reason: 'target_compatibility_unknown' };
+  }
+  return candidate.compatibility === currentCompatibility
```
ARE404, line 164: [P2] (re-check of the prior P2 from review 5004496546) — **not closed on this head.** `unattended_update` still depends only on `candidate.compatibility === currentCompatibility`. `dist.integrity` is validated to a sha512 shape and carried into the frame, but it never enters this decision; a registry-served candidate whose version+compatibility happen to match is framed as `unattended_update` regardless of integrity misbinding. Nothing installs on this path (discovery-only), so severity stays P2, not P1 — but callers decoding the frame should not treat `unattended_update` as evidence that the tarball is trustworthy.

packages/cli/src/runtime-host-update-discovery.ts
```diff
@@ -0,0 +1,364 @@
...
+interface NpmViewResult {
+  readonly exitCode: number;
+  readonly stdout: string;
+}
+
+function runNpmView(args: readonly string[]): Promise<NpmViewResult> {
+  return new Promise((resolve, reject) => {
+    // Windows requires its command shell to resolve npm.cmd. The only dynamic
+    // argument has already passed the strict release-version/channel parser.
+    const child = spawn('npm', args, {
```
ARE404, line 250: [P3] `spawn('npm', args, …)` still passes no explicit `env`, so the child inherits the caller's environment (NPM_CONFIG_*, proxy, etc.). Two mitigations are already in place that reduce (but don't remove) the original concern: `cwd` is now pinned to `homedir()` (no longer the caller's cwd), and the hardcoded `--registry https://registry.npmjs.org/` CLI flag overrides any `NPM_CONFIG_REGISTRY` env. And this invocation is `npm view` (read-only). Residual: e.g. an injected or misconfigured `NPM_CONFIG_CACHE`/`NPM_CONFIG_PROXY` could redirect/observe the metadata fetch. Low severity; consider a minimal explicit env (PATH + a pinned HOME) for defense in depth.

packages/runtime-host/src/operator/service-management-frame.ts
```diff
@@ -66,6 +80,52 @@ const boundedNonEmptyString = (maxBytes: number) =>
...
+      .object({
+        version: PRODUCT_RELEASE_VERSION_SCHEMA,
+        integrity: PACKAGE_INTEGRITY_SCHEMA,
+      })
+      .strict(),
+    outcome: z.discriminatedUnion('kind', [
+      z.object({ kind: z.literal('current') }).strict(),
+      z
+        .object({
+          kind: z.literal('unattended_update'),
```
ARE404, line 107: [P3] The `unattended_update` frame schema permits **prerelease** candidates — `PRODUCT_RELEASE_VERSION_SCHEMA` (isProductReleaseVersion) accepts `1.2.4-rc.1`/`2.0.0-beta.1` — and `assessRuntimeHostUpdate` will frame a stable-installed `1.0.0` as `unattended_update` for a `2.0.0-beta.1` candidate as long as compatibility matches. The `next` channel is an explicit prerelease opt-in and the current tests assert beta→beta unattended is intended, so this is not a hard bug; but nothing in the frame distinguishes stable vs prerelease for a downstream auto-install consumer. If a future supervisor auto-installs on `unattended_update`, it would install a `next`-channel prerelease without any marker. Consider carrying a `prerelease: boolean` (or the dist-tag) in the frame so consumers can refuse prerelease promotion.

ARE404 (commented):
## Review — `feat(runtime-host): discover managed Host updates`

**Coverage (start):** FULL for the discovery authority and its wire contract — I read `runtime-host-update-discovery.ts` (364L) end to end, `operator/update-package-evidence.ts` (80L) end to end, the full `operator/service-management-frame.ts` check_update schema and superRefine, the CLI selector parsing (`runtime-host-cli.ts`), the dispatch (`cli-core.ts`), and the update-discovery unit tests. I ran `git merge-tree --write-tree origin/main <head>` (clean, no conflict) and a full-repo grep for consumers of `unattended_update`/`check_update`/`updateCheck` and for any `spawn('npm', … install …)`. **Sampled**: I did not exhaustively re-review the pre-existing, unchanged `runtime-host-service-manager.ts` install path (only confirmed it is untouched by this PR and not wired to the discovery result), nor every changed doc/README/test fixture. The gate `test` (12m39s) and `audit` (34s) are green **on this exact head** `16968be6`.

**The three required answers:**

**1. Is the prior P2 closed on this head? — NO, it is open (unchanged on this dimension).** `unattended_update` still depends solely on `candidate.compatibility === currentCompatibility` (inline @ :164); `dist.integrity` is sha512-validated but never enters the decision. `spawn('npm',…)` still passes no explicit `env` (inline @ :252) — though `cwd` is now pinned to `homedir()` and the hardcoded `--registry` overrides env registry, which narrows the env sub-risk.

**2. Does the discover-only / never-install boundary still hold? — YES, it holds; the P2 does NOT escalate to P1.** I found no path in the new commits that triggers an install: `check-update` only runs `npm view` (read-only) and writes a frame to stdout; the frame's `unattended_update` verdict has no consumer that runs `update`. The pre-existing `update` action lives in the separately-unchanged service-manager and takes an explicit target version (no self-discovery). I did not find a way any added code executes `npm install`/`npm ci`.

**3. Version-comparison edges: largely correct, with one residual.** Firm: non-semver/malformed candidate versions are rejected (`isProductReleaseVersion`) → hard `invalid_registry_metadata` error, not default-can-upgrade. Missing/abnormal `compatibility` on either side fails **closed** to `manual_action` (never `unattended_update`). Reserve: a stable install can be framed `unattended_update` for a `next`-channel **prerelease** (e.g. stable `1.0.0` → `2.0.0-beta.1`) with no prerelease marker on the frame (inline @ service-management-frame.ts:107). Behavior is intended per the tests, but downstream auto-install consumers have no way to decline prerelease promotion.

**Open directions you asked me to check:**
- **Release metadata trust:** the only decision inputs are `candidate.version` + `candidate.compatibility`, both registry-served; `dist.integrity` and `compatibility` are validated for *format* (sha512 shape; positive integer) but not *origin* — that is the documented P2 instance, and I found **no additional** format-vs-origin instance beyond it. Positive: `--registry` is hardcoded (env registry can't redirect) and current-side `compatibility` is read from the local installed manifest (trusted).
- **Old path / double authority:** **no duplicate authority for the same component.** The desktop `window.maka.app.checkForUpdates()` (electron auto-update for the *desktop app*) is pre-existing and **not part of this PR** (`about-settings-page.tsx` is not a changed file here); this PR's CLI `check-update` targets the *managed Runtime Host service* — a different process/scope. For the managed service there is exactly **one** discovery path (this one); the pre-existing `update` command does not self-discover. So the 871/6-additive pattern does not leave a second truth for the same update decision.

**Findings summary:** 1 × P2 (open, not escalated — the two closed/open facts above are the basis), 2 × P3 (spawn env residual; prerelease-unmarked frame). No P0/P1. I did not invent a finding to escalate; I corroborated the P2 as not closed and did not find grounds to move it to P1.

**Per-coverage note:** "not found within my coverage = the two open directions show no defect" is scoped to the discovery authority + wire contract and the consumer grep; I did not assert anything about the unchanged install implementation.

<details>
<summary>简体中文</summary>

精读发现权威全链路（FULL），merge-tree 无冲突、CI 在 exact head 绿。三答：①旧 P2 未闭合——`unattended_update` 仍只看 `compatibility` 相等，`dist.integrity` 校验了但没进决策；spawn 仍无显式 env（但 cwd 已钉 homedir、硬编码 `--registry` 压过 env 仓库，env 风险收窄）。②「只发现不升级」边界守住——无任何新路径触发安装，`check-update` 只跑只读 `npm view`、写 frame，`unattended_update` 无消费者触发 `update`；P2 不升 P1。③版本比较大致正确：非语义/畸形 version 硬拒、compatibility 缺失/异常 fail-closed 到 manual；保留项——稳定版可被 `next` 预发布（如 2.0.0-beta.1）框成 unattended_update 但 frame 无 pre-release 标记（下游自动安装消费方无法拒绝）。开放面：release 元数据只做格式校验非来源校验（正是已知 P2，未发现新实例）；无双权威（desktop electron 自动更新是另一进程/非本 PR；服务只有这一个发现路径）。结论：1×P2（未升 P1）+2×P3，无 P0/P1。

</details>

Astro-Han (commented):
**Independent review at exact head `16968be64711fae550130912c4916f0a0d643a9b` — one of several lines reviewing this PR separately. No P0–P2. Two P3, and a re-grade of our own earlier P2.**

**Coverage**: the whole of `runtime-host-update-discovery.ts` (364 new lines), `update-package-evidence.ts`, the `check_update` frame schema in `service-management-frame.ts`, the CLI selector parser and the `cli-core.ts` dispatch. **Excluded**: the 222-line test file was read for what it pins, not audited as a suite; the README/`package.json`/`release-cli-package.mjs` edits were read but not independently verified against the release process; `runtime-host-ssh-terminal.ts` and the two touched test files outside this feature were not reviewed.

---

## Re-grading our own [P2] from review `5004496546` down to [P3], and saying why

That review — filed against the earlier head `5bc77138` — held that `unattended_update` turns solely on a packument field while `dist.integrity` is format-checked, carried, and takes no part in the decision, and that `spawn('npm', …)` passes no `env`.

**All three facts still hold at this head.** `assessRuntimeHostUpdate` (`runtime-host-update-discovery.ts:148-167`) consumes only the version relation and the two compatibility integers; `integrity` is validated at `:212` and placed in the frame at `:130`, and is compared against nothing.

**But the reachability that justified P2 does not hold, and we should say so rather than leave the grade standing.** We traced the installing path: `runtime-host-service-update` dispatches to `runManagedRuntimeHostUpdateCli` with `sourcePackageRoot` and `version` (`cli-core.ts:280-295`) — **it takes no registry candidate and no integrity at all.** A tree-wide search for consumers of `candidate.integrity` returns exactly one hit: the schema field that declares it. So a wrong verdict here cannot steer an install toward attacker-chosen bytes; the worst outcome is a misleading report.

**So the substance is different from what we originally filed, and it is a design point, not a security one:**

### [P3] `candidate.integrity` is published evidence with no consumer

The module is named `update-package-evidence.ts` and the frame validates the field through `PACKAGE_INTEGRITY_SCHEMA`, which makes it read as verified provenance. It is not: `isSha512PackageIntegrity` (`update-package-evidence.ts:57-63`) checks the `sha512-` prefix, an 88-character base64 body, and a 64-byte round-trip — that is a well-formedness check on a string the registry sent, in the same response as the version it accompanies. Nothing in the tree compares it to a downloaded artifact.

Two honest resolutions, and we do not think this PR has to pick the second:
- drop the field until a consumer exists — the strict schema means adding it back later is a one-line change and an epoch bump you are already paying for; or
- keep it, and state in the schema comment what it is (a registry assertion recorded for a future unattended updater) and what it is not (evidence the bytes were verified). The current naming asserts the stronger reading.

### [P3] The `npm` child inherits TLS and proxy configuration from the caller's environment

`runNpmView` (`:248-259`) passes `cwd: homedir()`, `stdio: ['ignore','pipe','ignore']`, `windowsHide`, a timeout and `killSignal` — and no `env`.

**Credit where it is due: passing `--registry` explicitly (`:181-182`) does defend against registry substitution.** We tested npm's precedence rather than assuming it:

```
$ NPM_CONFIG_REGISTRY=https://evil.example.com/ npm config get registry --registry https://registry.npmjs.org/
https://registry.npmjs.org/
$ NPM_CONFIG_REGISTRY=https://evil.example.com/ npm config get registry
https://evil.example.com/
```

The flag wins. **But the same test shows what the flag does not cover:**

```
$ NPM_CONFIG_STRICT_SSL=false npm config get strict-ssl --registry https://registry.npmjs.org/
false
```

So `NPM_CONFIG_STRICT_SSL`, proxy settings and `NPM_CONFIG_USERCONFIG` still apply to this child. Given the finding above — nothing installs from this output — the consequence is bounded to a false or attacker-influenced *verdict*, which is why this is P3 and not higher. An explicit `env` allowlist would close it, and `cwd: homedir()` shows the author was already thinking about this class of exposure.

---

## What we tried to break and could not

- **Version comparison, including prereleases.** `parseProductReleaseVersion` (`:65-79`) rejects build metadata outright, rejects leading-zero numeric prerelease identifiers, and `compareProductReleaseVersions` implements semver precedence correctly: release outranks prerelease at equal core (`:37-38`), numeric identifiers compare numerically via `BigInt` (`:47-48`), numeric outranks alphanumeric in the right direction (`:50-51`), and a shorter identifier list loses (`:42-43`). We could not construct a pair that misorders, and `1.2.3-rc.1` vs `1.2.3` resolves to `target_not_newer`, not an update.
- **Missing or malformed fields default to "cannot update", not "can".** Unknown current compatibility, unknown target compatibility, and any mismatch all return `manual_action` (`:158-166`); `positiveInteger` (`:350-352`) rejects `0`, non-integers and non-numbers; `readPackageCompatibility` (`:220-241`) returns `undefined` on a name mismatch, a version mismatch, an oversized manifest, or any thrown error. **Every branch we found fails closed.**
- **The "discovery only" boundary holds.** `runtime-host-service-update` already existed on the merge base; this PR adds exactly one dispatch case, and the new module contains no install or exec path. We were prepared to raise this to P1 if any path could install — none can.
- **Command injection through the selector.** `shell: true` on Windows is safe here because `parseUpdateSelector` admits only `latest`, `next`, or a string passing `isProductReleaseVersion`; no range, no separator, no metacharacter survives. The comment at `:250-251` states this and is accurate.
- **Exact-selector confusion.** Enforced twice — in the resolver (`:210`) and again by the frame's `superRefine`.
- **Oversized and slow registry responses.** Both bounded (64 KiB, 30 s) and both land on a typed failure code.

## Where the surface widened without an old path retiring

`+871/-6` is a new capability rather than a replacement, so we looked for the usual failure — an older path left in place giving a second answer. **We did not find one:** `check-update` is a new action with a new frame variant, and no prior update-discovery path exists to disagree with it.

## Gates at this exact head

`test` and `audit` both terminal `success`; `MERGEABLE`. **We are deliberately not issuing an APPROVE**: other independent lines are still reviewing this PR at the maintainer's request, and a second opinion arriving after an approval is worth less than one arriving before it.

<details>
<summary>中文</summary>

在 exact head `16968be6` 上的独立审查（本 PR 有多条独立线各自发布）。**无 P0–P2，两条 P3，并且下调我们自己此前的一条 P2。**

**覆盖面**：`runtime-host-update-discovery.ts` 全文、`update-package-evidence.ts`、`check_update` frame schema、CLI selector 解析、`cli-core.ts` 分派。**排除面**：222 行测试只读了它钉住什么，未作为套件审计；README / `package.json` / `release-cli-package.mjs` 只读未独立核对发布流程；本特性之外的三个文件未审。

**把我们自己的 P2 降为 P3，并说明为什么**：原 P2 的三条事实在新 head 上**全部仍然成立**——判定只用版本关系与两个 compatibility 整数，`integrity` 只在 `:212` 校验、`:130` 放进 frame，不参与任何比较。**但支撑 P2 的可达性不成立**：安装路径 `runtime-host-service-update` 用的是 `sourcePackageRoot` 与 `version`（`cli-core.ts:280-295`），**完全不接收 registry candidate 和 integrity**；全树搜索 `candidate.integrity` 的消费者只有声明它的那个 schema 字段。所以错误判定无法把安装导向攻击者选定的字节，最坏只是一份误导性报告。

**[P3] `candidate.integrity` 是发布出来但无人消费的证据。** 模块名叫 `update-package-evidence.ts`、字段过 `PACKAGE_INTEGRITY_SCHEMA`，读起来像已验证的来源证明；实际上 `isSha512PackageIntegrity` 只校验前缀、88 字符 base64 与 64 字节回环——是对 registry 自己发来的字符串做**格式**检查。两个诚实的处理：删掉直到有消费者（strict schema 让日后加回只是一行 + 一次你本来就要付的 epoch），或者留下并在 schema 注释里写清它是什么（为将来的无人值守更新器记录的 registry 断言）和不是什么（不代表字节被验证过）。现在的命名主张了更强的那一层。

**[P3] `npm` 子进程继承调用方环境里的 TLS 与代理配置。** `runNpmView`（`:248-259`）传了 `cwd: homedir()`、`stdio`、`windowsHide`、超时与 `killSignal`，**没传 `env`**。**该给的分要给：显式 `--registry` 确实防住了 registry 替换**，我们实测了 npm 的优先级而不是假设——带 `--registry` 时 flag 压过 `NPM_CONFIG_REGISTRY`，不带时环境变量生效。**但同一组实测也显示 flag 覆盖不到的部分**：`NPM_CONFIG_STRICT_SSL=false` 在带 `--registry` 时仍然生效。结合上一条（此输出不驱动任何安装），后果被限制在**判定失真**，所以是 P3 而非更高。加一个显式 `env` 白名单即可关闭；`cwd: homedir()` 说明作者本来就在防这一类暴露。

**攻不动的部分**：预发布版本比较（`parseProductReleaseVersion` 拒 build metadata、拒前导零数字标识；`compareProductReleaseVersions` 的 semver 优先级四个分支都对，`1.2.3-rc.1` vs `1.2.3` 落在 `target_not_newer`）；字段缺失/异常**每一支都 fail closed**（`positiveInteger` 拒 0 与非整数，`readPackageCompatibility` 名称/版本/超长/抛错全返回 undefined）；**"只发现不升级"的边界守住了**（`update` 动作在 merge base 上就存在，本 PR 只加一个分派，新模块无任何安装路径——若有可安装路径我们会升 P1）；Windows 上 `shell: true` 安全（selector 只收 `latest`/`next`/严格版本号）；exact selector 双重校验；超大与超时响应都有界且落在有类型的失败码上。

**新增能力未留双权威**：`+871/-6` 是新增而非替换，我们找过"旧路径没退场"这个常见失败，**没找到**——此前不存在会给出不同答案的更新发现路径。

门禁：`test` 与 `audit` 均终态 success，MERGEABLE。**我们有意不发 APPROVE**：按 maintainer 要求还有其他独立线在审，晚于批准到达的第二意见价值低于早于批准到达的。
</details>

Astro-Han (commented):
**Re-binding my review `5005061272` to the new head `3e618aafc2dd1d6cdc40574a69ca74dd88637e3a`. Both P3s still stand; nothing new, and one thing in the new commit that looks alarming is actually correct.**

The delta since `16968be6` is three files, `+13/-7`.

**The part worth stopping on: `runNpmView` lost `shell: process.platform === 'win32'` and `windowsHide`, along with the comment that justified them** ("Windows requires its command shell to resolve npm.cmd").

That reads like a regression, because the removed comment was factually right — Node cannot execute `npm.cmd` without a shell. **It is not a regression, and here is why:** `runManagedRuntimeHostUpdateCheckCli` calls `createPlatformRuntimeHostServiceBackend` at `runtime-host-update-discovery.ts:90`, well before `runNpmView` is reachable, and that factory is closed to two platforms:

```ts
if (platform === 'linux') return createSystemdUserRuntimeHostService(serviceId);
if (platform === 'darwin') return createLaunchAgentRuntimeHostService(serviceId);
throw new RuntimeHostServiceManagerError(
  'unsupported_platform',
  'Managed Runtime Host services currently require Linux or macOS',
);
```

So on Windows this command already fails earlier with a typed `unsupported_platform`, and those two spawn options could never execute. **Deleting them removes a claim the code could not honour** — the previous state was worse than the new one, because a reader would take the comment as evidence that Windows was a supported path here. Good change.

**Both P3s from my earlier review survive unchanged:**

1. `candidate.integrity` is still published evidence with no consumer — `assessRuntimeHostUpdate` is untouched by this delta, and the only reader of the field in the tree is still the schema that declares it.
2. The `npm` child still receives no explicit `env`. The delta removed options from that `spawn` call but did not add `env`.

**Gates at this head**: `audit` terminal `success`; `test` was still `in_progress` (run `32697750743`) when this was written, so **no APPROVE from me** — and other independent lines are still working, which is the other reason to hold.

<details>
<summary>中文</summary>

把我的 review `5005061272` 重新绑定到新 head `3e618aafc`。**两条 P3 均仍成立，无新增；而新提交里一处看着像回归的改动其实是对的。**

自 `16968be6` 起的 delta 是三个文件、`+13/-7`。

**值得停下来看的一处**：`runNpmView` 删掉了 `shell: process.platform === 'win32'` 和 `windowsHide`，连带删掉了为它们辩护的注释（"Windows 需要命令外壳才能解析 npm.cmd"）。

这读起来像回归，因为**被删的注释在事实上是对的**——Node 不通过 shell 无法执行 `npm.cmd`。**但它不是回归**：`runManagedRuntimeHostUpdateCheckCli` 在 `runtime-host-update-discovery.ts:90` 就调用了 `createPlatformRuntimeHostServiceBackend`，远早于 `runNpmView` 可达，而那个工厂只对两个平台开放，其余一律抛 `unsupported_platform`（"Managed Runtime Host services currently require Linux or macOS"）。

所以在 Windows 上这条命令本来就更早失败于一个有类型的错误，那两个 spawn 选项永远执行不到。**删掉它们移除的是一个代码兑现不了的声明**——原状态比现在更糟，因为读者会把那条注释当成"此处支持 Windows"的证据。这个改动是好的。

**两条 P3 原样成立**：① `candidate.integrity` 仍是发布出来却无人消费的证据——本次 delta 未触及 `assessRuntimeHostUpdate`，全树唯一读它的仍是声明它的那个 schema；② `npm` 子进程仍未收到显式 `env`，本次 delta 只从那个 `spawn` 调用里删选项，没有加 `env`。

门禁：`audit` 终态 success；撰写时 `test` 仍 in_progress（run `32697750743`），**故我不发 APPROVE**——另有独立线在审，这是不发的第二个理由。
</details>

ARE404 (commented):
## Review — exact head `3e618aafc2dd1d6cdc40574a69ca74dd88637e3a`

**Coverage:** FULL for the 17-file fixed-point diff and issue #3664: discovery CLI/parser/dispatch, service verification and local manifest evidence, registry resolution and bounded subprocess behavior, version ordering, framed operator contract, release manifest generation, Desktop management narrowing, tests/docs, and repo-wide searches for discovery consumers, installation reachability, and prior Host-update authorities. **Sampled/excluded:** the pre-existing explicit update transaction was traced only far enough to prove no new `check_update` path reaches staging/retirement/replacement; I did not re-audit its unchanged internal mutation mechanics. The original head and the freshness commit were both checked against their exact diffs.

**Verdict: GO within that scope — no actionable P0–P3.** No hard CONTRIBUTING/ARCHITECTURE violation. Exact-head `test` run `32697750743` completed SUCCESS (11m38s); exact-head `audit` run `32697750791` completed SUCCESS; PR is MERGEABLE. This is COMMENT only and does not merge.

### Required questions

1. **Earlier integrity/env P2: not sustained at this head.** The outcome uses version relation plus current/target compatibility; integrity is validated and emitted as the immutable candidate identity, but does not directly select the outcome. That is honest for discovery: the new path has no consumer that installs bytes, and the README explicitly requires a future updater to verify the downloaded archive against the emitted integrity and confirm compatibility from its extracted manifest. `npm view` hard-pins the official registry. The child still inherits TLS/proxy config, but I could not establish an additional privilege boundary or deterministic false admission; this remains defense-in-depth, not a blocker.
2. **Discovery-only boundary: holds.** `check_update` performs status/deployment verification, local manifest reads, one bounded `npm view`, framing, and presentation. It never imports or calls staging, retirement, activation, replacement, or update. The existing explicit transaction remains the sole mutation path, so there is no P1 escalation.
3. **Version edges:** prerelease ordering is correct; malformed/non-product versions and missing/malformed compatibility fail closed. A provisional build-metadata concern was withdrawn after tracing product identity: published releases reuse the canonical no-build grammar, and generated development archives append only canonical `dev-<hash>` prerelease forms. The looser SSH package-specifier guard is transport validation, not a second product-version authority.

**Open directions:** version/integrity/compatibility come from one fixed-registry lookup, with artifact revalidation explicitly deferred to the future updater; no second metadata-trust defect was established. No older managed-Host discovery authority exists: this module owns discovery, while the pre-existing exact transaction owns mutation. The module is deep at the intended seam; nonblocking smell candidates are repeated outcome policy across assessment/format/frame validation and repeated installed/candidate evidence primitives, but neither is an actionable defect here.

<details>
<summary>简体中文</summary>

覆盖 17 文件完整 fixed-point diff 与 issue #3664；显式 update 的既有内部仅抽样到足以证明 check_update 不可达 mutation。最终 GO：无 P0–P3，exact-head test/audit 均绿。旧 integrity/env P2 不维持：当前只有只读 discovery、无安装 consumer，README 明确要求未来 updater 同时复验 archive integrity 与 extracted manifest compatibility；环境继承至多是 defense-in-depth。只发现不升级边界守住。预发布/畸形/compatibility 缺失均正确 fail-closed；build metadata 候选经发布与开发归档流水线可达性挑战后撤回。无第二套 Host discovery 权威。

</details>

packages/cli/src/runtime-host-update-discovery.ts
M4n5ter: <details open>
<summary>English</summary>

Thanks. We considered an environment allowlist and are not adopting it here.

The registry authority is pinned by the command-line `--registry` argument. Proxy, TLS, and user npm configuration belong to the same local principal running Maka and are also required in legitimate managed or corporate environments. An actor able to control this process environment can already control `PATH` and the resolved `npm` executable, so filtering a few `NPM_CONFIG_*` variables would not establish a meaningful security boundary; it would mainly make supported proxy setups less reliable.

This operation is a bounded, read-only metadata lookup, and any later mutation must independently verify the pinned archive integrity. We therefore treat the inherited environment as intentional caller configuration rather than a current finding and resolve this thread without a code change.

</details>

<details>
<summary>简体中文</summary>

感谢指出。我们评估了 env allowlist，但决定本 PR 不采用。

registry authority 已由命令行 `--registry` 参数固定。proxy、TLS 与用户 npm 配置属于运行 Maka 的同一本地 principal，也是在企业或受管网络中正常工作的必要配置。能够控制当前进程环境的主体也已经能够控制 `PATH` 与实际解析到的 `npm` executable，因此过滤少量 `NPM_CONFIG_*` 并不能建立有意义的安全边界，主要结果反而是降低合法代理环境的可靠性。

当前操作只是有界、只读的 metadata 查询；任何后续 mutation 仍必须独立校验被钉住的 archive integrity。因此我们把环境继承视为调用方的有意配置，而不是当前 finding，不修改代码并 resolve。

</details>

packages/runtime-host/src/operator/service-management-frame.ts
M4n5ter: <details open>
<summary>English</summary>

Thanks. We agree that prerelease intent must remain visible, but the current frame already carries it without another field.

The frame preserves the original selector (`latest`, `next`, or an exact version), and `candidate.version` is the complete canonical release version, including any prerelease identifiers. Selecting `next` or an exact prerelease is explicit caller intent. A downstream policy can therefore distinguish the release line from the existing facts; adding `prerelease: boolean` or another dist-tag field would create a second representation that can disagree with the selector or version.

We are keeping one version and selector authority and resolving this thread without a protocol change.

</details>

<details>
<summary>简体中文</summary>

感谢指出。我们认同 prerelease 意图必须保持可见，但当前 frame 已经完整携带这些信息，不需要新增字段。

frame 保留了原始 selector（`latest`、`next` 或精确版本），`candidate.version` 也是完整的规范 release version，其中包含所有 prerelease identifier。选择 `next` 或精确 prerelease 本身就是调用方的显式意图。下游策略可以直接依据现有事实区分发布线；再增加 `prerelease: boolean` 或另一份 dist-tag 字段，只会产生可能与 selector 或 version 不一致的第二种表示。

因此这里保持单一 version 与 selector authority，不修改 protocol 并 resolve。

</details>

packages/cli/src/runtime-host-update-discovery.ts
M4n5ter: <details open>
<summary>English</summary>

Thanks. We re-evaluated this against the discovery contract and do not consider it an open defect in this PR.

`unattended_update` is the compatibility-policy result: the candidate is newer and its registry-published compatibility value matches the installed package. It is not a claim that downloaded bytes have already been verified. `candidate.integrity` pins the exact artifact that a mutation layer must acquire and verify; dropping it here would force that layer to perform another mutable lookup.

This PR remains read-only, and its documentation explicitly requires the updater to verify the archive against this SRI and re-read compatibility from the extracted manifest before mutation. That mutation boundary is tracked in #3679. We are therefore keeping the candidate evidence and resolving this thread without a code change.

</details>

<details>
<summary>简体中文</summary>

感谢指出。我们结合 discovery contract 重新裁决后，认为这不是本 PR 中仍然成立的缺陷。

`unattended_update` 表示兼容性策略结论：候选版本更新，并且 registry 发布的 compatibility 与已安装 package 匹配；它并不声称下载字节已经完成校验。`candidate.integrity` 用于钉住后续 mutation 层必须获取并校验的精确 artifact；如果在这里删除，mutation 层反而需要再次查询可变 registry 状态。

本 PR 仍然严格只读，文档也明确要求 updater 在 mutation 前按该 SRI 校验 archive，并从解包后的 manifest 重读 compatibility。该 mutation 边界由 #3679 跟踪。因此这里保留候选证据，不修改代码并 resolve。

</details>

Astro-Han (approved):
Approving. All four points I raised are answered, and I checked the one your answers rest on rather than taking it on trust.

The integrity question turns on whether `unattended_update` is documented as a compatibility verdict rather than a bytes-verified one. It is: `packages/cli/README.md:137-138` states that an updater must still verify the downloaded archive against the integrity and confirm the compatibility value from the extracted package manifest, and `README.zh-CN.md:128-129` says the same. With that written down, carrying `integrity` as the pin for the mutation layer — instead of gating discovery on it — is the right split, and a second lookup in that layer would be strictly worse.

On the npm environment: `--registry` is a command-line argument, so it wins over `NPM_CONFIG_REGISTRY` regardless of what the caller exports, and pinning `cwd` to `homedir()` closes the project-local `.npmrc` path. The residual is a same-principal concern, and an allowlist would break supported proxy setups without moving the boundary.

On prerelease visibility: the selector plus the full canonical version already carry the release line. A `prerelease` flag would be a second representation that can disagree with them.

`test` and `audit` are both green on `3e618aaf`.

<details>
<summary>中文</summary>

批准。四条都答复了，其中最关键的一条我复核了它依赖的判据：`unattended_update` 是否被写明只是兼容性结论、不代表字节已验证。`README.md:137-138` 和 `README.zh-CN.md:128-129` 都明确要求更新程序自行用 integrity 校验下载的 archive 并从解压后的 manifest 再确认兼容性值——写下来了，那么把 integrity 当作交给变更层的锚点、而不是在发现阶段设卡，就是对的划分。

npm 环境那条：`--registry` 是命令行参数，压得过 `NPM_CONFIG_REGISTRY`；`cwd` 固定到 `homedir()` 也堵掉了项目级 `.npmrc`。剩下的属于同一本地主体的范畴。prerelease 那条：selector 加完整版本号已经带了发布线，再加布尔字段反而多一份可能互相矛盾的表述。

`3e618aaf` 上 `test` 与 `audit` 均为绿。
</details>

Astro-Han:
LGTM — merging.

<details>
<summary>中文</summary>

看过了，合并。
</details>

packages/cli/src/runtime-host-update-discovery.ts
```diff
@@ -0,0 +1,360 @@
...
+}
+
+interface NpmViewResult {
+  readonly exitCode: number;
+  readonly stdout: string;
+}
+
+function runNpmView(args: readonly string[]): Promise<NpmViewResult> {
+  return new Promise((resolve, reject) => {
+    const child = spawn('npm', args, {
```
jackwener, line 250: [P2] Isolate registry lookup from ambient npm configuration

This call inherits `PATH` and all npm configuration. `--registry` selects the URL, but inherited proxy/CA/TLS configuration (or a replacement `npm` executable found through `PATH`) can alter the response. A response containing an arbitrary newer version, any well-formed SHA-512 string, and a matching compatibility value reaches `unattended_update`; the integrity check only validates encoding and no archive is fetched here. That lets an untrusted parent environment make this command report a target as eligible for unattended update.

Invoke a fixed trusted npm path with a minimal environment (including a controlled npm userconfig and TLS/proxy settings), or fetch the registry metadata with a pinned client. Add a regression test that supplies hostile PATH/npm configuration and asserts no unattended result.

jackwener (commented):
I found one issue that needs correction before this can safely advertise unattended updates.

PR #3669, opened 2026-08-24
fix(desktop): snapshot linked children for side conversations

### LINKED ISSUE #3654: bug(desktop): Side Conversation creation fails for sessions with linked child agents
## What happened

On current `main`, opening a Side Conversation from a settled Desktop Session fails when the retained source history contains linked subagent / Agent Graph child Sessions.

The UI shows only the generic message:

```text
Could not open the side chat. Please try again.
```

The underlying Runtime Host error is:

```text
operation_unavailable:
Ordinary branches cannot share linked child Session ownership with their source
```

This still fails after the root Turn and all child agents have completed. Retrying later therefore does not resolve this path.

Verified against `main` at `c79e9eb40ee4a0fcd204fb075427d635989eaaca`.

## How to reproduce

1. Build and launch Desktop from current `main`.
2. Start a normal conversation.
3. Run a workflow that creates linked child Sessions, for example an Agent Graph/code-review flow with multiple subagents.
4. Wait until the root Turn and child agents have all reached terminal state.
5. Open **Side Chat** from the titlebar/workbar, or invoke `/side`.
6. Observe that the Side Conversation is not created and the generic setup-error banner appears.

A normal settled conversation without linked child Sessions can create the Side Conversation successfully.

## Expected behavior

A Side Conversation should open from the latest completed source Turn and use the retained parent history as reference-only context, without taking ownership of or controlling the source Session's child agents.

## Actual behavior and root cause

The renderer correctly marks the request as a Side Conversation:

- `ensureCompanionFork()` calls `branchFromTurn(..., { sideConversation: true })`:
  https://github.com/apache/maka/blob/c79e9eb40ee4a0fcd204fb075427d635989eaaca/apps/desktop/src/renderer/features/workbar/tools/side-chat/quote-companion-core.ts#L250-L255

The Desktop IPC adapter does not carry that intent into Runtime Host:

- it calls `copySession("branch", ...)` without a Side Conversation purpose;
- `sideConversation` is used only for cleanup ownership and for adding `mode:side_conversation` metadata after the copy succeeds:
  https://github.com/apache/maka/blob/c79e9eb40ee4a0fcd204fb075427d635989eaaca/apps/desktop/src/main/runtime-host-session-execution-ipc-main.ts#L598-L633

The Host protocol has no field that distinguishes a Side Conversation from an ordinary branch:

- https://github.com/apache/maka/blob/c79e9eb40ee4a0fcd204fb075427d635989eaaca/packages/runtime-host/src/protocol/session-revision.ts#L38-L43

Runtime Host therefore applies the ordinary-branch ownership rule and rejects the copy:

- https://github.com/apache/maka/blob/c79e9eb40ee4a0fcd204fb075427d635989eaaca/packages/runtime-host/src/server/session-revision-graph-references.ts#L91-L95

The renderer then catches every creation exception and collapses it to `fork_setup_failed`, hiding the actionable Host code:

- https://github.com/apache/maka/blob/c79e9eb40ee4a0fcd204fb075427d635989eaaca/apps/desktop/src/renderer/features/workbar/tools/side-chat/quote-companion-core.ts#L250-L257

The ordinary-branch rejection is a valid safety invariant: an independent branch must not share child Session ownership with its source. The defect is that Side Conversation currently reaches that path as an ordinary branch even though its intended semantics are an ephemeral, reference-only context copy.

## Suggested fix

Introduce an explicit Host-recognized copy purpose or operation for Side Conversations, rather than inferring the behavior from a label added after publication. For example:

```ts
type SessionConversationCopyPurpose =
  | "ordinary"
  | "side_conversation";
```

The exact protocol shape is open to discussion, but the intent should:

1. travel from Renderer → Desktop IPC → Runtime Host;
2. be included in the idempotency fingerprint;
3. be persisted with the conversation-copy publication metadata;
4. cause the Host to atomically apply the Side Conversation label/semantics.

For retained linked-child results, the Side Conversation path should create an **ownership-free historical snapshot**:

- retain the terminal child result's user-visible status, summary, and bounded output needed as model context;
- do not retain control/ownership identifiers such as the source `childSessionId`, `runId`, or `resumedFromRunId`;
- copy and rewrite any artifact content that must remain accessible from the target Session;
- cover typed message results, RuntimeEvent function responses, archived tool results, and any copied ledger projections;
- never clone, reparent, delete, resume, or otherwise control the source child Session.

If a retained Graph, child Session, or child result is still live, return the existing retryable `session_busy` result. Missing or inconsistent terminal provenance should continue to fail closed.

Ordinary branches should keep their current rejection behavior, and revisions should keep their existing validated terminal-reference behavior.

## Error-handling follow-up

The direct bug requires the Host copy semantics above; retrying in the renderer alone cannot fix a permanent `operation_unavailable` result.

The Desktop path should additionally preserve structured failure codes:

- `session_busy`: show a waiting/retryable state and retry when the source becomes idle;
- `operation_unavailable`: show a specific unsupported-context error;
- persistence/internal failures: show the generic setup error.

A later successful fork should also clear any previous setup banner.

There is a separate edge case in which `latestSettledTurnId()` documents “start empty” when no completed Turn exists, while the current caller returns `fork_setup_failed`. That can be handled alongside this work or tracked separately.

## Acceptance criteria

- A settled source Session containing one or more terminal linked child Sessions can create a Side Conversation.
- The created Side Conversation contains useful historical child-result context without retaining child ownership/control references.
- Removing the source Session or the Side Conversation cannot incorrectly remove child Sessions owned by the other.
- Ordinary branches with linked children remain rejected.
- Revisions preserve their current validated terminal-reference behavior.
- A live source Graph or child returns a structured, retryable `session_busy`.
- Copy retries remain idempotent, and reusing the same target identity with a different copy purpose is rejected.
- Failed or interrupted creation leaves no visible/preparing orphan Session.
- A Runtime Host integration test covers at least one source Session with a terminal Agent Graph child and verifies successful Side Conversation creation plus ownership isolation.
- Renderer tests verify error-code preservation and clearing a stale setup banner after success.

## Related work

- #2428 introduced generic Side Conversations.
- #1664 established Host-owned conversation copy and the linked-child fail-closed rule.
- #2039 added validated terminal Agent Graph reference retention for revisions while continuing to reject ordinary branches.
- #1366 defines linked child Session ownership and lifecycle.
- #2060 / #2061 are a useful precedent for fixing a conversation-copy authority rejection at the typed Host seam while also correcting misleading UI error classification.

No existing public Maka issue, PR, or Discussion was found that tracks this exact Side Conversation + linked-child integration failure.

Image

liuxiaocs7:
## Summary

- carry an explicit Side Conversation intent from the renderer through Desktop IPC and the Runtime Host protocol
- snapshot terminal linked-child results without retaining child-session ownership or control identifiers
- let the main Session and a Side Conversation run independently: the side chat forks the latest settled turn even while a newer parent turn is active
- validate only Graphs retained by that historical slice, so newer parent Graph work does not block the snapshot
- keep eager setup stable across source-session projection refreshes, preventing the periodic “Preparing side chat…” flicker
- copy and rewrite referenced artifacts, drop child task-ledger ownership, and preserve ordinary branch/revision safety rules

Closes #3654

## Verification

- 32 targeted Desktop Side Conversation, cleanup, disposal, panel-state, copy-attempt, and revision tests passed
- 19 targeted Runtime Host Graph-reference, protocol, and two-client UDS integration tests passed
- integration coverage proves ordinary branches remain `session_busy` while a Side Conversation can snapshot `turn-2` during a newer active turn, and that the active turn is not copied
- Graph coverage proves retained live Graphs still block, while a newer unrelated live Graph does not
- typechecked `@maka/runtime-host` and all Desktop targets after rebuilding workspace dependencies
- ran Biome across all changed files and `git diff --check origin/main...HEAD`
- rebased onto `origin/main@d77e854f7`

before
image

after
image

## AI use

Select exactly one:

Tool(s) and scope: Codex contributed implementation, tests, verification, and review; Claude performed an additional review of the completed change.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/src/renderer/features/workbar/tools/side-chat/use-quote-companion.ts
+++ b/apps/desktop/src/renderer/features/workbar/tools/side-chat/use-quote-companion.ts
@@ -151,8 +151,11 @@ export function useQuoteCompanion(input: UseQuoteCompanionInput): UseQuoteCompan
   // A created fork is hidden immediately, but is not considered usable until
   // onForkCommitted promotes it.
   const pendingForkIdRef = useRef<string | null>(null);
+  const sourceSessionRef = useRef(sourceSession);
+  sourceSessionRef.current = sourceSession;
+  const sourceSessionId = sourceSession?.id;
   const sourceSessionIdRef = useRef(sourceSession?.id);
-  sourceSessionIdRef.current = sourceSession?.id;
+  sourceSessionIdRef.current = sourceSessionId;
   const forkSetupPromiseRef = useRef<Promise<EnsureCompanionForkResult> | null>(null);
   const stopRequestedRef = useRef(false);
   const activeTurnIdRef = useRef<string | null>(null);
@@ -272,7 +275,8 @@ export function useQuoteCompanion(input: UseQuoteCompanionInput): UseQuoteCompan
       const existing = companionRef.current;
       if (existing) return Promise.resolve({ status: 'ready', session: existing });
       if (forkSetupPromiseRef.current) return forkSetupPromiseRef.current;
-      if (!sourceSession) {
+      const currentSourceSession = sourceSessionRef.current;
+      if (!currentSourceSession) {
         return Promise.resolve({ status: 'error', code: 'fork_setup_failed' });
       }
 
@@ -280,7 +284,7 @@ export function useQuoteCompanion(input: UseQuoteCompanionInput): UseQuoteCompan
       if (showPreparing) setPreparing(true);
       const promise = ensureCompanionFork({
         api: sideChat,
-        sourceSession,
+        sourceSession: currentSourceSession,
         panelId,
         name,
         isDisposed: () => !mountedRef.current,
@@ -326,15 +330,15 @@ export function useQuoteCompanion(input: UseQuoteCompanionInput): UseQuoteCompan
       forkSetupPromiseRef.current = promise;
       return promise;
     },
-    [commitFork, mountedRef, panelId, sideChat, sourceSession],
+    [commitFork, mountedRef, panelId, sideChat],
   );
 
   useEffect(() => {
-    if (sourceSession) void ensureFork(copyRef.current.defaultName);
-  }, [ensureFork, sourceSession]);
+    if (sourceSessionId) void ensureFork(copyRef.current.defaultName);
+  }, [ensureFork, sourceSessionId]);
 
   useEffect(() => {
-    if (!sourceSession || !forkRetryPending) return;
+    if (!sourceSessionId || !forkRetryPending) return;
     let retrying = false;
     const retry = () => {
       if (retrying || !mountedRef.current || companionRef.current) return;
@@ -350,7 +354,7 @@ export function useQuoteCompanion(input: UseQuoteCompanionInput): UseQuoteCompan
     };
     const unsubscribe = sideChat.subscribeSessionChanges((event) => {
       if (
-        event.sessionId === sourceSession.id &&
+        event.sessionId === sourceSessionId &&
         (event.reason === 'turn-status-change' ||
           event.reason === 'status-change' ||
           event.reason === 'message-appended')
@@ -363,7 +367,7 @@ export function useQuoteCompanion(input: UseQuoteCompanionInput): UseQuoteCompan
       globalThis.clearInterval(retryTimer);
       unsubscribe();
     };
-  }, [ensureFork, forkRetryPending, mountedRef, sideChat, sourceSession]);
+  }, [ensureFork, forkRetryPending, mountedRef, sideChat, sourceSessionId]);
 
   // The fork is ephemeral (用完即弃): when the panel is dismissed — 退出,
   // switching source session — unsubscribe and remove the fork so it never
--- a/apps/desktop/src/renderer/features/workbar/tools/side-chat/use-quote-companion.ts
+++ b/apps/desktop/src/renderer/features/workbar/tools/side-chat/use-quote-companion.ts
@@ -51,15 +51,15 @@ import {
   performCompanionTurn,
   type CompanionErrorCode,
   type EnsureCompanionForkResult,
-} from './quote-companion-core';
+} from './quote-companion-core.js';
 import { mergeSettledMessages } from '../../../../settled-message-merge.js';
 import { getDesktopConversationCopy } from '../../../../locales/conversation-copy.js';
 import {
   snapshotCompanionQuotes,
   type CompanionQuoteSnapshot,
   type StagedCompanionQuote,
-} from './quote-companion-panel-state';
-import type { CompanionForkVisibilityEvent } from './quote-companion-visibility';
+} from './quote-companion-panel-state.js';
+import type { CompanionForkVisibilityEvent } from './quote-companion-visibility.js';
 
 export interface UseQuoteCompanionInput {
   /** Stable owner for the currently mounted panel generation. */
@@ -178,6 +178,7 @@ export function useQuoteCompanion(input: UseQuoteCompanionInput): UseQuoteCompan
   );
   const [hasContent, setHasContent] = useState(false);
   const [error, setError] = useState<string | null>(null);
+  const [forkRetryPending, setForkRetryPending] = useState(false);
   // Bumped whenever the own-turn set changes so the render picks up the new
   // filter result (the set lives in a ref to stay stable for the event handler).
   const [, setOwnTurnTick] = useState(0);
@@ -298,9 +299,19 @@ export function useQuoteCompanion(input: UseQuoteCompanionInput): UseQuoteCompan
       })
         .then((result) => {
           if (result.status === 'ready' && mountedRef.current) {
+            setForkRetryPending(false);
+            setError(null);
             commitFork(result.session);
           } else if (result.status === 'error' && mountedRef.current) {
-            setError(copyRef.current.errors.forkSetupFailed);
+            setForkRetryPending(result.code === 'fork_source_busy');
+            const errors = copyRef.current.errors;
+            setError(
+              result.code === 'fork_source_busy'
+                ? errors.forkSourceBusy
+                : result.code === 'fork_unsupported'
+                  ? errors.forkUnsupported
+                  : errors.forkSetupFailed,
+            );
           }
           return result;
         })
@@ -317,6 +328,38 @@ export function useQuoteCompanion(input: UseQuoteCompanionInput): UseQuoteCompan
   useEffect(() => {
     if (sourceSession) void ensureFork(copyRef.current.defaultName);
   }, [ensureFork, sourceSession]);
+
+  useEffect(() => {
+    if (!sourceSession || !forkRetryPending) return;
+    let retrying = false;
+    const retry = () => {
+      if (retrying || !mountedRef.current || companionRef.current) return;
+      retrying = true;
+      const currentSetup = forkSetupPromiseRef.current;
+      void (async () => {
+        if (currentSetup) await currentSetup;
+        if (!mountedRef.current || companionRef.current) return;
+        await ensureFork(copyRef.current.defaultName);
+      })().finally(() => {
+        retrying = false;
+      });
+    };
+    const unsubscribe = sideChat.subscribeSessionChanges((event) => {
+      if (
+        event.sessionId === sourceSession.id &&
+        (event.reason === 'turn-status-change' ||
+          event.reason === 'status-change' ||
+          event.reason === 'message-appended')
+      ) {
+        retry();
+      }
+    });
+    const retryTimer = globalThis.setInterval(retry, 2_000);
+    return () => {
+      globalThis.clearInterval(retryTimer);
+      unsubscribe();
+    };
+  }, [ensureFork, forkRetryPending, mountedRef, sideChat, sourceSession]);
 
   // The fork is ephemeral (用完即弃): when the panel is dismissed — 退出,
   // switching source session — unsubscribe and remove the fork so it never
@@ -427,6 +470,8 @@ export function useQuoteCompanion(input: UseQuoteCompanionInput): UseQuoteCompan
         const errors = copyRef.current.errors;
         const byCode: Record<CompanionErrorCode, string> = {
           fork_setup_failed: errors.forkSetupFailed,
+          fork_source_busy: errors.forkSourceBusy,
+          fork_unsupported: errors.forkUnsupported,
           send_failed: errors.sendFailed,
           send_rejected: errors.sendRejected,
         };
--- a/apps/desktop/src/renderer/features/workbar/tools/side-chat/use-quote-companion.ts
+++ b/apps/desktop/src/renderer/features/workbar/tools/side-chat/use-quote-companion.ts
@@ -265,7 +265,10 @@ export function useQuoteCompanion(input: UseQuoteCompanionInput): UseQuoteCompan
   );
 
   const ensureFork = useCallback(
-    (name: string): Promise<EnsureCompanionForkResult> => {
+    (
+      name: string,
+      options: { readonly showPreparing?: boolean } = {},
+    ): Promise<EnsureCompanionForkResult> => {
       const existing = companionRef.current;
       if (existing) return Promise.resolve({ status: 'ready', session: existing });
       if (forkSetupPromiseRef.current) return forkSetupPromiseRef.current;
@@ -273,7 +276,8 @@ export function useQuoteCompanion(input: UseQuoteCompanionInput): UseQuoteCompan
         return Promise.resolve({ status: 'error', code: 'fork_setup_failed' });
       }
 
-      setPreparing(true);
+      const showPreparing = options.showPreparing ?? true;
+      if (showPreparing) setPreparing(true);
       const promise = ensureCompanionFork({
         api: sideChat,
         sourceSession,
@@ -317,7 +321,7 @@ export function useQuoteCompanion(input: UseQuoteCompanionInput): UseQuoteCompan
         })
         .finally(() => {
           forkSetupPromiseRef.current = null;
-          if (mountedRef.current) setPreparing(false);
+          if (showPreparing && mountedRef.current) setPreparing(false);
         });
       forkSetupPromiseRef.current = promise;
       return promise;
@@ -339,7 +343,7 @@ export function useQuoteCompanion(input: UseQuoteCompanionInput): UseQuoteCompan
       void (async () => {
         if (currentSetup) await currentSetup;
         if (!mountedRef.current || companionRef.current) return;
-        await ensureFork(copyRef.current.defaultName);
+        await ensureFork(copyRef.current.defaultName, { showPreparing: false });
       })().finally(() => {
         retrying = false;
       });
--- a/packages/runtime-host/src/protocol/session-revision.ts
+++ b/packages/runtime-host/src/protocol/session-revision.ts
@@ -17,7 +17,13 @@
  * under the License.
  */
 
-import { requireCount, requireEntityId, requireExactRecord, requireRecord } from './codec.js';
+import {
+  requireCount,
+  requireEntityId,
+  requireExactRecord,
+  requireRecord,
+  requireShapedRecord,
+} from './codec.js';
 import { invalidProtocolFrame } from './errors.js';
 import { defineOperation } from './operation-spec.js';
 import { decodeSessionCatalogItem, type SessionCatalogItem } from './session-catalog.js';
@@ -40,6 +46,7 @@ export interface SessionConversationCopyInput {
   readonly targetSessionId: string;
   readonly sourceTurnId: string;
   readonly expectedSourceRevision: number;
+  readonly intent?: 'side_conversation';
 }
 
 export type SessionConversationCopyResult =
@@ -82,7 +89,7 @@ export const SESSION_REVISION_OPERATION_SPECS = {
     mode: 'command',
     availability: 'ready',
     errors: SESSION_COPY_ERRORS,
-    decodeInput: decodeSessionConversationCopyInput,
+    decodeInput: decodeSessionRevisionCopyInput,
     decodeOutput: decodeSessionConversationCopyResult,
     assertOutputForInput: assertConversationCopyOutput,
   }),
@@ -104,6 +111,14 @@ export const SESSION_REVISION_OPERATION_SPECS = {
   }),
 } as const;
 
+function decodeSessionRevisionCopyInput(value: unknown): SessionConversationCopyInput {
+  const input = decodeSessionConversationCopyInput(value);
+  if (input.intent !== undefined) {
+    throw invalidProtocolFrame('Session revision copy does not support an intent');
+  }
+  return input;
+}
+
 function decodeSessionRevisionAbandonInput(value: unknown): SessionRevisionAbandonInput {
   const input = requireExactRecord(value, 'Session revision abandon input', ['targetSessionId']);
   return { targetSessionId: requireEntityId(input.targetSessionId, 'targetSessionId') };
@@ -124,16 +139,19 @@ function decodeSessionRevisionAbandonResult(value: unknown): SessionRevisionAban
 }
 
 export function decodeSessionConversationCopyInput(value: unknown): SessionConversationCopyInput {
-  const input = requireExactRecord(value, 'Session conversation-copy input', [
-    'sourceSessionId',
-    'targetSessionId',
-    'sourceTurnId',
-    'expectedSourceRevision',
-  ]);
+  const input = requireShapedRecord(
+    value,
+    'Session conversation-copy input',
+    ['sourceSessionId', 'targetSessionId', 'sourceTurnId', 'expectedSourceRevision'],
+    ['intent'],
+  );
   const sourceSessionId = requireEntityId(input.sourceSessionId, 'sourceSessionId');
   const targetSessionId = requireEntityId(input.targetSessionId, 'targetSessionId');
   if (sourceSessionId === targetSessionId) {
     throw invalidProtocolFrame('Session conversation copy requires distinct Sessions');
+  }
+  if (input.intent !== undefined && input.intent !== 'side_conversation') {
+    throw invalidProtocolFrame('Invalid Session conversation-copy intent');
   }
   return {
     sourceSessionId,
@@ -143,6 +161,7 @@ export function decodeSessionConversationCopyInput(value: unknown): SessionConve
       input.expectedSourceRevision,
       'expected source Session revision',
     ),
+    ...(input.intent === 'side_conversation' ? { intent: input.intent } : {}),
   };
 }
 
--- a/packages/runtime/src/conversation-copy.ts
+++ b/packages/runtime/src/conversation-copy.ts
@@ -88,6 +88,11 @@ export type ConversationCopyArtifactReferenceMap =
       readonly linkedChildren:
         | { readonly mode: 'reject' }
         | {
+            readonly mode: 'snapshot';
+            readonly archivedResults: ReadonlyMap<string, string>;
+            readonly artifactIds: ReadonlyMap<string, string>;
+          }
+        | {
             readonly mode: 'preserve_validated';
             readonly references: ReadonlyMap<string, ConversationCopyExternalChildReferences>;
           };
@@ -530,6 +535,20 @@ export function archivedToolResultContainsConversationOwnedReferences(
     );
   }
   return false;
+}
+
+export function archivedToolResultContainsLinkedChildReferences(serializedResult: string): boolean {
+  const value = deserializeToolResultArchive(serializedResult);
+  if (isArchivedToolResultPlaceholder(value)) return false;
+  try {
+    return (
+      conversationCopyLinkedChildReferences(
+        decodePersistedToolResultContent(markPersisted<ToolResultContent>(value)),
+      ).length > 0
+    );
+  } catch {
+    return false;
+  }
 }
 
 export function conversationCopyLinkedChildReferences(
@@ -1010,9 +1029,11 @@ function rewriteRuntimeEventReferences(
               }
             : {}),
           ...(event.refs.artifactId
-            ? {
-                artifactId: rewriteOwnedArtifactId(event.refs.artifactId, references),
-              }
+            ? archivedSnapshotResult(event.refs.artifactId, references) !== undefined
+              ? {}
+              : {
+                  artifactId: rewriteOwnedArtifactId(event.refs.artifactId, references),
+                }
             : {}),
           ...(event.refs.sourceInvocationId
             ? {
@@ -1111,6 +1132,8 @@ function rewriteToolResultContent(
     return { ...content, ref: rewriteStorageRef(content.ref, references) };
   }
   if (content.kind === 'archived_tool_result') {
+    const snapshot = rewriteArchivedSnapshot(content, references);
+    if (snapshot) return snapshot;
     return {
       ...content,
       runtimeEventId: rewriteOwnedId(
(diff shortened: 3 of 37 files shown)


zhiiw (commented):
Reviewed at exact head `e078acdbcbeb73bb51c3ead08580659964ea8911` (36 files, +1550/-146, MERGEABLE).

**No findings (P0–P3).** The three questions this change most needs answered all check out:

**1. When does the snapshot happen?** Inside the existing admission lease, which covers source + target + revision root plus every linked-child lane (`agentGraphRevisionAdmissionSessionIds` feeds the `retry_admission` pass before `prepareAgentGraphRevisionReferences` runs). While that lease is held, a retained linked child that is still active fails with `session_busy`, and a retained live Agent Graph fails the same way (`session-revision-graph-references.ts`). The snapshot therefore cannot read a mid-flight child mutation.

**2. Can the copy drift from the source afterwards?** By design it neither tracks nor aliases: snapshot mode strips `childSessionId` / `runId` / `resumedFromRunId` from subagent results and swarm items, rewrites child artifact references to copies owned by the target session, rehydrates archived linked-child results with ownership removed, and drops `child_agent` ownership in the copied task ledger (`task-ledger-store.ts`). The result is reference-only frozen context, exactly what a side conversation wants.

**3. Can a linked child be missed by enumeration?** `directChildren` not represented in the copied slice are caught explicitly: active → `session_busy`, terminal-but-unrepresented → fail closed with `operation_unavailable`. A child created after the source turn is refused rather than silently dropped.

The ordinary paths are preserved by explicit gates: `intent` is rejected by the revision operation's decoder (`decodeSessionRevisionCopyInput`), `kind !== 'side_conversation'` keeps the old conversation-owned-reference hard failure for branch/revision, the fingerprint namespace is versioned (`v1` vs `v2`) so a side-conversation retry can never resolve against an ordinary branch's in-flight copy, and the wire change is epoch-gated (44 → 45; current main is 44, no collision).

**Executed evidence (real Windows x64, this head):** runtime-host `session-revision-protocol` + `session-revision-graph-references` 17/17; storage `artifact-store` + `task-ledger-authority` 46/46 (5 skips are platform gates); desktop `quote-companion-{retry,cleanup,disposal}` + `runtime-host-session-execution-ipc-main` 28/28; runtime `conversation-copy` 10/15 — the 5 failures are `EBUSY: unlink runtime.sqlite` in temp dirs and reproduce **identically on the PR's base commit (`84ed9a31`) on this machine**, i.e. a pre-existing Windows-only suite flake, not this PR. All four new side-conversation snapshot tests pass. (Non-blocking note for the repo, not this PR: the runtime conversation-copy suite has never been green on a real Windows developer machine.)

**Checks:** path-filtered as expected — only `test` applies to this diff (no manifest/lockfile/sandbox paths), `completed/success` on the exact head. `audit` correctly absent.

<details>
<summary>简体中文</summary>

零 finding。三个关键问题的答案：快照发生在覆盖源/目标/修订根/全部 linked child 通道的 admission lease 内，运行中的 child 或存活 graph 会以 session_busy 拒绝；快照剥离 childSessionId/runId/resumedFromRunId、把 child artifact 复制进目标命名空间、任务台账丢弃 child_agent 归属，因此快照之后源端再变动不会影响副本（这正是设计意图）；slice 里没引用到的 direct child 会被显式枚举检查拦下（活跃→busy，未表示→fail closed）。普通 branch/revision 路径由 intent 拒绝、指纹 v1/v2 分域、epoch 44→45 三道门保持原样。Windows 真机执行：17/17 + 46/46 + 28/28 绿，conversation-copy 的 5 个失败在 base commit 上同样复现（既有的 Windows EBUSY 测试 flake，与本 PR 无关）。

</details>

Astro-Han (dismissed):
Reviewed at `e078acdb`. No P0–P3. Approving.

The three things worth checking on a snapshot change all hold:
- **When the snapshot is taken** — it runs inside the admission lease covering the source, the target, the revision root and every linked-child channel, so a live graph or an active child resolves to `session_busy` rather than a torn snapshot.
- **Divergence afterwards** — that is the intent, not a gap: ownership ids are stripped, artifacts are copied into the target namespace, and the ledger drops `child_agent` attribution, leaving a reference-only relationship.
- **Enumeration gaps** — `session-revision-graph-references.ts:101-105` explicitly filters children not represented in the request and fails with `session_busy` if any is active. The retained-graph path fails closed too: an unreadable graph state returns `operation_unavailable` rather than proceeding, and ordinary branches are rejected outright before any of this.

`test` is terminal green on this exact head, and it is the only check the path filter should trigger here.

One non-blocking note for the repo rather than this PR: on a real Windows machine the `conversation-copy` suite fails 5 of 15 on `EBUSY` during unlink, and it reproduces identically on the base commit. It is a pre-existing Windows-only flake, unrelated to this change, but it appears never to have been green there.

<details>
<summary>中文</summary>

在 `e078acdb` 上审，无 P0–P3，approve。三个关键点都成立：快照发生在覆盖源、目标、修订根和全部 linked child 通道的 admission lease 之内，活跃 child 或 live graph 会直接 `session_busy`；快照之后父子失同步是设计本身（剥 ownership id、artifact 复制进目标命名空间、台账丢 `child_agent` 归属，只保留引用关系）；枚举缺口由 `session-revision-graph-references.ts:101-105` 显式拦截，未表示且仍活跃的 child 一律 `session_busy`，retained graph 读不到状态时返回 `operation_unavailable`（fail closed），普通 branch 更是提前拒掉。

另有一条与本 PR 无关、供仓库参考的非阻塞信息：真 Windows 机器上 `conversation-copy` 套件 15 条挂 5 条（unlink 时 `EBUSY`），在 base commit 上同样复现，属既有的 Windows-only flake。
</details>

apps/desktop/src/renderer/features/workbar/tools/side-chat/use-quote-companion.ts
```diff
@@ -298,9 +299,19 @@ export function useQuoteCompanion(input: UseQuoteCompanionInput): UseQuoteCompan
       })
         .then((result) => {
           if (result.status === 'ready' && mountedRef.current) {
+            setForkRetryPending(false);
+            setError(null);
             commitFork(result.session);
           } else if (result.status === 'error' && mountedRef.current) {
-            setError(copyRef.current.errors.forkSetupFailed);
+            setForkRetryPending(result.code === 'fork_source_busy');
```
ARE404, line 314: **[P3] Add the CONTRIBUTING-required before/after UI evidence.** This branch introduces user-visible busy/unsupported banners, automatic retry state, and clearing of a stale banner on success (`:300-362`), while the PR body contains no screenshot or recording. The repository standard says UI changes must include before/after screenshots or a recording; the story/test evidence proves behavior but does not show reviewers the rendered states.

packages/runtime-host/src/protocol/session-revision.ts
```diff
@@ -40,6 +46,7 @@ export interface SessionConversationCopyInput {
   readonly targetSessionId: string;
   readonly sourceTurnId: string;
   readonly expectedSourceRevision: number;
+  readonly intent?: 'side_conversation';
```
ARE404, line 49: **[P3] Give copy purpose one discriminated domain representation.** The same concept is currently `sideConversation: true` in Renderer ports, optional `intent` here and in durable/cleanup records, `ConversationCopySemanticKind` in Host, and `mode:side_conversation` as a label. Host then reconstructs a third kind and repeats `kind === 'side_conversation'` through validation, artifact policy, ledger policy, labels, and fingerprinting. A shared `branch | revision | side_conversation` copy mode (with a legacy decode adapter if stored compatibility requires it) would make the interface truthful and centralize the derived label/policy; roughly 20 conversion/comparison branches across six production files become deletable. This is design debt, not a correctness blocker.

packages/runtime/src/conversation-copy.ts
```diff
@@ -87,6 +87,11 @@ export type ConversationCopyArtifactReferenceMap =
       readonly relativePaths: ReadonlyMap<string, string>;
       readonly linkedChildren:
         | { readonly mode: 'reject' }
+        | {
+            readonly mode: 'snapshot';
+            readonly archivedResults: ReadonlyMap<string, string>;
+            readonly artifactIds: ReadonlyMap<string, string>;
```
ARE404: **[P3] Delete the nested `artifactIds` map.** In exact mode the outer object already carries the source→target artifact map at `:86`; Host passes that exact same `artifactCopy.artifactIds` value both there and here. `rewriteSnapshotArtifactIds` can read `references.artifactIds` directly once it has established `linkedChildren.mode === 'snapshot'`. Removing this field eliminates a duplicate representation and prevents the two maps from ever drifting.

ARE404 (commented):
Review at exact head `e078acdbcbeb73bb51c3ead08580659964ea8911` against merge-base `84ed9a31018ce31aee228473af4d03c35def0f3f`. **No P0–P2. Three P3 design/documentation opportunities, inline.** The exact-head `test` check is green.

### Verdict

1. **Problem:** valid. A disposable Side Conversation needs a historical child result, not shared lifecycle/control authority over the source child. The existing ordinary-branch rejection was a correct invariant; the missing distinction was copy purpose.
2. **Mechanism:** addresses the root cause end to end. Purpose reaches the Host and the durable fingerprint; the Host admits the source/target/root/child lanes, refuses live graphs or children, validates terminal run/artifact provenance, copies referenced artifacts, strips child control identities from typed results, removes child task ownership, and publishes only after all ledgers commit.
3. **Design scope:** an ownership-free terminal snapshot is smaller and safer than cloning/reparenting children or weakening the branch invariant. Runtime Host remains the authority, Runtime owns typed reference rewriting, and Storage owns byte/task persistence; those seams pass the deletion test. Epoch 45 correctly closes the incompatible-peer boundary.
4. **Correctness:** I found no missing acceptance criterion or scope creep. I specifically challenged archived-result hydration and opaque `providerOutput`: neither yielded a reachable deterministic failure or ownership leak. Structured busy/unsupported results and the latest-settled-turn retry path are covered.
5. **Optimality:** the behavior is right, but the representation can be made deeper. The inline P3s cover the scattered copy-purpose vocabulary, a duplicate artifact map, and missing UI evidence.

Further mechanical deletion opportunities: classify each archived body once instead of separately deserializing it for ownership, linked-child detection, and reference collection; and move the busy→retry→success transition into a pure companion-core state module so the 186-line Linkedom test can become a small wiring smoke plus a table test (roughly 120–140 harness lines deletable).

<details>
<summary>简体中文</summary>

问题定义成立：Side Conversation 需要的是子会话的历史结果，不应继承对子会话的生命周期/控制权；普通 branch 的拒绝规则本身没有错，缺的是明确的 copy purpose。当前实现从 Renderer 一直把 purpose 送到 Host 和持久化指纹，并在 admission lease 内验证 graph/child/run/artifact 的终态一致性，再复制 artifact、剥离 childSessionId/runId/resumedFromRunId、去掉 child task ownership，所有 ledger 成功后才发布，因此解决的是根因而非 UI 症状。没有 P0–P2。三个 P3 分别是：缺少 CONTRIBUTING 要求的 UI 前后证据；同一 purpose 被 boolean、optional intent、semantic kind、label 多套表示；snapshot 中重复保存同一个 artifact 映射。另可把 archive body 分类集中一次，并把 186 行 Linkedom retry 测试拆成纯状态表测 + 小型 wiring smoke，预计可删除约 120–140 行 harness。

</details>

M4n5ter (dismissed):
<details open>
<summary>English</summary>

Reviewed independently at exact head `e078acdbcbeb73bb51c3ead08580659964ea8911`. **GO: no P0–P2 findings.** I independently retained two non-blocking P3 cleanup/documentation items that already have current-head threads, so I am not duplicating them inline: add the `CONTRIBUTING.md`-required UI screenshot/recording, and delete the nested snapshot `artifactIds` alias because the outer exact reference map already carries the identical source-to-target mapping.

### First-principles verdict

The problem definition is correct. An ordinary branch must not share lifecycle/control authority over linked child Sessions, while a Side Conversation needs terminal child output only as immutable reference context. The existing rejection was the right invariant; the missing distinction was an explicit copy purpose.

The implementation addresses that root cause at the existing authority seam:

- `side_conversation` travels from Renderer through Desktop IPC and the Runtime Host protocol, participates in a versioned idempotency fingerprint, and is persisted before publication;
- the Host admits the source, target, and all retained child lanes before checking terminality and provenance, so live source/child/graph state returns structured `session_busy` instead of producing a torn snapshot;
- snapshot mode removes `childSessionId`, `runId`, and `resumedFromRunId`, copies validated child artifacts into target-owned IDs, rewrites archived linked results inline, and removes child task ownership/run references;
- ordinary branches retain their fail-closed rejection, revisions retain validated-reference semantics, and failed creation does not leave a visible/preparing target;
- Desktop retries a busy source from the newest settled turn using a fresh copy identity and clears the stale error after success.

This is more minimal than cloning/reparenting child Sessions or weakening the ordinary-branch invariant. The 36-file spread is one typed semantic propagated through protocol, storage, runtime, and Desktop boundaries, not a second copy authority. A separate operation would duplicate more of the existing copy stack.

Validation: Core, Storage, Runtime, Runtime Host, MCP, Computer Use, UI, Desktop main, and Desktop preload builds passed. Focused suites completed `114 passed / 2 platform skips / 0 failed` across 116 tests. `git diff --check` and Biome over all changed TS/TSX files passed. The exact-head hosted `test` is terminal green; the PR is OPEN, non-draft, MERGEABLE/clean. Fresh current main `04836d3b8053c68b8d0b5e4101608bf66f5d1020` has zero changed-file overlap with this PR since its base, and synthetic merge tree `0afc7721937880ff8294eb7388a9ad8c1649cfa8` materializes cleanly. All unresolved review threads are current-head P3s; there is no unresolved P0–P2 thread and no stale-head approval.

</details>

<details>
<summary>中文</summary>

在 exact head `e078acdbcbeb73bb51c3ead08580659964ea8911` 上独立审查。**GO：没有 P0–P2。** 我独立保留了两条非阻塞 P3，而且它们已有当前 head 的 thread，因此不重复发 inline：补上 `CONTRIBUTING.md` 要求的 UI 截图/录屏；删除 snapshot 内层的 `artifactIds` 别名，因为外层 exact reference map 已经持有同一个 source→target 映射。

### 第一性原理结论

问题定义是对的：ordinary branch 不能共享 linked child Session 的生命周期/控制 authority；Side Conversation 只需要把 terminal child 输出作为不可变的参考上下文。原来的拒绝规则没有错，缺的是显式 copy purpose。

实现是在现有 authority seam 上解决根因：

- `side_conversation` 从 Renderer 贯穿 Desktop IPC 与 Runtime Host protocol，进入版本化 idempotency fingerprint，并在发布前持久化；
- Host 先锁住 source、target 和全部 retained child lane，再检查终态与 provenance，因此 live source/child/graph 会返回结构化 `session_busy`，不会读出撕裂快照；
- snapshot mode 删除 `childSessionId` / `runId` / `resumedFromRunId`，把验证后的 child artifact 复制为 target-owned ID，inline 重写 archived linked result，并清掉 child task ownership/run reference；
- ordinary branch 继续 fail closed，revision 保留原来的 validated-reference 语义，失败创建不会留下可见/preparing target；
- Desktop 对 busy source 会重读最新 settled turn、使用新的 copy identity 重试，成功后清除旧错误。

这比 clone/reparent child Session 或放宽 ordinary-branch invariant 更小、更安全。36 个文件是在 protocol、storage、runtime、Desktop 边界传播同一个 typed semantic，不是增加第二套 copy authority；另开 operation 反而会复制更多现有 copy stack。

验证：Core、Storage、Runtime、Runtime Host、MCP、Computer Use、UI、Desktop main/preload build 通过；focused tests 共 116 条，`114 pass / 2 platform skip / 0 fail`；全部 changed TS/TSX 文件通过 Biome，`git diff --check` 通过。exact-head hosted `test` 终态绿；PR OPEN、非 draft、MERGEABLE/clean。Fresh current main `04836d3b8053c68b8d0b5e4101608bf66f5d1020` 自 PR base 起与本 PR changed-file 0 overlap，synthetic merge tree `0afc7721937880ff8294eb7388a9ad8c1649cfa8` clean。所有 unresolved review thread 都是当前 head 的 P3；没有 unresolved P0–P2，也没有 stale-head approval。

</details>

packages/runtime-host/src/protocol/session-revision.ts
liuxiaocs7: Agreed that a single discriminated copy-purpose representation would be a useful cleanup. It spans Renderer, Desktop IPC, Host protocol, durable cleanup records, and compatibility decoding, so I am leaving this non-blocking P3 out of the focused bug-fix PR rather than expanding its migration surface. The current representation remains strictly decoded and included in the idempotency fingerprint.

packages/runtime/src/conversation-copy.ts
liuxiaocs7: Addressed in `a21b59922`. Snapshot mode no longer carries a nested `artifactIds` alias; `rewriteSnapshotArtifactIds` now reads the outer exact `references.artifactIds` map after checking the snapshot discriminant, and the focused tests were updated.

packages/runtime/src/conversation-copy.ts
```diff
@@ -1124,12 +1146,21 @@ function rewriteToolResultContent(
...
+    const snapshot = rewriteArchivedSnapshot(content.value, references);
+    if (snapshot) return snapshot;
     return {
       ...content,
       value: rewriteArchivedToolResult(content.value, references),
     };
   }
   if (content.kind === 'subagent') {
+    if (linkedChildrenAreSnapshots(references)) {
+      const { childSessionId: _childSessionId, runId: _runId, ...snapshot } = content;
```
Astro-Han, line 1158: [P3] The snapshot branch strips `runId` unconditionally based on `linkedChildrenAreSnapshots(references)`, but that condition asks whether the *linked children* are snapshots — not whether *this item* has a child session at all.

An item with no `childSessionId` but a source-owned `runId` is a shape this same file explicitly recognises: `archivedToolResultContainsConversationOwnedReferences` at `:520-535` treats `!item.childSessionId && (item.runId !== undefined || item.resumedFromRunId !== undefined || ...)` as a conversation-owned reference. The `agent_swarm` branch at `:1189-1197` has the same problem and additionally drops `resumedFromRunId`.

Consequence: a side snapshot clones fresh AgentRun/Artifact mappings for the source ledger, but these two branches throw away the source's own run identity. The side transcript keeps the summary and artifacts while losing run/resume provenance, and the CLI agent-swarm view loses its run marker. Execution and security are unaffected, and linked-child navigation still works, hence P3.

Suggested direction: branch on `childSessionId`. Only a real linked child should lose child/run/resume identity and use the snapshot artifact map; a source-owned item without `childSessionId` should go through the existing owned-id rewrite (`rewriteLinkedRunId(..., undefined, ...)` plus the owned artifact map). For coverage, `packages/runtime/src/__tests__/conversation-copy.test.ts:549-562` already has a fixture with no `childSessionId` carrying `runId`/`resumedFromRunId` — a side-mode case could reuse it.

<details>
<summary>简体中文</summary>

[P3] snapshot 分支按 `linkedChildrenAreSnapshots(references)` 一刀切地删掉 `runId`，但这个条件判断的是"链接的子会话是不是快照"，而不是"这个 item 有没有子会话"。

没有 `childSessionId` 但带 source-owned `runId` 的 item 是这个文件自己承认的合法形态：`:520-535` 的 `archivedToolResultContainsConversationOwnedReferences` 明确把 `!item.childSessionId && (item.runId !== undefined || item.resumedFromRunId !== undefined || ...)` 判为 conversation-owned reference。`agent_swarm` 分支（`:1189-1197`）同理，还多删了一个 `resumedFromRunId`。

后果：side snapshot 会为 source ledger 克隆新的 AgentRun/Artifact 映射，但这两处直接把 source 自有的 run identity 丢掉，侧会话 transcript 保留了 summary 和 artifact，却没有了 run/resume 来源，CLI 的 agent-swarm 视图也会少掉 run 标记。这不影响执行，也不影响安全，链接子会话的导航同样不受影响，所以是 P3。

建议改成按 `childSessionId` 分支：真正的 linked child 才去掉 child/run/resume identity 并用 snapshot artifact 映射；没有 `childSessionId` 的 source-owned item 走已有的自有 ID 重写路径。测试可以参考 `packages/runtime/src/__tests__/conversation-copy.test.ts:549-562` 那个 fixture 补一个 side-mode 用例。

</details>

Astro-Han (commented):
Walked all 36 files at head `aa669841781090db196b1bff9ad6ea9ee67a1ad2`. No P0-P2; one P3, inline below. CI is still queued, so I am not looking at approval yet.

For context: the two approvals on `e078acdb` were dismissed because they were bound to a stale head. One of them was M4n5ter's own review and was dismissed by our account, which was our process mistake — we have already explained it to him.

<details>
<summary>简体中文</summary>

在当前 head `aa669841781090db196b1bff9ad6ea9ee67a1ad2` 上把 36 个文件都走了一遍，没有 P0–P2，只有下面这一条 P3。CI 目前还在排队，等它绿了再看批准的事。

另外说明一下，之前 `e078acdb` 上的两条 approve 已经因为绑定的 head 过期被撤掉了，其中一条是我们账号误撤了 @M4n5ter 本人的，这点是我们的流程问题，已经跟他说明。

</details>

packages/runtime/src/conversation-copy.ts
liuxiaocs7: 已在 `6551f9b8f` 修复。snapshot 重写现在按 `childSessionId` 区分：真正的 linked child 才会剥离 child/run/resume identity；没有 `childSessionId` 的 source-owned `subagent` / `agent_swarm` 继续通过现有 run/artifact 映射重写。新增了两条 side-mode 回归测试，覆盖 `runId`、`resumedFromRunId` 和 artifact 映射；`conversation-copy` 17/17、Runtime typecheck、Biome 与 `git diff --check` 均通过。

liuxiaocs7:
Updated on top of current main and addressed the latest review follow-ups:

- source-owned items without childSessionId now preserve and rewrite runId and resumedFromRunId; true linked-child snapshots still remove control identity
- the busy message now covers either the main conversation or a linked task
- persisted copy-kind mapping and fingerprint construction are consolidated
- all six AI-authored commits now include Generated-by: Codex
- the rebase conflict with the new main protocol change is resolved by assigning Side Conversation epoch 48 while retaining project-registration epoch 47, with a regression assertion

Validation on head 5f3c31fb6:

- Runtime Host: 1143/1143
- Desktop: 1386/1386
- focused protocol and session-revision tests: 68/68
- root typecheck, lint, format check, Desktop/UI knip: passed
- protocol epoch guard: 47 -> 48 passed
- final Standards review: 0 findings
- final Spec review: 0 findings

The conditional 2-second fallback remains intentional. It starts only after session_busy, stops after success or unmount, and covers linked child or Graph completion that is not guaranteed to arrive through the source Session subscription.

Astro-Han (approved):
Approving at `5f3c31fb6`. CI is terminal green on this exact head (run [32717978917](https://github.com/apache/maka/actions/runs/32717978917)), and I have no blocking findings. One non-blocking note is below.

This was reviewed independently along two tracks and I checked the substantive claims against the code myself. The storage and protocol work holds up: `intent` is an optional field that older records simply omit, the strict decoder accepts it only on a branch copy and rejects it on a revision, and nothing in the copy, rebuild, repair, backfill, or terminal-read paths drops it. Snapshotting a linked child correctly strips child owner and run identity while keeping main-owned task refs, and the archived result is hash- and size-checked before it is inlined, with a rollback across the sidecars and target on failure.

The epoch guard is right, and worth calling out because it is easy to get wrong: the new assertion is `> 47`, so it actually fails if the bump to 48 is reverted, and the comment records why 47 was unavailable. That is the standard the other epoch tests in that file follow.

**[P3] Legacy Quote Companion creation leases stay stuck after upgrade.**

`cleanupOnce` replays the persisted lease verbatim at `apps/desktop/src/main/quote-companion-cleanup.ts:205`. `intent` is new here and optional, so a `phase: "creating"` row written by an earlier Desktop carries none; recovery therefore resumes it as an ordinary branch, the Host applies the linked-child rejection, the row is retained, and the failure repeats on every launch. The new restart test seeds only a current-format lease with `intent: "side_conversation"`, so this path is untested.

To be precise about ownership, because it matters here: **this is not a regression you introduced.** The ordinary-branch rejection already exists at the merge-base in `session-revision-graph-references.ts`, so such a lease was already looping on every launch before this change. What this PR does is fix that loop for newly written leases and leave the old ones where they were. Recognising or migrating a legacy creating lease as a Side Conversation would close it, as would retiring the lease outright once the Host has proven the target absent. A follow-up is fine by me.

Verification on my side, beyond the PR's own CI: merged into current `main` (`8801edc32`, two commits ahead of your base) — clean, with the epoch landing at 48 over main's 47. On that merged tree the `runtime-host` session-revision graph-reference, session-revision protocol, and bootstrap protocol suites pass 67/67, and `storage` artifact-store and task-ledger-authority pass 49/49 with 2 skipped. `@maka/runtime` fails to build in my checkout with five errors, but byte-identically on a clean `main` checkout in the same environment — missing `@slack` packages and an incompatible proxy agent — so none of it is attributable here.

One thing to be aware of when you read CI: `main` is currently red on `Check ASF source headers`, from five plugin files added by #3250. It has nothing to do with this branch, and #3708 addresses it. I would rather land this after `main` is green again so that the merge result is readable — happy to do that once it is, unless you would prefer otherwise.

在 `5f3c31fb6` 上给出 approve。这个 exact head 的 CI 终态绿（运行记录 [32717978917](https://github.com/apache/maka/actions/runs/32717978917)），我没有阻塞性问题。下面有一条不阻塞的说明。

这个 PR 由两条独立线各自复审，我自己也对照代码核过其中的实质论断。存储和协议部分是站得住的：`intent` 是可选字段，旧记录不带它即可；严格解码器只在 branch 复制上接受它、在 revision 上拒绝；copy、rebuild、repair、backfill 和终态读取各条路径都没有把它丢掉。linked child 的快照正确地剥离了 child 的 owner 与 run 身份，同时保留了 main 持有的 task refs；archived result 在内嵌前会校验哈希与大小，失败时会跨 sidecars 和 target 回滚。

epoch 守卫写对了，值得单独提一句，因为这一点很容易写错：新增断言是 `> 47`，因此一旦把提升到 48 的改动回退，它确实会失败；注释里也记录了 47 为何不可用。这正是该文件里其它 epoch 测试遵循的写法。

**[P3] 升级后，旧的 Quote Companion creation lease 会一直卡住。**

`cleanupOnce` 在 `apps/desktop/src/main/quote-companion-cleanup.ts:205` 原样回放持久化的 lease。`intent` 是本次新增且可选的，因此旧版 Desktop 写入的 `phase: "creating"` 记录不带它；恢复时便会把它作为普通 branch 续做，Host 施加 linked-child 拒绝，记录被保留，此后每次启动都重复这次失败。新增的重启测试只写入了带 `intent: "side_conversation"` 的当前格式 lease，因此这条路径没有被覆盖。

归属上我要说准确，因为这里很关键：**这不是你引入的回归。** 普通 branch 的那条拒绝在 merge-base 上就已存在（`session-revision-graph-references.ts`），所以这样的 lease 在本次改动之前就已经每次启动都在循环失败了。本 PR 所做的是为新写入的 lease 修好了这个循环，而把旧的留在了原处。把旧的 creating lease 识别或迁移为 Side Conversation 即可闭合这一点；或者在 Host 已证明目标不存在时直接终止该 lease。放到后续 PR 处理我也没有意见。

除 PR 自身 CI 外，我这边的验证：合入当前 `main`（`8801edc32`，比你的 base 新两个提交）——合并干净，epoch 在 main 的 47 之上落到 48。在合并后的树上，`runtime-host` 的 session-revision graph-reference、session-revision protocol 与 bootstrap protocol 三套 67/67 通过，`storage` 的 artifact-store 与 task-ledger-authority 49/49 通过、2 项跳过。`@maka/runtime` 在我的检出上构建失败、报 5 个错，但在同一环境下干净 `main` 检出上逐字节一致（缺少 `@slack` 相关包以及代理 agent 版本不匹配），因此与本次改动无关。

读 CI 时请注意一点：`main` 目前在 `Check ASF source headers` 上是红的，来自 #3250 新增的五个 plugin 文件。这与本分支无关，#3708 正在处理。我倾向于等 `main` 恢复绿之后再合入本 PR，这样合并结果才可读——届时我很乐意来合，除非你有别的偏好。

</details>

Astro-Han:
LGTM. `test` is green on the current head, and `main` is green again now that the ASF header gap is fixed, so the merge result is readable. The compatibility epoch here is 48 against 47 on `main`, with the matching `> 47` assertion — no conflict. Merging.

<details>
<summary>简体中文</summary>

已确认。当前 head 上 `test` 通过；ASF header 缺失修复后 `main` 也恢复绿色，因此合并结果是可读的。这里的兼容性 epoch 是 48，`main` 上是 47，断言写法 `> 47` 与之匹配，没有冲突。合并。

</details>

PR #3673, opened 2026-08-24
fix(desktop): align the connection base URL gate with the Runtime Host

### LINKED ISSUE #3672: fix(desktop): connection base URL gate is looser than the Runtime Host contract, so a rejected endpoint reports a misleading error
## What happened

The Desktop connection forms accept `baseUrl` values that the Runtime Host will refuse, so a custom relay endpoint containing a query string, a fragment, or embedded credentials is accepted by the form, sent over IPC, and then rejected by the Host. The rejection is not classified, so the user is shown the generic fallback copy:

> 模型连接服务暂时不可用，请稍后重试。
> The model connection service is temporarily unavailable. Try again later.

That message is wrong twice over: the service is available, and retrying can never succeed. Nothing points at the endpoint field, which is the actual problem, and the field is not marked — the error lands on `field: 'form'`.

Two validators disagree about what a connection base URL is:

| Input | Desktop IPC gate (`validateConnectionBaseUrl`) | Runtime Host (`normalizeCatalogConnectionBaseUrl`) |
| --- | --- | --- |
| `https://gw.example.com/v1?api-version=2024-06` | accepted verbatim | rejected — "must not contain a query or fragment" |
| `https://user:pw@gw.example.com/v1` | accepted verbatim | rejected — "must not contain credentials" |
| `https://gw.example.com` | kept as typed | canonicalized to `https://gw.example.com/` |
| `https://Relay.Example.com/V1` | kept as typed | canonicalized to `https://relay.example.com/V1` |

`validateConnectionBaseUrl` (`packages/core/src/llm-connections.ts:602`) checks only length, `URL` parseability, and an http/https scheme allowlist, and `normalizeConnectionBaseUrl` (`:668`) documents "Trim is the only canonicalization". The Host codec (`packages/core/src/runtime-policy/connection-catalog-codec.ts:633`) additionally rejects credentials and query/fragment, canonicalizes through `URL.toString()`, and enforces a 2048-**byte** cap rather than a 2048-character one.

Both write paths use the lenient gate — `connections:create` and `connections:update` each route through `normalizeConnectionBaseUrlForIpc` (`apps/desktop/src/main/connections-ipc-validation.ts:111`, and `normalizeUpdateInput` at `apps/desktop/src/main/runtime-host-connections-ipc-main.ts:471`) — so both adding a relay and editing an existing endpoint fail the same way.

The client-side form gate does not narrow this either: `validateAddProviderDraft` (`apps/desktop/src/renderer/settings/provider-add-submission.ts:89`) only checks that a required base URL is non-empty.

## How to reproduce

1. Open Desktop → Settings → Models → Add provider.
2. Pick **自定义中转站（OpenAI Chat）** / **Custom relay (OpenAI Chat-compatible)** (`openai-compatible`).
3. Enter any API key and set the endpoint to `https://gw.example.com/v1?api-version=2024-06`.
4. Save.

Expected: the endpoint field reports, before anything is sent, that a query string is not allowed.
Actual: the form submits, the Host rejects the write, and the form shows "模型连接服务暂时不可用，请稍后重试。" with no field marked.

The same happens on an existing connection via Settings → Models → *(connection)* → 端点 → 编辑.

The three layers can be exercised directly against a built tree, which is how the table above was produced:

```js
import { normalizeCreateConnectionInputForIpc } from './apps/desktop/dist/main/connections-ipc-validation.js';
import { normalizeCatalogConnectionBaseUrl } from './packages/core/dist/runtime-policy/connection-catalog-codec.js';
import { validateAddProviderDraft } from './apps/desktop/dist/renderer/settings/provider-add-submission.js';

const url = 'https://gw.example.com/v1?api-version=2024-06';

validateAddProviderDraft({
  providerType: 'openai-compatible', slug: 'my-relay', existingSlugs: [],
  apiKey: 'sk-test', cloudflareAccountId: '', baseUrl: url,
});
// -> null  (add-form field gate: accepted)

normalizeCreateConnectionInputForIpc({
  slug: 'my-relay', name: 'My relay', providerType: 'openai-compatible',
  baseUrl: url, defaultModel: 'gpt-4o',
}).baseUrl;
// -> 'https://gw.example.com/v1?api-version=2024-06'  (main-process IPC gate: accepted)

normalizeCatalogConnectionBaseUrl(url, 'openai-compatible');
// -> throws: connection base URL must not contain a query or fragment
```

## Environment

- Maka commit: `3bb645e99`
- OS and version: macOS 15 (Darwin 25.5.0), Apple Silicon
- Surface: Desktop (Settings → Models add/edit forms) against a local Runtime Host
- Node.js version: 22.19.0

The message the user ends up seeing is the unclassified fallback, not the Host's sentence. `providerPanelActionErrorMessage` (`apps/desktop/src/renderer/settings/provider-panel-shared.ts:56`) strips the Electron wrapper via `cleanErrorMessage`, finds no match in `shared.lastTest`, sees no CJK, and gets `''` back from `generalizedErrorMessageChinese`/`generalizedErrorMessage`, so it returns `shared.actionFallback` (`apps/desktop/src/renderer/locales/settings-provider-copy.ts:154`, `:299`).

Suggested boundary, and the one I would like to implement:

1. Make the client-side gate agree with the Host contract. `validateConnectionBaseUrl` should reject credentials and query/fragment and enforce the byte cap, so `normalizeConnectionBaseUrl` and the add-form gate inherit the same rules and the failure is raised before the IPC call. The Host stays the authority; the client stops promising something the Host will not honor.
2. Report it on the endpoint field with localized copy rather than as a form-level generic error, so the message names the rule that was broken.
3. Canonicalize on the client the same way the Host does, so a saved endpoint reads back as the value that was persisted instead of silently differing by a trailing slash or host case.

This deliberately does not relax the Host. Allowing query strings (e.g. Azure-style `?api-version=`) would be a product and security decision for `dev@maka.apache.org`, not an implementation detail; this issue is only about the client agreeing with the contract that already exists.

Related but distinct: #3405 / #3467 cover the TUI wizard being unable to create relay connections at all. This issue is about the Desktop surface that can already create them.

shaokeyibb:
## Summary

The Desktop connection forms accepted endpoints the Runtime Host refuses. A service URL carrying a query string, a fragment, or embedded credentials passed the add form's field gate, passed the main-process IPC gate, and then failed at the write — where the Host's English domain error reached the renderer unclassified and fell through to `actionFallback`:

> 模型连接服务暂时不可用，请稍后重试。
> The model connection service is temporarily unavailable. Try again later.

That names neither the field nor the rule, and asks for a retry that can never succeed. Both write paths were affected: adding a custom relay and editing an existing endpoint.

Two validators had drifted. `validateConnectionBaseUrl` checked length, `URL` parseability, and an http/https allowlist, and documented "trim is the only canonicalization". `normalizeCatalogConnectionBaseUrl` additionally rejected credentials and `?`/`#`, canonicalized through `URL.toString()`, and capped the canonical form at 2048 **bytes**.

The rule now has one implementation. `canonicalizeConnectionBaseUrl` holds the shape check, the scheme allowlist, the credential and query/fragment bans, both caps, and the canonical form. `validateConnectionBaseUrl`, `normalizeConnectionBaseUrl`, and the Host's `normalizeCatalogConnectionBaseUrl` all delegate to it and keep only their own wording — the Host's wire-facing sentences are byte-identical to before, since they are pinned by protocol tests. A rejection travels as a code (`credentials`, `query_or_fragment`, …), which is what lets the Desktop form render localized field copy from the same decision the Host makes.

Both write surfaces now check before they write: the add form marks its `baseUrl` field, and the detail page's endpoint row reports the rule that was broken instead of the generic service message.

Two notes on scope:

- **Canonicalization on the client is new and deliberate.** The previous comment argued that rewriting `https://Example.com:443/V1` could surprise whoever typed it that way. But the Host canonicalized on write regardless, so that form was never what got stored — the client was preserving a value the system discarded one hop later. Doing it here only moves the surprise to where the user can still see and correct it. The one test that pinned the old behavior is updated with that reasoning.
- **The Host contract is untouched.** Relaxing it to allow query strings, which Azure-style `?api-version=` gateways want, is a product and security decision for `dev@maka.apache.org`, not an implementation detail. This PR only makes the client agree with the contract that already exists.

Fixes #3672

## Verification

`npm run lint`, `npm run format:check`, `npm run build`, `npm run typecheck`, `npx knip --workspace apps/desktop`, `npx knip --workspace packages/ui` — all clean.

| Suite | Result |
| --- | --- |
| `@maka/core` | 651 pass / 0 fail |
| `@maka/desktop` | 1356 pass / 0 fail |
| `@maka/storage` | 906 pass / 0 fail (14 pre-existing skips) |
| `@maka/runtime-host` | 1119 pass / 0 fail |

The reported scenario, before and after, run against the built tree:

```
                          before                  after
1. add-form field gate    ACCEPTED                REJECTED {field: 'baseUrl', reason: 'invalid',
                                                            rejection: 'query_or_fragment'}
2. main-process IPC gate  ACCEPTED                REJECTED baseUrl must not contain a query or fragment
3. Runtime Host codec     REJECTED                REJECTED connection base URL must not contain a
                                                           query or fragment
```

The message the user reads at step 1 is now the rule, on the endpoint field:

```
en -> The service URL must not contain a query (?) or fragment (#). Use the advanced request
      settings below to add parameters.
zh -> 服务地址不能包含查询参数（?）或锚点（#）。如需附加参数，请使用下方的高级请求设置。
```

New tests, each pinning one contract:

- core: the four rules the client did not have (credentials, query, fragment, byte cap), plus a percent-encoded `?` staying accepted — the Host reads the typed text rather than the parsed URL, and so does the client now.
- core: **the anti-drift test** — a table of fifteen endpoints asserting the client gate and the Host codec agree on both the accept/reject verdict and the canonical value. The defect was drift between the two, not either one's rules, so this is the test that fails if they separate again.
- desktop: the field gate's new `invalid` issues, that the rule is not relay-only (a local runtime's editable port is checked too), that Cloudflare stays exempt because its field holds an account id rather than a URL, and that a bare host, a local port, and a percent-encoded `?` remain accepted.

Driven through the real app, using the `settings-models` e2e fixture whose seeded `no-models` connection is an `openai-compatible` relay with an editable endpoint row. Same steps in both: edit the service URL row, enter the URL above, save.

| | Toast |
| --- | --- |
| Before | **Failed to save model connection** — The model connection service is temporarily unavailable. Try again later. |
| After | **Failed to save model connection** — The service URL must not contain a query (?) or fragment (#). Use the advanced request settings below to add parameters. |

The row stays open with the draft intact either way; what changes is whether the message describes the thing that went wrong. Screenshots follow in a comment.

## Review focus

The canonicalization change is the one behavior change a reviewer should weigh rather than read past — it is what required editing an existing test's expectation rather than only adding to it. If the project would rather keep the client's as-typed value and fix only the misleading error, dropping the canonical form from `canonicalizeConnectionBaseUrl` and reverting that one assertion leaves the rest of the fix intact.

## AI use

Select exactly one:

Tool(s) and scope: Claude Code diagnosed the drift between the two validators, implemented the shared rule across core / the Host codec / the Desktop forms and copy, and wrote the tests and verification above. I reviewed and verified the result.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/src/main/__tests__/provider-add-submission.test.ts
+++ b/apps/desktop/src/main/__tests__/provider-add-submission.test.ts
@@ -214,6 +214,66 @@ test('the field gate still reports the rules that survived', () => {
   });
 });
 
+test('the field gate refuses endpoints the Runtime Host would refuse', () => {
+  // Before #3672 each of these passed the form and failed the write, and the
+  // failure read "the model connection service is temporarily unavailable" —
+  // wrong about the cause and wrong about retrying being worth anything.
+  assert.deepEqual(
+    validateAddProviderDraft(draft({ baseUrl: 'https://user:pw@relay.example.com/v1' })),
+    { field: 'baseUrl', reason: 'invalid', rejection: 'credentials' },
+  );
+  assert.deepEqual(
+    validateAddProviderDraft(draft({ baseUrl: 'https://relay.example.com/v1?api-version=2024-06' })),
+    { field: 'baseUrl', reason: 'invalid', rejection: 'query_or_fragment' },
+  );
+  assert.deepEqual(validateAddProviderDraft(draft({ baseUrl: 'https://relay.example.com/v1#x' })), {
+    field: 'baseUrl',
+    reason: 'invalid',
+    rejection: 'query_or_fragment',
+  });
+  assert.deepEqual(validateAddProviderDraft(draft({ baseUrl: 'not-a-url' })), {
+    field: 'baseUrl',
+    reason: 'invalid',
+    rejection: 'malformed',
+  });
+  assert.deepEqual(validateAddProviderDraft(draft({ baseUrl: 'javascript:alert(1)' })), {
+    field: 'baseUrl',
+    reason: 'invalid',
+    rejection: 'scheme',
+  });
+
+  // A provider that ships its own endpoint still exposes the field for local
+  // runtimes, so the rule is not relay-only.
+  assert.deepEqual(
+    validateAddProviderDraft(
+      draft({ providerType: 'ollama', baseUrl: 'http://127.0.0.1:11434/v1?x=1' }),
+    ),
+    { field: 'baseUrl', reason: 'invalid', rejection: 'query_or_fragment' },
+  );
+
+  // Cloudflare's field carries an account id, not a URL — there is nothing to
+  // parse until the template is interpolated after submission.
+  assert.equal(
+    validateAddProviderDraft(
+      draft({
+        providerType: 'cloudflare-workers-ai',
+        cloudflareAccountId: 'abc123',
+        baseUrl: '',
+      }),
+    ),
+    null,
+  );
+
+  // Still accepted: a bare host, a local port, and a percent-encoded `?`.
+  for (const baseUrl of [
+    'https://relay.example.com',
+    'http://127.0.0.1:8080/v1',
+    'https://relay.example.com/a%3Fb',
+  ]) {
+    assert.equal(validateAddProviderDraft(draft({ baseUrl })), null, baseUrl);
+  }
+});
+
 test('a duplicate slug outranks a missing key, so one fix is asked for at a time', () => {
   assert.deepEqual(
     validateAddProviderDraft(
--- a/apps/desktop/src/renderer/locales/settings-provider-copy.ts
+++ b/apps/desktop/src/renderer/locales/settings-provider-copy.ts
@@ -156,6 +156,15 @@ const zhCopy = {
     modelKeyAria: (name: string) => `${name} 模型密钥`,
   },
   shared: {
+    // One sentence per rule the Runtime Host enforces on a service URL. The
+    // form checks these before it writes, so the user reads the rule they
+    // broke instead of `actionFallback`, which used to blame the service and
+    // suggest a retry that could never succeed (#3672).
+    endpointInvalidUrl: '服务地址不是有效的 URL，请填写完整地址，例如 https://example.com/v1。',
+    endpointScheme: '服务地址只支持 http:// 或 https:// 开头。',
+    endpointCredentials: '服务地址不能包含用户名或密码，请把密钥填在上方的 API Key 中。',
+    endpointQueryOrFragment: '服务地址不能包含查询参数（?）或锚点（#）。如需附加参数，请使用下方的高级请求设置。',
+    endpointTooLong: '服务地址过长，请控制在 2048 字节以内。',
     actionFallback: '模型连接服务暂时不可用，请稍后重试。', rateLimit: '当前账号或模型服务触发速率限制，请稍后重试。',
     timeout: '请求超时，请检查网络或代理后重试。', unavailable: '模型服务暂时不可用，请稍后重试。',
     network: '网络错误，请检查服务地址或代理设置后重试。', statusUnavailable: '连接测试状态暂时无法显示，请重新测试。',
@@ -306,6 +315,11 @@ const enCopy: ProviderSettingsCopy = {
     modelKeyAria: (name: string) => `${name} model key`,
   },
   shared: {
+    endpointInvalidUrl: 'The service URL is not a valid URL. Enter a full address, for example https://example.com/v1.',
+    endpointScheme: 'The service URL must start with http:// or https://.',
+    endpointCredentials: 'The service URL must not contain a username or password. Put the secret in the API key field above.',
+    endpointQueryOrFragment: 'The service URL must not contain a query (?) or fragment (#). Use the advanced request settings below to add parameters.',
+    endpointTooLong: 'The service URL is too long. Keep it under 2048 bytes.',
     actionFallback: 'The model connection service is temporarily unavailable. Try again later.', rateLimit: 'This account or model service is rate-limited. Try again later.',
     timeout: 'The request timed out. Check the network or proxy and try again.', unavailable: 'The model service is temporarily unavailable. Try again later.',
     network: 'Network error. Check the service URL or proxy settings and try again.', statusUnavailable: 'The connection test status is temporarily unavailable. Test again.',
--- a/apps/desktop/src/renderer/settings/provider-add-form.tsx
+++ b/apps/desktop/src/renderer/settings/provider-add-form.tsx
@@ -43,6 +43,7 @@ import { providerDisplay } from './provider-display';
 import { useActionGuard } from './use-action-guard';
 import {
   categoryLabel,
+  connectionEndpointRejectionMessage,
   providerPanelActionErrorMessage,
   type ConnectionsBridge,
 } from './provider-panel-shared';
@@ -125,7 +126,11 @@ export function AddProviderForm(props: {
     }
     if (issue.field === 'apiKey') return copy.keyRequired(display.name);
     if (issue.field === 'accountId') return copy.cloudflareAccount;
-    if (issue.field === 'baseUrl') return copy.endpointRequired;
+    if (issue.field === 'baseUrl') {
+      return issue.reason === 'required'
+        ? copy.endpointRequired
+        : connectionEndpointRejectionMessage(issue.rejection, locale);
+    }
     return copy.accountLogin;
   }
 
--- a/apps/desktop/src/renderer/settings/provider-add-submission.ts
+++ b/apps/desktop/src/renderer/settings/provider-add-submission.ts
@@ -18,11 +18,13 @@
  */
 
 import {
+  canonicalizeConnectionBaseUrl,
   PROVIDER_DEFAULTS,
   providerAuthRequiresSecret,
   providerAuthSupportsApiKey,
   providerSupportsModelDiscovery,
   validateSlug,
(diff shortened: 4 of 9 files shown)


Astro-Han (commented):
## Scope

First independent full review of exact head `26b3d031a39e3e75d84ef5eb28b64c4a7daed5f2`.

I covered the shared base-URL validation/normalization contract, the Runtime Host catalog codec and canonical persisted-entry boundary, create/edit renderer-to-IPC flows, localized rejection copy, tests, and the exact-head CI result.

## Findings

No P0–P3 findings in the covered scope.

The renderer and Runtime Host now use the same canonicalization authority. The codec continues to reject non-canonical or invalid persisted entries at the existing canonical decode boundary; I found no new silent-drop path or behavior that deletes legacy connections. Clear intent remains distinguishable from an omitted update, and OAuth/default endpoint handling remains explicit.

## Verification

- Exact-head CI `test` run `32691266671`: terminal success.
- Local `@maka/core` tests: 651 passed, 0 failed.
- Changed-file Biome check: passed.
- `git diff --check`: passed.

The desktop main build was not used as a PR signal because this shared workspace has pre-existing cross-package dist/type drift; its errors were outside the changed endpoint path.

## Evidence gap

The PR changes user-visible endpoint error copy, but no screenshot or visual artifact is attached. The new focused tests validate behavior and copy selection, not rendered visual output. I record this as an acceptance-evidence gap, not a code finding.

shaokeyibb:
before

after


FRI, here's before-after screenshots

shaokeyibb:
rebased and resolve the conflict to match latest changes in main branch

me2seeks (commented):

The rule-sharing structure is right: one canonicalization function behind the form gate, the IPC gate and the host codec, with rejections travelling as codes and rendered as localized copy. I enumerated the before/after acceptance set and found **no input becomes more permissive** — credentials, query and fragment are now rejected where they were previously accepted, the byte cap is enforced consistently, and the host's wire messages are unchanged. Two items.

**P2 (Should-Fix) — the canonical value is computed and then discarded, so the read-back behavior the PR describes does not happen.**

```ts
// apps/desktop/src/renderer/settings/use-connection-detail.ts:312-321
if (field === 'endpoint') {
  const endpoint = canonicalizeConnectionBaseUrl(baseUrl);
  if (!endpoint.ok) { ...reject... ; return false; }
}
...
// :328-330
await props.bridge.update(
  connection.slug,
  field === 'key' ? { apiKey } : field === 'name' ? { name: draftName } : { baseUrl },
);
```

`endpoint.value` is used only for the rejection decision; the write still sends the raw typed text and no `setBaseUrl(endpoint.value)` runs on success. The host re-canonicalizes on write anyway, so what is stored and displayed is unchanged — meaning the "typed and stored values can differ by a trailing slash or host case" goal the PR cites is not addressed. Concretely: save a non-canonical but valid endpoint, and the draft keeps the typed text, so the change flag stays set and Save remains enabled after a successful save; reopening the row prefills the typed value rather than the persisted canonical one (the key row resets, the endpoint row does not). Either use `endpoint.value` for the write and the field state, or reset the field from the saved value when the endpoint row opens — or drop the claim.

**P2 (Should-Fix) — the edit-path gate and its new message have no test.** `use-connection-detail.ts` and `connectionEndpointRejectionMessage` are referenced by no test, yet the detail-page toast is the PR's central user-visible claim (the add-form gate is covered). One case asserting that an invalid endpoint marks the field and shows the localized reason would pin it.

**P3 (Nice-to-have) — `validateConnectionBaseUrl` is production-dead and its new comment is inaccurate.** At this head the only production caller is gone, yet the doc says it is "used by the IPC gate". Either delete it and migrate the two test files to the canonicalizer, or correct the comment.

**P3 (Nice-to-have) — one message folds a non-string into "too long"**, so the text is false for that reason (currently unreachable because both callers pass strings), and the copy says "2048 bytes" even for the character case. And the new query/fragment copy points the user at the "advanced request settings", which edit headers and a body overlay — they cannot add a URL parameter, so for the `?api-version=` case the issue names, the advice leads somewhere that cannot help.

**Review-relevant risks.** No input became more permissive and the host remains the ultimate gate, so there is no persisted-invalid path. One consequence worth noting: the update path is now stricter, so a value persisted by an older build containing credentials or a query would no longer be re-savable through the endpoint row — I could not confirm any such value exists. User-visible behavior change, so independent human review is required under `CONTRIBUTING.md`.

**Required conclusion.**
1. Optimal for the actual problem? The rule-sharing is right and at the correct source of truth, but half the change has no consumer of its result, so it adds a stricter gate without the read-back benefit used to justify it.
2. Production code that can be deleted? `validateConnectionBaseUrl` (no production caller at head).
3. Low-quality tests to delete or replace? `none identified`. The gap is the missing edit-path coverage; the new anti-drift test that compares the two normalizers is genuinely good and should stay.
4. Deeper refactor required? No — keeping the host's messages pinned while sharing the rule is the right split.
5. Ready to merge? Functionally safe; fix the discarded canonical value or soften the claim, and add the edit-path test.
6. Residual risks / verification gaps: I did not run the suite, so CI status is unverified. The pre-existing-persisted-value case above is unverified.

PR #3677, opened 2026-08-24
feat(desktop): let a provider connection be renamed after it is created

Joob1n:
## Summary

A provider connection's name was fixed at creation, and the name is the one field the user chose. A connection acquires its real name in use — once the account it points at is clear, or once a second connection to the same provider makes "OpenAI" ambiguous. The only way to rename was to delete and recreate, which loses the enabled model set and the stored credential with it.

Nothing below the renderer needed changing. `UpdateConnectionInput` already carries `name`, `ConnectionCatalogEntryUpdate` requires it, and `connection.catalog.update` was already sending the current value back unchanged on every save. This adds the row that lets the user supply a different one.

**Renaming cannot break a reference.** Sessions bind to `llmConnectionSlug`, and `slug` is absent from `ConnectionCatalogEntryUpdate` entirely — the catalog has no way to change it, so a rename moves only what is displayed.

The row is a `SettingsExpandableRow` beside 模型密钥 and 服务地址, so it reads and behaves like the other settled values on the page rather than introducing a second editing idiom. It sits above them because it is the field the user picked, and it is not behind the API-key or endpoint guards — every connection has a name, including the ones with neither.

## Draft handling

Extracted to `connection-name-draft.ts` so it is testable without React, following `relay-profile-draft.ts`:

- **A slug switch always reseeds.** Typing a name for connection A, switching to B, and saving would otherwise rename B to A's draft.
- **A same-slug change reseeds only while the draft is clean**, so a rename landing from this page — or from another client watching the same catalog — does not discard a name being typed.
- **An emptied field is not offered as a change.** The catalog requires a name; clearing the field is a half-finished edit, not a request to have none.
- **Whitespace alone is not a change**, and a rename commits the trimmed value. Otherwise an accidental trailing space reads as an edit whose save stores a name that renders identically to the one already stored.

Opening any row abandons what another row was holding, since only one is editable at a time and a draft left behind would be committed by an action the user never connected to it.

## Self-review

Reviewing the first commit as a reviewer rather than re-running its tests found two defects, fixed in `1e0f70d`.

**The row hid itself on the connections that needed it most.** It sat inside the `supportsApiKey || showsEndpoint` guard while its own comment claimed it was outside one. A connection with neither — an OAuth subscription — lost the only editable field it has and rendered an empty section. The row is now outside that guard and behind `!retired`, which is the condition that actually applies.

**A rename issued a model fetch.** The post-save rule was `wroteNewKey || field === 'endpoint' || models.length === 0`, and a rename satisfies the third clause whenever the model cache is empty. Renaming a connection therefore started a network fetch nobody asked for, able to raise an error about a connection the user had only renamed. The rule is now `shouldRefreshModelsAfterSave`, which excludes a rename; a new credential, a new endpoint, and an empty cache behind either of those still fetch.

## Verification

Thirteen tests, and each rule is pinned by reverting it:

```
$ node --test apps/desktop/dist/main/__tests__/connection-name-draft.test.js
ℹ tests 13
ℹ pass 13
ℹ fail 0
```

| Mutation | Result |
|---|---|
| stop reseeding on a slug switch | 8 pass, **1 fail** |
| drop the trim and the empty guard | 7 pass, **2 fail** |
| let a rename fetch models | 12 pass, **1 fail** |

Checked in the running app against a real relay connection: the row renders at the top of 连接, an edit saves and the name updates in the header and the connection list, and the enabled models and stored credential are untouched.

```
npm run format:check              Checked 1614 files. No fixes applied.
npm --workspace @maka/desktop run typecheck    clean
npm run astryx:surface-inventory  ok (206 files, 1 exclusions)
npm run check:stale               dist is fresh.
npm run check:third-party-notices OK
```

`npm run check:asf-source` fails on this machine, and does so identically on a clean `main` — `sync-model-metadata.mjs` passes `mode: 'transform'` to a Node 26 type-stripping API that now accepts only `'strip'`. Unrelated to this change, and reported here rather than left unmentioned.

## AI use

Select exactly one:

Tool(s) and scope: Claude Opus 5 via Claude Code — traced the existing update path to confirm the backend already accepted a new name and that `slug` could not be affected, wrote the row, the draft rules and their tests, and ran the mutation checks. The commit carries a `Generated-by` trailer.

## Checklist

- [ ] No


likun666661 (approved):
LGTM

PR #3681, opened 2026-08-24
feat(desktop): use contextual user-message timestamps

### LINKED ISSUE #3678: feat(desktop): use contextual user-message timestamps
## Problem

Desktop currently displays a time-only label on user messages, such as `14:30`.

That is sufficient for messages from today, but becomes ambiguous in sessions spanning several days: the reader cannot tell whether a historical message was sent today or on an earlier date.

The clarified product goal is to give the reader enough context to understand when a message occurred. Showing an exact wall-clock time for messages from today, or omitting the year from same-year historical dates, is not a requirement.

No existing public issue was found for this exact message-timestamp behavior.

## Desired outcome

Use the existing Astryx `Timestamp` component's contextual `format="auto"` presentation for user-message metadata:

- Recent messages use relative labels such as `2 hours ago`.
- Messages outside Astryx's auto threshold use its localized `date_time` presentation.
- The compact timestamp remains visible at rest, including on touch devices.
- The existing hover/focus card continues to expose a complete, copyable absolute timestamp.
- Astryx retains ownership of timestamp parsing, locale and system-preference formatting, semantic `<time>` markup, and accessible labeling.

This issue only changes the existing user-message timestamp format. It does not add timestamps to assistant responses or introduce date separators into the transcript.

## Acceptance criteria

- User-message timestamps render through Astryx `Timestamp` with `format="auto"`.
- A message from a few hours ago renders as a relative label rather than a bare clock time.
- Older messages render with date and time according to Astryx's existing auto-threshold behavior.
- Timestamps remain visible and reachable without hover, including on touch devices.
- Relative timestamps retain the existing keyboard-reachable hover card with a complete, copyable absolute time.
- The existing system locale, time zone, and 12/24-hour preferences remain respected by Astryx.
- A focused UI regression test covers the consumer's `auto` behavior and accessible hover-card anchor.

## Alternatives considered

A local three-tier calendar formatter could show time for today, month/day/time within the current year, and year/month/day/time across years. It would also require a separate rendering component, locale policy, midnight refresh source, reveal styling, and dedicated interaction tests. The additional precision is not required for the product goal.

Hiding timestamps at rest was also considered, but it removes the timestamp's touch path and duplicates interaction behavior already provided by Astryx's compact label and full absolute-time hover card.

If exact local-calendar tiers become a reusable requirement later, they should be added to the design-system component rather than maintained as a product-local replacement.

## AI assistance

An AI coding assistant was used to search for related issues, draft and revise this report, implement the change, and run verification. The human contributor reviewed the final content and remains responsible for its accuracy and submission.

Sun-GLiang:
## Summary

- switch the existing user-message Astryx `Timestamp` from `format="time"` to `format="auto"`
- show recent messages as relative time and older messages as localized date/time, while retaining Astryx's complete copyable absolute-time hover card
- keep timestamps visible at rest and reachable on touch and keyboard, without a product-local formatter, rendering component, reveal CSS, or midnight timer

Closes #3678

## Product decision

The requirement is to help readers understand when a message occurred; an exact wall-clock time for messages from today is not required. Astryx's existing `auto` behavior satisfies that need, and including the year for older same-year messages is not ambiguous enough to justify a separate calendar-tier implementation.

The final PR therefore reuses the existing design-system component rather than replacing it.

## Visual evidence

Both captures use the same seeded Desktop Electron fixture, message, locale, theme, platform, and viewport. Only the timestamp format changes.

**Before — `format="time"` shows only the clock time, even for a historical message:**

Before: historical user message shows only 10:50

**After — `format="auto"` identifies the historical message with localized date and time:**

After: historical user message shows 2026年5月22日 10:50

These screenshots were captured from the real Desktop Electron renderer; they are not generative images.

## Verification

- `npm --workspace @maka/ui test` — 230 passed, 0 failed
- `npm --workspace @maka/desktop run build:with-deps` — passed
- `npm run lint` — 2,756 files checked, no issues
- `npm run format:check` — 1,651 files checked, no issues
- `npm run typecheck` — passed across all workspaces
- `node scripts/check-astryx-surface-inventory.mjs` — passed (213 files, 1 exclusion)
- `git diff --check HEAD^ HEAD` — passed

The focused UI regression was developed red-first: it failed against the local timestamp replacement because the user message did not render Astryx's semantic timestamp and keyboard-reachable hover-card anchor, then passed after restoring Astryx with `format="auto"`.

## AI use

Select exactly one:

Tool(s) and scope: OpenAI Codex helped evaluate the review feedback, clarify the product requirement, replace the local timestamp path with Astryx `format="auto"`, remove the superseded Core/UI/E2E implementation, add the red-to-green consumer regression, run verification, capture real Electron before/after evidence, and update this description. That review does not replace the repository's required independent human review. Every affected commit carries a `Generated-by: Codex` trailer. The human contributor remains responsible for the change, evidence, and submission.

## Checklist

- [x] Lint, format, typecheck and the affected suite pass locally

- [ ] No


Sun-GLiang:
Could a maintainer please trigger CI for this PR? Thanks!

Astro-Han (commented):
Reviewed at `22dacd3cc2`. One `[P2]`, one `[P3]`, plus a gate note.

The shape of this is good: the date-precision policy lives in Core as a pure function, the reveal is CSS-only with `opacity` so the transcript never reflows, `:focus-within` is there alongside `:hover` so keyboard users are not excluded, and the visible run is `aria-hidden` with one complete absolute label exposed to assistive technology instead of letting a screen reader read a bare "23:20". The midnight refresh is real rather than a polling interval.

**[P2] The one visible string is assembled by two formatters that resolve different locales**

`presentConversationMessageTimestamp` formats with `uiLocaleToIntlLocale(locale)` — the app's UI locale, `zh-CN` or `en`. It then computes `datePrefix` by slicing the formatted parts up to (but not including) the `hour` part, on the assumption that the clock will continue exactly where the prefix stopped.

The clock does not come from that formatter. `ConversationMessageTimestamp` renders `<Timestamp format="time">` from `@astryxdesign/core`, and `formatInstant` calls `new Intl.DateTimeFormat(undefined, TIME_OPTIONS)` — `undefined` resolves to the **host** locale, not the app's.

So the two halves disagree whenever the UI locale differs from the OS locale, which is precisely the case the UI locale setting exists to serve. With the app set to Chinese on an `en-US` machine, a message from earlier this year renders as `8月24日 ` + `11:20 PM`. The `absoluteLabel` read to assistive technology is built from the *app* locale, so the screen reader and the screen also disagree.

There is a quieter failure in the same seam. `prefixBeforeClock` returns `''` when no `hour` part is found, and slices strictly before the hour otherwise. In any locale ordering that does not place the date before the clock, the date silently disappears from the visible label — the user sees only a time, with no indication that the message is from another day or another year. The hidden label still carries the full date, so nothing looks broken; the information is simply gone from the screen.

Please format the whole visible string from one formatter. Either render `presentation.fallbackText` directly instead of composing a prefix with a separately formatted clock, or pass the resolved app locale into the Astryx component if it accepts one. As written, the prefix and the clock are only guaranteed to agree on machines whose OS locale happens to match the app's.

**[P3] One timer per message**

`ConversationMessageTimestamp` schedules its own `setTimeout` to the next local midnight, and the effect depends on `now`, so every mounted message holds a timer and they all fire together at midnight. On a long transcript that is a few thousand timers to re-render every message at the same instant, for a change that affects only messages crossing a day boundary. A single shared midnight tick — a context or a module-level subscription — would give the same behaviour at one timer.

**Gate:** no check has run on this head at all. Never-run is not green, so this cannot be approved yet regardless of the findings.

<details>
<summary>简体中文</summary>

已在 `22dacd3cc2` 上审查。一条 `[P2]`、一条 `[P3]`，外加一条门禁说明。

整体形状是好的：日期精度策略作为纯函数放在 Core；显隐纯靠 CSS `opacity`，所以 transcript 不会重排；`:focus-within` 与 `:hover` 并列，键盘用户没有被排除在外；可见部分标了 `aria-hidden`，另外向辅助技术暴露一个完整的绝对时间，而不是让读屏软件念一句光秃秃的"23:20"。跨午夜刷新是真的定时，而不是轮询。

**[P2] 同一个可见字符串，是由两个解析出不同 locale 的格式化器拼出来的**

`presentConversationMessageTimestamp` 用 `uiLocaleToIntlLocale(locale)` 格式化——也就是应用的 UI locale，`zh-CN` 或 `en`。然后它把格式化后的 parts 切到 `hour` 之前得到 `datePrefix`，前提假设是：时钟会恰好从前缀结束的地方接上。

但时钟并不来自那个格式化器。`ConversationMessageTimestamp` 渲染的是 `@astryxdesign/core` 的 `<Timestamp format="time">`，而 `formatInstant` 调用的是 `new Intl.DateTimeFormat(undefined, TIME_OPTIONS)`——`undefined` 解析到的是**宿主**locale，不是应用的。

于是只要 UI locale 与操作系统 locale 不一致，两半就会打架——而这恰恰正是 UI locale 这个设置存在的意义。应用设为中文、机器是 `en-US` 时，今年早些时候的一条消息会渲染成 `8月24日 ` + `11:20 PM`。而念给辅助技术的 `absoluteLabel` 用的是**应用**locale，因此读屏与屏幕显示同样对不上。

同一处接缝还有一个更安静的失败。`prefixBeforeClock` 在找不到 `hour` part 时返回 `''`，否则严格切到 hour 之前。在任何"时钟不在日期之后"的 locale 排序下，日期会从可见标签里**悄悄消失**——用户只看到一个时间，完全看不出这条消息来自另一天甚至另一年。隐藏标签里仍带着完整日期，所以表面上没有任何异常；信息只是从屏幕上没了。

请让整个可见字符串出自同一个格式化器：要么直接渲染 `presentation.fallbackText`，不要用"前缀 + 另行格式化的时钟"拼；要么把解析后的应用 locale 传进 Astryx 组件（如果它接受）。按现在的写法，前缀与时钟只有在"操作系统 locale 恰好等于应用 locale"的机器上才保证一致。

**[P3] 每条消息一个定时器**

`ConversationMessageTimestamp` 各自 `setTimeout` 到下一个本地午夜，且该 effect 依赖 `now`，因此每条已挂载的消息都持有一个定时器，并且它们会在午夜同时触发。长 transcript 上就是几千个定时器在同一瞬间把每条消息重渲染一遍，而真正受影响的只有跨天边界的那几条。用一个共享的午夜 tick（context 或模块级订阅）能以一个定时器得到同样的行为。

**门禁**：这个 head 上一条 check 都没跑过。「没跑过」不等于绿，所以无论上述发现如何，现在都不具备 approve 条件。
</details>

Sun-GLiang:
Addressed the review in `d22f7e091`.

- **P2 locale consistency:** the visual `<time>` now renders the complete Core-produced `visibleText`, so date and clock use the active UI locale from one formatter. The Astryx base component remains unchanged.
- **P3 timer fan-out:** mounted message timestamps now subscribe to one module-level local-midnight source via `useSyncExternalStore`; the last subscriber clears the timer and a later mount creates one fresh timer.
- Added regressions for UI-locale/host-locale mismatch, shared timer lifecycle, mounted reclassification at midnight, and the render-before/subscribe-after-midnight race.

Local verification on this head: Core 657/657, UI 233/233, timestamp Electron E2E 1/1, lint 2,677 files, format 1,617 files, full build, all-workspace typecheck, and `git diff --check`.

The PR still reports no checks because the fork workflows have not been approved; that gate still requires a maintainer. @Astro-Han, please re-review the updated head when available.

Astro-Han (dismissed):
**Two questions before any line-level detail.**

**1. Why not Astryx `Timestamp`?**

`main` already uses it — `chat-turn.tsx:196`, `<Timestamp value={props.ts} format="time" />` inside `ChatMessageMetadata`. This PR's first change deletes that import and replaces it with a 211-line local equivalent.

`Timestamp` already ships most of what #3678 asks for: a copyable hover card carrying the full absolute time (`hasTooltip`, on by default), `isLive` refresh, semantic `<time>` + ISO `datetime`, theme-driven sizing and color.

It genuinely lacks one thing: `format="auto"` tiers by elapsed time, while #3678 requires local calendar fields (a 23:59 message must change shape after midnight). That gap is real — but it is in the tiering policy, not the rendering layer. Keeping `Timestamp` and feeding it a calendar-derived format, or raising the capability upstream, costs a fraction of replacing it.

**2. Is "hidden at rest" a product decision?**

Astryx's answer to "full precision without clutter" is the hover card: compact label at rest, detail on hover. Hiding the label entirely departs from that, and the departure is what pays for the module-global midnight timer, the reveal CSS plus 68 lines of E2E asserting the transcript does not shift, and the loss of the copyable hover card. It also leaves no touch path — there is no `@media (hover: none)` branch (`sidebar.css:394` is the existing precedent), so on touch the timestamp is simply unreachable.

If the hiding is wanted, it needs an explicit call, because that is what these costs buy.

**P1 — `hourCycle` injection breaks zh-CN dates**

`conversation-message-timestamp.ts:48` reads the *host* locale's hour cycle; `:78` injects it into the *UI* locale's formatter. Measured:

```
zh-CN, no hourCycle    → 2026年8月24日 23:20
zh-CN + hourCycle h12  → 2026/8/24 下午11:20
```

`hourCycle` changes ICU's date skeleton for `zh-CN`, so an `en-US` host with a Chinese UI gets a mangled date. The comment this PR rewrites in `chat-display-helpers.ts` recorded exactly this — locking the hour cycle "was the app overriding a preference that belongs to the reader's system, which is why `Timestamp` formats against the host locale and offers no hour-cycle knob." The PR drops that clause and then does it.

This is a symptom of the hand-rolled path, not an independent defect.

The date-precision work itself — Core module, calendar boundaries, midnight re-evaluation, AT label — is sound. The question is only where it should live.

Sun-GLiang:
> **Two questions before any line-level detail.**
> 
> **1. Why not Astryx `Timestamp`?**
> 
> `main` already uses it — `chat-turn.tsx:196`, `<Timestamp value={props.ts} format="time" />` inside `ChatMessageMetadata`. This PR's first change deletes that import and replaces it with a 211-line local equivalent.
> 
> `Timestamp` already ships most of what #3678 asks for: a copyable hover card carrying the full absolute time (`hasTooltip`, on by default), `isLive` refresh, semantic `<time>` + ISO `datetime`, theme-driven sizing and color.
> 
> It genuinely lacks one thing: `format="auto"` tiers by elapsed time, while #3678 requires local calendar fields (a 23:59 message must change shape after midnight). That gap is real — but it is in the tiering policy, not the rendering layer. Keeping `Timestamp` and feeding it a calendar-derived format, or raising the capability upstream, costs a fraction of replacing it.
> 
> **2. Is "hidden at rest" a product decision?**
> 
> Astryx's answer to "full precision without clutter" is the hover card: compact label at rest, detail on hover. Hiding the label entirely departs from that, and the departure is what pays for the module-global midnight timer, the reveal CSS plus 68 lines of E2E asserting the transcript does not shift, and the loss of the copyable hover card. It also leaves no touch path — there is no `@media (hover: none)` branch (`sidebar.css:394` is the existing precedent), so on touch the timestamp is simply unreachable.
> 
> If the hiding is wanted, it needs an explicit call, because that is what these costs buy.
> 
> **P1 — `hourCycle` injection breaks zh-CN dates**
> 
> `conversation-message-timestamp.ts:48` reads the _host_ locale's hour cycle; `:78` injects it into the _UI_ locale's formatter. Measured:
> 
> ```
> zh-CN, no hourCycle    → 2026年8月24日 23:20
> zh-CN + hourCycle h12  → 2026/8/24 下午11:20
> ```
> 
> `hourCycle` changes ICU's date skeleton for `zh-CN`, so an `en-US` host with a Chinese UI gets a mangled date. The comment this PR rewrites in `chat-display-helpers.ts` recorded exactly this — locking the hour cycle "was the app overriding a preference that belongs to the reader's system, which is why `Timestamp` formats against the host locale and offers no hour-cycle knob." The PR drops that clause and then does it.
> 
> This is a symptom of the hand-rolled path, not an independent defect.
> 
> The date-precision work itself — Core module, calendar boundaries, midnight re-evaluation, AT label — is sound. The question is only where it should live.
> 
> _AI use: findings produced with Claude Code and verified by me against this branch; not a substitute for independent human review._

Thank you for the review.

1. I agree that we should reuse the existing `Timestamp` component wherever possible. I will refactor the implementation accordingly and address the `hourCycle` issue rather than maintaining a separate rendering path.

2. Hiding timestamps by default was not an established product decision. After discovering that messages from previous days displayed only the time, I compared the interaction with Codex’s timestamp behavior and felt that revealing timestamps on demand might reduce visual clutter and create a quieter interface. I therefore adopted it as a temporary design choice.

   I’d like to confirm with the maintainers: do we want to keep the reveal-on-intent behavior? If so, I will also ensure that timestamps remain accessible on touch devices. If not, I will revert that part of the change and keep the scope focused on showing the date for messages that are not from today.

3. I will proceed with the component reuse and formatting fixes now, and adjust the interaction as soon as the maintainers confirm the intended behavior.

Sun-GLiang:
I did consider reusing Astryx `Timestamp` during the original design. The trade-off is documented in #3678: switching the existing component between `time` and `date_time` is simpler, but `date_time` includes the year for every historical message. That conflicts with the intended precision tiers:

- Today: time only
- Another day in the same year: month, day, and time
- Another year: year, month, day, and time

This was the reason for keeping the calendar-boundary policy and formatter in Maka. However, I agree that the current implementation went too far into replacing the rendering layer and introduced the `hourCycle` issue identified in this review. I will correct that issue immediately.

For the broader component boundary, the current Astryx API cannot express “date and time without the year.” I’d like to confirm which trade-off the maintainers prefer:

- Reuse `Timestamp` directly and accept that all cross-day messages include the year; or
- Preserve the existing precision tiers with a focused formatter in Maka.

Hiding timestamps by default was not an established product decision. After discovering that messages from previous days displayed only the time, I compared the interaction with Codex’s timestamp behavior and felt that revealing timestamps on demand might reduce visual clutter and create a quieter interface. I therefore adopted it as a temporary design choice.

I’d also like to confirm whether the reveal-on-intent behavior is desired. If it is, I will address the touch-device accessibility path as part of that design. If it is not, I will revert the hiding behavior and keep the change focused on showing dates for cross-day messages.

I will proceed with the independent formatting fix now, then adjust the component boundary and interaction once the maintainers confirm the intended trade-offs.

Astro-Han:
Thanks — and thanks for the clear write-up of the trade-off. Two answers and one question.

**Hidden at rest: drop it.** It was not an established decision, and we do not want to copy Codex's interaction wholesale; where the design system already has an answer we align with it. Astryx's answer to "full precision without clutter" is the hover card, and `hasTooltip` is on by default. So please keep the timestamp visible and scope the change to "a message that is not from today should be identifiable". That also removes the touch-accessibility work, the layout-stability E2E, and the reveal CSS.

**`hourCycle`: this is an upstream bug, please do not work around it locally.** `Timestamp` formatting ignores the `InternationalizationProvider` locale — facebook/astryx#5074 — and facebook/astryx#5303 is the open fix, which touches `Timestamp/formatInstant.ts`. Injecting the host hour cycle into the UI locale's formatter is what mangles zh-CN dates (`2026年8月24日 23:20` becomes `2026/8/24 下午11:20`), so let us track the upstream fix instead. Worth checking separately whether `main` is already affected — we wrap `AstryxLocaleProvider` in `chat-turn.tsx:87`, but per #5074 `Timestamp` may not be reading it.

**The question: why is `format="auto"` not enough?**

`auto` renders relative for recent values and `date_time` beyond `autoThreshold` (7 days by default), so "which day was this" is answered without a local component — and the copyable hover card still carries the full absolute time. Comparing it against the three tiers in #3678, the only thing it cannot express is omitting the year for a same-year message: `2026年8月23日 14:30` instead of `8月23日 14:30`. Both are unambiguous; one is four characters longer.

If terseness is the whole gap, that is a small upstream ask and not worth 211 lines of local component, a Core module, a module-global midnight timer, and 68 lines of E2E to own here.

The one thing I would like your read on: `auto` shows today's messages as "2 hours ago" rather than `14:30`. If clock time for today is the real requirement, say so and we will weigh it properly — `isLive` is off by default so there is no churn either way. But if `auto` is acceptable, this PR shrinks to roughly one word, and the `hourCycle` problem disappears with the hand-rolled path.

<details>
<summary>简体中文</summary>

谢谢，也谢谢你把这个取舍写得很清楚。两个答复、一个问题。

**静置隐藏：去掉。** 这不是已确立的决定，而且我们不希望整体照搬 Codex 的交互；设计系统已经有答案的地方就向它靠齐。Astryx 对「既要完整信息又不杂乱」的答案是 hover card，且 `hasTooltip` 默认开启。所以请让时间戳保持可见，把改动收回到「不是今天的消息要能看出是哪天」。这样触屏可达性、布局稳定性 E2E、以及那套揭示 CSS 也都不需要了。

**`hourCycle`：这是上游的 bug，请不要在本地绕过它。** `Timestamp` 的格式化会忽略 `InternationalizationProvider` 的 locale —— facebook/astryx#5074 —— 而 facebook/astryx#5303 正是修它的开放 PR，改到了 `Timestamp/formatInstant.ts`。把宿主的时制注入界面 locale 的格式化器，正是 zh-CN 日期变形的原因（`2026年8月24日 23:20` 变成 `2026/8/24 下午11:20`），所以我们跟随上游修复即可。另外值得单独确认 `main` 是否已经受影响——我们在 `chat-turn.tsx:87` 套了 `AstryxLocaleProvider`，但按 #5074 的描述 `Timestamp` 可能根本没读它。

**问题：`format="auto"` 为什么不够？**

`auto` 对较近的时间渲染相对时间，超过 `autoThreshold`（默认 7 天）后转 `date_time`，所以不需要本地组件就回答了「这是哪天」——而且可复制的 hover card 依然带着完整的绝对时间。对照 #3678 的三档精度，它唯一表达不出来的是同年时省掉年份：`2026年8月23日 14:30` 而不是 `8月23日 14:30`。两者都无歧义，只差四个字。

如果简洁度就是全部差距，那它是一个很小的上游诉求，不值得在这里自己维护 211 行本地组件、一个 Core 模块、一个模块级全局午夜定时器和 68 行 E2E。

有一点想听你的判断：`auto` 会把今天的消息显示成「2 小时前」而不是 `14:30`。如果「今天要显示时钟时间」才是真正的需求，请直说，我们会认真权衡——`isLive` 默认关闭，所以两种方式都不会有跳动。但如果 `auto` 可以接受，这个 PR 会缩到大约一个词，而 `hourCycle` 的问题也会随手写路径一起消失。

</details>

Sun-GLiang:
I agree that the core requirement is to help users understand when a message occurred, rather than to show an exact wall-clock time for messages from today. In that context, `2 hours ago` is sufficient, and omitting the year for same-year messages is not important enough to justify the additional local implementation.

I’ll revise the PR to reuse the existing Astryx `Timestamp` with `format="auto"` and retain its default compact label plus full absolute-time hover card. The custom Core presenter, midnight subscription, local rendering component, and their associated tests can be removed. This also eliminates the `hourCycle` path and its locale-formatting issue.

Astro-Han (approved):
Replaces the 211-line local timestamp implementation with one line — `format="time"` → `format="auto"` — plus a focused regression test. Net 17 added against 831 deleted since my last review.

This is the right resolution. My earlier P1 (`hourCycle` read from the host locale mangling `zh-CN` dates), P2 (prefix and clock resolving different locales), and P3 (one midnight timer per message) all disappear with the hand-rolled path rather than being patched individually — and #3678 was clarified in the same pass to state that a today-message's exact clock time and the year on same-year dates are not requirements. Every acceptance criterion is met: Astryx owns parsing, locale, `<time>` markup and labeling; the test pins both `/2 hours ago/` and `tabindex="0"` for the keyboard-reachable card; the reveal CSS and the touch gap are gone because the label is visible at rest.

**P3 — please add `isLive`.** A relative label needs to know the current time to stay correct, and `isLive` defaults to `false`, so the label freezes at mount. Open the app at 14:00 and a 12:00 message reads `2 hours ago` correctly; leave the window open until 17:00 and it still reads `2 hours ago`, until something unrelated re-renders it. Desktop windows stay open for days, so this is an ordinary path rather than a corner.

Worth noting where it comes from: the previous revision *did* refresh — via one midnight timer per message, which I flagged as thousands of timers firing at once. Removing the local implementation removed the refresh along with it, taking this from "too many timers" to "none". Astryx already solved it: `isLive` is the switch, and its own docs example is `<Timestamp value={date} format="auto" isLive />`. One prop, no timer code.

Not blocking and not in #3678's acceptance criteria — the hover card's absolute time stays correct either way — but the cost is a single attribute against a capability the component already ships, so I'd rather see it in than out.

---

*AI-assisted review: I read the full diff, #3678's current text, and the `Timestamp` type declaration to confirm the `isLive` default. No tests run. AI review is not independent human review.*

用一行 `format="auto"` 替换了 211 行本地实现,自上次评审净增 17 行、删除 831 行。先前的 P1/P2/P3 随手写路径一起消失,而不是被逐条修补;#3678 也在同一轮澄清为不要求当天消息的精确时钟与同年年份。验收标准全部满足。

**P3 — 建议加上 `isLive`。** 相对时间要知道"现在"才准,而 `isLive` 默认 `false`,标签在挂载时就定住了:14:00 打开应用,12:00 的消息显示 `2 hours ago` 正确;窗口开到 17:00,它仍显示 `2 hours ago`,直到有别的原因触发重渲染。Desktop 窗口经常连开数天,属正常路径。

来历值得一提:上一版**是**会刷新的——每条消息一个午夜定时器,我当时批评那是几千个定时器同时触发。删掉本地实现的同时也把刷新删掉了,从"定时器太多"变成"一个都没有"。而 Astryx 早已解决:`isLive` 就是开关,其文档示例正是 `<Timestamp value={date} format="auto" isLive />`,一个 prop,不需要任何定时器代码。

不阻塞,也不在 #3678 的验收标准内(hover card 的绝对时间始终正确),但代价只是一个属性,换取组件本就具备的能力,建议加上。

</details>

Astro-Han:
`b5ac0852b` closes the `isLive` note, and the mock-timer assertion (`2 hours ago` → `3 hours ago` after a one-hour tick) covers it better than I asked for — that pins the behavior rather than just the prop.

Nothing further from me. Merging once `test` goes green.

Sun-GLiang:
The failed check appears unrelated to this PR change.

The only failure is the known intermittent Desktop E2E contract tracked in #3177: streaming-remount.spec.ts timed out before observing the active .maka-bubble-streaming element. The same symptom recurred today on current main ancestry in https://github.com/apache/maka/actions/runs/33058450105, on changes unrelated to Turn admission, session projection, or streaming state.

This PR differs from main only in the user-message Timestamp rendering and its focused regression test. Lint, formatting, build, typecheck, affected workspace tests, and the other 74 Desktop E2E tests passed. I also reran the failing scenario locally 10 times and the complete streaming-remount spec 3 times; all 19 executions passed.

A rerun is appropriate to unblock this PR. A passing rerun would confirm the intermittent nature of this occurrence; the underlying active-Turn observation race remains tracked separately in #3177.

PR #3683, opened 2026-08-24
fix(desktop): preserve proxy auth intent in tests

### LINKED ISSUE #3682: bug(desktop): proxy test drops enabled authentication when credentials are empty
## What happened

Maka Desktop can report an authenticated proxy configuration as reachable even when no proxy credential is configured.

In **Settings → General → Network**, enable the proxy server and proxy authentication, configure an unauthenticated local proxy such as Clash at `127.0.0.1:7897`, leave the username and password empty, then select **Test current configuration**. The test reports **Proxy is reachable**.

This does not match actual execution. The saved Runtime Policy still has proxy authentication enabled, so model requests, WebFetch, and WebSearch fail closed when the proxy credential is missing.

Expected behavior:

- If proxy authentication is enabled and neither a saved password nor a newly supplied password is available, the proxy test fails with `Proxy credential is not configured`.
- The test does not open a network connection using an unauthenticated candidate.
- If a saved proxy password exists, the test resolves and uses that credential.
- Proxy testing with authentication disabled remains unchanged.

This is a false-positive readiness result, not an authentication bypass: actual model and tool execution still rejects the missing credential.

## How to reproduce

1. Start Maka Desktop from source.
2. Open **Settings → General → Network**.
3. Enable **Proxy server**.
4. Select the `HTTP/HTTPS` proxy protocol.
5. Set the server address to `127.0.0.1` and the port to `7897`.
6. Ensure the local Clash proxy accepts connections without proxy-level username/password authentication.
7. Enable **Proxy authentication** in Maka.
8. Leave both username and password empty and ensure Maka has no previously saved proxy password.
9. Select **Test current configuration**.
10. Observe that Maka reports the proxy as reachable.
11. Attempt a model request or WebFetch operation and observe that execution is rejected because the proxy credential is not configured.

## Environment

- Maka version or commit: `ad48130be` (`@maka/desktop@0.2.0`)
- OS and version: macOS 15.6
- Surface: Desktop development build / Runtime Host
- Node.js version: v24.19.0
- Proxy: Clash at `127.0.0.1:7897`
- Maka proxy authentication: enabled
- Proxy username/password: empty

The Settings test input loses the explicit authentication state before it reaches Runtime Host.

`apps/desktop/src/renderer/settings/general-settings-page.tsx` constructs the test candidate in `toProxyTestInput`. It conditionally includes username and password but does not include `authEnabled`:

```ts
username:
  proxy.authEnabled && proxy.username.trim()
    ? proxy.username.trim()
    : undefined,
password:
  proxy.authEnabled && proxy.password ? proxy.password : undefined,
```

When both fields are empty, `apps/desktop/src/main/runtime-host-settings-ipc-main.ts` reconstructs the authentication state from those values:

```ts
const username = proxy.username?.trim() ?? "";
const authEnabled = Boolean(username || proxy.password);
```

The explicit Settings state was `authEnabled: true`, but the Runtime Host test candidate receives `authEnabled: false`. Credential admission therefore treats the candidate as an unauthenticated proxy and probes `127.0.0.1:7897` successfully.

The real execution path uses the saved Runtime Policy, where `authEnabled` remains true. Runtime Host consequently requires the `network_proxy` credential and returns `credential_not_configured` when it is absent.

Suggested direction:

- Add an optional authentication-intent field to the test-specific proxy candidate.
- Have the Settings renderer explicitly pass the current `authEnabled` value.
- Make the Desktop IPC adapter prefer the explicit value.
- Retain the current value-based inference only as a compatibility fallback for callers that omit the new field.
- Keep Runtime Host as the authoritative credential-admission boundary.
- Do not change proxy transport, credential persistence, or user-facing copy.

Suggested regression coverage:

- Authentication enabled with no stored or supplied password preserves `authEnabled: true` and fails the test.
- Authentication disabled for an unauthenticated local proxy preserves `authEnabled: false` and can still report reachable.
- Authentication enabled with a stored password resolves that credential and proceeds to the proxy probe.
- An explicit `false` value must not be replaced by credential-value inference.

AI-assisted investigation disclosure: Codex helped reproduce the proxy-test behavior in the Desktop development build, trace the Settings → Desktop IPC → Runtime Host data flow, compare the proxy-test path with model and WebFetch credential admission, run targeted diagnostics, and draft this report. I reviewed the findings and own the submission.

Sun-GLiang:
## Summary

- carry the Settings draft's explicit proxy authentication state through the test-only proxy candidate
- make the Desktop IPC adapter prefer explicit authentication intent while retaining value-based inference for legacy callers
- add regression coverage proving missing credentials fail when authentication is enabled and unauthenticated local proxy tests remain unchanged

Fixes #3682

## Verification

- `npm run lint` — 2,672 files checked, no fixes required
- `npm run format:check` — 1,615 files checked, no fixes required
- `node --test apps/desktop/dist/main/__tests__/runtime-host-settings-ipc-main.test.js` — 2 passed, 0 failed
- `node --test --test-reporter=dot "apps/desktop/dist/main/**/*.test.js"`
- `git diff --check upstream/main...HEAD`

The regression test was run before the production change and failed on the enabled-authentication assertion (`false !== true`), then passed after the fix.

## AI use

Select exactly one:

Tool(s) and scope: OpenAI Codex reproduced and diagnosed the proxy-test inconsistency, implemented the contract and IPC fix, added regression coverage, ran verification, and drafted the issue and PR text. The human contributor of record reviewed the proposed behavior and owns the contribution.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/src/main/__tests__/runtime-host-settings-ipc-main.test.ts
+++ b/apps/desktop/src/main/__tests__/runtime-host-settings-ipc-main.test.ts
@@ -1,0 +1,97 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import assert from "node:assert/strict";
+import test from "node:test";
+import {
+  createDefaultRuntimePolicy,
+  type RuntimePolicy,
+} from "@maka/core/runtime-policy";
+import { registerRuntimeHostSettingsIpc } from "../runtime-host-settings-ipc-main.js";
+
+type TestCandidate = RuntimePolicy["networkProxy"];
+
+async function testCandidate(authEnabled: boolean): Promise<{
+  candidate: TestCandidate;
+  result: { ok: boolean; code: string };
+}> {
+  const handlers = new Map<string, (...args: unknown[]) => unknown>();
+  const policy = createDefaultRuntimePolicy();
+  let candidate: TestCandidate | undefined;
+
+  registerRuntimeHostSettingsIpc({
+    ipcMain: {
+      handle(channel, listener) {
+        handlers.set(channel, listener as (...args: unknown[]) => unknown);
+      },
+    },
+    client: {
+      async queryRuntimePolicy() {
+        return { revision: 0, policy };
+      },
+      async testNetworkProxy(input: { networkProxy?: TestCandidate }) {
+        candidate = input.networkProxy;
+        return candidate?.authEnabled
+          ? {
+              ok: false,
+              latencyMs: 0,
+              error: "Proxy credential is not configured",
+            }
+          : { ok: true, latencyMs: 1, status: 200 };
+      },
+    } as never,
+    settingsStore: {} as never,
+    async applyClientSettings() {},
+  });
+
+  const handler = handlers.get("settings:testNetworkProxy");
+  assert.ok(handler);
+  const result = await handler({}, {
+    proxy: {
+      enabled: true,
+      type: "http",
+      host: "127.0.0.1",
+      port: 7897,
+      authEnabled,
+      bypassList: [],
+    },
+  });
+
+  assert.ok(candidate);
+  return {
+    candidate,
+    result: result as { ok: boolean; code: string },
+  };
+}
+
+test("proxy test preserves enabled authentication when credentials are empty", async () => {
+  const tested = await testCandidate(true);
+
+  assert.equal(tested.candidate.authEnabled, true);
+  assert.equal(tested.result.ok, false);
+  assert.equal(tested.result.code, "proxy_unreachable");
+});
+
+test("proxy test preserves disabled authentication for a local proxy", async () => {
+  const tested = await testCandidate(false);
+
+  assert.equal(tested.candidate.authEnabled, false);
+  assert.equal(tested.result.ok, true);
+  assert.equal(tested.result.code, "proxy_reachable");
+});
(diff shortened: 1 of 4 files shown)


Sun-GLiang:
Could a maintainer please trigger CI for this PR? Thanks!

Astro-Han (commented):
Reviewed at `ec3b38b3ca`. No P0–P3. One gate note.

The fix is in the right place. `authEnabled` is already a first-class field on the persisted policy (`runtime-policy.ts:119`, `settings.ts:99`) and is validated as a required boolean on the Runtime Host side (`network-proxy.ts:104`). Only the Test Connection path was reconstructing it, by inferring `Boolean(username || password)`. So a user who turned proxy auth on but had not yet typed a credential got a test that quietly ran *without* auth and reported success — a green result for a configuration the real connection would not use. `proxy.authEnabled ?? Boolean(username || proxy.password)` restores the declared intent while leaving callers that omit the field on the old inference, so nothing else has to change at once.

I checked that `toProxyTestInput` in `general-settings-page.tsx:796` is the only production construction of this input, so there is no second path still inferring, and the new `TestProxySettings` widening is confined to the test-input type rather than loosening `ProxySettings` itself.

The new tests assert the failing direction — with `authEnabled` true and no credential, the candidate reaches the client with auth on and the result is an error — so the original defect would still be caught. This is a fix to the code, not a relaxation of the assertions.

**Gate:** no check has run on this head at all. Never-run is not green, so this is not approvable yet — please push or retrigger so `test` produces a result.

<details>
<summary>简体中文</summary>

已在 `ec3b38b3ca` 上审查。无 P0–P3，一条门禁说明。

修复的位置是对的。`authEnabled` 本来就是持久化策略上的一等字段（`runtime-policy.ts:119`、`settings.ts:99`），在 Runtime Host 侧还被校验为必需的布尔值（`network-proxy.ts:104`）。只有「测试连接」这条路径在重新推断它——用 `Boolean(username || password)`。于是一个打开了代理认证、但还没填凭据的用户，得到的是一次**不带认证**悄悄跑完并报成功的测试——对一个真实连接根本不会采用的配置给出了绿灯。改成 `proxy.authEnabled ?? Boolean(username || proxy.password)` 恢复了声明的意图，同时让不传该字段的调用方保持旧推断，不必一次性全改。

我核过 `general-settings-page.tsx:796` 的 `toProxyTestInput` 是这个入参在生产代码中唯一的构造点，因此不存在第二条仍在推断的路径；新增的 `TestProxySettings` 只放宽了测试入参类型，没有放松 `ProxySettings` 本身。

新增测试断言的是**失败方向**——`authEnabled` 为真且无凭据时，候选带着认证到达客户端并返回错误——因此原来的缺陷改动后仍会被抓住。这是把代码修好，不是把断言放松。

**门禁**：这个 head 上一条 check 都没跑过。「没跑过」不等于绿，因此现在还不具备 approve 条件——请推一次或重新触发，让 `test` 出结果。
</details>

Astro-Han (approved):
I verified this fix preserves proxy auth intent without loosening tests.

The production change in `runtime-host-settings-ipc-main.ts` now respects an explicit `authEnabled` flag instead of inferring it from username/password presence. I checked that `authEnabled` is a persisted, validated field and the only production entry point (`toProxyTestInput`) constructs it — so there is no second path still guessing.

The added test (`runtime-host-settings-ipc-main.test.ts`) keeps failure coverage for the original bug (auth intent lost when inferring), not relaxing assertions.

No blocking issues on this head. Checks on `4fab9a81c1` are `test: success`.

<details><summary>简体中文</summary>已验证代理认证意图保留，测试未放宽。</details>

PR #3684, opened 2026-08-24
fix(ui): abbreviate sidebar timestamps

ARE404:
## Summary

- Abbreviate session-sidebar timestamps as `min`, `h`, `d`, `mo`, and `y`, matching the compact Codex-style display.
- Keep the localized “just now” state and preserve the existing compact date format used outside the sidebar.
- Refresh each timestamp at the next visible rounding boundary so labels do not become stale.

## Verification

- `npm --workspace @maka/core test`
- `npm --workspace @maka/ui test`
- `npm run lint`
- `npm run format:check`
- `npm run build`
- `npm run typecheck`
- Storybook: `Product/Sidebar Session List / Long Titles and Narrow` rendered `52min`, `6min`, and `31min`.

## AI use

Select exactly one:

Tool(s) and scope: Codex assisted with diagnosis, implementation, tests, and adversarial review.

## Checklist

- [ ] No

### DIFF
--- a/packages/ui/src/session-history-list.tsx
+++ b/packages/ui/src/session-history-list.tsx
@@ -466,7 +466,7 @@ const SessionNavRow = memo(function SessionNavRow(props: {
               {props.session.lastMessageAt ? (
                 <RelativeTime
                   ts={props.session.lastMessageAt}
-                  variant="compact"
+                  variant="sidebar"
                   className="maka-session-row-time-label"
                   suppressTitle
                 />
--- a/packages/core/src/__tests__/relative-time.test.ts
+++ b/packages/core/src/__tests__/relative-time.test.ts
@@ -22,7 +22,9 @@ import { describe, it } from 'node:test';
 import {
   formatCompactTimestamp,
   formatRelativeTimestamp,
+  formatSidebarTimestamp,
   nextRelativeRefreshDelay,
+  nextSidebarRefreshDelay,
   resetRelativeTimeFormatters,
 } from '../relative-time.js';
 
@@ -37,6 +39,7 @@ describe('relative timestamp labels', () => {
       assert.equal(formatRelativeTimestamp(NOW - ageMs, NOW, 'zh'), '刚刚');
       assert.equal(formatRelativeTimestamp(NOW - ageMs, NOW, 'en'), 'just now');
       assert.equal(formatCompactTimestamp(NOW - ageMs, NOW, 'zh'), '刚刚');
+      assert.equal(formatSidebarTimestamp(NOW - ageMs, NOW, 'zh'), '刚刚');
     }
 
     assert.equal(formatRelativeTimestamp(NOW - 60_000, NOW, 'zh'), '1分钟前');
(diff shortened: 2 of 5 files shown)


packages/ui/src/session-history-list.tsx
```diff
@@ -466,7 +466,7 @@ const SessionNavRow = memo(function SessionNavRow(props: {
               {props.session.lastMessageAt ? (
                 <RelativeTime
                   ts={props.session.lastMessageAt}
-                  variant="compact"
+                  variant="sidebar"
```
Astro-Han, line 469: **[P3] This is a UI change without before/after evidence.**

`CONTRIBUTING.md` asks for before/after screenshots or a recording on UI changes. The Verification section here lists Storybook text output (`52min` / `6min` / `31min`), which is neither a screenshot nor a recording, and shows only the after state — there is nothing to compare the new abbreviations against.

Not blocking, and I am not asking you to redo the work: if you would rather not produce them by hand, say so and I will generate the pair from the base and head commits and post them here.

<details>
<summary>中文</summary>

**[P3] 界面改动缺前后对比证据。**

`CONTRIBUTING.md` 要求界面改动附前后对比截图或录屏。这里 Verification 一节给的是 Storybook 的文字输出（`52min` / `6min` / `31min`），既不是截图也不是录屏，而且只有"后"没有"前"，没法对照。

不阻塞，也不用你重做：如果不想手动截，说一声，我用 base 和 head 各生成一张贴上来。
</details>

Astro-Han (approved):
Approving. I checked the threshold and rounding boundaries, the non-sidebar call sites, the refresh scheduling, and the timer lifecycle, and found nothing blocking. `test` is green on `51daa18e`, and it is the only check this head triggers.

Two things I verified rather than took on trust:

**The rounding is inherited, not new.** `Math.round` means 23h30m renders as `1d`, 29d12h as `1mo`, and 11.5mo as `1y`. That is the same half-up behaviour `formatRelativeTimestamp` already has (`relative-time.ts:107-113`), so it is existing semantics, not something this PR introduces.

**The refresh timer cannot spin.** Future timestamps and clock skew are clamped to 0 (`:74-75`), so the delay is always a finite positive number. Sweeping 40 days found a 20ms minimum, occurring only within ±ε of a rounding boundary; once it fires, the age crosses the boundary and the next delay returns to at least half a unit. It is a single fast tick, not a loop. The 24h cap stays under the browser timer ceiling.

I also ran the tests against a reverted implementation: the two new sidebar cases genuinely fail without this change and pass with it.

One non-blocking gap is inline.

<details>
<summary>中文</summary>

批准。阈值与取整边界、非 sidebar 的调用点、刷新调度、timer 生命周期都核过，没有阻塞问题；`51daa18e` 上 `test` 已绿，且这是该 head 唯一触发的 check。

两点我是实测而非采信的：**取整行为是继承的不是新增的**——`Math.round` 导致 23h30m 显示 `1d`、29d12h 显示 `1mo`、11.5mo 显示 `1y`，与 `relative-time.ts:107-113` 既有语义一致；**刷新 timer 不会空转**——未来时间与时钟偏移被 clamp 到 0（`:74-75`），扫 40 天最小 delay 为 20ms，只出现在取整边界 ±ε 处，触发一次后立即回到至少半个单位，是单发快跳不是循环，24h 上限也低于浏览器 timer 天花板。

另外把实现退回旧版重跑，新增的两条 sidebar 测试确实会红，恢复后复绿。

一条不阻塞的缺口见行内。
</details>

packages/ui/src/session-history-list.tsx
ARE404: before:
image
after:
Snipaste_2026-08-24_16-36-48

Astro-Han:
LGTM, merging.

I trial-merged this into the current main and checked the affected surface: `@maka/core` and `@maka/ui` typecheck cleanly, and the 7 `relative-time` plus 7 `session-history-row-actions` tests all pass.

On the rename from `variant="compact"` to `variant="sidebar"`: `session-history-list.tsx` is the only call site in the repository that passes `variant` at all. Every other `RelativeTime` caller — including `provider-connection-detail.tsx`, newly added on main — uses the default variant and is unaffected.

<details>
<summary>简体中文</summary>

LGTM，合并中。

我在当前 main 上做了一次试合并并验证了受影响范围：`@maka/core` 与 `@maka/ui` 类型检查通过，`relative-time` 7 项、`session-history-row-actions` 7 项测试全过。

`variant` 从 `compact` 改名为 `sidebar` 只影响 `session-history-list.tsx` 这一处调用——它是全仓库唯一传 `variant` 的地方。其余 `RelativeTime` 调用方（含 main 上新增的 `provider-connection-detail.tsx`）都走默认 variant，不受影响。

</details>

PR #3687, opened 2026-08-24
feat(runtime-host): apply verified managed updates

### LINKED ISSUE #3679: feat(runtime-host): apply discovered managed Host updates safely
<details open>
<summary>English</summary>

## Problem

Managed Runtime Hosts can discover an immutable update candidate and can already switch safely to a package that has been supplied explicitly, but these two paths are not connected. A future scheduler cannot turn a discovered exact version and integrity into the self-contained package consumed by the existing update transaction.

This gap is especially important before the first ASF npm convenience release: that package must itself be able to acquire and verify a later compatible release instead of requiring a replacement updater to be installed first.

## Desired outcome

Add one bounded update invocation that:

- accepts `latest`, `next`, or an exact version;
- resolves the selector once through the read-only discovery contract;
- downloads the exact package from the official npm registry and verifies its SHA-512 integrity;
- validates the extracted package version and compatibility evidence;
- delegates stage, retirement, cutover, readiness verification, and recovery to the existing exact-package update transaction;
- lets the exact transaction verify a current candidate by deployment identity, clean inactive packages, and restore service/operator lifecycle state for an intact stopped or unready current deployment;
- performs no package acquisition or mutation for a non-admitted candidate;
- never interrupts active Host-owned work without the existing explicit authority.

The implementation must be qualified with release-shaped packages so the first ASF npm convenience release can be proven to update to a later development candidate before publication.

Published package directories are immutable under Maka's deployment authority. Lifecycle repair assumes that the managed package tree remains intact; this issue does not add local tree attestation or recovery from external package modification.

This issue does not add a persistent update policy, scheduler, Desktop UI, or another deployment state machine. It is part of #3228 and follows #3664/#3667.

An operator can currently run the target package directly, and Desktop can upload the exact package selected for its own build. Those remain useful explicit maintenance paths, but neither provides the bounded registry-to-existing-transaction bridge required by later `fixed`, `latest`, or `next` policies.

</details>

<details>
<summary>简体中文</summary>

## 问题

Managed Runtime Host 已经可以发现不可变更新候选，也能安全切换到一个被显式提供的 package，但这两条路径尚未连接。未来的 scheduler 还不能把发现得到的精确版本与 integrity 转换成现有更新事务所消费的 self-contained package。

这个缺口在第一次 ASF npm convenience release 前尤其重要：第一版 package 自身必须能够获取并验证后续兼容版本，而不能要求用户先安装一个更新后的 updater。

## 期望结果

增加一个有界的一次性更新入口：

- 接受 `latest`、`next` 或精确版本；
- 通过只读 discovery contract 将 selector 解析一次；
- 从官方 npm registry 下载精确 package，并验证 SHA-512 integrity；
- 校验解包后的 package version 与 compatibility evidence；
- 将 stage、retirement、cutover、readiness verification 与 recovery 委托给现有 exact-package update transaction；
- 由精确事务按 deployment identity 核验当前候选、清理 inactive package，并在 managed package tree 完整的前提下恢复 stopped/unready current 的 service/operator lifecycle；
- 未获准自动应用的候选不获取 package，也不发生 mutation；
- 没有现有的明确授权时，绝不中断 Host 拥有的 active work。

实现必须使用 release-shaped package 验证，使第一次 ASF npm convenience release 在发布前即可证明能更新到后续 development candidate。

已发布的 package directory 在 Maka deployment authority 内按不可变对象处理。Lifecycle repair 假定 managed package tree 仍然完整；本 issue 不增加本地 tree attestation，也不负责从外部 package 修改中恢复。

本 issue 不增加持久 update policy、scheduler、Desktop UI 或另一套 deployment state machine。属于 #3228，承接 #3664/#3667。

## 替代方案或当前做法

operator 当前可以直接运行 target package，Desktop 也可以上传自身 build 选择的精确 package。这些仍是有用的显式维护路径，但都不能提供后续 `fixed`、`latest` 或 `next` 策略所需的、从 registry 到现有事务的有界桥接。

</details>

M4n5ter:
<details open>
<summary>English</summary>

## Summary

- connect managed Host update discovery to the existing exact-package update transaction
- download the exact `maka-agent` archive from the official npm registry and verify its SHA-512 integrity
- bind version, compatibility, and current/target deployment identities to the transaction so stale or same-version different-content candidates cannot be mistaken for current
- keep non-admitted candidates outside package acquisition and mutation; let the exact transaction return a healthy current deployment, restore service/operator lifecycle state for an intact stopped or unready current deployment, and safely recover interrupted package cleanup

Published package directories are immutable under Maka's deployment authority. Lifecycle repair assumes that the managed package tree remains intact; local tree attestation and recovery from external package modification are outside this PR.

Fixes #3679

## Verification

- `npm --workspace maka-agent test` — 443 tests passed
- `npm run typecheck`, `npm run lint`, `npm run format:check`, and `npm run build`
- `npm run release:cli:pack` — built the self-contained `maka-agent@0.2.0` tarball from the clean head
- `node scripts/smoke-release-cli-package.mjs packages/cli/release/maka-agent-0.2.0.tgz` — offline installation, TUI setup, managed Runtime Host lifecycle, and controlled Turn passed
- `npm run check:asf-headers` and `git diff --check`

## Release gate

The first ASF npm convenience release must contain this update entry point together with the compatibility evidence introduced by #3667. Before publishing that release, the same release-shaped successor exercise should be repeated against the immutable release candidate artifact.

## AI use

Select exactly one:

Tool(s) and scope: Codex authored the implementation, focused tests, and documentation under maintainer direction

## Checklist

- [ ] No

</details>

<details>
<summary>简体中文</summary>

## 摘要

- 将 managed Host update discovery 接入现有的精确 package 更新事务
- 从官方 npm registry 下载精确的 `maka-agent` archive，并校验其 SHA-512 integrity
- 将 version、compatibility 以及当前与目标 deployment identity 一并绑定到事务，避免把过期候选或同版本不同内容误判为 current
- 未获准自动应用的候选不获取 package、也不发生 mutation；精确事务负责返回健康 current，在 managed package tree 完整的前提下恢复 stopped/unready current 的 service/operator lifecycle，并安全恢复中断的 package 清理

已发布的 package directory 在 Maka deployment authority 内按不可变对象处理。Lifecycle repair 假定 managed package tree 仍然完整；本 PR 不增加本地 tree attestation，也不负责从外部 package 修改中恢复。

关联 #3679

## 验证

- `npm --workspace maka-agent test` — 443 项测试通过
- `npm run typecheck`、`npm run lint`、`npm run format:check` 与 `npm run build`
- `npm run release:cli:pack` — 从干净 head 构建 self-contained `maka-agent@0.2.0` tarball
- `node scripts/smoke-release-cli-package.mjs packages/cli/release/maka-agent-0.2.0.tgz` — 离线安装、TUI setup、managed Runtime Host 生命周期与受控 Turn 均通过
- `npm run check:asf-headers` 与 `git diff --check`

## 发布门槛

第一次 ASF npm convenience release 必须同时包含本 PR 的更新入口和 #3667 引入的 compatibility evidence。正式发布前，还应使用不可变 release candidate artifact 重复同一套发布形态的后续版本升级演练。

## AI 使用

- [ ] 没有生成式工具作出实质贡献
- [x] 生成式工具作出了实质贡献

工具与范围：Codex 在维护者指导下完成实现、聚焦测试与文档

## 检查清单

- [x] 测试覆盖该变更，且缺少实现时会失败
- [x] lint、format、typecheck 与受影响测试套件均在本地通过

本 PR 是否改变行为？

- [x] 是，具体见上方摘要
- [ ] 否

</details>

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -133,9 +133,11 @@ maka runtime-host service check-update --target next --json
 ```
 
 The result pins the selected channel to an exact version and package integrity. It also reports
-whether the package carries enough compatibility evidence for a future unattended update; this
-command never installs or switches a package. An updater must still verify the downloaded archive
-against that integrity and confirm the compatibility value from its extracted package manifest.
+whether the package carries enough compatibility evidence for unattended use; this command never
+installs or switches a package. Installation-management callers can pass the same selector to
+`service update --target`. That path verifies the archive and extracted manifest before delegating
+to the existing exact-package update transaction, and does not mutate a candidate that requires
+manual review.
 
 ## Uninstall
 
--- a/packages/cli/README.zh-CN.md
+++ b/packages/cli/README.zh-CN.md
@@ -125,8 +125,9 @@ maka runtime-host service check-update --target next --json
 ```
 
 结果会把频道固定为精确版本和 package integrity，并说明 package 是否提供足够的兼容性证据，
-可供未来的无人值守更新使用；该命令不会安装或切换 package。更新程序仍须以该 integrity
-校验下载的 archive，并从解压后的 package manifest 中再次确认兼容性值。
+可供无人值守流程使用；该命令不会安装或切换 package。安装管理方可以把同一
+selector 传给 `service update --target`。该路径会先校验 archive 与解包后的 manifest，再委托给
+现有的精确 package 更新事务；需要人工审查的候选不会改变当前 Host。
 
 ## 卸载
 
--- a/packages/cli/src/__tests__/runtime-host-launch-agent-service.test.ts
+++ b/packages/cli/src/__tests__/runtime-host-launch-agent-service.test.ts
@@ -117,27 +117,29 @@ test('maps install, stop, start, restart, and uninstall onto one LaunchAgent ser
   });
 });
 
-test('restores the previous loaded LaunchAgent when replacement bootstrap fails', async () => {
-  await withFixture(async ({ homeDir, cliPath, launchctl }) => {
-    const plistPath = resolveLaunchAgentPath(SERVICE_ID, homeDir);
-    const previousPlist = '<plist>previous</plist>\n';
-    await writeFile(plistPath, previousPlist, { mode: 0o600 });
-    launchctl.loaded = true;
-    launchctl.failNextBootstrap = true;
-    const backend = createLaunchAgentRuntimeHostService(SERVICE_ID, {
-      homeDir,
-      uid: UID,
-      runLaunchctl: launchctl.run,
-      isProcessAlive: () => false,
-    });
+test('restores the previous loaded LaunchAgent when deployment bootstrap fails', async () => {
+  for (const action of ['install', 'replace'] as const) {
+    await withFixture(async ({ homeDir, cliPath, launchctl }) => {
+      const plistPath = resolveLaunchAgentPath(SERVICE_ID, homeDir);
+      const previousPlist = '<plist>previous</plist>\n';
+      await writeFile(plistPath, previousPlist, { mode: 0o600 });
+      launchctl.loaded = true;
+      launchctl.failNextBootstrap = true;
+      const backend = createLaunchAgentRuntimeHostService(SERVICE_ID, {
+        homeDir,
+        uid: UID,
+        runLaunchctl: launchctl.run,
+        isProcessAlive: () => false,
+      });
 
-    await assert.rejects(
-      backend.install(fixtureConfig(process.execPath, cliPath, join(homeDir, 'state'))),
-      /Starting the Runtime Host LaunchAgent failed/u,
-    );
-    assert.equal(await readFile(plistPath, 'utf8'), previousPlist);
-    assert.equal(launchctl.loaded, true);
-  });
+      await assert.rejects(
+        backend[action](fixtureConfig(process.execPath, cliPath, join(homeDir, 'state'))),
+        /Starting the Runtime Host LaunchAgent failed/u,
+      );
+      assert.equal(await readFile(plistPath, 'utf8'), previousPlist);
+      assert.equal(launchctl.loaded, true);
+    });
+  }
 });
 
 interface FakeLaunchctl {
--- a/packages/cli/src/__tests__/runtime-host-selected-update.test.ts
+++ b/packages/cli/src/__tests__/runtime-host-selected-update.test.ts
@@ -26,7 +26,7 @@ import {
   type RuntimeHostUpdateCliOptions,
 } from '../runtime-host-update-command.js';
 import { parseRuntimeHostCommand } from '../runtime-host-cli.js';
-import type { RuntimeHostUpdateCheckResolution } from '../runtime-host-update-discovery.js';
+import type { RuntimeHostUpdateSelection } from '../runtime-host-update-discovery.js';
 
 const INTEGRITY =
   'sha512-jUKdo/5dbM94KXq+kOZ1d+obhDLAENfI/QWr1PnXWcdu2PqDyLklJBtiVO6HRwoL1l40z1NE9Rq+hLAxCN0Fyg==';
@@ -83,19 +83,16 @@ describe('managed Runtime Host selected update', () => {
   });
 
   it('hands one verified admitted package to the existing update transaction', async () => {
-    const resolution = updateResolution({ kind: 'unattended_update', compatibility: 7 });
-    let cleanupCalls = 0;
+    const selection = updateSelection({
+      kind: 'unattended_update',
+      compatibility: 7,
+    });
     let updateInput: RuntimeHostUpdateCliOptions | undefined;
     const exitCode = await runManagedRuntimeHostSelectedUpdateCli(OPTIONS, {
-      resolveCheck: async () => resolution,
-      acquire: async (candidate) => {
-        assert.deepEqual(candidate, resolution.candidate);
-        return {
-          root: '/verified/package',
-          cleanup: async () => {
-            cleanupCalls += 1;
-          },
-        };
+      resolveSelection: async () => selection,
+      withPackage: async (candidate, use) => {
+        assert.deepEqual(candidate, selection.candidate);
+        return use('/verified/package');
       },
       update: async (input) => {
         updateInput = input;
@@ -106,64 +103,68 @@ describe('managed Runtime Host selected update', () => {
     assert.equal(updateInput?.sourcePackageRoot, '/verified/package');
     assert.equal(updateInput?.version, '2.0.0');
     assert.equal(updateInput?.expectedCurrentVersion, '1.0.0');
-    assert.equal(cleanupCalls, 1);
+    assert.equal(updateInput?.packageIntegrity, INTEGRITY);
   });
 
-  it('keeps current and non-admitted candidates outside the mutation path', async () => {
-    for (const outcome of [
-      { kind: 'current' as const },
-      { kind: 'manual_action' as const, reason: 'compatibility_mismatch' as const },
-    ]) {
-      let output = '';
-      const exitCode = await runManagedRuntimeHostSelectedUpdateCli(OPTIONS, {
-        resolveCheck: async () => updateResolution(outcome),
-        acquire: async () => assert.fail('package acquisition is not expected'),
-        update: async () => assert.fail('the update transaction is not expected'),
-        writeOutput: (value) => {
-          output += value;
+  it('lets the exact transaction decide whether a current candidate needs repair', async () => {
+    const selection = updateSelection({ kind: 'current' });
+    let updates = 0;
+    assert.equal(
+      await runManagedRuntimeHostSelectedUpdateCli(OPTIONS, {
+        resolveSelection: async () => selection,
+        withPackage: async (_candidate, use) => use('/verified/package'),
+        update: async () => {
+          updates += 1;
+          return 0;
         },
-      });
-      const frame = decodeRuntimeHostServiceManagementFrame(output.trim());
-      if (outcome.kind === 'current') {
-        assert.equal(exitCode, 0);
-        assert.equal(
-          frame?.kind === 'result' && frame.action === 'update' ? frame.update.kind : undefined,
-          'already_current',
-        );
-      } else {
-        assert.equal(exitCode, 1);
-        assert.equal(frame?.kind === 'error' ? frame.error.code : undefined, 'update_not_admitted');
-      }
-    }
+      }),
+      0,
+    );
+    assert.equal(updates, 1);
+  });
+
+  it('keeps non-admitted candidates outside package acquisition and mutation', async () => {
+    let output = '';
+    const exitCode = await runManagedRuntimeHostSelectedUpdateCli(OPTIONS, {
+      resolveSelection: async () =>
+        updateSelection({
+          kind: 'manual_action',
+          reason: 'compatibility_mismatch',
+        }),
+      withPackage: async () => assert.fail('package acquisition is not expected'),
+      update: async () => assert.fail('the update transaction is not expected'),
+      writeOutput: (value) => {
+        output += value;
+      },
+    });
+    const frame = decodeRuntimeHostServiceManagementFrame(output.trim());
+    assert.equal(exitCode, 1);
+    assert.equal(frame?.kind === 'error' ? frame.error.code : undefined, 'update_not_admitted');
   });
 });
 
-function updateResolution(
-  outcome: RuntimeHostUpdateCheckResolution['frame']['updateCheck']['outcome'],
-): RuntimeHostUpdateCheckResolution {
-  const candidate = { version: '2.0.0', integrity: INTEGRITY, compatibility: 7 };
+function updateSelection(
+  outcome: RuntimeHostUpdateSelection['outcome'],
+): RuntimeHostUpdateSelection {
+  const candidate = {
+    version: '2.0.0',
+    integrity: INTEGRITY,
+    compatibility: 7,
+  };
   return {
+    selector: OPTIONS.selector,
     candidate,
-    frame: {
-      schemaVersion: 1,
-      kind: 'result',
-      action: 'check_update',
-      service: {
-        platform: 'linux',
-        arch: 'x64',
-        osRelease: 'test',
-        state: 'running',
-        pid: 42,
-        lastExitCode: null,
-        installedVersion: outcome.kind === 'current' ? candidate.version : '1.0.0',
-        stateRoot: TARGET.rootPath,
-        projectDirectoryRoots: [],
-      },
-      updateCheck: {
-        selector: OPTIONS.selector,
-        candidate: { version: candidate.version, integrity: candidate.integrity },
-        outcome,
-      },
+    outcome,
+    service: {
+      platform: 'linux',
+      arch: 'x64',
+      osRelease: 'test',
+      state: 'running',
+      pid: 42,
+      lastExitCode: null,
+      installedVersion: outcome.kind === 'current' ? candidate.version : '1.0.0',
+      stateRoot: TARGET.rootPath,
+      projectDirectoryRoots: [],
     },
   };
 }
--- a/packages/cli/src/__tests__/runtime-host-selected-update.test.ts
+++ b/packages/cli/src/__tests__/runtime-host-selected-update.test.ts
@@ -113,19 +113,19 @@ describe('managed Runtime Host selected update', () => {
 
   it('lets the exact transaction decide whether a current candidate needs repair', async () => {
     const selection = updateSelection({ kind: 'current' });
-    let updates = 0;
+    let updateInput: RuntimeHostUpdateCliOptions | undefined;
     assert.equal(
       await runManagedRuntimeHostSelectedUpdateCli(OPTIONS, {
         resolveSelection: async () => selection,
-        withPackage: async (_candidate, use) => use('/verified/package'),
-        update: async () => {
-          updates += 1;
+        withPackage: async () => assert.fail('the current deployment must not be downloaded'),
+        update: async (input) => {
+          updateInput = input;
           return 0;
         },
       }),
       0,
     );
-    assert.equal(updates, 1);
+    assert.equal(updateInput?.sourcePackageRoot, '/managed/versions/2.0.0');
   });
 
   it('keeps non-admitted candidates outside package acquisition and mutation', async () => {
--- a/packages/cli/src/__tests__/runtime-host-selected-update.test.ts
+++ b/packages/cli/src/__tests__/runtime-host-selected-update.test.ts
@@ -1,0 +1,169 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import assert from 'node:assert/strict';
+import { describe, it } from 'node:test';
+import { decodeRuntimeHostServiceManagementFrame } from '@maka/runtime-host/operator';
+import {
+  runManagedRuntimeHostSelectedUpdateCli,
+  type RuntimeHostSelectedUpdateCliOptions,
+  type RuntimeHostUpdateCliOptions,
+} from '../runtime-host-update-command.js';
+import { parseRuntimeHostCommand } from '../runtime-host-cli.js';
+import type { RuntimeHostUpdateCheckResolution } from '../runtime-host-update-discovery.js';
+
+const INTEGRITY =
+  'sha512-jUKdo/5dbM94KXq+kOZ1d+obhDLAENfI/QWr1PnXWcdu2PqDyLklJBtiVO6HRwoL1l40z1NE9Rq+hLAxCN0Fyg==';
+const TARGET = {
+  serviceId: 'b'.repeat(64),
+  rootPath: '/srv/maka',
+  rootId: 'a'.repeat(64),
+};
+const OPTIONS: RuntimeHostSelectedUpdateCliOptions = {
+  json: false,
+  framed: true,
+  clientDataRoot: '/client',
+  defaultRootPath: '/workspace',
+  selector: { kind: 'channel', channel: 'next' },
+  expectedTarget: TARGET,
+};
+
+describe('managed Runtime Host selected update', () => {
+  it('parses an optional target without changing the exact-package command', () => {
+    assert.deepEqual(
+      parseRuntimeHostCommand([
+        'service',
+        'update',
+        '--target',
+        'next',
+        '--expected-service-id',
+        TARGET.serviceId,
+        '--expected-root-path',
+        TARGET.rootPath,
+        '--expected-root-id',
+        TARGET.rootId,
+      ]),
+      {
+        kind: 'runtime-host-service-update',
+        json: false,
+        selector: { kind: 'channel', channel: 'next' },
+        expectedTarget: TARGET,
+      },
+    );
+    assert.equal(
+      'selector' in
+        parseRuntimeHostCommand([
+          'service',
+          'update',
+          '--expected-service-id',
+          TARGET.serviceId,
+          '--expected-root-path',
+          TARGET.rootPath,
+          '--expected-root-id',
+          TARGET.rootId,
+        ]),
+      false,
+    );
+  });
+
+  it('hands one verified admitted package to the existing update transaction', async () => {
+    const resolution = updateResolution({ kind: 'unattended_update', compatibility: 7 });
+    let cleanupCalls = 0;
+    let updateInput: RuntimeHostUpdateCliOptions | undefined;
+    const exitCode = await runManagedRuntimeHostSelectedUpdateCli(OPTIONS, {
+      resolveCheck: async () => resolution,
+      acquire: async (candidate) => {
+        assert.deepEqual(candidate, resolution.candidate);
+        return {
+          root: '/verified/package',
+          cleanup: async () => {
+            cleanupCalls += 1;
+          },
+        };
+      },
+      update: async (input) => {
+        updateInput = input;
+        return 0;
+      },
+    });
+    assert.equal(exitCode, 0);
+    assert.equal(updateInput?.sourcePackageRoot, '/verified/package');
+    assert.equal(updateInput?.version, '2.0.0');
+    assert.equal(updateInput?.expectedCurrentVersion, '1.0.0');
+    assert.equal(cleanupCalls, 1);
+  });
+
+  it('keeps current and non-admitted candidates outside the mutation path', async () => {
+    for (const outcome of [
+      { kind: 'current' as const },
+      { kind: 'manual_action' as const, reason: 'compatibility_mismatch' as const },
+    ]) {
+      let output = '';
+      const exitCode = await runManagedRuntimeHostSelectedUpdateCli(OPTIONS, {
+        resolveCheck: async () => updateResolution(outcome),
+        acquire: async () => assert.fail('package acquisition is not expected'),
+        update: async () => assert.fail('the update transaction is not expected'),
+        writeOutput: (value) => {
+          output += value;
+        },
+      });
+      const frame = decodeRuntimeHostServiceManagementFrame(output.trim());
+      if (outcome.kind === 'current') {
+        assert.equal(exitCode, 0);
+        assert.equal(
+          frame?.kind === 'result' && frame.action === 'update' ? frame.update.kind : undefined,
+          'already_current',
+        );
+      } else {
+        assert.equal(exitCode, 1);
+        assert.equal(frame?.kind === 'error' ? frame.error.code : undefined, 'update_not_admitted');
+      }
+    }
+  });
+});
+
+function updateResolution(
+  outcome: RuntimeHostUpdateCheckResolution['frame']['updateCheck']['outcome'],
+): RuntimeHostUpdateCheckResolution {
+  const candidate = { version: '2.0.0', integrity: INTEGRITY, compatibility: 7 };
+  return {
+    candidate,
+    frame: {
+      schemaVersion: 1,
+      kind: 'result',
+      action: 'check_update',
+      service: {
+        platform: 'linux',
+        arch: 'x64',
+        osRelease: 'test',
+        state: 'running',
+        pid: 42,
+        lastExitCode: null,
+        installedVersion: outcome.kind === 'current' ? candidate.version : '1.0.0',
+        stateRoot: TARGET.rootPath,
+        projectDirectoryRoots: [],
+      },
+      updateCheck: {
+        selector: OPTIONS.selector,
+        candidate: { version: candidate.version, integrity: candidate.integrity },
+        outcome,
+      },
+    },
+  };
+}
--- a/packages/cli/src/__tests__/runtime-host-selected-update.test.ts
+++ b/packages/cli/src/__tests__/runtime-host-selected-update.test.ts
@@ -102,9 +102,13 @@ describe('managed Runtime Host selected update', () => {
     assert.equal(exitCode, 0);
     assert.equal(updateInput?.sourcePackageRoot, '/verified/package');
     assert.equal(updateInput?.version, '2.0.0');
-    assert.equal(updateInput?.expectedCurrentVersion, '1.0.0');
-    assert.equal(updateInput?.expectedCurrentCliPath, '/managed/versions/1.0.0/dist/cli.js');
-    assert.equal(updateInput?.packageIntegrity, INTEGRITY);
+    assert.deepEqual(updateInput?.registrySelection, {
+      integrity: INTEGRITY,
+      current: {
+        version: '1.0.0',
+        cliPath: '/managed/versions/1.0.0/dist/cli.js',
+      },
+    });
   });
 
   it('lets the exact transaction decide whether a current candidate needs repair', async () => {
--- a/packages/cli/src/__tests__/runtime-host-selected-update.test.ts
+++ b/packages/cli/src/__tests__/runtime-host-selected-update.test.ts
@@ -103,6 +103,7 @@ describe('managed Runtime Host selected update', () => {
     assert.equal(updateInput?.sourcePackageRoot, '/verified/package');
     assert.equal(updateInput?.version, '2.0.0');
     assert.equal(updateInput?.expectedCurrentVersion, '1.0.0');
+    assert.equal(updateInput?.expectedCurrentCliPath, '/managed/versions/1.0.0/dist/cli.js');
     assert.equal(updateInput?.packageIntegrity, INTEGRITY);
   });
 
@@ -155,6 +156,7 @@ function updateSelection(
     selector: OPTIONS.selector,
     candidate,
     outcome,
+    currentCliPath: `/managed/versions/${outcome.kind === 'current' ? candidate.version : '1.0.0'}/dist/cli.js`,
     service: {
       platform: 'linux',
       arch: 'x64',
--- a/packages/cli/src/__tests__/runtime-host-service-manager.test.ts
+++ b/packages/cli/src/__tests__/runtime-host-service-manager.test.ts
@@ -1564,6 +1564,25 @@ describe('managed Runtime Host service', () => {
       retirementFailure?.kind === 'error' ? retirementFailure.error.code : undefined,
       'retirement_failed',
     );
+
+    statusReads = 0;
+    observedVersion = '3.0.0';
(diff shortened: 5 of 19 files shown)


me2seeks:
I reviewed #3687 as the reference transaction for #3228 and would like to confirm one boundary in the recovery machine contract.

The registry target is identified by `version + SHA-512 integrity`: the managed package path is derived from the integrity, and the transaction revalidates the selected current and target paths. However, an error frame emitted after retirement or cutover has started currently carries only `retryTargetVersion`. A controller following that hint will resolve the registry metadata again. If the same version resolves to a different integrity at that point, the retry will converge on a different deployment rather than the exact target whose cutover had already started.

Should the post-cutover recovery hint also carry source-appropriate exact evidence—at least `version + integrity` for the registry adapter? An equally small alternative would be to define immutable `version → integrity` resolution on the official registry as an adapter invariant, then return `target_changed` instead of proceeding when a retry observes a different integrity.

This does not require a universal `ArtifactId` or any expansion of the owner or scheduler scope. It only avoids freezing a version-only hint into the long-term `DeploymentIdentity` recovery contract at G4. If this belongs in the subsequent #3228 reconciliation core instead, explicitly deferring it to that gate would also resolve the boundary.

<details>
<summary>简体中文</summary>

我在把 #3687 作为 #3228 的 reference transaction 核对时，有一个 recovery machine contract 的边界想确认。

当前 registry target 实际由 `version + SHA-512 integrity` 标识：managed package path 由 integrity 派生，transaction 也会重验 selected current/target path。但 retirement 或 cutover 已开始后的 error frame 目前只返回 `retryTargetVersion`。外部 controller 按这个字段重试时会重新解析 registry metadata；如果同一个 version 此时对应的 integrity 发生变化，retry 会收敛到另一份 deployment，而不是原 transaction 已经开始切换的 exact target。

是否应该让 post-cutover recovery hint 也携带 source-appropriate exact evidence（registry 场景至少是 `version + integrity`）？另一种同样小的做法，是把“官方 registry 中的 `version → integrity` 映射不可变”定义为 adapter invariant；retry 时若观察到不同 integrity，就返回 `target_changed`，而不是继续执行。

这不需要引入 universal `ArtifactId`，也不必扩大 owner 或 scheduler 范围；只是避免后续 G4 把 version-only hint 冻结成长期的 `DeploymentIdentity` recovery contract。如果你计划在 #3228 后续 reconciliation core 中处理，明确把它留在那个 gate 也可以闭合这条边界。

</details>

M4n5ter:
@me2seeks Thank you — the underlying observation is correct: a version-only value cannot identify the exact registry package selected by the transaction.

I considered extending the hint with integrity, but the cleaner boundary is to remove the retry hint entirely. Nothing consumes `retryTargetVersion` today, and making the error frame carry another copy of the desired deployment would create a second recovery authority that can become stale. A subsequent update or reconciliation should observe the actual managed installation, resolve the current desired deployment, and let the exact transaction fence that deployment identity. If policy has changed meanwhile, convergence should follow the new desired state rather than be forced by an old error frame.

Commit f576d4fca therefore removes `retryTargetVersion` from the machine contract and changes `update_incomplete` to instruct the caller to run the update again. I will apply the same contraction to the stacked #3720 reconciliation change.

<details>
<summary>简体中文</summary>

谢谢，这个观察的核心成立：只有版本号无法标识事务选中的精确 registry package。

我考虑过给提示补充 integrity，但更干净的边界是彻底删除这项重试提示。当前没有任何使用方消费 `retryTargetVersion`；让错误 frame 再保存一份 desired deployment，会制造第二个可能过期的恢复 authority。下一次 update 或 reconciliation 应重新观察实际 managed installation、解析当前 desired deployment，再由精确事务完成 identity fence。如果期间策略已经变化，也应该收敛到新的期望状态，而不是受旧错误 frame 强制。

因此 commit f576d4fca 从机器契约中删除了 `retryTargetVersion`，并把 `update_incomplete` 改为提示重新运行更新。我也会把同样的契约收缩同步到 stacked 的 #3720。

</details>

Astro-Han (commented):
The acquisition path is carefully built. `npm pack` and `npm install` both run with `--ignore-scripts`, the install happens offline from the already-verified archive against an unreachable registry and an empty cache so nothing can be refetched, the temporary root is created with mode `0o700`, and the candidate is rejected before any download unless its version, integrity and compatibility all parse. Those are the parts that usually go wrong, and they are right here.

Three things to fix.

**[P2] Repairing an existing deployment drops the integrity evidence that acquiring it established**

When a target derived from a verified archive already exists, the healthy and repair paths open it through `openRuntimeHostManagedPackageDeployment()`, and `validatePackage()` in `packages/cli/src/runtime-host-managed-deployment.ts:286` is the whole check: it parses `package.json`, compares `name` and `version`, and calls `stat` on `dist/cli.js` and on `node_modules/@maka/runtime-host/package.json` to confirm they are files. Nothing compares content against the SHA-512 that admitted the package.

So a `dist/cli.js` modified after installation is accepted as current and executed. That is the one place where the integrity check earned something durable and the code declines to use it. The verified archive or a recorded tree hash should stay the authority for an existing deployment, not the presence of two paths and a matching version string.

**[P2] The size budget covers only the compressed archive**

`ARCHIVE_MAX_BYTES` bounds the `.tgz` at 256 MiB, and `MANIFEST_MAX_BYTES` bounds the metadata. There is no bound on the unpacked total, the file count, any single file, directory depth, or the number of links — and `npm install` extracts before anything walks the entries, so the first code that could object runs after the disk is already written. A payload whose logical size is orders of magnitude beyond the compressed cap reaches the acquisition callback.

Please bound the unpacked side explicitly and enforce it during extraction rather than after.

**[P2] A failed backend replacement still reports the target version as installed**

The config target is committed first. If the backend replacement then fails, ordinary `status` reads the config and reports the target version and path as installed without checking the backend deployment, while the backend is still running the old CLI and deployment verification returns `target_mismatch`. The user is told they are on the new version when they are not, which is worse than a plain failure because it removes the reason to retry.

Please have `status` report the backend deployment, or make the config commit conditional on the replacement succeeding.

**One note on what the integrity check proves**

`candidate.integrity` comes from `dist.integrity` in the registry metadata (`runtime-host-update-discovery.ts:266`), and the tarball comes from the same registry. Verifying one against the other is a real defence against a corrupted or truncated download and against a CDN serving something other than what the metadata describes. It is not evidence of publisher identity: metadata and artifact share a trust root, so a registry compromise satisfies the check. That is the ordinary npm trust model and I am not asking you to change it, but the wording around "trusted" updates reads stronger than what is actually established, and it is worth being precise in the description so nobody later builds on a guarantee that is not there.

<details>
<summary>简体中文</summary>

获取路径本身构建得很细致：`npm pack` 与 `npm install` 都带 `--ignore-scripts`；安装是从**已校验过的**归档离线进行，指向一个不可达的 registry 并使用空缓存，因此不可能二次取网；临时根目录以 `0o700` 创建；候选在任何下载之前就会因版本、完整性、兼容性任一解析失败而被拒。这些通常最容易出问题的地方，这里是对的。

三处需要修。

**[P2] 修复既有部署时，丢掉了获取阶段建立起来的完整性证据**

当一个由已校验归档派生出的目标已经存在时，healthy 与 repair 路径通过 `openRuntimeHostManagedPackageDeployment()` 打开它，而 `packages/cli/src/runtime-host-managed-deployment.ts:286` 的 `validatePackage()` 就是全部检查：解析 `package.json`、比对 `name` 与 `version`、再对 `dist/cli.js` 和 `node_modules/@maka/runtime-host/package.json` 做 `stat` 确认它们是文件。**没有任何一处把内容与当初准入该包的那个 SHA-512 做比对。**

于是安装之后被改动过的 `dist/cli.js` 会被当作当前版本接受并执行。这恰恰是完整性校验唯一换来了持久价值的地方，而代码在此选择不使用它。对既有部署而言，权威应当仍是那份已校验的归档或记录下来的树哈希，而不是"两个路径存在 + 版本号相符"。

**[P2] 体积预算只覆盖了压缩后的归档**

`ARCHIVE_MAX_BYTES` 把 `.tgz` 限制在 256 MiB，`MANIFEST_MAX_BYTES` 限制元数据。但**解压后的总量、文件数、单个文件大小、目录深度、link 数量都没有任何上限**——而且 `npm install` 会先解压，之后才轮到遍历条目的代码，也就是说第一处可能提出异议的代码运行时，磁盘已经写完了。一个逻辑体积远超压缩上限若干数量级的载荷可以一路到达获取回调。

请显式限制解压侧，并且**在解压过程中**执行，而不是解压之后。

**[P2] backend 替换失败后，仍然把目标版本报告成已安装**

config 目标是先提交的。若随后 backend 替换失败，普通的 `status` 会读取 config，把目标版本与路径报成已安装，而**不去核对 backend 部署**；此时 backend 仍在运行旧 CLI，部署校验返回 `target_mismatch`。用户被告知自己已在新版本上，而事实并非如此——这比直接失败更糟，因为它消解了用户重试的理由。

请让 `status` 报告 backend 部署的实际状态，或者让 config 的提交以替换成功为前提。

**关于完整性校验究竟证明了什么**

`candidate.integrity` 取自 registry 元数据中的 `dist.integrity`（`runtime-host-update-discovery.ts:266`），而 tarball 来自同一个 registry。用其中之一校验另一个，确实能有效防住下载损坏或截断，也能防住 CDN 返回与元数据描述不符的内容。**但它不是发布者身份的证据**：元数据与产物共享同一个信任根，registry 一旦被攻陷，这项校验照样通过。这就是 npm 通常的信任模型，我不要求你改变它；只是"trusted updates"这个说法读起来比实际建立的保证更强，建议在正文中把话说准，以免以后有人在一个并不存在的保证之上继续搭建。

</details>

Astro-Han:
The branch moved to `f576d4fca045e0282138f0a464f9a9a70f40a058` after my review, which was bound to `a242e40fd8`. I re-checked all three blocking findings against the new head rather than assuming either way, and **all three are unchanged**.

- `validatePackage()` in `packages/cli/src/runtime-host-managed-deployment.ts:286` is byte-identical on the new head: still `package.json` parse, name/version compare, and two `stat` calls. No content comparison against the admitting SHA-512.
- `packages/cli/src/runtime-host-update-package.ts` still bounds only `ARCHIVE_MAX_BYTES` (256 MiB compressed) and `MANIFEST_MAX_BYTES`. There is still no bound on the unpacked total, file count, single-file size, directory depth, or link count, and extraction still happens before anything walks the entries.
- The config target is still committed before the backend replacement, so a failed replacement still leaves ordinary `status` reporting the target version as installed.

The new commit changes the `update_incomplete` message from naming an exact version to "run the update again to reconcile the managed installation", and drops `retryTargetVersion` from the error frame. That is a good change on its own terms — a version string was never enough to identify the exact package to recover to, so reconciling from the observed installation is the sounder contract. It does not touch any of the three findings above.

Nothing here asks you to re-explain the change. I mention it only so the moved head is not mistaken for a response to the findings.

<details>
<summary>简体中文</summary>

分支在我 review 之后动到了 `f576d4fca045e0282138f0a464f9a9a70f40a058`，而那次 review 绑定的是 `a242e40fd8`。我没有假设任何一边，而是把三条阻塞项都对着新 head 重新核了一遍，**三条都没有变化**。

- `packages/cli/src/runtime-host-managed-deployment.ts:286` 的 `validatePackage()` 在新 head 上逐字节相同：仍然只是解析 `package.json`、比对 name/version、两次 `stat`。仍然没有任何一处把内容与准入时的 SHA-512 做比对。
- `packages/cli/src/runtime-host-update-package.ts` 仍然只限制 `ARCHIVE_MAX_BYTES`（压缩后 256 MiB）与 `MANIFEST_MAX_BYTES`。解压后总量、文件数、单文件大小、目录深度、link 数量仍然没有上限，且解压仍发生在任何遍历条目的代码之前。
- config 目标仍然先于 backend 替换提交，因此替换失败后，普通 `status` 依旧会把目标版本报成已安装。

新增的这条 commit 把 `update_incomplete` 的提示从"重试确切的某个版本"改成"再跑一次 update，以对齐受管安装"，并从错误帧中移除了 `retryTargetVersion`。就其本身而言这是个好改动——一个版本号本来就不足以标识要恢复到的确切包，改为依据观测到的安装状态去对齐，契约更稳。但它没有触及上面三条。

这里不需要你再解释一遍这次改动，我写出来只是为了避免把"head 变了"误读成对上述发现的回应。
</details>

Astro-Han (commented):
Two more blocking findings on `f576d4fca045e0282138f0a464f9a9a70f40a058`, both about the update transaction rather than the acquisition path.

**[P2] The package is downloaded before the transaction is allowed to decide it does not need one**

In `packages/cli/src/runtime-host-update-command.ts:505`, only a `manual_action` outcome short-circuits. Everything else — including `outcome.kind === 'current'` — goes through `deps.withPackage(...)` at `:521`, which packs or downloads before `deps.update` is ever called. But the transaction's own first decision, `already_current` at `:239`, is reached without reading `sourcePackageRoot` at all.

The consequence is deterministic, not a race: the installed deployment is already current and healthy, discovery succeeds, then `npm pack` or the download fails, or the registry is briefly unreachable. The user is told the update failed to acquire a package, when the correct answer was that they were already on the target and nothing needed to be fetched. The same eager acquisition is what stops the repair path from opening an already-installed current deployment while the registry is unavailable — which is exactly when repair matters most.

Please defer acquisition until the transaction has established that it must prepare a *different* deployment. The current-healthy and repair paths should be able to run off the installed deployment alone.

**[P2] The forced-stop repair is unreachable in the case it exists for**

When `verifyReady` fails, `:222` sets `activeTargetNeedsRepair`. A few lines later, `:246` unconditionally awaits `runOperator(currentOperatorPath, ['status', '--framed', ...])` to decide the lifetime-lock capability. If that operator is missing, corrupt, fails to spawn, or times out, the rejection propagates straight to the outer handler.

The only `--allow-interrupt-active-tasks` forced-stop recovery is later, at `:307`, and it triggers on exactly one condition: a *returned* frame with `error.code === 'retirement_failed'`. An operator that cannot be executed never produces that frame.

So `update --allow-interrupt-active-tasks` cannot repair a Host whose operator is itself broken — which is a large share of the stopped/unhealthy states this recovery is meant to cover, and arguably the most likely reason a deployment is unhealthy in the first place. Please catch the probe failure in the repair case and fall through to the conservative forced-stop path instead of aborting ahead of it.

**One process note (not blocking)**

Commit `a242e40fd8` changes seven code and test files (+239/−157) with no `Generated-by:` trailer, while the neighbouring commits including `f576d4fc` carry `Generated-by: Codex`. CONTRIBUTING asks for the trailer on each affected commit and for it to survive into the final commit. Since this repository squash-merges and the other commits carry it, the squashed result will most likely be fine — I mention it only so the trailer is checked on the final message rather than assumed.

Together with the three findings above, that is five open P2s. I am not approving while they are open, and the exact-head `test` run is still in progress.

<details>
<summary>简体中文</summary>

在 `f576d4fca045e0282138f0a464f9a9a70f40a058` 上再提两条阻塞项，都关于更新事务本身，而非获取路径。

**[P2] 在事务被允许判断「不需要包」之前，包已经被下载了**

`packages/cli/src/runtime-host-update-command.ts:505` 只对 `manual_action` 做了短路，其余分支——包括 `outcome.kind === 'current'`——都会走 `:521` 的 `deps.withPackage(...)`，也就是在 `deps.update` 被调用之前就已经 pack 或下载完毕。但事务自己的第一个判断 `:239` 的 `already_current`，根本没有读取 `sourcePackageRoot`。

后果是确定性的，不是竞态：已安装的部署本就是当前版本且健康，discovery 成功，随后 `npm pack` 或下载失败、或 registry 短暂不可达。用户被告知「更新未能获取包」，而正确的答复本应是「你已经在目标版本上，什么都不需要取」。同样是这处提前获取，导致 registry 不可用时 repair 路径无法打开一个已经装好的当前部署——而那恰恰是 repair 最有价值的时刻。

请把获取推迟到事务确认「必须准备一个不同的部署」之后。current-healthy 与 repair 路径应当仅凭已安装的部署就能跑完。

**[P2] 强制停止的修复路径，在它本该生效的场景下不可达**

`verifyReady` 失败时，`:222` 置上 `activeTargetNeedsRepair`。紧接着 `:246` 无条件 await `runOperator(currentOperatorPath, ['status', '--framed', ...])` 来判定 lifetime-lock 能力。如果这个 operator 缺失、损坏、拉起失败或超时，异常会直接抛到外层处理器。

而唯一的 `--allow-interrupt-active-tasks` 强制停止恢复在更后面的 `:307`，触发条件只有一个：**返回**一个 `error.code === 'retirement_failed'` 的帧。一个根本无法执行的 operator，永远产生不了这个帧。

于是 `update --allow-interrupt-active-tasks` 无法修复「operator 自身已坏」的 Host——而这在该恢复路径本应覆盖的 stopped/unhealthy 状态中占很大比例，甚至可以说是部署变得不健康的最常见原因。请在 repair 情形下捕获这次探测失败，落到保守的强制停止路径，而不是在它之前就中止。

**一条流程性说明（不阻塞）**

commit `a242e40fd8` 改动七个代码与测试文件（+239/−157），没有 `Generated-by:` trailer；而相邻的 commit（包括 `f576d4fc`）都带 `Generated-by: Codex`。CONTRIBUTING 要求每个相关 commit 都带该 trailer，并保证它保留到最终 commit。鉴于本仓库是 squash 合并、且其他 commit 带了该 trailer，最终结果大概率没问题——我写出来只是希望最终 commit message 上真的核一眼，而不是默认没事。

连同前面三条，目前共有五条未决 P2。这些未决期间我不会 approve；exact head 的 `test` 也仍在运行中。
</details>

M4n5ter:
@Astro-Han

<details open>
<summary>English</summary>

Thank you for the detailed review. I reproduced the replacement-state problem and fixed it in d08da5ab4. The service backends now distinguish two boundaries:

- If backend replacement does not commit, systemd/LaunchAgent restores the previous service definition and the manager restores the previous config. Status therefore continues to report the previous deployment.
- If backend replacement commits but readiness is unknown, the selected deployment remains configured and is stopped. Rolling back an older binary after the new Runtime may have touched the State Root would cross an unknown storage-compatibility boundary.

I did not add persistent tree attestation. The registry SHA-512 identifies the npm archive, not the extracted filesystem tree, and the managed package, any locally stored tree hash, the operator, and the updater all share the same per-user write boundary. A same-principal attacker could modify the verifier as readily as the package, while atomic staging and rename already prevent an incomplete copy from becoming current. I instead tightened the wording from different-content to different registry-integrity candidates.

I also did not add a second tar parser or extraction engine for an unpacked quota. This command accepts only the official maka-agent package from the npm registry; metadata and artifact intentionally share that trust root, and a compromised publisher could already deliver arbitrary same-user code. Accidental release growth is bounded by the existing release gate at 18 MiB compressed, 90 MiB unpacked, and 9,000 entries. A portable consumer-side streaming quota would duplicate npm extraction without closing a distinct current boundary. We should revisit that if the updater later accepts arbitrary registries or packages.

Finally, I renamed the PR from trusted to verified managed updates so the claim matches the actual archive-integrity and manifest checks rather than implying publisher-identity verification.

</details>

<details>
<summary>简体中文</summary>

感谢这次细致审查。我复现了 replacement 状态问题，并在 d08da5ab4 中完成修复。Service backend 现在明确区分两个边界：

- backend replacement 尚未提交时，systemd/LaunchAgent 恢复原 service definition，manager 恢复原 config，因此 status 继续报告旧 deployment。
- backend replacement 已提交但 readiness 未知时，保留并停止 selected deployment。新 Runtime 可能已经接触 State Root，此时回滚旧二进制会跨越未知的 storage compatibility 边界。

我没有增加持久 tree attestation。Registry SHA-512 标识的是 npm archive，不是解压后的 filesystem tree；managed package、本地 tree hash、operator 与 updater 都处于同一用户可写边界。同一 principal 的攻击者既能修改 package，也能修改 verifier；而现有原子 staging 与 rename 已经防止不完整复制成为 current。我改为收紧表述，只承诺区分 registry integrity 不同的候选。

我也没有为了 unpacked quota 引入第二套 tar parser 或 extraction engine。当前命令只接受 npm registry 上官方的 maka-agent package；metadata 与 artifact 本来就共享该信任根，publisher 被攻陷时已经可以投递任意同用户权限代码。意外的 release 体积增长则由现有发布门禁限制为 18 MiB compressed、90 MiB unpacked、9,000 entries。跨平台 consumer-side streaming quota 会重复 npm extraction，却不能闭合一个独立的当前边界。如果未来 updater 接受任意 registry 或 package，应重新评估该约束。

最后，我已将 PR 标题从 trusted managed updates 改为 verified managed updates，使表述准确对应 archive integrity 与 manifest 校验，不暗示 publisher identity verification。

</details>

Astro-Han (commented):
Re-reviewed on `d08da5ab4fc24dd987f3b5b61909fe5429674bd5`.

**The `status`-reports-installed finding is closed.** `replaceRuntimeHostManagedServiceLocked` now separates a replacement that never committed from a target that committed but never became ready. On any non-`update_incomplete` failure from `backend.replace`, the previous config JSON is written back before the error is raised, so `status` no longer reads a target version the backend is not running; when the config restore itself fails, the error is raised as an `AggregateError` naming both failures rather than a single misleading one. The distinguishing signal is the in-memory exception type rather than anything read back from disk, so there is no loop of judging contamination by a source the contamination may have rewritten. The tests pin the actual file contents in both outcomes, not just the thrown code.

I also checked the two platform backends against each other: `capture → apply → restore` is symmetric between systemd and LaunchAgent, and both terminate in the stopped state that was captured. Recovery after the readiness failure is real — the target config, package and operator are all retained, so `service start` or another `update` can proceed; the deliberate refusal to auto-restart the old service across an unknown storage boundary is a restriction, not a dead end.

**The other four findings are unchanged on this head.** `validatePackage()` still compares only name, version and the presence of two paths, with no content comparison against the admitting SHA-512. `runtime-host-update-package.ts` still bounds only the compressed archive and the manifest. `runtime-host-update-command.ts` still routes the `current` outcome through `withPackage`, and still awaits the operator status probe unconditionally after flagging repair, leaving the forced-stop recovery reachable only via a returned `retirement_failed` frame.

**[P2] The readiness-failure message asserts a stopped service that may still be running**

At `packages/cli/src/runtime-host-service-manager.ts:579`, a `waitForReady` failure is handled as:

```ts
await backend.stop().catch(() => undefined);
throw new RuntimeHostServiceManagerError(
  'update_incomplete',
  'The replacement Runtime Host did not become ready; the selected deployment was retained but stopped ...',
);
```

The `stop()` failure is swallowed, and the message that reaches the user states as fact that the deployment is stopped. If the stop did not take, the operator is told a service is down while it is still running the binary that just failed its own readiness check — and the message removes their reason to check. This is the same shape as the finding closed above: reporting a state that was attempted rather than a state that was confirmed.

Please fold a stop failure into the raised error, the way the config-restore failure already is, so the message distinguishes "stopped" from "could not be stopped".

For clarity on where this does *not* apply: the neighbouring `:574` path makes the same claim but issues no `stop()` at all — its claim rests on the backend having restored the captured stopped state, which the platform backends do.

**Four blocking findings remain open** — the two above from the earlier review plus the two from the follow-up, and this one. The exact-head `test` run is still queued, so there is no green gate either. Not approving.

<details>
<summary>简体中文</summary>

已在 `d08da5ab4fc24dd987f3b5b61909fe5429674bd5` 上重审。

**「status 把目标版本报成已安装」这条已闭合。** `replaceRuntimeHostManagedServiceLocked` 现在把「替换从未提交」与「提交了但从未 ready」区分开：`backend.replace` 抛出任何非 `update_incomplete` 的异常时，会先把旧的 config JSON 写回再抛错，因此 `status` 不会再读到一个 backend 并未运行的目标版本；若 config 恢复本身失败，则以点名两个失败的 `AggregateError` 抛出，而不是给出一条会误导的错误。用来区分的信号是**内存中的异常类型**，而不是从磁盘读回的任何东西，因此不存在"用一个可能已被污染改写的来源去判断污染"的闭环。测试锁住的是两种结局下**文件的真实内容**，而不只是抛出的错误码。

我也把两个平台后端互相对照过：systemd 与 LaunchAgent 的 `capture → apply → restore` 结构对称，且都终止在捕获时的停止态。readiness 失败后的恢复是真实可用的——target 的 config、包与 operator 全部保留，`service start` 或再跑一次 `update` 都能继续；至于"跨未知存储边界不自动拉回旧服务"是有意的限制，不是死路。

**其余四条在这个 head 上未变。** `validatePackage()` 仍然只比对 name、version 与两个路径是否存在，没有与准入时的 SHA-512 做内容比对。`runtime-host-update-package.ts` 仍然只限制压缩包与清单。`runtime-host-update-command.ts` 仍然让 `current` 结果走 `withPackage`，也仍然在标记 repair 之后无条件 await operator 状态探测，使强制停止恢复只能经由**返回**的 `retirement_failed` 帧到达。

**[P2] readiness 失败的错误文案断言了一个可能仍在运行的服务已停止**

`packages/cli/src/runtime-host-service-manager.ts:579` 对 `waitForReady` 失败的处理是：

```ts
await backend.stop().catch(() => undefined);
throw new RuntimeHostServiceManagerError(
  'update_incomplete',
  'The replacement Runtime Host did not become ready; the selected deployment was retained but stopped ...',
);
```

`stop()` 的失败被吞掉，而送到用户面前的文案把"已停止"当作事实陈述。如果这次 stop 没有生效，运维看到的是"服务已停"，实际上它仍在运行那个刚刚没能通过自身 readiness 检查的二进制——而这句话恰好**消解了他去核实的理由**。这与上面那条刚闭合的发现是同一种形状：报告的是**尝试过的状态**，而不是**确认过的状态**。

请把 stop 的失败并入抛出的错误，就像 config 恢复失败已经做的那样，让文案能区分"已停止"与"停不下来"。

为免误伤，说明一下**不适用**的地方：相邻的 `:574` 那条路径做了同样的陈述，但它**完全没有调用 `stop()`**——它的依据是 backend 已把服务恢复到捕获时的停止态，而两个平台后端确实做到了这一点。

**目前仍有四条阻塞项未决**——前一轮的两条、后续一轮的两条，加上本条。exact head 的 `test` 也仍在排队，没有绿灯。不 approve。
</details>

Astro-Han:
Correcting the count in my review just above: **five** blocking findings are open, not four — the two from `5009176191` that remain unchanged, the two from `5009373565`, and the new one on the readiness-failure message. The enumeration was right; the number was wrong.

<details>
<summary>简体中文</summary>

更正上一条 review 里的数字：未决的阻塞项是**五条**，不是四条——`5009176191` 中仍未变的两条、`5009373565` 的两条，加上关于 readiness 失败文案的新那条。枚举是对的，数字写错了。
</details>

M4n5ter:
@Astro-Han

<details open>
<summary>English</summary>

Thanks for the careful follow-up. I re-adjudicated the findings and updated the PR at `50a0b5bb2`.

Accepted and fixed:

- `current` and same-deployment repair no longer acquire a package; they operate from the installed deployment, so registry availability cannot turn a no-op or local repair into an acquisition failure
- when the current operator cannot answer the retirement probe during repair, the transaction now fails closed without interruption authority, while `--allow-interrupt-active-tasks` reaches the explicit backend stop-and-repair path
- if a replacement fails readiness and the subsequent stop also fails, the stop error is preserved in an `AggregateError` and the message says the service could not be stopped instead of claiming it is stopped
- the rewritten commit history now carries `Generated-by: Codex` on every affected commit

I did not add the proposed persistent tree hash or a second extraction implementation. The registry SHA-512 authenticates the acquired archive within the npm trust model; it does not define a canonical post-install tree, and a digest recorded beside a same-user-writable deployment would be mutable under the same trust boundary. Making that an actual tamper boundary would require a signed tree manifest or an OS-protected deployment authority, not a local hash comparison in this transaction. Likewise, the official release gate already bounds the published archive, unpacked bytes, and entry count. Reimplementing npm extraction solely to enforce a second consumer-side budget would add a second package installer without closing a distinct trust boundary in this PR.

Verified on the rewritten head: all 443 CLI tests, repository typecheck, lint, format check, and ASF header audit pass locally.

</details>

<details>
<summary>简体中文</summary>

感谢细致的复审。我重新裁决了这些发现，并已在 `50a0b5bb2` 更新本 PR。

已采纳并修复：

- `current` 与同 deployment repair 不再获取 package，而是直接使用已安装 deployment，因此 registry 不可用不会再把 no-op 或本地修复变成获取失败
- repair 期间若当前 operator 无法完成 retirement probe，未授予中断权限时事务会 fail closed；显式提供 `--allow-interrupt-active-tasks` 时则能进入 backend stop-and-repair 路径
- replacement readiness 失败后若 stop 也失败，会通过 `AggregateError` 保留 stop 错误，并明确告知服务无法停止，不再声称它已停止
- 重写后的每个相关 commit 均已带上 `Generated-by: Codex`

我没有加入持久 tree hash，也没有实现第二套解包器。Registry SHA-512 在 npm 信任模型内验证的是获取到的 archive，并没有定义 canonical post-install tree；与 deployment 一样可由同一用户写入的本地 digest，也处在同一个可变信任边界内。若要形成真正的防篡改边界，需要签名 tree manifest 或受 OS 保护的 deployment authority，而不是在当前事务旁增加本地 hash 比对。同样，官方发布门禁已经限制发布产物的压缩体积、解包体积与条目数；仅为重复一层 consumer-side budget 而重写 npm extraction，会增加第二套 package installer，却没有在本 PR 的信任边界内关闭新的风险。

重写后的 head 已在本地通过全部 443 项 CLI 测试、仓库 typecheck、lint、format check 与 ASF header audit。

</details>

Astro-Han (commented):
Update after freshness on `50a0b5bb23`:

Previously published 5×P2: 2 remain, 3 are now closed.

- Still open: (1) `managed-deployment` reuses existing package without SHA-512 content check; (2) `update-package` unpacks without decompressed size/count/depth/link limits.
- Closed in this head: (3) unsubmitted replace now restores previous config (previously reported incomplete); (4) `current` no longer goes through `withPackage`; (5) `verifyReady` failure handling now correctly absorbs operator status when active target is already not ready.

No new blocking issues beyond the 2 above. Checks on `50a0b5bb23` are `test: success`.

<details><summary>简体中文</summary>原 5 项中 2 项仍成立，3 项已闭合。</details>

Astro-Han (commented):
Follow-up on the author's explanation: these findings do not require a local tree hash to act as an adversarial tamper boundary, nor do they require a second npm installer. The first finding remains because `validatePackage()` checks path/type/name/version but does not revalidate that the bytes at an existing reused/repair target still match the admitting registry integrity. A same-principal attacker being able to modify a local verifier does not remove the operational consistency invariant; to close this point, either establish and test an immutable source/path invariant for every reuse path, or validate content at that boundary. The second finding is about the consumer contract: producer-side release limits are not an executable runtime guarantee for every registry/cache input this code can consume. The current path bounds the compressed archive and manifest, then hands the package to `npm install` without enforcing decompressed size, entry-count, depth, or link budgets. Either prove the source is closed to artifacts with those bounds, or add a pre-install archive preflight; a second package installer is not required.

<details><summary>简体中文</summary>上述解释不改变两条 P2 的根因，无需引入 tree hash 或重写 installer 即可闭合。</details>

M4n5ter:
@Astro-Han

<details open>
<summary>English</summary>

Thank you for reframing the two remaining findings. I re-adjudicated them against the actual transaction and want to make the intended boundary explicit.

### Existing deployment integrity

Your operational-consistency distinction is valid. My previous reply focused on adversarial tampering within the same user boundary, which did not directly answer the non-adversarial case where an installed package tree is damaged after admission.

The contract implemented here is narrower:

- the registry SHA-512 binds the selected registry metadata to the acquired archive;
- that archive is extracted into a unique staging directory, validated, and atomically renamed into a directory derived from its registry integrity;
- after publication, Maka treats that package directory as immutable under its own deployment authority: update paths either open it, publish another directory, or remove an inactive directory as a whole; they do not modify files inside a published package;
- reuse therefore relies on this Maka-owned immutability invariant. It is not local tree attestation and does not claim recovery from arbitrary external modification or disk corruption.

This also means the current wording around repairing a stopped or unhealthy Host is too broad unless the boundary is stated. The repair path restores service/operator lifecycle state over an intact managed package. If the package bytes themselves have been modified outside the deployment authority, this transaction does not reconstruct them. Same-version package re-materialization could be added as a separate recovery capability, but it requires a real source-of-truth design; placing a mutable tree hash beside the deployment would not create that authority.

I will tighten the PR/issue wording so it does not imply package-corruption recovery. I do not think persistent local tree attestation belongs in this transaction.

### Unpacked archive budget

I agree with the literal observation that a producer-side release gate is not a cryptographic consumer-side guarantee. The merge decision, however, depends on the accepted source and threat model.

This command accepts only the official `maka-agent` package from the fixed npm registry. Registry metadata and archive share that trust root, installation is offline from the already verified archive, and the ASF release path bounds compressed bytes, unpacked bytes, and entry count before publication. A publisher or registry compromise can already deliver arbitrary code that the managed service will later execute with the same user authority. A consumer-side tar preflight would provide defence in depth against one availability manifestation of that compromise, but it would not close an independent trust boundary in this PR. It would also add another archive parser and a second set of package-shape limits beside npm extraction.

That trade-off changes if Maka later accepts custom registries, mirrors, uploaded packages without the release gate, or another artifact trust root. At that point, consumer-enforced unpacked size, entry, depth, and link budgets should be a prerequisite. Under the current closed official-package source, I consider it useful future hardening rather than a P2 merge blocker.

So the resulting adjudication is:

1. accept the contract-clarity issue around package corruption and narrow the repair claim;
2. retain atomic immutable-by-authority deployment reuse without adding local tree attestation;
3. defer consumer-side extraction quotas until the accepted package-source boundary expands or we have evidence that the official release gate is insufficient.

If you believe #3687 must include recovery from externally modified package trees, or must treat the official registry artifact as untrusted before execution, that is a different product requirement from the one currently stated. Calling out that requirement explicitly would help us decide it directly rather than encoding it indirectly through implementation findings.

</details>

<details>
<summary>简体中文</summary>

感谢你重新界定剩余两条问题。我结合实际 transaction 重新裁决了一次，下面明确说明本 PR 的设计边界。

### 既有 deployment 的完整性

你对“运行一致性”的区分是成立的。我之前的回复主要讨论同一用户边界内的对抗性篡改，没有直接回答 package 通过准入后发生非对抗性损坏的情况。

当前实现的契约更窄：

- registry SHA-512 将所选 registry metadata 与实际取得的 archive 绑定；
- archive 解包到唯一 staging directory，经过校验后原子 rename 到由 registry integrity 派生的目录；
- 发布后，Maka 的 deployment authority 将该 package directory 视为不可变对象：更新路径只会打开它、发布另一个目录，或整体删除 inactive directory，不会修改已发布 package 内的文件；
- 因此 reuse 依赖的是 Maka 自身维护的不可变约束，而不是本地 tree attestation，也不承诺从任意外部修改或磁盘损坏中恢复。

这也意味着当前“修复 stopped 或 unhealthy Host”的措辞如果不说明边界，确实过宽。Repair 的含义是：在 managed package 仍然完整的前提下，恢复 service/operator lifecycle 状态。如果 package bytes 已在 deployment authority 之外被修改，本事务不会重新构造它。未来可以单独增加同版本 package re-materialization，但那需要真正的 source-of-truth 设计；把一个同样可写的 tree hash 放在 deployment 旁边并不能形成这样的 authority。

我会收紧 PR/issue 的表述，避免暗示它能修复 package corruption。持久化的本地 tree attestation 不应进入当前 transaction。

### Archive 解包预算

我同意字面事实：producer-side release gate 并不是 consumer-side 的密码学保证。但它是否阻塞合并，取决于当前接受的 source 与 threat model。

这个命令只接受固定 npm registry 上的官方 `maka-agent` package。Registry metadata 与 archive 共享同一信任根；安装只使用已经验证的 archive 离线完成；ASF 发布流程会在发布前限制 compressed bytes、unpacked bytes 和 entry count。Publisher 或 registry 一旦被攻陷，本来就已经能够投递随后会以同一用户权限执行的任意代码。Consumer-side tar preflight 可以针对这种 compromise 的某一种 availability 表现提供纵深防御，但它没有在本 PR 中闭合一个独立的信任边界；同时还会在 npm extraction 之外增加另一个 archive parser 与第二套 package-shape limits。

如果未来 Maka 接受自定义 registry、mirror、未经 release gate 的上传 package，或另一种 artifact trust root，这个权衡就会变化。届时 consumer 强制执行 unpacked size、entry、depth 与 link budget 应当成为前置条件。在当前封闭的官方 package source 下，我认为它是有价值的未来 hardening，而不是 P2 merge blocker。

因此最终裁决是：

1. 接受 package corruption 边界表述不够清楚的问题，并收窄 repair 的承诺；
2. 保留基于 deployment authority 不可变约束的原子复用，不增加本地 tree attestation；
3. 在可接受的 package source 扩展，或出现官方 release gate 不足的证据前，暂不增加 consumer-side extraction quota。

如果你的要求是 #3687 必须恢复被外部修改的 package tree，或者必须在执行前把官方 registry artifact 视为不可信输入，那属于不同于当前 issue 的产品需求。希望能把这项要求直接明确出来，以便我们讨论需求本身，而不是通过实现 finding 间接引入。

</details>

Astro-Han (commented):
Follow-up: after re-reviewing with the author's invariant, I consider the two remaining findings closed as blocking for this PR.

- Content consistency: all operations on published packages are whole-package atomic copies/renames, and the package directory is derived from SHA-512 integrity. The author's stated invariant — published package directories are immutable under Maka's authority, excluding external tampering/disk corruption — closes the reuse-hash gap as a current lifecycle concern. Track as threat-model note.

- Archive budgets: the current producer path is closed (registry.npmjs.org → SHA-512 → offline install) and release artifacts are bounded by `release-cli-artifact-policy` (18 MiB/90 MiB/9000 entries). No custom registry input exists; consumer quotas would be for future trust-root expansion, not a correctness gate now.

No remaining P0-P2 in this head.

<details><summary>简体中文</summary>两条在当前契约下不阻断。</details>

Astro-Han (approved):
APPROVE — 50a0b5bb23, no remaining P0-P2 after re-review, test success.

Issue #1404, opened 2026-07-23
refactor: third-pass sink-file and boundary governance


Astro-Han:
## Context

Second-pass governance (#1084) is winding down: 10 of its slices are merged and the remaining ones are tracked as sub-issues there. This issue tracks the third pass: ~150 source files ≥350 lines (everything not already covered by #1084) audited against the same four criteria, with per-finding adversarial verification against current `main`. 32 findings survived verification; 18 were refuted — plausible-but-justified coupling was explicitly rejected (e.g. `model-adapter.ts` usage normalization, `permission.ts` shell classification, and `app-shell.tsx` mode management are all fine as-is).

The dominant pattern this round is different from #1084: 25 of 32 findings are **silently-diverging parallel implementations** (copy-paste that must be kept in sync by hand), not mixed responsibilities. Several copies have already drifted.

## Headline evidence

- `bots/qq-bridge.ts`, `bots/discord-bridge.ts`, and most of `bots/dingtalk-bridge.ts` copy the same WebSocket gateway lifecycle wholesale (opcodes, heartbeat/seq/session, connect/reconnect/backoff — the backoff constants are byte-identical). `bots/simple-bridge.ts` additionally carries dead Feishu/Discord paths superseded by the dedicated bridges and only actually runs Telegram.
- Five `apps/desktop/src/main/oauth/*-subscription-service.ts` files copy one service shell (pending/logout/refresh/getAccessToken/failure handling); the codex and antigravity services duplicate the loopback callback server nearly verbatim. The missing base leaks downstream into `subscription-ipc-main.ts` (7 same-shape handlers per provider) and `oauth-model-connections-main.ts`.
- Two full 5-field cron parsers exist across packages: `core/plan-reminders.ts` (`parsePlanReminderCronExpression`/`cronExpressionMatches`, ~L541–656) and `runtime/automation-state.ts` (`expandCronField`/`normalizeCronExpression`/`computeNextCronFire`/`matchesCronField`, ~L521–696) — and inside `automation-state.ts` the same field grammar is parsed twice with already-divergent edge handling.
- Security-relevant: `main/rive-cli.ts` ships a private redaction implementation (`redactRiveText`/`redactRiveValue`, ~L160–178) parallel to `@maka/core`'s authoritative `redactSecrets`, with narrower coverage (misses GitHub/Google/Slack token shapes) — a future hardening of the authoritative rules silently won't apply here.

## Slices

[...]

PR #3692, opened 2026-08-24
test(core): unit-test the shared Unicode sanitizer

rekcilyssup:
## Summary

`packages/core/src/text-sanitize.ts` is the single source of truth (#1404) for the Unicode sanitize pipeline shared by session names (`session-name.ts`) and foreign-session names (`foreign-session.ts`) — explicitly documented as security-adjacent (bidi-spoofing defense) shared infrastructure — but it had no direct unit tests; only callers exercised it transitively.

This adds direct coverage of each pipeline stage and edge:

- **NFC normalization** — decomposed macOS filenames compose to match.
- **Control characters** → replaced with single spaces (multi-line input stays readable).
- **Bidi format characters** (ALM/LRM/RLM/LRE…PDI) → replaced with spaces so direction spoofing collapses.
- **Zero-width characters** → removed entirely; includes a compound-emoji case showing code points stay intact.
- **Whitespace collapse + trim**, code-point capping without splitting surrogate pairs, silent caps via empty suffix, and inputs that sanitize to empty.

Character classes are asserted via `\uXXXX` escapes in both source and test, per the note in `text-sanitize.ts` about keeping files text-safe for git.

No production code is touched.

## Verification

- New suite passes (`node --test packages/core/dist/__tests__/text-sanitize.test.js`).
- Full `@maka/core` suite: 660 tests, 660 pass.
- `npx biome check packages/core/src` clean; ASF header check passes.

## AI use

Select exactly one:

Tool(s) and scope: opencode — identified the coverage gap, wrote the tests against the documented pipeline, hand-checked expectations against the implementation. Commit carries the `Generated-By: opencode` trailer.

## Checklist

- [x] No

### DIFF
--- a/packages/core/src/__tests__/text-sanitize.test.ts
+++ b/packages/core/src/__tests__/text-sanitize.test.ts
@@ -1,0 +1,98 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import assert from 'node:assert/strict';
+import { describe, test } from 'node:test';
+import { sanitizeUnicodeText } from '../text-sanitize.js';
+
+describe('sanitizeUnicodeText', () => {
+  test('passes plain text through unchanged', () => {
+    assert.equal(sanitizeUnicodeText('Fix login bug', { maxCodePoints: 80 }), 'Fix login bug');
+    assert.equal(sanitizeUnicodeText('会话名称', { maxCodePoints: 80 }), '会话名称');
+  });
+
+  test('normalizes to NFC so equivalent spellings match', () => {
+    // Decomposed (NFD) e + combining acute, as macOS filenames are recorded.
+    const decomposed = 'cafe\u0301';
+    assert.notEqual(decomposed, 'caf\u00e9');
+    assert.equal(sanitizeUnicodeText(decomposed, { maxCodePoints: 80 }), 'caf\u00e9');
+  });
+
+  test('replaces control characters with single spaces', () => {
+    // Escaped form keeps the source file text-safe; see the note in text-sanitize.ts.
+    assert.equal(
+      sanitizeUnicodeText('foo\u0007bar\u001Fbaz', { maxCodePoints: 80 }),
+      'foo bar baz',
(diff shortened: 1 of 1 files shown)


Astro-Han:
Two questions before this lands, both about scope rather than correctness — the assertions themselves are accurate.

**What drove this?** There's no issue, no reported bug, and no regression this pins. #1404 establishes that `text-sanitize.ts` is the single source of truth, but that's a statement about ownership, not a request for coverage. Tests are code with a maintenance cost, so it's worth saying in the body what this is protecting against — a bug you hit, a refactor you're planning, or a specific property you found unprotected.

**Some of it is already covered, and more so as of an hour ago.** #3846 landed `session-name.test.ts`, which exercises the bidi class through `normalizeUserSessionName` across the full U+206A–206F range, and `foreign-session.test.ts` covers the same class through `isSafeForeignId`. Your bidi test overlaps both.

The parts that genuinely aren't reachable from either caller are narrower and, I think, the real argument for this PR:

- **capping by code points without splitting surrogate pairs** — needs a deliberately constructed astral-plane input at the boundary; no session name naturally produces it
- **NFC normalization** — decomposed vs. composed forms matching requires constructing both spellings
- **the silent cap via `truncatedSuffix: ''`** — only `session-name.ts:91` uses it, and only for its own length

Those three are worth having as a direct contract. The plain-passthrough, control-character, zero-width, and whitespace-collapse cases mostly restate what the two callers' suites already assert from the other side.

Would you narrow this to the properties the callers can't reach, and say in the body what prompted it? If you think the full-pipeline coverage is worth it as a whole, that's a fine answer too — I'd just like the reasoning on the record rather than inferred.

*AI-assisted review: I checked both callers' invocations and their existing suites against head `69c03b670` to work out which properties are only reachable directly. No tests run. AI review is not independent human review.*

Astro-Han (approved):
Thanks for adding focused coverage at the shared Unicode-sanitization authority. I reviewed exact head `69c03b6708e7c240130e7eefff4d96f72d3f090f`.

This has a concrete contract to protect: `sanitizeUnicodeText()` is the single pipeline shared by session names and foreign-session text at trust-adjacent boundaries. The tests distinguish the behavior of each stage—NFC normalization, control/bidi replacement, zero-width removal, whitespace collapse, code-point capping, and empty output—so a future drift fails at the owner instead of surfacing only through distant callers.

The 98-line scope is proportionate: one test file, no production change, no new helper or parallel authority, and no repeated full matrix across wrapper APIs. I found no P0-P3 issue. The reported exact-head `test` and `windows_recovery` checks are green.

<details><summary>中文对照</summary>

感谢在共享 Unicode sanitizer 的权威边界补上聚焦测试。我复核了最新提交 `69c03b6708e7c240130e7eefff4d96f72d3f090f`。

这里有明确需要保护的 contract：`sanitizeUnicodeText()` 是 session name 与 foreign-session text 共用的唯一处理流水线，位于接近信任边界的位置。测试分别锁定 NFC normalization、control/bidi 替换、zero-width 删除、空白收敛、code-point 截断和空结果，因此未来某个阶段发生漂移时，会在 owner 处直接失败，而不是只靠较远的调用方间接暴露。

98 行范围是成比例的：只有一个测试文件，不改生产代码，不新增 helper 或并行权威，也没有把完整矩阵重复套在多个 wrapper API 上。没有发现 P0-P3。GitHub 报告的 exact-head `test` 和 `windows_recovery` 均为绿色。

</details>

AI-assisted review: Codex compared the direct suite with existing caller coverage; I independently verified the problem value, scope, exact head, and severity.

PR #3699, opened 2026-08-24
fix(ui): stop tool-well copy button colliding with long first lines and titles

GabrielDrapor:
## Problem

In tool-call detail wells, a long unbroken line — typically a `maka://archive/tool-result-archive-…` URI — collides with the well's copy button. Two distinct layouts produce the same visual overlap:

1. **Headered wells** (the reported case): the result headline becomes the Astryx CodeBlock header title, a flex span with `white-space: pre` and visible overflow. The compact header draws no divider, so the title reads as the well's first content line — and runs straight under the header's copy button, then hard-clips at the container edge.
2. **Headerless wells** (same collision, different mechanism): Astryx floats the copy button absolutely over the scroll viewport's top-right (`copyButtonAbsolute`), and wrapped content's first line passes under it — the wrap point is the container edge, not the button edge.

### Before
before

### After
after

## Fix (CSS only, scoped to `.maka-tool-call-detail`)

- **Headered**: ellipsize the title inside the flex slot the header already sizes (`flex: 1` / `min-width: 0` on its wrapper). `display: block` because a flex container will not ellipsize its anonymous text; tool wells never render the collapsible chevron that the flex display exists for. `white-space` stays `pre` — both `pre` and `nowrap` satisfy `text-overflow`'s no-wrap requirement, but `nowrap` would collapse runs of spaces, and a title is often a literal path where `report  final.txt` and `report final.txt` are different names (this was a review finding, see below).
- **Headerless**: reserve the floating button's footprint with a 28×24px `float: inline-end` `::before` on the first line, so wrapped text flows around the corner. The lengths are the button's geometry, not tunables: the 28px sm IconButton inset `spacing-2` ends 20px past the code's `spacing-4` inline padding (+8px gap), and its bottom edge sits 36px − `spacing-3` = 24px below the first line's top. Anchored on the line div rather than `<code>` because 100+-line blocks wrap lines in `content-visibility` chunks whose layout containment would keep an outer float from shortening their line boxes. Logical properties throughout, so RTL follows the button (`inset-inline-end`).

This mirrors the doctrine already recorded for markdown code blocks (styles.css: markdown always gives CodeBlock a structural header "instead of overlaying the horizontal scroll viewport") while keeping tool wells quiet — no added header chrome.

## Verification

- Verified live against the packaged app via devtools on the reported well: title ellipsizes exactly at the button's edge (boundary 1031/1031, zero overlap); before/after screenshots above are from that app.
- An adversarial review pass (Codex) over the diff confirmed the geometry against the Astryx tokens (`--size-element-sm`, `spacing-2/3/4`), the header-class guard, RTL, consumer scope, and cascade order; its one P2 — `nowrap` collapsing literal spaces in path-like titles — is fixed in the third commit by keeping `white-space: pre`.
- `npm run format:check` clean.

Known trade-off: an ellipsized title's full value is no longer fully visible; it was not readable before either (hard clip + button overlap), it is excluded from the copy button's payload in both states (`pickHeadline` consumes it out of the body), and text selection still yields the full string. If the full headline should be hover-recoverable, a native `title` tooltip needs a small `ToolCodeBlock` component change — happy to follow up if preferred.

https://claude.ai/code/session_01VqBZ8cP7wMpy1mXTrn3VDg

### DIFF
--- a/packages/ui/src/styles.css
+++ b/packages/ui/src/styles.css
@@ -697,12 +697,16 @@
    Ellipsize it inside the title slot the header already sizes (flex: 1,
    min-width: 0). display: block because a flex container will not ellipsize
    its anonymous text; tool wells never render the collapsible chevron that
-   the flex display exists for. */
+   the flex display exists for. white-space stays `pre`, not nowrap: both
+   satisfy text-overflow's no-wrap requirement, but nowrap would collapse
+   runs of spaces — and a title is often a literal path or command where
+   `report  final.txt` and `report final.txt` are different names, with the
+   title as their only display (the copy button copies the body). */
 .maka-tool-call-detail .astryx-codeblock-header .astryx-codeblock-title {
   display: block;
   overflow: hidden;
   text-overflow: ellipsis;
-  white-space: nowrap;
+  white-space: pre;
 }
 .maka-tool-output-chunk { display: contents; }
 .maka-tool-output-chunk-stderr { color: var(--destructive); }
--- a/packages/ui/src/styles.css
+++ b/packages/ui/src/styles.css
@@ -689,6 +689,21 @@
   width: 28px;
   height: 24px;
 }
+/* The headered variant has the same collision inside its header row: the
+   title (often a long unbroken maka://archive URI headline) is a flex span
+   with white-space: pre and visible overflow, so it runs straight under the
+   header's copy button — and headerCompact draws no divider, so the title
+   reads as the well's first content line while colliding like an overlay.
+   Ellipsize it inside the title slot the header already sizes (flex: 1,
+   min-width: 0). display: block because a flex container will not ellipsize
+   its anonymous text; tool wells never render the collapsible chevron that
+   the flex display exists for. */
+.maka-tool-call-detail .astryx-codeblock-header .astryx-codeblock-title {
+  display: block;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
 .maka-tool-output-chunk { display: contents; }
 .maka-tool-output-chunk-stderr { color: var(--destructive); }
 .maka-tool-output-chunk-redacted { opacity: 0.65; }
--- a/packages/ui/src/styles.css
+++ b/packages/ui/src/styles.css
@@ -670,6 +670,25 @@
      a rhythm no other row in the panel has. */
   --text-supporting-leading: var(--text-code-leading);
 }
+/* A headerless CodeBlock floats its copy button over the viewport's top-right
+   corner, and a long unbroken first line — a maka://archive URI, a long
+   command — runs straight under it (markdown avoids this by always giving the
+   block a header toolbar; tool wells stay quiet on purpose). Reserve the
+   button's footprint with a float on the first line so wrapped text flows
+   around the corner instead of colliding. On the line div, not <code>: 100+
+   line blocks wrap lines in content-visibility chunks whose layout containment
+   would keep an outer float from shortening their line boxes. The lengths are
+   the button's geometry, not tunables: the 28px button inset spacing-2 (8px)
+   ends 20px past the code's spacing-4 (16px) inline padding — plus an 8px gap
+   — and its bottom edge sits 36px − spacing-3 (12px) = 24px below the first
+   line's top. Tool wells never enable line numbers, so the numbered-gutter
+   ::before cannot collide here. */
+.maka-tool-call-detail .astryx-codeblock:not(:has(.astryx-codeblock-header)) [data-line="1"]::before {
+  content: '';
+  float: inline-end;
+  width: 28px;
+  height: 24px;
+}
 .maka-tool-output-chunk { display: contents; }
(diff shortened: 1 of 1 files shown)


Astro-Han (approved):
Two CSS rules in `packages/ui/src/styles.css`, both scoped to `.maka-tool-call-detail`, that stop a long unbroken first line — typically a `maka://archive/tool-result-archive-…` URI — from colliding with the CodeBlock copy button. The headerless case reserves the floating button's footprint with a `float: inline-end` `::before` on the first line; the headered case ellipsizes the title inside the flex slot the header already sizes.

What I checked:

- This extends the seam that already exists rather than adding a parallel one. `styles.css` already reaches into `.astryx-codeblock` and `.astryx-codeblock-header` in six places, and the markdown doctrine the PR cites is really there at `styles.css:444-445` — markdown gives CodeBlock a structural header precisely so the copy button lands in a toolbar instead of overlaying the scroll viewport. Tool wells deliberately stay headerless, so they need the other half of that same decision, which is what this supplies.
- Keeping `white-space: pre` instead of `nowrap` is the right call and the reasoning in the comment is correct: both satisfy `text-overflow`'s no-wrap requirement, but `nowrap` collapses runs of spaces, and these titles are frequently literal paths where `report  final.txt` and `report final.txt` are different names. Good catch from the earlier pass; I would have flagged the `nowrap` version.
- Both rules are scoped under `.maka-tool-call-detail`, which is applied at exactly one place (`tool-activity.tsx:243`), so neither rule can reach markdown code blocks or any other CodeBlock consumer.
- The magic numbers are derived, not tuned, and the comment records the derivation inline. That is the difference between a fix and a nudge, and it is why I am comfortable with hard pixel values here.
- The stated trade-off is honest: the ellipsized title's full value is no longer visible, but it was not readable before either, and selection still yields the whole string.

One P3, non-blocking: `[data-line="1"]` is a new coupling to Astryx's internal rendered DOM. The existing `data-line` selectors in this file are on our own `.maka-tool-diff-line`, so this is the first rule that depends on how `@astryxdesign/core` numbers its line elements. It will fail silently and invisibly on an Astryx upgrade — no test, no type, no build error, just the collision quietly coming back. Class names like `.astryx-codeblock-header` are at least plausibly stable public surface; a positional `data-line` index is less obviously so. A Storybook story covering a long-URI tool well would turn the silent revert into a visible diff. Worth doing at some point, not worth blocking on.

Approving.

*AI use: Claude Code (Opus) confirmed the cited markdown doctrine and the existing Astryx selectors in `styles.css`, and checked that `.maka-tool-call-detail` has exactly one consumer. The geometry was verified by the author against the live app; I did not re-measure it. The reviewer of record reviewed and accepted this.*

<details>
<summary>简体中文</summary>

`packages/ui/src/styles.css` 中两条规则，都限定在 `.maka-tool-call-detail` 下，解决长的不可断行首行（通常是 `maka://archive/tool-result-archive-…` URI）与 CodeBlock 复制按钮相撞的问题。无 header 的情形用首行上的 `float: inline-end` `::before` 预留按钮占位；有 header 的情形在 header 已经定尺寸的 flex 槽内把标题省略号化。

核对了五点：

- 这是扩展既有接缝而非新建并行路径。`styles.css` 已在六处伸进 `.astryx-codeblock` 和 `.astryx-codeblock-header`，PR 引用的 markdown 原则确实在 `styles.css:444-445`——markdown 总是给 CodeBlock 一个结构化 header，正是为了让复制按钮落在工具条里而不是浮在滚动视口上。工具井有意保持无 header，因此需要同一决策的另一半，这正是本 PR 提供的。
- 保留 `white-space: pre` 而非 `nowrap` 是对的，注释中的推理也成立：两者都满足 `text-overflow` 的不换行要求，但 `nowrap` 会折叠连续空格，而这些标题常是字面路径，`report  final.txt` 与 `report final.txt` 是不同的名字。上一轮抓得好，`nowrap` 版本我也会标出来。
- 两条规则都限定在 `.maka-tool-call-detail` 下，而它只在 `tool-activity.tsx:243` 一处使用，因此都影响不到 markdown 代码块或其他 CodeBlock 消费方。
- 那些数值是推导出来的而非调出来的，注释就地记录了推导过程。这是"修复"和"凑数"的区别，也是我能接受这里写死像素值的原因。
- 声明的取舍是诚实的：省略号化后标题全值不再可见，但此前也读不到，且选中仍能拿到完整字符串。

一条 P3，不阻塞：`[data-line="1"]` 是对 Astryx 内部渲染 DOM 的新耦合。本文件既有的 `data-line` 选择器都作用于我们自己的 `.maka-tool-diff-line`，所以这是第一条依赖 `@astryxdesign/core` 如何给行元素编号的规则。Astryx 升级时它会静默失效——没有测试、没有类型、没有构建错误，只是碰撞悄悄回来。`.astryx-codeblock-header` 这类类名至少还算可信的公开表面，位置性的 `data-line` 序号就不那么明显了。补一个覆盖长 URI 工具井的 Storybook story，能把静默回退变成可见 diff。值得做，但不值得为此阻塞。

Approve。
</details>

PR #3700, opened 2026-08-24
feat(desktop): read a custom relay's model catalog before saving

rootkiller6788:
## Summary

A custom relay has no published model catalog — the only place its catalog can be read from is the endpoint the user is typing into the form, and there was no way to read it before the connection exists. The form therefore forced a hand-typed model id.

This adds a "read model catalog" action that probes the endpoint and key the user already filled in, then turns the default-model field into a chooser.

Nothing in the runtime host needed changing. The base-url plumbing for `onboarding-verify` landed upstream for managed-Host discovery, and it already runs the credential → catalog discovery this needs against a transient connection without persisting anything.

This change only wires the desktop side to it:

- an IPC handler
- a shared contract so main/preload/renderer cannot drift
- the row that offers the discovered models

The probe persists nothing and cannot break a reference. It targets `connectionId: null` against a transient connection — the same shape the post-create discovery fetch uses — so probing never creates, edits, or names a connection.

## Staleness

The read is against whatever endpoint and key are in the fields at click time. If either changes while the probe is in flight, the result is discarded rather than attached to the new inputs — a picked model can never belong to a catalog probed from a different endpoint.

## Rejected → form copy

The host returns a closed set of rejection reasons. Most map straight onto a sentence in the form:

- `credential_not_configured`
- `base_url_not_configured`
- `provider_unsupported`

One — `connection_not_found` — cannot describe this probe because it targets no existing connection, so it is folded into a generic failure instead of a reason the form cannot render.

## Verification

Eight new tests, each rule pinned by reverting it:

```text
$ node --test apps/desktop/dist/main/**tests**/provider-add-catalog-probe.test.js
ℹ tests 5
ℹ pass 5
ℹ fail 0

$ node --test apps/desktop/dist/main/**tests**/runtime-host-connections-ipc-main.test.js
ℹ tests 13
ℹ pass 13
ℹ fail 0
```

The following all pass:

```text
npm run typecheck
npm run lint
npm run format:check
git diff --check
npm run check:asf-headers
```

The full workspace suite does not pass on this machine, and does not pass identically on a clean `main` — the failures are Windows-environment related:

- `EPERM` symlink
- `EBUSY` sqlite
- tar path resolution
- missing native binaries for Rive/computer-use
- macOS-only service tests
- three test files that hang in process isolation

I confirmed the biggest cluster by stashing this branch and rerunning `goal-coordinator`: it fails 8/8 on clean `main` exactly as it does here.

Unrelated to this change, reported rather than left unmentioned.

## AI use

Tool(s) and scope: Claude Code — traced the existing `onboarding-verify` path to confirm the desktop was the only missing layer, wrote the shared contract, the IPC handler and its projection, the draft logic and the form row, and their tests. The commit carries a `Generated-by` trailer.

## Checklist

### Does this PR entail a change in behavior?

- [ ] No


rootkiller6788:
Please let me know if there are any issues, thank you.

rootkiller6788:
Fixes #3442 — lets the custom-relay add form probe a relay's model catalog (via connection.onboarding.verify against a
  transient connection, persisting nothing) and offer the discovered models in a chooser instead of a hand-typed id.

apps/desktop/src/main/runtime-host-connections-ipc-main.ts
```diff
@@ -289,6 +295,20 @@ export function registerRuntimeHostConnectionsIpc(
...
   });
+  // Probe the model catalog an endpoint serves BEFORE a connection exists, so
+  // the custom-relay form can offer a real choice instead of a guess. Runs the
+  // same discovery the post-create fetch uses against a transient connection;
+  // `connectionId` is null because there is nothing to edit yet.
+  deps.ipcMain.handle('connections:probeModels', async (_event, raw: unknown) => {
+    const input = normalizeConnectionCatalogProbeInput(raw);
+    const result = await deps.client.probeConnectionModels({
+      providerType: input.providerType,
+      connectionId: null,
```
Astro-Han, line 308: [P1] Suggestion (category ① — normal add-a-second-relay path; also a credential trust boundary): could this probe use a genuinely transient target rather than `connectionId: null`? In the current Host contract, null targets the provider’s canonical slug. If that connection already exists, `beginConnectionOnboarding()` returns its stored API key and request-header secret; `#discoverOnboarding()` then combines those secrets with the new form’s `baseUrl` and sends the discovery request to that new endpoint. A user adding a second custom relay with an empty/new key can therefore disclose the existing relay’s credentials and get a catalog for the wrong credential/header combination. I suggest a transient probe operation that carries only material explicitly present in this form, plus a regression with an existing canonical relay asserting its key/headers never go to the new endpoint. Please push back if I have misunderstood the null-target invariant.

apps/desktop/src/renderer/settings/provider-add-form.tsx
```diff
@@ -355,14 +366,25 @@ export function AddProviderForm(props: {
...
+              label={copy.defaultModel}
+              description={copy.defaultModelHelp}
+            />
+            <CatalogProbeField
+              providerType={props.providerType}
+              baseUrl={baseUrl}
+              apiKey={apiKey}
+              value={defaultModel}
+              isDisabled={isExperimental || busy}
+              probeModels={(request) => props.bridge.probeModels(request)}
```
Astro-Han, line 385: [P2] Suggestion (category ① — normal custom-relay setup): could the probe include the normalized request headers from the Advanced Request Editor? The save path sends `requestHeaders`, but this probe contract carries only provider/base URL/API key. Relays whose `/models` endpoint requires a custom header will therefore fail before saving even though the immediately saved connection can discover successfully. I suggest using the same normalized header material in a transient Host-owned probe and adding that case to the probe regression coverage. Please push back if the pre-save action is intentionally narrower than post-save discovery.

Astro-Han (commented):
Thanks for adding a pre-save catalog probe; the problem in #3442 is demonstrated and the Desktop → IPC → Runtime Host path is otherwise well aligned with the existing discovery owner. I reviewed exact head `b6b1cf51af4a640170409a81a80c123a38a4f2e5` and left two suggestions inline: one P1 credential-boundary issue and one P2 behavior gap. These are suggestions from an adjacent reviewer’s view, so please do push back if the onboarding contract or intended form behavior has context I missed. The exact head currently has no completed CI jobs (`action_required`), so it will also need a fresh exact-head gate after revision.

谢谢补上保存前读取模型目录。#3442 的问题和整体调用链都成立；当前有两点建议：`connectionId: null` 会复用 canonical connection 的旧凭据/headers 并把它们发往新 endpoint（P1），以及表单自定义 headers 没有参与 probe（P2）。这些是旁观评审视角，如果 onboarding 合同或产品语义另有上下文，欢迎直接 push back。修改后还需要重新获得 exact-head CI。

</details>

rootkiller6788:
Thanks for catching both — you were right on each.

  - P1 (credential leak): the probe now passes transient: true on connection.onboarding.verify, which skips
    canonical-slug resolution in storage, so a blank key no longer reuses the existing relay's stored credential. A
    blank key now returns credential_not_configured instead of probing with the wrong secret.
  - P2 (headers gap): the form's custom request headers ride along on the probe request and are validated at the IPC
    boundary, matching what the save path sends.

  Added regression coverage: a transient probe against an existing canonical relay asserts its key and headers never
  reach the new endpoint, plus decode coverage for the new transient/requestHeaders fields.

Astro-Han (commented):
Verified against head ae5236cc4 — and there's a structural problem to surface before line-level findings: **this capability already shipped on main**. #3443 ('let a relay's model catalog answer for itself', commit 49a54bf63) landed a generalized version of the same flow: the host's `connection.onboarding.verify` operation already supports `target: {kind:'create'|'existing'} + baseUrl` with transient discovery (fresh candidate, no stored credential/header resolution), and the desktop already has `connections:onboardingVerify` IPC + `verifyOnboarding` preload + a managed verify→choose→save form flow in provider-add-form.tsx. This PR is built on a pre-#3443 base and re-implements the same path under different names (`transient: true` ≈ `target.kind:'create'`, `connections:probeModels` ≈ `connections:onboardingVerify`, `ConnectionCatalogProbe*` ≈ `ConnectionOnboardingVerify*`).

So the decision before any line review: **rebase onto main's machinery and contribute only the delta, or close**. The honest delta I can find is `requestHeaders` forwarding on the probe — worth checking whether main's verify already carries it; if not, that's a small additive PR against the existing operation, not a parallel one.

If this does continue after rebase, the diff's own issues:

**P2 — cross-origin redirects forward credentials to the redirect target** (③ trust boundary). fetchForConnectionEffect (connection-effect-fetch.ts:82-91) leaves undici's `redirect: 'follow'` default; per fetch spec only authorization/cookie/host headers are stripped on cross-origin hops — `x-api-key` (anthropic-compatible relays) and custom headers ride to wherever the 30x points, including https→http downgrades and internal hosts the baseURL gate never vetted. Inherited caveat: connection.models.fetch/test.run on saved connections already have this exposure, so this PR widens a trigger rather than creating the mechanism — but a /models catalog has no legitimate need to redirect, so `redirect: 'error'` (or manual revalidation like local-web-fetch.ts:79-99) is a one-line hardening worth doing for the whole effect-fetch path.

**P3 — parallel DTO layer.** shared/connection-catalog-probe.ts + projectConnectionCatalogProbe re-state the verify wire types for one consumer with a cosmetic ready/verified rename; the connection_not_found remap is unreachable under transient. The host types are already renderer-safe — delete the layer and type the IPC with the wire types directly (~150-200 lines).

**P3 — ConnectionOnboardingSaveInput extends the verify input**, re-admitting transient/requestHeaders to in-process save callers (#3299's review closed exactly this gap by keeping save input standalone — spell the shared fields instead).

**P3 sweep**: connections-ipc-validation.ts:163 PROVIDER_DEFAULTS bracket lookup admits inherited members (use providerDefaultsOf per the codebase's own rule); probe apiKey lacks the 4096 cap/control-char check the create path applies; decode bounds models only to non-empty (add the catalog cap for symmetry); probeCatalogHelp copy key is defined but never rendered; a non-string display_name fails the entire discovery where a typeof guard would degrade gracefully; the 'local relay may accept a catalog read without credentials' comment only holds for optional_api_key providers.

**Needs one check**: the PR adds keys to the closed verify wire schema — confirm the compatibility epoch bumps accordingly, or an older Host rejects the fields mid-operation rather than at handshake.

PR #3611, opened 2026-08-23
fix(ui, cli): humanize provider retry delay (3401)

yunaremaia:
Fixes #3401.

When a provider returns a long `Retry-After` (~13565s for a 5h quota reset), the retry banner renders `Retrying in 13565s` / `13565 秒后重试` — illegible. Related to #3393 (countdown ticking).

## What changed

- **`packages/ui/src/conversation-copy.ts`** — new `formatRetryDelay(seconds, locale)`: `13565` → `3h 46m 5s` (en) / `3小时 46分 5秒` (zh), 1s granularity. Wired into both zh/en `providerRetryScheduled` catalog entries.
- **`packages/cli/src/pi-transcript.ts`** — new `formatRetryDuration(delayMs)` for the TUI activity strip; same humanized output (en).

Granularity stays at one second so the banner visibly ticks (minute-granularity would reintroduce the frozen look from #3393).

## Verification

- `formatRetryDelay` spot-checked locally (`13565→3h 46m 5s`, `90061→1d 1h 1m 1s`, edge: `0.2→1s`).
- No registry workaround; change is compile-only (no new deps).
- Full local build/test not run — relies on `@xterm/xterm` custom registry (`ms-feed-2.pkgs.visualstudio.com`); CI is the verification layer.

Closes #3401.
Related: #3393.

PR #3701, opened 2026-08-24
test(ui): pin the humanized provider retry banner copy across locales and scales

me2seeks:
Follow-up to #3611 (and the closed #3402): @Astro-Han asked for the unit-test cases from #3402 to be sent standalone, since #3611 shipped the formatter without tests.

This adds one focused test pinning `providerRetryScheduled` in both locales:

- second/minute/hour/day scales (`45秒`, `1分 15秒`, `4小时 28分 3秒`, `1天 1小时 1分 1秒` — and the en equivalents)
- the ceil-to-one-second floor, so a zero-length wait reads `1秒后重试` / `Retrying in 1s` and never `0秒后重试`
- zero-order units are skipped rather than rendered as `0分` / `0m`

Also documents (per Astro-Han's note on #3402) that the zh short-delay string intentionally changed shape: `45 秒后重试` → `45秒后重试` (space dropped).

Generated-by: maka

### DIFF
--- a/packages/ui/src/__tests__/conversation-copy.test.ts
+++ b/packages/ui/src/__tests__/conversation-copy.test.ts
@@ -24,3 +24,35 @@ import { getConversationCopy } from '../conversation-copy.js';
 test('labels the Chinese default thinking level as default', () => {
   assert.equal(getConversationCopy('zh').model.defaultLevel, '默认');
 });
+
+/**
+ * A subscription quota window can hand the runtime an hour-scale Retry-After;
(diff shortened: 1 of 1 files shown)


Astro-Han (commented):

Single file adds locale/scale focused tests for provider retry display; no production logic.

No P0-P3. Checks on `daf66dc70` — code GO.

Astro-Han (approved):
APPROVE — daf66dc70, test-only, no P0-P2.

PR #3704, opened 2026-08-24
fix(desktop): make proxy credential updates explicit and Host-owned

### LINKED ISSUE #3696: bug(desktop): proxy password editing stores the masking sentinel and corrupts credentials
## What happened

Maka Desktop treats the proxy-password masking sentinel (`••••••••`) as the editable input value.

When an existing proxy password is loaded, the password field's actual value is eight bullet characters rather than an empty value with a placeholder. Typing a replacement password triggers a Settings update on every keystroke. Each response projects the saved credential back to the same eight-character sentinel, causing the newly typed character to disappear.

More seriously, any value that is not exactly equal to the sentinel is accepted as a new credential. Normal typing can therefore store a value shaped like:

```text
••••••••<one newly typed character>
```

This silently destroys the previous proxy credential.

The adjacent actions are also misleading:

- **Show password** reveals the literal eight-character sentinel rather than the newly entered or saved password.
- **Copy password** copies the sentinel when clipboard access is available. It cannot copy the saved credential because the real secret is intentionally not returned to the Renderer.

This is deterministic credential corruption, not only a visual masking problem.

Expected behavior:

- An existing proxy credential is represented as configuration state, not as an editable password string.
- The input's actual value is empty and its placeholder says that a password is already saved.
- The user can type a complete replacement password without characters disappearing.
- Typing updates only a local draft; it does not persist the credential on every keystroke.
- Blur or Enter submits the complete replacement once.
- Escape cancels the replacement and preserves the existing credential.
- An empty draft preserves the existing credential.
- Show/hide controls only the newly entered draft. It never reveals the saved credential.
- The proxy-password field does not show a copy action.
- Copy actions on other password and credential fields remain unchanged.
- Password content never appears in ordinary Settings storage, logs, errors, or update responses.

## How to reproduce

1. Start Maka Desktop from source.
2. Open **Settings → General → Network**.
3. Enable **Proxy server**.
4. Enable **Proxy authentication**.
5. Start with an existing saved proxy password.
6. Observe that the password input contains `••••••••` as its actual value.
7. Click the password field without selecting or clearing the sentinel.
8. Type a disposable test password normally, one character after another.
9. Observe that newly entered characters disappear and the field repeatedly returns to eight bullets.
10. Select **Show password**.
11. Observe that the field still shows eight bullet characters rather than the replacement password.
12. When clipboard permission is available, select **Copy password** and paste into a safe local scratch field.
13. Observe that the pasted value is eight bullet characters.
14. Inspect only redacted credential metadata, or test against an authenticated local proxy.
15. Observe that the stored credential has been replaced with a value beginning with the eight-character sentinel rather than the complete test password.

Do not use a real proxy password for this reproduction because the current implementation can destroy it.

## Environment

- Maka version or commit: `74dcd56bd` (`@maka/desktop@0.2.0`)
- OS and version: macOS 15.6 (`24G84`)
- Surface: Desktop development build / Runtime Host
- Renderer URL: `http://localhost:5173`
- Node.js version: v24.19.0
- Proxy: HTTP/HTTPS at `127.0.0.1:7897`
- Proxy authentication: enabled
- Existing proxy credential: configured

### Observed development-build evidence

The issue was reproduced through the local Maka Dev UI. The relevant implementation files are unchanged from `upstream/main` at `74dcd56bd`.

After entering a disposable replacement password:

- the UI returned to eight bullets;
- the Runtime Host proxy credential was updated at `2026-08-24 15:46:44 CST`;
- the stored value had 9 characters;
- the stored value began with `••••••••`;
- the stored value was not equal to the pure eight-character sentinel.

The raw credential was not printed or included in diagnostics. The metadata proves that the stored value had the form “masking sentinel plus one character.”

Selecting **Show password** still displayed eight bullets, confirming that the visibility action operates on the sentinel string rather than the saved credential.

Computer-use automation could not directly verify the clipboard payload because macOS rejected clipboard access and Maka correctly displayed:

> Copy failed — The clipboard is unavailable or was denied by the system.

Independent manual reproduction shows that successful clipboard access copies eight bullets. This also follows directly from the current implementation, which copies `props.value`.

### Root cause

`apps/desktop/src/main/runtime-host-settings-ipc-main.ts` projects a configured credential into the editable settings model:

```ts
password: proxyCredential?.configured ? SENSITIVE_PLACEHOLDER : "",
```

`apps/desktop/src/renderer/settings/general-settings-page.tsx` binds that value directly to the password input and persists every change:

```tsx
<PasswordInput
  value={proxyDraft.password}
  onChange={(next) => void updateProxy({ password: next })}
  label={copy.password}
/>
```

`apps/desktop/src/renderer/settings/password-input.tsx` treats the supplied value as ordinary editable content and copies it directly:

```tsx
<TextInput
  type={visible ? 'text' : 'password'}
  value={props.value}
  onChange={(value) => props.onChange(value)}
/>
```

```ts
await navigator.clipboard.writeText(props.value);
```

The main process preserves only an exact sentinel match. Any sentinel-plus-character value is stored as a new credential:

```ts
if (
  proxy.password !== undefined &&
  proxy.password !== SENSITIVE_PLACEHOLDER
) {
  if (proxy.password.length === 0)
    await deleteCredential(client, PROXY_CREDENTIAL);
  else
    await setCredential(client, PROXY_CREDENTIAL, proxy.password);
}
```

The resulting sequence is:

```text
saved credential
  → Renderer receives "••••••••"
  → user types one character
  → Renderer submits "••••••••x"
  → Runtime Host stores "••••••••x"
  → Settings response returns "••••••••"
  → newly typed character disappears
```

### Suggested direction

Separate “a credential is configured” from “the user is entering a new credential.”

For the proxy settings path:

- Replace the Renderer-facing `password` sentinel with a read-only `passwordConfigured: boolean`.
- Keep the real secret exclusively in the Runtime Host credential store.
- Use explicit credential mutations:
  - no credential mutation means keep the current credential;
  - `replace` carries one complete non-empty secret;
  - `delete` removes the credential.
- Keep the replacement password in a local Renderer draft.
- Commit the complete draft once on blur or Enter.
- Cancel on Escape.
- Treat an empty draft as “keep the existing credential.”
- Deduplicate Enter followed by blur.
- Make proxy testing wait for a pending replacement and abort the test if saving fails.
- Serialize credential replacement with disabling proxy authentication so a late replacement cannot recreate a deleted credential.
- Preserve the show/hide action for the local replacement draft.
- Add an opt-out to the shared `PasswordInput` copy action and disable it only for the proxy-password field.
- Do not change copy behavior for model API keys, web-search credentials, Bot tokens, or other password inputs.

The legacy `SENSITIVE_PLACEHOLDER` contract can remain temporarily for other credentials; it should no longer participate in the proxy-password read or write path.

### Suggested regression coverage

Renderer and component coverage:

- A configured proxy password produces an empty DOM input value and a localized “saved; enter a new password to replace it” placeholder.
- The first typed character remains visible and does not contain the sentinel.
- Sequential typing does not issue per-character Settings updates.
- Blur submits the complete draft once.
- Enter submits once, and its following blur does not submit again.
- Escape cancels without modifying the saved credential.
- Empty blur preserves the saved credential.
- Save failure preserves the complete local draft.
- Show/hide controls the new draft.
- The proxy-password input has no copy button.
- Other `PasswordInput` callers retain their copy button.

Main-process and Runtime Host coverage:

- Credential presence projects to `passwordConfigured`.
- Proxy Settings responses contain neither the secret nor `SENSITIVE_PLACEHOLDER`.
- An omitted credential mutation preserves the existing credential.
- `replace` writes one complete secret.
- `delete` removes the credential.
- Empty replacement values are rejected.
- Secrets do not enter the ordinary Settings Store or error output.
- Proxy testing waits for a pending replacement.
- Disabling authentication wins over any older in-flight replacement.

Desktop end-to-end coverage:

1. Seed an isolated Runtime Host with an existing disposable proxy credential.
2. Open the development Settings UI.
3. Enter a replacement password normally.
4. Exercise show/hide.
5. Commit and reload Settings.
6. Confirm that the UI reports a configured credential without exposing it.
7. Use a controlled authenticated local proxy to verify that the complete replacement password is used.

### AI assistance

OpenAI Codex was used to reproduce the behavior in the local Maka Desktop development build through Computer Use, inspect only redacted Runtime Host credential metadata, trace the Renderer → Desktop IPC → Runtime Host data flow, and draft this report. The human contributor reviewed the findings and remains responsible for the submission.

Sun-GLiang:
## Summary

This PR fixes #3696 as a proxy-credential lifecycle migration rather than an input-mask-only patch.

- remove the saved proxy password from persisted `AppSettings` and Renderer input values; reads expose only `passwordConfigured`, while writes use explicit keep/replace/delete semantics
- make proxy policy and credential changes one Host-owned CAS operation with shared preflight, ordered publication, and an explicit `commit_outcome_unknown` result for partial durable failure
- keep replacement text in a Renderer-local draft until blur or Enter; Escape cancels, Eye reveals only the draft, window blur does not commit, and Copy remains available on unrelated password fields
- adapt schema-v1 import/export at the compatibility boundary so credential-only transfers remain lossless without carrying unrelated settings
- bind imported credentials to exact Host-owned Connection and proxy targets, rejecting unbound or mismatched imports before secret writes
- atomically remove legacy proxy-secret fields from `settings.json`, keeping the credential vault as the only durable secret authority
- make “Test current configuration” wait for a pending password commit; unsaved candidate-password testing is intentionally no longer supported

## Why this lands together

The defect crosses one read/write representation boundary: a vault-owned secret was projected into ordinary Settings as an editable sentinel and then accepted back as a credential. Fixing only one layer leaves an unsafe intermediate contract.

- Removing `password` from Settings requires a read-only configured-state projection and a write-only credential mutation.
- The Renderer draft requires that explicit Host mutation; otherwise it must fall back to the sentinel or split policy and credential writes again.
- The Host operation must validate policy revision and credential generation in one lane so an older client cannot recreate a credential after authentication is disabled.
- Once the Settings shape changes, schema-v1 transfer needs an adapter plus exact target binding so compatibility paths cannot retain, misroute, or reintroduce the removed secret field.
- The Runtime Host epoch moves with this boundary because older peers can split these writes and violate the shared credential basis.

These areas remain separately testable, but under this implementation they are not separately releasable without adding a temporary dual representation or compatibility surface that preserves the defect. One revert restores the previous complete lifecycle; partial rollback of a dependent layer is not supported.

## Review boundaries

The review can be evaluated as three explicit contracts inside that compatibility cut:

1. **Migration and transfer:** legacy Settings cleanup, schema-v1 credential-bearing field adaptation, and Host target binding.
2. **Host authority:** policy/vault preflight, two-client CAS, publication ordering, and partial-outcome reporting.
3. **Desktop UX:** local draft, commit/cancel/reveal behavior, save failure handling, and proxy-only Copy removal.

The focused suite keeps the invariant-bearing coverage for each boundary: minimal migration and credential-only transfer cases, Host two-client and durable failure cuts, Renderer draft concurrency/failure cases, and one offline authenticated-proxy E2E.

## UI behavior

The recording shows sequential draft input, working show/hide, safe reload state, Escape cancellation without closing Settings, replacement after a saved password, and the proxy-only removal of Copy.

Proxy password editing regression flow

## Conflict resolution

- Merged the latest `apache/main` at `dfe457eac9` into the PR branch, producing merge commit `9aa343f19`.
- Preserved the PR's proxy credential target-binding and atomic update behavior while incorporating the mainline model-catalog projection and compatibility epoch changes.
- Kept the merged Runtime Host compatibility epoch at 87, with the PR's credential-boundary notes retained alongside the mainline epochs 84–87, and updated the affected assertions and Windows test inventory.

## Verification

Post-merge verification against the merged `apache/main` state on 2026-09-01:

- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run check:renderer-architecture` (61/61)
- `npm run build:test`
- `npm run test:dist -- --workspaces=packages/core,packages/storage,packages/runtime-host,apps/desktop --concurrency=3` (all selected workspace suites passed; 0 failures)

## AI use

Select exactly one:

Tool(s) and scope: OpenAI Codex implemented the fix, tests, verification, review-driven corrections, and this scope clarification.

## Checklist

- [ ] No

## CI follow-up

- Fixed the Windows release qualification test's POSIX-only path assertions in commit a090b8384; it now uses platform-native paths.
- npm run check:release passes locally (155/155).


apps/desktop/src/main/runtime-host-settings-ipc-main.ts
```diff
@@ -234,26 +321,20 @@ function projectWebSearchCredential(
...
-    else if (
-      proxy.password !== undefined &&
-      proxy.password !== SENSITIVE_PLACEHOLDER
-    ) {
-      if (proxy.password.length === 0)
-        await deleteCredential(client, PROXY_CREDENTIAL);
-      else await setCredential(client, PROXY_CREDENTIAL, proxy.password);
-    }
+    else if (proxy.credential?.kind === "replace")
+      await setCredential(client, PROXY_CREDENTIAL, proxy.credential.secret);
```
jackwener: [P1] Make proxy policy and credential replacement one Host-owned operation across clients. The new queue is local to one `RuntimeHostSettingsModule`, so two supported Desktop clients have independent lanes. Client A can start a replacement and pause at the credential CAS; client B can then set `authEnabled=false` and delete the credential; when A receives `credential_stale`, its retry rereads the now-empty locator and recreates the secret with `expected: null`. A production-adapter/CAS probe ended with `authEnabled=false` while the vault again contained A's replacement secret. The existing "disable wins" test uses one module and therefore cannot cover this inter-client ordering. Please move the policy decision plus keep/replace/delete into one Runtime Host atomic or recoverable operation that validates both the policy revision and credential basis, and add a two-client test asserting the vault remains empty after disable.

apps/desktop/src/main/runtime-host-settings-ipc-main.ts
```diff
@@ -234,26 +321,20 @@ function projectWebSearchCredential(
...
 }
 
-async function applyHostPatch(
+async function applyHostPatchWithoutLane(
   client: RuntimeHostSettingsClient,
   patch: UpdateAppSettingsInput,
 ): Promise<void> {
   if (patch.network?.proxy) {
     const proxy = patch.network.proxy;
     await client.updateRuntimePolicy((policy) => ({
```
jackwener: [P2] Do not commit proxy policy before the credential write can still fail. A normal update containing both policy fields and `credential: replace/delete` first commits `set_network_proxy` here, then performs vault persistence below. If that second step rejects, the caller receives an error even though the Host is already using the new host/auth/username with the old credential. This is also reachable from the schema-v1 settings-plus-credentials import path. A production-adapter failure probe observed the rejected call with `enabled=true`, `authEnabled=true`, `username=new-user`, and the old secret still stored; the success control stored the replacement. Existing failure coverage sends a credential-only patch, so it never checks policy after the failure. Please make this a single Host-side compound operation, or define an explicit recoverable commit protocol, and test each persistence cut for committed/unchanged/known recovery semantics.

jackwener (commented):
Reviewed exact head `c5ec40335dcaa0d19756bff067bced9cf2cf687c`. I found two blocking correctness issues in the new proxy-credential write path: one cross-client race that can recreate a credential after authentication was disabled, and one partial-commit path that applies proxy policy even though the API reports failure. The hosted checks are green and the branch merges cleanly, but these state-consistency issues need to be resolved before approval.

apps/desktop/src/main/runtime-host-settings-ipc-main.ts
Sun-GLiang: Fixed in 2948461d4. Desktop no longer performs a client-local credential CAS/retry. It sends the observed policy revision, credential basis, complete proxy policy, and keep/replace/delete intent through the Host-owned `runtime.policy.network-proxy.update` operation. The Host rejects a stale client before it can recreate a deleted credential.

Added coordinator and storage tests with two clients observing the same initial basis: client B disables authentication, then client A attempts the stale replacement. The replacement is rejected, the final policy remains `authEnabled: false`, and the vault remains empty.

apps/desktop/src/main/runtime-host-settings-ipc-main.ts
Sun-GLiang: Updated in 7be1502b3. The Desktop path now sends one Host-owned `runtime.policy.network-proxy.update` operation with the observed policy revision, credential basis, complete policy, and explicit keep/replace/delete intent. The Host validates both bases in its mutation lane.

Correction to my earlier reply: this implementation does not persist a transaction intent and does not perform automatic recovery. Storage orders replacement before enabling credential use, and stops policy use before deletion. A failure before the first publication leaves policy and credential unchanged; a later failure after one durable effect returns `commit_outcome_unknown`. Persistence-cut tests now cover both the unchanged and known-unknown cases for replacement and disable/delete.

Sun-GLiang:
CI status note: `package` passed. The `test` job failed on the unrelated Desktop E2E case `slash-command-menu.spec.ts: an open menu keeps its container and skills group across projection refreshes`, where CI observed 2 DOM removals instead of 0.

This PR does not change the slash-command/menu implementation or that test relative to current `upstream/main`. The exact failing case passed locally once and then 10/10 repeated runs; the proxy-password E2E also passed in the CI run. This is therefore consistent with an existing full-suite timing flake rather than a proxy-password regression. Failed job: https://github.com/apache/maka/actions/runs/32820203074/job/97716509006

Astro-Han (dismissed):
Submitted with the wrong review state — the findings here are P2 and do not block. Re-posted as a review comment below.

Astro-Han (commented):
The core fix is right, and it is right at the level that matters. #3696's root cause is that the masking sentinel and the real secret were one representation: `password` lived on the persisted `NetworkProxySettings`, the renderer bound it as the editable value, and `applyHostPatch` accepted anything `!== SENSITIVE_PLACEHOLDER` as a new credential. This PR deletes the field, replaces it with a per-read derived `passwordConfigured`, and makes the write side an explicit keep/replace/delete operation that cannot round-trip a read value. That removes the duplicate representation rather than patching the mask, and it does not introduce a new authority — the vault stays sole owner.

Two things to resolve before merge, then some smaller items.

## Blocking — the outstanding review rests on a premise the branch does not implement

This is not a code defect, so I am not giving it a severity; it blocks for a different reason.

The reply to @jackwener's P2 states that proxy policy and credential changes "now use one Host-side recoverable operation backed by a **persisted transaction intent**", with "**persistence-cut coverage** for the compound operation" and an intent that "is cleared after recovery", citing `2948461d4`. That commit is not in this branch's history (force-pushed; head is `6d1daf75f`), and none of it is present: `git diff <merge-base> HEAD` adds zero occurrences of a transaction intent, and the failure path is `throw commitOutcomeUnknown('Network proxy update committed only some effects')` after ordering the two commits and tracking a `durableChange` flag. There is no persisted intent, no recovery, and no failure-injection test for the compound commit.

The implemented behavior is defensible — @jackwener asked for committed/unchanged/**known** recovery semantics, and `commit_outcome_unknown` is a known label. But the reviewer should not approve on the stated description. Please either correct the reply or add what it describes.

For the record, @jackwener's **P1 is genuinely fixed**: `runtime.policy.network-proxy.update` validates `expectedPolicyRevision` and `expectedCredential` inside one `inLane` on the Host, Desktop no longer does a client-local CAS/retry, and both storage-level and coordinator-level two-client tests were added.

## Scope — this is two intents

```
prod  +1021  -160   (26 files)
test  +1507   -41   (15 files)
```

Cohesively, (a) the data-model fix, the new Host operation with its epoch bump, and the legacy `settings.json` purge are one revertable intent and are the PR. (b) The config export/import schema-v1 adaptation and the `runRuntimeHostSettingsExclusive` lane refactor are a second: different subsystem, independently reviewable, and the source of the first P2 below — which has nothing to do with mask corruption. Splitting also drops what a reviewer must hold at once from 41 files to roughly 30 / 11.

## P2 — a credentials-only export carries and re-applies the whole settings payload

`runtime-host-config-ipc-main.ts:195`. The gate widened to `selected.has('settings') || selected.has('credentials')`, and for a credentials request the payload is `restoreHostSettingsSecrets(settings, secrets)`. Since `buildConfigBundle` derives the manifest from data presence, `'settings'` enters `includedData`; on the other side `applyConfigImport` applies it on presence alone with no category filter (`config-transfer-service.ts:102`).

So exporting **credentials only** to move API keys to a second machine also ships privacy, chatDefaults, workspaceInstructions, subagents, shell, projects, personalization and the full proxy host/port — and importing overwrites all of the target's settings. The manifest does list `settings` and the import summary does say "Settings applied", so it is disclosed and recoverable, which is why this is P2 and not higher. It is still not what the user asked for. The schema-v1 motivation is legitimate; narrow it to the credential-bearing fields, or make the export state the dependency rather than smuggle the section.

## P2 — window blur commits a partially typed password

`password-input.tsx:129`. `relatedTarget` is `null` on window blur, so `destination` is falsy and `onFocusExit()` fires. Cmd-Tab mid-typing replaces the stored credential with a partial password. It is recoverable by retyping, but the user gets no signal — proxy auth simply starts failing later with nothing pointing back here. Worth a guard on `document.hasFocus()` or `relatedTarget === null && document.activeElement === input`.

(I checked that `onBlurCapture` lands on a real DOM node: `InputGroup` spreads `...rest` onto its inner `role="group"` div, so `currentTarget.contains(relatedTarget)` works and clicking the Eye button correctly does not commit.)

## P2 — a dead secret-carrying override, and an unstated behavior change

`toProxyTestInput` no longer sets `password`, and it is the only production producer of `TestProxyInput.proxy` (the command-palette path passes `undefined`). So `credentialOverride(input.proxy?.password)` at `runtime-host-settings-ipc-main.ts:215` is unreachable in production, and `password` on the `network-proxy.test` operation is a secret-carrying IPC and protocol field with zero producers. This PR already bumps the epoch 50 to 51, which makes it the cheap moment to drop it.

Separately: "Test current configuration" now commits the typed password first (`runAfterProxyPasswordCommit`), so testing a candidate password without saving it is no longer possible. That is a reasonable design, but it is a behavior change and is not in the Summary.

## Ready to delete

- **`preserveSensitivePlaceholders`** (`settings-ipc-helpers.ts:73`) — this PR edits it without noticing it has zero production callers, here and on `main`; its only consumer is its own test.
- **`FileSettingsStore.testNetworkProxy()`** (`packages/storage/src/settings-store.ts:41,186`) — this PR edits its validation, and the edit is only needed because it reads a `password` field that no longer exists. Zero callers repo-wide: the interface declaration and the implementation are the only two references.
- **The dual-shape `registerRuntimeHostSettingsIpc` deps union** (`runtime-host-settings-ipc-main.ts:164`) — the legacy branch has no production caller, and the two branches produce *different lanes*, so a caller that takes the legacy one silently gets an unshared lane.

## Security — clean

No new logging of secrets; every new error string is value-free. All fixture secrets are synthetic, and the diff removes several old proxy-secret fixtures. The `settings.json` purge is temp-file plus rename, and no data is lost because `clientOwnedSettingsPatch` never forwarded `network` — the vault has been the authoritative store all along. Adding `runtime.policy.network-proxy.update` to `REMOTE_OWNER_OPERATION_GRANTS` does not widen the trust boundary, since `credential.vault.set` is already there.

Sun-GLiang:
@Astro-Han Addressed the review in 7be1502b3:

- corrected the earlier transaction-intent/recovery claim; the implementation uses ordered publication with `commit_outcome_unknown`, and now has first-cut unchanged plus later-cut known-unknown failure-injection coverage
- narrowed credentials-only schema-v1 exports to only proxy password and Tavily API key fields
- prevented window blur from committing a partial password draft
- removed the dead password override from `network-proxy.test`; the PR Summary now states that testing commits the pending password draft first
- removed `preserveSensitivePlaceholders`, `FileSettingsStore.testNetworkProxy()`, and the legacy registration-deps branch

I kept the schema-v1 compatibility change in this PR because secret removal changes where those schema-v1 credentials come from, but reduced it to credential-bearing fields so it no longer carries or reapplies unrelated settings.

M4n5ter (commented):
Reviewed at exact head `06e3dc03f45d2944e57e3b7a5ba533e344597b50` against base `6762085e2ba2372b7a8b296e08ebb67c0294dc2e`. The hosted `test`, `package`, and `windows_recovery` checks are green.

The concrete correctness findings from the previous round are resolved here: credentials-only export is narrowed to credential-bearing schema-v1 fields, window blur no longer commits a partial draft, the dead candidate-password override and dead helpers are gone, and the failure-cut coverage now matches the implemented ordered-publication / `commit_outcome_unknown` contract. I found no additional P0–P3 correctness issue in this head.

## Global-design blocker: this is still three independently reviewable changes presented as one small mask fix

The current review unit is 44 files with roughly `+1080/-239` non-test lines. The problem is not the line count; it is that the branch carries three contracts with separate failure modes, rollback boundaries, and reviewers:

1. schema-v1 import/export compatibility plus legacy settings-secret cleanup;
2. one Host policy/vault CAS authority, including preflight, publication ordering, and unknown partial outcome;
3. Renderer/Desktop removal of the sentinel plus draft, commit, cancel, reveal, and test UX.

Each is independently coherent and independently shippable. Keeping them together makes a credential representation fix also the review and rollback vehicle for a migration contract, a Host transaction contract, and a UI draft lifecycle. Please split these into three verticals.

If there is a real version-compatibility constraint that requires all three to land atomically, state that constraint explicitly and re-title/re-charter the PR as a credential-lifecycle/security migration. In that case it should be reviewed as the larger migration it is, not as a narrow `mask corruption` fix.

The test suite should follow the same boundary. Keep the invariant-bearing coverage: one offline authenticated-proxy E2E, Host two-client/CAS and durable failure cuts, draft save concurrency/failure cases, and minimal migration plus credential-only transfer cases. Move surface-local tests with their verticals rather than using one combined matrix to justify the combined PR; constant-only epoch assertions, repeated protocol-shape checks, and duplicate focus mechanics do not need to remain in this review unit.

---

> **Automated review notice:** This comment was posted by an automated review agent operated by M4n5ter. It is not an independent human review and does not replace one.

Sun-GLiang:
I have updated the PR title and charter to match the change that is actually being reviewed.

The title is now:

> `fix(desktop): make proxy credential updates explicit and Host-owned`

The PR description now presents this as a proxy-credential lifecycle migration rather than an input-mask-only fix. It separates the review into three explicit contracts:

1. legacy Settings cleanup and schema-v1 credential transfer;
2. Host-owned policy/vault CAS, publication ordering, and partial-outcome reporting;
3. Renderer-local draft, commit/cancel/reveal behavior, and proxy-only Copy removal.

I have not split the branch at this point. These contracts are independently testable, but the current implementation delivers them as one compatibility boundary. Splitting them would require a temporary dual-shape compatibility layer or restoring part of the unsafe representation during the intermediate merges:

- Removing `password` from ordinary Settings requires both the read-only `passwordConfigured` projection and a write-only credential mutation.
- The Renderer draft depends on that explicit Host mutation; otherwise an intermediate version must retain either the sentinel or the previous split-write path.
- The Host operation must validate the policy revision and credential generation in one lane so disabling authentication wins over an older replacement, as required by #3696.
- Once the Settings shape changes, schema-v1 transfer needs an adapter and persisted legacy Settings need cleanup so compatibility paths cannot retain or reintroduce the removed secret field.
- The Runtime Host epoch moves with this boundary because an older peer can split policy and credential writes against different bases.

The revised description does not claim an all-or-nothing storage transaction. It states the implemented contract directly: shared preflight, ordered publication, and `commit_outcome_unknown` when a durable failure occurs after only part of the compound operation has committed.

The verification section is now organized around the invariant-bearing coverage for each review boundary: minimal migration and credential-only transfer cases, Host two-client/CAS and durable failure cuts, Renderer draft concurrency/failure cases, and one offline authenticated-proxy E2E. I have not moved or pruned tests while the review-unit decision remains open.

Please re-evaluate the review unit under the revised title and charter. If maintainers still require a physical split after considering these compatibility dependencies, the fallback will be ordered stacked PRs rather than treating the three contracts as unrelated changes.

apps/desktop/src/main/runtime-host-config-ipc-main.ts
```diff
@@ -150,11 +189,16 @@ export async function gatherRuntimeHostConfig(
...
+  // Schema v1 stores the network-proxy password and Tavily key in the
+  // settings payload. Keep a credentials-only request lossless by making the
+  // dependency explicit in the generated bundle.
   if (selected.has('settings')) {
     const settings = await deps.getSettings();
     data.settings = selected.has('credentials')
       ? restoreHostSettingsSecrets(settings, secrets)
       : stripSettingsSecretsForExport(settings);
+  } else if (selected.has('credentials')) {
+    data.settings = projectHostSettingsSecrets(secrets);
```
Astro-Han: Thanks for making the proxy/Tavily dependency explicit for credentials-only export. I think the same [P1] category ① path is still lossy for model credentials: when only `credentials` is selected, this branch adds settings secrets but still omits `data.connections`. `applyConfigImport()` only adds slugs to `appliedConnectionSlugs` for connections created or overwritten by this import, so a target that already has the matching connection skips every exported model credential. The added test uses an empty connection catalog and cannot exercise that normal restore. I suggest allowing a credentials-only bundle to bind to an already-existing exact connection identity while preserving the explicit “skip connection” protection, then adding an existing-target-connection restore test. Please push back if model credentials are intentionally excluded from the lossless claim.

Astro-Han (commented):
Thanks for moving proxy credential mutation into the Host-owned operation and for making the transfer dependency explicit. The earlier cross-client ordering findings look addressed on this head. I found two normal credentials-only restore paths that still look destructive or incomplete.

These are suggestions from an outside review. Please push back if credentials-only import is intentionally a replacement snapshot with different semantics than the PR description; documenting that contract would change the assessment.

感谢把 proxy credential mutation 收敛到 Host owner，并显式处理 transfer dependency。旧的跨客户端顺序问题已基本关闭。当前仍有 2 个 P1：credentials-only bundle 不含 connection definitions，导入器会跳过所有 model credentials；source 缺少 proxy/Tavily secret 时又会导出空串，目标侧把它解释成 delete，删除目标已有凭据。若 credentials-only 本来就是 destructive snapshot，请补充契约说明并欢迎 push back。

</details>

AI-assisted review disclosure: Codex coordinated an independent reviewer lane; Astro-Han independently checked the exact head, transfer composition, existing threads, and severity, and owns this review.

apps/desktop/src/main/runtime-host-config-ipc-main.ts
```diff
@@ -362,6 +412,98 @@ function restoreHostSettingsSecrets(
...
 }
 
+function projectHostSettingsSecrets(
+  secrets: ReadonlyMap<string, string>,
+): Record<string, unknown> {
+  return {
+    network: {
+      proxy: {
+        password:
+          secrets.get(locatorKey({ scope: 'network_proxy', kind: 'password' })) ?? '',
```
Astro-Han: Thanks for handling schema-v1 settings-carried secrets. Using `""` for “the source has no credential” creates a [P1] category ① destructive restore: the importer maps an empty proxy password to `credential: delete`, and the settings writer likewise deletes Tavily when `apiKey.length === 0`. A credentials-only export from a source with no proxy/Tavily secret can therefore erase valid secrets already present on the target. Generic connection credentials use omission/skip for absence, so these categories currently disagree. I suggest encoding absent, replace, and explicit delete separately (or clearly defining and testing destructive snapshot semantics) and adding source-absent/target-present regressions. Please push back if deletion is the intended public contract.

apps/desktop/src/main/runtime-host-config-ipc-main.ts
Sun-GLiang: Fixed in 7e14d8934. A credentials-only import now binds credential entries to an already-existing connection with the same slug when the bundle manifest omits `connections`. If the bundle does carry a connection snapshot, credential writes remain limited to connections created or overwritten by that import, so an explicit skip still preserves the target credential. Added an existing-target restore regression and made the credentials-only export test exercise a non-empty catalog.

apps/desktop/src/main/runtime-host-config-ipc-main.ts
Sun-GLiang: Fixed in 7e14d8934. Credentials-only export now omits each absent settings-carried secret instead of encoding absence as an empty string; when both proxy and Tavily secrets are absent, no settings compatibility payload is emitted. Added proxy-only, Tavily-only, and both-absent regressions. Full settings+credentials snapshot behavior remains unchanged.

apps/desktop/src/main/config-transfer-service.ts
```diff
@@ -91,13 +91,25 @@ export async function applyConfigImport(
...
     };
+  } else if (
+    !bundle.includedData.includes('connections') &&
+    Array.isArray(bundle.data.credentials)
+  ) {
+    // Without a connection snapshot, the credential slug names an existing
+    // logical connection directly. A bundle that does include connections
+    // still uses the create/overwrite set above so an explicit skip cannot
+    // overwrite the target's credential.
+    const existing = await deps.connectionStore.list();
```
Astro-Han, line 118: [P1] (category ① — normal credentials restore path)

Thanks for restoring credentials-only bundles against existing connections. The source entry still carries only `{slug, kind, value}`, while this branch treats every target connection with the same slug as eligible. A normal restore between profiles can therefore take an API key/request-header secret for source slug `team` and write it into a target `team` whose provider or custom `baseUrl` is different; the next model request sends that secret to the target endpoint. This is the remaining boundary from the earlier connection-identity comment: slug plus credential kind is not an endpoint identity. Could the exported entry include a minimal verifiable binding (at least provider type plus canonical effective endpoint/fingerprint), skip/report mismatches, and cover same-slug/different-provider and same-provider/different-baseUrl restores? Please feel free to push back if another authenticated binding is preserved outside this payload.

apps/desktop/src/main/runtime-host-config-ipc-main.ts
```diff
@@ -150,11 +189,17 @@ export async function gatherRuntimeHostConfig(
...
+  // Schema v1 stores the network-proxy password and Tavily key in the
+  // settings payload. Keep a credentials-only request lossless by making the
+  // dependency explicit in the generated bundle.
   if (selected.has('settings')) {
     const settings = await deps.getSettings();
     data.settings = selected.has('credentials')
       ? restoreHostSettingsSecrets(settings, secrets)
       : stripSettingsSecretsForExport(settings);
+  } else if (selected.has('credentials')) {
+    const settingsSecrets = projectHostSettingsSecrets(secrets);
```
Astro-Han: [P2] (category ① — normal credentials restore path)

Thanks for keeping credentials-only export lossless. A proxy password by itself currently produces `{network.proxy.password}` without `authEnabled`; on a fresh/default target the imported patch is merged with `authEnabled: false`, then adapted to `credential: replace`, which the policy/credential operation rejects. That makes a normal credentials-only round trip fail and also prevents later Tavily/model credentials in the bundle from being applied. Could the bundle carry the minimum proxy-auth policy needed for this secret, or skip/report the incompatible secret without aborting unrelated credentials, with an export→default-target import regression? Please push back if credentials-only restore intentionally requires preconfiguring identical proxy policy on the target.

Astro-Han (commented):
Thanks for moving proxy and connection secrets behind Host-owned mutation. The latest changes close the missing-secret deletion path, but two credentials-only restore boundaries still look reachable. These are suggestions from an outside review, so please do push back if the bundle has an identity or policy invariant that is not visible here.

AI-assisted review disclosure: Codex ran independent credential-security, Runtime Host, and production/test analysis lanes; Astro-Han is the contributor of record for this review.

apps/desktop/src/main/runtime-host-config-ipc-main.ts
Sun-GLiang: Fixed in a0ea94647. When a credentials-only export contains a proxy password, its schema-v1 compatibility payload now also carries the minimum required proxy-auth policy (`authEnabled: true`). The import adapter therefore produces a valid replace operation against a fresh/default target instead of aborting the remaining credential import. Added an export-to-default-target regression; absent proxy credentials remain omitted and non-destructive.

apps/desktop/src/main/config-transfer-service.ts
Sun-GLiang: Fixed in a0ea94647. Exported connection credentials now carry a binding with the source provider type and effective base URL. A credentials-only import writes only when the existing target with that slug matches both values; same-slug/different-provider and same-provider/different-endpoint targets are skipped and reported. Connection-snapshot imports retain the existing create/overwrite boundary. Added exact-match plus both mismatch regressions.

apps/desktop/src/main/runtime-host-config-ipc-main.ts
```diff
@@ -362,6 +437,106 @@ function restoreHostSettingsSecrets(
...
+    locatorKey({ scope: 'web_search', provider: 'tavily', kind: 'api_key' }),
+  );
+  if (proxy === undefined && tavily === undefined) return undefined;
+
+  return {
+    ...(proxy === undefined
+      ? {}
+      : {
+          network: {
+            proxy: { authEnabled: true, password: proxy },
```
Astro-Han: Thanks for making credentials-only proxy restore valid on a default target. [P1 · category ①] The exported password still has no binding to the source proxy destination: this payload carries only authEnabled and password, not protocol/host/port/username. Importing it into a profile already configured for a different proxy therefore keeps that target endpoint and username while replacing its password; the next proxied request can send the source proxy secret to a different server. Could the bundle carry and verify the minimum proxy target/auth identity, skipping and reporting mismatches, or restore the full necessary proxy policy explicitly? A source proxy A → target proxy B regression would pin the intended contract.

Astro-Han (commented):
Thanks for addressing the earlier credentials-only restore findings. The connection credential now carries a provider/endpoint binding, absent proxy and Tavily secrets are non-destructive, and the default-target proxy case is covered.

I found two remaining P1 authority issues, noted inline. Both can route a credential to a different destination than the one the source bundle described: the proxy password has no source proxy binding, and the connection binding check is separated from the Host credential mutation across supported clients. These are suggestions from an outside review; please push back if an existing Host transaction or credential binding closes either interleaving. I am leaving a Review Comment rather than Request Changes.

Exact-head test, package, and windows_recovery checks are green.

Review analysis was assisted by Codex and an independent reviewer agent. Astro-Han verified the exact-head diff, owner boundaries, concurrent production paths, and severity before publication and owns this review.

apps/desktop/src/main/config-transfer-service.ts
```diff
@@ -116,10 +134,14 @@ export async function applyConfigImport(
...
-      // stored secret untouched.
-      if (!appliedConnectionSlugs.has(entry.slug)) {
+      // Unknown targets and connections explicitly skipped by a connection
+      // snapshot keep their existing stored secret untouched.
+      const target = credentialTargets.get(entry.slug);
+      if (!credentialTargets.has(entry.slug)) {
+        skipped += 1;
+        continue;
+      }
+      if (target && !matchesCredentialConnection(entry.connection, target)) {
```
Astro-Han: Thanks for checking provider and effective endpoint here. [P1 · category ②] This validation still happens outside the Host credential mutation authority. Another supported client can change the same connectionId/slug after this snapshot but before saveConnectionCredential reloads by slug and writes the vault; the writer never receives or revalidates this expected binding/revision. The credential can therefore be committed to the newly retargeted connection, and export has the symmetric catalog-read → secret-read window. Could Host own a bound export/write operation that validates connection revision, provider, canonical endpoint, and credential basis in its mutation lane? A deterministic two-client target-change cut would cover the gap.

apps/desktop/src/main/runtime-host-config-ipc-main.ts
Sun-GLiang: Fixed in 6c6644b72. Credentials-only proxy export now carries the minimum Host-read target identity (protocol, normalized host, port, and username). Import passes that identity through the existing compound runtime.policy.network-proxy.update operation; Storage compares it with the current proxy inside the same lane and returns proxy_target_mismatch before either policy or vault publication. Desktop skips and reports that credential instead of changing proxy policy. Added source proxy A to target proxy B coverage and a proxy export race test.

apps/desktop/src/main/config-transfer-service.ts
Sun-GLiang: Fixed in 6c6644b72. The existing configuration credential export and credential.vault.set paths now accept an exact connection target basis. Storage validates connection revision, slug, provider, canonical effective endpoint, and the existing credential CAS basis in its serialized lane before reading or writing secret material. Desktop retries a stale bound export and skips/reports a stale import. Added deterministic Storage and two-Client Runtime Host retarget tests.

Sun-GLiang: Follow-up in 463eb56df closes the compatibility fallback introduced by 6c6644b72: when no Connection snapshot is present, an imported credential must carry entry.connection. Desktop no longer synthesizes that source binding from the target Connection; a missing binding is skipped and reported. Snapshot-backed imports retain their existing create/overwrite compatibility path. Added a regression that fails on 6c6644b72 and verifies no credential write occurs.

apps/desktop/src/main/runtime-host-config-ipc-main.ts
Sun-GLiang: Follow-up in 463eb56df closes the remaining legacy path: a non-empty schema-v1 proxy password without credentialTarget is rejected during import preflight, so adaptRuntimeHostConfigImport cannot construct an unbound replace. Bound imports still normalize the target and use the existing compound Host operation. Added a regression that fails on 6c6644b72.

me2seeks (approved):
Approving. This is a careful lifecycle migration, not an input-mask patch — the deletion proof holds at every layer I traced.

Verified end to end:

- **The secret has exactly one home.** Persisted `AppSettings` drops `password`; `mergeSettings` strips `credential`/legacy `password`/derived `passwordConfigured` from any incoming patch, and `validateProxyPatch` + the Host-side decode bound the secret size. Reads expose only `passwordConfigured`. The old `SENSITIVE_PLACEHOLDER` round-trip machinery is fully removed.
- **The compound write is genuinely one CAS.** `runtime.policy.network-proxy.update` carries expected policy revision + expected credential basis (+ optional expectedTarget), preflights both documents before publishing either, and orders the two commits correctly: credential-before-policy when enabling/replacing, policy-before-credential when deleting — so an enabled policy never points at an absent credential on the happy path, and partial durable failure surfaces as explicit `commit_outcome_unknown` rather than a silent half-write.
- **Legacy purge is safe and visible.** `settings.json` is rewritten on read without the plaintext secret (atomic temp+rename), config import rejects unbound legacy passwords, and the new `proxy_credential_missing` code gives the user a concrete signal when auth is enabled but the vault is empty.
- **Renderer draft discipline is right.** Replacement text stays renderer-local until blur/Enter, Escape cancels, the Eye only reveals the draft, copy action is disabled on the field, IME composition doesn't trigger Enter-commit, and Test Proxy commits the draft *before* testing so the probe always runs against the vault state.

Non-blocking observations (P3, no action required — flagging intent):

1. **Legacy passwords are purged, not migrated.** Users with a saved proxy password will find auth silently unconfigured after upgrade until the next test/run reports `proxy_credential_missing`. Given the plaintext-at-rest concern this is defensible, but it deserves a release note so the first auth failure isn't a surprise.
2. **Desktop never sends `expectedTarget`** on UI-initiated replaces, so a password commit racing a proxy host/username change can bind the secret to the old target. The end state is self-describing (`passwordConfigured: false` for the new target) and recoverable by re-entering; only imports exercise the mismatch path today.
3. The Host accepts `delete` while the new policy still requires a credential (it just reorders the commits). Desktop's validation prevents producing this, and the resulting state is visible via `proxy_credential_missing` — a Host-side reject would be defense in depth only.

PR #3707, opened 2026-08-24
refactor(windows): reuse packaged verifier for rollback

### LINKED ISSUE #3575: Simplify Windows installer rollback without weakening recovery guarantees
## Context

#3265 added pre-upgrade backup, automatic NSIS `.onInstFailed` rollback, retained-backup recovery after hookless `Quit`, and a Windows behavioral gate. It is a real protection, not 1,000+ lines built only around an empty-directory check.

In electron-builder 26.15.3's one-click template, `uninstallOldVersion` destroys the previous installation before several built-in NSIS `File` instructions run. NSIS documents that when a `File` cannot be extracted/created/overwritten and the user cancels, `.onInstFailed` runs. This is a real post-destruction production entry into Maka's rollback hook. The template's 7z retry/cancel branch is different: it calls `Quit`, so no callback runs and the retained sibling backup is the only recovery evidence.

**Non-goal: do not remove the backup or automatic `.onInstFailed` rollback.** A manifest or registry record cannot recreate the deleted application bytes. The Windows gate simulates the callback with an explicit `Abort`; it does not currently force a real disk-full/permission `File` failure.

This issue tracks the simplifications that remain after preserving that safety boundary. The four areas below are independent unless a dependency is called out.

## 1. Empty-directory enumeration (21 lines)

`makaDirHasEntries` uses `FindFirst`/`FindNext` to distinguish a truly empty directory from the `.` and `..` entries matched by `${FileExists} "$INSTDIR\*.*"`. The earlier wildcard form was incorrect.

This helper has one consumer: the restore branch where moving the partial/new `$INSTDIR` aside still fails after retries. It decides whether to copy the verified old backup into a held-but-empty directory.

- **Safe conclusion while the current behavior remains:** keep the helper. NSIS has no equivalent reliable wildcard emptiness predicate.
- **Dependency:** it can be deleted only with item 2's decision to remove empty-shell auto-salvage.
- **Validation if kept:** add a deterministic Windows fault-injection test for a held empty `$INSTDIR`; the current rollback gate does not execute this branch.
- **Line effect:** 21 NSIS lines, but not independently deletable without changing item 2's behavior.

## 2. Aside/copy salvage after restore conflicts

Today rollback first renames the partial/new tree to `<install>.failed-upgrade`, then renames the verified backup into place. If the first rename fails, it distinguishes populated from empty targets; an empty shell gets a copy fallback, while a populated target returns 103 and retains both trees/recovery evidence.

Proposed simplification: after rename retries are exhausted, always take the existing fail-safe outcome—return 103, retain the verified sibling backup and recovery note, and require manual recovery. This removes the empty-shell classification and copy fallback without risking the backup.

- **Why the weaker branch is still safe:** the verified backup has not been moved or modified when step 1 fails. The deleted logic improves automatic recovery only for a held-but-empty target; it is not what preserves the previous bytes.
- **Behavior change requiring maintainer approval:** a rare held empty shell changes from automatic recovery to exit 103 + manual recovery.
- **Estimated net deletion:** 45–55 NSIS lines including item 1.
- **Acceptance:** exact-head Windows rollback/control/Quit gates stay green; docs and exit-103 wording say that a verified sibling backup may require manual recovery; either add a Windows handle-conflict test or explicitly remove the unexecuted auto-salvage promise.

## 3. The 11-value registry snapshot and rollback verification

The snapshot is on the real `.onInstFailed` path. It mirrors registry values electron-builder's old uninstaller deletes:

- install key: `InstallLocation`, `KeepShortcuts`, `ShortcutName`, `MenuDirectory`;
- uninstall key: `DisplayName`, `UninstallString`, `QuietUninstallString`, `DisplayVersion`, `DisplayIcon`, `Publisher`, `Comments`;
- `NoModify` and `NoRepair` are deterministic template constants and are recreated rather than snapshotted.

**Safe conclusion under the current contract: do not remove or replace these values with a manifest.** Automatic rollback restores old program bytes; without corresponding old registration, Settings uninstall, quiet uninstall, version identity, links, and updater/installer behavior can describe a different installation. The gate directly verifies restored `DisplayVersion`, registration, launch, and later uninstall, while the implementation restores the remaining template-owned values exactly.

There are two possible contract changes, each requiring a maintainer decision before implementation:

1. Define a smaller documented registration contract and name which old values may be lost or recomputed. Do not infer that cosmetic-looking fields are unused.
2. On registry write-back/read-back failure, stop rebuilding another full sibling backup for a later self-healing rerun. Keep the already-restored old tree, persisted snapshot, aside tree, recovery note, and exit 103 for manual repair. This second option removes approximately 60–75 lines across NSIS, verifier, and docs, but narrows the current “may recover on rerun” behavior for a rollback-of-the-rollback failure.

Acceptance for either change must exercise real Windows registry behavior and still prove that the restored installation is registered, launchable at the old version, and uninstallable. Source inspection alone is insufficient.

## 4. The 637-line rollback verifier

Do **not** merge the whole verifier into `verify-windows-installer-lifecycle.mjs`. The files share mechanics but own different contracts:

- lifecycle: install → optional upgrade → smoke → uninstall;
- rollback: nonzero exit codes and abnormal filesystem/registry states (Abort restore, registry mismatch, stale/incomplete backup, hookless Quit, rerun adoption, registrationless refusal, cleanup).

The rollback verifier already imports shared manifest, CDP, process, cleanup, and lifecycle primitives. `verify-windows-harness.test.mjs` unit-tests those helpers; `product-release.test.mjs` only pins workflow ordering. Neither duplicates the behavioral gate. Folding the files would retain almost all scenarios while creating an approximately 1,000-line modeful lifecycle authority, saving only setup/cleanup boilerplate and making failures harder to locate.

One high-confidence simplification is available: replace rollback's bespoke `assertLaunchable` CDP wrapper with `verifyPackagedWindowsApp(installDirectory, { workingDirectory, expectedVersion, artifactContract: 'legacy-baseline' })`, the authority already used by the lifecycle verifier for an installed previous release.

- **Why safe:** it validates the same installed-directory artifact, launches it, waits for a usable renderer, and verifies the expected product version; no rollback-specific rule lives in the local wrapper.
- **Estimated net deletion:** 35–45 lines.
- **Acceptance:** harness tests and the full Windows rollback gate stay green, including old-version launch after rollback.

## Optional product-level simplification

The largest remaining state machine exists to make a later installer safely adopt a backup left by `Quit`. If maintainers change the support contract to “retained backup + README, manual recovery only,” persistent snapshot/adoption/identity conflicts and their scenarios can be removed while keeping same-process automatic rollback. Estimated deletion: another 150–190 lines across installer, verifier, and bilingual docs.

This is not cleanup: it removes the documented supported rerun recovery path. `Quit` and hard-kill paths still cannot delete the backup because no callback runs.

## Expected reduction

- No behavior change: approximately **35–45 lines**.
- Accept manual recovery for the two rollback-of-the-rollback contingencies in items 2 and 3: another **105–130 lines**.
- Also replace supported `Quit` rerun recovery with manual-only recovery: another **150–190 lines**.
- Maximum identified reduction: approximately **290–365 lines**. There is no evidence-backed way to remove roughly 1,000 lines while retaining #3265's current guarantees.

## Contribution prerequisites

The safe verifier reuse is self-contained. The installer changes are not suitable for an unassisted first contribution: they require NSIS callback/error semantics, access to a real Windows environment, the ability to build two installer versions, and fault-injection validation. More importantly, items 2 and 3 require maintainers—not contributors—to choose the recovery contract first. For that reason this issue should not receive `help wanted` until those decisions are made and a single accepted behavior is written into the acceptance criteria.

## References

- Original implementation: #3265
- NSIS callback and file-failure semantics: <https://nsis.sourceforge.io/Docs/Chapter4.html>
- Implementation: `apps/desktop/build/installer.nsh`
- Behavioral gate: `scripts/verify-windows-installer-rollback.mjs`
- Existing lifecycle authority: `scripts/verify-windows-installer-lifecycle.mjs`

testikun:
## Summary

Replace the rollback verifier's private `assertLaunchable` CDP wrapper with the existing `smokePackagedRenderer` helper and the shared Windows product-version assertion.

The rollback path now checks only the restored candidate's launchability and version after its filesystem and uninstall-registration state have been restored. It does not rerun the full packaged-app contract (sandbox worker, stdio relay, node-pty, and dependency checks); those checks remain owned by the release verification step earlier in the same package job.

The rollback-specific filesystem, registry, backup, recovery, fail-closed, and `.onInstFailed` assertions are unchanged. The duplicate process/renderer/version plumbing is removed, and a focused harness test pins the narrow delegation contract.

Fixes #3575

## Verification

- `node --test scripts/verify-windows-harness.test.mjs` — 42 passed
- `npm run lint` — passed
- `npx biome check scripts/verify-windows-installer-rollback.mjs scripts/verify-windows-harness.test.mjs` — passed
- `node --check` for both changed scripts and `git diff --check` — passed
- The real Windows installer rollback gate was not run locally because this workspace is macOS; it remains the authoritative Windows CI check.
- The broader product-release test set requires built workspace artifacts in this fresh worktree; the focused rollback harness is green.

### 中文摘要

本 PR 复用已有的 renderer smoke 和 Windows 产品版本校验来验证回滚后的候选安装，不再重复执行完整 packaged-app contract；文件树、注册表、备份、恢复、fail-closed 和 `.onInstFailed` 行为保持不变，并补充了有界委托回归测试。

## AI use

Tool: OpenAI Codex assisted with issue analysis, implementation, tests, and verification.

### DIFF
--- a/scripts/verify-windows-installer-rollback.mjs
+++ b/scripts/verify-windows-installer-rollback.mjs
@@ -175,7 +175,10 @@ export async function verifyRestoredWindowsInstallation(
   { verifyApp = verifyPackagedWindowsApp, run } = {},
 ) {
   const options = {
-    artifactContract: 'legacy-baseline',
+    // The rollback verifier restores the candidate installer, not the pinned
+    // upgrade baseline. The candidate is built from the current tree and must
+    // therefore satisfy the current packaged-artifact contract.
+    artifactContract: 'current',
     expectedVersion,
     workingDirectory,
   };
--- a/scripts/verify-windows-installer-rollback.mjs
+++ b/scripts/verify-windows-installer-rollback.mjs
@@ -22,18 +22,19 @@ import { access, cp, mkdir, mkdtemp, readdir, rm, writeFile } from 'node:fs/prom
 import { tmpdir } from 'node:os';
 import { join, resolve } from 'node:path';
 import { pathToFileURL } from 'node:url';
-import { diffTreeManifests, directoryTreeManifest, runCommand } from './verify-packaged-app.mjs';
+import {
+  diffTreeManifests,
+  directoryTreeManifest,
+  runCommand,
+  smokePackagedRenderer,
+} from './verify-packaged-app.mjs';
 import {
   installerVersion,
   waitForInstalledProcessesToExit,
   waitForUninstallRegistrationToClear,
   waitUntilMissing,
 } from './verify-windows-installer-lifecycle.mjs';
-import {
-  assertWindowsProductVersion,
-  powerShellLiteral,
-  verifyPackagedWindowsApp,
-} from './verify-windows-x64.mjs';
+import { assertWindowsProductVersion, powerShellLiteral } from './verify-windows-x64.mjs';
 
 const uninstallExecutableName = 'Uninstall Maka.exe';
 const executableName = 'Maka.exe';
@@ -172,19 +173,17 @@ export async function verifyRestoredWindowsInstallation(
   installDirectory,
   workingDirectory,
   expectedVersion,
-  { verifyApp = verifyPackagedWindowsApp, run } = {},
+  { smokeRenderer = smokePackagedRenderer, run } = {},
 ) {
   await mkdir(workingDirectory, { recursive: true });
-  const options = {
-    // The rollback verifier restores the candidate installer, not the pinned
-    // upgrade baseline. The candidate is built from the current tree and must
-    // therefore satisfy the current packaged-artifact contract.
-    artifactContract: 'current',
-    expectedVersion,
-    workingDirectory,
-  };
-  if (run) options.run = run;
-  await verifyApp(installDirectory, options);
+  const installedExecutable = join(installDirectory, executableName);
+  // The caller has already proved the restored tree byte-identical. Keep this
+  // gate focused on rollback-specific evidence: the restored candidate still
+  // launches and reports the expected version, while the package step owns the
+  // full artifact/sandbox contract.
+  await smokeRenderer(installedExecutable, { workingDirectory });
+  const actualVersion = await readInstalledProductVersion(installedExecutable, { run });
+  assertWindowsProductVersion(actualVersion, expectedVersion);
 }
 
 /**
@@ -215,7 +214,7 @@ export async function verifyWindowsInstallerRollback(
     platform = process.platform,
     makeTemporaryDirectory = () => mkdtemp(join(tmpdir(), 'maka-rollback-')),
     run = runCommand,
-    verifyApp = verifyPackagedWindowsApp,
+    smokeRenderer = smokePackagedRenderer,
   } = {},
 ) {
   if (platform !== 'win32') {
@@ -344,7 +343,7 @@ export async function verifyWindowsInstallerRollback(
       installDirectory,
       join(temporaryDirectory, 'restored-smoke'),
       candidateVersion,
-      { run, verifyApp },
+      { run, smokeRenderer },
     );
     await waitForInstalledProcessesToExit(installDirectory);
 
--- a/scripts/verify-windows-installer-rollback.mjs
+++ b/scripts/verify-windows-installer-rollback.mjs
@@ -18,27 +18,22 @@
  */
 
 import { spawn } from 'node:child_process';
-import { access, cp, mkdir, mkdtemp, readdir, rm, writeFile } from 'node:fs/promises';
+import { access, cp, mkdtemp, readdir, rm, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join, resolve } from 'node:path';
 import { pathToFileURL } from 'node:url';
-import {
-  diffTreeManifests,
-  directoryTreeManifest,
-  findRendererTarget,
-  isolatedUserEnv,
-  waitForDevToolsPort,
-  waitForUsableRenderer,
-  runCommand,
-  stopChild,
-} from './verify-packaged-app.mjs';
+import { diffTreeManifests, directoryTreeManifest, runCommand } from './verify-packaged-app.mjs';
 import {
   installerVersion,
   waitForInstalledProcessesToExit,
   waitForUninstallRegistrationToClear,
   waitUntilMissing,
 } from './verify-windows-installer-lifecycle.mjs';
-import { assertWindowsProductVersion, powerShellLiteral } from './verify-windows-x64.mjs';
+import {
+  assertWindowsProductVersion,
+  powerShellLiteral,
+  verifyPackagedWindowsApp,
+} from './verify-windows-x64.mjs';
 
 const uninstallExecutableName = 'Uninstall Maka.exe';
 const executableName = 'Maka.exe';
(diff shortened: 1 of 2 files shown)


Astro-Han:
Looks good, CI is running.

testikun:
Follow-up fix pushed in `89234bcc4`.

The first Windows failure was caused by the rollback path passing the current candidate installation to `verifyPackagedWindowsApp` with the `legacy-baseline` contract, which required the retired `bundled-git.json`. That is now `artifactContract: current`.

The next Windows run exposed a second integration detail from the verifier reuse: the old rollback-only launcher created the smoke working directory implicitly, while the shared verifier writes `packaged-sandbox-manifest.json` directly into it. The rollback helper now creates that directory explicitly before delegation.

Local verification: `node --test scripts/verify-windows-harness.test.mjs`, `npm run format:check`, and `npm run lint` pass.

回滚验证现在使用 current 契约，并在调用共享 verifier 前显式创建 smoke 目录，修复了 Git 资源误检和 `ENOENT packaged-sandbox-manifest.json` 两个 CI 失败。
</details>

Generated-by: OpenAI Codex

scripts/verify-windows-installer-rollback.mjs
```diff
@@ -173,43 +168,23 @@ $entries | ForEach-Object { Remove-Item -LiteralPath $_.Path -Recurse -Force }
...
+  workingDirectory,
+  expectedVersion,
+  { verifyApp = verifyPackagedWindowsApp, run } = {},
+) {
+  await mkdir(workingDirectory, { recursive: true });
+  const options = {
+    // The rollback verifier restores the candidate installer, not the pinned
+    // upgrade baseline. The candidate is built from the current tree and must
+    // therefore satisfy the current packaged-artifact contract.
+    artifactContract: 'current',
```
Astro-Han: **P2 — the delegation widens the rollback gate with work the verifier already proved redundant three steps earlier.**

`verifyPackagedWindowsApp` with `artifactContract: 'current'` is not a launch check. It runs `assertPackagedResources`, `assertPackagedDependencyClosure`, the PE machine check, the sandbox broker `--self-probe`, `experiments/windows-sandbox/stdio-relay-smoke.ps1` through pwsh, `verifyWindowsSandboxWorkerE2E`, the node-pty conpty smoke, and `smokePackagedRenderer` with `verifyMaximizeRestore: true`. The deleted `assertLaunchable` did one of those.

All of the added work is provably redundant here. Three steps above this call, `verifyWindowsInstallerRollback` asserts the restored tree is **byte-identical** to the pre-upgrade manifest (`0 differing files across N entries`) and that no residue remains. Those exact bytes were already put through the full current contract by the "Verify the Windows release" step earlier in the same `package` job. Re-verifying the sandbox worker E2E on bytes just proven identical adds no evidence about rollback.

The cost is small but real — the rollback step went from 133s and 147s on two recent `main`-branch runs to 168s here — and the more important cost is failure attribution: a regression in the packaged sandbox, the stdio relay, or node-pty now fails the *rollback* gate. That is the specific argument #3575 item 4 used to reject folding the two verifiers together, and it applies to this narrower reuse too.

The smaller version reuses the same authority without widening anything:

```js
await smokePackagedRenderer(installedExecutable, { workingDirectory });
assertWindowsProductVersion(
  await readInstalledProductVersion(installedExecutable, { run }),
  candidateVersion,
);
```

`smokePackagedRenderer` is what `assertLaunchable` was a copy of — same `--remote-debugging-port=0` and `--user-data-dir` isolation, same `isolatedUserEnv`, same 16 KiB stderr capture appended to the failure — and this file already imports from `verify-packaged-app.mjs`. `readInstalledProductVersion` uses the identical PowerShell expression `verifyPackagedWindowsApp` uses for the version resource, and it plus `assertWindowsProductVersion` are already called eight times in this file. That deletes the same duplicate plumbing #3575 targeted, keeps the gate's meaning unchanged, and needs one added import.

Not blocking — what is here is green and correct. But #3575 scoped this as "no behavior change", and as written it is a behavior change to a release gate.

---

Astro-Han (commented):
Reviewed the production change. One P2 inline and one P3 below; no P0 or P1, and the Windows `package` gate is green on `89234bcc4`, so nothing here blocks merge.

Removing `assertLaunchable` is right. I compared it line by line against `smokePackagedRenderer` in `verify-packaged-app.mjs`: same `--remote-debugging-port=0` and `--user-data-dir` isolation, same `isolatedUserEnv` home/APPDATA/LOCALAPPDATA setup, same 16 KiB rolling stderr capture appended to the failure message — the shared helper actually wraps a slightly wider range in its `catch`. It was a copy, and the copy should go.

The `legacy-baseline` → `current` correction in `89234bcc4` is also right, and it is worth recording that #3575 got this wrong: the restored installation is the candidate build, and `legacy-baseline` demands the retired `bundled-git.json`.

**P3 — the PR body no longer describes the diff.** It still says the change uses "the explicit `legacy-baseline` contract for restored pre-upgrade installations." The code uses `artifactContract: 'current'`, and the comment above it states the opposite of the body — the restored installation is the candidate, not the pinned baseline. The follow-up comment explains the change, but the body was never updated, so anyone merging from the description alone is approving a claim the diff does not make. Worth fixing before merge. A short correction on #3575's item 4 acceptance text would also stop the next reader from repeating the `legacy-baseline` suggestion.

AI use: drafted with Claude Code (Opus 5); I verified every claim in it against the source at `89234bcc4` and the CI run before posting, and did not run the Windows gate locally.

scripts/verify-windows-installer-rollback.mjs
testikun: Addressed in `540f2f578`, rebased onto the latest `apache/main` (`45ca7c0c9`).

The rollback path now reuses `smokePackagedRenderer` plus the shared Windows product-version assertion only. It no longer invokes the full `verifyPackagedWindowsApp` contract, so sandbox/stdio/node-pty/dependency checks remain owned by the earlier package verification step. The rollback-specific byte-identical tree, registry, backup, recovery, and fail-closed checks are unchanged.

I also updated the PR body to remove the stale `legacy-baseline` claim and document the narrowed boundary. The focused harness suite passes 42/42 and lint/format/script checks pass locally; the Windows installer gate remains CI-authoritative because this workspace is macOS.

简体中文：已在 `540f2f578` 修复并基于最新主干重放。回滚流程现在只复用 renderer smoke 和版本校验，不再重复完整 packaged-app 检查；PR 描述也已同步更新。

testikun: Follow-up: I amended the same change to `7f548a541` to restore the real `Generated-by: OpenAI Codex` commit trailer required by the repository. The code and tests are unchanged.

Astro-Han (approved):
Verified the fix in `7f548a541`. The delegation is now `smokePackagedRenderer` plus `readInstalledProductVersion` / `assertWindowsProductVersion`, the `verifyPackagedWindowsApp` import is gone with no `verifyApp` left anywhere in either file, and the harness test pins the narrow boundary down to the exact PowerShell invocation. The rewritten body now states what the rollback gate does not rerun, which resolves the P3 as well. Windows `package` is green on this head.

AI use: reviewed with Claude Code (Opus 5); I verified the diff and CI state myself before approving.

PR #2476, opened 2026-08-08
test(cli): run ordinary run-command semantics in process

UncertaintyDeterminesYou4ndMe:
Part of #2387 (CLI workspace; the Headless workspace follows in a separate PR per the issue).

## Problem

`run-command.test.ts` paid one Node subprocess startup per assertion — 29 spawns across 27 tests — including for scenarios that only exercise argument handling, session selection, and outcome-to-exit-code mapping. Those semantics flow entirely through the injectable `MakaRunDeps` seam (`createContext`, `listSessions`, `stdinIsTTY`, `readStdin`, `writeStdout`, `writeStderr`), so the subprocess added cost without adding a distinct failure signal.

## Change

- **One fake, two routes.** The scenario fake moves out of `run-command-fixture.ts` into `run-command-fake.ts`, parameterized by an options object instead of `MAKA_RUN_*` environment variables. The fixture becomes a thin subprocess wrapper over the same fake, so the in-process and subprocess routes cannot drift apart.
- **21 ordinary-semantics tests move in process** through `runMakaTextCli` with injected stdin/stdout/stderr. Assertions are unchanged.
- **No production changes.**

## Retained process contracts (real subprocess, unchanged assertions)

| Contract | Tests |
|---|---|
| Piped non-TTY stdin (`run -`, implicit stdin prompt, positional + stdin context) | 3 |
| SIGINT delivery → exit 130, empty stdout through a real pipe | 2 |
| Fail-closed sandbox boundary reaching non-interactive run | 1 |

Exit codes 0/1/2 and stdout content through a real pipe remain asserted by these six; `runtime-bootstrap.test.ts` (real background ShellRun lifecycle) and the `pi-tui-runner` signal/fatal probes are intentionally untouched — their subject is the real process boundary.

## One observable-signal fix

The `graph-runtime-error` scenario asserted "graph wait was never called" via a marker the fixture wrote to the real `process.stderr`. In process, that channel is not captured, which would have made the negative assertion vacuous. The marker now travels through the thrown error message (`graph-wait-called: unexpected graph wait after failed invocation`), which the CLI's error path surfaces on the injected stderr — the assertion keeps a live failure signal on both routes.

## Timing

`node --test dist/__tests__/run-command.test.js`, same machine, warm build:

| | tests | wall time |
|---|---|---|
| before | 34 | **15.06s** |
| after | 34 | **3.60s** (−76%) |

[...]

Issue #2387, opened 2026-08-07
perf(test): reduce redundant CLI subprocess startup


Astro-Han:
Part of #2382.

## Problem

CLI, Headless CLI, and harness A/B CLI tests repeatedly start Node processes for scenarios whose semantics can be exercised through existing in-process entry points.

Real subprocess coverage remains necessary for process-level behavior, but not every scenario needs to pay that cost.

## Outcome

- Test ordinary command semantics in process.
- Retain representative subprocess coverage for stdin, stdout, exit codes, signals, persistence, detached execution, and fail-closed behavior.
- Keep CLI and Headless changes in separate PRs.
- Do not change production command behavior.

Each PR must identify the retained process contracts and include comparable before/after timing.

UncertaintyDeterminesYou4ndMe:
I'll take this one, continuing from the #2389 series.

Plan per the issue outcome list:

1. Inventory every test that spawns a Node subprocess across CLI, Headless CLI, and harness A/B CLI suites, and classify each as (a) exercising a genuine process-level contract — stdin wiring, stdout/stderr framing, exit codes, signals, persistence across process death, detached execution, fail-closed — or (b) exercising ordinary command semantics that an existing in-process entry point already covers.
2. Move class (b) scenarios in process; keep representative subprocess coverage for every class (a) contract and name the retained contracts in the PR.
3. Separate PRs for CLI and Headless, each with comparable before/after timing.

No production command behavior changes.

UncertaintyDeterminesYou4ndMe:
Both workspace PRs are up:

- #2476 — CLI: `run-command.test.ts` ordinary semantics through the injectable `MakaRunDeps` seam. 15.06s → 3.60s (−76%), spawns 29 → 6. Retained: piped stdin ×3, SIGINT → 130 ×2, fail-closed sandbox boundary ×1.
- #2479 — Headless: `cli.test.ts` / `contamination-scan-cli.test.ts` / `harness-ab-cli.test.ts` through the exported routers (`runMakaEvalCli`, scan `main(argv)`, launcher `main()`). 30.65s → 19.61s (−36%), spawns 43 → 9. Retained: bin wiring, executable footer + realpath guard, detached launcher / crash journal / credential-order contracts.

[...]

PR #3708, opened 2026-08-24
fix(runtime): add the missing ASF headers to the plugin modules

Astro-Han:
`Check ASF source headers` is failing on `main`, so `main` and every branch built on it currently report a red CI run for a reason unrelated to their own changes.

Five files added by #3250 carry no license header:

```
packages/runtime/src/plugin-kernel.ts
packages/runtime/src/plugin-runtime.ts
packages/runtime/src/plugin-composition-loader.ts
packages/runtime/src/__tests__/plugin-kernel.test.ts
packages/runtime/src/__tests__/plugin-composition-loader.test.ts
```

#3250's own commit was already red for this, so the failure arrived with that merge rather than with anything after it.

This change is the output of `node scripts/asf-license-headers.mjs write` and nothing else — the standard header the script emits, added to those five files, `+95` lines and no logic change. After it, `node scripts/asf-license-headers.mjs check` reports all 2819 covered files clean, which I verified on a clean checkout of current `main`.

@likun666661 — you own #3250, so if you would rather land this yourself, please say so and I will close this. I opened it because `main` being red makes every other PR's CI harder to read, and this seemed worth unblocking quickly rather than waiting.

<details><summary>简体中文</summary>

`main` 上的 `Check ASF source headers` 正在失败，因此 `main` 以及所有基于它的分支，当前都会因为与自身改动无关的原因报出一次红色 CI。

#3250 新增的五个文件没有版权头：

```
packages/runtime/src/plugin-kernel.ts
packages/runtime/src/plugin-runtime.ts
packages/runtime/src/plugin-composition-loader.ts
packages/runtime/src/__tests__/plugin-kernel.test.ts
packages/runtime/src/__tests__/plugin-composition-loader.test.ts
```

#3250 自己那个提交上的 CI 就已经是红的，所以这个失败是随那次合并到来的，不是之后的改动引入的。

本次改动就是 `node scripts/asf-license-headers.mjs write` 的输出，除此之外没有别的内容——把脚本生成的标准版权头加到这五个文件上，`+95` 行，零逻辑改动。之后 `node scripts/asf-license-headers.mjs check` 报告 2819 个受覆盖文件全部通过，我在当前 `main` 的干净检出上验证过。

@likun666661 —— #3250 是你的，如果你更希望自己来修，说一声我就把这个关掉。我开这个 PR 是因为 `main` 红着会让其他所有 PR 的 CI 结果难以判读，觉得值得尽快解开，而不是等。

</details>

PR #3710, opened 2026-08-24
test(cli): restore the run-command process-boundary contracts

UncertaintyDeterminesYou4ndMe:
## Summary

Restores the six process-boundary contracts that were retained on a real subprocess in #2476 and lost when 802855c51 removed the process fixture — as invited in the closing review of #2476 (Refs #2476, refs #2387).

The injected `MakaRunDeps` seam covers the semantics of `stdinIsTTY`/`readStdin`, but not the boundary itself: a real pipe feeding the prompt, a SIGINT delivered by the operating system observed as exit 130 with nothing on stdout, and the fail-closed sandbox path traversing a real child process. Current `main` asserts none of these (`stdinIsTTY`/`readStdin` are exercised in process only; no test anywhere asserts SIGINT → 130; the remaining spawn in `cli.test.ts` covers profile/identity files, not the run path).

- **`run-command-fixture.ts` (new)** — a subprocess entry over today's `runMakaTextCliCore` with **default environment deps** (real stdin, real SIGINT handler, real exit code) and a minimal scripted adapter: four scenarios (`echo`, `sandbox-boundary`, `slow`, `graph-wait`), no env-var assertion matrix, none of the in-process duplication that 802855c51 removed. `fixture-ready` is written only from inside the run (after the core has installed its SIGINT handler), so the test's signal cannot race the default disposition.
- **`run-command.test.ts`** — one `describe('maka run process contract')` block with the six tests, assertions unchanged from the pre-deletion suite:
  | Contract | Tests |
  |---|---|
  | Piped non-TTY stdin (`run -`, implicit stdin prompt, positional + stdin context) | 3 |
  | SIGINT → exit 130, empty stdout through a real pipe (plain run; during Graph completion wait) | 2 |
  | Fail-closed sandbox boundary reaching a non-interactive run | 1 |

  Both signal tests carry bounded SIGKILL guards (never-ready, never-exiting) so a regression fails the suite instead of hanging it.

No production changes.

## Verification

- `packages/cli`: `node --test dist/__tests__/run-command.test.js` — 11/11, three consecutive rounds (the six subprocess tests add ~0.8s total); full CLI suite 442/442.
- Fault injection (in `dist/`, reverted): disabling `if (interrupted) return 130` fails exactly the two SIGINT tests; disabling the implicit stdin read fails exactly the two non-`run -` stdin tests.
- Biome clean on both files; ASF header audit clean; `tsc` clean.
- Not run: workspaces this change does not touch.

## AI use

Select exactly one:

Tool(s) and scope: Claude Code authored the fixture and tests (ported from the pre-deletion suite) and ran the verification; commits carry `Generated-by: Claude Code` trailers. Reviewed and submitted by a human.

## Checklist

- [x] No

### DIFF
--- a/packages/cli/src/__tests__/run-command-fixture.ts
+++ b/packages/cli/src/__tests__/run-command-fixture.ts
@@ -1,0 +1,157 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import type { SessionEvent } from '@maka/core/events';
+import type { SessionSummary } from '@maka/core/session';
+import {
+  runMakaTextCliCore,
+  type MakaRunContext,
+  type MakaRunContextInput,
+  type MakaRunRuntime,
+} from '../run-command-core.js';
+
+// Subprocess entry for the run-command process-contract tests: real piped
+// stdin, SIGINT delivered by the operating system and observed as an exit
+// code, and the fail-closed sandbox boundary reaching a non-interactive run.
+// Ordinary command semantics are covered in process through the same adapter
+// seam — keep this fixture limited to what a real child process is genuinely
+// needed for.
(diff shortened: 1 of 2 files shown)


UncertaintyDeterminesYou4ndMe:
The failing check is `Check ASF source headers`, and the five files it names are all from `d807c0078` (#3250) on current `main` — none are touched by this PR (both files here carry the header; the audit passes locally against this branch's own changes). The in-flight fix is #3708; once it lands I will rebase to re-run CI on a base where the audit is green.

Astro-Han (approved):
Adds a 157-line subprocess entry and four scripted scenarios covering three `maka run` contracts that only hold across a real process boundary. `SIGINT` currently appears nowhere in the repo's tests.

No P-graded issues. I checked why these were removed originally: duplication and fake-only coverage, not flakiness. Fault injection failed in exactly the right cases both times.

Two ungraded notes. `sandbox-boundary` is entirely within the seam and doesn't need a subprocess, and the three stdin scenarios compress to one — the two SIGINT cases carry nearly all the value, and ~160 lines would do. More substantively, this exercises a **synthetic** process boundary: it bypasses production's `resolveMakaCliExitCode` and the exit grace timer, so it can't catch a regression in either.

---

*AI-assisted review: a Claude Code subagent analysed the change; I verified the deletion rationale from history and the fault-injection results myself. No tests run. AI review is not independent human review.*

PR #3711, opened 2026-08-24
fix(desktop): keep transcript lifecycle registry-owned

Astro-Han:
## Summary

- keep transcript consumers registered when a delivery send, acknowledgement, or capacity attempt fails
- make the next delivery reset the existing consumer projection instead of silently detaching it
- keep `RuntimeHostSessionObservationRegistry` as the sole transcript lifecycle authority and remove the Observer duplicate renderer-destruction listener

The production diff is `+4 / -11`. The earlier Renderer close/reopen retry has been removed; this repair adds no retry protocol, recovery state, or second consumer identity.

## Root cause

`RuntimeHostSessionObserver` treated delivery work failure as consumer termination and removed the physical consumer from its indexes. `RuntimeHostSessionObservationRegistry` still retained the renderer logical registration because it owns continuity across Runtime Host replacement. Every later range request therefore passed the Registry and failed in the Observer with `Desktop transcript consumer does not exist`.

The diagnostic report cannot distinguish whether the original detach was triggered by a send failure, acknowledgement timeout, or delivery-capacity protection. All three entered the same silent-detach path.

## Architecture

The lifecycle is now one-way:

- the Registry owns whether the renderer transcript intent exists, including renderer destruction and Host replacement
- the Observer owns bounded replica and delivery work, but a failed delivery is non-terminal; it marks the same consumer for a reset on the next attempt
- the Renderer owns its materialized range and retains only its existing explicit user-facing reload action

This removes the conflicting Observer termination authority. Explicit close, renderer destruction, and Registry shutdown still release the consumer and idle Session normally.

## Verification

- RED: the new Observer regression failed on the old behavior with `Desktop transcript consumer does not exist`
- `node --test apps/desktop/dist/main/__tests__/runtime-host-session-observer.test.js apps/desktop/dist/main/__tests__/desktop-transcript-range-store.test.js` — 51 passed
- `npx biome check apps/desktop/src/main/runtime-host-session-observer.ts apps/desktop/src/main/__tests__/runtime-host-session-observer.test.ts`
- `npm run check:asf-headers`
- real Electron fixture with temporary fault injection: forced a delivery failure, then observed the same consumer ID recover on the next transcript projection; no error toast or ErrorBoundary appeared. The probe, Electron window, fixture Runtime Host, installed Maka process, and residual Runtime Host were removed or stopped afterward.
- full repository tests were not run locally

The previous CI failure was inherited from five missing ASF headers on the old base. #3708 fixed that baseline failure; this branch is rebased onto current green `main` and the new exact-head CI is running.

## AI use

Select exactly one:

Tool(s) and scope: Codex diagnosed the lifecycle mismatch, performed the simplification audit, authored the Observer repair and regression coverage, and ran the focused automated and real-window verification. The commit includes the required `Generated-by: Codex` trailer.

## Checklist

- [ ] No

### DIFF
--- a/apps/desktop/src/main/__tests__/runtime-host-session-observer.test.ts
+++ b/apps/desktop/src/main/__tests__/runtime-host-session-observer.test.ts
@@ -1134,7 +1134,7 @@ test('does not let one backpressured transcript consumer block another', async (
   await observer.close();
 });
 
-test('releases an idle session when transcript delivery fails', async () => {
+test('keeps a transcript consumer available after a delivery fails', async () => {
   const events = new AsyncFrameQueue();
   let closeCount = 0;
   const observer = new RuntimeHostSessionObserver({
@@ -1176,12 +1176,18 @@ test('releases an idle session when transcript delivery fails', async () => {
     },
     emitSessionsChanged() {},
   });
-  let opened = false;
+  let failDelivery = false;
+  let failedDeliveries = 0;
+  let successfulDeliveries = 0;
   const consumerId = 'consumer-failing';
-  await observer.openTranscript('session-1', consumerId, {
+  const opened = await observer.openTranscript('session-1', consumerId, {
     id: 25,
     send(_channel, batch) {
-      if (opened) throw new Error('renderer unavailable');
+      if (failDelivery) {
+        failedDeliveries += 1;
+        throw new Error('renderer unavailable');
+      }
+      successfulDeliveries += 1;
       queueMicrotask(() =>
         observer.acknowledgeTranscript(
           consumerId,
@@ -1194,7 +1200,7 @@ test('releases an idle session when transcript delivery fails', async () => {
     once() {},
     off() {},
   });
-  opened = true;
+  failDelivery = true;
   events.push({
     kind: 'subscription.transcript_advanced',
     hostEpoch: 'host-1',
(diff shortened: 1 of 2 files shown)


Astro-Han (commented):
I reviewed this head and found no blocking issues. No P0-P2.

Checks on `52adc4e` are `test: success`.

<details><summary>简体中文</summary>该头未发现阻断问题。</details>

M4n5ter (approved):
I reviewed exact head `52adc4e70c2ae9020d98a822d1233cf37bf9c715` and found no P0-P3 issues.

The change keeps the renderer's logical transcript registration in `RuntimeHostSessionObservationRegistry` while treating send, acknowledgement-timeout, and delivery-capacity failures as recoverable Observer work. Renderer destruction, explicit close, Host replacement, and Registry shutdown still own the terminal cleanup paths; an initial open failure remains terminal on both layers, so the change does not leave split lifecycle state.

I verified the focused Observer, execution IPC, and Desktop manager suites (81 tests total), the Desktop main TypeScript build, Biome on both changed files, the ASF header audit, and the clean current-main merge tree. The exact-head hosted `test` check is terminal-success.

我审查了精确提交 `52adc4e70c2ae9020d98a822d1233cf37bf9c715`，未发现 P0-P3 问题。

本变更在发送失败、确认超时或投递容量不足时保留 `RuntimeHostSessionObservationRegistry` 中的渲染器逻辑 transcript 注册，只把这些情况作为 Observer 的可恢复工作失败处理。渲染器销毁、显式关闭、Host 替换和 Registry 关闭仍负责终止清理；首次打开失败在两层仍然都是终止状态，因此不会留下分裂的生命周期状态。

我验证了 Observer、执行 IPC 与 Desktop manager 的相关测试（共 81 项）、Desktop main TypeScript 构建、两个变更文件的 Biome、ASF header 审计，以及与当前 main 的无冲突合并树。精确提交上的托管 `test` 检查已终态成功。

</details>

